PR #3, opened 2017-11-21
Expose vue routes

mcstover:


mcstover:
Def a couple ways we could go about processing these. Will revisit if we deem this useful moving forward.

PR #10, opened 2018-01-21
esure a string is passed when no cookies are returned from the reques…

mcstover:
…t header.


server/vue-middleware.js
```diff
@@ -33,7 +33,7 @@ module.exports = function createMiddleware({ serverBundle, clientManifest }) {
 			runInNewContext: false,
 		});
 
-		const cookies = cookie.parse(req.headers.cookie);
+		const cookies = cookie.parse(req.headers.cookies || '');
```
emuvente: Isn't the header name `cookie` not `cookies`?

PR #22, opened 2018-03-06
New colors

emuvente:


emuvente:
The coveralls check has completed already, no problems. Not sure why github is still reporting it as pending. Merging...

PR #25, opened 2018-03-07
Prefetch graphql fragment types

emuvente:


server/vue-middleware.js
```diff
@@ -45,11 +46,20 @@ module.exports = function createMiddleware({ serverBundle, clientManifest }) {
 
 		res.setHeader('Content-Type', 'text/html');
 
-		renderer.renderToString(context).then(html => {
-			res.send(html);
-			if (!isProd) {
-				console.log(`whole request: ${Date.now() - s}ms`);
-			}
-		}).catch(err => handleError(err, req, res));
+		getGqlFragmentTypes(config.server.graphqlUri)
```
mcstover: Seems like this will likely need to be a more generic method as we move forward with setting up / fetching other config. Totally ok for now!

emuvente: agreed

PR #26, opened 2018-03-08
Redirect undefined routes for links wrapped in router-link. Add a TOD…

mcstover:
…O + some scratch code for redirecting to the main kiva 404 page.


mcstover:
The answer on the following page is not a bad idea, moving the redirect to it's own component which is a fallback * route.
https://stackoverflow.com/questions/40193634/vue-router-redirect-on-page-not-found-404

emuvente:
I think what would happen in the solution suggested in that stackoverflow is that it would transition to a blank page, and then start the redirect. I think what we have here with starting the progress bar and then leaving the app before the router does anything works well.

PR #31, opened 2018-03-14
Dependency updates

emuvente:
A bit of dependency cleanup


.babelrc
```diff
@@ -8,11 +8,11 @@
 		}],
 		"stage-2"
 	],
-	"plugins": ["transform-runtime"],
+	"plugins": ["transform-vue-jsx", "transform-runtime"],
```
mcstover: Was this needed for a testing framework?

emuvente: Updates from the vue-cli webpack template that were missed when the transform-vue-jsx was added to the dependencies.

PR #34, opened 2018-03-15
ENG-1788 Implement jest (replacing karma,mocha,chai,sinon)

emuvente:


emuvente:
Coverage decrease is from more accurate coverage reporting from jest. The previous reported coverage from karma was only the coverage of the files that were touched during testing. The coverage reported from jest includes all (most) of /src/

PR #42, opened 2018-03-24
ENG-1941 type styles

emuvente:


src/assets/scss/global/typography.scss
```diff
@@ -0,0 +1,25 @@
+@import 'settings';
+
+h1 {
+	font-weight: bold;
```
mcstover: This can be declared in the header-style for "small" and will apply to h1's on all screen sizes.

emuvente: Actually, the $header-style thing currently only supports font-size, line-height, margin-top, and margin-bottom. I originally had the font-weight in there, but it was not being applied.

mcstover: ahh gotcha...sorry about that. wishful thinking

emuvente: Yeah... though we should still consider moving the header styles somewhere else so they can be edited in one place

BoulderBrains:
Late the the party, but still looking 👀 👍

PR #53, opened 2018-04-13
Firefox header alignment

mcstover:
@emuvente Split these fixes into 2 commits. The icon fix should be good but the header-button width and centering fix was a little more involved.


mcstover:
@emuvente A little clean up and fix up for the search icon button.

These updates sync up the the layout for Safari and Firefox which both had the same issues with the original styles setup for Chrome.

mcstover:
@emuvente btw I'm not sure how css changes dropped coverage... ignoring.

emuvente:
The drop in coverage was b/c you merged in master which had lower coverage. The coveralls check isn't the most accurate...

PR #58, opened 2018-04-17
Add https to prod deps, updating package-lock while in the dev vm see…

mcstover:
…ms to have removed fs-events as it's an optional dep on linux.


mcstover:
It's only needed when running in OSX so we can save a few lines in the package-lock file without it.
Ref: https://stackoverflow.com/questions/40226745/npm-warn-notsup-skipping-optional-dependency-unsupported-platform-for-fsevents

PR #70, opened 2018-04-24
Set cors headers for requests to font directory when serving static f…

mcstover:
…iles.


mcstover:
Based on:
https://expressjs.com/en/starter/static-files.html
https://expressjs.com/en/resources/middleware/serve-static.html

A we could also check against the mime types but would then need multiple checks to cover the various font file types. Ex. application/font-woff is returned for .woff, application/vnd.ms-fontobject for .eot, and image/svg+xml for .svg.

PR #79, opened 2018-04-27
Eng 1979 basic ga snowplow tracking

mcstover:


src/client-entry.js
```diff
@@ -32,6 +32,11 @@ if (window.__APOLLO_STATE__) {
 	apolloClient.cache.restore(window.__APOLLO_STATE__);
 }
 
+// setup global analytics data
+app.$setKvAnalyticsData(window.__INITIAL_STATE__);
```
emuvente: Instead of passing in this global, would it be possible to just use Vue.$store in the plugin itself?

PR #81, opened 2018-04-30
Basic click events based on current values and tracking libs.

mcstover:
This is how I've started out...thoughts?


emuvente:
Looks good! What do you think of skipping the parsing step and just using an array as the directive argument, since it has to be an expression anyway? So instead of
```
v-kv-track-event="`TopNav|click-Lend-Category|${ category.name }|${ category.index }`"
```
it could look like
```
v-kv-track-event="['TopNav', 'click-Lend-Category', category.name, category.index]"
```

PR #94, opened 2018-05-04
ENG-2060 set min-height for large-screen footer

emuvente:


emuvente:
small css fix, so just merging. fyi @mcstover

PR #95, opened 2018-05-07
ENG-2060 ie defaults flex-shrink to 1

emuvente:


src/components/WwwFrame/WwwPage.vue
```diff
@@ -29,6 +29,10 @@ export default {
 	display: flex;
 	flex-flow: column nowrap;
 
+	& > * {
```
mcstover: @emuvente maybe add a comment here about what this does...? Just for future reference

emuvente: I think it might be better as `header, main, footer` anyway

PR #96, opened 2018-05-08
Eng 1944

BoulderBrains:
This gets Raven/sentry hooked up for client and server side. Future refinements will need to be discussed. 

Ran into issues many duplicate errors being fired in sentry when we tried to configured Raven for server in the main.js file. We just implemented in client for now which didn't seem to cause any issues. 

Open questions for discussion: 
- Do we want to push through more useful debugging info?


server/index.js
```diff
@@ -4,10 +4,17 @@ const serverBundle = require('../dist/vue-ssr-server-bundle.json');
...
 const config = require('../config/selectConfig')(argv.config);
+const Raven = require('raven');
 
 const app = express();
 const port = argv.port || config.server.port;
 
+if (config.app.enableSentry) {
+	Raven.config(config.app.sentryURI).install();
+	app.use(Raven.requestHandler());
+	app.use(Raven.errorHandler());
```
emuvente: Express error handlers usually need to be the last thing put into "use", so this line may need to be moved to just before the `app.listen()`

BoulderBrains:
@emuvente Do these changes bring the file more inline with how you've been using Express error handlers? Or would it be better to just bring the whole if block down to the bottom of the file with the app.use calls?

server/index.js
```diff
@@ -33,4 +31,9 @@ app.use(vueMiddleware({
 	config,
 }));
 
+if (config.app.enableSentry) {
+	app.use(Raven.requestHandler());
+	app.use(Raven.errorHandler());
```
mcstover: I think in this case we should only move the errorHandler and test to see if it makes a difference in sending errors automatically.

In our app, both our request and error handling are actually implemented inside the vue-middleware.js file. Theoretically, we would actually need them both declared before that. However we aren't using the standard express onError directly (it may be used in the vue bundle renderer) so moving the errorHandler may have no change.

As a test to see if changing the errorHandler placement has an effect, you could comment out the Manual statement we added in vue middleware and then trigger the server side error to see if it gets sent without us manually capturing the exception.

mcstover: Created ENG-2078 to revisit global error handling and logging in the node app.

BoulderBrains:
Made some updates in here @mcstover & @emuvente. I tested putting the the error handler above the app.listen line at the bottom of server/index.js file, but couldn't get errors to trigger with that configuration. Otherwise, I left some comments, moved some code.

PR #97, opened 2018-05-09
ENG-2069 handle starting session on ui-server

emuvente:


emuvente:
depends on kiva/kiva#6133

PR #106, opened 2018-05-18
ENG-2103 install Helmet for Express

emuvente:


emuvente:
Yeah, fortunately helmet is written with a secure-by-default mindset. You can see which of its modules are turned on by default: https://helmetjs.github.io/docs/
After reviewing the defaults, it didn't seem to be necessary to make any changes for our situation. We will likely come back to this in the future for setting the content security policy.

PR #110, opened 2018-05-22
Remove Sentry/Raven config and usage in server environments. Attempte…

mcstover:
…d to move Raven config for clientside version of Vue app into the client-entry but configuration must be done before boostrapping the app.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -35,8 +35,8 @@ export default function createApp({ apollo = {}, appConfig = {} } = {}) {
 
 	sync(store, router);
 
-	// TODO: Try to get this working without duplicates for server entry.
 	// Checking that sentry is enabled & is not server side
+	// TODO: Enable custom errorHandler for server side to log server errors
 	if (appConfig.enableSentry && typeof window !== 'undefined') {
 		Raven.config(appConfig.sentryURI);
 		Raven.addPlugin(RavenVue, Vue);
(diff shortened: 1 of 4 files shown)


mcstover:
This is the first stage. Please DO NOT delete this branch, I'll keep working against it for other logging tasks.

@emuvente I attempted to move the initialization for Raven into client entry but this process is similar to the Vue.use functionality and must be done before bootstrapping (altering Vue.config.errorHandler in this case) an app instance. https://vuejs.org/v2/api/#Global-Config

Once the app is initialized, Vue.config is was NOT available for alteration.

emuvente:
@mcstover What if we `config.enableSentry = false;` in `server-entry.js`?

src/main.js
```diff
@@ -35,8 +35,8 @@ export default function createApp({ apollo = {}, appConfig = {} } = {}) {
 
 	sync(store, router);
 
-	// TODO: Try to get this working without duplicates for server entry.
 	// Checking that sentry is enabled & is not server side
+	// TODO: Enable custom errorHandler for server side to log server errors
 	if (appConfig.enableSentry && typeof window !== 'undefined') {
```
mcstover: @emuvente We could override the enableSentry value to false and simplify this check. It may be also be useful to set a flag for serverEntry as a more generalized value to check though that would still check 2 items.

mcstover: @emuvente merging for now, both server and client entry get executed on the client in some form so forcing this to false in server actually turns it off for client as well.

PR #122, opened 2018-05-31
vueql-17 Adding leading text span and styles to handle touch & active…

BoulderBrains:
… states


BoulderBrains:
Feel free to review but please don't merge yet, I thought of a couple things last night that I want to verify before merging. Thanks.

PR #129, opened 2018-06-05
Use MemJs instead of Memcached node module. MemJs uses a slightly dif…

mcstover:
…ferernt approach to setting and getting data in that it will return a buffer in some instances that you need to parse into a string.

https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end
https://hackernoon.com/https-medium-com-amanhimself-converting-a-buffer-to-json-and-utf8-strings-in-nodejs-2150b1e3de57

### DIFF
--- a/server/util/initCache.js
+++ b/server/util/initCache.js
@@ -9,7 +9,7 @@ function FakeMemcached(options) {
 	const oldGet = lru.get;
 	lru.get = (key, callback) => {
 		const result = oldGet.call(lru, key);
-		return callback ? callback(result) : result;
+		return callback ? callback((result === undefined ? undefined : null), result) : result;
 	};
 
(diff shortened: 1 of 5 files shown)


mcstover:
Pulling this branch down into a "local" environment to make sure the LRU cache works.

server/util/initCache.js
```diff
@@ -9,7 +9,7 @@ function FakeMemcached(options) {
 	const oldGet = lru.get;
 	lru.get = (key, callback) => {
 		const result = oldGet.call(lru, key);
-		return callback ? callback(result) : result;
+		return callback ? callback((result === undefined ? undefined : null), result) : result;
```
mcstover: put a switch in here so we could differentiate between "undefined" which is what the error state returns and null for no error.

PR #135, opened 2018-06-06
defining text color for IWD promo banner.

BoulderBrains:
Small update to fix the text color for the IWD banner, good catch Matt.


src/components/WwwFrame/PromotionalBanner/Banners/WomensDayBanner.vue
```diff
@@ -11,6 +11,9 @@
 @import 'settings';
 
 .iwd-banner {
+	.content {
+		color: $kiva-darkgreen;
+	}
```
mcstover: was this enough to cover the hover/active/visited states?

BoulderBrains:
Thanks @mcstover, last commit fixes those states as well. These changes appear to be complete now.

PR #136, opened 2018-06-07
Initial content for Refugee Banner within new GenericPromoBanner comp…

mcstover:
…onent.

It seems that in kiva/kiva land we have just been updating the copy for this simple banner. So it doesn't seem we have any flag to flip between different simple banner types. Thus, I've started the GenericPromoBanner component with the thought that maybe we can make the content dynamic...

What other thoughts do you guys have to making this simple to manage?


BoulderBrains:
As discussed with you earlier Matt, it would be great to be able to pass in the 2 pieces of text & the URL via the CMS or settings manager. This would be nice so we don't have a multiple .vue banners with the same HTML structure and different text and call to actions.

mcstover:
@BoulderBrains  I think that dynamic banner content to power both the Ui server and Kiva repo could be a future enhancement.

src/components/WwwFrame/PromotionalBanner/Banners/DefaultPromoBanner.vue
```diff
@@ -0,0 +1,30 @@
...
+	display: block;
+	color: $kiva-accent-green;
+	font-weight: 700;
+
+	&:hover,
+	&:active {
+		color: $kiva-accent-green;
+	}
+
+	@include breakpoint(xlarge down) {
```
mcstover: I think this should probably be moved into PromoBannerSmall if the dark background and border should apply to all the other cases...

emuvente: agreed

BoulderBrains:
FYI: I didn't merge this and delete the branch because tests are still running on the branch.

PR #144, opened 2018-06-14
vue-67 borrower info component built, still needs real data.

BoulderBrains:
First of the loan card components


src/pages/Lend/LendPage.vue
```diff
@@ -6,16 +6,19 @@
...
 				future for themselves and their families.</p>
 			</div>
+			<BorrowerInfo />
 		</div>
 	</www-page>
 </template>
 
 <script>
 import WwwPage from '@/components/WwwFrame/WwwPage';
+import BorrowerInfo from '@/components/LoanCards/BorrowerInfo';
```
jedlin-kiva: @BoulderBrains -Do you think we ought to build at least a first cut loan-card component (just a div with a border around it), and have that be the component that specifies and holds all the loan card sub-components? So LendPage.vue would only import LoanCard.

jedlin-kiva: Oh - I see you already did gridLoanCard

jedlin-kiva: so maybe revise this with that context

BoulderBrains: I believe my second commit has this covered. I've replaced BorrowerInfo with GridLoanCard, then piped the BorrowerInfo through GridLoanCard.

src/pages/Lend/LendPage.vue
```diff
@@ -6,16 +6,19 @@
 				<p class="small-12 xlarge-9">Each Kiva loan helps people build a better
 				future for themselves and their families.</p>
 			</div>
+			<GridLoanCard />
```
jedlin-kiva: this will work for now, but probably we need a wrapping div here that is something like loan-cards-container with a v-for to iterate over the returned loans.

PR #145, opened 2018-06-14
VUE-66 - first cut at loan card image

jedlin-kiva:
Not for merge yet - just for feedback

### DIFF
--- a/src/components/LoanCards/LoanCardImage.vue
+++ b/src/components/LoanCards/LoanCardImage.vue
@@ -1,0 +1,63 @@
+<!--
+	issues:
+
+	* tracking click on image - do here or bubble up?
+		kvtrackevent="Lending|click-Read more|Photo|{{id}}|true"
+
+	* do we need to deal with image editing?
+		{{#is_admin}}
+			<div class="image-edit-buttons">
+				<a class="edit" href="{{image.edit_url}}" target="_blank"><svg class="icon icon-edit"><use xlink:href="#icon-edit"/></svg></a>
+				<a class="refresh" href="#"><svg class="icon icon-refresh"><use xlink:href="#icon-refresh"/></svg></a>
+			</div>
+		{{/is_admin}}
+ -->
+
+
+
+<template>
+	<a href="{{borrowerPageUrl}}">
+		<!-- current implementation has image as background to <a> rather than <img />
+	 			any reason to continue that?
+				also - do we need an alt tag, if so, what should be in it? "photo of {borrower-name}"-->
+		<img class="loan-card-borrower-image"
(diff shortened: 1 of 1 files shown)


src/components/LoanCards/LoanCardImage.vue
```diff
@@ -0,0 +1,63 @@
+<!--
+	issues:
+
+	* tracking click on image - do here or bubble up?
+		kvtrackevent="Lending|click-Read more|Photo|{{id}}|true"
```
emuvente: Doing here makes sense. We have a global directive for supplying that event info (example from TheHeader.vue):
```
v-kv-track-event="'TopNav|click-Portfolio-My public Trustee page'"
```

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -0,0 +1,63 @@
...
+
+	* do we need to deal with image editing?
+		{{#is_admin}}
+			<div class="image-edit-buttons">
+				<a class="edit" href="{{image.edit_url}}" target="_blank"><svg class="icon icon-edit"><use xlink:href="#icon-edit"/></svg></a>
+				<a class="refresh" href="#"><svg class="icon icon-refresh"><use xlink:href="#icon-refresh"/></svg></a>
+			</div>
+		{{/is_admin}}
```
emuvente: I think we should at least file a ticket for this. We can then discuss the priority further. We potentially don't have to worry about this until the logged-in-user loan card case, as you have to be logged in to be an admin.

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -0,0 +1,63 @@
...
+ -->
+
+
+
+<template>
+	<a href="{{borrowerPageUrl}}">
+		<!-- current implementation has image as background to <a> rather than <img />
+	 			any reason to continue that?
+				also - do we need an alt tag, if so, what should be in it? "photo of {borrower-name}"-->
```
emuvente: I think it was a background image to help with image scaling and to have an appropriately sized container while the image is loading. That shouldn't be required for those things though.

+1 to that alt value

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -0,0 +1,63 @@
...
+				default: false
+			}
+        }
+    };
+</script>
+
+<style scoped>
+	.loan-card-borrower-image {
+		width: 100%;
+		max-width: rem-calc(480);  /* necessary? */
```
emuvente: I would imagine that we would want to limit the width of the entire loan card, not just the image.

PR #146, opened 2018-06-18
Vue 70

BoulderBrains:
Matching text functional component in place.

### DIFF
--- a/src/components/LoanCards/GridLoanCard.vue
+++ b/src/components/LoanCards/GridLoanCard.vue
@@ -1,13 +1,13 @@
 <template>
 	<div class="column column-block">
-		<BorrowerInfo
+		<borrower-info
 			:id="loan.id"
 			:name="loan.name"
 			:amount="loan.loanAmount"
 			:use="loan.use"
 			:country="loan.geocode.country.name" />
 
-		<MatchingText :matching-text="loan.matchingText" />
+		<matching-text :matching-text="loan.matchingText" />
 	</div>
 </template>
 
--- a/src/components/LoanCards/GridLoanCard.vue
+++ b/src/components/LoanCards/GridLoanCard.vue
@@ -7,7 +7,7 @@
 			:use="loan.use"
 			:country="loan.geocode.country.name" />
 
-		<MatchingText />
+		<MatchingText :matching-text="loan.matchingText" />
 	</div>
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/MatchingText.vue
```diff
@@ -0,0 +1,13 @@
+<template functional>
+	<div class="matching-text-wrapper" v-if="props.matchingText">
+		<span>2x matching by {{ props.matchingText }}</span>
+	</div>
+</template>
+
+<style lang="scss">
```
jedlin-kiva: It may not be necessary, but I've been marking all my component <style> sections as "scoped". Wonder if we should make this a general practice?

BoulderBrains: Just looked that up @jedlin-kiva, thanks for catching that miss of mine. Seems to make sense to me to add the <style scoped> if it's a functional component like this one.

src/components/LoanCards/GridLoanCard.vue
```diff
@@ -6,15 +6,19 @@
 			:amount="loan.loanAmount"
 			:use="loan.use"
 			:country="loan.geocode.country.name" />
+
+		<MatchingText :matching-text="loan.matchingText" />
```
emuvente: Generally we've been favoring using kebab-case syntax when adding components to the template, as that is more in line with the html spec. In this case: `<matching-text :matching-text="loan.matchingText" />`

src/components/LoanCards/MatchingText.vue
emuvente: "scoped" can have some unexpected behavior, and in general I think it's only useful for pages or components that will only be on the page once.

src/components/LoanCards/GridLoanCard.vue
BoulderBrains: @emuvente changes made to both the borrower-info and matching-text components.

BoulderBrains:
FYI: @jedlin-kiva, @emuvente & @mcstover I'm going to go forward and merge this into the vue-67 (loan card master branch currently) and once we answer the <style scoped> question in a day or two we can come back to BorrowerInfo & MatchingText component files and add scoped if we decide to go that direction.

PR #148, opened 2018-06-19
Vue 69

BoulderBrains:
Here are changes brought together @jedlin-kiva. 

Upon suggestion from Shua, I've removed the wrapper divs from the MatchingText and ActionButton components. 

Please note: The way that I've anchored the .loan-card-footer-wrap to the bottom of the loan card doesn't work across all screen sizes. Tomorrow my first task is going to be implementing this using flex-box instead of the position: absolute; bottom: 0; method that I've implemented here. I was hoping to get around the the flex-box fix tonight, but I'm running out of steam. I just wanted to make sure these changes were accessible for you in the morning to ensure I wasn't blocking you at all.

### DIFF
--- a/src/components/LoanCards/ActionButton.vue
+++ b/src/components/LoanCards/ActionButton.vue
@@ -8,8 +8,14 @@
 @import 'settings';
 
 .action-button-wrapper {
-	margin: 0 rem-calc(20);
 	text-align: center;
+	margin-top: rem-calc(30);
 
+	.button {
+		height: 2.9375rem;
+		line-height: 0.9375rem;
+		font-size: 1.125rem;
+		width: 100%;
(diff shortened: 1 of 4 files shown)


src/components/LoanCards/ActionButton.vue
```diff
@@ -1,21 +1,15 @@
...
+.button {
 	margin-top: rem-calc(30);
-
-	.button {
-		height: 2.9375rem;
-		line-height: 0.9375rem;
-		font-size: 1.125rem;
-		width: 100%;
-	}
+	height: 2.9375rem;
```
jedlin-kiva: wondering why you didn't use rem-calc in this section?

PR #159, opened 2018-06-22
Vue 77

BoulderBrains:
These changes cover the fixes for vue-76 (lowerCasing first letter of loan use if it's not the borrower or group name.)

Changes also cover fix for vue-77 (adding group language to loan used based on borrowerCount of the loan.)

I have an open question out for to @brandonatkiva about the wording of a couple of these changes so don't merge this branch yet please.

### DIFF
--- a/src/components/LoanCards/BorrowerInfo.vue
+++ b/src/components/LoanCards/BorrowerInfo.vue
@@ -42,28 +42,30 @@ export default {
 	},
 	computed: {
 		helpedLanguage() {
-			if (this.status === 'fundRaising' || this.status === 'inactive' || this.status === 'reviewed') {
+			if (this.status === 'fundRaising'
+			|| this.status === 'inactive'
+			|| this.status === 'reviewed') {
 				return 'helps';
-			} return 'helped';
+			}
+			return 'helped';
 		},
 		borrowerCountLanguage() {
 			if (this.borrowerCount > 1) {
 				return ' a member ';
-			} return ' ';
+			}
+			return ' ';
 		},
 		shortenedLoanUse() {
 			const maxLength = 100;
-			const nameLength = this.name.length;
 			const lowerCaseUse = this.use.toString().charAt(0).toLowerCase() + this.use.toString().slice(1);
-			const convertedUse = (this.use.substring(0, nameLength) === this.name) ? this.use : lowerCaseUse;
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -29,17 +30,39 @@ export default {
...
+		borrowerCount: {
+			type: Number,
+			default: 1
 		}
 	},
 	computed: {
+		helpedLanguage() {
+			if (this.status === 'fundRaising' || this.status === 'inactive' || this.status === 'reviewed') {
+				return 'helps';
+			} return 'helped';
```
mcstover: Move this return to it's own line for legibility sake

src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -29,17 +30,39 @@ export default {
...
+		},
+		borrowerCountLanguage() {
+			if (this.borrowerCount > 1) {
+				return ' a member ';
+			} return ' ';
```
mcstover: same move this return to it's own line

src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -29,17 +30,39 @@ export default {
...
+		},
 		shortenedLoanUse() {
 			const maxLength = 100;
+			const nameLength = this.name.length;
+			const lowerCaseUse = this.use.toString().charAt(0).toLowerCase() + this.use.toString().slice(1);
+			const convertedUse = (this.use.substring(0, nameLength) === this.name) ? this.use : lowerCaseUse;
```
mcstover: don't think you need the nameLength var as it's only used 1 time.

src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -29,17 +30,39 @@ export default {
...
+
 			if (this.use.length === 0) {
 				return 'For the borrower\'s privacy, this loan has been made anonymous.';
-			} else if (this.use.length > maxLength) {
-				return `${this.use.substring(0, maxLength)}...`;
 			} else if (this.use.length < maxLength) {
```
mcstover: Currently if a loan use is exactly 100 chars it won't match either of your blocks so nothing will be returned.

You could make this line us <= so that a use of 100 characters would match...or

Another option would be move the > maxLength check block up to where then <  maxLength check is . Then just have an else (no if check needed) block that returns the convertedUse last in the chain...

PR #160, opened 2018-06-22
Update graphql queries and data locations used to assign fetched data…

mcstover:
… to match new api structure.

PENDING merge of https://github.com/kiva/kiva/pull/6316

### DIFF
--- a/src/graphql/query/promotionalBanner.graphql
+++ b/src/graphql/query/promotionalBanner.graphql
@@ -13,22 +13,34 @@ query {
 		}
 		lendingRewardOffered
 	}
-	holiday_enabled: setting(setting: "feature.holiday_mode.enabled") {
-		value
+	holiday_enabled: general {
+		setting(setting: "feature.holiday_mode.enabled") {
+			value
+		}
 	}
-	holiday_start_time: setting(setting: "feature.holiday_mode.start_time") {
-		value
+	holiday_start_time: general {
+		setting(setting: "feature.holiday_mode.start_time") {
+			value
+		}
 	}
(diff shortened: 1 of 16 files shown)


src/graphql/query/wwwHeader.graphql
```diff
@@ -19,6 +19,6 @@ query {
 		}
 	}
 	shop {
-		headerItemCount
+		nonTrivialItemCount
```
emuvente: This name changed, but I don't see it updated in any js?

src/graphql/query/promotionalBanner.graphql
```diff
@@ -13,22 +13,34 @@ query {
...
+		setting(setting: "promo.topnav_promo.start_time") {
+			value
+		}
 	}
-	promo_end_time: setting(setting: "promo.topnav_promo.end_time") {
-		value
+	promo_end_time: general {
+		setting(setting: "promo.topnav_promo.end_time") {
+			value
+		}
```
emuvente: Instead of aliasing the entire 'general' tree repeatedly, the aliases could be nested under 'general'
```
general {
  holiday_enabled: setting(setting: "feature.holiday_mode.enabled") {
    value
  }
  ...
}
```
and then the get in the js becomes `_get(data, 'general.holiday_enabled.value')`

src/graphql/query/wwwHeader.graphql
mcstover: ahh so true

src/graphql/query/promotionalBanner.graphql
mcstover: cool, can do. also found some test stuff I'm fixin  up now

PR #161, opened 2018-06-23
ensure the dark background is not used on xxlarge header and up.

mcstover:


mcstover:
I checked the holiday version and it didn't have this issue.

PR #164, opened 2018-06-25
vue-75 This should do the trick for link text on hover.

BoulderBrains:


BoulderBrains:
This is just preventing the <anchor> text from becoming underlined in the bonus banner when user hovers on desktop or taps on mobile.

PR #168, opened 2018-06-26
adding tracking to borrower-info pieces of loancard.

BoulderBrains:
Adding tracking to borrower name and "read more" text on loan cards.

### DIFF
--- a/src/components/LoanCards/BorrowerInfo.vue
+++ b/src/components/LoanCards/BorrowerInfo.vue
@@ -3,7 +3,7 @@
 		<router-link
 			:to="`/lend/${ id }`"
 			class="name"
-			v-kv-track-event="`['Lending', 'click-Read more', 'Name', ${ id }, 'true']`">
+			v-kv-track-event="['Lending', 'click-Read more', 'Name', id, 'true']">
 			{{ name }}
 		</router-link>
 		<div class="country">{{ country }}</div>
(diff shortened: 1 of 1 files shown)


mcstover:
@BoulderBrains Need to update both of these to be a string literal `` and then use ${id} for that var.

mcstover:
I'm not sure if that will work... Usually the `` wrap the full string like you have in the :to attribute.

How did it render the attribute with this in place?

src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -1,13 +1,22 @@
 <template>
 	<div class="borrower-info-wrapper">
-		<router-link :to="`/lend/${ id }`" class="name">{{ name }}</router-link>
+		<router-link
+			:to="`/lend/${ id }`"
+			class="name"
+			v-kv-track-event="['Lending', 'click-Read more', 'Name', id, 'true']">
```
jedlin-kiva: I think this one should be something like 'click-borrower name' so we can distinguish which elements are getting clicked (I'll change the link around the borrower photo to 'click-borrower photo') for the same reason.

mcstover:
@jedlin-kiva I think it's already using the other parameter to make that distinction.

That said, ultimately, any tracking we add that is being duplicated from existing tracking, should match exactly, the parameters in use so that any reports will function as they do today.

jedlin-kiva:
@mcstover - yeah, I figured that out after I made the comment, so deleted the comment. :0

BoulderBrains:
Thank you both @jedlin-kiva & @mcstover

PR #169, opened 2018-06-28
Vue 82

BoulderBrains:
Putting the whole team on the review for this one. These changes include adding a loan to basket with the header count updated and the "lend" button transitioned to a "checkout now" button.


src/components/LoanCards/Buttons/LendButton.vue
```diff
@@ -0,0 +1,29 @@
...
+		KvButton,
+	},
+	props: {
+		id: {
+			type: Number,
+			default: null
+		},
+		listeners: {
+			type: Number,
+			default: null
```
BoulderBrains: When I didn't have listeners defined in props it was giving me a console error, however I'm not sure I've actually defined this correctly, @mcstover thoughts?

emuvente: It's `$listeners`. All the vue internal variables/functions start with a dollar sign. So you should change the usage in the template to `v-on="$listeners"` and remove the extra prop.

src/components/LoanCards/Buttons/CheckoutNow.vue
```diff
@@ -0,0 +1,58 @@
...
+		},
+	},
+};
+
+</script>
+
+<style lang="scss">
+@import 'settings';
+
+.success.button {
```
emuvente: The styles for this `.success` class look pretty similar to the `.secondary` class styles that KvButton already defines. Could that be used instead? If they are different enough then maybe these should be added into the KvButton component so that they can be used elsewhere.

BoulderBrains: I'll look into this, thanks for the review @emuvente

PR #170, opened 2018-06-28
Vue twenty three favorite icon

jedlin-kiva:


src/components/LoanCards/FavoriteStar.vue
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	.icon-star {
+		height: 2rem;
+		width: 2rem;
+		fill: transparent;
+		color: $white;
+		opacity: 0.5;
+		background-color: $kiva-text-dark;
+		padding: .5rem;
```
emuvente: Stylelint is complaining about this not having a leading zero.

src/components/LoanCards/GridLoanCard.vue
```diff
@@ -88,31 +111,55 @@ export default {
 			}
 			return 'Expiring now!';
 		}
+	},
+	methods: {
+		toggleFavorite() {
+			// optimistically toggle it locally first
+			this.isFavorite = !this.isFavorite;
+
+			// @todo - does this need a .catch? if so, do we log error?
```
emuvente: You would only use `catch` for network errors. Any graphql errors are going to come through the `then` callback as part of the result. It can be accessed through destructuring like `data` is:
```.then({ data, errors }) => {```

File: src/components/LoanCards/GridLoanCard.vue [hunk 3]
```diff
@@ -88,31 +111,55 @@ export default {
...
-		flex-grow: 0;
-		padding: rem-calc(20) rem-calc(20) rem-calc(16);
-		text-align: center;
-		width: 100%;
-	}
+.loan-card-footer-wrap {
+    flex-grow: 0;
+    padding: rem-calc(20) rem-calc(20) rem-calc(16);
+    text-align: center;
+    width: 100%;
```
emuvente: Stylelint is complaining about these being spaces instead of tabs.

src/components/LoanCards/FavoriteStar.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+	<button
+		v-kv-track-event="['Lending','click-Favorite star','Favorite',undefined,true]"
+		@click.prevent.stop='toggleFavorite'
```
mcstover: It may be valuable to pass the loan id in place of undefined here.

mcstover:
Ready to go!

PR #172, opened 2018-06-29
View twenty three lend again button

BoulderBrains:


src/components/LoanCards/Buttons/CheckoutNowButton.vue
```diff
@@ -1,7 +1,8 @@
 <template>
 	<kv-button
 		class="secondary"
-		v-kv-track-event="['Lending', 'click-Read more', 'checkout-now-click', id, 'true']">
+		v-kv-track-event="['Lending', 'click-Read more', 'checkout-now-button-click', id, 'true']"
+		:to="`/basket`">
```
mcstover: Shouldn't need the wrapping `` as there are no dynamic values.

emuvente: since this is just a static string you don't need to use v-bind and can just pass the value:
```
to="/basket"
```

PR #176, opened 2018-07-03
vue-92 adding loanLength|lenderRepaymentTerm to loan cards.

BoulderBrains:
Adding loanLength to loan cards.

### DIFF
--- a/src/components/LoanCards/BorrowerInfo.vue
+++ b/src/components/LoanCards/BorrowerInfo.vue
@@ -17,6 +17,7 @@
 				v-kv-track-event="['Lending', 'click-Read more', 'Read more', id, 'true']">
 				Read more
 			</router-link>
+			<div>{{ loanLength }} months</div>
 		</div>
 	</div>
 </template>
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/BorrowerInfo.vue
```diff
@@ -17,6 +17,7 @@
 				v-kv-track-event="['Lending', 'click-Read more', 'Read more', id, 'true']">
 				Read more
 			</router-link>
+			<div>{{ loanLength }} months</div>
```
mcstover: The comp in Vue-22 shows some bolded copy before the loan term.

BoulderBrains: 👍

BoulderBrains:
I've added in the intro text. Additional work, I've created a universal .strong class in typography.scss file. We have one in the current styleguide so I've added this in for use. If we didn't have the class for a specific reason, that's cool, I'm interested in hearing why and I can change it over and set it under the .loan-use class in the BorrowerInfo.vue file. Otherwise the new .strong class will be there for future use.

PR #178, opened 2018-07-05
Vue 37 global tip message

mcstover:


src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -0,0 +1,171 @@
...
+				this.tipMsg = _get(data, 'tipMsg');
+				this.tipMsgType = _get(data, 'tipMsgType');
+				this.tipVisible = _get(data, 'tipVisible');
+			}
+		});
+	}
+};
+</script>
+
+<style lang="scss">
```
BoulderBrains: Should these styles be "scoped" ?

src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -0,0 +1,171 @@
...
+@import 'settings';
+
+.fade-enter-active,
+.fade-leave-active {
+	transition: opacity 0.5s;
+}
+
```
BoulderBrains: nice fade out!

mcstover:
I'll probably add some documentation in the Mixin File about usage and then reference that in the Component file.

src/plugins/tip-message-mixin.js
```diff
@@ -0,0 +1,55 @@
...
+		/*
+			Show global tip message
+			@param String tipmsg
+			@param Optional String tipMsgType (pass null or 'info' if using tipPersist param)
+			@param Optional Boolean tipPersist
+		*/
+		$showTipMsg(tipMsg, tipMsgType = '', tipPersist = false) {
+			if (!this.$options.inject || !this.$options.inject.apollo) {
+				throw new Error('No apollo client provided! Add "inject: [\'apollo\']" to this component definition.');
+			}
```
emuvente: Since we're using this check twice here and also in the apollo-mixin, maybe it should be moved to some utility file?

src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -0,0 +1,199 @@
...
+		this.apollo.watchQuery({ query: tipMessageData }).subscribe({
+			next: ({ data }) => {
+				// set/update our local data
+				this.tipMsg = _get(data, 'tipMsg');
+				this.tipMsgType = _get(data, 'tipMsgType');
+				this.tipVisible = _get(data, 'tipVisible');
+				this.tipPersist = _get(data, 'tipPersist');
+				this.tipInitUrl = _get(data, 'tipInitUrl');
+			}
+		});
```
emuvente: This could also be done with the `apollo` config option:
```
apollo: {
    query: tipMessageData,
    result({ data }) {
        this.tipMsg = _get(data, 'tipMsg');
        ...
    },
}
```

src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -0,0 +1,199 @@
...
+		closeTip() {
+			this.apollo.mutate({
+				mutation: updateTipMessage,
+				variables: {
+					tipMsg: this.tipMsg,
+					tipMsgType: this.tipMsgType,
+					tipVisible: false,
+					tipPersist: false,
+					tipInitUrl: ''
+				}
```
emuvente: I would think that the only variable you would want to set for this mutation is `tipVisible: false`. Maybe the resolver should be changed so that it merges in variables instead of just replacing them from the query. Or maybe you could use some default values in the graphql query?

BoulderBrains:
Just swinging through to stay updated on these changes 👀

PR #180, opened 2018-07-08
Initial setup for global lightbox in Ui server. Optional white conten…

mcstover:
…t background and title. TODO: optional action buttons and events.


src/components/Kv/KvLightbox.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+	<transition name="fade">
+		<div v-show="isShown" class="kv-lightbox">
+			<div class="row">
+				<div class="columns small=12">
+					<div class="lightbox-content" :class="removeContentBg">
+						<a @click.stop.prevent="closeLightbox" class="close-lightbox" aria-label="Close">
```
emuvente: this should be a `<button>` since it's a clickable that isn't navigating anywhere.

src/components/Kv/KvLightbox.vue
```diff
@@ -0,0 +1,168 @@
...
+	},
+	computed: {
+		removeContentBg() {
+			return this.inverted ? 'inverted' : '';
+		}
+	},
+	// Create and set our internal visibility property
+	watch: {
+		// eslint-disable-next-line object-shorthand
+		'visible'() {
```
emuvente: I think if you do this line as just `visible() {` then  you won't need the eslint-disable

src/components/Kv/KvLightbox.vue
```diff
@@ -0,0 +1,168 @@
...
+			this.isShown = this.visible;
+		}
+	},
+	methods: {
+		closeLightbox() {
+			this.isShown = false;
+			// listen for this event for notice of lightbox being closed internally
+			this.$emit('lightboxClosed');
```
emuvente: Always use kebab-case for event names, b/c camel-case can't be used in templates and event names won't go through the same interpolation that props do. See https://vuejs.org/v2/guide/components-custom-events.html#Event-Names

mcstover:
@emuvente This should be good to review

src/components/Kv/KvLightbox.vue
```diff
@@ -0,0 +1,217 @@
...
+
+.fade-enter-active,
+.fade-leave-active {
+	transition: opacity 0.5s;
+}
+
+.fade-enter,
+.fade-leave-to {
+	opacity: 0;
+}
```
emuvente: it's not a problem here, but these rules are overriding the same rules in TheTipMessage.vue

mcstover: I thought of that as well and think we should probably create a "transitions" scss file we can import that has our global styles for different transitions... timing, tweens, etc...

emuvente: I like that idea

PR #182, opened 2018-07-10
Try event handling first

mcstover:


mcstover:
@emuvente I'd like to try this approach first... The v-show is out in dev now but it has introduced a weird change where the reset state is switched to as the fade is occurring...

PR #185, opened 2018-07-11
VUE-96 action button re-org, use loanId instead of id, add Read more button

emuvente:


src/components/LoanCards/Buttons/ActionButton.vue
```diff
@@ -1,35 +1,20 @@
...
 import CheckoutNowButton from './CheckoutNowButton';
 import LendAgainButton from './LendAgainButton';
-import PlaceholderButton from './PlaceholderButton';
+import ReadMoreButton from './ReadMoreButton';
 
 export default {
-	components: {
-		LendButton,
-		CheckoutNowButton,
-		LendAgainButton,
```
mcstover: Nice that we don't have "assign" these to this component and can just pass them through.

BoulderBrains:
👀

PR #192, opened 2018-07-14
vue-90 fixing action button, now lend25->AddingtoBasket->Checkoutnow

BoulderBrains:


BoulderBrains:
FYI this fixes the issue mentioned in VUE-90 that david passed back to me.

PR #197, opened 2018-07-17
Update lightbox z-indexes to prevent footer links with translate3d fr…

mcstover:
…om showing above the lightbox bg and content.


src/components/Kv/KvLightbox.vue
```diff
@@ -126,6 +126,7 @@ body.scroll-locked {
 	bottom: 0;
 	padding: 4.5rem 0;
 	overflow-y: scroll;
+	z-index: 1500;
```
mcstover: Applying a z-index on this element fixed the issue. Other edits were just to "group" z-indexes for this component.

PR #200, opened 2018-07-18
Feck 126

BoulderBrains:
This code review covers getting a basic version of the login & register forms created in vue. All additional feature for form have been handled by creating individual tickets for those features. The basic layout and field are in place.

### DIFF
--- a/src/assets/scss/global/forms.scss
+++ b/src/assets/scss/global/forms.scss
@@ -41,3 +41,15 @@ input[type="submit"].hidden-submit {
 	width: 1px;
 	height: 1px;
 }
+
+.input-set {
+	label {
+		font-weight: 400;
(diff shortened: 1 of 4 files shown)


mcstover:
@BoulderBrains Can you please go ahead and break these out into LoginForm and RegistrationForm components? There may be some of your styles that can move to the global/forms.scss file and be imported into both that way.

BoulderBrains:
Alright @mcstover, I've done some refactoring, would yo mind taking another look at this for me?

mcstover:
@BoulderBrains We'll have a look at adding some tests to help this coverage drop... I don't believe it's all your code that did that...The graph on coveralls is quite jumpy for this time frame.

PR #204, opened 2018-07-26
Feck 145

BoulderBrains:


src/components/Forms/LoginForm.vue
```diff
@@ -94,7 +107,21 @@ export default {
 	mounted() {
 		this.currUrl = window.location.href;
 	},
+	apollo: {
+		query: SalesforceHelpTextQuery,
+		preFetch: true,
+		result({ data }) {
+			this.salesforceHelpText.name = _get(data, 'general.salesforceSolution.name');
```
mcstover: I think you can simplify this to 1 line, just assigning salesforceHelpText directly as you're operating with an object already. We also know from the query you are only retrieving those 2 specific fields.

PR #211, opened 2018-08-07
LC-23 - add vue-touch-events for row scroling on mobile

jedlin-kiva:


package.json
```diff
@@ -57,6 +57,8 @@
 		"vue-progressbar": "^0.7.3",
 		"vue-router": "^3.0.1",
 		"vue-server-renderer": "^2.5.7",
+		"vue-touch": "^2.0.0-beta.4",
```
emuvente: This should be removed since it didn't work out.

jedlin-kiva: I noticed that. I don't understand why npm uninstall didn't remove it, but have removed manually now.

PR #212, opened 2018-08-07
Feck 140

BoulderBrains:


src/components/Forms/RegisterForm.vue
```diff
@@ -253,7 +254,7 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
 @import 'settings';
```
BoulderBrains: I had to remove scoped from this file's css because I needed to override the height attribute of the password strength meter badge.

emuvente: you can style child components using the `/deep/` selector: https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors

BoulderBrains: 😎 thanks @emuvente

mcstover:
In general, you've done a fine job hooking up this integration...

There is one thing that concerns me quite a bit though which is this peer dependency:
https://github.com/dropbox/zxcvbn

It has a very large file for to check passwords against which will increase the resource load for our "streamlined" checkout if registration is embedded there. Just this include adds 400k gzipped which is 830k parsed. Current recommendations are that bundles should be below 250k.

That file also hasn't seen any updates in quite a while...though it may not be a big deal as most of the easy/stupid password patterns aren't changing...

@BoulderBrains It would be useful here to have a look at the final size of the resources that are loaded in browser when loading that component. You can do this buy running the app in prod mode...or (On your Mac in a local repo) you can run the build and use the report tooling to see the new bundle sizes.

@emuvente  Additional thoughts?

emuvente:
If the component is loaded async and it's the only module with the dependency, then webpack should make it its own chunk that will only be loaded when the form needs to be displayed.

BoulderBrains:
The password meter and field validation is now working correctly on blur events. @mcstover & @emuvente I haven't addressed loading in this library in an async way, but as matt stated above "we will look at optimizing the loading of code when we embed the form in checkout beta"

PR #216, opened 2018-08-08
LC-39 pre-fetch page data

emuvente:


src/graphql/query/loanCardBasketed.graphql
```diff
@@ -11,12 +13,7 @@ query($id: Int!) {
...
 	lend {
 		loan(id: $id) {
-			id
-			loanFundraisingInfo {
-				fundedAmount
-				reservedAmount
-				isExpiringSoon
-			}
+			...loanCardFields
 		}
```
jedlin-kiva: do you actually want the full fragment here, or should the fundrasingInfo be a small fragment of its own?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<www-page class="lend-by-category-page">
 		<div class="heading-region">
-			<h1>Make a loan, change a life</h1>
+			<h1 @click="isAdmin=!isAdmin">Make a loan, change a life</h1>
```
jedlin-kiva: what's this toggle for?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -33,17 +31,23 @@
...
+		return JSON.parse(JSON.parse(_get(data, key)));
+	} catch (e) {
+		return null;
+	}
+}
 
 export default {
 	components: {
-		CategoryAdminControls,
+		CategoryAdminControls: () => import('./admin/CategoryAdminControls'),
```
jedlin-kiva: just curious about the benefit of importing this here rather than above?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -52,13 +56,8 @@ export default {
 	},
 	data() {
 		return {
-			isAdmin: true,
-			isVisitor: true,
-			itemsInBasket: [],
-			loading: false,
-			rows: [],
+			isAdmin: false,
```
jedlin-kiva: won't we need some of these deleted data items for responding to lend-25 events?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -33,17 +31,23 @@
...
-import LoadingOverlay from './LoadingOverlay';
+
+// Parse a SettingsManager value
+function readSetting(data, key) {
+	try {
+		return JSON.parse(JSON.parse(_get(data, key)));
+	} catch (e) {
+		return null;
+	}
+}
```
jedlin-kiva: factor this out to be a utility function?

src/graphql/query/loanCardBasketed.graphql
emuvente: We could do that. I figured we might as well just use the whole fragment so that we don't have to update this every time we change the data fetched for the loan card.

src/pages/Lend/LendByCategoryPage.vue
emuvente: I changed isAdmin to default to false so that the admin tool will not display. But we don't actually check if a user is an admin yet, so I added this hidden toggle as a temporary way to be able to see the tool.

src/pages/Lend/LendByCategoryPage.vue
emuvente: This turns it into an async component, which means it and its dependencies won't be fetched until the component is used in the page. So if a user is not an admin, we avoid even downloading the js for the admin component.

src/pages/Lend/LendByCategoryPage.vue
emuvente: The lend-25 action is entirely contained within the lend button now

src/pages/Lend/LendByCategoryPage.vue
emuvente: My thoughts exactly! I'm planning on pulling that out into a settings manager utility that would also have a function for checking if a feature/experiment is enabled based on start/end time and the enabled flag.

PR #217, opened 2018-08-08
LC-2 lend by category

emuvente:
This is at a stage where we can work with it out of master instead of needing a feature branch, so we wanted to get a spot-check and provide an fyi before we merged.


src/pages/Lend/admin/CategoryAdminControls.vue
```diff
@@ -0,0 +1,176 @@
...
+
+.admin-controls {
+	margin-bottom: 1rem;
+
+	.button .loading-spinner {
+		width: 1.5rem;
+		height: 1.5rem;
+		vertical-align: bottom;
+
+		& /deep/ .line {
```
BoulderBrains: Glad to see I'm not the only one who had to use this /deep/ selector.

src/pages/Lend/admin/CategorySet.vue
```diff
@@ -0,0 +1,151 @@
...
+		stroke: $kiva-text-dark;
+	}
+
+	select {
+		width: auto;
+		margin: 0 0.5rem;
+	}
+}
+
+.sortable-ghost {
```
BoulderBrains: 👻

PR #235, opened 2018-08-21
Set up component demo pages with one for general components and anoth…

mcstover:
…er for components that require kiva app access.


mcstover:
@jedlin-kiva @BoulderBrains This was just to move the components that need kiva app off of the ui-site-map so it is still reachable when in local dev mode...ie no access to the kiva app through graphql.

mcstover:
Also feel free to do other things with this concept of a demo area... Just wanted to make it easy for people to run this repo locally as we talk about it tonight.

BoulderBrains:
👀

PR #236, opened 2018-08-22
Remove rem unit from global line height variables. Forcing the rem un…

mcstover:
…it at this level causes compilation errors when importing the vars globally.


mcstover:
For ref: https://github.com/kiva/nuxt-sandbox/pull/6

PR #239, opened 2018-08-24
Updating the style of the password meter

BoulderBrains:
Made some updates to the vue password meter.


BoulderBrains:
@mcstover after reviewing the meter height this morning, I noticed a bit of a difference in height. I went back in and made the adjustments by setting the rem to exactly the height of the progress meter that's present on the GridLoanCards.

mcstover:
👍

PR #242, opened 2018-08-27
Additional work on payment via paypal

mcstover:


src/components/Checkout/PayPalExpress.vue
```diff
@@ -16,58 +18,119 @@ export default {
...
+									payerId: data.payerID
+								},
+							})
+								.then(ppResponse => {
+									console.log(ppResponse);
+									if (ppResponse.errors) {
+										// Transaction is complete
+										// Currently returns error from findAccountForIpn
+										// "Can not process without account id"
+										console.log(ppResponse);
```
mcstover: @jsaur I was getting an error as mentioned in the comments on the VM. Hopefully it's just that context that fails...so I'm looking to test this out in Dev

src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -53,14 +55,17 @@ export default {
...
+			donations: [],
+			loading: false,
 		};
 	},
 	apollo: {
 		query: initializeCheckout,
-		prefetch: true,
-		result({ data }) {
-			console.log(data);
+		preFetch: true,
```
mcstover: @BoulderBrains I had originally put 'prefect' instead of 'preFetch' so now we're getting the server render...

src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -82,4 +87,9 @@ export default {
 .page-content {
 	padding: 1.625rem 0;
 }
+
+// Hide Basket Bar
+.basket-bar {
+	display: none;
+}
```
mcstover: @BoulderBrains I think this is all it takes to cover FECK-138... will add notes there

src/pages/Checkout/CheckoutBetaPage.vue
BoulderBrains: 👍

src/pages/Checkout/CheckoutBetaPage.vue
BoulderBrains: Nice

src/components/Checkout/PayPalExpress.vue
jsaur: Sounds reasonable, let's see what happens. 👍

PR #244, opened 2018-08-27
add redirect on successful transaction

mcstover:


mcstover:
@BoulderBrains @jsaur FYI this should redirect on successful paypal transaction

PR #247, opened 2018-08-27
LC-12 LC-28 LC-29 LC-47 ability to experiment with which categories are on the page

emuvente:


src/graphql/fragments/loanChannelFields.graphql
```diff
@@ -3,7 +3,6 @@
 fragment loanChannelFields on LoanChannel {
 	id
 	name
-	description
```
jedlin-kiva: I suspect this may come back at some point, but I guess there's no harm in not pulling it until we need it.

PR #248, opened 2018-08-27
Feck 134

BoulderBrains:
Thanks for the hand with this one this afternoon. I added the additional else if to cover the case of <=1min too


BoulderBrains:
I had some git merge issues with this branch, I think I got them all sorted out and the page is loading correctly again. I'll be sure to verify on my VM right before I merge this branch into master.

src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -55,6 +55,7 @@ import LoginForm from '@/components/Forms/LoginForm';
 import RegisterForm from '@/components/Forms/RegisterForm';
 import FacebookLoginRegister from '@/components/Forms/FacebookLoginRegister';
 import BasketItemsList from '@/components/Checkout/BasketItemsList';
+import loanCardQuery from '@/graphql/query/loanCardData.graphql';
```
mcstover: Don't think we need this. Any query adjustments can be made to initializeCheckout which I think you've done already.

src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -63,7 +64,8 @@ export default {
 		LoginForm,
 		RegisterForm,
 		FacebookLoginRegister,
-		BasketItemsList
+		BasketItemsList,
+		loanCardQuery
```
mcstover: same

PR #251, opened 2018-08-28
Feck 133 order totals

mcstover:


src/components/Checkout/OrderTotals.vue
```diff
@@ -0,0 +1,138 @@
...
+<style lang="scss">
+@import 'settings';
+
+.order-totals {
+	margin: $featured-text-line-height $global-margin;
+	text-align: center;
+
+	@include breakpoint(medium) {
+		text-align: right;
+	}
```
BoulderBrains: fancy 🎩🧐

mcstover:
@emuvente @jedlin-kiva @BoulderBrains In this PR, I've created a new icon for the checkout design (small-x) which is a bit more bold than I could adjust our existing X icon to be. You could also make this one more bold by adding a stroke.

I've also added a new color $subtle-gray which is #CCC.

PR #253, opened 2018-08-28
Fix a few loan card click events and convert the rest to the array sy…

mcstover:
…ntax


mcstover:
Hey Guys, fixed up the few issues with Loan card tracking events and converted the rest of the repo to use the array syntax in the v-kv-track-event directive.

Remember you don't need to bind directive with a : as the value passed is a JS expression.

PR #260, opened 2018-08-29
Feck 130

BoulderBrains:


src/components/Checkout/DonationItem.vue
```diff
@@ -0,0 +1,158 @@
...
+			</div>
+		</span>
+		<span class="small-12 medium-4 columns featured-text">
+			<div class="small-12 medium-6 align-center">$
+				<input
+					type="input"
+					class="donation-amount-input"
+					name="donation"
+					id="donation"
+					v-model="amount">
```
mcstover: prob should add some validation at to this input at some point...need to determine what that would be

BoulderBrains: I just created feck-184 to cover the validation rules for this input field.

PR #269, opened 2018-08-31
Initial logic for loan amounts select box. Takes into consideration r…

mcstover:
…eserved loan amounts to prevent invalid options from showing. Also applies an arbitrary limit of 150 for loan shares.


mcstover:
FYI @jsaur 

This process is an additive approach starting with the initial lend amount, considering the reserves available then applying the arbitrary 150 limit.

BoulderBrains:
👍

PR #270, opened 2018-08-31
make sure signin sign up links don't overlap fb link. tweak column cl…

mcstover:
…asses to ensure fb button text isn't cut off


mcstover:
I'll merge this after your code so you don't have to worry about conflicts...

PR #272, opened 2018-08-31
Task to validate graphql queries against an api schema

emuvente:
Currently just setup to compare against the dev api. We should consider what would be the best way to make that configurable. Eventually we should incorporate this into a deploy test.


emuvente:
Yeah, that was the first valid linting error that this task caught!

PR #288, opened 2018-09-11
Moved initialize basket into checkout query folder. Added setting and…

mcstover:
… session call for activeLoginStatus and implemented additional checks in the checkout beta page to ensure the user is actively logged in.


mcstover:
@ksekhri graphql query updates then the usage of the fetched data

PR #298, opened 2018-09-13
Feck 187

BoulderBrains:
I've tagged you in this @mcstover so you can get eyes on it. I'm working on replacing some rem-calc() values with defined values from settings.scss and verifying all these styles are needed. I'll have another more refined commit soon.

### DIFF
--- a/src/assets/scss/_settings.scss
+++ b/src/assets/scss/_settings.scss
@@ -236,6 +236,8 @@ $normal-text-font-size: 1rem;  // because the global "100% isn't useful for all 
 $small-text-font-size: 0.875rem;
 $small-text-line-height: 1.357rem;
 
+$medium-text-font-size: 1.25rem;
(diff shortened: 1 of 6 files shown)


src/assets/scss/_settings.scss
```diff
@@ -236,6 +236,8 @@ $normal-text-font-size: 1rem;  // because the global "100% isn't useful for all
 $small-text-font-size: 0.875rem;
 $small-text-line-height: 1.357rem;
 
+$medium-text-font-size: 1.25rem;
```
mcstover: I have needed this setting SO many times! Thank you!

BoulderBrains: Happy to put this in there, Adam requested a text size in between featured and body text. This font size is right in the middle of those two existing font types.

PR #299, opened 2018-09-13
Feck 177 v2

BoulderBrains:
This code looks really solid @mcstover, thanks for taking a crack at it. Putting the div after the content and then pulling it up to the top of the container makes a lot of sense. Me putting my background div at the top before the content I think was the downfall of my attempt at this ticket. Again, thanks for taking a look at it. This looks good to me to merge.


src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -60,56 +60,60 @@
...
+								@click.prevent.native="validateCreditBasket">Complete order</kv-button>
+						</div>
+
+						<loading-overlay v-if="updatingTotals" class="updating-totals-overlay" />
+					</div>
+
+					<div class="container basket-overlay-bg"></div>
+					<div @click="overlayMouseover"
+						class="basket-overlay-fg">
+						<div class="basket-overlay row align-center align-middle"
```
BoulderBrains: Nice use of the foundation classes

PR #307, opened 2018-09-14
Basket update optimization

mcstover:


src/components/Checkout/KivaCardRedemption.vue
```diff
@@ -59,7 +59,7 @@ export default {
 
 .kiva-card-toggle-wrapper {
 	display: flex;
-    align-items: center;
+	align-items: center;
 }
```
BoulderBrains: Weird, I changed these spaces to a tab character in my last branch I had you review too. 🤔

PR #315, opened 2018-09-18
feck-194 removing dev-vm from this url, updating to production

BoulderBrains:


src/components/Forms/FacebookLoginRegister.vue
```diff
@@ -57,9 +57,9 @@
 							v-model="newAcctTerms"
 							@click="showNewAcctTermsError = validateTerms()">
 						I have read and agree to the Kiva
-						<a href="https://dev-vm-01.kiva.org/legal/terms" target="_blank">Terms of Use</a>
+						<a href="https://www.kiva.org/legal/terms" target="_blank">Terms of Use</a>
 						and
-						<a href="https://dev-vm-01.kiva.org/legal/privacy" target="_blank">Privacy Policy</a>
+						<a href="https://www.kiva.org/legal/privacy" target="_blank">Privacy Policy</a>
```
mcstover: can we make this relative? Removing the domain that is... "/legal/privacy"

BoulderBrains: @mcstover We sure can, see latest commit for fix.

PR #318, opened 2018-09-18
Add clientside check for bonus credit after login

mcstover:
We can only do a redirect from the client at this point so this check will typically fire after the page refreshes due to a login event.


mcstover:
@jsaur is that b/c any availableBonus is ALWAYS applied?

jsaur:
No it’s more that having bonus available is the main trigger, if they do then we have to display different UI components regardless of whether it’s applied or not. By default it’s applied but they can always remove it, the amount available stays the same and the amount applied goes to zero but we still show a link to reapply.

PR #324, opened 2018-09-19
VUE-49 Add country flags back to /portfolio/lending-stats

ksekhri:


src/components/Kv/KvFlag.vue
```diff
@@ -0,0 +1,1038 @@
...
+
+<script>
+export default {
+	props: {
+		country: { type: String, required: true }
+	},
+};
+</script>
+
+<style lang="scss">
```
mcstover: I think the scoped attribute would be appropriate here.

ksekhri: Cool, I just updated the PR. I like that better 👍

src/components/Kv/KvFlag.vue
```diff
@@ -0,0 +1,1038 @@
...
+export default {
+	props: {
+		country: { type: String, required: true }
+	},
+};
+</script>
+
+<style lang="scss" scoped>
+	.flag-icon {
+		background: url('../../assets/images/flags.png') no-repeat;
```
emuvente: We have `@` built-in to our webpack config as a shortcut to the `src` directory, so the url can be written as `url('~@/assets/images/flags.png')` (ex: [fonts.scss](https://github.com/kiva/ui/blob/25cb1aedd602678ab7eddf485e4765b55f7e919c/src/assets/scss/global/fonts.scss#L4)). Doing that will make it so that we don't have to update this line if we move this component around.

ksekhri: Oh awesome, thank you for showing me that 👍 just updated my PR

PR #325, opened 2018-09-19
Redirect on Server and Client if bonusAvailableCredit is greater than 0

mcstover:


src/pages/Checkout/CheckoutBetaPage.vue
```diff
@@ -181,24 +180,29 @@ export default {
...
+		preFetch(config, client) {
+			return client.query({
+				query: initializeCheckout
+			}).then(({ data }) => {
+				const totals = _get(data, 'shop.basket.totals');
+				// check for bonus credit and redirect if present
+				// TODO: remove once bonus credit functionality is added
+				// TODO: bonusAvailableTotal is reporting 0 once the credit has been removed in legacy basket
+				if (parseFloat(totals.bonusAvailableTotal) > 0) {
+					return Promise.reject({ path: '/basket?kexpn=checkout_beta.minimal_checkout&kexpv=a', query: {} });
```
emuvente: The parameters in the path can be moved to the query if you want:
`{ path: '/basket', query: { kexpn: 'checkout_beta.minimal_checkout', kexpv: 'a' } }`
If not you can also drop the empty `query`. This just needs to be a valid [VueRouter Route object](https://router.vuejs.org/api/#route-object-properties).

mcstover: good call...updating now

jsaur:
@mcstover ah you're right. Bonus available is how much bonus is in their basket. Bonus applied is how much of that bonus is actually applied to loans. So when they click remove, their bonus available (in the basket) goes to 0. In the BasketView we also keep track of the bonus amount tied to the user (separate from whether the credit is in their basket). At some point we'll need to do that here as well, and the good thing is it's already in graphql under { my { userAccount { promoBalance } } }. The names of things are a bit confusing right now (ie promoBalance vs bonus) but there is an open task to make all the names consistent.

mcstover:
@jsaur Thanks for the additional insight here, we'll use that when adding the functionality later. I think we're still in good shape here. Once they are kicked out from out page, they remain in the legacy basket. Someone would have to manually go to /checkout-beta to get there and they could checkout, without any bonus credit as there is no way to add it back it. The link in the header to find loans for you bonus credit is still there but I think this is a highly unlikely edge case.

PR #329, opened 2018-09-20
Feck 198

BoulderBrains:
Moving remove loan functionality into a updateLoanAmount if condition.

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -24,11 +24,6 @@
 		</span>
 		<span class="small-3 show-for-small-only"></span>
 		<span class="small-9 medium-3 large-2 loan-res-price-wrapper">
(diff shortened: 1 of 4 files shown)


src/components/Checkout/BasketItem.vue
```diff
@@ -15,15 +15,20 @@
...
 		</span>
 		<span class="small-3 show-for-small-only"></span>
 		<span class="small-9 medium-3 large-2 loan-res-price-wrapper">
-			<loan-reservation
+			<!-- <loan-reservation
 				:is-expiring-soon="loan.loan.loanFundraisingInfo.isExpiringSoon"
 				:is-funded="loan.isFunded"
 				:expiry-time="loan.expiryTime"
-			/>
+			/> -->
```
mcstover: Prob can remove this commented out code, unless you think it will come back here...

PR #331, opened 2018-09-21
VUE-19 - Appeal Banner

horationunez:
Connect to global setting using gql.
Implement design clone from styleguide.

To come next:
- Connection to CMS (must discuss with team)
- Responsive (somewhat related to CMS)
- Analytics


horationunez:
Will review why the build is broken after lunch.

mcstover:
@horationunez 

Try to add the following to your git config to see if you can clear up all of the weird permission changes that are present in this branch. We don't want to manipulate the entire app code permissions at this tim.
`config.FileMode flag = false`

As for your build error it seems quite minor:

src/components/WwwFrame/AppealBanner.vue
 138:11  ✖  Expected single space after ":" with a single-line declaration   declaration-colon-space-after

mcstover:
Builds are passing so that's good, May need to checkout those other remaining permissions affected files from master to see if that is cleared.

horationunez:
@mcstover I did a full reset on the files (664) and folder (775) permissions. Some of the assets showed diffs there since they have +x ... so I just restored them manually. We could other PR for that.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -1,248 +1,252 @@
...
-						:aria-pressed="searchOpen ? 'true' : 'false'"
-						aria-controls="top-nav-search-area"
-						@click="toggleSearch"
-						v-kv-track-event="['TopNav','click-search-close-mobile']"
-					>
-						<kv-icon class="close-icon" name="x" />
-					</button>
-					<search-bar ref="search" />
+	<div>
+		<appeal-banner v-show="showAppealBanner" />
```
mcstover: v-if would better b/c it won't add this when disabled

src/components/WwwFrame/AppealBanner.vue
```diff
@@ -0,0 +1,150 @@
...
+	},
+};
+
+</script>
+<style lang='scss'>
+	@import 'settings';
+	@import 'foundation';
+
+	@include foundation-flex-classes;
+	@include foundation-flex-grid;
```
mcstover: @horationunez you only need to import 'settings' here. 'foundation' along with -flex-classes and -flex-grid are all global and already present.

horationunez:
@mcstover ready for another pass :)

src/components/WwwFrame/TheHeader.vue
```diff
@@ -326,15 +334,15 @@ export default {
...
 	},
 };
 </script>
 
 <style lang="scss">
-@import 'settings';
+@import "settings";
 
-$top-nav-font-size: 1.125rem;
+$top-nav-font-size: 1.125 _rem;
```
mcstover: ooops

src/components/WwwFrame/AppealBanner.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+	<div class="appeal-banner">
+
+		<div class="appeal-banner-layout close" v-show="!isOpen">
```
emuvente: You could use KvExpandable here instead of having two duplicate divs

src/components/WwwFrame/TheHeader.vue
```diff
@@ -1,248 +1,252 @@
...
-						:aria-pressed="searchOpen ? 'true' : 'false'"
-						aria-controls="top-nav-search-area"
-						@click="toggleSearch"
-						v-kv-track-event="['TopNav','click-search-close-mobile']"
-					>
-						<kv-icon class="close-icon" name="x" />
-					</button>
-					<search-bar ref="search" />
+	<div>
+		<appeal-banner v-if="showAppealBanner"/>
```
emuvente: I think it would be cleaner to include the appeal-banner component in the WwwPage component instead of TheHeader. Additionally, this v-if could be placed on the root element of the AppealBanner component, so then the checks for whether it should be displayed can be entirely contained within that component.

PR #339, opened 2018-09-24
Feck 182 fb error handling

mcstover:


src/components/Forms/FacebookLoginRegister.vue
```diff
@@ -190,6 +191,9 @@ export default {
 		clickFbLogin() {
 			this.initiateFbLogin();
 
+			// close any open tip message
+			this.$closeTipMsg();
+
```
BoulderBrains: Nice catch here

PR #341, opened 2018-09-25
Feck 184

BoulderBrains:
This is up and running well on my machine with 1 caveat. 

Donation values of $1000 or more don't update the basket total correctly anymore, I've added a comment in my code with a little bit of an explanation if you want to take a look. I would love to debug this together though, I'm really interested in figuring out what's going on when the donation amount reaches the thousands that different than when the donation amount is in the hundreds.


mcstover:
@BoulderBrains Looking good, you're very close. Check for graphql errors when making a donation over 1000...

PR #347, opened 2018-09-26
VUE-19 Initial UI

horationunez:


mcstover:
Please do a checkout from master on TheHeader.vue & wwwHeader.graphql seeing as no changes are necessary in those files now that you moved things.

PR #359, opened 2018-09-26
Fix up duplicate tags by adding vmid attributes

mcstover:


mcstover:
@ksekhri FYI had to add some more vmid attributes to keep these tags from being duplicated. also had to null out the title template to get that override to work for protocol

PR #366, opened 2018-09-28
expand css linter to enforce specific units

horationunez:


BoulderBrains:
Nice! I like the thought around this change. Let's all be reminded to stay on the same page. 

My only question is, how does this interact with the rem-calc() declarations we have in the /ui repo? I'm curious if rem-calc() meets the stylelint criteria of a "unit" and will require an ignoreProperty to pass over those instances or if it's not a "unit" and therefore not an issue at all. I'll pull this down in the morning and find out.

horationunez:
@BoulderBrains @emuvente ready for another pass.

mcstover:
Thanks @horationunez  that was a good catch.

For a little background a declaration like this (padding-right: 12 _rem;) was not triggering an error...

mcstover:
@BoulderBrains That was a good question, I had no idea how rem-calc would be handled with new restrictions...

horationunez:
@mcstover it flags `12_rem` but not `12 _rem`. This is what I was referring the other day the 12 alone seems to be perceived as pixels an _rem is an undefined token.

PR #367, opened 2018-09-28
Handle errors returned from validate basket method in Kiva Credit and…

mcstover:
… PayPal checkout flows.


mcstover:
Found a US loan, added it to basket, then found that user through admin and tried to lend to it. Also, cleared cookies after logging in on the /checkout-beta page to trigger the login issue.

PR #371, opened 2018-10-01
ENG-2271 Port build pages to Vue

ksekhri:
This is the initial PR for ENG-2271, to test the new pages


src/components/WwwFrame/Menus/CustomSecondaryMenu.vue
```diff
@@ -0,0 +1,216 @@
...
+		menuItems: {
+			type: Array,
+			required: true,
+		}
+	},
+	apollo: {
+		query: customSecondaryMenuQuery,
+		preFetch: true,
+		result({ data }) {
+			this.usingTouch = _get(data, 'usingTouch');
```
horationunez: @ksekhri in case this value isn't provided or is ill formed will it default to the initial value? if this is done implicitly i will prefer to do it explicitly.

ksekhri: I did a test and when the value is missing `_get` returns undefined which implicitly is casted to false. I believe in the situations where this fails, false would work so I've replaced that line with:
```
this.usingTouch = !!_get(data, 'usingTouch');
```

horationunez:
LGTM

src/components/WwwFrame/SecondaryMenu.vue
```diff
@@ -29,12 +29,12 @@
...
 
 				&:hover,
-				&.router-link-active {
+				&.router-link-exact-active {
 					color: $kiva-accent-green;
 					border-bottom: 2px solid $kiva-accent-green;
 				}
 
-				&.router-link-active {
+				&.router-link-exact-active {
```
mcstover: @ksekhri We need to keep both of these general (remove the 'exact') in order to preserve the selected styling for the parent of the selected page, ie the Selected Section. See the secondary nav on /portfolio/lending-stats

src/graphql/query/customSecondaryMenuQuery.graphql
```diff
@@ -0,0 +1,3 @@
+query {
+	usingTouch @client
+}
```
mcstover: Please rename this file to something like "usingTouchClient.graphql" as it's so general we can use it elsewhere.

src/components/WwwFrame/Menus/CustomSecondaryMenu.vue
```diff
@@ -0,0 +1,225 @@
+<template>
```
mcstover: This really gets the secondary menu into a more reusable state. At a later time, there may be a way to use this to cleanup the portfolio secondary stuff too.

src/router/routes.js
```diff
@@ -12,4 +12,6 @@ export default [
 	{ path: '/checkout-vue', component: () => import('@/pages/Checkout/CheckoutPage') },
 	{ path: '/checkout', component: () => import('@/pages/Checkout/CheckoutBetaPage') },
 	{ path: '/protocol', component: () => import('@/pages/Protocol') },
+	{ path: '/build-vue', component: () => import('@/pages/Build/BuildPage') },
+	{ path: '/build-vue/docs', component: () => import('@/pages/Build/DocsPage') },
```
mcstover: I think you can place the docs component route into a 'children' array to the /build-vue object.
https://github.com/vuejs/vue-router/blob/dev/examples/nested-routes/app.js
* Note the addition of the Default component which may or may not be required...have to test that out.

src/graphql/query/customSecondaryMenuQuery.graphql
ksekhri: Agreed, added it to a `shared` folder and named it `usingTouchClient.graphql`

src/components/WwwFrame/SecondaryMenu.vue
mcstover: The more practical approach would probably be to just add &.router-link-exact-active to both groups so this style is applied in all 3 scenarios.

ksekhri: > The more practical approach would probably be to just add &.router-link-exact-active to both groups so this style is applied in all 3 scenarios.

When I tried that the "Developer Home" heading was highlighted even when the user is navigated to the sub page `/build-vue/docs`

I've come up with something new, let me know if this approach looks good to you @mcstover I've let users of the component decide whether they want the matches to be exact or not based on props.

PR #386, opened 2018-10-12
try to validate in prefetch and re-fetch loan data if any errors occur

mcstover:


src/graphql/query/checkout/shopBasketUpdate.graphql
```diff
@@ -3,8 +3,9 @@
 query {
 	my {
 		userAccount {
-			promoBalance
 			id
+			balance
+			promoBalance
 		}
```
BoulderBrains: aesthetically, i appreciate how you lined these up. 🀳

mcstover: lol, yea, I was adding and removing stuff to try and get more info there around the promo balance and just ended up matching it to the other file...

PR #394, opened 2018-10-16
add gaurd for experiment version and assignment

mcstover:
@emuvente Your start and end time guards for experiments work well! And serve as great reminders as to what they are set as lol...

Realized today that the endTime for uiexp.register_fields was 2018-10-14 which was causing a page level 500 error due to the expVersion not being assigned.

Added this guard which will ensure I don't use it if it's absent...could have just checked for undefined but wanted to make sure the variant values were covered too.

### DIFF
--- a/src/components/Forms/RegisterForm.vue
+++ b/src/components/Forms/RegisterForm.vue
@@ -139,6 +139,7 @@
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


mcstover:
@emuvente @BoulderBrains Updated this to use _get as suggested.

PR #406, opened 2018-10-23
Add ability to pass targets object in experiment data. You can target…

mcstover:
… by 'cookied' or 'uncookied' users. Setup basic experiment in Lend by Category in prep for LW 1753

### DIFF
--- a/src/api/localResolvers/experiment.js
+++ b/src/api/localResolvers/experiment.js
@@ -35,6 +35,39 @@ function serializeExpCookie(assignments) {
 }
 
 /**
+ * Cycle through targets object and determine matches
+ *
+ * @param {object} targets
+ * @param {object} cookieStore
+ * @returns {boolean}
+ */
+function matchTargets(targets, cookieStore) {
+	// return true if no targets are set, aka everyone matches!!!
+	if (_isUndefined(targets)) return true;
+
+	// re-start if targets are present
+	let matched = false;
+
+	// User Segment Targets
+	if (!_isUndefined(targets) && !_isUndefined(targets.users)) {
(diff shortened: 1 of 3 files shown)


mcstover:
Setup the following experiment data in Settings manager to test:
Name of Setting: uiexp.lbc_message
Data type: String
Value:
`{"name":"LBCMessage","enabled":true,"startTime":"2018-08-29T00:00","endTime":"2020-10-14T00:00","control":{"key":"control","name":"Control"},"distribution":{"control":0, "shown": 1},"targets":{"users":["cookied"]}}`

I'm open to suggestions or other ways to approach this. The kvu cookie which we use to target rather often is an http only cookie so we can't read it from the client, if it were, I would have pushed this task out longer...

src/api/localResolvers/experiment.js
```diff
@@ -34,6 +34,39 @@ function serializeExpCookie(assignments) {
...
+ * @returns {boolean}
+ */
+function matchTargets(targets, cookieStore) {
+	// return true if no targets are set, aka everyone matches!!!
+	if (_isUndefined(targets)) return true;
+
+	// re-start if targets are present
+	let matched = false;
+
+	// User Segment Targets
```
mcstover: I can simplify some of these nested ifs...and we'll likely want to separate the 'users' checks at some point...
- targets will be defined if we get here...so just check for targets.users
- should set a var for cookieStore.get('kvu')
- combine kvuValue with check for 'cookied' or 'uncookied'

mcstover: Did thEse things...

PR #407, opened 2018-10-23
Add lightbox explanation for forthcoming redirect in the case that a …

mcstover:
…Promo card has been applied or basket validation finds that free credits are in use after logging in.


mcstover:
FYI @ksekhri

PR #409, opened 2018-10-24
Cash 26 basket items validation

mcstover:


mcstover:
Must be merged alongside https://github.com/kiva/kiva/pull/6635

src/plugins/checkout-utils-mixin.js
```diff
@@ -59,26 +64,26 @@ export default {
 		 * @param {Object} errorResponse contains errors node with array of errors
 		 */
 		showCheckoutError(errorResponse) {
-			const errors = _get(errorResponse, 'errors');
```
jsaur: I'm not sure if you're accounting for this already, but when a loan expires or is fully bought out, those error messages will also come through here, and I believe will have the error code: RESERVATION_NO_RESERVATION_FOUND

mcstover: Gonna hold off until after branch cut for these PR's.

Good call on the "extraneous" errors. I also think we may want to suppress the donation removal message that comes along with a loan being removed.

Thanks for having a look, we'll get it out there and start hardening things this upcoming sprint.

PR #411, opened 2018-10-24
Add refreshTotals after validatePreCheckou method. Add check on loan …

mcstover:
…item to trigger validatePreCheckout if it gets into a state where the loan price is 0.00


mcstover:
@ksekhri FYI

PR #421, opened 2018-10-30
Cash 28 basic tooltip component

mcstover:


mcstover:
I've started in on a tooltip component based on KvDropdown using popper.js Any thoughts on this? Kept most of the utility and event stuff you added for Using Touch as it seemed to work well for this case too. The tooltip gets closes on mobile if you start swiping or clicking outside of it.

The design were provided by Adam save for the Caret style which I've improvised. The caret is generated using borders so in order to apply a border on the caret you have to duplicate + stack the markup and styles...like an inner arrow.

FYI @BoulderBrains  @ksekhri

src/components/Kv/KvTooltip.vue
```diff
@@ -0,0 +1,274 @@
...
+	font-weight: 800;
+	margin-bottom: 0.75rem;
+}
+
+.tooltip-arrow {
+	width: 0;
+	height: 0;
+	border-style: solid;
+	position: absolute;
+	margin: 14px;
```
mcstover: The styles for the Caret/Arrow are from the popper site. Suppose I could set a var for rem-calc(14) and reuse it everywhere...

BoulderBrains:
This looks really good Matt. I love the arrow affect along the bottom of the tipmsg that has the thicker drop shadow that spills out into the arrow. Shua's refactor suggestion makes sense too, but look great over all. Has adam/design approved this aqua-blue/green color?

mcstover:
@emuvente Good idea, I'll do some work in that direction. There are some differences with how we apply the show and hide in order to enable the vue transition to work. There are also some differences in the config of the popper but I believe those can be passed down. Will update soon.

mcstover:
@emuvente Here's an initial KvPopper component and an implementation of KvTooltip to use it. The defaults in KvPopper are actually taken from KvDropdown so I think it will convert easily. I'll work on updating KvDropdown to use KvPopper in a separate PR.

PR #425, opened 2018-11-01
Cash 40

BoulderBrains:
Password strength meter reworked.

### DIFF
--- a/src/components/Forms/RegisterForm.vue
+++ b/src/components/Forms/RegisterForm.vue
@@ -71,20 +71,19 @@
 
 			<div class="input-set">
 				<label for="password">Password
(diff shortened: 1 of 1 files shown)


emuvente:
👏

mcstover:
Just for Reference: Prior to removing the vue pw strength meter and the zxcvbn dependency. All "Parsed" code was 2.04mb with zxcvbn accounting for 834k. "Gzipped"  the app was 754kb with zxcvbn accounting for 399kb.

BoulderBrains:
Now: 
"Parsed" code is 1.23mb
"Gzipped" the app is 355kb

Big wins!

PR #429, opened 2018-11-05
Fixing the numeral formatting for kiva card values over 3 digits

BoulderBrains:
Is there an associated ticket for this issue @mcstover ?


BoulderBrains:
FYI @ksekhri

PR #430, opened 2018-11-06
Something about hitting this query in the mounted hooked trickles int…

mcstover:
…o multiple calls to mounted within milliseconds. Paypal scripts are currently initialized on mounted and the duplicated mounted calls were causing duplicate initializations of the button and showing 2 buttons.

### DIFF
--- a/src/pages/Checkout/CheckoutBetaPage.vue
+++ b/src/pages/Checkout/CheckoutBetaPage.vue
@@ -299,6 +299,13 @@ export default {
 			this.preCheckoutStep = 'register';
 			this.switchToRegister();
 		}
+
+		this.holidayModeEnabled = settingEnabled(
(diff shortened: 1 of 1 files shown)


mcstover:
@emuvente @ksekhri @BoulderBrains  Something about calling this query in mounted hooked was causing a "chain" of mounted hooks to be called resulting in a duplicate paypal button. I think it may have to do with the query being an async operation that updated the state...

This was my suggestion to Karan as I thought it would just hit the cache and not be an issue. Moving up to created seems to work fine and it has the benefit of server rendering the banner when active.

@emuvente we'll be moving this basket holiday banner into it's own component soon. Just had get something out for verification as Karan's schedule is restricted this week.

BoulderBrains:
👀 Thanks for the heads up and explanation.

PR #438, opened 2018-11-08
CASH-41

ksekhri:


src/components/Forms/LoginForm.vue
```diff
@@ -50,6 +50,23 @@
 				</p>
 			</div>
 
+			<div class="input-set" v-if="teamName">
+				<label for="team">
+					Request to join team.
+					<a class="helpTip sfHelpTip_old" id="team-info" @click.prevent="triggerTeamLightbox">
```
mcstover: I think we can remove these classes on the <a> these shouldn't be applying any styles as we use kebab lowercase for classes these days and these are likely not defined in the ui styles.

PR #446, opened 2018-11-09
make sure scroll unlock is called when the lightbox is set to invisible

mcstover:


mcstover:
@emuvente @BoulderBrains We found an issue with the Lightbox wherein if you closed it with some action/button in your slot content the internal scrollLock wouldn't be turned off. This seems to fix it up.

PR #448, opened 2018-11-10
use 0 as team id instead of null

mcstover:


mcstover:
Depends on https://github.com/kiva/kiva/pull/6735

PR #451, opened 2018-11-13
Only calculate reserved shares for loans that are not expiring soon. …

mcstover:
…For loans expiring soon we just show remaining shares which are all unreserved


mcstover:
FYI @BoulderBrains @ksekhri @jsaur Fixing un-reserved shares on expiring soon loans. Can't test in VM so pushing to dev.

Standard loans with Reserved shares are looking fine.

PR #457, opened 2018-11-16
Set up basic Algolia Vue Instant search demo with how to on importing…

mcstover:
… components


mcstover:
@coderintherye Whipped this up for a start
@emuvente meant to do this a while back but something came up (aka revenue lol) Change as you wish...

mcstover:
It will also be important to keep any eye on their V2 implementation as it has some significant changes but they are for the better. The lack of SSR in V2 shouldn't really be a problem...

mcstover:
@coderintherye Merge this is you'd like or keep working against it as needed.

PR #471, opened 2018-11-28
CASH-381 Update category rows for 'New countries for you'

foochris:


src/components/LoansByCategory/CategoryRow.vue
```diff
@@ -122,6 +130,11 @@ export default {
 		}
 	},
 	mounted() {
+		// hide 'new countries for you' category if user is not logged in
+		if (this.isLoggedIn === false && this.url.includes('new-countries-for-you') === true) {
+			this.showCategory = false;
+		}
```
emuvente: `showCategory` could be a computed property so that it updates if the user's logged-in state changes

foochris: @emuvente I've updated `showCategory' to be a computed property. Correct?

PR #472, opened 2018-11-28
CASH-373: Fix IE11 Win 7 Bug by using v-show instead of v-if

ksekhri:
Adding @horationunez and @mcstover for their reference


ksekhri:
@BoulderBrains no, sadly. Let's stop supporting IE 11, hahaha.

PR #475, opened 2018-11-30
Travis build matrix

emuvente:
This will make it so that pull requests are only built in node 8 (once). This also makes it so that only the master branch is built in both node 8 and the latest node version. It does this using the [Travis Build Matrix](https://docs.travis-ci.com/user/build-matrix/) to limit the branches that are built when pushed to to only master and any release branches, and only building in the latest node version when the branch is master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 language: node_js
-node_js:
-  - "6"
+matrix:
+  fast_finish: true
+  include:
+    - node_js: '8'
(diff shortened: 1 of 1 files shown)


emuvente:
Note that only one travis job is running for this pull request!

emuvente:
Yeah. Every time a branch with a pull request is pushed to, the pull request job on travis runs again.

PR #479, opened 2018-12-03
ENG-2333 - first commit for build docs 

go-bears:
change string variable to html text


go-bears:
@ksekhri @mcstover added some new text to the developers home page for your review. thanks!

PR #483, opened 2018-12-05
Add kiva algolia configs, prob a better wway to get global config...

mcstover:
@emuvente I'm starting to think we should just add the global config to our store in App.vue so it's available later...open to other suggestions on how to use global config in multiple components.


emuvente:
We could have an apollo local resolver that provides the values. Also, in the future that resolver could provide other methods for interacting with Algolia.

emuvente:
Nice idea with the injected config! That makes a lot of sense.

src/components/LoanCards/AlgoliaLoanCardAdapter.vue
```diff
@@ -0,0 +1,69 @@
...
+			default: () => {}
+		},
+		itemsInBasket: {
+			type: Array,
+			default: () => []
+		},
+	},
+	computed: {
+		adaptedLoan() {
+			// TODO: There must be a better way
```
emuvente: you could do each of these as a separate computed field

src/components/LoanCards/AlgoliaLoanCardAdapter.vue
```diff
@@ -0,0 +1,69 @@
...
+			const defaultImage = `https://www-dev-kiva-org.global.ssl.fastly.net/img/w480h360/${_get(this.loan, 'image.hash')}.jpg`; // eslint-disable-line
+			const retinaImage = `https://www-dev-kiva-org.global.ssl.fastly.net/img/w960h720/${_get(this.loan, 'image.hash')}.jpg`; // eslint-disable-line
+			const exprirationDate = new Date(_get(this.loan, 'plannedExpirationDate') * 1000);
```
emuvente: We have the [date-fns](https://date-fns.org/) library included. It might have something to make this easier.

src/components/LoanCards/AlgoliaLoanCardAdapter.vue
mcstover: agree. I'm a little concerned about what we're losing between this raw loan data the data we get back from graphql...so just hacked something together for now.

src/components/LoanCards/AlgoliaLoanCardAdapter.vue
mcstover: totally would. I think the next step will be to refactor this adapter to serve either a grid loan card or the new list loan card so there is def more work.

PR #484, opened 2018-12-05
Setup Experiment for CASH-101 EXP track loans you might like visibilt…

mcstover:
…y and initial page level tracking


mcstover:
@BoulderBrains @ksekhri You'll need to add the setting "uiexp.checkout_lyml" from settings manager in Dev (set to 0/100) or Prod (set to 50/50) to settings in your vm.

PR #485, opened 2018-12-06
POC for using a dummy loan channel but replacing it's loans with a cu…

mcstover:
…stom query.

### DIFF
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -87,6 +87,7 @@
 import _each from 'lodash/each';
 import _get from 'lodash/get';
 import _map from 'lodash/map';
+import _findIndex from 'lodash/findIndex';
 import { readJSONSetting } from '@/util/settingsUtils';
 import experimentQuery from '@/graphql/query/lendByCategory/experimentAssignment.graphql';
 import lendByCategoryQuery from '@/graphql/query/lendByCategory/lendByCategory.graphql';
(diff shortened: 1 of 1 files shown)


emuvente:
@mcstover Please take a look at my refactor. It's different in that it uses the existing query so that we get all the existing benefits of pre-fetching and reactive updates. One improvement could be to actually get the loan channel name for the custom channels from the query as well.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -242,14 +263,25 @@ export default {
...
-				this.categories = _get(data, 'lend.loanChannelsById') || [];
+				this.realCategories = _get(data, 'lend.loanChannelsById') || [];
+				this.customCategories = [];
+				this.customCategories.push({
+					id: 62,
+					name: 'Custom category',
+					loans: {
+						values: _get(data, 'lend.loans.values') || [],
+					},
+				});
```
mcstover: By pushing this "Structured" custom category data with the id, name and then loans object we should preserve the tracking setup.

mcstover:
@emuvente I think those are nice updates and should cover our needs. I've made a few tweaks to fix up and issue with url. If that field is missing there was a js error. So I added the field when we push to categories and also updated the category row code not to show a link if that field is empty.

Additionally, I wanted to test out setting up 2 custom rows as both @foochris & @ksekhri are working new rows. We'll have to some specific work for each of those to get their "special" vars prepped for use in the loan queries but it shouldn't be an issue.

Anything else you think is crucial to include? Trying to get help these guys get their tickets done this week...

mcstover:
@ksekhri @foochris We've worked up a way to let us use a "dummy" loan channel for each of your custom rows. You'll need to create a loan channel for your new row then update the query and code as shown in this pr.

mcstover:
@emuvente I'm trying to track down a "Mismatching childNodes vs. VNodes" with this current code setup if you have a sec.

PR #490, opened 2018-12-07
Cash 101 slide functionality

BoulderBrains:
Slide functionality part of the way there. loan cards do not stay in a row when screen size is small, they stack instead.


BoulderBrains:
👍 🙏

PR #493, opened 2018-12-07
Add UI server script for status/log/restart

foochris:


BoulderBrains:
👀

mcstover:
@foochris Can you please add some comments to your new script that indicate this is a Kiva server only tool. No one trying to download this repo outside of should think it's for them...

foochris:
@mcstover Good suggestion. Will add to script and readme.

PR #497, opened 2018-12-08
CASH-101: Styling

ksekhri:


src/components/LoansYouMightLike/MinimalLoanCard.vue
```diff
@@ -145,30 +148,58 @@ export default {
...
+	line-height: 1.3;
 }
 
 .card-action {
-	margin-bottom: 15px;
 	font-weight: 400;
+	margin-bottom: 0;
+}
+
+.minimal-loan-card >>> .borrower-image-wrapper .borrower-image {
```
ksekhri: @mcstover not sure how to make this work

src/components/LoansYouMightLike/MinimalLoanCard.vue
```diff
@@ -145,30 +148,58 @@ export default {
...
+	height: auto;
+	object-fit: cover;
+}
+
+.minimal-loan-card >>> .borrower-image-wrapper {
+	padding-top: rem-calc(135);
```
ksekhri: @mcstover same issue

mcstover:
@ksekhri @BoulderBrains I've updated the query to use an existing image size. We should revisit them if this experiment is successful.

PR #501, opened 2018-12-12
Eng 2336 add Code of Conduct

go-bears:
Adding new Code of Conduct vue page
updated routes.js to point to new page
added page to Secondary Menu


go-bears:
@ksekhri are we writing unit tests for new pages or that data on pages exist, like the one we broke when removing API Status?  I'm only seeing unit tests for components and util folders.

src/pages/Build/CodeOfConductPage.vue
```diff
@@ -0,0 +1,136 @@
...
+		DeveloperSecondaryMenu,
+	},
+	metaInfo: {
+		title: 'Code of Conduct'
+	}
+};
+</script>
+
+<style lang="scss">
+@import 'settings';
```
mcstover: You'll only need this if you are using variable in the same style element. Not a biggie.

mcstover:
@go-bears No need for tests around static html content.

src/pages/Build/CodeOfConductPage.vue
go-bears: Sure thing. But it's a good idea to remove Line 131 to keep the file clean?

go-bears:
woot! it was hard to find those 2 darn spaces. lol!

PR #502, opened 2018-12-13
CASH-80: Add donation nudge lightbox and experiment

ksekhri:
Settings Manager:
```
uiexp.donation_nudge_lightbox_custom_tip
String
{"name": "DonationNudgeLightboxCustomTip","enabled": true,"startTime": "2018-08-29T00:00","endTime": "2020-10-14T00:00","control": {"key": "variant-a","name": "Control (variant-a)"},"variants": {"variant-b": {"name": "Custom donation field"}},"distribution": {"variant-a": 1,"variant-b": 0}}
Custom Donation Nudge Experiment
```

### DIFF
--- a/src/components/Checkout/DonationNudgeLightbox.vue
+++ b/src/components/Checkout/DonationNudgeLightbox.vue
@@ -93,11 +93,11 @@ export default {
 					appealIsHorizontallyPadded: false,
 				},
 			];
-			const lowPercentage = {
+			const lowPercentage = [{
 				percentage: 10,
 				appeal: 'Cover some of Kiva\'s costs',
 				appealIsHorizontallyPadded: true,
-			};
+			}];
 			return this.hasCustomDonation ? basePercentageRows : lowPercentage.concat(basePercentageRows);
 		},
 	},


src/components/Checkout/DonationItem.vue
```diff
@@ -32,6 +32,14 @@
 					v-kv-track-event="['basket', 'Donation Info Lightbox', 'Open Lightbox']">
 					How Kiva uses donations
 				</a>
+				<donation-nudge-lightbox
```
mcstover: I don't think this should be rendered unless you have an assignment returned on the donationNudgeLightboxExpVersion variable. If that is null the exp is not enabled so we shouldn't inject the contnet.

src/components/Checkout/DonationNudgeLightbox.vue
```diff
@@ -0,0 +1,175 @@
...
+	.nudge-lightbox-row-padding {
+		padding-left: 3rem;
+		padding-right: 3rem;
+	}
+
+	text-align: center;
+
+	#nudge-donation-top {
+		padding-bottom: 3rem;
+		background: #F8F8F8;
```
mcstover: @BoulderBrains @ksekhri Please keep an eye out for these colors being requested in other new designs... They really should be using colors in the palette from our styleguide rather than breaking consistency to infinitely expand the palette.

If they start showing up in other designs we need to start replacing the values for existing variables so these colors are applied site wide.

src/components/Checkout/DonationItem.vue
mcstover: Probably not an issue seeing as this is the new nudge and replaces the simple input field when loans are in the basket. No action needed.

PR #507, opened 2018-12-14
remove import settings

go-bears:
cleaning up the file to remove import settings. The branch was merged and deleted before I could push the change.


go-bears:
thanks all! 
image

PR #508, opened 2018-12-14
Cash 414 is appeal banner shrunk

mcstover:


mcstover:
@BoulderBrains @ksekhri This update moves a lot around but ultimately gets all of the setting and session checks in place and uses them to show an hide the various content blocks. It will probably be easiest to get this merged into 411 before commencing any other work.

We also hide the appeal on checkout and save the users action of closing or opening the appeal banner back to their site session ( this is dependent upon https://github.com/kiva/kiva/pull/6844 but should just console the error until it's present )

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -1,22 +1,17 @@
 <template>
-	<div class="appeal-banner-wrapper">
+	<div class="appeal-banner-wrapper" v-if="showAppeal">
```
mcstover: @BoulderBrains We had lost the visibility check somewhere along the way. This new variable ensures at least 1 of the Appeal features is turned on and doesn't show if we're on checkout.

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -1,22 +1,17 @@
...
+					<h2>
 						<!-- IF BONUS APPEAL BANNER -->
-						<!-- PUT CONDITIONAL HERE -->
-						<!-- <h2>Donate today and receive a bonus to lend!</h2> -->
-
+						<span v-if="appealBonusEnabled">Donate today and receive a bonus to lend!</span>
+						<!-- IF MATCHED APPEAL BANNER -->
+						<span v-else-if="appealMatchEnabled">Double or triple the impact of your donation!</span>
+						<!-- ELSE STANDARD APPEAL BANNER -->
+						<span v-else>Your donation will power impact and hope in 2019</span>
```
mcstover: It's easiest to put our Default content as the last option in the v-if cascade.

src/graphql/query/appealBanner.graphql
```diff
@@ -1,10 +1,29 @@
 query {
 	general {
-		setting(setting: "feature.endofyear_appeal.enabled") {
+		appeal_enabled: setting(setting: "feature.endofyear_appeal.enabled") {
```
mcstover: Aliases for all settings and checks against their start and end time.

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -101,19 +93,68 @@ export default {
...
 		query: appealBannerQuery,
 		preFetch: true,
 		result({ data }) {
-			this.isVisible = _get(data, 'general.setting.value') === 'true' || false;
+			this.appealEnabled = settingEnabled(
+				data,
+				'general.appeal_enabled.value',
+				'general.appeal_start_time.value',
+				'general.appeal_end_time.value'
+			);
```
mcstover: Check enabled + start and end times.

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -101,19 +93,68 @@ export default {
...
+				'general.appeal_match_enabled.value',
+				'general.appeal_match_start_time.value',
+				'general.appeal_match_end_time.value'
+			);
+
+			// This setting SHOULD be temporary and CANNOT reveal this appeal alone.
+			this.appealBonusEnabled = readBoolSetting(data, 'general.appeal_bonus_active.value');
+
+			// the IsAppealBannerShrunk session value returns either false or '"1"'
+			this.isAppealShrunk = readJSONSetting(data, 'general.appeal_banner_shrunk.data') === 1;
```
mcstover: Check session for appeal shrunk, then watch this value below to close the appeal on load if this is set to true.

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
BoulderBrains: 🆒

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
BoulderBrains: Makes sense 👍

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
BoulderBrains: This makes more sense than what was in there. 🧠Looking at all the setting options.

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -101,19 +93,68 @@ export default {
...
 		},
 	},
+	computed: {
+		showAppeal() {
+			// make sure the appeal is enable + we're not on checkout
+			return (this.appealEnabled || this.appealMatchEnabled) && this.$route.path !== '/checkout';
```
BoulderBrains: Nice or/and statement here to hide it on checkout.

src/graphql/query/appealBanner.graphql
BoulderBrains: Nice, it's great we can pass through and check the group vs all the individual settings.

PR #522, opened 2018-12-17
Ensure value JSON.parse acts upon is not undefined

mcstover:


mcstover:
FYI @BoulderBrains @ksekhri Had to add this default value to _get so that JSON.parse didn't try to parse undefined when you are working outside of the VM context.

BoulderBrains:
👍

PR #524, opened 2018-12-17
Update foundation sites 6.5.x

mcstover:


mcstover:
Foundation docs state that the text helper classes will be moving to a prototype mixin include but it hasn't happened yet. So, I pinned the version until the decided when these changes will occur.

https://foundation.zurb.com/sites/docs/typography-helpers.html

FYI @BoulderBrains @ksekhri

mcstover:
@emuvente I think you did all the work on this when you set it up!

PR #529, opened 2018-12-18
ENG-2350 - add more links

go-bears:
adding more terms of service links
porting overview text to Build landing page


go-bears:
thanks for the 👀 !

PR #531, opened 2018-12-19
CASH-501 Update styles on 'Loan you might like'

foochris:


src/components/LoansYouMightLike/MinimalLoanCard.vue
```diff
@@ -16,7 +16,7 @@
 					:loan-id="loan.id"
 					:name="loan.name"
 					class="small-text name"
-				>{{ loan.name }}</p>
+				>{{ loan.name | truncate(20) }}</p>
```
mcstover: Also, please add a title attribute with the full name to this element.

foochris: @mcstover PR updated with requested changes.

PR #538, opened 2018-12-26
lower v8 memory limit + add new npm command to turn on inspect

mcstover:


mcstover:
@emuvente Ref VUE-141 + 142

PR #539, opened 2018-12-29
update vue and associated libs

mcstover:
There have been an array of fixes to vue (ssr too) that could help us. Been running this locally in prod mode for a few days and things seem good.


BoulderBrains:
👀

PR #540, opened 2018-12-29
Update global mixins to use plugin wrapper

mcstover:


coderintherye:
seems like the right idea, if that mixin issue is part of the memory leak at least then this seems like it would fix it, should get a look from another set of eyes though

BoulderBrains:
👀

BoulderBrains:
👀

PR #544, opened 2019-01-03
Donation controls

mcstover:


mcstover:
Added the optional title prop to KvIcon so we can pass that value in. We may also want to add props to power that aria-label too.

Ref: https://css-tricks.com/accessible-svgs/ + https://stackoverflow.com/questions/4697100/accessibility-recommended-alt-text-convention-for-svg-and-mathml

PR #550, opened 2019-01-05
Vue algolia beta plugin

mcstover:


mcstover:
Here's the conversion to use the V2 Algolia plugins. Feel free to add or update parts of this that move us in the direction of the designs in CASH-321. There is a new page in confluence with more resources: Find "Filter+Sort+with+Algolia+in+Vue" under Ui Server Docs

PR #551, opened 2019-01-07
Limit memory usage by creating one renderer per process instead of pe…

mcstover:
…r request


mcstover:
Spent some time running this in compiled/prod mode and I think we should run in dev for a bit.

PR #552, opened 2019-01-07
App now starts using the nodejs cluster module when starting in produ…

mcstover:
…ction mode.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -1,3 +1,4 @@
+const cluster = require('cluster');
 const express = require('express');
 const helmet = require('helmet');
 const vueMiddleware = require('./vue-middleware');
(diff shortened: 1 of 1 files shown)


mcstover:
@emuvente This would be more of a performance enhancement, allowing the 2 workers to do "stuff" at the same time. Memory usage per process is about the same with the addition of a master process. Here's my tests from the vm.

2cpus in VM
- A master process is spawned that holds between 46-48mb of memory
- 2 child processes are spawned
- - each uses between 80-135mb
- running 2-3 browser sessions

mcstover:
@coderintherye @mikeleach @jmallatt-kiva Gonna put this into dev. It allows the the node process to run on 2 cpus/threads/workers/processes or whatever you want to call them...

PR #556, opened 2019-01-09
Cash 462

BoulderBrains:
This takes care of all the requested changes in cash-462

### DIFF
--- a/src/components/Checkout/OrderTotals.vue
+++ b/src/components/Checkout/OrderTotals.vue
@@ -1,27 +1,25 @@
 <template>
-	<div class="row">
-		<div class="order-totals small-12 text-right">
-			<div v-if="showKivaCredit" class="kiva-credit">
-				<span v-if="showRemoveKivaCredit">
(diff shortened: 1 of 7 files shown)


BoulderBrains:
FYI @ksekhri

src/components/Checkout/OrderTotals.vue
```diff
@@ -1,27 +1,25 @@
 <template>
-	<div class="row">
```
BoulderBrains: Just removed this row, was causing conflict with other elements on page that make it not uniform.

src/components/Checkout/OrderTotals.vue
```diff
@@ -120,7 +118,7 @@ export default {
 	.kiva-credit {
 		font-weight: $global-weight-highlight;
 		margin-bottom: 1rem;
-		font-size: rem-calc(18);
+		font-size: $medium-text-font-size;
```
BoulderBrains: down with custom styles!

PR #558, opened 2019-01-11
AUTH-58 Auth0 login for UI server

emuvente:
These are the base changes to get ui server working using auth0. This works by having the server-side and client-side code each authenticate separately. The client mainly uses KvAuth0.js and Auth0Link.js, and the server mainly uses auth-router.js. This pr includes a flag to enable/disable auth0 per-environment (currently all set to disabled).  

There are a few things that also need to be taken care of before the login will work entirely, including figuring out what to do with features that currently rely on the php session (like the basket and the holiday promo), and what to do in various edge cases like being logged in on the client but not the server, or being logged in so long that the tokens need to be renewed.


emuvente:
Pairs with kiva/kiva#6901

src/util/KvAuth0.js
```diff
@@ -0,0 +1,135 @@
+// These symbols are unique, and therefore are private to this scope.
+// For more details, see https://medium.com/@davidrhyswhite/private-members-in-es6-db1ccd6128a5
```
BoulderBrains: 👍

src/util/KvAuth0.js
```diff
@@ -0,0 +1,135 @@
...
+				// Successful authentication
+				this.user = result.idTokenPayload;
+				this.accessToken = result.accessToken;
+				// handle expiration?
+				resolve(result);
+			}
+		});
+	}
+
+	// Throw an error if this gets called while not in a browser
```
BoulderBrains: In what situation would a user be logging in and not be in a browser? Is this attempting to prevent command line hacking?

emuvente: The is to keep devs from accidentally calling this from server-side code

BoulderBrains: Thanks @emuvente

PR #560, opened 2019-01-14
change footer url link

go-bears:
update developer api link to point to kiva.org/build


src/components/WwwFrame/TheFooter.vue
```diff
@@ -35,7 +35,7 @@
 						<li><router-link to="/gifts">Gifts</router-link></li>
 						<li><router-link to="/live">Happening now</router-link></li>
 						<li><router-link :to="sitemapUrl">Site map</router-link></li>
-						<li><a href="https://build.kiva.org" title="Developer API">Developer API</a></li>
+						<li><a href="https://www.kiva.org/build" title="Developer API">Developer API</a></li>
```
emuvente: Since that page is served by vue now, you should use a router-link here with `to="/build"`

PR #562, opened 2019-01-15
Cash 348 recent loans local storage

mcstover:


mcstover:
@emuvente @BoulderBrains @ksekhri 

Ultimately, I found a lot of places were we should eventually re-factor various pieces of the Grid loan card to enable better reuse of independent pieces of information that may show on a loan card. We have another mini loan card ticket coming up which could be a good time to do that. 

For this experiment I just went with overriding styles to hide stuff but still had to instantiate a bunch of data that was expected in the nested components even though I wasn't showing it. The style overrides mostly worked with regard to hard coded styles in the deepest components so long as I didn't "scope" my parent component but this isn't ideal.

PR #565, opened 2019-01-16
ML-81 Create new GDPR cookie to support additional fields

foochris:


src/components/WwwFrame/CookieBanner.vue
```diff
@@ -48,14 +48,30 @@ export default {
 			this.showBanner = false;
 			this.$kvTrackEvent('global', 'gdpr-notice', 'click-close');
 		},
+		migrateCookie() {
+			const cookieStore = new CookieStore();
```
mcstover: @foochris Could you please move this up above the export default object to create a re-usable instance. Then remove the other instantiations in setGdprCookie + down in mounted?

@emuvente I think that's the right approach...

emuvente: We should provide the cookieStore when creating the vue app, which will allow it to be injected and used here. it's important to use the cookieStore instances created by the entry.js files, because that ensures all the cookies get set appropriately during server render

foochris: @mcstover PR updated.

foochris: @emuvente Should I move

```
const cookieStore = new CookieStore();
```

to client/server-entry.js?

mcstover: @emuvente I think that's probably a good idea but all of this is currently in mounted so we may not need it just yet.

foochris: @emuvente PR updated per your suggestion.

PR #572, opened 2019-01-17
Check for experiment enabled flag even if cookie is already present. …

mcstover:
…Set the version to null in the apollo cache if not enabled

### DIFF
--- a/src/api/localResolvers/experiment.js
+++ b/src/api/localResolvers/experiment.js
@@ -158,7 +158,7 @@ export default ({ cookieStore }) => {
 						id,
 						// if experiment exist & enabled = false return a null version
 						// > we don't want to render a disabled experiment even if a cookie version is present
-						version: (experiment !== null && !experiment.enabled) ? null : version,
+						version: (experiment === null || !experiment.enabled) ? null : version,
 						__typename: 'Experiment',
 					};
 				},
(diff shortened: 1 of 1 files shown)


mcstover:
This will make sure we verify the experiment is enabled even if an assignment is already in the cookie. If it's not enabled we'll return null for the version so it's not shown.

src/api/localResolvers/experiment.js
```diff
@@ -139,11 +139,11 @@ export default ({ cookieStore }) => {
...
 
-					// assign an experiment version if it's currently undefined
-					if (_isUndefined(version)) {
-						// read the experiment data from the cache
-						const experiment = readJSONSetting(context, `cache.data.data['Setting:uiexp.${id}'].value`);
+					// read the experiment data from the cache
+					const experiment = readJSONSetting(context, `cache.data.data['Setting:uiexp.${id}'].value`);
 
+					// assign an experiment version if it's currently undefined
+					if (experiment && _isUndefined(version)) {
```
mcstover: also added this check for cases when the exp may not be in the cache when we look for it.

mcstover:
@emuvente wondering what we could do in here to cover the case where the setting isn't in the cache at all as well...

PR #573, opened 2019-01-17
Make compatible with lend increment button by passing loan in micro l…

mcstover:
…oan card and updating styles.


mcstover:
Gonna merge and get out there to sync this up with the regular grid loan card.

PR #574, opened 2019-01-17
Cash 461 cleanup

mcstover:


mcstover:
FYI @BoulderBrains moving this into release candidate.

PR #578, opened 2019-01-22
Setup limit for rows that are prefetched on the server load with the …

mcstover:
…remaining rows being queried after page load in the client. Moved watcher functions and analytics to happen in sequence after all categories have been loaded.

### DIFF
--- a/src/components/LoansByCategory/CategoryRow.vue
+++ b/src/components/LoansByCategory/CategoryRow.vue
@@ -178,7 +178,8 @@ export default {
 	methods: {
 		saveWindowWidth() {
 			this.windowWidth = window.innerWidth;
-			this.wrapperWidth = this.$refs.innerWrapper.clientWidth;
+			// TODO: New Countries for You code is getting executed even for NON Logged in lenders (no loans, no width)
+			this.wrapperWidth = this.$refs.innerWrapper ? this.$refs.innerWrapper.clientWidth : 0;
 		},
 		scrollRowLeft() {
 			if (this.scrollPos < 0) {
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -257,7 +257,6 @@ export default {
 			}).then(({ data }) => {
 				// add our remaining loan channels
 				this.clientCategories = _get(data, 'lend.loanChannelsById') || [];
-				Promise.resolve();
 			});
 		},
 		fetchRecentlyViewed() {
@@ -303,18 +302,6 @@ export default {
 					});
 				},
 			});
-
-			// React to changes to the category data
-			// TODO: DO we really need this?
-			// > It triggeres an additional full network call to graphql for all loanChannels + their loans
-			// > This could have been part of the long ssr load times when this was in the created method.
-			// categoryObserver.subscribe({
-			// 	next: ({ data }) => {
-			// 		this.realCategories = _merge(this.realCategories, _get(data, 'lend.loanChannelsById')) || [];
-			// 		// If active, update our custom categories on any query change
-			// 		// this.setCustomRowData(data);
-			// 	},
-			// });
 		}
 	},
 	apollo: {
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -94,7 +94,6 @@ import _drop from 'lodash/drop';
 import _each from 'lodash/each';
 import _get from 'lodash/get';
 import _map from 'lodash/map';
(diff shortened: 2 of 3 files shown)


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -225,6 +237,84 @@ export default {
...
+				const recentlyViewed = WebStorage('recentlyViewedLoans');
+				// decode, parse then set recently viewed loan data
+				try {
+					this.recentLoanIds = JSON.parse(atob(recentlyViewed));
+				} catch (e) {
+					// no-op
+				}
+			}
+		},
+		activateWatchers() {
```
mcstover: Moving these out to mounted and after we've fetched all loans eliminated the need to do multiple versions for the prefetch/ssr and client sets of loans.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -225,6 +237,84 @@ export default {
...
+					this.itemsInBasket = _map(_get(data, 'shop.basket.items.values'), 'id');
+					// Update the categories observer with the new setting, triggering updates
+					categoryObserver.setVariables({
+						ids: this.realCategoryIds,
+						// @todo variables for fetching data for custom channels
+					});
+				},
+			});
+
+			// React to changes to the category data
```
mcstover: This seems unnecessary and I couldn't figure out what was enabled with it in place.

I did find this was triggering an additional full query for all loan data that was redundant.

emuvente: You can drop categoryObserver entirely. It enabled admins to immediately see the changes they made to which categories were shown on the page. Definitely not essential, as they can always just refresh the page.

The reason is triggers another network call is because of how the loading of categories is split between the server and the client now. The queries in the apollo cache are keyed based on their arguments as well, so the results of two different api calls using the same query but different arguments will both be cached. This observer is setup ti watch the results of a query for *all* the categories, but now that the prefetching is split, the cache has a query for some categories (from the server) and a second query for some others (from the client). Since the cache doesn't have the results for a single query that fetched all the categories, creating this observer triggers a full api request. 

Since this isn't an important feature, I think we can just get rid of it. If we wanted to continue supporting this feature, we could maybe wait to create the observer until after an admin makes changes.

btw, this wouldn't have affected ssr load times because any promises or observers made in the created() callback (or anywhere in the vue app) never resolve on the server, as the app is rendered synchronously. Prefetching (which is obviously async) only works because it is handled before the app is given to the vue server renderer.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -225,6 +237,84 @@ export default {
...
+				query: loanChannelQuery,
+				variables: {
+					ids: _drop(this.realCategoryIds, ssrRowLimiter),
+					excludeIds: ssrLoanIds,
+					// @todo variables for fetching data for custom channels
+				},
+			}).then(({ data }) => {
+				// add our remaining loan channels
+				this.clientCategories = _get(data, 'lend.loanChannelsById') || [];
+				Promise.resolve();
```
emuvente: was this meant to be `return Promise.resolve();`?

src/pages/Lend/LendByCategoryPage.vue
mcstover: Thanks for the info that makes sense about caching the initial queries and not a single unified query.

Also the ssr statement mostly makes sense... I do wonder if it would still try though.

src/pages/Lend/LendByCategoryPage.vue
mcstover: I just want to make sure this returns and completes the promise... guess I should just 'return someData'...

src/pages/Lend/LendByCategoryPage.vue
mcstover: Had to preserve these watchers to ensure the loans updated after basket update. May be able to whittle them down more...

src/components/LoansByCategory/CategoryRow.vue
```diff
@@ -178,7 +178,8 @@ export default {
 	methods: {
 		saveWindowWidth() {
 			this.windowWidth = window.innerWidth;
-			this.wrapperWidth = this.$refs.innerWrapper.clientWidth;
+			// TODO: New Countries for You code is getting executed even for NON Logged in lenders (no loans, no width)
+			this.wrapperWidth = this.$refs.innerWrapper ? this.$refs.innerWrapper.clientWidth : 0;
```
mcstover: Waiting on future plans for New countries category row... so just introducing a check until that plan is finalized.

PR #580, opened 2019-01-22
Feature auth 72 team joining

ebalcomb:


src/graphql/query/teamInfoFromId.graphql
```diff
@@ -1,7 +1,12 @@
-query($id: Int) {
+query($team_id: Int, $team_recruitment_id: Int) {
```
emuvente: SInce this changes the variable names we should either update the other usage of this query or wait to merge this pr until the checkout page has been updated to use an Auth0 popup instead of an embedded login/register form.

PR #582, opened 2019-01-24
Set proposed SSR row limit on lend by category for testing and verifi…

mcstover:
…cation on dev.


mcstover:
FYI: Actually setting the SSR row limit... @emuvente @BoulderBrains @ksekhri

BoulderBrains:
👀 What does ssr Row stand for?

PR #583, opened 2019-01-25
Cash 566

BoulderBrains:
Hey guys, here is the explore/filters toggle for the /kiva/ui stack. @ksekhri Lets connect on Monday and review our toggles side by side and make adjustments accordingly.


build/webpack.client.base.conf.js
```diff
@@ -35,7 +35,11 @@ module.exports = merge.smart(baseWebpackConfig, {
 				'style': 'display:none;',
 			},
 			svgoOptions: {
-				plugins: [],
+				plugins: [
+					{
+						convertPathData: false,
```
mcstover: @emuvente We found that without this the svgs were being over-optimized to the point of losing and changing portions of the original image. Found this tip (https://github.com/svg/svgo/issues/1074) to fix that along with the options we may want to use in the future (https://github.com/svg/svgo/blob/master/examples/test.js)

src/assets/icons/filters.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcstover: @BoulderBrains You've replaced the exisiting filters.svg file. Need to check that file out from origin/master and put it back in

src/components/LoansByCategory/ViewToggle.vue
```diff
@@ -0,0 +1,88 @@
...
+			class="filter-box">
+			<span class="divider">
+				<kv-icon
+					name="filters-toggle"
+					class="filter-toggle-icon"/>
+				<span class="small-text hide-for-small-only view-text">Filter</span>
+			</span>
+		</router-link>
+	</div>
+</template>
```
mcstover: @BoulderBrains Need to add in the analytics for these links, basic reqs are in 473...
['Lending', 'click-filter-toggle', ##link name##]

build/webpack.client.base.conf.js
emuvente: Which svgs did it affect? How much did the final icon bundle size increase? Another thing that may be responsible for the problems is the floatPrecision below. It might be worth experimenting with changing that to 3 (which is the svgo default), which might result in more favorable final file sizes without affecting the quality.

mcstover: We can go back and see how the file sizes were affected. Unfortunately, this wasn't a precision issue. It was removing entire elements from the svg that was rendered.

src/components/LoansByCategory/ViewToggle.vue
ksekhri: I'm going to use upper case for my `link name`: `Browse`

build/webpack.client.base.conf.js
emuvente: That hadn't happened before with our svgs, so maybe something is different in some of our recent icons that we haven't had to deal with. btw, I think turning off this plugin is fine, I'm just curious about what might have been the issue and how best to optimize our svgs going forward. I've found svgs that have high precision in their numeric values often don't actually need that precision, especially at the size of icons. But sometimes they do, so as we add icons we should keep an eye on the final quality and adjust the floatPrecision if needed.

mcstover: Ok, So turning off "convertPathData" doubles the size of the compiled icons.svg file (85kb to 172kb).

@BoulderBrains I ran a test locally of just updating the precision to 3 as @emuvente suggested and that seemed to preserve the new filter-toggle graphic. There may have been some degradation to the "browse-toggle" graphic but the warping wasn't really noticeable the way it was before.

Could you please try to just change the floatPrecision to 3 (still increases the file size to about 100kb)?

src/components/LoansByCategory/ViewToggle.vue
BoulderBrains: FYI: I went with the following line: 
For the explore/browse icon:
v-kv-track-event="['Lending', 'click-filter-toggle', 'Explore']"
For the filter icon: 
v-kv-track-event="['Lending', 'click-filter-toggle', 'Filter']"

src/components/LoansByCategory/ViewToggle.vue
```diff
@@ -0,0 +1,90 @@
...
+				<kv-icon
+					name="browse-toggle"
+					class="browse-toggle-icon"/>
+				<span class="small-text hide-for-small-only view-text">Explore</span>
+			</span>
+		</router-link>
+		<router-link
+			to="/lend"
+			class="filter-box"
+			v-kv-track-event="['Lending', 'click-filter-toggle', 'Explore']">
```
mcstover: Oops ... should be "Filter"

src/components/LoansByCategory/ViewToggle.vue
```diff
@@ -0,0 +1,90 @@
...
+<script>
+import KvIcon from '@/components/Kv/KvIcon';
+
+export default {
+	components: {
+		KvIcon,
+	},
+	data() {
+		return {
+			isSelected: false,
```
mcstover: Don't think you need this any longer

build/webpack.client.base.conf.js
BoulderBrains: FYI: @emuvente I've gone through, removed the convertPataData and updated the floatPercision to 3. This method preserved the correct shape of the svgs I was working with.

mcstover: @emuvente in the end upping the floatPrecision got us close enough! as mentioned above, the file size will increase a little bitt but not as much as turning off convertPathData did.

PR #584, opened 2019-01-25
prevent load of 3rd party scripts if lender has opted_out

mcstover:
Here's one approach to using the new opted_out param in the kvgdpr cookie.


foochris:
@mcstover Can I merge this PR once https://github.com/kiva/kiva/tree/ML-84 has been merged?

PR #586, opened 2019-01-28
Cash 575 h1 defs

mcstover:


mcstover:
First of many upcoming style updates. Please keep your eyes peeled for anything that goes off the rails . For instance, this changes the font-weight of h1s which you'll certainly notice but that's the intent so it's ok.

PR #591, opened 2019-01-30
Update lend-by-category page designs to sync margins on touch devices…

mcstover:
… along with responsive sizes. Includes new transparent carousel buttons and arrow links

### DIFF
--- a/src/components/LoansByCategory/CategoryRow.vue
+++ b/src/components/LoansByCategory/CategoryRow.vue
@@ -212,20 +212,19 @@ $row-max-width: 63.75rem;
 }
 
 .arrow {
-	display: block;
-    position: absolute;
-    background: rgba(255, 255, 255, 0.8);
-    width: 2.5rem;
-    margin: 0;
-    text-align: center;
-    line-height: 640px;
+	display: flex;
+	position: absolute;
+	background: rgba(255, 255, 255, 0.8);
+	width: 2.5rem;
+	margin: 0;
+	text-align: center;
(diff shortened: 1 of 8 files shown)


src/components/LoansByCategory/ViewToggle.vue
```diff
@@ -39,51 +39,48 @@ export default {
 @import 'settings';
 
 .view-toggle {
```
mcstover: most of the changes below are b/c I nested everything inside of view-toggle, which moved it over. 


Style changes were show in text labels on large up (instead of medium up) and setting the spacing to .75 rem on either side of the dividers when the labels are hidden.

src/components/LoansByCategory/CategoryRow.vue
```diff
@@ -200,15 +200,32 @@ export default {
...
 .cards-and-arrows-wrapper {
+	max-width: $row-max-width;
+	margin: 0 auto;
 	align-items: center;
 	display: flex;
+	position: relative;
+	justify-content: center;
 }
 
 .arrow {
```
mcstover: New html markup and design for arrows which are now semi transparent. The arrows also dictate the offsets for page margins as they had a hard requirement to fall "outside" of the cards on the left. This requirement sets the left position of the first loan card which dictates the title position of each row and requires special margins on the header region.

*Arrow and row styles continue to be duplicated in feature rows for the moment which is current state. I do see a path however for consolidation of the layouts and styles.

PR #595, opened 2019-01-30
Gather any availabled loan ids from local storage for reporting

mcstover:
Moved check for recently loans in local storage up a level so we can report if those in control "could" have been shown their recently viewed loans.


mcstover:
@BoulderBrains yea it's an experiment so along with the control signifier we would also see if they had any loan ids stored.

PR #596, opened 2019-01-31
Fix up loans you might like display when we only have 1 loan of a typ…

mcstover:
…e. Ensure it's there before we try to fetch it. Fix up arrow visibility. Add ability to click loan card image and open in new window. Fix mobile basket bar z-index

### DIFF
--- a/src/components/LoanCards/LoanCardImage.vue
+++ b/src/components/LoanCards/LoanCardImage.vue
@@ -4,6 +4,7 @@
 			:to="`/lend/${loanId}`"
 			v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId, 'true']"
 			class="borrower-image-link"
+			:target="linkTarget"
 		>
 
 			<img class="borrower-image"
@@ -58,6 +59,15 @@ export default {
 		isFavorite: {
 			type: Boolean,
 			default: false
+		},
+		openInNewTab: {
+			type: Boolean,
+			default: false
+		}
+	},
+	computed: {
+		linkTarget() {
+			return this.openInNewTab ? '_blank' : '_self';
 		}
 	}
 };
--- a/src/components/LoansYouMightLike/lymlContainer.vue
+++ b/src/components/LoansYouMightLike/lymlContainer.vue
@@ -80,16 +80,16 @@ export default {
 			return _map(this.loans, 'id');
 		},
 		hasLoansInBasket() {
-			return this.loans || this.loans.length > 0;
+			return this.loans.length || 0;
 		},
 		sameCountry() {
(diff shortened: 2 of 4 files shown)


src/components/LoanCards/LoanCardImage.vue
```diff
@@ -4,6 +4,7 @@
 			:to="`/lend/${loanId}`"
 			v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId, 'true']"
 			class="borrower-image-link"
+			:target="linkTarget"
```
ksekhri: This is cool

mcstover: @emuvente for these loans shown on the basket we don't want to close that page so we open a new tab. An interesting feature for "router-link" but hey...guess we weren't the only ones that needed it.

src/components/WwwFrame/TheBasketBar.vue
```diff
@@ -41,6 +41,7 @@ export default {
 	background-color: $kiva-accent-blue;
 	font-size: rem-calc(20);
 	font-weight: normal;
+	z-index: 1000;
```
mcstover: Recent changes to the category rows carousel introduced z-indexes that placed them above the basket bar. This fixes that.

src/components/LoansYouMightLike/lymlContainer.vue
```diff
@@ -322,17 +325,40 @@ export default {
...
+	// hide arrows if screen is wide enough
+	@media only screen and (min-width: rem-calc(864)) {
+		.lyml-row-wrapper .arrow {
+			visibility: hidden;
+		}
+	}
+}
+
+// hide arrows on mobile
+@media (hover: none) {
```
mcstover: hide on mobile like we do for category rows.

src/components/LoansYouMightLike/lymlContainer.vue
```diff
@@ -192,15 +192,17 @@ export default {
...
 
 				// same Activity loans
-				if (this.hasLoansInBasket && data.data.lend.sameActivity) {
-					loansYouMightLike.push(_get(data.data.lend, 'sameActivity.values[1]'));
+				const sameActivityLoans = _get(data, 'data.lend.sameActivity') || [];
+				if (sameActivityLoans.length > 1) {
+					loansYouMightLike.push(sameActivityLoans.values[1]);
 				} else {
 					loansYouMightLike.push(randomLoans[2]);
 				}
```
BoulderBrains: These if/else statements checking for the data and if it's not there passing through random loans make sense for sure. Thanks for tagging me here on the solution. 👍

PR #601, opened 2019-02-01
Fix category + featured row header font-weight and remove left + righ…

mcstover:
…t margin override on veiw all categories link at the bottom.

### DIFF
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -405,7 +405,11 @@ export default {
 	}
 
 	.loan-category-row {
-		margin: 0 0 rem-calc(40);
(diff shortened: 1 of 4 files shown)


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -409,13 +409,18 @@ export default {
 	}
 
 	.pre-footer {
-		margin: 2rem 0;
+		margin-top: 2rem;
+		margin-bottom: 2rem;
```
mcstover: @ksekhri this is one you found the other day. I didn't realize I had overridden the margin of the default .row styles...

mcstover:
@ksekhri moving on with these tweaks so we can verify in QA

PR #603, opened 2019-02-04
CASH-561 Create view container for filter loan view

foochris:


foochris:
@mcstover 

Let me know if you prefer that:
1. this new view be moved to a new directory `src/components/Lend/`
2. the header for `lend-by-category` be refactored to use this new view

PR #605, opened 2019-02-06
Fix up bottom margin on featured column block to prevent the bottom o…

mcstover:
…f the card from being cut off.

### DIFF
--- a/src/components/LoanCards/GridLoanCard.vue
+++ b/src/components/LoanCards/GridLoanCard.vue
@@ -242,6 +242,16 @@ export default {
 .is-in-featured {
 	flex: 0 0 auto;
 
+	.grid-loan-card {
+		width: rem-calc(280);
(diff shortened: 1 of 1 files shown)


src/components/LoanCards/GridLoanCard.vue
```diff
@@ -250,12 +260,15 @@ export default {
...
 
-	.grid-loan-card {
-		width: rem-calc(280);
-		@include breakpoint(340px down) {
-			min-width: rem-calc(256);
-			width: rem-calc(256);
+.is-in-featured {
+	&.column-block {
+		padding: 0 rem-calc(10);
+		margin-bottom: 2.5rem;
```
mcstover: this was the actual fix. just had to move stuff around in order to only apply that to the is-in-featured class in a way the lint rules allowed.

PR #608, opened 2019-02-08
CASH-578 Implement Cloudinary image enhancenment experiment

foochris:


foochris:
@mcstover / @emuvente The Cloudinary setting that is used is `w_970,h_720,c_fill,g_faces`.

`c_fill`
https://cloudinary.com/documentation/image_transformations#automatic_cropping_with_the_fill_mode

`g_faces`
Cloudinary supports built-in face detection capabilities that allow you to intelligently crop your images. To automatically crop an image so that the detected face(s) is used as the center of the derived picture, set the gravity parameter to one of the following values:

faces - the region of the image that includes all the faces detected (g_faces for URLs).

https://cloudinary.com/documentation/image_transformations#face_detection_based_transformations

emuvente:
Please use all the transformations that we currently apply: https://github.com/kiva/kiva/blob/development/httpdocs/Utils/ImageUtils.php#L735

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -63,13 +63,40 @@ export default {
...
-	}
+	},
+
+	methods: {
+		getImageUrl(imageUrl, transformations) {
+			return (this.imageEnhancementExperimentVersion === 'variant-b' ? this.getCloudinaryUrl(transformations) : imageUrl);  // eslint-disable-line
+		},
+
+		getCloudinaryUrl(transformations) {
+			return `https://res.cloudinary.com/kiva/image/upload/${transformations}/e_viesus_correct/e_sharpen:150/remote/${this.loanImageHash}.jpg`; // eslint-disable-line
```
emuvente: I'm not sure that `/image/upload/` part is needed, see https://github.com/kiva/kiva/blob/development/httpdocs/Utils/ImageUtils.php#L735

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -138,6 +143,10 @@ export default {
 		realCategoryIds() {
 			return _without(this.categoryIds, ...customCategoryIds);
 		},
+		getImageEnhancementExperimentVersion() {
```
emuvente: You don't need all these getters, you can just use `:image-enhancement-experiment-version="imageEnhancementExperimentVersion"` in the template

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -63,13 +63,40 @@ export default {
...
+			type: String,
+			default: ''
+		},
+		loanImageHash: {
+			type: String,
+			default: ''
 		}
 	},
+
 	computed: {
```
emuvente: Computed properties shouldn't be named 'get'; you can just call these imageRetinaUrl and imageStandardUrl

emuvente: Or maybe something like formattedRetinaImageUrl since retinaImageUrl is already a property.

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -63,13 +63,40 @@ export default {
...
+		getImageRetinaUrl() {
+			return this.getImageUrl(this.retinaImageUrl, 'w_970,h_720,c_fill,g_faces');
+		},
+
+		getImageStandardUrl() {
+			return this.getImageUrl(this.standardImageUrl, 'w_480,h_360,c_fill,g_faces');
```
emuvente: Would it possibly make more sense to parse the width/height out of the retinaImageUrl and standardImageUrl props instead of forcing every image to be this size? Or is this done to limit the number of transformations that are performed?

emuvente:
As a side note, this isn't great for caching as this is hitting cloudinary directly instead of going through fastly, but it should be fine as an experiment

src/components/LoanCards/LoanCardImage.vue
foochris: Sure, the parsing could be done. And it's not done to limit the number of transactions since once the transformation is complete, it's cached on Cloudinary's side.

src/components/LoanCards/LoanCardImage.vue
foochris: I've updated the PR. Could you verify the transformation order is correct?

src/components/LoanCards/LoanCardImage.vue
emuvente: I was referring to that by parsing out the size, you could end up with more sizes being requested than just these 2, which would result in more transformations being performed.

emuvente:
👍

PR #609, opened 2019-02-08
CASH-610: Fix location of image on mobile, cleanup code

ksekhri:


src/components/Checkout/DonationNudge/DonationNudgeIntro.vue
```diff
@@ -9,6 +9,14 @@
 		>
 			{{ props.header }}
 		</div>
+		<div
+			v-if="props.mobileImageUrl"
+			id="borrower-image-mobile-container"
+			class="hide-for-large"
+			:class="{'has-description': props.description}"
```
ksekhri: This is here because when the description exists, we need a tiny bit more margin at the bottom

src/components/Checkout/DonationNudge/DonationNudgeIntro.vue
```diff
@@ -9,6 +9,14 @@
 		>
 			{{ props.header }}
 		</div>
+		<div
+			v-if="props.mobileImageUrl"
```
ksekhri: Gated by a v-if the url exists

PR #612, opened 2019-02-09
Cash 407 loan channel cats

mcstover:


src/components/LoansByCategory/CategoryRow.vue
```diff
@@ -3,14 +3,24 @@
 		<div class="row title-row">
 			<div class="column small-12">
 				<h2 class="category-name">
-					<a class="view-all-link"
+					<router-link
```
mcstover: converted to router-link so we can hop to ui page without refresh!

BoulderBrains: Nice

src/components/LoansByCategory/CategoryRow.vue
```diff
@@ -3,14 +3,24 @@
...
+							'Lending',
+							'click-Category-View-All',
+							`View all ${cleanName} loans`]">
+						{{ cleanName }} <span v-if="showViewAllLink" class="view-all-arrow">&rsaquo;</span>
+					</router-link>
+					<!-- <a class="view-all-link"
 						:href="cleanUrl"
 						:title="`View all ${cleanName} loans`"
 						v-kv-track-event="[
 							'Lending',
```
BoulderBrains: Should we just delete this commented out code since it's been changed from an <a> to a <router-link>?

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -0,0 +1,293 @@
...
+		// 	variables: _merge(
+		// 		this.loanQueryVars,
+		// 		fromUrlParams(this.pageQuery)
+		// 	),
+		// 	fetchPolicy: 'network-only',
+		// }).subscribe({
+		// 	next: ({ data }) => {
+		// 		this.loanChannel = _get(data, 'lend.loanChannelsById[0]');
+		// 	},
+		// });
```
mcstover: This watchQuery is how things were done in the original lend-vue page. However, I couldn't get that work with these queries so I moved to an actual "watch" setup against the query vars below. @emuvente I'm open to revisiting this if there is a better way. Otherwise, I'll remove after you verify.

src/components/LoansByCategory/ViewToggle.vue
```diff
@@ -32,6 +32,16 @@ export default {
...
+	props: {
+		browseUrl: {
+			type: String,
+			default: '/lend-by-category'
+		},
+		filterUrl: {
+			type: String,
+			default: '/lend'
+		},
+	}
```
mcstover: new props so we can customize the url used in the toggle based on page context.

src/components/LoansByCategory/CategoryRow.vue
mcstover: yes. sorry I missed that

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -0,0 +1,293 @@
...
+				});
+			}).then(() => {
+				// TODO: REMOVE Once Lend Increment Button EXP ENDS
+				// Pre-fetch the setting for lend increment button
+				return client.query({ query: experimentSetting, variables: { key: 'uiexp.lend_increment_button' } });
+			}).then(() => {
+				// TODO: REMOVE Once Lend Increment Button EXP ENDS
+				// Pre-fetch the assigned version for lend increment button
+				return client.query({ query: experimentQuery, variables: { id: 'lend_increment_button' } });
+			});
```
emuvente: Since these experiment queries don't depend on the results of the other queries at all, you could actually run them in parallel with the first query by using Promise.all

mcstover: ooh good idea!

PR #616, opened 2019-02-11
CASH-452: Instantiate Algolia List View

ksekhri:


src/pages/Lend/LendFilterPage.vue
```diff
@@ -1,81 +1,132 @@
...
+				return Promise.all(cash103);
+			});
+		}
 	},
 };
 </script>
 
 <style lang="scss" scoped>
 @import 'settings';
-@import 'global/transitions';
```
ksekhri: @mcstover removed `global/transitions` import, were we using this in this page?

mcstover: yea that's not needed at this point.

src/pages/Lend/LendFilterPage.vue
```diff
@@ -1,81 +1,132 @@
...
+	inject: [
+		'apollo',
+		'algoliaConfig'
+	],
+	apollo: {
+		preFetch(config, client) {
+			return client.query({
+				query: itemsInBasketQuery,
+			}).then(({ data }) => {
+				this.itemsInBasket = _map(_get(data, 'shop.basket.items.values'), 'id');
```
mcstover: When using the prefetch function you're not in the context of the component so you don't need this here.

In the created hook you'll read from your basket items query and then set this data, There will also need to be a watch query at some point but we can wait on that part and deal with it alongside the adapter component work.

src/pages/Lend/LendFilterPage.vue
```diff
@@ -1,81 +1,132 @@
...
+			}).then(() => {
+				// TODO: REMOVE Once CASH-103: Lend Increment Button EXP ENDS
+				const cash103 = [
+					// Pre-fetch the setting for lend increment button
+					client.query({ query: experimentSetting, variables: { key: 'uiexp.lend_increment_button' } }),
+					// Pre-fetch the assigned version for lend increment button
+					client.query({ query: experimentQuery, variables: { id: 'lend_increment_button' } }),
+				];
```
mcstover: You can't actually chain these under promise all. The 2nd one requires the 1st so you'll just have to use .then for both. I tried to do this earlier and the experiment assignment would never occur.

ksekhri: Yes! Good catch hahaha.

src/pages/Lend/LendFilterPage.vue
```diff
@@ -1,81 +1,130 @@
...
+	apollo: {
+		preFetch(config, client) {
+			return client.query({
+				query: itemsInBasketQuery,
+			}).then(() => {
+				// TODO: REMOVE Once CASH-103: Lend Increment Button EXP ENDS
+				return client.query({ query: experimentSetting, variables: { key: 'uiexp.lend_increment_button' } });
+			}).then(() => {
+				// TODO: REMOVE Once CASH-103: Lend Increment Button EXP ENDS
+				return client.query({ query: experimentQuery, variables: { id: 'lend_increment_button' } });
```
ksekhri: @mcstover cleaned up a bit

src/pages/Lend/LendFilterPage.vue
```diff
@@ -1,81 +1,130 @@
...
+		AlgoliaPagination,
 	},
-
 	metaInfo: {
 		title: 'Lend Filter'
 	},
-
 	created() {
+		const itemsInBasketResults = this.apollo.readQuery({ query: itemsInBasketQuery });
+		this.itemsInBasket = _map(_get(itemsInBasketResults, 'shop.basket.items.values'), 'id');
```
ksekhri: @mcstover Moved this into created as we discussed

PR #621, opened 2019-02-13
set cookie before going back to lend, wwwApp code will read this and …

mcstover:
…setup the full query param filter url.


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -242,6 +241,13 @@ export default {
 		this.updateFromParams(to.query);
 		next();
 	},
+	beforeRouteLeave(to, from, next) {
+		if (typeof window !== 'undefined' && to.path.indexOf('/lend/') !== -1) {
```
emuvente: is window required? are we not expecting this to be called on the server?

mcstover: I only want this called when we click a router link in the client. no server redirect for this.

emuvente: 👍

PR #622, opened 2019-02-13
CASH-349 Add view all loans card on lend by category view

foochris:


foochris:
@mcstover PR updated and remove link on text.

PR #634, opened 2019-02-15
Update visibility checks for kiva credit button now that we have brai…

mcstover:
…ntree in place as well.

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -115,7 +115,7 @@
 									@updating-totals="setUpdatingTotals" />
 
 								<kiva-credit-payment
-									v-else
+									v-if="showKivaCreditButton"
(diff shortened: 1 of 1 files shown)


mcstover:
FYI @BoulderBrains

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -115,7 +115,7 @@
 									@updating-totals="setUpdatingTotals" />
 
 								<kiva-credit-payment
-									v-else
+									v-if="showKivaCreditButton"
```
mcstover: With the addition of the v-if for Braintree above we broke the simple v-else for this. Now we have an explicit check tied to the need for credit, intentionally independent from Paypal and Braintree which will both be active when credit is needed.

BoulderBrains: ✔️

PR #635, opened 2019-02-18
Braintree style starter

mcstover:


mcstover:
@BoulderBrains I did some additional work around the markup (setup to match comps) and styles for the braintree component. For the styles, i've extended what you had started with applying style directly (this is totally how you have to do most styles) and tested out the the options for applying styles via a class (this works for only the base styles). Overall this should be a good started for us with the upcoming tickets.

PR #636, opened 2019-02-18
CASH-550: Standardize light-gray main section

ksekhri:


mcstover:
@ksekhri This should be fine...there may be fallout with any loaders on the pages. Also, this should probably be added to the LoanChannelCategory page.

ksekhri:
@mcstover added 👍

PR #641, opened 2019-02-20
Cash 441

BoulderBrains:
FYI: @ksekhri


src/components/Checkout/PayPalExpress.vue
```diff
@@ -242,7 +248,7 @@ export default {
 						color: 'blue',
 						shape: 'rect',
 						size: (typeof window === 'object' && window.innerWidth > 480) ? 'medium' : 'responsive',
-						fundingicons: true
+						fundingicons: this.showHidePaypalIcons()
```
mcstover: you should be able to just "!showBraintree" here too...

BoulderBrains:
@mcstover correction made, code simplified.

PR #646, opened 2019-02-22
update eslint configuration info for vscode

mcstover:


mcstover:
@BoulderBrains This may help you too.

PR #650, opened 2019-02-22
CASH-555 Implement loan detail filter view (loan length, delinquency …

foochris:
…rate, default rate, risk rating)

### DIFF
--- a/src/pages/Lend/Filter/FilterRangeSlider.vue
+++ b/src/pages/Lend/Filter/FilterRangeSlider.vue
@@ -1,0 +1,216 @@
+<template>
+	<ais-range-input :attribute="queryAttribute" :min="minimum" :max="maximum">
+		<div slot-scope="{ range, refine }">
+			<div class="header">
+				<span class="title">
+					{{ title }}
+				</span>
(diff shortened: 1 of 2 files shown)


foochris:
@ksekhri PR updated with lower/upper range in green/gray:

https://github.com/kiva/ui/pull/650/files#diff-4b4fdc2cd36efbc8e47c92db13a72827R146

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -0,0 +1,221 @@
...
+					({{ rangeLabel }})
+				</span>
+			</div>
+			<input
+				:id="id"
+				type="range"
+				:min="range.min"
+				:max="range.max"
+				:value="rangeValue"
+				:style="showSliderStyle"
```
mcstover: @foochris did you try just "sliderStyle" here... That's already a reactive property... it should just update without an additional computed layer.

foochris: @mcstover PR updated with your suggestion.

PR #651, opened 2019-02-22
Wrap filters in their own file to allow DOM manipulation without resetting filters

ksekhri:
Matt can you test to see if this fixes the issue on your machine too?


mcstover:
@foochris FYI new structure coming your way on these filter sections. Should help us get around the weird loss of refinements...

PR #652, opened 2019-02-23
update algolia configs to point to sandbox instance

mcstover:


mcstover:
FYI @ksekhri @foochris @coderintherye 
Ui Algolia use in VM, Dev + QA will now look at the sandbox account.

PR #653, opened 2019-02-25
fix invalid result count property type

foochris:


src/pages/Lend/Filter/FilterSectionCategories.vue
```diff
@@ -4,7 +4,7 @@
 		file as Algolia filters.
 		https://github.com/algolia/vue-instantsearch/issues/618#issuecomment-466543784
 	-->
-	<filter-menu-section :title="title" :open="open" :result-count="10">
+	<filter-menu-section :title="title" :open="open" :result-count=10>
```
mcstover: this should be in quotes.

src/pages/Lend/Filter/LendFilterMenu.vue
```diff
@@ -7,7 +7,7 @@
 				<kv-icon class="filter-toggle-chevron" name="large-chevron" />
 			</div>
 			<div id="filter-menu">
-				<filter-section-categories result-count="10" />
+				<filter-section-categories :result-count=10 />
```
mcstover: the missing bind ':' was the problem here.

PR #659, opened 2019-02-26
CASH-661 Update style and functionality of KvPillToggle

foochris:


src/components/Kv/KvPillToggle.vue
```diff
@@ -1,14 +1,20 @@
...
-			v-for="option in options" :key="option.key"
+	<div class="pill-toggle">
+		<span
+			v-for="(option, index) in options" :key="option.key"
 			@click="pillToggled(option.key)"
-			class="pill-option"
-			:class="{ active: selected === option.key }">
-			<h3>
+			:class="[
+				'pill-option',
```
ksekhri: Let's use `first-child` and `last-child` for this instead of array indices

foochris: @ksekhri PR updated with `first-child` and `last-child` and other minor SASSy changes.

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,71 @@ export default {
...
+	width: 100%;
 
-.pill-option:first-child {
-	border-top-left-radius: $form-border-radius;
-	border-bottom-left-radius: $form-border-radius;
-}
+	span {
+		background: #F7F7F7;
+		border: rem-calc(1) solid #C4C1C6;
+		color: #808080;
```
mcstover: What are the vars for these colors?

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,71 @@ export default {
...
+			border-bottom-left-radius: 0;
+			border-bottom-right-radius: 0;
+			border-top-left-radius: 0;
+			border-top-right-radius: 0;
+		}
+	}
 
+	.active {
+		background: #8CCB8C;
+		border: rem-calc(1) solid #4FAF4E;
```
mcstover: vars

src/components/Kv/KvPillToggle.vue
foochris: @mcstover Unfortunately, there are no vars for the L54-56 vars.

mcstover: Please create in _settings.scss

foochris: @mcstover Color vars added to `_settings.scss` in a new block.

src/components/Kv/KvPillToggle.vue
foochris: @mcstover Color vars added to `_settings.scss` in a new block.

src/assets/scss/_settings.scss
```diff
@@ -1002,6 +1002,13 @@ $tab-content-border: $light-gray;
 $tab-content-color: $body-font-color;
 $tab-content-padding: 1rem;
 
+// 51.1 Tabs - Pill - vars added by Kiva
+// ------------------
+$tab-pill-border: rem-calc(1) solid #c4c1c6;
+$tab-pill-background: #f7f7f7;
+$tab-pill-color: #808080;
+$tab-pill-active-background: #8ccb8c;
+
```
ksekhri: Nice

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,71 @@ export default {
 	},
 	methods: {
 		pillToggled(key) {
-			this.$emit('pillToggled', key);
+			this.selected = key;
```
ksekhri: Couple of notes here:
1. This is actually an anti-pattern in Vue. We only pass data down the component tree using props, not back up. This causes a console error warning us: `Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "selected"`. Basically whenever our parent re-renders any changes made by this line of code will be reset. @emuvente shared a useful link with more info here: http://michaelnthiessen.com/avoid-mutating-prop-directly/
2. As this is a controlled component, we'll update this value in the parent. We can make other changes to make this a mixed (controlled/uncontrolled) component too but for now we can remove this line and read the `pillToggled` event in the parent.

tl;dr: We can delete line 29

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,71 @@ export default {
...
+			text-align: center;
+			width: 100%;
+		}
 
-.pill-option.active {
-	border: 1px solid $kiva-darkgreen;
-	cursor: default;
-}
+		&:first-child {
+			border-bottom-left-radius: $form-border-radius;
```
ksekhri: Nice!

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,71 @@ export default {
...
-}
+		&:last-child {
+			border-bottom-right-radius: $form-border-radius;
+			border-top-right-radius: $form-border-radius;
+		}
 
-.pill-option.active h3 {
-	background-color: $kiva-green;
-}
+		&:not(:first-child):not(:last-child) {
```
ksekhri: Can we move these lines 82-25 up under line 60 (under selector `.pill-toggle span`)? It'll have the same effect up there.

ksekhri:
@BoulderBrains I think this should work with your code fine, no interface changes

src/components/Kv/KvPillToggle.vue
foochris: @ksekhri PR updated per your suggestion.

src/components/Kv/KvPillToggle.vue
foochris: @ksekhri PR updated per your suggestion.

src/components/Kv/KvPillToggle.vue
mcstover: Thank you!

BoulderBrains:
bitmoji

src/components/Kv/KvPillToggle.vue
```diff
@@ -26,45 +26,61 @@ export default {
...
+			border-top-right-radius: $form-border-radius;
+		}
+	}
 
-.pill-option.active h3 {
-	background-color: $kiva-green;
+	.active {
+		background: $tab-pill-active-background;
+		border: rem-calc(1) solid $kiva-green;
+		color: $white !important;
```
ksekhri: We should remove `!important` here

foochris: @ksekhri PR updated with removal of `!important`

PR #661, opened 2019-02-27
CASH-661: Cleanup borders

ksekhri:
There was a visual issue where there were two borders in between elements (sometimes of differing colors. @horationunez came up with a clean solution to it.


foochris:
Changes merged into CASH-661 PR

PR #665, opened 2019-02-28
One small fix to get this exp up and running correctly

BoulderBrains:
FYI @mcstover missed this previously somehow. Just gonna merge it in. Tested on my VM


mcstover:
👍

PR #666, opened 2019-02-28
CASH-556 Enable filter by gender

foochris:


foochris:
@BoulderBrains Assuming that this PR is approved, the toggle emit object will be changing:

https://github.com/kiva/ui/pull/666/files#diff-4be7f5dba997b4240c798e3bce9fb99aR32

So PaymentWarapper.vue will have to be updated:

https://github.com/kiva/ui/blob/d1450305b56aefb99426073e53275dee6a4d920e/src/components/Checkout/PaymentWrapper.vue#L70

ksekhri:
Looks good @foochris, just working on a review. Let's not merge this yet.

foochris:
@ksekhri New `FilterSectionGender` file created

src/components/Kv/KvPillToggle.vue
```diff
@@ -23,10 +23,15 @@ export default {
...
+			default: null,
+		}
 	},
 	methods: {
-		pillToggled(key) {
-			this.$emit('pill-toggled', key);
+		pillToggled(key, callback) {
+			// eslint-disable-next-line quote-props
+			this.$emit('pill-toggled', { 'key': key, 'callback': callback });
 		}
```
ksekhri: We’ll have to find a different solution for this. Passing a function to a component then getting it back from that component isn’t a pattern that will be reused and as such doesn’t belong in a reusable widget.

src/pages/Lend/Filter/FilterSectionGender.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+	<!--
+		This category page was added as a temporary workaround to a bug preventing DOM manipulation in the same Vue
```
ksekhri: We can get rid of these comments

src/components/Kv/KvPillToggle.vue
ksekhri: My suggestion is passing the refine function as a prop to a component that encapsulates the toggle. Then when the toggle sends the event we can call this.refine.

foochris: @ksekhri PR updated with your suggestions. Thanks!

foochris:
@ksekhri FYI, pill set to 'All genders' when no gender refinements are applied and filter menu is closed.

https://github.com/kiva/ui/pull/666/files#diff-c6cd67350f044b8524f9282a762f2a8cR46

src/pages/Lend/Filter/FilterSectionGenderPillToggle.vue
```diff
@@ -0,0 +1,90 @@
...
+	watch: {
+		items(data) {
+			// set gender pill to 'all genders' when menu is closed and gender filter has been removed
+			if (this.filterMenuOpen === false && _filter(data, { isRefined: true }).length === 0) {
+				this.genderOptionSelected = 'all_genders';
+			}
+		}
+	},
+	methods: {
+		genderPillToggled(key) {
```
ksekhri: Very well thought out :+1:

PR #667, opened 2019-02-28
CASH-555: Wrap filter section range slider in component, support IE slider style, enable computed/watch

foochris:


foochris:
@ksekhri / @mcstover The slider has been refactored and it's probably easier to view the entire file rather than the diff. 

CASH-669 has also be combined with this PR.

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -100,16 +98,65 @@ export default {
...
+				value = `${this.selectedRangeValue}${this.inbetweenLabel}`;
+			}
+
+			return value;
+		},
+	},
+	watch: {
+		// upon deletion of refinement (filter menu closed) update range: label, value and style (liner gradient)
+		currentRefinement(data) {
+			if (this.filterMenuOpen === false) {
```
mcstover: This is confusing me...Not sure why we would want to do anything b/c you closed the Filter drop down.

foochris: @mcstover It's when a user removes a facet when the filter menu is closed; the slider, label and gradient (gray / green) have to be updated.

mcstover: oh ok, thanks, that makes sense.

ksekhri: Let's discuss this `watch` over our chat tomorrow. This runs 8 times every time we set or unset a filter. Is there a way to replace this functionality by hooking into the `transformItems` method?

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -100,16 +98,65 @@ export default {
...
+				if (data.min === null && data.max === null) {
```
ksekhri: Why/when does minimum value for a max only slider change from -1 to null?

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -119,112 +166,158 @@ export default {
...
-	border: 0;
-}
+	.range-label {
+		color: $kiva-text-light;
+	}
 
-input[type=range]::-moz-range-progress {
-	background-color: $kiva-green;
-	border: rem-calc(1) solid $kiva-green;
+	.slider {
```
ksekhri: This is a nice block of SASS

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -119,112 +166,158 @@ export default {
...
 	},
 	created() {
 		if (this.minimumMaximumSwap === true) {
-			this.rangeLabel = this.minimumLabel;
-			this.sliderStyle = 'background: #dbdbdb';
+			this.selectedRangeValue = this.range.min;
+
+			// set slider to all gray
+			// eslint-disable-next-line max-len
+			this.sliderStyle = 'background-image: -webkit-linear-gradient(left, #4faf4e 0%, #4faf4e 0%, #dbdbdb 0%, #dbdbdb 100%)';
```
ksekhri: Let's chat about `this.sliderStyle` tomorrow, it would be interesting to learn about the challenges in making this a computed property. Will also be interesting to understand why this code differs from `setSlideStyle`.

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -119,112 +166,158 @@ export default {
...
 
 			if (this.minimumMaximumSwap === true) {
-				if (minimumValue === null && selectedRangeValue === null) {
-					minimumValue = range.max;
-					selectedRangeValue = range.max;
+				if (minimumValue === 0 && selectedRangeValue === 0) {
+					minimumValue = -1;
+					selectedRangeValue = 1000;
+
+					// eslint-disable-next-line no-param-reassign
```
ksekhri: We shouldn't be reassigning parameters

src/pages/Lend/Filter/FilterRangeSlider.vue
foochris: @ksekhri Unfortunately, `ais-range-input` does not have `transformItems`

src/pages/Lend/Filter/FilterRangeSlider.vue
foochris: @ksekhri The max/min value(s) are set to null when the refinement is removed from the SelectedRefinements.

src/pages/Lend/Filter/FilterRangeSlider.vue
foochris: @ksekhri I've added this computed method. It's somewhat a hack, but it does work and is not as called as often as the watch method:

```
		// set slider position
		sliderPosition() {
			//	eslint-disable-next-line max-len
			if (this.filterMenuOpen === false && this.currentRefinement.min === null && this.currentRefinement.max === null) {
				this.setSliderValue();
			}

			return '';
		},
```

foochris:
@ksekhri `FilterRangeSlider` completely refactored based on your comments.

foochris:
@ksekhri PR updated

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -1,66 +1,60 @@
...
 				:style="sliderStyle"
 				@input="onSliderChange(
 					refine,
 					range,
-					$event)"
-			>
+					$event)">
+
+			<div class="hide">
+				{{ sliderPosition }}
```
ksekhri: Lets add a comment with a `// TODO: ` signifying that this was added to make the component reactive as well as a quick reminder of why we needed to add a hidden div and why line 132 doesn't need to return anything

src/pages/Lend/Filter/FilterRangeSlider.vue
```diff
@@ -87,144 +79,201 @@ export default {
...
+				? (this.sliderValue / this.maximum)
+				: (Math.abs(this.sliderValue - this.minimum) / Math.abs(this.maximum - this.minimum));
+
+			const basicStyle = 'background-image: -webkit-linear-gradient(left,';
+			const stop1 = `${sliderLeftColor} 0%,`;
+			const stop2 = `${sliderLeftColor} ${numeral(selectedStop).format('0%')},`;
+			const stop3 = `${sliderRightColor} ${numeral(selectedStop).format('0%')},`;
+			const stop4 = `${sliderRightColor} 100%);`;
+
+			return `${basicStyle}${stop1}${stop2}${stop3}${stop4}`;
```
ksekhri: I like this :+1: very readable

PR #674, opened 2019-03-05
CASH-551 - Sort loans in lend/filter

foochris:


src/pages/Lend/Filter/FilterSectionSort.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+	<filter-menu-section :title="title" :open="open" :result-count="resultCount">
+		<ais-sort-by :items="defaultSortIndices">
```
mcstover: Please pass defaultSortIndices down into this component. This so we don't have to include configure or the mixin here.

Alternatively, you could pull the fundraisingIndices + defaultSortIndices() out of algolia-config-mixin into a 2nd algolia-sort-indices-mixin.

foochris: @mcstover PR updated.

src/pages/Lend/Filter/FilterSectionSort.vue
```diff
@@ -0,0 +1,74 @@
...
+import FilterMenuSection from '@/pages/Lend/Filter/FilterMenuSection';
+
+export default {
+	components: {
+		AisConfigure,
+		AisSortBy,
+		FilterMenuSection,
+	},
+	filters: {
+		capitalize(value) {
```
foochris: @mcstover Should we move this globally? There may be another use for it in `SelectedRefinements.vue` as the label 'female / male' should be upper case.

src/pages/Lend/Filter/LendFilterMenu.vue
```diff
@@ -78,11 +80,13 @@
 
 <script>
 
+import algoliaConfig from '@/plugins/algolia-config-mixin';
```
mcstover: @foochris this is already included in the page. please pass down from there.

foochris: @mcstover PR updated.

PR #676, opened 2019-03-05
cash-568 fixing alignment of lyml header

BoulderBrains:
Another small change that fixes cash-568

### DIFF
--- a/src/components/LoansYouMightLike/lymlContainer.vue
+++ b/src/components/LoansYouMightLike/lymlContainer.vue
@@ -265,7 +265,6 @@ export default {
 	font-weight: 400;
 	margin-bottom: 1rem;
 	line-height: 0.8;
-	margin-left: 1.9rem;
 }
 
 .arrow {


src/components/LoansYouMightLike/lymlContainer.vue
```diff
@@ -265,7 +265,6 @@ export default {
 	font-weight: 400;
 	margin-bottom: 1rem;
 	line-height: 0.8;
-	margin-left: 1.9rem;
```
mcstover: what does not having this style this do at wider screen sizes?

BoulderBrains: It just keeps the heading left aligned to that area, which looked the most consistent with the way the headings respond on the /lend-by-category page.

mcstover: Ok so long as it's aligned with the left edge of the first image we're good. Should have been more specific in my first comment.

BoulderBrains: It's aligned across all screen sizes as long as the user doesn't scroll the cards left or right on a smaller screen. Once the scroll happens, the cards don't always scroll back all the way to the left and the alignment can be a little off. I think this is the best we can do without having the card lock into specific scroll position. More than willing to talk about this in person with you tomorrow if you want.

mcstover: thanks for verification on that. The beginning state before scrolling is all we're really worried about.

PR #682, opened 2019-03-07
Add config to run Ui Server in a non-VM environment

foochris:


mcstover:
Hey, there might be another way to do this that doesn't end up with so much repetition... please hold for now.

config/dev-local.js
```diff
@@ -0,0 +1,45 @@
+var merge = require('webpack-merge')
+var base = require('./index.js')
+
```
mcstover: You can import dev-vm here too....

config/dev-local.js
```diff
@@ -0,0 +1,45 @@
...
+module.exports = merge(base, {
+	app: {
```
mcstover: then `merge(base, dev-vm, {....`

And you'd only need to include items that need altering...app { publicPath }, server { memcached } etc...

foochris:
@mcstover PR updated.

mcstover:
FYI @foochris I merged in the new apache configs earlier today so the proxy pass for /uiext should show up in the vms soon.

PR #684, opened 2019-03-08
CASH-640: Loan Card Controller

ksekhri:


src/components/LoanCards/LoanCardController.vue
```diff
@@ -0,0 +1,186 @@
...
+		:items-in-basket="itemsInBasket"
+		:is-visitor="isVisitor"
+		:lend-increment-button-version="lendIncrementButtonVersion"
+		:image-enhancement-experiment-version="imageEnhancementExperimentVersion"
+		:is-favorite="isFavorite"
+		:expiring-soon-message="expiringSoonMessage"
+		:percent-raised="percentRaised"
+		:is-funded="isFunded"
+		:amount-left="amountLeft"
+		@track-interaction="trackInteraction"
```
mcstover: also need to listen for the toggle-favorite event here.

src/components/LoanCards/LoanCardController.vue
```diff
@@ -0,0 +1,186 @@
...
+			}
+			const hours = differenceInHours(this.loan.plannedExpirationDate, Date.now());
+			if (hours >= 2) {
+				return `Only ${hours} hours left! `;
+			}
+			const mins = differenceInMinutes(this.loan.plannedExpirationDate, Date.now());
+			if (mins >= 2) {
+				return `Only ${mins} minutes left! `;
+			}
+			return 'Expiring now!';
```
mcstover: I think our next step (not for this pr) with all of these computed properties and even the methods would be to move them to a standard loan-display-utils.js file in our utils folder. We could then have methods for each which we could import individually and write better tests around.

ksekhri: I like that idea, I'll keep thinking about that

PR #688, opened 2019-03-12
CASH-402: Refactor BorrowerInfo to increase shared code

ksekhri:


src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
```diff
@@ -0,0 +1,96 @@
...
+					interactionType: 'viewBorrowerPage',
+					interactionElement: 'readMore'
+				})"
+			>Read more</span>
+		</router-link>
+		<!--
+        <div v-if="activeSort === 'loanLength'" class="loan-length">
+            <strong>Loan length:</strong> {{ loanLength }} months
+        </div>
+        -->
```
ksekhri: It appears this has been deprecated

mcstover: I think we will need something like this to activate when on the popularity/trending now sort via Algolia. Good to leave commented out for now.

src/components/LoanCards/BorrowerInfo/BorrowerInfoHeader.vue
```diff
@@ -0,0 +1,58 @@
...
+			<span
+				@click="$emit('track-loan-card-interaction', {
+					interactionType: 'viewBorrowerPage',
+					interactionElement: 'borrowerName'
+				})"
+			>
+				{{ name }}
+			</span>
+		</router-link>
+		<div class="country" v-if="country">{{ country }}</div>
```
mcstover: This will likely need to come out of here too at some point. Pretty sure the mini loan cards show country + activtiy under the title. As you work on the list card, I would move the borrower info wrapper div into your card and then include each item directly, preferably as a "dumb" components that just receive the single value for the name...or country...or activity. This as opposed to combining them as a group which is not versatile.

ksekhri: Interesting, I like the idea here. Let's discuss this further in our sync up.

PR #691, opened 2019-03-13
CASH-402: Add ListLoanCard

ksekhri:
* The only file changed in this PR is `ListLoanCard.vue`. I'll merge master into this branch once the supporting PRs are approved to ensure it reflects that.


src/components/LoanCards/ListLoanCard.vue
```diff
@@ -1,13 +1,154 @@
 <template>
-	<div class="list-loan-card">
-		ListLoanCard Placeholder
+	<div class="list-loan-card row">
+		<div class="list-loan-card-desktop-column show-for-large large-4 xlarge-4 columns">
```
mcstover: This should be your first column which stacks at the smaller sizes rather than duplicates the content.

src/components/LoanCards/ListLoanCard.vue
```diff
@@ -1,13 +1,154 @@
...
+			<div class="list-loan-card-desktop-column-image-wrapper">
+				<loan-card-image
```
mcstover: We need to structure the markup so that the loan card image component is not duplicated. Text copy can be duplicated where needed but not the images.

PR #692, opened 2019-03-13
Emit an event with upon remove of a facet in current refinements alon…

mcstover:
…g with information about that facet.


mcstover:
This enables us to perform clean up on the global disjunctiveRefinements object as those items get removed.

PR #695, opened 2019-03-13
Switching pay with card/ pay with paypal order

BoulderBrains:
This change swaps the pay with paypal/ pay with card options in the payment toggle component.


BoulderBrains:
Going to merge this in before branch cut!

PR #696, opened 2019-03-15
CASH-404: Add isSelectedByAnother status to ActionButton

ksekhri:


src/components/LoanCards/Buttons/LoanSelectedText.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+	<div class="loan-selected-text small-text">
+		Another lender has selected this loan. Please choose a different borrower to support.
```
mcstover: Where did this copy come from? I don't think it represents the actual situation (a funded loan) very well.

PR #697, opened 2019-03-15
CASH-635 Update Algolia pagination component to align with current UI

foochris:


src/pages/Lend/AlgoliaPagination.vue
```diff
@@ -1,103 +1,155 @@
...
-			isFirstPage,
-			isLastPage,
-			refine,
-			createURL
-			}"
+	<ul class="ais-Pagination-list">
+		<!-- previous page -->
+		<li
+			class="ais-Pagination-item ais-Pagination-item--previousPage"
+			:class="isFirstPage ? 'ais-Pagination-item--disabled' : ''">
```
ksekhri: Let's use object syntax for this
https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax
`:class="{'ais-Pagination-item--disabled': isFirstPage}"`

src/pages/Lend/AlgoliaPagination.vue
```diff
@@ -1,103 +1,155 @@
...
+				v-if="number - 1 === currentRefinement"
+				class="ais-Pagination-link">
+				{{ number }}
+			</span>
+		</li>
+
+		<!-- next page -->
+		<li
+			class="ais-Pagination-item ais-Pagination-item--nextPage"
+			:class="isLastPage ? 'ais-Pagination-item--disabled' : ''"
```
ksekhri: Object syntax

src/pages/Lend/AlgoliaPagination.vue
foochris: @ksekhri PR updated with object syntax.

src/pages/Lend/AlgoliaPagination.vue
foochris: @ksekhri PR updated with object syntax.

src/pages/Lend/AlgoliaPagination.vue
```diff
@@ -1,103 +1,155 @@
...
+				@click.prevent="refine(currentRefinement - 1)"
+				aria-label="Previous"
+				class="ais-Pagination-link"
+			>
+				<kv-icon name="triangle" />
+			</a>
+		</li>
+
+		<li v-for="(number, index) in numbers"
+			:key="number || -index"
```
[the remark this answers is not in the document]
ksekhri: cc @emuvente would this be an ok change to make?

src/pages/Lend/AlgoliaPagination.vue
```diff
@@ -1,103 +1,155 @@
...
+			required: true,
+		},
+	},
+	data() {
+		return {
+			maximumExtraPages: 3,
+		};
+	},
+	computed: {
+		numbers() {
```
emuvente: This looks like it was copied directly from KvPagination. Maybe this needs to be done as a mixin?

mcstover: It has some differences so we have another tech-debt ticket to de-dupe the code later

src/pages/Lend/AlgoliaPagination.vue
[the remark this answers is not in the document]
emuvente: `key` is used to prevent unnecessary re-rendering. What makes each of these elements most unique is their page number, so basing the key on that allows more elements to be re-used between renders. However, not all elements have a number (i.e. ellipsis elements), and so for those we choose a key that won't interfere with other keys, thus `-index`.

src/pages/Lend/AlgoliaPagination.vue
emuvente: 👍

src/pages/Lend/AlgoliaPagination.vue
```diff
@@ -1,103 +1,155 @@
...
+			:class="{ 'ais-Pagination-item--disabled' : isFirstPage }">
+			<span
+				v-if="isFirstPage"
+				aria-label="Previous (Inactive)"
+				class="ais-Pagination-link"
+			>
+				<kv-icon name="triangle" class="disabled" />
+			</span>
+			<a
+				v-if="!isFirstPage"
```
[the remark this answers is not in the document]
foochris: @ksekhri PR updated with v-else and other suggestions

PR #701, opened 2019-03-18
Cash 634 algolia url state

mcstover:
Demo of how to create custom formatted url segments as a store for the page state and source for instantiating the page with a specified state.


mcstover:
@emuvente This has an initial direction for converting Algolia search state to a url. We'll catch up with you soon on planning a more robust spec for this soon.

PR #702, opened 2019-03-18
AUTH-114 fix missing basketId arg when reading the basket from the ca…

emuvente:
…che after a custom prefetch

### DIFF
--- a/src/graphql/query/loanChannelPage.graphql
+++ b/src/graphql/query/loanChannelPage.graphql
@@ -1,4 +1,4 @@
-query($basketId: String) {
+query {
 	my {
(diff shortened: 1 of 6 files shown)


src/graphql/query/loanChannelPage.graphql
```diff
@@ -13,15 +13,6 @@ query($basketId: String) {
...
 	}
-	shop (basketId: $basketId) {
-		basket {
-			items {
-				values {
-					id
-				}
-			}
-		}
-	}
```
emuvente: This shop data wasn't being used in the only component that used this query, so I removed it.

PR #705, opened 2019-03-19
updated algolia dependencies and dotenv

mcstover:


mcstover:
FYI @emuvente @ksekhri @BoulderBrains Started updating some deps and will continue in chunks... probably starting with Build deps, then Vue + it's deps then whatever is left.

PR #708, opened 2019-03-21
Add new client state directives for mutating and managing the add to …

mcstover:
…basket interstitial lightbox. This gives us a global default state of false which is altered when the experiment has been assigned to active. Add to basket buttons monitor the global state and trigger a mutation setting a loan id for the interstitial if the exp is active.

### DIFF
--- a/src/api/StateLink.js
+++ b/src/api/StateLink.js
@@ -12,7 +12,7 @@ export default ({ cache, ...options }) => {
 			experiment(options),
 			tipMessage(options),
 			usingTouch(options),
-			basketAddInterstitial(options)
+			basketAddInterstitial(options),
 		),
 	});
 };
--- a/src/api/StateLink.js
+++ b/src/api/StateLink.js
@@ -3,6 +3,7 @@ import { withClientState } from 'apollo-link-state';
 import experiment from './localResolvers/experiment';
 import tipMessage from './localResolvers/tipMessage';
 import usingTouch from './localResolvers/usingTouch';
+import basketAddInterstitial from './localResolvers/addToBasketInterstitial';
 
 export default ({ cache, ...options }) => {
 	return withClientState({
@@ -10,7 +11,8 @@ export default ({ cache, ...options }) => {
 		..._merge(
 			experiment(options),
 			tipMessage(options),
-			usingTouch(options)
+			usingTouch(options),
+			basketAddInterstitial(options)
 		),
 	});
 };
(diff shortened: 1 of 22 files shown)


src/api/StateLink.js
```diff
@@ -3,14 +3,16 @@ import { withClientState } from 'apollo-link-state';
...
 
 export default ({ cache, ...options }) => {
 	return withClientState({
 		cache,
 		..._merge(
 			experiment(options),
 			tipMessage(options),
-			usingTouch(options)
+			usingTouch(options),
+			basketAddInterstitial(options)
```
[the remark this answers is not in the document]
mcstover: Not ready for a policy change but I will add a trailing comma in my updates

src/components/Kv/KvButton.vue
```diff
@@ -77,11 +77,19 @@ button.button {
...
+
 	&.setting {
 		@include secondary-styles();
 		@include button-setting-styles();
 	}
 
+	&.setting.smallest {
+		@include button-smallest-styles();
+	}
+
```
mcstover: applying smallest styles with both of these.

src/components/Kv/KvLightbox.vue
```diff
@@ -160,6 +160,7 @@ export default {
 
 			#lightbox-columns {
 				position: relative;
+				width: 100%;
```
mcstover: ensuring this parent container is set to 100% of the parent instead of stretching to the full width of child content.

src/components/LoanCards/Buttons/LendIncrementButton.vue
```diff
@@ -18,15 +18,20 @@
...
 </template>
 
 <script>
 import numeral from 'numeral';
 import LendButton from '@/components/LoanCards/Buttons/LendButton';
+import addToBasketInsterstitial from '@/plugins/add-to-basket-show-interstitial';
 
 export default {
+	mixins: [
+		addToBasketInsterstitial
```
ksekhri: Understanding this is currently a WIP, could this become an event that bubbles up and is handled by `ActionButton` in the future?

src/components/Kv/KvLightbox.vue
ksekhri: What was the issue this fixed? Would adding a class of `small-12` have the same effect?

mcstover: I'll look into this.

src/components/LoanCards/Buttons/LendIncrementButton.vue
mcstover: Definitely planning to add a loading sequence as I shore up the image + structure of the loan preview.

I'll also work to cleanup the duplication with that event, good call.

PR #710, opened 2019-03-22
CASH-737 Add featured hero wrapper to use loan card controller

foochris:


src/components/LoanCards/LoanCardController.vue
```diff
@@ -13,12 +14,15 @@
 		:is-funded="isFunded"
 		:is-selected-by-another="isSelectedByAnother"
 		:amount-left="amountLeft"
+		:show-category-description="showCategoryDescription"
+		:experiment-data="experimentData"
```
foochris: @ksekhri  / @mcstover I can imagine one or more experiments and hence the creation of `experimentData` (type: Array). Also, I believe having the `loanChannel` (type: Object) would be useful for any future loan card views.

foochris:
@ksekhri / @mcstover All data retrieval (including the experiment query) for the featured hero loan occurs in `FeaturedHeroLoanWrapper` and the child component, `FeaturedHeroLoan`, has been refactored to reflect this change.

src/components/LoanCards/LoanCardController.vue
mcstover: @foochris I think the experimentData array could be useful. You'll want to change how your accessing it though to use find or filter instead of keying off of an index.

As for the loan channel, I don't want to add this to the loan as it will in your case contain the same loan you passed in and in other cases contain all of the loans we queried for a loan channel. If there is a need to pass down the Loan Channel ID, Name or other meta data we can extract that when the time comes.

foochris: @mcstover If the loan channel object is not to be passed (and used to display the description), what if we had a property for generic text instead ('text', type: String)?

https://github.com/kiva/ui/blob/a688861be1529e03e95f9f67791f4b750f673d3c/src/components/LoansByCategory/FeaturedHeroLoan.vue#L7

foochris: @mcstover PR updated with your suggestion of moving the loan description to the wrapper.

src/components/LoansByCategory/FeaturedHeroLoan.vue
```diff
@@ -256,99 +220,24 @@ export default {
...
-		this.experimentVersion = _get(experimentQuery, 'experiment.version') || null;
-
-		if (this.experimentVersion === 'variant-a') {
-			this.$kvTrackEvent('Lending', 'EXP-CASH-684-Mar2019', 'a');
-			this.showFundraisingThermometer = false;
-		} else if (this.experimentVersion === 'variant-b') {
-			this.$kvTrackEvent('Lending', 'EXP-CASH-684-Mar2019', 'b');
-			this.showFundraisingThermometer = true;
+		if (this.experimentData.length > 0) {
+			this.experimentVersion = _get(this.experimentData, '0.experiment.version') || null;
```
ksekhri: I think you can simplify this by removing the if wrapper on line 231:
```
this.experimentVersion = this.experimentData.featured_hero_loan_fundraising_thermometer;
if (this.experimentVersion === 'variant-a') {
...
```

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -0,0 +1,163 @@
...
+	},
+	created() {
+		// CASH-684 : Experiment : Featured hero loan fundraising thermometer
+		const experimentQuery = this.apollo.readQuery({
+			query: experimentAssignmentQuery,
+			variables: { id: 'featured_hero_loan_fundraising_thermometer' },
+		});
+
+		// push the query to the experiment data array
+		this.experimentData = [experimentQuery];
```
ksekhri:
```
this.experimentData = {featured_hero_loan_fundraising_thermometer: _get(experimentQuery, 'experiment.version')};
```

src/components/LoanCards/LoanCardController.vue
```diff
@@ -90,6 +94,14 @@ export default {
 			type: String,
 			default: ''
 		},
+		showCategoryDescription: {
+			type: Boolean,
+			default: false,
+		},
+		experimentData: {
+			type: Array,
+			default: () => [],
```
ksekhri: Let's change this to an object. The key can be the experiment id.

src/components/LoanCards/LoanCardController.vue
foochris: @ksekhri / @mcstover PR updated with `experimentData` as `type: Object`.

src/components/LoanCards/LoanCardController.vue
```diff
@@ -13,12 +13,15 @@
 		:is-funded="isFunded"
 		:is-selected-by-another="isSelectedByAnother"
 		:amount-left="amountLeft"
+		:show-category-description="showCategoryDescription"
```
mcstover: We don't need to include this here...it's shouldn't be used in the loan card view.

src/components/LoanCards/LoanCardController.vue
```diff
@@ -90,6 +94,14 @@ export default {
 			type: String,
 			default: ''
 		},
+		showCategoryDescription: {
```
mcstover: same as above.

src/components/LoansByCategory/FeaturedHeroLoan.vue
```diff
@@ -164,54 +151,42 @@ export default {
...
+					},
+					image: {},
+				};
+			}
+		},
+		percentRaised: {
+			type: Number,
+			default: 0,
+		},
 		showCategoryDescription: {
```
mcstover: no longer part of this component.

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -0,0 +1,162 @@
...
+			:is-visitor="!isLoggedIn"
+			:items-in-basket="itemsInBasket"
+			:key="loan.id"
+			:lend-increment-button-version="lendIncrementButtonVersion"
+			:loan="loan"
+			loan-card-type="FeaturedHeroLoan"
+			:loan-channel="loanChannel"
+			:position="1"
+			:row-number="0"
+			:show-category-description="showCategoryDescription"
```
mcstover: remove

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -0,0 +1,162 @@
...
+			:enable-tracking="true"
+			:experiment-data="experimentData"
+			:image-enhancement-experiment-version="imageEnhancementExperimentVersion"
+			:is-visitor="!isLoggedIn"
+			:items-in-basket="itemsInBasket"
+			:key="loan.id"
+			:lend-increment-button-version="lendIncrementButtonVersion"
+			:loan="loan"
+			loan-card-type="FeaturedHeroLoan"
+			:loan-channel="loanChannel"
```
mcstover: remove

src/components/LoanCards/LoanCardController.vue
```diff
@@ -90,6 +94,14 @@ export default {
...
+			type: Boolean,
+			default: false,
+		},
+		experimentData: {
+			type: Object,
+			default: () => [],
```
ksekhri: We should update the default to be an object

src/components/LoansByCategory/FeaturedHeroLoan.vue
```diff
@@ -144,18 +119,30 @@ export default {
...
-		enableTracking: {
-			type: Boolean,
-			default: false,
+		amountLeft: {
+			type: Number,
+			default: 0,
+		},
+		experimentData: {
+			type: Object,
+			default: () => [],
```
ksekhri: Default object

foochris: @ksekhri PR updated.

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
foochris: @ksekhri PR updated.

PR #711, opened 2019-03-22
Ensure basket id is retrieved and passed to queries that expect it. T…

mcstover:
…here seems to be a race between resolving pre fetch data for multiple components so this may pop up in others too.

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBanner.vue
+++ b/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBanner.vue
@@ -7,6 +7,7 @@
 <script>
 import numeral from 'numeral';
 import _get from 'lodash/get';
+import cookieStore from '@/util/cookieStore';
 import { settingEnabled } from '@/util/settingsUtils';
 import promoQuery from '@/graphql/query/promotionalBanner.graphql';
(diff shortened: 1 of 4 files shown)


mcstover:
FYI @ksekhri @BoulderBrains @foochris @emuvente  I was running into 500 errors on server render of several pages after running the dev server for about 30 minutes. The issue was always that 1 of the queries didn't have the basket id in the shop query.

Please chime in if you've seen the same.

mcstover:
@emuvente Should we move forward with this?

emuvente:
The basket id is already being added to all variables and all prefetch variables in https://github.com/kiva/ui/blob/master/src/api/BasketLink.js#L13 and https://github.com/kiva/ui/blob/master/src/plugins/apollo-plugin.js#L27, so I don't see how this pr would change anything. There must be something else going on that's occasionally causing the basketid to be missing. Since it happens every 30 minutes, it's probably related to the ui server being restarted by puppet.

emuvente:
Actually, I'm sorry, it looks like I forgot to add it here: https://github.com/kiva/ui/blob/master/src/util/apolloPreFetch.js#L37. I think the problem can be fixed by including cookiestore in that file and adding the basketId to the prefetchVariables the same way as in apollo-plugin.

PR #712, opened 2019-03-25
Cash 721

BoulderBrains:


BoulderBrains:
@mcstover,
@ksekhri and I to work through some of these changes. Please note that we tried to use .nextTick() instead of the setTimeout(), however we were not able to get that working. This code works as it's written though I'm open to suggestions on how to make it better. Currently working as requested in the ticket.

BoulderBrains:
Also, the padding request within this ticket is not yet made, Thuan wanted to see the changes in dev then work with what's there to tweak the spacing on everything.

src/components/Checkout/DonationNudge/DonationNudgeBoxes.vue
```diff
@@ -55,6 +54,9 @@
 								ref="customDonationInputDesktop"
 								name="customDonationInputTextDesktop"
 								maxlength="10"
+								tabindex="10"
+								id="customDonationInput"
+								@click.capture="formatAndSubmitOnEnter"
 								class="nudge-box-input nudge-box-input-desktop"
```
mcstover: The event and value for this input would be better handled with v-model assigned to a data item and @click.enter with your method call.

src/components/Checkout/DonationNudge/DonationNudgeBoxes.vue
```diff
@@ -250,6 +264,20 @@ export default {
...
+		formatAndSubmitOnEnter() {
+			const customInput = document.getElementById('customDonationInput');
+			const customInputButton = document.querySelector('.custom-amount-submit');
+
+			customInput.addEventListener('keyup', event => {
+				if (event.keyCode === 13) {
+					customInput.blur();
+					customInputButton.click();
+				}
+			});
```
mcstover: I think everything here could go away and you could just call setCustomInputAndClose... may even be able to call that directly from the @click.enter="setCustomInputAndClose"

mcstover: if this would require a significant refactor of the component you can skip for now...

src/components/Checkout/DonationNudge/DonationNudgeBoxes.vue
mcstover: if this would require a significant refactor of the component you can skip for now...

PR #723, opened 2019-03-27
first attempt to always inject basketId into our prefetchApolloQuery.

mcstover:


mcstover:
@emuvente Here's where I ended up...but was ultimately not successful. Any thoughts?

emuvente:
config.variables and config.preFetchVariables are always functions. I don't think you want to be referencing or changing config.variables at all though, as that function is expected to be called in the context of a vue component, which we don't have yet in this pre-fetch.
I think you might just need something like this:
```
...
const basketId = cookieStore.get('kvbskt');
const prefetchVariables = config.preFetchVariables ? config.preFetchVariables(args) : {};
...
	client.query({
		query: config.query,
		variables: { basketId, ...prefetchVariables }
...
```

mcstover:
That's where I started but I still had intermittent failures. Considered that maybe it needed to take into consideration the promise returned above before getting to the query.

PR #724, opened 2019-03-27
CASH-66: Anchor 17% as low tip option in new checkout

ksekhri:


ksekhri:
cc @BoulderBrains for reference

BoulderBrains:
👀 Just swinging by.

PR #728, opened 2019-03-28
CASH-232: Update basket when adding loan

ksekhri:
~@emuvente I think Matt and Josh are out today, can you take a look at this tiny bugfix?~
Oh cool @mcstover just reviewed, thanks Matt


BoulderBrains:
👀 Thanks @ksekhri

PR #729, opened 2019-03-29
Reduce client/server compile times by using Babel cache directory

foochris:


foochris:
A minor change to the webpack configuration that should reduce client compile times significantly (50-75%). No significant change in server times were observed.

Change uses the Babel cache directory option (`cacheDirectory`):

https://github.com/babel/babel-loader#options

BoulderBrains:
👀 Playing catch up after being on PTO 😃

PR #731, opened 2019-03-29
Fetch mostRecentPaymentType on initializebasket query to inform the n…

mcstover:
…ew PaymentWrapper when to default to the paypal or braintree form.


src/components/Checkout/PaymentWrapper.vue
```diff
@@ -61,9 +65,12 @@ export default {
 					key: 'pp',
 				},
 			],
-			selectedOption: 'pp',
+			selectedOption: 'bt',
```
mcstover: We now default to Braintree unless we know they used PayPal in their last transaction.

PR #734, opened 2019-04-01
use random instead of expringSoon to help mitigate un-actionable loans

mcstover:


mcstover:
FYI. Nick wanted to change this back b/c of the frequency of expired loans that were not allowing you to add them to basket.

mcstover:
The errorred build above is for the same branch on a closed pr. all tests pass.

BoulderBrains:
👀 👍 That change makes sense, I was having issue developing while it was pulling in the expringSoon loans too. ✔️

PR #742, opened 2019-04-04
CASH-617 Add brower/filter url to LendHeader

foochris:


src/pages/Lend/LendHeader.vue
```diff
@@ -17,6 +17,16 @@ import ViewToggle from '@/components/LoansByCategory/ViewToggle';
 export default {
 	components: {
 		ViewToggle,
+	},
+	props: {
+		browseUrl: {
+			type: String,
+			default: '/lend-by-category'
```
mcstover: I think these defaults can prob be empty at this level

foochris: @mcstover Will update to blank/empty.

PR #744, opened 2019-04-05
Fix favoriting functionality and title on loan channel cateogry page.

mcstover:


mcstover:
Couple of small fixes.

PR #756, opened 2019-04-09
CASH-765: Desktop design of filter menu

ksekhri:


src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -132,6 +135,7 @@ export default {
 		return {
 			itemsInBasket: null,
 			isLoggedIn: false,
+			filterMenuOpen: false,
```
ksekhri: Moved up a level to match new mocks

PR #762, opened 2019-04-11
Positioning fix for category row carousel controls on IE 11. JS error…

mcstover:
… fix for add to basket on lend-by-cat when logged in with a kiva admin user. The fix was a sync of values returned for similar graphql queries in lend by cat exp manager.


mcstover:
This fixes the /lend-by-cat exp admin error that was popping up and positioning of carousel controls on IE 11.
FYI @foochris

PR #763, opened 2019-04-11
CASH-765: Cleanup pill toggle

ksekhri:


src/components/Kv/KvPillToggle.vue
```diff
@@ -39,7 +39,7 @@ export default {
 $form-border-radius: rem-calc(3);
 
 .pill-toggle {
-	font-size: rem-calc(14);
+	font-size: 1rem;
 	height: rem-calc(34);
```
ksekhri: Hey @BoulderBrains I tagged you on this one because I know you're using it too. I think moving the font size of 1 rem to the root shouldn't hurt your implementation since `rem-calc(14)` seems overridden later anyways. Moving this to the root lets us override it easier in parent components

src/pages/Lend/Filter/FilterSectionGenderPillToggle.vue
```diff
@@ -1,11 +1,10 @@
...
-			:selected="genderOptionSelected"
-			@pill-toggled="genderPillToggled"
-		/>
-	</div>
+	<kv-pill-toggle
+		class="gender-pill-toggle"
+		:options="genderOptions"
+		:selected="genderOptionSelected"
+		@pill-toggled="genderPillToggled"
+	/>
```
ksekhri: Just some tidying up

PR #765, opened 2019-04-11
Add ability to accept/pass basketId in these shop queries.

mcstover:


mcstover:
This should be all of them. 2 were somewhat new files that came in through an older poc branch and 1 existing.

PR #766, opened 2019-04-11
Restrict image height on lyml image

mcstover:


src/components/Checkout/DonationItem.vue
```diff
@@ -27,7 +27,7 @@
 				<span
 					class="donation-amount"
 					v-kv-track-event="['basket', 'Edit Donation']"
-					@click.prevent="enterEditDonation"
+					@click="enterEditDonation"
```
mcstover: try to ensure event propagation is fully allowed.

src/components/Checkout/DonationItem.vue
```diff
@@ -482,6 +482,7 @@ export default {
 		cursor: pointer;
 		font-weight: $global-weight-highlight;
 		font-size: $medium-text-font-size;
+		min-height: 2rem;
```
mcstover: ensure the clickable element always has height

PR #771, opened 2019-04-12
CASH-793: Hide Categories, add Attributes and Sectors

ksekhri:


src/pages/Lend/Filter/FilterSectionAttributes.vue
```diff
@@ -0,0 +1,41 @@
...
+export default {
+	components: {
+		FilterMenuSection,
+		AisRefinementList,
+	},
+	data() {
+		return {
+			visibleThemes: {
+				'conflict zone': true,
+				'clean energy': true,
```
mcstover: There is only 1 of these in our current indexes...

src/pages/Lend/Filter/FilterSectionAttributes.vue
```diff
@@ -0,0 +1,41 @@
...
+				'displaced populations': true,
+				'rural exclusion': true,
+				'social enterprise': true,
```
mcstover: This doesn't exist in our current indices. "Small Enterprise" does though...

ksekhri: Let's merge this then we can update this list as necessary

PR #774, opened 2019-04-13
Dep updates 04132019

mcstover:


mcstover:
@emuvente This wipes out some of those green keepers and all of the in-range. The apollo-client update to 2.5.x seemed to work but threw warnings due to the deprecation of apollo-link-state. Pinned to the latest 2.4.x release and am planning to update and refactor our client state resolvers in a future pr. Also pinned webpack to 4.28.4 for now as moving to 4.30.0 requires some other new peer deps.

PR #781, opened 2019-04-16
AUTH-85 sync www login state with main

emuvente:


src/util/KvAuth0.js
```diff
@@ -76,12 +54,34 @@ export default class KvAuth0 {
...
 		this[sessionPromise] = this.webAuth.then(webAuth => new Promise((resolve, reject) => {
 			webAuth.checkSession({}, (err, result) => {
-				this[authCallback](err, result).then(resolve).catch(reject);
+				if (err) {
+					this[setAuthData]();
+					if (err.error === 'login_required') {
+						// User is not logged in, so continue without authentication
+						cookieStore.set('kvls', 'o', { secure: true });
+						resolve();
+					} else if (err.error === 'consent_required' || err.error === 'interaction_required') {
```
mcstover: What might trigger this clause?

emuvente: These are possible errors that can be returned from silent authentication (where no login form is actually displayed, the user is just redirected or hidden iframes are used). From the auth0 [docs](https://auth0.com/docs/api-auth/tutorials/silent-authentication):

- `consent_required`: The user was logged in at Auth0, but needs to give consent to authorize the application
- `interaction_required`: The user was logged in at Auth0 and has authorized the application, but needs to be redirected elsewhere before authentication can be completed; for example, when using a [redirect rule](https://auth0.com/docs/rules/redirect).

I don't think we'll ever encounter the `consent_required` error, and we may even want to log or report it if it comes up, as that should only be an error for third-party apps.
We might possibly use a redirect rule, so it might be possible that we encounter the `interaction_required` error once we implement that.

PR #785, opened 2019-04-17
CASH-802: Add JS smooth scrolling

ksekhri:
This approach should work in more browsers than CSS based `smooth-scroll` and it lets us set the animation time (this timing is much more pleasant in my opinion). If we like it we can easily add an `xPosition` scrolling option in the future as well.

### DIFF
--- a/src/plugins/smooth-scroll-mixin.js
+++ b/src/plugins/smooth-scroll-mixin.js
@@ -1,36 +1,24 @@
 export default {
 	methods: {
-		smoothScrollTo({ yPosition, millisecondsToAnimate = 250, stops }) {
+		smoothScrollTo({ yPosition, millisecondsToAnimate = 200 }) {
 			const startingPosition = window.scrollY;
 			const endTime = (new Date()).getTime() + millisecondsToAnimate;
-			const numberOfStops = stops || Math.ceil(millisecondsToAnimate / 2);
 			const yDifference = startingPosition - yPosition;
 
-			let animationComplete = false;
+			const scroll = () => {
+				const currentTime = (new Date()).getTime();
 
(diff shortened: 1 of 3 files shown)


src/plugins/smooth-scroll-mixin.js
```diff
@@ -0,0 +1,36 @@
...
+			const startingPosition = window.scrollY;
+			const endTime = (new Date()).getTime() + millisecondsToAnimate;
+			const numberOfStops = stops || Math.ceil(millisecondsToAnimate / 2);
+			const yDifference = startingPosition - yPosition;
+
+			let animationComplete = false;
+
+			for (let i = 0; i < numberOfStops; i += 1) {
+				// eslint-disable-next-line no-loop-func
+				setTimeout(() => {
```
mcstover: I think we should probably convert this setTimeout to use requestAnimationFrame instead. You may also want to create an instance of either setTimeout/requestAnimationFrame that you can stop/cancel/clear out when done.

Some links:
https://flaviocopes.com/requestanimationframe/
https://stackoverflow.com/questions/38709923/why-is-requestanimationframe-better-than-setinterval-or-settimeout

mcstover:
@ksekhri I think we should move in this direction. Can you please grab the copy changes from SelectedRefinements and LendFilterMenu and pull them over here...or isolate them in their own PR.

src/plugins/smooth-scroll-mixin.js
ksekhri: Very cool, I'll do some research

ksekhri:
@mcstover I moved this to `window.RequestAnimationFrame`

PR #788, opened 2019-04-19
CASH-804: Cleanup margins on lend/filter

ksekhri:
* Note: This doesn't affect how `/lend-by-category` looks in my testing, only affects `/lend/filter`

### DIFF
--- a/src/components/LoanCards/AlgoliaLoanCardAdapter.vue
+++ b/src/components/LoanCards/AlgoliaLoanCardAdapter.vue
@@ -143,9 +143,3 @@ export default {
 	}
 };
 </script>
-
-<style lang="scss">
-.algolia-loan-card-adapter {
-	padding: 0;
-}
-</style>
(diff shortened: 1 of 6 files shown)


src/components/LoanCards/AlgoliaLoanCardAdapter.vue
```diff
@@ -143,9 +143,3 @@ export default {
 	}
 };
 </script>
-
-<style lang="scss">
```
mcstover: Removing this probably affects the grid loan card display on /algolia-vue but seeing as that's all going to get a re-work soon it shouldn't be an issue.

src/components/LoanCards/ListLoanCard.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="list-loan-card row">
+	<div class="row list-loan-card">
```
mcstover: whoa now, risky biz lol

ksekhri: Hahahaha

PR #790, opened 2019-04-19
Initial implementation of global experiment prefetch and assignment. …

mcstover:
…Demo of how that can be used on LendFilter page along with the update of an experiment version based on a temporary legacy exp cookie.


mcstover:
@emuvente What are you thoughts on the changes to server-entry.js and the new src/util/experimentPreFetch.js file? This is only acting on 1 new experiment at the moment and I'm not set on keeping the assignment portion active yet due to some of the cross stack stuff.

But, this could be part of a way forward as it puts both the setting and assignment into the cache so it's available for any operations in the created method of components.

emuvente:
How about doing it in a custom prefetch function in the WwwPage component?

mcstover:
@emuvente Yea that seems to work too!

PR #791, opened 2019-04-19
CASH-822: Correct loan cards for expired loans

ksekhri:
* There are still some questions left on the ticket but this is done for now

### DIFF
--- a/src/components/LoanCards/LoanCardController.vue
+++ b/src/components/LoanCards/LoanCardController.vue
@@ -118,6 +118,9 @@ export default {
 		isSelectedByAnother() {
 			return this.amountLeft <= 0 && !this.isFunded;
 		},
+		isExpired() {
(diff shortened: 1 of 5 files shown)


src/components/LoanCards/LoanCardController.vue
```diff
@@ -118,6 +118,9 @@ export default {
 		isSelectedByAnother() {
 			return this.amountLeft <= 0 && !this.isFunded;
 		},
+		isExpired() {
+			return this.loan.status === 'expired';
+		},
```
ksekhri: This is the core of this PR. I'm not able to find an expired loan to test on my VM, does this look correct to you?

mcstover: Yes that should do it.

ksekhri:
@mcstover Cool, done

PR #792, opened 2019-04-20
Ensure we use basketId currently set in cookie. This resolves situati…

mcstover:
…ons wherein initial prefetch queries were using an old basketId.


mcstover:
@emuvente As you suggested in a prior attempt, ensuring we use the basketId from the current cookie does the trick. With this setup, a page refresh will fix any basketId mismatch.

FYI @ksekhri

PR #798, opened 2019-04-23
CASH-778 Update countdown timer for loan reservations

foochris:


mcstover:
@foochris Looks like a good start. I do like that you were able to condense some of the message switching logic too.

Next step is to animate the time. One approach to this would be an interval setup in the mounted() hook that repeatedly calls the "reservationMessage" method.

foochris:
@mcstover There was a bug in the message switching logic; if minute = 1 as it would never reach the last minute displayed message. Will add animation Add `setInterval` with 1000 ms delay?

mcstover:
You can use the setInterval. make sure to clear it on component destroy.

foochris:
@mcstover `setInterval()` added to `mounted()` and `clearInterval()` called when destroyed or timer at 0.

PR #802, opened 2019-04-23
removed some console logs and an unused event watcher in braintree

mcstover:


src/components/Checkout/BraintreeCheckout.vue
```diff
@@ -321,19 +321,6 @@ export default {
 						return;
 					}
 
-					// Watch for validity change on hosted field inputs
-					hostedFieldsInstance.on('validityChange', event => {
```
mcstover: No need for this seeing as we've got other error handling in place.

src/pages/AlgoliaPOC.vue
```diff
@@ -327,35 +322,8 @@ export default {
 			// console.log(`New Search State: ${JSON.stringify(newDisjunctiveFacets)}`);
 			return newDisjunctiveFacets;
 		},
-		getRefinementCounts() {
```
mcstover: never used. too janky to get accurate data.

PR #808, opened 2019-04-24
Move lend filter experiment handling to mixin and apply to the lend/f…

mcstover:
…ilter and /lend-by-cat/##loan-channel## page. This allows activation of the lend filter experiment when arriving by redirect from /lend/##loan-channel## pages.


mcstover:
No new code here guys, just moved some stuff to a mixin so it could be reused elsewhere.

PR #812, opened 2019-04-26
Add mapping for most loan channel urls to show similar results on /le…

mcstover:
…nd/filter. Update the LoanChannelCategoryPage to build it's filter url with the algolia facet params if the lend filter experiment is active. Refined handling of redirects and assignment for category pages.

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -49,6 +49,7 @@ import loanChannelQuery from '@/graphql/query/loanChannelDataExpanded.graphql';
 import experimentQuery from '@/graphql/query/lendByCategory/experimentAssignment.graphql';
 import updateAddToBasketInterstitial from '@/graphql/mutation/updateAddToBasketInterstitial.graphql';
 import lendFilterExpMixin from '@/plugins/lend-filter-page-exp-mixin';
+import loanChannelQueryMapMixin from '@/plugins/loan-channel-query-map';
 import WwwPage from '@/components/WwwFrame/WwwPage';
 import LoanCardController from '@/components/LoanCards/LoanCardController';
 import KvPagination from '@/components/Kv/KvPagination';
@@ -109,6 +110,7 @@ export default {
 	inject: ['apollo'],
 	mixins: [
 		lendFilterExpMixin,
+		loanChannelQueryMapMixin,
 	],
 	metaInfo() {
 		return {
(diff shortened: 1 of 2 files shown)


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -274,6 +276,12 @@ export default {
 	},
 	mounted() {
 		this.updateLendFilterExp();
+		// check for newly assigned bounceback
+		const redirectFromUiCookie = cookieStore.get('redirectFromUi') || '';
+		if (redirectFromUiCookie === 'true') {
+			cookieStore.remove('redirectFromUi');
+			this.$router.push(this.getAlgoliaFilterUrl());
+		}
```
mcstover: This covers the scenario when visitor with no 'lend_filter" exp assignment clicks the 'filter' link on a new /lend-by-cat/some-cat page. At that time they are redirected to the legacy /lend/some-cat page and may be assigned into the experiment. If so they bounce back here so we catch that and display the /lend/filter page with query params for their intended filters applied. This is an auxillary behavior enabled by the new getAlgoliaFilterUrl method and mapping of old loan channel sets to new algolia facet params.

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -289,6 +297,27 @@ export default {
 			if (!_isEqual(this.$route.query, this.urlParams)) {
 				this.$router.push({ query: this.urlParams });
 			}
+		},
+		getAlgoliaFilterUrl() {
```
mcstover: This method builds the algolia compatible facet params if we have them in the new map file.

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -301,7 +330,9 @@ export default {
 		next();
 	},
 	beforeRouteLeave(to, from, next) {
-		if (typeof window !== 'undefined' && to.path.indexOf('/lend/') !== -1) {
+		if (typeof window !== 'undefined'
+			&& to.path.indexOf('/lend/') !== -1
+			&& to.path.indexOf('/lend/filter') === -1) {
```
mcstover: new check so we DO NOT set this cookie if they are going to /lend/filter. Only when directed to legacy /lend/some-cat.

src/plugins/loan-channel-query-map.js
```diff
@@ -0,0 +1,316 @@
+export default {
```
mcstover: New Map of existing loan channel query params converted to new Algolia facet query params where possible.

This file covers all of the urls that are shown on the lend menu, category grid, all categories page and lend by category page. Almost all of them can be recreated using new Algolia facet params.

Also note some urls aren't supported yet and should probably have fallbacks setup similar to the 'livestock' url which uses a facet we don't have yet.

ksekhri: This is awesome!

src/pages/Lend/LoanChannelCategoryPage.vue
ksekhri: Nice

PR #817, opened 2019-04-29
Add search box and update layout with sidebar of components

mcstover:


mcstover:
FYI @ksekhri I'm updating the Algolia Vue page with a bunch of raw algolia components including the search box. This is so we can start to experiment with the behavior of the searchbox in isolation.

PR #818, opened 2019-04-29
Add loan details from lend/filter

mcstover:


mcstover:
@ksekhri Just adding this so we can interact with every facet we're currently using on lend/filter

PR #822, opened 2019-04-30
Add fallback content for missing loan channel name and description.

mcstover:


mcstover:
@BoulderBrains here's the second part of that pr. Forgot to add the file.

PR #823, opened 2019-04-30
Add perf timing to pageview

mcstover:


src/plugins/kv-analytics-plugin.js
```diff
@@ -20,15 +20,24 @@ export default Vue => {
 				fromUrl = window.location.origin + from.fullPath;
 			}
 
-			// Snowplown pageview
+			// Snowplow pageview
```
BoulderBrains: I prefer snowplown 😉

src/plugins/kv-analytics-plugin.js
```diff
@@ -128,7 +137,10 @@ export default Vue => {
 	Vue.prototype.$fireServerPageView = () => {
 		const to = { path: window.location.pathname };
 		const from = { path: document.referrer };
-		kvActions.pageview(to, from);
+		// delay pageview call to ensure window.performance.timing is fully populated
+		setTimeout(() => {
+			kvActions.pageview(to, from);
+		}, 500);
```
emuvente: I feel like this should be some sort of dom-interactive check if we want to ensure that it's populated

src/plugins/kv-analytics-plugin.js
```diff
@@ -128,7 +137,22 @@ export default Vue => {
...
+		}, 1000);
+		document.onreadystatechange = () => {
+			// fire on complete if not already fired
+			if (document.readyState === 'complete') {
+				if (!pageviewFired) {
+					clearInterval(fallbackPageview);
+					kvActions.pageview(to, from);
+				}
+			}
+		};
```
mcstover: @emuvente Sometimes the 'complete' state can take a while so I use that as default but have a timeout for a fallback just incase it hangs we can still get the pv event.

mcstover: Also, I tried a readystate of 'interactive' but then we lose some of the additional measurement on the performance.timing object.

emuvente: I think that the only things that are loading between when this function gets called and the 'complete' event are stylesheets and images, right? My concern with using a fallback or waiting for 'complete' is that the user might navigate somewhere else before either of those happen.

BoulderBrains:
bitmoji

PR #827, opened 2019-04-30
CASH-867: Add a more prominent pathway to /lend from /lend/filter

ksekhri:


src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -155,6 +158,8 @@ export default {
 			userId: '',
 			filterMenuOpen: false,
 			selectedCustomCategories: {},
+			// eslint-disable-next-line
+			topMessage: `Welcome to Kiva's new filter page! Take it for a spin below, or <a v-kv-track-event="['Lending', 'click-return-classic', 'Exit-CASH-545-2019']" href="https://www.kiva.org/lend?kexpn=lend_filter.lend_filter_versions&kexpv=c">return to the classic view</a> at any time.`,
```
ksekhri: @mcstover will `v-kv-track-event` work here with v-html?

src/components/Kv/KvMessage.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+	<div class="kv-message message-text text-center message-text-confirmation" v-if="safeMessage">
+		<div class="kv-message-row">
+			<img class="kv-message-image" src="~@/assets/images/beta-icon.svg">
+			<div class="kv-message-message">
+				<p class="message" v-html="safeMessage"></p>
```
mcstover: Please make this a slot instead of using v-html

src/pages/Lend/Filter/LendFilterPage.vue
mcstover: I think you need to attach an on click handler like we do for clicking Advanced Filters. Clicking on Advanced Filters bubbles up and then fires the "exitLendFilterExp" method from the lend-filter-exp.mixin.js file.

All of those operations also need to be called with your click. In other words you need to do four things:
Update this: `this.lendFilterExpVersion = 'c';`
Call this:	`this.updateLendFilterExp();`
Track your event...
Then perform your redirect.

These first two steps are necessary to ensure the continuity of the opted out state of the experiment.

ksekhri: Thanks Matt, missed some steps there

src/components/Kv/KvMessage.vue
ksekhri: Updated :+1:

ksekhri:
Cool @mcstover, this is now updated :+1:

PR #828, opened 2019-05-01
Cash 771

BoulderBrains:


src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -1,17 +1,14 @@
 <template>
-	<div class="appeal-banner-wrapper" v-if="showAppeal">
-		<div class="appeal-banner">
+	<div class="sitewide-appeal-wrapper" v-if="showAppeal">
+		<div class="appeal-banner sitewide-appeal row">
```
mcstover: Shouldn't need the row class here. It's on the div just below.

mcstover: Also, please remove "appeal-banner" here and move the css for it into "sitewide-appeal"

PR #833, opened 2019-05-02
AUTH-137 - add RegisterSocial for checking TOS box

timledlie:


timledlie:
The UI could probably be better, but it's something.

timledlie:
Merging now to test, feedback still welcome.

timledlie:
When I loaded this page up on dev, the big "Donate to Kiva today and earn a free loan" banner came up which was pretty distracting / misleading for when we're in the middle of registering a user and need them to check the TOS box.  Is there a way to disable those kinds of banners for this page?

emuvente:
That's a good point, and I agree that the banner is pretty distracting. It makes me wonder how much of the frame we actually need for this page. For instance, we could drop the header and footer entirely by removing the www-page tags. We could also copy the minimal frame we use for the auth0 pages over to the ui repo so we could use it for this page and others as well

mcstover:
I think we have a whitelist to prevent the appeal from showing... showing. Just a sec...

mcstover:
@timledlie we'll need to add an exclude for your new page url to the "showAppeal" method inside of AppealBanner.vue.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+	<www-page>
+		<div class="page-content">
+			<form
+				id="registerSocialTermsForm"
+				ref="registerSocialTermsForm"
```
emuvente: This reference isn't being used, so it can probably be dropped

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -0,0 +1,98 @@
...
+				class="promptForm"
+				@submit.prevent.stop="postRegisterSocialForm">
```
emuvente: You can also add `action="."` which makes it so that the submit event is fired when pressing 'enter' or submitting the form in some other way besides clicking the submit button

timledlie:
@mcstover - thanks for the suggestion, I've done https://github.com/kiva/ui/pull/840

src/pages/LoginAndRegister/RegisterSocial.vue
timledlie: Thanks, done with https://github.com/kiva/ui/commit/cf2b65dc1799d4278f3521a78b8d1c199c3b488e

emuvente:
I think we should probably add in a few words to help explain what is happening so the user understands why they're being asked to click a checkbox. Maybe something like
```
<h1>Almost done...</h1>
<p>To complete your registration, please agree to the Terms of Use</p>
```
although there are probably better words than that.

Also, should we show an error message or redirect if someone visits this page without the state parameter?

timledlie:
Okay, added that language (tweaked a bit): https://github.com/kiva/ui/commit/709ed7c9e39d78694b0aea1d39937a57decde966

Erm, I'm not sure how / why anyone would visit this page if they weren't in the login flow, but if you think that's possible, I'm happy to add a redirect to the home page or whatever...  How would I code that in the page -- before loading, check if a query param is set and redirect if not?

timledlie:
Also @emuvente, if someone did attempt to reach register/social with no state and then submitted the page, they would get to `https://login.kiva.org/continue?agree=yes&state=undefined` with an error message `Unauthorized`, so that seems fine.

timledlie:
@emuvente - https://github.com/kiva/ui/commit/626c4d1f42ca07ad848c4eef5ca00019e49014ba
I don't need a `return` after the router push, do I?

emuvente:
@timledlie no, that looks good 👍

PR #837, opened 2019-05-03
CASH-849: Add pagination size control to loans page

foochris:


src/pages/Lend/Filter/FilterComponents/LendFilterMenu.vue
```diff
@@ -6,6 +6,9 @@
 				<div>Filter and sort</div>
 				<kv-icon class="filter-toggle-chevron" name="large-chevron" />
 			</div>
+			<div id="algolia-pagination-stats">
+				<algolia-pagination-stats :padding="2" class-names="show-for-small-only"/>
```
ksekhri: Can we remove padding here?

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -44,12 +44,15 @@
 						@remove-custom-category="removeCustomCategory"
 						@clear-custom-categories="clearCustomCategories"
 					/>
+
+					<algolia-pagination-stats :padding="2" class-names="hide-for-small-only" />
```
ksekhri: Is padding needed?

foochris:
@ksekhri I've updated the PR to reflect your requested changed and the views should match the Zeplin mockups.

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -318,6 +322,14 @@ export default {
 						opacity: 0.2;
 					}
 				}
+
+				.algolia-pagination-stats-component {
+					display: none;
+
+					@include breakpoint(1194px) {
+						display: block;
```
ksekhri: I believe this needs to be namespaced into `filter-menu-pinned` to prevent it from displaying when the pinned experiment is inactive?

mcstover:
@foochris I've spoken with @ksekhri on this and it's good to merge for now. There are likely some issues with the new stats positioning that we'll need to pin-point once this is in dev and fix up before next release.

src/pages/Lend/Filter/LendFilterPage.vue
ksekhri: We can merge this and make tweaks later as per Matt

foochris: Will merge. Thanks.

PR #840, opened 2019-05-06
flesh out whitelist of routes where we don't show banner

timledlie:


mcstover:
@timledlie 'blacklist' probably does make more sense. Also a few lint need to be fixed up.

timledlie:
Good call on blacklist vs whitelist.  Fixed.

Wow, I managed to make 4 lint violations with three lines of changes!  I like the Travis job a lot.  Fixed those too, and even learned a bit about var vs let vs const in the meantime.

PR #842, opened 2019-05-06
Ensure we have the required String! type on the query definition

mcstover:


mcstover:
FYI @emuvente Just getting things into master too.

emuvente:
👍

PR #843, opened 2019-05-06
Extra gaurd to skip processing legacyExp data if it's empty

mcstover:


mcstover:
The usage of legacyExpData will eventually be removed. This just cleans up the computed property so it doesn't try to parse the data before it arrives.

PR #845, opened 2019-05-07
Create new process for turning on a specific experiment version using…

mcstover:
… a url query parameter. Passing the setuiab query param with an 'active' exp id and version will switch to that version. Also setup prelimiary work for global active experiment fetching and cookie clean up.


mcstover:
@emuvente any thoughts on this?

I need to finalize before branch cut tomorrow

PR #856, opened 2019-05-09
Small tweaks to margin on elements around stats to equalize spacing a…

mcstover:
…cross the standard view as well as upcoming experiments with the pinned filter menu and search bar.


mcstover:
I'll move this into the release branch shortly

PR #861, opened 2019-05-10
Perform async update of authentication state after popup login. Turn …

mcstover:
…off loading state if the popup is closed and after the completion of the async login update.


mcstover:
@emuvente  added a couple ids to help QA/Automation find the correct buttons

PR #866, opened 2019-05-13
Eslint lib updates

mcstover:
eslint updates plus a few other libs too. Had to fix a few newly surfaced errors due to new eslint presents for vue + airbnb. One particular rule we should change is to allow if else conditions which I have just disabled checking for in several places.
 
This leaves about ~684 warnings (100+ in dev mode) for syntax updates, mostly around tab position, putting html element content on it's own line and a few other things.

Once we get this update and error round merged, I would suggest we tackle all of those over the course of a day or 2 this week just to clean em up.


ksekhri:
@emuvente I get why the rule is there, it'll make diffs clearer. We can add `ignoreWhenNoAttributes` and `ignoreWhenEmpty` for the simpler cases.

mcstover:
@emuvente I'm fine with that... These were the ones I had considered turning off the following which cuts the warnings down to 14 from 685 lol...

'vue/html-closing-bracket-spacing': 'off',
'vue/html-closing-bracket-newline': 'off',
'vue/multiline-html-element-content-newline': 'off',
'vue/singleline-html-element-content-newline': 'off',

ksekhri:
@mcstover We should keep `vue/html-closing-bracket-newline`, it makes the code so much more readable

emuvente:
@ksekhri That's a good point about making diffs cleaner, and it looks like `ignoreWhenNoAttributes` and `ignoreWhenEmpty` are true by default, so my example would actually be allowed, my mistake.

@mcstover I think all those others we should keep for readability.

src/components/Checkout/LoanPrice.vue
```diff
@@ -4,10 +4,13 @@
...
-				<option v-for="price in prices"
-					:key="price"
-					:value="price">${{ price }}
+				@change="updateLoanReservation()"
+			>
+				<option v-for="priceOption in prices"
+					:key="priceOption"
+					:value="priceOption"
+				>
+					${{ priceOption }}
```
mcstover: I've updated price to priceOption in this context as we already have a 'price' prop coming in and were getting a warning for having a 2nd price variable declaration/assignment.

mcstover:
@emuvente @BoulderBrains @ksekhri @foochris Here it is, your daily developer eye test. Please take a little time to review.

Hats off to Shua for being bold enough to consider the --fix option. It worked just fine!

Notes:
- I don't really like dangling '>' all over the place, but if you're all cool with it so am I.
- in the relatively near future, we need to update our eslint rule to allow else if statements so we can clean out the disabled blocks around those.

foochris:
@mcstover Although I can rationalize the requirement to have a closing bracket on a new line, I don't agree with it. I think it's too visually jarring when one is reading a line and all of a sudden, there's an orphan character on it's own line. 

I think, however, that the number of lines of markup per file is not large, so having orphan closing bracket(s) shouldn't cause too much developer friction.

emuvente:
I do think it's jarring (and ugly) to have that floating closing bracket, but I do prefer it for clearly indicating where the attributes end and the content begins, as the attributes and content are at the same indentation level, which can make things hard to differentiate when skimming over the code

src/pages/Lend/AlgoliaPaginationWrapper.vue
```diff
@@ -1,14 +1,14 @@
...
-			}"
+			currentRefinement,
+			nbPages,
+			pages,
+			isFirstPage,
+			isLastPage,
+			refine,
+			createURL
+		}"
 		>
```
emuvente: I think this is probably the worst consequence of having the dangling closing bracket.

src/pages/Build/Research.vue
```diff
@@ -5,13 +5,15 @@
 			<div class="small-12 columns">
 				<div v-for="year in years" :key="year">
 					<h3>{{ year }}</h3>
+					<!-- eslint-disable vue/no-use-v-if-with-v-for -->
```
emuvente: We should come back and fix this too. It could use a computed property that returns the papers grouped by year so that the template can be a simple v-for

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBanner.vue
```diff
@@ -37,6 +40,7 @@ export default {
 			} else if (this.promoEnabled) {
```
emuvente: Since these are all returning something, I think in this case this just needs to be an `if` on it's own line, and the linting should pass

src/components/WwwFrame/PromotionalBanner/PromotionalBanner.vue
```diff
@@ -18,12 +18,15 @@ export default {
 		};
 	},
 	computed: {
+		/* eslint-disable */
 		currentActivePromo() {
+			// TODO: Fix lint rule to allow else if
 			if (this.lendingRewardOffered) {
 				return LendingRewardsBanner;
 			} else if (this.bonusBalance > 0) { // TODO: skip if on a checkout/basket page
```
emuvente: This also can just be an `if`, since the previous statement returns, and that should clear up the linting

src/components/WwwFrame/LendMenu/LendMegaMenu.vue
```diff
@@ -74,19 +80,22 @@
...
 			</ul>
 		</div>
 		<kv-expandable property="width" :skip-leave="true">
+			<!-- eslint-disable -->
 			<div
 				v-for="region in regions"
 				:key="region.name"
 				v-if="isOpenSection(region.name)"
 				class="right-section"
 			>
```
emuvente: We should come back to this one soon too

src/components/Checkout/LoanPrice.vue
```diff
@@ -75,7 +78,9 @@ export default {
 		};
 	},
 	computed: {
+		/* eslint-disable */
 		prices() {
+			// TODO: convert to switch statement with default value returned
```
emuvente: We could also change that else if to an if since the first if returns something

src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
```diff
@@ -69,13 +70,15 @@ export default {
 			const maxLength = 1000;
 			const lowerCaseUse = this.use.toString().charAt(0).toLowerCase() + this.use.toString().slice(1);
 			const convertedUse = (this.use.substring(0, this.name.length) === this.name) ? this.use : lowerCaseUse;
-
+			/* eslint-disable */
+			// TODO: Fix eslint rule to allow else if
```
emuvente: I think the rule is just complaining because the else if is not needed when the first if returns something

BoulderBrains:
Catching back up with the conversation here. FYI: I don't have strong opinions with regards to the closing brackets on a new line or the same line as the last of the element. Readability is slightly better with the new line. Overall "look" and length of the files suffer a little bit with the new line for the closing bracket, but it's nothing I can't get use to. All the other changes look good. Big ups for the "--fix option" making short work on these changes.

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBanner.vue
mcstover: resolved

src/components/WwwFrame/PromotionalBanner/PromotionalBanner.vue
mcstover: resolved

src/components/Checkout/LoanPrice.vue
mcstover: resolved

src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
mcstover: resolved

mcstover:
@emuvente I think this is ready

BoulderBrains:
👀

PR #867, opened 2019-05-13
Add prompt option for Auth0 popup login flow so we can force the logi…

mcstover:
…n form to show if the lender is not actively logged in.

### DIFF
--- a/src/util/KvAuth0.js
+++ b/src/util/KvAuth0.js
@@ -90,7 +90,7 @@ export default class KvAuth0 {
 	}
 
 	// Open a popup window to the login page
-	popupLogin() {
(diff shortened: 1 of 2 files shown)


src/util/KvAuth0.js
```diff
@@ -90,7 +90,7 @@ export default class KvAuth0 {
 	}
 
 	// Open a popup window to the login page
-	popupLogin() {
+	popupLogin(prompt) {
```
emuvente: How about doing this as a options object that is spread after any defaults we want to set? like
```
webAuth.popup.authorize({
	prompt: 'none'
	popupOptions: { .. },
	...options
}
```

PR #872, opened 2019-05-14
Begin clean up featured loan experiments so we now only show the full…

mcstover:
… width card. Update the Featured loan wrapper with processing to show or fetch and show new loans if the initial loan is funded.

remove recently viewed loans component

preserve featured loan variants for now


BoulderBrains:
Late to the party, thanks for the tag. 👀

PR #878, opened 2019-05-15
CASH-916: Sort order gets overridden on prod

ksekhri:


src/plugins/algolia-init-mixin.js
```diff
@@ -14,13 +14,25 @@ const indexKeys = [
 
 // Rebuild index name as set in Algolia
 function setSortByEnv(selectedRouteSort) {
-	let expandedRouteSort = '';
-	['dev', 'qa', 'prod'].forEach(env => {
-		if (typeof window !== 'undefined' && window.location.host.indexOf(env) !== -1) {
```
ksekhri: I noticed this was only happening on prod, this line was checking for prod.kiva.org as the host name

src/plugins/algolia-init-mixin.js
```diff
@@ -14,13 +14,25 @@ const indexKeys = [
...
+			envName: 'dev',
+			hostNameField: 'dev',
+		},
+		{
+			envName: 'qa',
+			hostNameField: 'qa',
+		},
+		{
+			envName: 'prod',
+			hostNameField: 'www.kiva.org',
```
ksekhri: I'm of two minds on this. I made it very specific (I believe you will always get a `www` before `kiva.org` as your host name. We can also make this more of a default case, replace `hostNameField` with something like `''` or `'kiva.org'`. What do you think @mcstover

ksekhri:
@mcstover updated with prod being a default if none of the others match :+1:

PR #879, opened 2019-05-16
Fix post-auth team redirect

emuvente:
Login sync was forcing users to go through the post-auth flow twice, causing the double team invite. Also, certain cookies being missing caused a new session to be generated, which would cause the invite handling to be skipped.

### DIFF
--- a/server/util/getSessionCookies.js
+++ b/server/util/getSessionCookies.js
@@ -1,10 +1,20 @@
 const fetch = require('./fetch');
 const setCookieParser = require('set-cookie-parser');
 
+const getCookieString = cookies => {
+	return Object.keys(cookies)
+		.map(key => `${key}=${cookies[key]}`)
+		.join(';');
+};
+
 module.exports = function getSessionCookies(url = '', requestCookies = {}) {
 	return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


server/util/getSessionCookies.js
```diff
@@ -1,10 +1,20 @@
...
+
 module.exports = function getSessionCookies(url = '', requestCookies = {}) {
 	return new Promise((resolve, reject) => {
 		if (url.length && (!requestCookies.kv || !requestCookies.kvis || !requestCookies.kvbskt)) {
-			fetch(url).then(res => {
+			fetch(url, {
+				headers: {
+					Cookie: getCookieString(requestCookies),
+				},
+			}).then(res => {
```
mcstover: What else might this block be helping with?

ebalcomb:
I don't really understand what's happening

server/util/getSessionCookies.js
mcstover: I suspect this will help keep the sessions in sync and that may help some of the other issues I had seen this week. Will report back after this is out in dev.

emuvente: Any situation where we want to look at cookies that last longer than the session during the session startup process. It seems like something related to the experiment cookies or something similar might be helped by this.

PR #884, opened 2019-05-16
Cleanup header when Algolia search is enabled

ksekhri:
* When Algolia search is enabled, currently the header shows 2 borders to the right of Lend. Hiding the `top-nav-search-area` when the search icon is hidden fixes this.

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -25,7 +25,7 @@
 				<kv-icon v-show="searchOpen" class="close-icon" name="x" />
 			</button>
 			<div class="flexible-center-area">
-				<div id="top-nav-search-area" :aria-hidden="searchOpen ? 'false' : 'true'">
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -25,7 +25,7 @@
 				<kv-icon v-show="searchOpen" class="close-icon" name="x" />
 			</button>
 			<div class="flexible-center-area">
-				<div id="top-nav-search-area" :aria-hidden="searchOpen ? 'false' : 'true'">
+				<div id="top-nav-search-area" v-if="!hideSearchInHeader" :aria-hidden="searchOpen ? 'false' : 'true'">
```
mcstover: I know I approved, but should we use a v-show instead here... Just wondering if there are js events and bindings that may break if it's missing from the dom.

ksekhri: @mcstover I did some testing toggling between /lend-by-category and /lend/filter and I searched through the code for references to the id `top-nav-search-area `. I feel we can keep this an if for now.

mcstover:
👍

PR #887, opened 2019-05-20
Use v-if to prevent initialization of braintree or paypal until that …

mcstover:
…component is visible. guard against null error logs + ensure props are passed down for content switching.

### DIFF
--- a/src/components/Checkout/PaymentWrapper.vue
+++ b/src/components/Checkout/PaymentWrapper.vue
@@ -10,7 +10,7 @@
 			</div>
 			<div>
 				<pay-pal-exp
-					v-show="selectedOption === 'pp'"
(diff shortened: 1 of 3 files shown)


src/components/Checkout/PaymentWrapper.vue
```diff
@@ -19,7 +19,7 @@
 			</div>
 			<div>
 				<braintree-checkout
-					v-show="selectedOption === 'bt'"
+					v-if="selectedOption === 'bt'"
```
mcstover: We could potentially go either way with this one, this setup ensure that each component is only initialized when it's needed.

BoulderBrains: That makes sense, no need to initialize unless we need it. 👍

PR #890, opened 2019-05-21
Add basketId variables to braintree mutation

mcstover:


mcstover:
@leoperry We patch in the basketId but hadn't made the variable available here :( Thank you for helping debug!

leoperry:
@mcstover Ah, that's cool to see! Thanks for the FYI.

src/components/Checkout/BraintreeCheckout.vue
```diff
@@ -448,9 +449,10 @@ export default {
 			this.apollo.mutate({
 				mutation: braintreeDepositAndCheckout,
 				variables: {
+					basketId: cookieStore.get('kvbskt'),
```
emuvente: I'm still concerned about adding in the basket id here when it should already be being added by the BasketLink, but theoretically this shouldn't cause problems.

PR #893, opened 2019-05-22
CASH-935: Update name of lend/filter experiment

ksekhri:


ksekhri:
Updated removing URL changes for /lend

PR #895, opened 2019-05-22
Ui routes endpoint

mcstover:


mcstover:
@emuvente still working out details of how to format dynamic routes (id. /lend-by-category/:category) for the proxy and if we need to delineate specific child routes or just format them as dynamic (ie. /styleguide/whatever...)

mcstover:
@emuvente This should get us going for some initial tests.

PR #896, opened 2019-05-22
hide basket bar on join team popup

mcstover:


ebalcomb:
sorry I was late on this, thanks so much for taking care of it!

PR #901, opened 2019-05-25
Ensure global Vue instance is clean when imported in the app

emuvente:
This removes the cached Vue module before the server-bundle is processed, ensuring that the next call to import Vue will return a fresh instance with no plugins installed yet (which will resolve the InvariantError caused by the apollo-plugin being installed multiple times)


emuvente:
Technique taken from http://derpturkey.com/reload-module-with-node-js-require/

PR #909, opened 2019-05-31
reduce max loan use length and set shorter use length for expandable …

mcstover:
…cards.

### DIFF
--- a/src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
+++ b/src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
@@ -50,6 +50,10 @@ export default {
 			type: String,
 			default: ''
 		},
+		maxUseLength: {
+			type: Number,
+			default: 500
+		},
 	},
 	computed: {
 		helpedLanguage() {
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/BorrowerInfo/BorrowerInfoBody.vue
```diff
@@ -50,6 +50,10 @@ export default {
 			type: String,
 			default: ''
 		},
+		maxUseLength: {
+			type: Number,
+			default: 500
```
ksekhri: Why did we reduce the default to 500 for GridLoanCard and ListLoanCard?

mcstover: Just wanted to put a more "sane" default in case there are real long loan use statements out there it would keep it from pushing that area too far down on the list item card.

mcstover: Checked my local db and there were 0 with 300+ characters or more, 97 with 200 chars or more.

PR #911, opened 2019-05-31
Fix sort order

ksekhri:


ksekhri:
@mcstover updated stage config too

PR #915, opened 2019-06-04
Remove unused styles

ksekhri:
These styles aren't applied since the SASS is `scoped`. I added them manually in prod and none of them made a difference to the styling except for `padding: 0`, which if actually applied broke the image altogether.


ksekhri:
This is used actually! Deleting PR.

PR #924, opened 2019-06-06
alternate graphql url for api debugging

mcstover:


mcstover:
FYI @ksekhri @BoulderBrains @foochris  Using this alternate url will trigger the debug in PHPStorm. (may require extra config in PHPStorm)

foochris:
Cool. Did the request for this feature from others?

mcstover:
I've been investigating the graphql batch requests and had to figure this out so I could debug the api code.

PR #926, opened 2019-06-07
CASH-1019: Make Kv-Flag Scalable

ksekhri:
* Required for HoverLoanCard

### DIFF
--- a/src/pages/BorrowerProfile/fundedBorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/fundedBorrowerProfile.vue
@@ -28,7 +28,7 @@
 				<!-- Borrower location -->
 				<div>
 					<span class="country-image">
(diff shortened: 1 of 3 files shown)


src/pages/BorrowerProfile/fundedBorrowerProfile.vue
```diff
@@ -28,7 +28,7 @@
 				<!-- Borrower location -->
 				<div>
 					<span class="country-image">
-						<kv-flag :country="`${ loan.geocode.country.isoCode }`" />
+						<kv-flag :country="`${ loan.geocode.country.isoCode }`" :custom-height="14" />
```
ksekhri: `customHeight` is a backwards compatibility feature. The flags are actually square with white bars on top and bottom, but we cut off the white pixels at the bottom of the flag historically.

PR #930, opened 2019-06-10
Fix a visual bug from mixing classes

ksekhri:
* When a child component has a bound class, a class placed on that child by the parent is ignored
* This was worked around using a wrapper div


ksekhri:
cc @BoulderBrains

PR #934, opened 2019-06-11
CASH-955: Hover Loan Cards Experiment

ksekhri:
Settings Manager:
```
uiexp.hover_loan_cards
string
{"name": "HoverLoanCards","enabled": false,"startTime": "2018-08-29T00:00","endTime": "2022-02-16T15:00","control": {"key": "variant-a","name": "Control (variant-a)"},"variants": {"variant-b": {"name": "Hover Loan Cards"}},"distribution": {"variant-a": 0.5,"variant-b": 0.5}}
CASH-521: Hover Loan Cards
```

### DIFF
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -366,6 +366,30 @@ export default {
 				);
 			}
 		},
+		initializeHoverLoanCard() {
+			// CASH-521: Hover loan card experiment
+			const hoverLoanCardExperiment = this.apollo.readQuery({
+				query: experimentQuery,
+				variables: { id: 'hover_loan_cards' },
+			});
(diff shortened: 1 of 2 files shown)


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -529,6 +535,29 @@ export default {
 				'b',
 			);
 		}
+
+		// CASH-521: Hover loan card experiment
```
mcstover: Could you please move all of this setup into it's own method similar to how initializeFavoriteCountryRowExp is handled?

emuvente: +1

ksekhri: @mcstover @emuvente Cool, encapsulated the code :+1:

PR #935, opened 2019-06-11
Cash 912

BoulderBrains:


BoulderBrains:
Only remaining issue is the console error that's thrown on page load: "Uncaught (in promise) TypeError: Cannot read property 'totalCount' of null"
Which I believe points to line 121 in AgoliaLoanCardAdapter

mcstover:
@BoulderBrains I think that totalCount error will go away once we're in a real environment. It's likely a mismatch of loans between Algolia Dev and trying to fetch the corresponding data from your local db.

PR #943, opened 2019-06-17
Query and pass down index level list of location.lvl1 facet values. U…

mcstover:
…se this list to persist location select options when other facets (attributes, tags, gender, etc)  are used for filtering.

### DIFF
--- a/src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
+++ b/src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
@@ -47,6 +47,7 @@
 				:checked="isCountryChecked(item.isRefined, item.region)"
 				:label="`${item.label} (${item.count})`"
 				@checkbox-input="handleCountryCheckboxInput(item)"
+				:disabled="item.count === 0"
 			/>
 		</div>
 	</div>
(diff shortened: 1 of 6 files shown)


mcstover:
Initial setup is here and seems to work well. This does not yet set the options with 0 loans to "disabled".

mcstover:
FYI @emuvente @BoulderBrains @foochris

BoulderBrains:
👀

src/plugins/algolia-init-mixin.js
```diff
@@ -184,5 +189,33 @@ export default {
...
+
+		// initialize utility instance of default index
+		this.defaultIndexInstance = this.searchClient.initIndex(this.algoliaDefaultIndex);
+
+		// set global data set for Lvl1 Locations
+		this.setAllLvl1Locations();
 	},
+	methods: {
+		setAllLvl1Locations() {
+			this.defaultIndexInstance.searchForFacetValues({
```
ksekhri: Cool solution :+1:

src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
```diff
@@ -47,6 +47,7 @@
 				:checked="isCountryChecked(item.isRefined, item.region)"
 				:label="`${item.label} (${item.count})`"
 				@checkbox-input="handleCountryCheckboxInput(item)"
+				:disabled="item.count === 0"
```
ksekhri: Nice

PR #944, opened 2019-06-17
Require graphql queries to be named

emuvente:
Assists with performance/usage monitoring, and mocking for testing


emuvente:
We don't have to, but I think using names will allow us to do things like refetching queries by name as opposed to providing the entire query to refetch

PR #948, opened 2019-06-19
encode doneurl query params

mcstover:


mcstover:
@emuvente This seems to do the trick for ui. However legacy login doesn't appear to preserve the query params at all. Also, I'm not really sure we should have to do this at all. The page is resolved, the query params should already be collected on the session...

PR #950, opened 2019-06-19
cash-1037 fixing up Cannot read property 'donationTotal' of undefined

BoulderBrains:
FYI: @mcstover

### DIFF
--- a/src/components/Checkout/DonateRepaymentsToggle.vue
+++ b/src/components/Checkout/DonateRepaymentsToggle.vue
@@ -54,7 +54,7 @@ export default {
 		this.apollo.watchQuery({ query: initializeCheckout }).subscribe({
 			next: ({ data }) => {
 				this.myDonateRepayments = _get(data, 'my.userAccount.donateRepayments');
-				this.totals = _get(data, 'shop.basket.totals');
+				this.totals = _get(data, 'shop.basket.totals') || {};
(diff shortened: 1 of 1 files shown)


src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -54,7 +54,7 @@ export default {
 		this.apollo.watchQuery({ query: initializeCheckout }).subscribe({
 			next: ({ data }) => {
 				this.myDonateRepayments = _get(data, 'my.userAccount.donateRepayments');
-				this.totals = _get(data, 'shop.basket.totals');
+				this.totals = _get(data, 'shop.basket.totals') || {};
```
mcstover: This won't fully fix the situation as you'll now have a totals object but will still get undefined below (line 71).

The root cause of the donationTotal and loanReservation undefined errors is that we don't have a basket. SO... I think it's worth looking into not rendering basket items at all if we don't have a response on "shop" from the graphql.

emuvente: `totals` is initialized as an empty object, so it should always be fine to set it to an empty object when we don't have totals. Line 71 will work fine because `numeral(undefined).value() === 0`

BoulderBrains: Thanks @mcstover, Shua and I just spoke about this and I filed this ticket to fix the root of the this issue (cash-1038). I believe that we still want to merge this change in, but will work on cash-1038 tomorrow.

PR #952, opened 2019-06-20
prevent error from being logged when pop up login window is closed. T…

mcstover:
…his will clean up this issue so we can focus on the actual errors.


mcstover:
@emuvente @BoulderBrains Wanted to try and extract a real error string so they get separated out when reported.

PR #953, opened 2019-06-20
CASH-960: Add support for end cards

ksekhri:
Additions:
* Add code for rearranging cards when expanding first card in row and template for rearranging when expanding last card in a row

Fixes:
* Increases code reuse
* Cleanup comments and transitions
* Fixes horizontal scrollbar appearing intermittently in row

Known Issues/Limitations:
* Currently moving from the first card in the row to the second card causes one card to be skipped, this will be fixed in the subsequent PR

### DIFF
--- a/src/components/LoansByCategory/CategoryRowHover.vue
+++ b/src/components/LoansByCategory/CategoryRowHover.vue
@@ -42,6 +42,8 @@
 					:style="{ marginLeft: scrollPos + 'px' }"
 					v-touch:swipe.left="scrollRowRight"
 					v-touch:swipe.right="scrollRowLeft"
+
+					ref="hoverCardsHolder"
 				>
 					<loan-card-controller
 						loan-card-type="HoverLoanCard"
@@ -64,7 +66,7 @@
 						@update-detailed-loan-index="updateDetailedLoanIndex"
 						@update-hover-loan-index="updateHoverLoanIndex"
 
-						ref="loanCards"
+						ref="hoverLoanCards"
 					/>
 					<!--
 						Blocks of attributes above:
@@ -105,6 +107,7 @@
 import _get from 'lodash/get';
 import _throttle from 'lodash/throttle';
 import LoanCardController from '@/components/LoanCards/LoanCardController';
+import categoryRowArrowsVisibleMixin from '@/plugins/category-row-arrows-visible-mixin';
 
 const cardWidthPlusPadding = 200;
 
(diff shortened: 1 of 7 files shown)


src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -236,6 +242,38 @@ export default {
...
+			const hoverCardDistanceFromLeft = hoverLoanCardLeft - hoverCardsHolderLeft;
+
+			return hoverCardDistanceFromLeft === 0;
+		},
+		hoverLoanIsRightMost() {
+			if (this.noHoverLoan) {
+				return false;
+			}
+
+			return false;
```
mcstover: I expect more to come here later...

src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -236,6 +242,38 @@ export default {
...
+			const arrowWidthInRem = 2.5;
+			const pxInRem = 16;
+			const leftPadding = this.categoryRowArrowsVisible()
+				? arrowWidthInRem * pxInRem
+				: noArrowPaddingSide * pxInRem;
+			const hoverCardsHolderLeft = hoverCardsHolderRect.left - bodyRect.left + leftPadding;
+
+			const hoverCardDistanceFromLeft = hoverLoanCardLeft - hoverCardsHolderLeft;
+
+			return hoverCardDistanceFromLeft === 0;
```
emuvente: Is 0 the only distance that qualifies or is there a range of distances that are ok? For instance if the distance is less than 1 rem.

mcstover:
@ksekhri This build is green if you check the link.

src/components/LoansByCategory/CategoryRowHover.vue
ksekhri: Hahahaha I don’t know, it looks nice so minimal ;)

src/components/LoansByCategory/CategoryRowHover.vue
ksekhri: Interesting question. I believe the scrolling behavior we use (scrolling by a multiple of the width of a card + margin-right) makes it so that it should always be zero but let me think about this and test it more. We can easily make this more lax if there’s a chance of outliers.

ksekhri: Rereading the code there shouldn't be a non-zero case. I'll continue to keep an eye on this and test :+1:

PR #964, opened 2019-06-26
Add legacy ga script to ensure pickup of gtm params across stacks.

mcstover:


src/plugins/kv-analytics-plugin.js
```diff
@@ -55,6 +58,17 @@ export default Vue => {
...
 			}
 
+			if (gaAltLoaded) {
+				let gaAltId;
+				/* eslint-disable no-underscore-dangle */
+				if (window.__KV_CONFIG__ && window.__KV_CONFIG__.gaAlternateId) {
+					gaAltId = window.__KV_CONFIG__.gaAlternateId;
+				}
+				window._gaq.push(['_setAccount', gaAltId]);
+				window._gaq.push(['_trackPageview']);
```
mcstover: If that config wasn't available these would fail silently.

PR #976, opened 2019-07-03
Mock graphql server for testing

emuvente:
The mocks themselves still need work, but this gets the server setup to allow for them. Read more about mocking graphql with apollo here: https://www.apollographql.com/docs/apollo-server/features/mocking/ and more about using lunar for mocking through cypress here: https://github.com/mjfaga/react-boston-2018-lunar-launch


src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
```diff
@@ -128,8 +128,17 @@ export default {
 		query: appealBannerQuery,
 		preFetch: true,
 		result({ data }) {
-			this.appealEnabled = JSON.parse(_get(data, 'general.appeal_enabled.value', false));
-			this.appealMatchEnabled = JSON.parse(_get(data, 'general.appeal_match_enabled.value', false));
+			try {
+				this.appealEnabled = JSON.parse(_get(data, 'general.appeal_enabled.value', false));
```
mcstover: @emuvente Did you see these blocks failing when working on setting this up?

src/util/experimentPreFetch.js
```diff
@@ -104,7 +104,11 @@ export function fetchAllExpSettings(config, apolloClient, route) {
 		// Check for active experiments listing
 		const activeExperimentsSettings = _get(results, 'data.general.activeExperiments');
 		if (typeof activeExperimentsSettings !== 'undefined' && activeExperimentsSettings !== null) {
-			activeExperiments = JSON.parse(activeExperimentsSettings.value).split(',');
+			try {
+				activeExperiments = JSON.parse(activeExperimentsSettings.value).split(',');
```
mcstover: Same ?

src/components/WwwFrame/EndOfYearAppealBanner/AppealBanner.vue
emuvente: Yeah these were failing with the mock data because apollo server defaults to using `"Hello, World!"` as the value for all `String` type fields. Obviously we should make the mock data better so that doesn't happen, but it did highlight how this code can fail if the data is in an unexpected shape.

emuvente:
@BoulderBrains The documentation is a little sparse, but the main repo is here: https://github.com/ezcater/lunar-core. I'd recommend looking at the examples in the lunar-launch repo I linked earlier, which includes examples showing how to start with a regular apollo server with mocking turned on and end up with a server that uses lunar to setup & teardown mocks

BoulderBrains:
Thanks for that @emuvente

PR #1003, opened 2019-07-15
CASH-1094: Borrower Image Component

ksekhri:


src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -137,6 +150,20 @@ export default {
...
 		},
+		loan() {
+			return this.apollo.readFragment({
+				id: this.loanId,
+				fragment: detailedLoanCardFragment,
+			});
+		},
+		retinaImageUrl() {
+			// eslint-disable-next-line quotes
+			return this.loan.image.retina.replace(`/w960h600/`, `/w960h720/`);
```
ksekhri: Since the taller image isn't in the cache, I ran a `String.replace()`

src/graphql/fragments/loanCardFields.graphql
```diff
@@ -4,6 +4,7 @@ fragment loanCardFields on LoanBasic {
 	name
 	borrowerCount
 	geocode {
+		city
```
ksekhri: My understanding was I needed to add `city` here if I wanted to access it from the GQL cache

emuvente: Since this fragment is used in a query for fetching the loans for the category rows, this will indeed populate the cache with that value. However, as we build more of this loan card we will probably want separate queries so that we only fetch the extra info we need for the card when we're displaying it

src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -137,6 +150,20 @@ export default {
...
+		},
+		standardImageUrl() {
+			// eslint-disable-next-line quotes
+			return this.loan.image.default.replace(`/w480h300/`, `/w480h360/`);
```
emuvente: This can fail if `this.loan` is ever null or undefined

src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -137,6 +150,20 @@ export default {
 		hasTrustee() {
 			return false;
 		},
+		loan() {
+			return this.apollo.readFragment({
```
emuvente: I like that you put this readFragment in a computed value!

src/graphql/fragments/loanCardFields.graphql
mcstover: Agree. We're currently fetching a lot of unused information for the initial rows with these tiny loan cards.

src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
ksekhri: Oh yes, nice eye. I'm putting in a couple of guards before merging :+1:

PR #1008, opened 2019-07-17
Cash 964

BoulderBrains:
Loan Story (loanDescription) loading on the story tab for the hover loan card. Loading state present.


src/components/LoanCards/HoverLoanCard/InfoPanels/BorrowerStoryPanel.vue
```diff
@@ -3,31 +3,90 @@
...
+	},
+	created() {
+		this.apollo.query({
+			query: loanDescriptionQuery,
+			variables: {
+				id: parseInt(this.loanId, 10)
+			}
+		}).then(({ data }) => {
+			this.loanStory = _get(data, 'lend.loan.description');
+		});
```
emuvente: Instead of this custom created function, you should take advantage of the apollo plugin so that this component will also update when the loan id changes
```
apollo: {
  query: loanDescriptionQuery,
  result({data}) {
    this.loanStory = _get(data, 'lend.loan.description');
  },
}
```
}

mcstover: Sorry @emuvente @BoulderBrains had apollo setup initially but the variables on 'this.' weren't coming through... so we moved it. I think he later found the issue was higher up and the var wasn't being passed into the component.

PR #1010, opened 2019-07-17
CASH-1095: Add shared loan object for our reference

ksekhri:


ksekhri:
@mcstover for reference, just making sure we're all using same loan object

emuvente:
I'd rather that we let the info panels each do their own querying/readFragment calls based on the passed loan id

ksekhri:
@emuvente let's discuss offline @BoulderBrains

mcstover:
@emuvente I can see that being useful for the Partner and Loan Detail panels but overkill for the Overview/Image + Story panel...

ksekhri:
Closing this in favor of #1012

PR #1024, opened 2019-07-19
CASH-1096: Wrap DetailedLoanCard in LoanCardController

ksekhri:
* We need many of the methods and computed values from LCC to complete the `DetailedLoanCard`

#### Note: 
A known UI bug exists in this change: When the DetailedLoanCard is open and the user hovers over another loan, the card closes and reopens. This is a bug we will have to chase down.


ksekhri:
@emuvente Heck yeah, nice fix

PR #1021, opened 2019-07-19
cash-1097 work toward matching text

BoulderBrains:

PR #1026, opened 2019-07-20
CASH-1096: Add FundraisingStatusLarge

ksekhri:


mcstover:
@ksekhri We should merge in https://github.com/kiva/ui/pull/1021/files seeing as it's been sitting there waiting...

PR #1027, opened 2019-07-20
Cleanup DetailedLoanCard

ksekhri:
* Cleanup location/sector on tablet
* Cleanup image on desktop
* Support full width, vertically centered images to prep for tablet


ksekhri:
@mcstover Hey man thanks for the review! I just pushed a solid update to the image to allow full width with vertical centering.

PR #1028, opened 2019-07-22
CASH-1093 expanded card spacing & style cleanup

emuvente:


src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -275,11 +274,11 @@ export default {
 		},
 		retinaImageUrl() {
 			// eslint-disable-next-line quotes
-			return _get(this.loan, 'image.retina', '').replace(`/w960h600/`, `/w960h720/`);
+			return _get(this.loan, 'image.retina', '').replace(`/w960h600/`, `/w1096h822/`);
```
mcstover: @emuvente Will these get requested "only" when the Overview section is shown?

We should be careful not to initialize new image sizes that will apply to all loans shown on the lbc page. If it's a very limited number (aka only when the image is shown) then it's probably ok for the experiment.

emuvente: Yeah, these are only loaded when the expanded loan card is shown

mcstover: Cool thank you!

PR #1031, opened 2019-07-23
cash-1124 adjusting margins and placements

BoulderBrains:
a couple more spacing adjustments that came down from Adam

### DIFF
--- a/src/pages/Search/SentenceSearch.vue
+++ b/src/pages/Search/SentenceSearch.vue
@@ -321,17 +321,25 @@ export default {
 	border-bottom: 1px solid $kiva-bg-darkgray;
 }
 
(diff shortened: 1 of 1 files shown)


src/pages/Search/SentenceSearch.vue
```diff
@@ -342,28 +350,30 @@ export default {
...
+		margin-bottom: rem-calc(16);
 	}
 }
 
 .ais-MenuSelect select {
 	border: none;
 	background-color: transparent;
 	color: $kiva-accent-blue;
-	font-size: 1.5rem;
+	font-size: 1.375rem;
```
mcstover: I have a feeling this size will be added to stylguide. It's also used a lot in the donation lightbox.

PR #1033, opened 2019-07-24
CASH-1067: Allow loan card rows to overlap page content

ksekhri:
From a user’s standpoint, cards now only expand when you enter the card body, not the card margin. This is implemented as follows:
* Hovering in margins will now only trigger card expansion when another card is already expanded
* To expand a card in a row where no other cards are expanded yet, you must actually enter the card body

### DIFF
--- a/src/components/LoansByCategory/CategoryRowHover.vue
+++ b/src/components/LoansByCategory/CategoryRowHover.vue
@@ -19,7 +19,7 @@
 						{{ cleanName }}
 					</template>
 				</h2>
-				<p v-if="showCategoryDescription" class="category-description show-for-large">
+				<p class="category-description show-for-large">
 					{{ loanChannel.description }}
 				</p>
 			</div>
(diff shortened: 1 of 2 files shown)


src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -19,7 +19,7 @@
 						{{ cleanName }}
 					</template>
 				</h2>
-				<p v-if="showCategoryDescription" class="category-description show-for-large">
+				<p class="category-description show-for-large">
```
ksekhri: I disabled this `v-if` since even if we don't have a description, we need a min height to allow clicking on the row title. This change is only for rows containing hover cards.

mcstover: Should be ok, may want to verify no errors are thrown on something without a description. If so you could throw a check in where we try to use "loanChannel.description"

PR #1057, opened 2019-07-31
CASH-1142, CASH-1087: Sync margins when expanding DetailedLoanCard and add scroll when expanding

ksekhri:
* Sync margin animation of DetailedLoanCard to be same as HoverLoanCard expansion
* On mobile: scroll to DetailedLoanCard when switching loans or initially expanding DetailedLoanCard
* On desktop: scroll to CategoryLoanRowHover only when initially expanding DetailedLoanCard

### DIFF
--- a/src/components/LoansByCategory/CategoryRowHover.vue
+++ b/src/components/LoansByCategory/CategoryRowHover.vue
@@ -107,8 +107,8 @@
 			>&rsaquo;</span>
 		</div>
 
-		<kv-expandable>
-			<div v-if="detailedLoanCacheId">
+		<kv-expandable :delay="150" easing="linear">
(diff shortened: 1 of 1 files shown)


src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -411,6 +431,27 @@ export default {
 		handleSetPreventUpdatingDetailedCard(newState) {
 			this.setPreventUpdatingDetailedCard(newState);
 		},
+		runningOnServer() {
```
mcstover: You should be able to just use `this.$isServer` to check this...

ksekhri: Nice, that's better! Updated :+1:

PR #1062, opened 2019-08-01
CASH-976: Track Desktop Tabs

ksekhri:
* Generate tabs from an array to reduce code repetition
* Track clicking on tabs on desktop


src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -24,42 +24,14 @@
 		<div class="main-panel columns small-12 medium-5 small-order-2 medium-order-1">
 			<div class="info-tab-selector show-for-medium">
 				<button
+					v-for="{title, id} in tabs"
```
mcstover: I like that these are dynamic now. You might want to create a computed version of the tab data so you can filter for active/inactive sections based on the loan type.

ksekhri: That’s a cool idea!

PR #1110, opened 2019-08-02
filter links for navigating between enabled panels

mcstover:


ksekhri:
@mcstover I'm going to merge this, need it for some bug testing

PR #1120, opened 2019-08-06
hover loan card panel tweaks

BoulderBrains:


BoulderBrains:
@mcstover Here are the tweaks I was making for the hotfix tomorrow. The timeOnKiva, NaN error in safari doesn't seem to be working, BUT the odd thing is, if I got to the dateFns docs site and enter this in the console: "dateFns.differenceInCalendarMonths(Date.now(), '2015-02-09T12:50:02Z');" I get back a valid # of months in the console. I took the timestamp from the vue inspector in chrome, so that's the exact format coming in. I'm not sure why this would be working in the console but not working in my code where the same thing is happening, perhaps we can talk through this more tomorrow morning? I also tried to install the vue inspector devtools in safari, but from what I read safari isn't supported yet. If I've got that wrong and you know a way to get it installed in safari, let me know.

BoulderBrains:
FYI: this is a draft PR for now.

BoulderBrains:
Cool, well I tested this out on my computer this morning and it appears to be working now 🧙‍♂️Would you be willing to pull this branch down and then verify you're getting a valid Time On Kiva value on the partner and or trustee tab of the hover loan card in Safarii? Or we can put it up on dev and I can retest it there, either way.

PR #1125, opened 2019-08-06
Cleanup functional events

mcstover:


mcstover:
code approved here: https://github.com/kiva/ui/pull/1122
Couldn't get Travis cache to clear fully so this build would pass. Still failing on the missing lib that we've since added to the config.

PR #1127, opened 2019-08-07
Kubernetes deploy ui

jmallatt-kiva:
@mcstover @emuvente 
Will likely want to re-introduce npm tests, etc.


go-bears:
🍿 !

config/k8sdev.js
```diff
@@ -0,0 +1,56 @@
+var merge = require('webpack-merge')
+var base = require('./index.js')
```
emuvente: We could use './dev.js' as the base, and then the rest of this could just be
```
module.exports = merge(base, {
	server: {
		memcachedServers: 'k8sdev-elasticache.bu9ifv.cfg.usw1.cache.amazonaws.com:11211',
	}
})
```

mcstover: @jmallatt-kiva  I'll make this update.

PR #1128, opened 2019-08-07
Cash 1181

BoulderBrains:


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -362,12 +350,11 @@ export default {
 		},
 		showPayPal() {
```
mcstover: We won't need this method any longer as we'll never show just paypal...

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -362,12 +350,11 @@ export default {
...
 			return parseFloat(this.creditNeeded) > 0
-			&& (this.braintreeExpVersion === 'control' || this.braintree === false);
+			&& this.braintree === false;
 		},
 		showBraintree() {
 			return parseFloat(this.creditNeeded) > 0
-				&& this.braintree === true
-				&& this.braintreeExpVersion === 'shown';
+				&& this.braintree === true;
```
mcstover: I think we should probably remove the Ui setting for braintree and just set it to true always...

BoulderBrains:
@mcstover this is ready for another review.

PR #1133, opened 2019-08-09
Update vue meta

mcstover:


mcstover:
FYI @ksekhri & @BoulderBrains

mcstover:
Guess I could have just used the Greenkeeper PR... instead of creating a new one after I tested this locally...

PR #1137, opened 2019-08-12
update css loader to 3.2.0

mcstover:


mcstover:
@emuvente Tested in dev and prod mode. Things look fine.

Couldn't get the greenkeeper pr to reset with the new travis config so just created a new one.

PR #1139, opened 2019-08-13
cash-1167 most of the exp setup, need to figure out why selectedOptio…

BoulderBrains:
…ns isn't being passed to payment wrapper

### DIFF
--- a/src/components/Checkout/PaymentWrapper.vue
+++ b/src/components/Checkout/PaymentWrapper.vue
@@ -65,6 +65,8 @@ export default {
 					key: 'pp',
 				},
 			],
+			// Uncomment this selectedOption line to set a default payment option
+			// bt = braintree, pp = paypal
 			// selectedOption: 'bt',
 		};
 	},
--- a/src/components/Checkout/PaymentWrapper.vue
+++ b/src/components/Checkout/PaymentWrapper.vue
@@ -50,7 +50,7 @@ export default {
 		},
 		selectedOption: {
 			type: String,
-			default: ''
+			default: 'bt'
 		}
 	},
 	data() {
@@ -65,9 +65,6 @@ export default {
 					key: 'pp',
 				},
 			],
-			// Uncomment this selectedOption line to set a default payment option
-			// bt = braintree, pp = paypal
-			// selectedOption: 'bt',
 		};
 	},
 	methods: {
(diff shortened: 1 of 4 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -202,6 +204,8 @@ export default {
 			redirectLightboxVisible: false,
 			teams: [],
 			holidayModeEnabled: false,
+			braintreeExpVersion: null,
```
mcstover: this should be braintreeVsPaypalVersion

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -202,6 +204,8 @@ export default {
...
+			selectedOption: null,
```
mcstover: don't need this as it's now computed > translateSelectedOption

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -337,6 +356,12 @@ export default {
 		showKivaCreditButton() {
 			return parseFloat(this.creditNeeded) === 0;
 		},
+		translateSelectedOption() {
+			if (this.braintreeExpVersion === 'control' || this.braintreeVsPaypalVersion === 'a') {
```
mcstover: you can drop the this.braintreeExpVersion check here...

src/components/Checkout/PaymentWrapper.vue
```diff
@@ -47,6 +47,10 @@ export default {
 		amount: {
 			type: String,
 			default: ''
+		},
+		selectedOption: {
+			type: String,
+			default: ''
```
mcstover: please set the default to 'bt' we should ensure that is always provided and we just switch if the exp assignment differs.

mcstover: Also this is just to ensure this component can work without a prop being passed.

src/components/Checkout/PaymentWrapper.vue
```diff
@@ -61,7 +65,9 @@ export default {
 					key: 'pp',
 				},
 			],
-			selectedOption: 'bt',
+			// Uncomment this selectedOption line to set a default payment option
```
mcstover: these commented lines can all be removed if you have a default in the prop!

PR #1148, opened 2019-08-20
cash-1219 new ui homepage skeleton

BoulderBrains:


src/pages/Homepage/Homepage.vue
```diff
@@ -0,0 +1,30 @@
...
+
+<script>
+import WwwPage from '@/components/WwwFrame/WwwPage';
+
+export default {
+	components: {
+		WwwPage
+	},
+	metaInfo: {
+		title: 'Sitemap'
```
emuvente: For the homepage I don't think we need to specify any title and we can just let the default be used

BoulderBrains: My bad, I just remove this copy/paste error.

PR #1159, opened 2019-08-28
Add client-side kount integration + update some loading sequencing.

mcstover:


mcstover:
Fixed new setting key and added to dev,qa,stage (true) + prod (false)

BoulderBrains:
👀

leoperry:
@mcstover: Kicking off a dev minor right now... to enable this, do I need to do anything other than set up the settings manager flag and set it to enabled?

mcstover:
@leoperry settings should be in place so you should be good to go.

PR #1160, opened 2019-08-29
CASH-1243 Update Braintree checkout mutation definition to include deviceData for Kount

leoperry:
I don't really know if this breaks anything else! I didn't see any other references to this mutation or to `braintreeDepositAndCheckout.graphql` outside the Braintree checkout component.


leoperry:
But I did scratch my head for a while trying to understand why I could see the device data on the client side, but then it was always null server side... fortunately the Apollo developer tool in Chrome helped me see that the device data ultimately wasn't being sent down to the server!

PR #1164, opened 2019-08-30
Cash 1217

BoulderBrains:


src/components/Homepage/CategoryGrid.vue
```diff
@@ -0,0 +1,192 @@
...
+				v-for="loanChannel in loanChannels" :key="loanChannel.id"
+				class="small-6 medium-4"
+				v-kv-track-event="[
+					'homepage',
+					'click-Category-card',
+					`${loanChannel.name}-category-card`,
+					'true']"
+			>
+				<div class="category-card-overlay">
+					<div>
```
emuvente: I think this div can be dropped since the .category-card-overlay div is already wrapping it

src/components/Homepage/CategoryGrid.vue
```diff
@@ -0,0 +1,190 @@
...
+<style lang="scss" scoped>
+@import 'settings';
+
+.heading-text {
+	padding: 3rem;
+	font-weight: 300;
+}
+
+.get-started-link {
+	display: contents;
```
emuvente: `contents` doesn't have good browser support yet, and I'm not sure if it's needed for this element since it doesn't have any children

PR #1166, opened 2019-09-03
cash-1217 fixing category card links

BoulderBrains:
This is going to be brought up to production once reviewed.


mcstover:
Why are you looking to change this? For most loan channels (aka categories) there is a ui page and that page should handle any urls that don't work...

PR #1167, opened 2019-09-03
cash-1247 if user logins from homepage, redirect them to the login URL

BoulderBrains:
Tested this out on my vm and it does work. Weird side effect: pop-up is shown really quickly before the redirect to the /login url. 

popup is shown, then closed and user is left on login page.


emuvente:
I think you can address the popup showing briefly by changing the header's `auth0Login` method to check `this.showPopupLogin` instead of `this.kvAuth0.enabled` in its if statement

PR #1172, opened 2019-09-09
Cash 1208

BoulderBrains:


src/components/LoanCards/HoverLoanCard/InfoPanels/LoanDetailsPanel.vue
```diff
@@ -42,10 +42,21 @@
 						{{ fundingModelFormatted }}
 					</p>
 				</li>
-				<li v-if="this.currencyExchangeLoss">
+				<li v-if="this.currencyLossLiability === 'shared'
```
mcstover: I think this can move to just wrap the label v-if the first group and just else the '' and none options... This way you won't have 2 instances of this li...

BoulderBrains: Thanks for that @mcstover, that's all cleaned up now.

BoulderBrains:
Just double verified, but the v-else in place for the currency exchange loss covers the needed case.

PR #1173, opened 2019-09-10
Initial set up of experiment to show favorite sector loan on the feat…

mcstover:
…ured card.

### DIFF
--- a/src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
+++ b/src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
@@ -3,7 +3,13 @@
 		<div class="row">
 			<div class="column small-12">
 				<h2 class="section-name">
-					Featured: Research-backed impact
+					Featured:
+					<template v-if="featuredSectorExpVersion === 'shown'">
+						{{ loanChannel.name }}
+					</template>
+					<template v-else>
+						Research-backed impact
+					</template>
 				</h2>
 				<p v-if="showCategoryDescription" class="section-description show-for-large">
 					{{ loanChannel.description }}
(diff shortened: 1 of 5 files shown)


mcstover:
Waiting on final copy for EXP title and description.

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -65,12 +65,19 @@ export default {
...
+			type: String,
+			default: null
+		}
 	},
 	data() {
 		return {
 			experimentData: {},
-			featuredCategoryIds,
-			initialLoanCount,
+			featuredCategoryIds: [56],
```
mcstover: These were never getting updated to have a value in the code below...So we'll just go with the hardcoded value for the default loan channel and the global limit.

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -80,72 +87,94 @@ export default {
...
-			this.loading = false;
+		if (this.favoriteSectorId !== null && this.featuredSectorExpVersion === 'shown') {
+			// set initial loan data so we ssr with a loan if possible
+			const loanArray = _get(allLoanData, 'lend.loans.values');
+			this.setInitialLoan(loanArray);
+			// display custom channel data
+			this.loanChannel = {
+				id: 999,
+				description: 'Some special text',
+				name: 'Some special title',
```
BoulderBrains: :shipit:

src/graphql/query/featuredLoansData.graphql
```diff
@@ -25,5 +26,21 @@ query featuredLoansData(
 				}
 			}
 		}
+		loans(
```
emuvente: Would it be much effort to do this as a separate query and select which loan query to call based on the experiment assignment?

mcstover: I think longer term I'd like to have a data provider we could alter params on to switch queries. But for this one, it was just more simple to run both to get to ensure this loan was rendered from the server using the prefetch.

PR #1181, opened 2019-09-17
cash-1326 changing btVSpp exp name

BoulderBrains:
Tested this out on my VM and it was still working with the new exp name. I also moved the new exp setting into the settingsManager on dev, qa & production. I'll deleted the old exp setting after verification of this work in dev.

### DIFF
--- a/src/components/Checkout/PaymentWrapper.vue
+++ b/src/components/Checkout/PaymentWrapper.vue
@@ -69,6 +69,7 @@ export default {
 		};
 	},
 	created() {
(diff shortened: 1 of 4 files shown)


src/components/Checkout/PaymentWrapper.vue
```diff
@@ -69,6 +69,7 @@ export default {
 		};
 	},
 	created() {
+		console.log('selectedOption', this.selectedOption);
```
mcstover: please remove

PR #1185, opened 2019-09-18
Add initial delete saved payment method functionality

mcstover:
As the VaultManager delete method is not currently functional, this approach compares payment method information from our backend api with corresponding information from the VaultManager.fetchPaymentMethods function. We can match the card targed for deletion and then use our backend api to delete the card.


mcstover:
Will add styles and the pop up soon.

BoulderBrains:
👀 Just swinging through to take a look.

mcstover:
@emuvente this is ready for review

PR #1190, opened 2019-09-22
Add hashCode function for use in creating the uiab cookie

mcstover:
This function will produce a one-way hash representing the value of a passed string.

### DIFF
--- a/src/api/localResolvers/experiment.js
+++ b/src/api/localResolvers/experiment.js
@@ -1,6 +1,7 @@
 import _isUndefined from 'lodash/isUndefined';
 import _filter from 'lodash/filter';
 import _map from 'lodash/map';
+import _pick from 'lodash/pick';
 import _toPairs from 'lodash/toPairs';
 import { isWithinRange } from 'date-fns';
 import cookieStore from '@/util/cookieStore';
@@ -159,8 +160,11 @@ export default () => {
 
 					// read the experiment data from the cache
 					const experiment = readJSONSetting(context, `cache.data.data['Setting:uiexp.${id}'].value`);
+					// create targeted subset of experiment setting to use in hash
+					// Changing the Name, Distribution, Variants or Control values will "reset" an experiment assignment
+					const experimentSubset = _pick(experiment, ['name', 'distribution', 'variants', 'control']);
 					// get the hash for our current experiment setting
-					const settingHash = hashCode(JSON.stringify(experiment));
+					const settingHash = hashCode(JSON.stringify(experimentSubset));
 
 					// Add hash to exisitng cookie exps if it's missing
 					if (typeof currentAssignment.hash === 'undefined') {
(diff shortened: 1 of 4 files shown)


src/util/settingsUtils.js
```diff
@@ -59,3 +59,29 @@ export function settingEnabled(data, enabledKey, startTimeKey, endTimeKey) {
...
+	// return 0 if empty or incorrect type
+	if (typeof source === 'undefined' || typeof source !== 'string' || source === '') {
+		return hash;
+	}
+	// shift the bits
+	for (let i = 0; i < source.length; i++) { // eslint-disable-line
+		// isolate character
+		const character = source.charCodeAt(i);
+		// original approach left shift
+		hash = ((hash << 5) - hash) + character; // eslint-disable-line no-bitwise
```
emuvente: Bit shifting is going to be a lot more efficient, so I think you can go ahead and disable the no-bitwise check for the whole file

test/unit/specs/util/settingsUtils.spec.js
```diff
@@ -0,0 +1,30 @@
...
+		});
+
+		it('converts a simple string into a 32bit integer', () => {
+			const simpleString = 'some string';
+			expect(hashCode(simpleString)).toEqual(1395333309);
+		});
+
+		it('converts a string including digits into a 32bit integer', () => {
+			const simpleString = 'some string 12345';
+			expect(hashCode(simpleString)).toEqual(-472388848);
```
emuvente: Should we maybe have hashCode return the absolute value so that we don't have to encode the negative sign as well? That will increase the chance of collision, but hopefully not by much.

src/api/localResolvers/experiment.js
```diff
@@ -137,64 +155,71 @@ export default () => {
 			Query: {
 				experiment(_, { id }, context) {
 					// get the existing assigned version for this experiment id
-					let version = assignments[id];
+					let currentAssignment = assignments[id] || {};
 
 					// read the experiment data from the cache
 					const experiment = readJSONSetting(context, `cache.data.data['Setting:uiexp.${id}'].value`);
+					// get the hash for our current experiment setting
+					const settingHash = hashCode(JSON.stringify(experiment));
```
emuvente: Should we maybe only hash the weights? I'm thinking that we don't necessarily want to force a reassignment just because dates change or the experiment gets disabled

mcstover: hmmm... That's prob a good call...I'll work on excluding the date change and enabled field. The cookie cleaner with remove disabled experiments as they won't be in the active list anymore.

I'll think about how to add a way to trigger a general reassignment of the experiment to cover allowing those fields to remain. Maybe just a simple incrementing value field in the settings object...

mcstover:
@emuvente Updated

PR #1192, opened 2019-09-25
cash-1313 exp initial setup, couple kinks to workout

BoulderBrains:


src/pages/Homepage/HeroSlideshow.vue
```diff
@@ -50,7 +50,32 @@
 				&nbsp;
 			</div>
 		</div>
-		<div class="action-button-wrapper">
+		<div
```
mcstover: @BoulderBrains you shouldn't need to duplicate this markup. You can just drop a span after the copy in each button `Sign up<span v-if="showDoubleArrowButton"> >></span>`

src/pages/Homepage/HeroSlideshow.vue
```diff
@@ -82,6 +107,7 @@ const mgPromoImageRequire = require.context('@/assets/images/mg-hero-slideshow/'
 
 export default {
 	name: 'HeroSlideshow',
+	// I need to return the props.doubleArrowButtonExp like the following example is doing for mgPromoExp
```
mcstover: For this you can add some logic above the return statement to determine which of your 4 cases should be the cacheKey. Make sure to give a unique key for your 2 new cases...then just return you final key.

src/pages/Homepage/Homepage.vue
```diff
@@ -41,6 +51,13 @@ export default {
 				this.mgPromoExp.version === 'shown' ? 'b' : 'a'
 			);
 		}
+		if (this.doubleArrowButtonExp.version !== null) {
+			this.$$kvTrackEvent(
```
mcstover: 1 $ no 2 $$

BoulderBrains: Mo money, Mo problems.

src/pages/Homepage/HeroSlideshow.vue
BoulderBrains: 👍

PR #1194, opened 2019-09-25
Disable popup login for now

emuvente:


mcstover:
@emuvente I think this will be fine for now. It should solve the issues with users being confused about the header being updated.

PR #1199, opened 2019-09-27
cash-1282 text update came in hot

BoulderBrains:
Brandon messaged me this morning saying the final text for the donation line exp had been updated since I last looked. This commit gets the final text in for this experiment.


BoulderBrains:
If this looks good to you, can you please merge this into master and I'll jump on dev and verify it later today. When tested on my VM all was working as expected with single loan and multi loan basket scenarios.

PR #1210, opened 2019-10-02
Add lendingRewardTag check to checkout

mcstover:


mcstover:
@emuvente For the time being we need to make sure people tagged to receive a lending reward get pushed to legacy checkout. I'll prob have an analogous kiva pr for this too.

PR #1211, opened 2019-10-02
cash-1364 remove the loanRes10 exp from our code. Exp was deemed not …

BoulderBrains:
…significant

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -46,7 +46,6 @@
 </template>
 
 <script>
-import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import CheckoutItemImg from '@/components/Checkout/CheckoutItemImg';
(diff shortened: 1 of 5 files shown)


src/components/Checkout/BasketItem.vue
```diff
@@ -76,24 +74,9 @@ export default {
 		return {
 			activateTimer: true,
 			loanVisible: true,
-			loanRes20Version: 'control',
-			setTimedMessage: false,
+			loanRes20Version: 'control'
```
mcstover: I think this can go to. it should be unused

src/components/Checkout/BasketItem.vue
```diff
@@ -75,8 +74,7 @@ export default {
...
```
BoulderBrains: I didn't see any other mentions of this loanRes20 exp, I'm thinking I should get rid of this line?

PR #1213, opened 2019-10-03
AUTH-244 Force install idtoken-verifier version 1.2.0

emuvente:


emuvente:
This came from this suggestion https://stackoverflow.com/a/48524488
I moved idtoken-verifier from `'required'` to `'dependencies'` with `"version": "1.2.0"`, then deleted `node_modules` and ran `npm install` which then made the other changes. I think this will only change if we upgrade auth0-js, so we will need to keep an eye out for that happening

PR #1214, opened 2019-10-03
Cash 1369

BoulderBrains:


mcstover:
@BoulderBrains We'll need to run a prefetch on promoQuery in Homepage like we do in the GlobalPromoBanner code. Then we can set and update the the promoEnabled values and show this content based on the promo setting in admin that already powers the banner in legacy.

PR #1223, opened 2019-10-06
Losslessly compress image assets

AndrewCraswell:
Noticed a lot of the site graphics were coming in large. It seems just loading the home page the initial download size on my retina screen is 1.3mb, jumping to 2.6mb after the carousel has cycled once. Tested on a mobile device and this was not much better. This is an initial attempt to just reduce the file sizes a bit, although some other optimizations should occur at some point.

In this PR:
- Compress the SVGs using SVGO CLI for 40-60% average size reduction. Most of these files are sized appropriately for inlining to reduce network requests.
- Compress PNGs, JPGs with FileOptimizer which uses the latest versions of the most optimized lossless compression binaries to get the best lossless compression rate. Total size-reduction amounts to over 500kb.

A few things should be followed up on:
- On the initial homepage load, the first and second hero images load both the "std" and "retina" images.
- The hero images alternate between blurred background, and sharp backgrounds. Standardizing to the former would give better opportunities for lossy compression to have a big impact.
- Several other images are loading twice, including the map .svg
- Hero images could probably have bitrate reduced, compression rate increased for slight lossiness, or have the size reduced to better fit various screen sizes (especially applicable for mobile)
- Images should have `alt` text set to a value or empty for accessibility (counted 11 violations).
- The `Loans By Category` page has loan avatar images compressed well, but should only be loading the first ~50 images that are visible, and lazyloading the remainder off-screen when needed (rather than 100+). A sizing strategy should be implemented as the images loaded are 300% larger than what the UI needs. Currently, I am seeing 10s load times on this page over my super-fast network.
- I saw a PR earlier referencing HTTP/2 CDNs. Not sure if that was targetting the loan avatar images, but it would certainly help when trying to hammer 100+ images requests at once.


mcstover:
@AndrewCraswell 

Thank you for the time and effort put into thinking about our image based performance. There are certainly improvements to be made and we will create some tickets in the backlog to implement some of your suggestions.

We already have svg compression built into our compilation process, trace the webpack base config to see how that works.

As for image assets, we appreciate the suggestions on using a stronger compression and will take care of that on our side.

Thanks again!

AndrewCraswell:
@mcstover Is the backlog public at all? My company donates money to non-profits if we contribute to open source. Might be interested in exploring some of the items planned if it's available

mcstover:
@AndrewCraswell Thanks for getting back in touch. That sounds like a nice program.

At this point our backlog is not public. However, we're working on a process to expose some of the tasks for the project as Github issues. We don't have an immediate timeline for this but will be working on it over the next few months.

PR #1226, opened 2019-10-11
AUTH-247 show tip message for auth0 errors

emuvente:


src/client-entry.js
```diff
@@ -44,6 +45,27 @@ const {
...
 
+// Show a tip message when there is an unhandled auth0 error
+kvAuth0.onError(({ eventId, user }) => {
+	let message = 'We\'re sorry, something went wrong.';
+	if (user) {
+		message = `${message} Please log out and try again.`;
+	} else {
+		message = `${message} Please clear your cookies and try again.`;
+	}
+	if (eventId) {
```
mcstover: Is the eventId only present when in "special" error scenarios?

src/util/KvAuth0.js
```diff
@@ -26,23 +30,20 @@ export default class KvAuth0 {
...
 	}) {
+		this[errorCallbacks] = [];
 		this.enabled = true;
 		this.user = user;
 		this.accessToken = accessToken;
 		this.isServer = typeof window === 'undefined';
 
 		if (!this.isServer) {
-			// delay loading of webauth until we've confirmed we're in the browser,
-			// as it complains about not having window
```
mcstover: Guess this was fixed in some update?

emuvente: Yes, with version 1.4.1 of idtoken-verifier https://github.com/auth0/idtoken-verifier/commit/7e25214d2a610a1597939fa8d113778f8ae5d0d8

src/client-entry.js
emuvente: The event id should be present after any message is sent to sentry. I think the only real case it will not be present is on the vm where sentry is disabled.

PR #1227, opened 2019-10-11
Add setupBasketForUser mutation as first step in checkout prefetch.

mcstover:


mcstover:
Depends on https://github.com/kiva/kiva/pull/7804

PR #1232, opened 2019-10-15
new green color

BoulderBrains:
New green defined to be used for the kv-toggle component


AndrewCraswell:
Just out of curiosity, is there any discussion around accessibility with the green on white theme colors throughout the site? I noticed most of the text doesn't pass color contrast tests... Any chance there might be some plans to update that?

PR #1233, opened 2019-10-16
CASH-1397 Local resolver for handling autolending settings page state

emuvente:


src/api/fixtures/AutolendProfile.js
```diff
@@ -0,0 +1,37 @@
...
+		isEnabled: false,
+		donationPercentage: 15,
+		enableAfter: 0,
+		loanSearchCriteria: LoanSearchCriteria(),
+	};
+}
+
+// Return a cleaned profile suitable for input
+export function getInputProfile(profile) {
+	const cleanProfile = _omit(profile, [
```
mcstover: The lodash docs call out omit as being much slower than pick so in this case it may be better to pick 3 instead of omit 3... but we can wait and see if it's an issue first.

PR #1235, opened 2019-10-17
KvDropdownRounded resuable component

BoulderBrains:


BoulderBrains:
FYI: I've configured the the width of this dropdown to take in any length of text and still have the background-image on the dropdown positioned correctly.

BoulderBrains:
FYI: I started thinking about customizing the look the dropdown when it's open. I know different browsers have their own way of rendering form elements and this one's no different. I did some reading and there's no easy and fast way to make this change. On top of that, I read quite a few places stating that changing the browser defaults for form elements completely destroys form accessibility. For the sake of time and pushing forward, I'm going to leave the default dropdown styles in place (for an open dropdown). https://css-tricks.com/dropdown-default-styling/

PR #1238, opened 2019-10-18
Turn on polling to check filesystem for changes when using NFS throug…

mcstover:
…h a Vagrant VM.


mcstover:
This activates file system polling to get recompilation working in the new Bionic VM. Do you all think 1 sec is too frequent?

@emuvente @BoulderBrains Could one of you pull this branch, restart ui then change a file and see how this affects your setup?

PR #1240, opened 2019-10-21
VUE-164

eddieferrer:
* Added ids and classes to DOM elements in filter section so that they can be targeted more easily by testing automation

### DIFF
--- a/src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
+++ b/src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
@@ -19,6 +19,7 @@
 					/>
 					<div
 						class="region-label"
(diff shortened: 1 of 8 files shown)


src/pages/Lend/Filter/FilterSections/LocationMulti/LocationMultiRefinements.vue
```diff
@@ -30,6 +31,7 @@
 					<div
 						v-if="!item.allCountriesSelected"
 						@click="selectAllCountries(item.countries, item.region)"
+						id="select-all"
```
mcstover: if possible please add the region to this (select-all-##region##) and to the deselection toggle as well.

eddieferrer: Done

PR #1243, opened 2019-10-22
Autolend gender radio

BoulderBrains:
Not yet working, but getting there.


src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,126 @@
...
+					mutation: gql`mutation {
+						autolending @client {
+							editProfile(profile: { isEnabled: ${enabled} })
+						}
+					}`,
+				});
+			}
+		},
+		gender(enable, previouslyEnabled) {
+			if (enabled !== previouslyEnabled) {
```
emuvente: For these you probably want to use `gender` and `previousGender`

src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,126 @@
...
+					mutation: gql`mutation {
+						autolending @client {
+							currentProfile {
+								loanSearchCriteria {
+									loanSearchFilters {
+										editProfile(profile: { gender: ${gender} })
+									}
+								}
+							}
+						}
```
emuvente: I think for this query you want
```
autolending @client {
	editProfile(profile: {
		loanSearchCriteria: {
			filters: {
				gender: ${gender}
			}
		}
	}
}
```

src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,126 @@
...
+				name-set="genderRadio"
+				value="both"
+				v-model="gender"
+				class="both-radio"
+			>
+				Both
+			</kv-radio>
+			<br>
+			<kv-radio
+				label-set="genderRadioSet"
```
emuvente: Since this is used for the input id, this value needs to be different for each radio button (and also unique on the entire page, so maybe something like `genderRadioSetMale`)

src/components/Kv/KvRadio.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-	<span class="styled-radio">
+	<div class="styled-radio">
 		<input
 			type="radio"
-			value="radio-input"
+			:value="value"
```
emuvente: You could also remove this line and the prop, which will result in the `value` attribute being applied automatically by the `v-bind="$attrs"` line (the same things could be done for the `name` attribute)

BoulderBrains:
Gender radios are up and working, tested locally. Updates to loan counter are occurring on click.

src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+	<div class="gender-radios">
```
emuvente: I think it might make sense to include the title in this component too `<h3>Gender</h3>`

BoulderBrains: 💯

src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,87 @@
...
+		<kv-radio
+			label-set="genderRadioSetBoth"
+			name-set="genderRadio"
+			radio-value="both"
+			v-model="gender"
+			class="both-radio"
+		>
+			Both
```
emuvente: We should make these match the labels and ordering from the copy doc:
```
Anyone
Women only
Men only
```

BoulderBrains: Just found the content doc and made these changes. I verified with Brandon about the using "Everyone" instead of "Anyone" was an alright change. I felt that "Everyone" was more inclusive sounding than "Anyone"

src/pages/Autolending/GenderRadios.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+	<div class="gender-radios">
+		<h3 class="filter-title">
+			Individuals/Groups
```
emuvente: Gender?

BoulderBrains: ffs juggling two different branches error.

BoulderBrains:
👍 Thanks

PR #1244, opened 2019-10-22
Turn off compile lint in favor of test stage linting and IDE function…

mcstover:
…ality.


mcstover:
Obvs some more clean up to be done here but wanted to start a discussion and see if there was anything else we should add.

mcstover:
Can I remove these too?
`"stylelint": "^8.4.0", "stylelint-config-standard": "^18.3.0", "stylelint-scss": "^3.11.0",`

emuvente:
@mcstover no we want to keep those for the linting script. We'll actually finally get to upgrade those now that we aren't waiting on stylelint-webpack-plugin

mcstover:
@emuvente no need for 2nd review....just blindly clicking new buttons...

BoulderBrains:
👀

PR #1245, opened 2019-10-22
Webpack dev compile improvements

eddieferrer:
**WIP
For testing purposes, do not merge yet.** 

First compile will take a long time, as it needs to build a cache. 

I added a script to manually destroy the cache:
`npm run destroyBuildCache`


build/webpack.server.conf.js
```diff
@@ -1,17 +1,13 @@
...
 var VueSSRServerPlugin = require('vue-server-renderer/server-plugin');
 var FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin');
 
-var noop = path.resolve(__dirname, 'no-op.js');
-
 module.exports = merge.smart(baseWebpackConfig, {
 	entry: './src/server-entry.js',
 	target: 'node',
-	devtool: '#source-map',
+	devtool: 'eval',
```
mcstover: If your running server debug processes on code outside of vue components it would be helpful to change this back. But for dev purposes I'm fine setting to eval.

That said, I don't think we should leave this as 'eval' for the compiled build that we deploy. Seems like we should switch back to 'source-map' for that... maybe just a prod toggle like in other places...

mcstover: Similar to webpack.client.base. where we import the config then run this...
`config.build.productionSourceMap ? '#source-map' : false,` <-making false 'eval' instead.

eddieferrer: `config.build.productionSourceMap ? '#source-map' : false`

I don't believe this code works at all. (See my new commit for changes) I tested and config.build.productionSourceMap is always true. It is a key in the config file and it is always true. For the dev server and builds the false condition of the ternary never ran. I changed this check to be isProd and based on the node env instead. This works as expected. 

Also, I could not find any reference to '#source-map' the WebPack 4 docs reference that option as 'source-map'

build/webpack.client.base.conf.js
```diff
@@ -7,11 +6,14 @@ var MiniCssExtractPlugin = require('mini-css-extract-plugin');
...
 
+const HardSourceWebpackPlugin = require('hard-source-webpack-plugin')
+const isProd = process.env.NODE_ENV === 'production';
+
 module.exports = merge.smart(baseWebpackConfig, {
 	entry: {
 		app: './src/client-entry.js'
 	},
-	devtool: config.build.productionSourceMap ? '#source-map' : false,
+	devtool: isProd ? 'source-map' : false,
```
mcstover: I'm fine with this. the '#' is webpack's default prefix for source map lines so I think it was unnecessary for our purposes.

package.json
```diff
@@ -15,6 +15,7 @@
 		"build:all": "rimraf dist && npm run build:client && npm run build:server",
 		"build:client": "rimraf dist/static && webpack --config build/webpack.client.prod.conf.js --progress --hide-modules",
 		"build:server": "webpack --config build/webpack.server.conf.js --progress --hide-modules",
+		"destroyBuildCache": "rimraf node_modules/.cache",
```
mcstover: Is there any concern from using the default cache location and consequently blowing away the entire node_modules cache which is used by other tools such as babel?

eddieferrer: Probably no concern, but I updated it to only destroy the hard-source folder.

mcstover: Thanks! @emuvente Let's chat about this briefly in our sync tomorrow then we'll merge and monitor through the next sprint.

PR #1252, opened 2019-10-24
Cash 1461

BoulderBrains:


BoulderBrains:
Still in draft, didn't seem to be hooking into the loan count display correctly. Need to look into that tomorrow morning.

src/pages/Autolending/GroupRadios.vue
```diff
@@ -0,0 +1,100 @@
...
+						filters {
+							isGroup
+						}
+					}
+				}
+			}
+		}`,
+		preFetch: true,
+		result({ data }) {
+			this.isGroup = !!_get(data, 'autolending.currentProfile.loanSearchCriteria.filters.isGroup') || 'both';
```
emuvente: `isGroup` can be true, false, or null. We need to map those values to something we can use for the radio options
```
const isGroup = _get(data, 'autolending.currentProfile.loanSearchCriteria.filters.isGroup');
if (isGroup === true) {
    this.isGroup = 'groups-only';
} else if (isGroup === false) {
    this.isGroup = 'individual-only';
} else {
    this.isGroup = 'both';
}
```
And then we need to do the same thing but in reverse in the watch method below to map the radio values into true/false/null

src/pages/Autolending/GroupRadios.vue
```diff
@@ -0,0 +1,115 @@
...
+									filters: {
+										isGroup: 
+											if (isGroup) === true) {
+												isGroup = "group-only";
+											} else if (isGroup === false) {
+												isGroup = "individual-only";
+											} else {
+												isGroup = "both";
+											}
+
```
BoulderBrains: I don't think this mapping is formatted correctly here.

emuvente: Do this before the apollo.mutate call, not inside of the template string

emuvente: You want to map from the group option string values to true/false/null, so you could have something like
```
let isGroup = null;
if (groupOption === 'group-only') {
    isGroup = true;
} else if (groupOption === 'individual-only') {
    isGroup = false;
}
```
and then you just need `isGroup: ${isGroup}` in the query

src/pages/Autolending/GroupRadios.vue
```diff
@@ -0,0 +1,113 @@
...
+						}
+					}
+				}
+			}
+		}`,
+		preFetch: true,
+		result({ data }) {
+			const isGroup = _get(data, 'autolending.currentProfile.loanSearchCriteria.filters.isGroup');
+			if (isGroup === true) {
+  				this.isGroup = 'group-only';
```
emuvente: It looks like travis is only failing because of the mixed tabs and spaces on this line

PR #1258, opened 2019-10-26
Require login to view autolending settings page and show tip messages…

emuvente:
… after saving


mcstover:
This will be good. We'll have some time to observe any common errors and figure out if there are other ways we want to respond to certain codes.

PR #1264, opened 2019-10-31
CASH-1482 Autolending opt-out

eddieferrer:
* Added auto lending opt out page/e2e tests
* Alphabetized routes in routes file for better organization
* Renamed e2e test for autolending settings page to match component name


src/pages/Autolending/AutolendingOptOutPage.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+	<www-page :gray-background="true">
+		<div class="title-area">
+			<div class="row column">
+				<h1>Auto-lending</h1>
```
eddieferrer: This title area is duplicate code. If this gets approved to stay on this opt-out page then we can move this page and the autolending settings page to share the title area with a child/parent route

src/router/routes.js
```diff
@@ -1,37 +1,43 @@
...
+	{ path: '/page-two', component: () => import('@/pages/PageTwo') },
+	{ path: '/portfolio/lending-stats', component: () => import('@/pages/LendingStats/LendingStatsPage') },
+	{ path: '/process-browser-auth', component: () => import('@/pages/ProcessBrowserAuth') },
+	{ path: '/protocol', component: () => import('@/pages/Protocol') },
 	{ path: '/register/social', component: () => import('@/pages/LoginAndRegister/RegisterSocial') },
-	{ path: '/build/research', component: () => import('@/pages/Build/Research') },
-	{ path: '/error', component: () => import('@/pages/Error') },
+	{ path: '/settings/autolending', component: () => import('@/pages/Autolending/AutolendingPage')	},
+	{
+		path: '/settings/autolending/opt-out',
```
eddieferrer: This is the new route. The rest of the changes here are just alphabetizing the routes to make them a little more organized

src/pages/Autolending/AutolendingOptOutPage.vue
mcstover: This minor amount of duplication is fine for now. After initial launch of this page we can consider some enhancements to using a parent view instead.

PR #1265, opened 2019-11-01
Cash 1463

BoulderBrains:


src/pages/Autolending/LoanTermDropdown.vue
```diff
@@ -0,0 +1,104 @@
...
+					}
+				}
+			}
+		}`,
+		preFetch: true,
+		result({ data }) {
+			const loanTermMax = _get(data, 'autolending.currentProfile.loanSearchCriteria.filters.lenderTerm.max');
+			if (loanTermMax <= 6) {
+				this.loanTerm = '6m-or-less';
+			} else if (loanTermMax > 6 && loanTermMax <= 12) {
```
eddieferrer: For my own understanding, why do we have to do these range checks? We can only set loanTermMax to 6/12/18/24 can lenderTerm.max be any other values?

mcstover: The api and loan search criteria allows any number to be set for the min or max. The 6/12/18/24 options represent our very specific business rules for this interface only.

src/pages/Autolending/LoanTermDropdown.vue
```diff
@@ -0,0 +1,104 @@
...
+					loanTerm = 12;
+				} else if (loanTermMax === '18m-or-less') {
+					loanTerm = 18;
+				} else if (loanTermMax === '24m-or-less') {
+					loanTerm = 24;
+				}
+				this.apollo.mutate({
+					mutation: gql`mutation {
+						autolending @client {
+							editProfile(profile: { lenderTerm: ${loanTerm} })
```
emuvente: You can base the structure of these mutations on the query you use to get the loan term, so in this case you want
```
editProfile(profile: {
	loanSearchCriteria {
		filters {
			lenderTerm {
				max: ${loanTerm}
			}
		}
	}
})
```

PR #1268, opened 2019-11-02
cash-1464 partnerDeliquency dropdown configured without data

BoulderBrains:


src/pages/Autolending/PartnerDelRateDropdown.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+	<div>
+		<h3 class="filter-title">
+			Partner delinquency rate
+		</h3>
+		<kv-dropdown-rounded v-model="delinquencyRate">
+			<option value="default">
```
emuvente: If you use numbers instead of strings for the option values (e.g. `0` for default, `5` for 5% or less, etc.) Then you won't need any of those if/else statements that are mapping the number values to string values.

BoulderBrains: Alright, I'll go through and make those changes in this branch and the other branches I've created using strings.

src/pages/Autolending/PartnerDelRateDropdown.vue
```diff
@@ -0,0 +1,117 @@
...
+					delinquencyRate = 15;
+				} else if (delinquencyRateMax === '20-or-less') {
+					delinquencyRate = 20;
+				} else if (delinquencyRateMax === '25-or-less') {
+					delinquencyRate = 25;
+				}
+				this.apollo.mutate({
+					mutation: gql`mutation {
+						autolending @client {
+							editProfile(profile: { delinquencyRate: ${delinquencyRate} })
```
emuvente: Similarly to my comment on #1265, I think the mutation you want to use here is
```
editProfile(profile: {
	loanSearchCriteria {
		filters {
			arrearsRate {
				max: ${delinquencyRateMax}
			}
		}
	}
})
```

BoulderBrains:
@emuvente ready for another review.

PR #1269, opened 2019-11-04
Cash 1465

BoulderBrains:


eddieferrer:
Seems like the Risk Rating dropdown (and also the Loan Term Dropdown) are not firing the isChanged flag on change. So, if I go into my settings, change Risk Rating, I do not see the Save button. Is this something that should be working at this time?

BoulderBrains:
Thanks for bringing that up @eddieferrer, taking a look into that now.

src/pages/Autolending/RiskRatingDropdown.vue
```diff
@@ -0,0 +1,95 @@
...
+import KvDropdownRounded from '@/components/Kv/KvDropdownRounded';
+
+export default {
+	inject: ['apollo'],
+	components: {
+		KvDropdownRounded,
+	},
+	data() {
+		return {
+			riskRating: '1',
```
emuvente: Use a number `1` instead of a string `'1'`, because the value fetched in the query will be a number

src/pages/Autolending/RiskRatingDropdown.vue
```diff
@@ -0,0 +1,95 @@
...
+							}
+						}
+					}
+				}
+			}
+		}`,
+		preFetch: true,
+		result({ data }) {
+			// eslint-disable-next-line max-len
+			const riskRatingMin = _get(data, 'autolending.currentProfile.loanSearchCriteria.filters.riskRating.min' || 1);
```
emuvente: The `|| 1` should go outside of the _get call:
```
_get(data, 'autolending.currentProfile.loanSearchCriteria.filters.riskRating.min') || 1;
```

src/pages/Autolending/RiskRatingDropdown.vue
```diff
@@ -0,0 +1,95 @@
...
+				riskRating = riskRatingMin;
+				this.apollo.mutate({
+					mutation: gql`mutation {
+						autolending @client {
+							editProfile(profile: {
+								loanSearchCriteria: {
+									filters: {
+										riskRating: {
+											min: 0
+											max: ${riskRating || null}
```
emuvente: For this filter, change the min not the max. The max should be hard-coded as 5

PR #1270, opened 2019-11-05
CASH-1198 Auto lend timing settings dropdown

eddieferrer:
* Adds auto lend timing dropdown to new autolend settings page

CASH-1198

### DIFF
--- a/src/pages/Autolending/AutolendingPage.vue
+++ b/src/pages/Autolending/AutolendingPage.vue
@@ -10,6 +10,7 @@
 			<main-toggle />
 		</div>
 		<div class="row column settings-area" :class="{ obscure: !isEnabled }">
+			<lend-timing-dropdown />
 			<!-- main toggles -->
 			<div class="setting-group">
 				<!-- timing dropdown -->
@@ -91,6 +92,7 @@ import autolendingQuery from '@/graphql/query/autolending/autolendingPage.graphq
 import AttributeFilter from './AttributeFilter';
 import CountryFilter from './CountryFilter';
 import DonationDropdown from './DonationDropdown';
+import LendTimingDropdown from './LendTimingDropdown';
 import FloatingCounter from './FloatingCounter';
 import MainToggle from './MainToggle';
 import GenderRadios from './GenderRadios';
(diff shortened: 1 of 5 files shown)


src/pages/Autolending/AutolendingPage.vue
```diff
@@ -10,6 +10,7 @@
 			<main-toggle />
 		</div>
 		<div class="row column settings-area" :class="{ obscure: !isEnabled }">
+			<lend-timing-dropdown v-if="isEnabled" />
```
emuvente: The v-if should not be needed as this dropdown will always be visible.

test/unit/specs/pages/Autolending/LendTimingDropdown.spec.js
```diff
@@ -0,0 +1,177 @@
...
+		it('Previous auto lend 18 - If I dont log in for 1.5 Years', () => {
+			expect(wrapper.vm.convertEnableAfterToNewSetting(18)).toBe(120);
+		});
+		it('Previous auto lend null or other value', () => {
+			expect(wrapper.vm.convertEnableAfterToNewSetting(1666)).toBe(0);
+			expect(wrapper.vm.convertEnableAfterToNewSetting(undefined)).toBe(0);
+			expect(wrapper.vm.convertEnableAfterToNewSetting(null)).toBe(0);
+		});
+	});
+	describe('Apollo mixin result callback should set the component variables correctly', () => {
```
emuvente: I think these might be better as cypress tests, where the api response to `{ my { autolendProfile } }` can be mocked with this data and then check that the correct option is currently selected in the dropdown.

src/pages/Autolending/AutolendingPage.vue
eddieferrer: I think it looks better if the drop down is hidden because the copy reads: 

_Auto-lending On
Kiva will automatically lend my balance
<dropdown>_

and 

_Auto-lending off
I’ll lend my balance myself_

If the dropdown was not hidden it would be: 

_Auto-lending off
I’ll lend my balance myself
<dropdown>_

emuvente: The design mocks have it still visible, but transparent and non-interactive. See the timing-donation-OFF.png attachment in https://kiva.atlassian.net/browse/CASH-1249

PR #1272, opened 2019-11-06
Modified Jest config to include testing/coverage reporting of Vue com…

eddieferrer:
…ponents and removed duplicated exclusion patterns

We should be including our Vue components in code coverage for unit testing.


emuvente:
The Vue components were removed from coverage reporting for Jest because we wanted Vue components to primarily be tested with Cypress, not Jest. We don't have coverage reporting added to Cypress yet, as that used to require manual instrumenting, but there are coverage plugins for Cypress now which should simplify adding coverage reporting to those tests.

eddieferrer:
I'm fine with this approach of focusing on cypress tests for testing the components.

PR #1274, opened 2019-11-06
Try using alternative domain for TOS consent page for mobile (dev-only)

emuvente:


src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -80,8 +80,14 @@ export default {
 				// show error here
 				this.showNewAcctTermsError = true;
 			} else {
-				window.location = `https://${this.auth0Config.domain}`
-				+ `/continue?agree=yes&state=${this.$route.query.state}`;
+				let auth0Domain = this.auth0Config.domain;
+
+				// Use alternatvie domain for mobile app
+				if (this.$route.query.clientID === 'p6WnYfHbJZiBtnIGAx55Yb2PSeL5MJgB' && this.auth0Config.altDomain) {
```
meganmcmillan: We will eventually need to compare against different clientIDs here, but this is good for testing if this will work!

meganmcmillan: Ohh, now I see the (dev-only) comment.

PR #1276, opened 2019-11-07
cash-1463 Strings to intergers and reworking graphql mutation, that d…

BoulderBrains:
…oesn't work

Made the string to integer changes here. 
Also made the mutation graphql update that you suggested, but the syntax of the mutation is erroring out when being called. 
error: 
GraphQLError {message: "Syntax Error: Expected :, found {", locations: Array(1), path: undefined, nodes: undefined, source: Source, …}

### DIFF
--- a/src/pages/Autolending/LoanTermDropdown.vue
+++ b/src/pages/Autolending/LoanTermDropdown.vue
@@ -7,16 +7,16 @@
 			<option value="default">
 				All loan terms
 			</option>
-			<option value="6m-or-less">
+			<option value="6">
 				6 months or less
 			</option>
-			<option value="12m-or-less">
+			<option value="12">
 				12 months or less
 			</option>
-			<option value="18m-or-less">
+			<option value="18">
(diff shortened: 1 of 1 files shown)


src/pages/Autolending/LoanTermDropdown.vue
```diff
@@ -4,19 +4,19 @@
 			Loan term
 		</h3>
 		<kv-dropdown-rounded v-model="loanTerm">
-			<option value="default">
+			<option value="70">
```
BoulderBrains: @emuvente 
I set this 'All loans' option to 70 because that's a little over the max value we currently have as a lenderLimit. I can't imagine we'll be doing many loans with a larger lenderLimit than 70. What do you think?

BoulderBrains:
I cleaned this up a lot, can you take another pass at this for me @emuvente

BoulderBrains:
@emuvente goddamn that was fast, thank you. 🙏

PR #1280, opened 2019-11-07
VUE-166 Stop vue-progressbar plugin from jumping to 100% when it's at…

ryan-ludwig:
… 95% or above

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -24,6 +24,7 @@ Vue.use(VueProgressBar, {
 		opacity: '0s',
 		termination: 300,
 	},
+	autoFinish: false,
 });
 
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -24,6 +24,7 @@ Vue.use(VueProgressBar, {
 		opacity: '0s',
 		termination: 300,
```
ryan-ludwig: @mcstover termination seems to be how many ms the progress bar stays on the screen after reaching 100%

go-bears:
first commit to kiva repo? 🎉
👋 @ryan-ludwig  welcome to kiva!

PR #1281, opened 2019-11-08
Cash 1513

BoulderBrains:


src/pages/Autolending/DefaultRateDropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+import KvDropdownRounded from '@/components/Kv/KvDropdownRounded';
+
+export default {
+	inject: ['apollo'],
+	components: {
+		KvDropdownRounded,
+	},
+	data() {
+		return {
+			defaultRate: '0',
```
emuvente: Use a number `0` instead of a string `'0'` because the value from the query will be a number

src/pages/Autolending/DefaultRateDropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+							}
+						}
+					}
+				}
+			}
+		}`,
+		preFetch: true,
+		result({ data }) {
+			// eslint-disable-next-line max-len
+			const defaultRateMax = _get(data, 'autolending.currentProfile.loanSearchCriteria.filters.defaultRate.max');
```
emuvente: You can add `|| 0` to the end of this line to make 0 the default value used when nothing is defined

src/pages/Autolending/DefaultRateDropdown.vue
```diff
@@ -0,0 +1,110 @@
...
+				defaultRate = defaultRateMax;
+				this.apollo.mutate({
+					mutation: gql`mutation {
+						autolending @client {
+							editProfile(profile: {
+								loanSearchCriteria: {
+									filters: {
+										defaultRate: {
+											min: 0
+											max: ${defaultRate}
```
emuvente: Use `${defaultRate || null}` so that the max will be set to null if "all" is selected

PR #1282, opened 2019-11-08
VUE-169 Page out of range in lend by category

eddieferrer:
* If the page query param is out of range in the lend by category page, instead of throwing a 500 error, navigate to last page and display a tip message

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -265,7 +265,7 @@ export default {
 				}
 			});
 		} catch (e) {
-			logReadQueryError(e, 'LoanChangeCategoryPage loanChannelPageQuery');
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -262,7 +265,7 @@ export default {
 				}
 			});
 		} catch (e) {
-			logReadQueryError(e, 'LoanChanngeCategoryPage loanChannelPageQuery');
+			logReadQueryError(e, 'LoanChangeCategoryPage loanChannelPageQuery');
```
mcstover: Should be 'LoanChannelCategoryPage'

eddieferrer: 👍

PR #1287, opened 2019-11-11
do not render these routes in prod just yet

mcstover:


eddieferrer:
You're going to have to comment out the e2e tests to visit those URL's. 

test/e2e/specs/AutolendingOptOutPage.spec.js
test/e2e/specs/AutolendingPage.spec.js

An easy way to not run these is to change the top level 'describe' to 'xdescribe'

PR #1291, opened 2019-11-11
cash-1514 fixing typo

BoulderBrains:


BoulderBrains:
@mcstover This is correctly spelled in the legacy stack.

PR #1293, opened 2019-11-13
Create Possibility page frame. Redirect to certain campaign pages depending on date

ryan-ludwig:
CASH-1540

### DIFF
--- a/src/pages/Possibility/Possibility.vue
+++ b/src/pages/Possibility/Possibility.vue
@@ -21,11 +21,11 @@ export default {
 		const jan1 = 157783680000;
 
 		let url = '/';
-		if (now >= dec3 && now <= dec14) {
+		if (now >= dec3 && now < dec14) {
 			url = '/possibility/giving-tuesday';
-		} else if (now >= dec14 && now <= dec26) {
+		} else if (now >= dec14 && now < dec26) {
 			url = '/possibility/12-days-of-lending';
-		} else if (now >= dec26 && now <= jan1) {
+		} else if (now >= dec26 && now < jan1) {
 			url = '/possibility/year-end';
 		}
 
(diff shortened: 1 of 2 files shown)


src/pages/Possibility/Possibility.vue
```diff
@@ -0,0 +1,33 @@
...
+import WwwPage from '@/components/WwwFrame/WwwPage';
+
+export default {
+	components: {
+		WwwPage
+	},
+	beforeCreate() {
+		const todaysDate = new Date();
+		const day = todaysDate.getDate();
+		const month = todaysDate.getMonth() + 1; // getMonth is 0 based
```
mcstover: @ryan-ludwig please look into making this switch at midnight PDT rather than basing it on system time.

src/pages/Possibility/Possibility.vue
```diff
@@ -0,0 +1,37 @@
...
+		const offset = 8 * 60 * 60 * 1000; // PDT is 8 hours behind UTC
+		const now = Date.now() - offset;
+
+		const dec3 = 1575331200000;
+		const dec14 = 1576281600000;
+		const dec26 = 1577318400000;
+		const jan1 = 157783680000;
+
+		let url = '/';
+		if (now >= dec3 && now <= dec14) {
```
eddieferrer: I believe your conditionals are not correct here. dec14, should display 12 days of lending, not giving tuesday.

eddieferrer:
In the end, we're still arriving at the client time using Date.now, so if I change the time on my machine, I can trigger different pages to come up. 

If we want to truly guarantee that these dates can't be messed with, we should get the _server time_ in the component. While we could do a check for the window object in the beforeCreate hook, we can't use $router.push there since the server doesnt know about $router. 

The way to do this would be to use, the serverPrefetch to get server date, then pass that to the component on the client side, then do the checks and push the route based on that value.

`// Server-side only
  serverPrefetch () {
   get server date
  },`

src/pages/Possibility/Possibility.vue
ryan-ludwig: You're right, thanks

PR #1296, opened 2019-11-14
Giving Tuesday draft and KvHero component

ryan-ludwig:
CASH-1521


src/components/Kv/KvHero.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+	<div class="hero">
+		<div class="images">
+			<div class="images-placeholder"></div>
+			<div class="images-container">
+				<slot name="images"></slot>
+			</div>
+		</div>
+		<div class="headline">
```
mcstover: For the 3rd phase I we'll prob want another computed check against "all" of the headline slots here so we turn on and off the block.

src/components/Kv/KvHero.vue
```diff
@@ -0,0 +1,182 @@
...
+
+.images-container {
+	position: absolute;
+	top: 0;
+}
+
+.images-placeholder {
+	background-color: $kiva-bg-lightgray;
+	width: 100%;
+	padding-bottom: 600/480 * 100%;
```
BoulderBrains: @ryan-ludwig Just curious, what's the value/purpose of setting this padding-bottom to a calculation instead of the hardcoded value of the calculation?

src/components/Kv/KvHero.vue
```diff
@@ -0,0 +1,182 @@
...
+	opacity: 0.9;
+}
+
+.headline-title {
+	position: relative;
+	display: block;
+	margin-bottom: 0;
+	font-size: 1.75rem;
+	line-height: 2rem;
+	font-weight: 400;
```
BoulderBrains: Minor, we could use this defined value here: $global-weight-highlight;

src/components/Kv/KvHero.vue
ryan-ludwig: I grabbed these styles from the homepage slider but I like how it's set up. It makes it a little more clear that this padding is determined by the width/height of the image inside (600px x 480px here), rather than seeing something arbitrary like 125%.

BoulderBrains: @ryan-ludwig I just spoke to matt and he informed me that you were basing this off of the HeroSlideshow.vue file, where I also see these styles. 👍 Would be good to have a conversation with you around the definitions we have defined in _settings.scss and how to fold them into these reusable components. Having that discussion will also bring up some larger questions around, how do we fix our "styleguide" to be more useful and reuseable.

ryan-ludwig: Totally! I'd love to chat about common variables and future styleguide plans.

PR #1297, opened 2019-11-14
Get content from Contentful

eddieferrer:
* Added local resolvers/schema to fetch content from contentful
* Added GlobalPromotionalBanner component to fetch banner settings and banner content from contentful
* Added contentful dependencies and alias to allow them to work
* Modified settingsUtils to handle dates from contentful
* Added contentful settings to app config

VUE-167

### DIFF
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -28,6 +28,7 @@ module.exports = {
 	resolve: {
 		extensions: ['.mjs', '.js', '.vue', '.json'],
 		alias: {
+			'~': resolve('node_modules'),
 			'@': resolve('src'),
 			'foundation': 'foundation-sites/js',
 		}
--- a/config/dev-vm.js
+++ b/config/dev-vm.js
@@ -49,6 +49,9 @@ module.exports = merge(base, {
 			serverCallbackUri: 'https://dev-vm-01.kiva.org/process-ssr-auth',
 			domain: 'login.dev.kiva.org',
 		},
+		contentful: {
+			environment: 'development'
+		},
 	},
 	server: {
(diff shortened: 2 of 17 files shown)


eddieferrer:
What we have so far:

* New global banner component that gets uiSetting > ui-global-promo field from contentful, then enables the promo based on the active field and date ranges of that setting. 
* This setting contains some validation and a reference to a promo content type in contentful
* The promo content type includes kvtracking, rich text and link destination
* The content is rendered in the new component. With support for the RichText coming from contentful

Questions:
* Do we want to remove the old GlobalPromotionalBanner component? 
* Are we going to make the switch for global banners to use contentful?
* Do we need to support text color, hover text color and background color for the banners? If we want to we can add more fields to contentful, there are even third party extensions for adding color pickers for color fields

eddieferrer:
This is done. Now supports an array of content from contentful, allowing us to fully automate the global banner stuff. 

This PR does not encompass the full scope of VUE-167, since it only deals with the global promo banner. 

This is really the work for CASH-1571, CASH-1573 for the new stack. It should probably not go live until the work for this is done on the old stack

mcstover:
thanks @eddieferrer Let's hold on any merges of this until early next week.

build/webpack.base.conf.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
 	resolve: {
 		extensions: ['.mjs', '.js', '.vue', '.json'],
 		alias: {
+			'~': resolve('node_modules'),
```
mcstover: @eddieferrer In your testing did this addition affect any of the image imports? Homepage components would be a quick place to check.

eddieferrer: No it does not affect them, everything still works.

PR #1304, opened 2019-11-20
Initial build of advent calendar

ryan-ludwig:
CASH-1527


src/pages/Possibility/TwelveDaysCalendar.vue
```diff
@@ -0,0 +1,469 @@
...
+import Number10 from '@/assets/inline-svgs/12-days-of-lending/10.svg';
+import Number11 from '@/assets/inline-svgs/12-days-of-lending/11.svg';
+import Number12 from '@/assets/inline-svgs/12-days-of-lending/12.svg';
+import SquigglyLine from '@/assets/inline-svgs/12-days-of-lending/squiggly_line.svg';
+import StarBig from '@/assets/inline-svgs/12-days-of-lending/star_big.svg';
+import StarLittle from '@/assets/inline-svgs/12-days-of-lending/star_little.svg';
+
+export default {
+	components: {
+		Number1,
```
ryan-ludwig: Currently using a [special webpack loader](https://github.com/visualfanatic/vue-svg-loader) to inline the SVGs instead of <img src=".svg" them. They need to be inlined so the fill colors can be set with CSS. Not sure this is the best way to get it done though. Would love some feedback.

eddieferrer: This approach is fine by me

mcstover: Yea we probably could have made it work with a similar approach to the icons but I don't want all of this loaded everywhere.

I also really like the added capability of including and using an svg ONLY where we need it.

ryan-ludwig:
TODO: Determine how we want to drive the dates

mcstover:
Let's hold this to merge sometime tomorrow. Lot's of in-progress work going into master right now that will be easier to transfer to our release-11252019 by direct merge instead of cherry-pick.

src/pages/Possibility/TwelveDaysCalendar.vue
emuvente: I like this approach for svgs that we want to be able to change/animate in the page

ryan-ludwig:
@mcstover Ready to go in?

mcstover:
Not just yet please. A few things happening in master still that I'd like to transfer to release

PR #1306, opened 2019-11-20
Cash 1523 kiva content block

BoulderBrains:
We want to get the majority of this component into master before branch cut. Still to come, small style tweaks across the screen sizes.


src/pages/Possibility/KivaContentBlock.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+	<kv-carousel @change="slideChange">
+		<div class="slide-placeholder"></div>
+		<kv-carousel-slide :key="headline" v-for="{ images, url, headline, subheadline } in kivaContentBlockData">
```
ryan-ludwig: This will result in duplicate :keys. Could pass an index into the v-for and do something like :key="`headline-${index}`"

emuvente: Each of the headlines is unique, so there won't be a problem with duplicate keys for the current set of slides, but adding the index would be a way to future-proof this component in case slides are added that have the same headline.

BoulderBrains: @ryan-ludwig Each of the headlines are unique. Just trying to understand why this will result in duplicate keys? Can you expand on this a little bit for me?

ryan-ludwig: Ah you're right, I was reading it as a string instead of a binding.

src/pages/Possibility/KivaContentBlock.vue
```diff
@@ -0,0 +1,182 @@
...
+					url: '/blog/ernestina-is-a-skilled-carver-making-a-name-for-herself-in-a-male-dominated-craft',
+					headline: 'Turning possibility into craft',
+					// eslint-disable-next-line max-len
+					subheadline: 'Carving a place for women in a male-dominated craft, Ernestina is more than a trailblazer'
+				}
+			]
+		};
+	},
+	methods: {
+		slideChange() {
```
emuvente: I think you can drop this method and the `@change="slideChange"` in the template, snice counter isn't being used anywhere

src/pages/Possibility/KivaContentBlock.vue
BoulderBrains: Made these changes, thanks for the discussion guys.

PR #1307, opened 2019-11-20
Default to /possibility/giving-tuesday 

ryan-ludwig:
if you land on /possibility and it's before dec3


eddieferrer:
What happens if you end up in /possibility after Jan 1? Now its going to redirect to giving-tuesday, which makes less sense...

ryan-ludwig:
> What happens if you end up in /possibility after Jan 1? Now its going to redirect to giving-tuesday, which makes less sense...

Addressed in #1309

PR #1308, opened 2019-11-20
Vue 176 auth0 superagent warning

eddieferrer:
VUE-176

Fixes a server compile warning about superagent in a non-browser environment


eddieferrer:
> Nice! Auth0Link should be updated also to use something besides the isServer value
> 
> https://github.com/kiva/ui/blob/4be2243eef0af9afbff5de842c659f8237c76dce/src/api/Auth0Link.js#L37

Good catch

PR #1312, opened 2019-11-21
cash-1523 css tweaks for the /possibility page

BoulderBrains:


src/pages/Possibility/GivingTuesday.vue
```diff
@@ -109,6 +108,8 @@ $cta-color: #02582e;
 
 .cta-button {
 	@include button-style($cta-color, auto, #fff);
+	// Include in order to override the default box-shadow of this button
+	box-shadow: 0 2px darken($cta-color, 10%);
 }
```
BoulderBrains: This was just a minor issue I saw while adding my component to the GivingTuesday page.

emuvente:
Since the KivaContentBlock component is only for the /possibility pages, let's keep it in the /pages/ directory. That way, the css for that component will only be loaded for those pages. When components are put in the /components/ directory, their css is added to the app bundle, making them load globally for every page.

src/pages/Possibility/KivaContentBlock.vue
```diff
@@ -142,39 +144,85 @@ $cta-color: #611b1f;
...
-}
 
-.slide-placeholder {
-	// background-color: red;
-	width: 100%;
-	padding-bottom: 600/480 * 100%;
+	// This color customization isn't working for the second slide, because the 
+	// kv-carousel uses the css display: none; property which removes the element from
+	// the DOM which then messes up the ability to apply css using :nth-child(#)
+
```
BoulderBrains: I'm open to suggestions here if anyone has them.

BoulderBrains: Alternatively we could just keep all the text and buttons the same color and remove all this extra css. Brandon left a comment in the ticket that reads, "Brandon Smith to update colors to be consistent" So maybe the color changes between slides was something that was supposed to be removed from the ticket?

ryan-ludwig: @BoulderBrains One option would be to add a unique class to each kv-carousel-slide as you loop through your data. Then you could use that class instead of nth-childing to customize the colors.

emuvente: ☝️

BoulderBrains: 👍

PR #1315, opened 2019-11-21
Added vue filter for taking loan params and returning loan use string

eddieferrer:
VUE-78


src/api/localResolvers/contentfulCMS.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-underscore-dangle */
```
eddieferrer: This is unrelated to this issue, just happened across this

PR #1317, opened 2019-11-22
CASH-1547 possibilities homepage promo

ryan-ludwig:
- Replace former promos and experiments with possibilities promo. Use new KvHero component.
- Use ContentfulCMS to determine when to show promo on the homepage

TODO: Some styling work for the possibility promo. Any other PR feedback.


src/pages/Homepage/Homepage.vue
```diff
@@ -29,69 +26,30 @@ export default {
...
-			this.$kvTrackEvent(
-				'Homepage',
-				'EXP-CASH-1313-Oct2019',
-				this.doubleArrowButtonExp.version === 'shown' ? 'b' : 'a'
+				uiPromoSetting,
+				'active',
+				'startDate',
+				'endDate'
 			);
+			// this.promoEnabled = uiPromoSetting.active; // use this to bypass date during dev
```
ryan-ludwig: To test the Contentful integration prior to Giving Tuesday, uncomment this line and toggle the ui-homepage-promo active setting within Contentful development space.

eddieferrer: It seems to me that this is prone to error, forgetting to uncomment this line, etc. We should uncomment this line. If you want to enable or disable it, you can change the setting in the contentful development space, that is the reason why development exists. It will also allow us to test this on dev.kiva.org

src/pages/Homepage/HeroSlideshow.vue
```diff
@@ -1,182 +1,128 @@
...
+			<template v-slot:headlineTitle>
+				Make opportunity possible <br class="smo">for borrowers around the world.
+			</template>
+			<template v-slot:action>
+				<div class="row">
+					<div class="small-12 large-6 columns action-1">
+						<kv-button
+							class="action-button"
+							to="/lend-by-category/"
+							v-kv-track-event="['Home', 'possibility', 'click-Lend-now', null, 'true']"
```
eddieferrer: These are the legacy params. You can see in src/plugins/kv-analytics-plugin.js that v-kv-track-event only takes 4 params.  this should be v-kv-track-event="['Home', 'possibility', 'click-Lend-now'] with the value param null. 'true' shouldn't be a param. Also applies to other v-kv-track events on this page

src/pages/Homepage/Homepage.vue
ryan-ludwig: Good points, thanks.

ryan-ludwig: I've removed this line, to test you can change the date range and active setting in Contentful development space.

src/pages/Homepage/HeroSlideshow.vue
```diff
@@ -1,182 +1,138 @@
...
-		</div>
+	<div>
+		<!-- Promo -->
+		<kv-hero v-if="promoEnabled" headline-bg-color="#00244e" class="possibilities-promo">
+			<template v-slot:images>
+				<kv-carousel @change="slideChange">
+					<kv-carousel-slide>
+						<kv-responsive-image
+							:images="possibilitiesImages(1)"
+							alt="An African American woman in a park wearing a leather jacket, smiling broadly"
```
ryan-ludwig: @mcstover Is there someone I can run wording like this past?

mcstover: Your best copy will work fine for these right now. There is no process (or time in this case) to go back through and do these.

In the future as we do these image heavy hero spaces we should start to ask for specific alt text against each image otherwise we'll just make it up according to what we see (this is what I usually do ;).

src/pages/Homepage/HeroSlideshow.vue
```diff
@@ -111,7 +111,7 @@ export default {
 	serverCacheKey: props => {
 		let cacheKey = 'DefaultHeroSlideshow';
 		if (props.promoEnabled) {
-			cacheKey = 'PromoSlideshow';
+			cacheKey = 'PromoSlideshowPossibility';
```
eddieferrer: I have my serious doubts that this is even doing anything. If you look at getCacheKey, its doing a substring(0,8) so any getCacheKey parameter shorter than 8 chars will be trimmed anyway.... so:
cacheKey = 'PromoSlideshow'; and cacheKey = 'PromoSlideshowPossibility';  will actually return the same cacheKey of 'PromoSli'

eddieferrer: Reference: https://ssr.vuejs.org/guide/caching.html#page-level-caching we should really take a second look at this cache key stuff...

emuvente: No that's incorrect. The substring(0,8) only applies to the commit hash, which is then appended to the full key string.

src/components/Kv/KvCarousel.vue
```diff
@@ -54,5 +54,6 @@ export default {
 	position: relative;
 	overflow: hidden;
 	width: 100%;
+	height: 100%;
```
emuvente: Is this going to have any unintended side effects? I didn't include this originally so that the content of the slides could drive the height of the carousel as opposed to needing to set the height in a wrapper.

src/pages/Homepage/HeroSlideshow.vue
ryan-ludwig: @eddieferrer I believe the substring will only be happen the ${UI_COMMIT} portion of that string.

eddieferrer: Whoops, misread that. You guys are right

src/components/Kv/KvCarousel.vue
ryan-ludwig: Seems OK so far. Currently KvCarousel is only used in KvHero and KivaContentBlock and those two look OK. Height is always kindof confusing to me in CSS but I think the content of the slides will still be driving that 100%.

BoulderBrains: ☝️ @ryan-ludwig 
FYI: I had to add a slew of media queries in KivaContentBlock.vue in order to get the height right across all the screen sizes we support. That fix was definitely less than ideal.

emuvente: Ok, this is fine for now then

emuvente:
👍

PR #1340, opened 2019-12-03
Cash 1627 contentful local caching

eddieferrer:
Implements local caching in localStorage and reduction of calls to contentful API. Client should not make same contentful request more than once every 30 minutes. 

CASH-1627

### DIFF
--- a/src/api/localResolvers/contentfulCMS.js
+++ b/src/api/localResolvers/contentfulCMS.js
@@ -46,7 +46,7 @@ export default context => {
 						contentfulQueryParams['fields.key'] = contentKey;
 					}
 
-					// Use the contentful URL as the cache key to sessionStorage
+					// Use the contentful URL as the cache key to localStorage
 					const cacheKey = 'https://cdn.contentful.com'
 									+ `/spaces/${context.appConfig.contentful.space}`
 									+ `/environments/${context.appConfig.contentful.environment}`
--- a/src/api/localResolvers/contentfulCMS.js
+++ b/src/api/localResolvers/contentfulCMS.js
@@ -1,5 +1,6 @@
 import { GraphQLJSONObject } from 'graphql-type-json';
 import { createClient } from 'contentful';
+import store2 from 'store2';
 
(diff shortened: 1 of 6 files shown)


src/pages/Homepage/Homepage.vue
```diff
@@ -30,16 +30,14 @@ export default {
 		};
 	},
 	inject: ['apollo'],
-	apollo: {
-		query: promoQuery,
-		variables() {
-			return {
+	mounted() {
+		this.apollo.query({
```
ryan-ludwig: By not prefetching, will the hero-slideshow component start showing the default content first and then show the promo content once the contentful response comes in?

src/api/localResolvers/contentfulCMS.js
```diff
@@ -37,15 +37,39 @@ export default context => {
...
 						contentfulQueryParams['fields.key'] = contentKey;
 					}
 
+					// Use the contentful URL as the cache key to localStorage
+					const cacheKey = 'https://cdn.contentful.com'
+									+ `/spaces/${context.appConfig.contentful.space}`
+									+ `/environments/${context.appConfig.contentful.environment}`
+									+ `/entries?${Object.entries(contentfulQueryParams).map(entry => entry.join('=')).join('&')}`;// eslint-disable-line max-len
+
+					const cachedResponse = localStorage.getItem(cacheKey);
```
emuvente: We have [store2](https://github.com/nbubna/store) as a dependency that you can use to help avoid browser-support issues that can crop up with using localStorage directly

src/pages/Homepage/Homepage.vue
eddieferrer: See update, but I'm open to suggestions for other ways to do it

ryan-ludwig: I think this works really well as a stopgap. Having Contentful cached on a backend server we control would probably be an ideal scenario, but this will work well to stop the bleeding.

PR #1343, opened 2019-12-04
Avoid cache if we're development channel

ryan-ludwig:


ryan-ludwig:
@eddieferrer Bumped into this testing the advent calendar and not seeing my changes, thought it'd be helpful for content people too to not have to clear their localstorage while testing.

mcstover:
I think we should probably make an enabledCache setting instead of just "development"...

Just keying off of 'development' means that all dev vms, dev, stage and qa will run without caching.

it would just be true unless the individual developer set it to false and restarted their vm.

eddieferrer:
> @eddieferrer Bumped into this testing the advent calendar and not seeing my changes, thought it'd be helpful for content people too to not have to clear their localstorage while testing.

I would rather be able to test that the local 

> I think we should probably make an enabledCache setting instead of just "development"...
> 
> Just keying off of 'development' means that all dev vms, dev, stage and qa will run without caching.
> 
> it would just be true unless the individual developer set it to false and restarted their vm.

Agreed

ryan-ludwig:
I think you're right that we need to see the cache functionality on stage and qa for sure. Let's chat some more about this offline, maybe we need to use contenful a little longer to see where all of the friction points are.

PR #1345, opened 2019-12-05
New Component banner area

eddieferrer:
* Consolidates logic of what should be shown above the header in 1 place.
* Fixes bug CASH-1629
* Removes duplicate code between appeal banner and global promo banner


BoulderBrains:
👀 👍 better late than never 😬

PR #1347, opened 2019-12-06
VUE-188 KvPillToggle Storybook and a11y improvements

ryan-ludwig:
- Change from @ clicks on spans to radio buttons for a11y. Add disabled, hover, active, and focus styles
- Add .stories for KvPillToggle
- Create Form Elements subfolder in Storybook

No visual differences from the previous component aside from the hover, focus, and active states being added. No changes to the props or events. Now usable with a keyboard. 

Verified on 
https://www.kiva.org/checkout
https://www.kiva.org/lend/filter?gender=female&sortBy=popularity

### DIFF
--- a/src/components/Checkout/BraintreeCheckout.vue
+++ b/src/components/Checkout/BraintreeCheckout.vue
@@ -104,11 +104,11 @@
 					<div id="kv-expiration-date" class="kv-braintree-wrapper"></div>
 				</div>
 				<div class="small-4 columns">
-					<label class="cvv-label">CVV</label>
+					<label for="kv-cvv" class="cvv-label">CVV</label>
 					<div id="kv-cvv" class="kv-braintree-wrapper"></div>
 				</div>
 				<div class="small-4 columns">
-					<label>Postal code</label>
+					<label for="kv-postal-code">Postal code</label>
 					<div id="kv-postal-code" class="kv-braintree-wrapper"></div>
 				</div>
 				<ul class="kv-multi-field-error-list"
--- a/src/components/Kv/KvPillToggle.vue
+++ b/src/components/Kv/KvPillToggle.vue
@@ -5,26 +5,23 @@
 			:key="option.key"
 			class="pill"
 		>
-			<input
-				class="radio"
-				type="radio"
-				:name="`pill-${uuid}`"
-				:id="`${option.key}-${uuid}`"
-				:class="option.key"
-				:disabled="option.disabled"
-				:checked="isSelected(option.key)"
-				@change="onChange(option.key)"
-			>
-			<label class="label" :for="`${option.key}-${uuid}`">
-				{{ option.title }}
+			<label class="label">
+				<input
+					class="radio"
+					type="radio"
+					:class="option.key"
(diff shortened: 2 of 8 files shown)


src/components/Checkout/BraintreeCheckout.vue
```diff
@@ -104,11 +104,11 @@
 					<div id="kv-expiration-date" class="kv-braintree-wrapper"></div>
 				</div>
 				<div class="small-4 columns">
```
ryan-ludwig: Sneaking this in, not related to KvPillToggle, just noticed it on the same page.

src/components/Kv/KvPillToggle.vue
```diff
@@ -1,34 +1,62 @@
...
+		<div
+			v-for="option in options"
+			:key="option.key"
+			class="pill"
 		>
-			<div class="title">
+			<input
+				class="radio"
+				type="radio"
+				:name="`pill-${uuid}`"
```
mcstover: why the uuid structure instead of just pulling index off of the v-for?

ryan-ludwig: It showed up when I put multiple kv-pill-toggle elements on a page that they'll have the same names / ids. So you click an input on the bottom kv-pill-toggle and the top one updates!

mcstover: Ahhh ok... So this is really a form input/label correlation specific issue. Just want to be careful of propagating this type of extra code throughout other components.

Should be ok with regard to form elements as there are only so many of them, though if it starts to go beyond that let's find a better way to eliminate the boilerplate.

eddieferrer: I don't like this. Is this a bug with storyBook? multiple components on the same page won't do this with Vue. If it is storybook related the fix should be in the story, not in the component

ryan-ludwig: @eddieferrer You're right! The bug wasn't in Storybook bug though, it was that my demo had all of the instances bound to the same Vue model. Doh! Will push a fix soon.

ryan-ludwig: Well now it seems like it was a combination of both duplicate model bindings and also duplicate IDs. By removing the IDs and for attributes on the labels and wrapping them instead it seems we're OK. For reference, here's a [paired down example](https://codesandbox.io/s/relaxed-cherry-1qfo2) of 
`<input id="test"><label for="test">`
 vs 
`<label><input></label>`.

src/components/Kv/KvPillToggle.vue
```diff
@@ -17,17 +26,34 @@ export default {
...
 		},
 		selected: {
 			type: String,
-			required: true
+			default: '',
 		},
 	},
+	data() {
+		return {
+			checked: ''
```
eddieferrer: Alternatively you can just bind checked to selected here and avoid the created hook all together...
```suggestion
			checked: this.selected
```

ryan-ludwig: Nice, thanks!

PR #1348, opened 2019-12-06
Cash 1533

BoulderBrains:


src/pages/Possibility/KivanSlider.vue
```diff
@@ -0,0 +1,371 @@
...
+				background-color: darken($cta-color3, 10%);
+			}
+		}
+	}
+	/* stylelint-enable */
+
+	// This is brute force solution to the varing heights of the content within this component
+	// In the furture slideshows should have a defined height, so we can avoid this customization.
+	.slide-placeholder {
+		padding-bottom: 14rem;
```
mcstover: Obvs not ideal but we can refactor later.

src/pages/Possibility/KivanSlider.vue
```diff
@@ -0,0 +1,371 @@
+<template>
+	<kv-carousel
+		:autoplay="false"
+		ref="KvCarousel"
+		@change="slideTransition($event)"
+	>
+		<div class="slide-placeholder"></div>
+		<div class="bullet-wrap text-center">
+			<div
```
ryan-ludwig: These bullets would be well served as a <button> element. Doing so gives us keyboard navigation out of the box, show the right cursor on hover, and announces to screenreaders better. Then we'll want to add :focus and :hover styling. The other piece for that is that inside of the button we'll need text, since you shouldn't have "empty" buttons. Something like "Go to slide ${index}". That text will have a class that hides it visually, but keeps it on the page for assistive technology, foundation has one, "show-for-sr".

src/pages/Possibility/KivanSlider.vue
```diff
@@ -0,0 +1,371 @@
...
+	margin-bottom: rem-calc(140);
+
+	.left-arrow,
+	.right-arrow {
+		z-index: 10000;
+		stroke: $white;
+		height: rem-calc(30);
+		width: rem-calc(30);
+		position: absolute;
+		top: 29%;
```
ryan-ludwig: We should be able to handle top without multiple media queries by using some calc() magic since we know the height of our arrows. Something along the lines of top: calc(50% - #{rem-calc(15)}).

src/pages/Possibility/KivanSlider.vue
```diff
@@ -0,0 +1,371 @@
...
+		<div class="slide-placeholder"></div>
+		<div class="bullet-wrap text-center">
+			<div
+				v-for="(slide , index) in slidesData"
+				:key="index"
+				:class="['bullet', {active: currentIndex === index}]"
+				@click="setSlide(index)"
+			></div>
+		</div>
+		<button @click="previousSlide">
```
ryan-ludwig: I think it would be good to have these 2 buttons be big fat hit targets. Probably fill the whole height of the carousel and have a little extra width. Also we should add :hover, :active, and :focus styling.

src/pages/Possibility/KivanSlider.vue
BoulderBrains: Creating a refactor ticket to handle early next week. https://kiva.atlassian.net/browse/CASH-1639

PR #1351, opened 2019-12-06
VUE-180 Sentry Upgrade

eddieferrer:
* Upgraded Sentry 2 major versions
* Changed exceptions to use new API

VUE-180

The second part to this is to connect Sentry with github, that is being done as well.


src/util/logReadQueryError.js
```diff
@@ -1,10 +1,9 @@
...
 
 export default (error, queryType = 'undefined_query_error_type') => {
 	console.error(queryType, error);
-	Raven.captureException(error, {
-		tags: {
-			query_error_type: queryType,
-		}
+	Sentry.withScope(scope => {
+		scope.setTag('query_error_type', queryType);
+		Sentry.captureException(error);
```
mcstover: @emuvente @eddieferrer 

I think we can drop the sentry call here and just leave the console.error which will continue to reveal the SSR invariant errors in our logs.

I'm going to take out the Raven call in the new release branch now.

emuvente: 👍

PR #1353, opened 2019-12-09
Add long term cache headers for static files

ryan-ludwig:
VUE-184

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -33,7 +33,7 @@ function setHeaders(res, path) {
 
 app.use(express.static('dist', {
 	setHeaders,
-	maxAge: '1d'
+	maxAge: '1y'
 }));
 
 // Setup Request Logger


ryan-ludwig:
@mcstover @emuvente It might be prudent to set this to 2days before going up to a year to make sure it's only applied to the files we want.

mcstover:
@emuvente I think your changes to the header settings are going out with this release...or did it go out with last release?

I don't want to push this until we verify that the client/browser/esp. Chrome has started to re-fetch or validate then refetch the files instead of reusing. Otherwise we'll have visitors getting very old content.

emuvente:
Yeah my changes should already be released, and also my changes only applied to the html we served, no the static files.

ryan-ludwig:
@mcstover What do you think about doing 2 days instead of 1 year for tomorrow's release? We'll be able to verify the effects with low risk.

PR #1362, opened 2019-12-11
Added ImageDensity check for MG promo grid cards

eddieferrer:
CASH-1426

### DIFF
--- a/src/components/LoanCards/PromoGridLoanCard.vue
+++ b/src/components/LoanCards/PromoGridLoanCard.vue
@@ -28,12 +28,18 @@
 
 <script>
 import KvButton from '@/components/Kv/KvButton';
+import { isHighDensity, isRetina } from '@/util/checkScreenDensity';
 
 const promoLoanImageRequire = require.context('@/assets/images/mg-promo-loan-card/', true);
 
 export default {
 	components: {
 		KvButton,
+	},
+	data() {
+		return {
+			isRetina: false,
+		};
 	},
 	props: {
 		experimentData: {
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/PromoGridLoanCard.vue
```diff
@@ -46,26 +52,33 @@ export default {
...
+			// if this.experimentVersion === 'control' or theres some error, return default
+			return promoLoanImageRequire(`./mg-promo-default-${this.imageDensity}.jpg`);
 		}
 	},
 	methods: {
 		trackInteraction(args) {
 			this.$emit('track-interaction', args);
 		}
 	},
+	mounted() {
```
mcstover: Would this work better to establish the retina url if it occurred in created?

eddieferrer: I was trying to keep it to client side only with mounted, similar how the KvResponsiveImage component does it. I dont think it matters in this particular case, the background image css can be changed on the fly after initial component load without any impact, and our check density utilities just return false if run on the server side.

mcstover: Ok, I'm mostly concerned with double image loading with the standard showing then loading the retina if we detect a higher density...This concern can be separated from this update though as it would affect other locations too.

emuvente: Calling this in created will result in them always being `false` as those check return false if window is not defined. Calling it in mounted is necessary to actually get a retina image at all

emuvente:
👍

PR #1365, opened 2019-12-11
KvRadio tuneup

ryan-ludwig:
- Use name property instead of name-set and label-set. 
- Add focus and disabled styles. 
- Wrap component in label el instead of using ids.
- Allow font-size to scale label without reaching inside the component styles. i.e., `<kv-radio style="font-size: 100px;">`Will scale both the circle and the label up
- Start the kitchen sink file.

VUE-192

### DIFF
--- a/src/assets/scss/_settings.scss
+++ b/src/assets/scss/_settings.scss
@@ -498,6 +498,17 @@ $button-smaller-medium-shadow: 0 2px $button-background-hover;
 	opacity: 0.3;
 }
 
+// Hide visually, but not from screenreaders
+// https://a11yproject.com/posts/how-to-hide-content/
+@mixin visually-hidden() {
+	position: absolute !important;
+	height: 1px;
+	width: 1px;
+	overflow: hidden;
+	clip: rect(1px, 1px, 1px, 1px);
+	white-space: nowrap;
+}
+
 // 11.2 Button - Foundation vars not currently used by Kiva
 // -----------
 
--- a/src/components/Kv/KvRadio.vue
+++ b/src/components/Kv/KvRadio.vue
@@ -11,11 +11,11 @@
 			v-bind="$attrs"
 		>
 		<label
-			class="wrapper"
+			class="label"
 			:for="id"
 		>
 			<div class="disc"></div>
-			<div class="label">
+			<div>
 				<slot></slot>
 			</div>
 		</label>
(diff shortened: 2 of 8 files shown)


ryan-ludwig:
I'm having some troubles logging in locally to verify these changes on the autolend page. Will double check tomorrow unless one of you can verify for me.

src/assets/scss/_settings.scss
```diff
@@ -498,6 +498,17 @@ $button-smaller-medium-shadow: 0 2px $button-background-hover;
 	opacity: 0.3;
 }
 
+// Hide visually, but not from screenreaders
```
eddieferrer: Foundation already has a utility class for this. Not sure if we want to use it or write our own

ryan-ludwig:
Right now <kv-radio> defaults to display: inline-block. How do you all feel about changing the default to block, so they stack vertically without any extra fuss? Currently it looks like all kv-radio elements are being styled to stack vertically.

eddieferrer:
> Right now `<kv-radio>` defaults to display: inline-block. How do you all feel about changing the default to block, so they stack vertically without any extra fuss? Currently it looks like all kv-radio elements are being styled to stack vertically.

Fine by me. I don't believe there are any horizontal kv-radios

src/components/Kv/KvRadio.vue
```diff
@@ -1,25 +1,29 @@
 <template>
-	<div class="styled-radio">
+	<label class="kv-radio">
```
mcstover: @ryan-ludwig we need to bring back the individual identifier on each individual radio control and label combo. This used to manifest like ":label-set="`${name}-radio-all`"" <- taken from autolend settings.

There may be a way to make this automatic based on the value but it's probably better to just have a prop that can then be applied to both the input and label. ie. `radioId = "group"` then the label has `id="label-group` and the input has `id="input-group"`

ryan-ludwig: @mcstover Is this for automation or some other purpose? Perhaps we can add the id to the outer element like `<kv-radio id="my-id">` and then automation could do `querySelector('#my-id label')`. Aiming to keep the API of the components themselves as simple and HTML-like as possible.

emuvente: For accessibility, labels should have a `for` attribute that matches the `id` attribute of the input element that it is a label for. It's best pract to do this even when nesting the input within the label.

mcstover: @ryan-ludwig Yes we can try that approach first.

mcstover:
@ryan-ludwig I think the stacked radios are a good default.

Could you also add the id attributes for the kv-radio component usages in this PR?

Then I think it will be good to go.

PR #1375, opened 2019-12-17
Intercom Messenger

eddieferrer:
* Implements intercom messenger on all pages based on build setting to enable it, and also experiment setting to control distribution

CASH-1680

### DIFF
--- a/config/dev.js
+++ b/config/dev.js
@@ -48,6 +48,9 @@ module.exports = merge(base, {
 		contentful: {
 			environment: 'development'
 		},
+		intercom: {
+			enable: false,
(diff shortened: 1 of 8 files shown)


config/dev.js
```diff
@@ -48,6 +48,9 @@ module.exports = merge(base, {
 		contentful: {
 			environment: 'development'
 		},
+		intercom: {
+			enable: false,
```
mcstover: let's enable in dev for testing. We can set the experiment setting to control(off) 100 but can turn on for a couple of verification tests with CS.

PR #1376, opened 2019-12-17
VUE-198 KvToggle 

ryan-ludwig:
- Make KvToggle's props and events mirror KvCheckbox. 
- Add focus, active, and disabled styling
- Add to storybook
- Update usage within the site

### DIFF
--- a/.storybook/stories/KvToggle.stories.js
+++ b/.storybook/stories/KvToggle.stories.js
@@ -1,5 +1,4 @@
 import KvToggle from '@/components/Kv/KvToggle';
-import { boolean } from '@storybook/addon-knobs';
 
 export default { title: 'Kv | Form Elements/KvToggle' };
 
@@ -7,19 +6,12 @@ export const Default = () => ({
 	components: {
 		KvToggle
 	},
-	props: {
-		option1: {
-			type: Boolean,
-			default: boolean('option1', false)
-		},
-		option2: {
-			type: Boolean,
-			default: boolean('option2', true)
-		},
-		option3: {
-			type: Boolean,
-			default: boolean('option3', false)
-		 },
+	data() {
+		return {
+			option1: false,
+			option2: true,
+			option3: false
+		}
 	},
 	template: `
 		<fieldset>
--- a/.storybook/stories/KvToggle.stories.js
+++ b/.storybook/stories/KvToggle.stories.js
@@ -1,4 +1,5 @@
 import KvToggle from '@/components/Kv/KvToggle';
+import { boolean } from '@storybook/addon-knobs';
 
 export default { title: 'Kv | Form Elements/KvToggle' };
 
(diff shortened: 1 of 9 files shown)


src/pages/Autolending/MainToggle.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-	<kv-toggle class="main-toggle" v-model="isEnabled">
+	<kv-toggle
+		class="main-toggle"
+		id="auto-lending-toggle"
```
mcstover: please do "auto-lending-enabled-toggle" here

src/pages/Autolending/CheckList.vue
```diff
@@ -6,9 +6,9 @@
 		>
 			<kv-toggle
 				class="list-item"
-				:name="id"
+				:id="`${id}`"
```
mcstover: can't we just use "id" here seeing as it's bound already... or is this to ensure it's cast to a string?

src/pages/Autolending/MainToggle.vue
```diff
@@ -1,5 +1,9 @@
...
+		v-model="isEnabled"
```
mcstover: Is v-model handled now? seems like it was removed in the new kv-toggle code.

src/pages/Autolending/CheckList.vue
ryan-ludwig: Yeah, this is just casting it to a string.

src/pages/Autolending/MainToggle.vue
ryan-ludwig: v-model is handled now with
```	
model: {
    prop: 'checked',
    event: 'change'
},
props: { ...
```
A little more info on that [here](https://vuejs.org/v2/guide/components-custom-events.html#Customizing-Component-v-model)

mcstover: Nice 👍

src/assets/scss/_settings.scss
```diff
@@ -493,6 +493,10 @@ $button-smaller-medium-shadow: 0 2px $button-background-hover;
 	}
 }
 
+@mixin input-focus() {
+	box-shadow: 0 0 0 0.25em rgba(174, 225, 92, 0.4); // $kiva-accent-green
+}
```
BoulderBrains: What if we updated this to use the variable name? 
IE: background-color: rgba($kiva-accent-green, 0.4); 
That way when/if we update the $kiva-accent-green color, we don't have to hunt down this hard coded value and update it?

.storybook/stories/KvToggle.stories.js
```diff
@@ -1,28 +1,157 @@
...
 
 export const Default = () => ({
 	components: {
 		KvToggle
 	},
-	data() {
-		return {
-			checked: true,
-		}
+	props: {
```
eddieferrer: You can't mutate props directly like this. See console: 

[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "option3"

src/pages/Autolending/CheckList.vue
eddieferrer: We should change this in the Checklist component. Right now it renders: id="<number>" and similarly in the for attribute, I think these should be descriptive, for the sake of automation testing as well. It should render an identifiable string, based on the name. So something like id="<name>" The name is unfortunately not slug friendly. See: src/components/WwwFrame/LendMenu/LendListMenu.vue Line 31

src/assets/scss/_settings.scss
ryan-ludwig: Love it

.storybook/stories/KvToggle.stories.js
ryan-ludwig: Good catch, switched to data() to demonstrate the v-model stuff

PR #1377, opened 2019-12-18
Give .stories access to the same plugins as the app

ryan-ludwig:


mcstover:
@ryan-ludwig I'm ok with adding this but we should call out or document explicitly when these are required and should also try to ensure the base components don't require these plugins as that will make it harder to use them in other apps without also importing the plugins...

eddieferrer:
Why don't we just add them on a per story basis. Like I did in: .storybook/stories/PromoGridLoanCard.stories.js ? If we're going to add them globally, like this, please remove the addition from .storybook/stories/PromoGridLoanCard.stories.js

ryan-ludwig:
@mcstover @eddieferrer A per-story basis sounds better 👍.

PR #1378, opened 2019-12-19
Sample story for component that depends on Apollo

eddieferrer:
* Story clean  up
* Added apollo mocking for global promo component
* Added svg sprite generation for storybook

VUE-196

### DIFF
--- a/.storybook/stories/GlobalPromoBanner.stories.js
+++ b/.storybook/stories/GlobalPromoBanner.stories.js
@@ -61,8 +61,6 @@ export const Default = () => ({
 	components: {
 		GlobalPromoContentful,
 	},
-	props: {
-	},
 	template: `
 		<global-promo-contentful />
 	`,
--- a/.storybook/stories/GlobalPromoBanner.stories.js
+++ b/.storybook/stories/GlobalPromoBanner.stories.js
@@ -6,7 +6,7 @@ import GlobalPromoContentful from '@/components/WwwFrame/PromotionalBanner/Globa
 import kivaPlugins from '@/plugins';
 Vue.use(kivaPlugins)
 
-let mockedContentfulAPIResponse = {
+const mockedContentfulAPIResponse = {
 	data: {
 		contentfulCMS: {
 			items: [{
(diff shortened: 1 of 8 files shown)


.storybook/stories/GlobalPromoBanner.stories.js
```diff
@@ -0,0 +1,69 @@
+import Vue from 'vue'
+import StoryRouter from 'storybook-vue-router';
+import GlobalPromoContentful from '@/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful';
+
+// import plugins
+import kivaPlugins from '@/plugins';
+Vue.use(kivaPlugins)
+
+let mockedContentfulAPIResponse = {
```
ryan-ludwig: Nitpick, should this be a const instead?

eddieferrer: 👍

mcstover:
Hmmm. I think this speaks to us needed to decouple the datasource and the markup for the global promo banner. We should setup global promo banner so it accepts an icon id, link, copy and click track data.

I think there is something useful in being able to render a component like this from mocked data but even then that could be a data provider for the base version of global promo banner.

eddieferrer:
> Hmmm. I think this speaks to us needed to decouple the datasource and the markup for the global promo banner. We should setup global promo banner so it accepts an icon id, link, copy and click track data.
> 
> I think there is something useful in being able to render a component like this from mocked data but even then that could be a data provider for the base version of global promo banner.

Yeah I agree, the component that is decoupled would be the <generic-promo-banner>, there might not be too much value in making a story for this component as opposed to the <generic-promo-banner> component...

eddieferrer:
Added some additional improvements: 
* src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue is just a wrapper now that gets contentful data and passes it to the generic promo banner
* generic promo banner is dumber, just takes in data now. 
* kept the contentful banner story as an example of how to do a story for component that depends on this.apollo.query with fake content
* added more useful story for generic promo banner with knobs for text/icon/link

PR #1380, opened 2019-12-19
VUE-199 KvRangeSlider

ryan-ludwig:
- Create a presentational component based on the slider in the lend filter page.
- Allow v-model or :value binding
- Ensure a label element is used and that the value of the slider is not part of the label.
- Update usage in the site

### DIFF
--- a/src/components/Kv/KvRangeSlider.vue
+++ b/src/components/Kv/KvRangeSlider.vue
@@ -58,19 +58,9 @@ export default {
 	},
 	computed: {
 		sliderStyle() {
-			const sliderLeftColor = '#8ccb8c';
-			const sliderRightColor = 'transparent';
-
 			const snapped = Math.round(this.value / this.step) * this.step; // snap to the nearest :step
-			const selectedStop = Math.abs(snapped - this.min) / Math.abs(this.max - this.min) * 100;
-
-			const basicStyle = 'background-image: linear-gradient(to right,';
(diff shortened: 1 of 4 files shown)


src/components/Kv/KvRangeSlider.vue
```diff
@@ -0,0 +1,220 @@
...
+		}
+	},
+	methods: {
+		onInput($event) {
+			this.$emit('input', parseFloat($event.target.value, 10));
+		},
+		onChange($event) {
+			this.$emit('change', parseFloat($event.target.value, 10));
+		},
+		sliderStyle() {
```
eddieferrer: This should be a computed property instead of a method.

ryan-ludwig:
@BoulderBrains Gave the CSS a refactor based on lessons [learned here](https://css-tricks.com/sliding-nightmare-understanding-range-input/,) 🤯 we're up to around 20px hit target area now. Still not amazing for touch, but better than 6px. Tested on Firefox, Chrome, Safari. Downloading XCode for iOS emulation now.

mcstover:
@ryan-ludwig is our scss compilation going to pick up and make the css variables compatible with IE?

ryan-ludwig:
@mcstover Good question. No, it won't make them compatible, but in this instance IE isn't using them. Webkit and blink are using that fancy CSS var gradient, while IE is using it's own "::-ms-fill-lower " pseudo element that uses a regular background color.

PR #1383, opened 2019-12-20
remove failed autolend test screenshots

mcstover:


eddieferrer:
Should we add these to gitignore?

mcstover:
@eddieferrer Yea I think if we aren't worried about collecting these we should either gitignore or tell cypress not to save them...

eddieferrer:
> @eddieferrer Yea I think if we aren't worried about collecting these we should either gitignore or tell cypress not to save them...

I added them to .gitignore, the screenshots are very helpful when running cypress in headless mode to see the failed test

PR #1384, opened 2019-12-20
Fix prop valued passed to OptIn controls component and ensure that is…

mcstover:
…n't overwritten.


mcstover:
@BoulderBrains @eddieferrer Fixed these things after finding the issue in dev.

PR #1387, opened 2020-01-02
Managed some UI dependencies

eddieferrer:
* Removed acorn and jquery dependencies
* Updated ora and winston dependencies to latest version


package.json
```diff
@@ -72,7 +72,6 @@
 		"https": "^1.0.0",
 		"instantsearch.js": "^3.6.0",
 		"isomorphic-fetch": "^2.2.1",
-		"jquery": "^3.4.1",
```
ryan-ludwig: 🎉

mcstover: Now we get the warning about missing jquery and we prob added just to suppress that...we an update coming for foundation-sites soon wherein I'll explore moving it into dev deps anyhow...

PR #1391, opened 2020-01-04
Update apollo-client, foundation-sites and memory-fs

mcstover:


build/style-loaders.js
```diff
@@ -21,11 +21,14 @@ module.exports = [
 		loader: 'sass-loader',
 		options: {
 			implementation: sass,
-			includePaths: [
-				'src/assets/fonts',
-				'src/assets/scss',
-				'node_modules/foundation-sites/scss'
-			]
+			sassOptions: {
+				// fiber: false, // to disable automatic use of fibers package
```
mcstover: sass-loader 8 will use fibers by default if it's present. We can disable it if it's too finicky. I've found that the trick is to make sure you run npm i from within your vm to be sure the fibers post-install tasks run. After this it works fine and saves me 3-5 seconds on subsequent compilations...

src/components/Checkout/LoanPrice.vue
```diff
@@ -264,7 +264,6 @@ export default {
 
 	@include breakpoint(medium) {
 		height: inherit;
-		line-height: $medium-text-line-height;
```
mcstover: This change was required after updating foundation-sites and I believe it was due to some changes in one of their scss files. I've reviewed all the major Ui pages and all the stories in storybook and didn't find any other anomalies.

mcstover:
Also you'll see new warnings during npm install for unnecessary optional peer deps for foundation-sites and for sass-loader, none of which need any further action.
ref. https://github.com/webpack-contrib/sass-loader/issues/749

PR #1395, opened 2020-01-07
cash-1645 thanks page in place with routes

BoulderBrains:


src/router/routes.js
```diff
@@ -28,6 +28,7 @@ export default [
 	{ path: '/lend/filter', component: () => import('@/pages/Lend/Filter/LendFilterPage') },
 	{ path: '/page-two', component: () => import('@/pages/PageTwo') },
 	{ path: '/portfolio/lending-stats', component: () => import('@/pages/LendingStats/LendingStatsPage') },
+	{ path: '/thanks', component: () => import('@/pages/Thanks/Thanks') },
```
mcstover: I think this is good for now as we won't be ready for launch this sprint... we'll need the .htaccess entry too.

@emuvente, do you have any thoughts on if we are even able to use the existing path `/basket/thanks` when we have 2 versions of thanks... seems like our current proxy setup won't allow use to use that path in both Ui and Php...

BoulderBrains: I added the .htaccess entry to this other code review: https://github.com/kiva/kiva/pull/8158

mcstover: Updated my comment above.

BoulderBrains: with my latest commit I've updated the route for the new thanks page to be /checkout/thanks since /thanks is taken by our existing thanks page.

src/router/routes.js
```diff
@@ -28,6 +28,7 @@ export default [
 	{ path: '/lend/filter', component: () => import('@/pages/Lend/Filter/LendFilterPage') },
 	{ path: '/page-two', component: () => import('@/pages/PageTwo') },
 	{ path: '/portfolio/lending-stats', component: () => import('@/pages/LendingStats/LendingStatsPage') },
+	{ path: '/checkout/thanks', component: () => import('@/pages/Thanks/Thanks') },
```
emuvente: It's strongly recommended that components are named with multiple words: https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential (`ThanksPage.vue` would work fine)

BoulderBrains: File name updated in order to prevent conflicts with existing or future HTML elements. Thanks bringing this up Shua.

BoulderBrains:
This is now fixed up @emuvente

PR #1396, opened 2020-01-07
Thanks page starting point

ryan-ludwig:
A starting point with some data mocks in storybook for the receipt and social share components. 

Sass is in BEM as I write quickly that way, if that's offputting or difficult to read, I'm happy to change it.

Both components are half-baked but wanted to get a branch up sooner than later. Is this how you guys envision the data being used and structured?

CASH-1647
CASH-1646
CASH-1694


src/components/Checkout/SocialShare.vue
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		&--email {
+			color: $kiva-textlink;
+		}
+	}
+
+	&__icon {
+		width: rem-calc(24px);
+		height: rem-calc(24px);
```
BoulderBrains: In the past when we've used rem-calc we leave off the px value. As long as this works I don't think it really matters, but thought I would mention it.

ryan-ludwig: Good to know, thanks!

mcstover:
@ryan-ludwig Looks like it's off to a great start. Please be sure to remove/hide the social share stuff from the print view.

I didn't know we already had bem compilation enabled with our setup so that's cool too!

emuvente:
@mcstover I think that hiding of the social share section can be handled when we do the print styles in CASH-1648

src/pages/Thanks/ThanksPage.vue
```diff
@@ -1,20 +1,113 @@
...
+			receipt: null,
+			isReceiptVisible: true,
+		};
+	},
+	apollo: {
+		query: checkoutReceiptQuery,
+		preFetch: true,
+		preFetchVariables() {
+			return {
+				checkoutId: 38649558, // TODO, have this passed in from the checkout page
```
ryan-ludwig: Oh, probably should fix this before merge. @emuvente I see a param coming across on /thanks?kiva_transaction_id=555555... Not quite sure how to get at that though. Thoughts?

emuvente: The `preFetchVariables` function is called with the route object for the url that is being loaded, so you can access it like this:
```
    return {
        checkoutId: route.query.kiva_transaction_id,
    };
}
```
For the `variables` function, you can access the current route through the component instance:
```
variables() {
    return {
        checkoutId: this.$route.query.kiva_transaction_id,
    };
}
```

ryan-ludwig: Awesome thanks!

src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -0,0 +1,258 @@
+<template>
+	<div class="checkout-receipt">
+		<h2 class="checkout-receipt__headline">
+			Order Confirmation
+		</h2>
+		<section class="section">
+			TODO: Date<br>
```
emuvente: It looks like we'll need to add this (transaction create time) to graphql so that we can display it here. Is there a ticket for that yet?

ryan-ludwig: No ticket that I could find.

src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -0,0 +1,258 @@
...
+						<div class="loan__amount">
+							${{ item.price }}
+						</div>
+					</div>
+					<div
+						class="loan"
+						v-else-if="item.basketItemType === 'kiva_card'"
+					>
+						<h3 class="loan__name">
+							Kiva Card
```
emuvente: For this section, I think we should mimic what what was done here https://github.com/kiva/kiva/blob/development/httpdocs/views/Basket/ThanksView.tpl#L63-L122

emuvente: This one is a little complicated when you include the “print kiva card” link, but we could do that as a separate ticket. That also involves displaying extra messages about the print-type gift cards https://github.com/kiva/kiva/blob/development/httpdocs/views/Basket/ThanksView.tpl#L17-L30

src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -0,0 +1,258 @@
...
+						<div class="loan__amount">
+							${{ item.price }}
+						</div>
+					</div>
+					<div
+						class="loan"
+						v-else-if="item.basketItemType === 'donation'"
+					>
+						<h3 class="loan__name">
+							Donation
```
emuvente: The old receipt has the donation included in the totals section, rather than as a line item https://github.com/kiva/kiva/blob/development/httpdocs/views/Basket/ThanksView.tpl#L125-L130

src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -0,0 +1,258 @@
...
+		</section>
+		<section class="section split total">
+			<h3 class="total__header split__item">
+				Total:
+			</h3>
+			<span class="total__amount split__item split__item--end">${{ receipt.totals.itemTotal }}</span>
+		</section>
+		<section class="section payments">
+			<h2 class="payments__header">
+				Payments
```
emuvente: I think we should include all of these payment types https://github.com/kiva/kiva/blob/development/httpdocs/views/Basket/ThanksView.tpl#L163-L197. And probably in the order that they're put in there too. These totals are all available through graphql already.

src/components/Checkout/CheckoutReceipt.vue
mcstover: Added CASH-1754 for this

PR #1400, opened 2020-01-11
Cash 1714

BoulderBrains:
How does monthly good work? component.


src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -0,0 +1,156 @@
...
+		<div class="row">
+			<h1 class="column small-12 text-center">
+				How does Monthly Good work?
+			</h1>
+			<div class="steps-container">
+				<div class="column small-12 medium-6 large-3 step-wrapper">
+					<span class="number">
+						1
+					</span>
+					<kv-icon name="hp-mg-1-signup" class="mg-image" />
```
ryan-ludwig: Since these .svgs are only used on the homepage I think they would be good candidates for loading as inline-svgs rather than adding to the svg sprite. By adding to the sprite that data will be loaded on every page.

There's an example of this inline loading on one of the year-end promo [pages with the "sketchy arrow"](https://github.com/kiva/ui/blob/master/src/pages/Possibility/YearEnd.vue#L71). Basically the .svg file needs to be in assets/inline-svgs and then you import it and use it as if it were a Vue component.

src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+	<div class="monthly-good-explained">
+		<div class="row">
```
emuvente: This div could be combined with the previous one to reduce the number of dom nodes used (good for page performance) `<div class="monthly-good-explained row">`

src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -0,0 +1,156 @@
...
+			<h1 class="column small-12 text-center">
+				How does Monthly Good work?
+			</h1>
+			<div class="steps-container">
```
emuvente: I would recommend removing this wrapper div since it's not changing the way things are displayed and it would cut down on the number of dom nodes used

src/components/Homepage/MonthlyGoodExplained.vue
eddieferrer: I agree with Ryan on this, I think it would be great to be able to move away from the SVG sprite going forward

src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -0,0 +1,156 @@
...
+			</div>
+			<div class="column large-12 text-center learn-more-wrapper">
+				<router-link
+					:to="`/monthlygood`"
+					class="learn-more-text"
+					v-kv-track-event="[
+						'homepage',
+						'click-monthly-good-Learn more',
+						'homepage-monthly-good-learn-more-link',
+						'true'
```
eddieferrer: This 4th param is unnecessary. The param just resolves to NaN in kvTrackEvent and isn't registered in snowplow/looker anyway. Also remove from the WhyKiva component

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: @emuvente Collapsing this into one line. "<div class="row monthly-good-explained"> causes me to lose the ability to color the entire section of the page (to the edges of the page). Currently I  have "monthly-good-explained" background color set to green, which extends to the edges of the page. Adding the row class to this div, prevents the green from running to the edges of the page and contains it to the grid. I'm up to other suggestions, but that's why I did it this way.

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: Removed.

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: Removed the 4th parameter in monthlyGoodExplained and WhyKiva

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: I've successfully transitioned to pulling this svg inline. However, when I attempt to load the homepage up now, it's throwing an error for each of the svgs i'm trying to import. 
"You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file."
- I'm wondering if I have to export these svgs differently because the 'sketchy-arrow.svg' file looks pretty different than any of my svg files ie: 'hp-mg-1-signup.svg'. 
- I also don't see any type of additional loader that was used for the 'sketchy-arrow.svg' file which leads me to believe i'll have to re-export these svg in a different way... 
Would love any tips or suggestions related to a fix for this.

eddieferrer: @BoulderBrains the svgs have to be inside of the inline-svg folder so that they get processed by the correct webpack loader

BoulderBrains: Thank you @eddieferrer

BoulderBrains:
@eddieferrer the spacing adjustments you previously mentioned have been corrected in my latest commit.

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: @ryan-ludwig pointed me to lines 89 -> 95 in webpack.base.conf.js which explained the additional loader required to load these svgs inline. Thanks for that @ryan-ludwig

BoulderBrains:
@eddieferrer @ryan-ludwig & @emuvente 
I believe all suggestions have been corrected, this is ready for another review.

src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -0,0 +1,162 @@
...
+		margin: 0 auto;
+		line-height: 1.4;
+
+		@include breakpoint(medium) {
+			margin: unset;
+		}
+	}
+
+	.mg-image {
+		width: 45%;
```
eddieferrer: The svgs are getting cut off at browser width ~681-690. I believe we can just remove this line completely.

src/components/Homepage/MonthlyGoodExplained.vue
emuvente: 👍 That makes sense, sorry I didn't notice that css

src/components/Homepage/MonthlyGoodExplained.vue
BoulderBrains: Corrected.

BoulderBrains:
Thanks for the taking the time to review this team.

PR #1404, opened 2020-01-15
Cash 1734 mg landing form

eddieferrer:
* Implemented MG form that is repeated twice on the page, forms sync values
* Form has validation using vuelidate library
* Form amount input is a currency type input with input sanitation and formatting and rounding

CASH-1734

** Remember to add - `RewriteCond %{REQUEST_URI} ^/monthlygood-ui$ [OR]` - to your server side .htaccess to be able to see the new page.

### DIFF
--- a/src/components/Kv/KvCurrencyInput.vue
+++ b/src/components/Kv/KvCurrencyInput.vue
@@ -1,0 +1,51 @@
+<template>
+	<input
+		type="text"
+		v-model="displayValue"
+		@blur="isInputActive = false"
+		@focus="isInputActive = true"
+		placeholder="$"
+	>
+</template>
+
+<script>
+export default {
+	props: {
+		value: {
+			type: Number,
+			default: 0,
+		}
+	},
+	data() {
+		return {
+			isInputActive: false
+		};
+	},
+	computed: {
+		displayValue: {
+			get() {
+				if (this.isInputActive) {
+					// Cursor is inside the input field. unformat display value for user
+					return this.value.toString();
+				}
+				// User is not modifying now. Format display value for user interface
+				return `$ ${this.value.toFixed(2).replace(/(\d)(?=(\d{3})+(?:\.\d+)?$)/g, '$1,')}`;
+			},
+			set(modifiedValue) {
+				// Recalculate value after ignoring "$" and "," in user input
+				let newValue = parseFloat(modifiedValue.replace(/[^\d\.]/g, ''));
+				// Ensure that it is not NaN
+				if (Number.isNaN(newValue)) {
+					newValue = 0;
+				}
+				// Round to 2 decimal places
+				if (!Number.isInteger(newValue)) {
+					newValue = parseFloat(newValue.toFixed(2));
+				}
+				// $emit the event to the parent component
+				this.$emit('input', newValue);
+			}
+		}
+	}
+};
+</script>
--- a/src/components/Kv/KvCurrencyInput.vue
+++ b/src/components/Kv/KvCurrencyInput.vue
@@ -9,6 +9,8 @@
 </template>
 
 <script>
+import numeral from 'numeral';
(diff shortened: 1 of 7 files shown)


src/components/Kv/KvCurrencyInput.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+	<input
```
ryan-ludwig: I think we'll need to add "id" as a required prop and use it on this input here so that <label for="my-cool-id"> outside of this component point to the input.

src/components/Kv/KvCurrencyInput.vue
```diff
@@ -0,0 +1,51 @@
...
+			default: 0,
+		}
+	},
+	data() {
+		return {
+			isInputActive: false
+		};
+	},
+	computed: {
+		displayValue: {
```
ryan-ludwig: 👀 this is slick

src/pages/MonthlyGood/LandingForm.vue
```diff
@@ -0,0 +1,126 @@
+<template>
+	<form @submit.prevent.stop="submit" novalidate>
+		<fieldset>
+			<label class="input-label" :class="{ 'error': $v.$invalid }" for="amount">
+				<kv-currency-input id="amount" :value="amount" @input="updateAmount" />
+				<ul class="validation-errors">
```
ryan-ludwig: Nitpick, this ul will always be on the dom even if no errors. Perhaps a `v-if="$v.$invalid"`.

src/pages/MonthlyGood/LandingForm.vue
```diff
@@ -0,0 +1,126 @@
+<template>
+	<form @submit.prevent.stop="submit" novalidate>
+		<fieldset>
+			<label class="input-label" :class="{ 'error': $v.$invalid }" for="amount">
+				<kv-currency-input id="amount" :value="amount" @input="updateAmount" />
```
ryan-ludwig: I think we'll want to have `<kv-currency-input>` outside of the the `<label>` element since we're the for and id attributes.
We'll also want to have some text inside the label even if it's visually hidden for screen readers to pick up.

src/components/Kv/KvCurrencyInput.vue
mcstover: agree

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -3,17 +3,27 @@
 		<div class="row page-content">
 			<div class="small-12 columns">
 				Monthly Good page placeholder text
+				<landing-form :amount.sync="monthlyGoodAmount" :selected-category.sync="selectedCategory" />
+				<landing-form :amount.sync="monthlyGoodAmount" :selected-category.sync="selectedCategory" />
```
ryan-ludwig: duplication?

src/components/Kv/KvCurrencyInput.vue
eddieferrer: This is already happening. The id from here -> `<kv-currency-input id="amount"` gets rendered unto the input element

src/components/Kv/KvCurrencyInput.vue
```diff
@@ -0,0 +1,51 @@
...
+			get() {
+				if (this.isInputActive) {
+					// Cursor is inside the input field. unformat display value for user
+					return this.value.toString();
+				}
+				// User is not modifying now. Format display value for user interface
+				return `$ ${this.value.toFixed(2).replace(/(\d)(?=(\d{3})+(?:\.\d+)?$)/g, '$1,')}`;
```
mcstover: I think you could use numeral in all of these cases to flip back and forth between the various formats + values...

PR #1405, opened 2020-01-16
CASH-1649 log purchase events

emuvente:


src/pages/Checkout/PostPurchase.vue
```diff
@@ -0,0 +1,63 @@
...
+								campaignContent,
+								medium,
+								snowplowSessionId,
+								snowplowUserId,
+								source: gclid || route.query.gclid ? 'google-cpc' : source,
+								transactionId,
+							},
+						})
+							// whether it succeeded or failed, redirect to the thanks page with the transaction id
+							.then(() => reject(successRoute))
```
eddieferrer: We can use .finally here and avoid the duplication of the .then/.catch

emuvente: I didn't use .finally because it's not supported in node 8: https://node.green/#ES2018-features-Promise-prototype-finally

eddieferrer: 👍

PR #1410, opened 2020-01-18
CASH-1648 add print styles for global elements and thanks page

emuvente:


src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -125,7 +125,6 @@
 					v-if="kivaCards.length > 0"
 				>
 					For more details about all your Kiva Card purchases, please visit your
-					<!-- TODO: is this the tracking we want? -->
```
emuvente: I removed this because this comment node was causing mismatched rendering error in the browser (which causes a full render to run and makes weird things happen with hot reloading)

PR #1413, opened 2020-01-22
Prevent errors when checking out without a purchased loan

ryan-ludwig:
Perhaps you only had a donation


ryan-ludwig:
Merging to keep QA going

PR #1414, opened 2020-01-22
CASH-1698 Copy url to clipboard

ryan-ludwig:


src/components/Checkout/SocialShare.vue
```diff
@@ -88,20 +97,29 @@ export default {
...
-			maxMessageLength: 280
+			selectedLoanIndex: 0,
 		};
 	},
 	computed: {
+		selectedLoan() {
+			return this.loans[this.selectedLoanIndex];
+		},
+		selectedLoanUrl() {
+			return `https://www.kiva.org/lend/${this.selectedLoan.id}`;
```
ryan-ludwig: @emuvente This could be useful for the other social share pieces

mcstover: I think this needs to include the lender specific user name (unless anonymous)... is this the format that is currently used on the legacy thanks page?

emuvente: @mcstover yeah there are other things that need to be included for that whole share link. I'm adding that to my branch where I'm doing the other social share buttons. I can update the url the copy link uses with that so we can merge this pr

src/components/Checkout/SocialShare.vue
```diff
@@ -88,20 +97,29 @@ export default {
...
+			},
+			maxMessageLength: 280,
 			message: '',
-			maxMessageLength: 280
+			selectedLoanIndex: 0,
 		};
 	},
 	computed: {
+		selectedLoan() {
+			return this.loans[this.selectedLoanIndex];
```
emuvente: We don't always have a loan on the thanks page, so it would be good to include a fallback here like `return this.loans[this.selectedLoanIndex] || {};`

src/components/Checkout/SocialShare.vue
```diff
@@ -88,20 +97,29 @@ export default {
...
 		placeholderMessage() {
-			return `Why did you lend to ${this.loans[this.selectedLoanIndex].name}?`;
+			return `Why did you lend to ${this.selectedLoan.name}?`;
 		},
 		suggestedMessage() {
-			const selectedLoan = this.loans[this.selectedLoanIndex];
-			const { name } = selectedLoan;
-			const location = selectedLoan.geocode.city || selectedLoan.geocode.country.name;
-			return `Kiva is an easy way to make a real difference in someone's life. Will you join me in helping ${name} in ${location} to pursue their dream?`; // eslint-disable-line max-len
+			const location = this.selectedLoan.geocode.city || this.selectedLoan.geocode.country.name;
```
emuvente: Since the selected loan won't always be defined, I would recommend using lodash.get here:
```
import _get from 'lodash/get';
...
const location = _get(this, 'selectedLoan.geocode.city') || _get(this, 'selectedLoan.geocode.country.name');
```

src/components/Checkout/SocialShare.vue
ryan-ludwig: 👍 I merged [a PR](https://github.com/kiva/ui/pull/1413) this morning that conditionally displays this social block only if we have loans.

emuvente: That's good! We should still write this code defensively in case this component gets used somewhere that forgets to add that check. "Cannot read x of undefined" errors are difficult to track down, so it's good practice to write components that can handle their props being malformed. Lodash.get can help a lot with that, especially when you need to get a deeply-nested piece of data.

PR #1415, opened 2020-01-22
Cash 1688

BoulderBrains:
These changes get some auto-lending explanation text displayed on the /settings/autolending page which inform users when they should expect auto-lending to be triggered and their funds are lent out.


BoulderBrains:
@mcstover @eddieferrer @emuvente 
I did some work around writing cypress tests for this feature, but ran into an issue while trying to mock the userBalance/balance, see my notes in AutolendingPage.spec.js. Most of the tests are written, I just need to be able to mock that data for the test to pass. At this point i'm pretty invest in the fix for this, if any of you have any ideas to get the mock data set correctly, please let me know. 

For now the test are commented out, because the code is good. @eddieferrer I added an additional null check into the vue file while should protect against the case you brought up to me. 

Let me know what you guys think and thanks for the review.

test/e2e/fixtures/wwwPageMock.js
```diff
@@ -22,7 +24,8 @@ export default function wwwPageMock(userId = null) {
 			redemptionCodeAvailableTotal: '0.00',
 		}),
 		UserAccount: () => ({
-			id: userId
+			id: userId,
```
emuvente: Based on what I've investigated, the root of the issue comes from this, `id: userId`, meaning that this problem existed before your pr, you just were unlucky enough to uncover it. It fails because when that gets serialized and sent over the wire to the mock api, `userId` doesn't get replaced with the number, it's just sent literally as `userId`. The resolver fails to read that undeclared identifier, and so it throws out the whole result. For now, I would recommend just hard-coding this as `id: 42`.

To actually set the balance in each of your tests, I recommend only using `Money: () => ...` within the cy.mock call in each test. Remove the `balance: 0` and `Money: () => '30',` from this file, because otherwise you won't be able to override the balance in each of your tests.

test/e2e/specs/AutolendingPage.spec.js
```diff
@@ -88,4 +88,76 @@ describe('Autolending Page Spec', () => {
 			// assert toggle says on
 		});
 	});
+
+	describe('Autolend explanation text', () => {
+		it('Verify user balance below lendable amount', () => {
+			// Mock date and idle start time
+			cy.mock({
+				AutolendProfile: () => ({
+					cIdleStartTime: null,
```
emuvente: You should also mock out isEnabled and enableAfter in each of these tests, because otherwise they will be assigned random values which can change what text is displayed.
```
isEnabled: true,
enableAfter: 0,
```

test/e2e/specs/AutolendingPage.spec.js
```diff
@@ -88,4 +88,76 @@ describe('Autolending Page Spec', () => {
...
+		// 		// Mock cIdleStartTime and lendAfterDaysIdle
+		// 		cy.mock({
+		// 			AutolendProfile: () => ({
+		// 				cIdleStartTime: '2020-01-17T00:00:00',
+		// 				lendAfterDaysIdle: 7,
+		// 				donationPercentage: 5
+		// 				// userBalance: '5', Nope
+		// 				// balance: '5' Nope
+		// 			}),
+		// 			Date: () => cy.clock(1579651200000),
```
emuvente: The cy.clock call should just be on it's own, not inside the mock

src/pages/Autolending/LendTimingDropdown.vue
```diff
@@ -75,6 +80,30 @@ export default {
...
 			notice += this.changedTiming ? '.' : ', the closest matching setting above will be applied upon save.';
 			return notice;
+		},
+		autolendExplanationText() {
+			const now = Date.now();
+			const idleStartTime = Date.parse(this.cIdleStartTime);
+			const daysIdle = differenceInCalendarDays(now, idleStartTime);
+			const daysUntilLend = this.lendAfterDaysIdle - daysIdle;
+			const userBalance = parseFloat(this.userBalance).toFixed(2);
+			const loanAndDonationAmount = ((1 + this.donationPercentage / 100) * 25).toFixed(2);
```
emuvente: You shouldn't use `toFixed` for the userBalance and the loanAndDonationAmount because it's turning them into strings instead of numbers, which results in them being compared alphabetically rather than numerically. For instance, `5 > 26 === false`, but `'5' > '26' === true` because when compared as strings they are compared character by character, and 5 is bigger than 2, so string '5' is therefore bigger than string '26'

BoulderBrains: Changed to use numeral, which on all accounts is much much better.

test/e2e/specs/AutolendingPage.spec.js
BoulderBrains: Done

test/e2e/fixtures/wwwPageMock.js
BoulderBrains: Thanks for explaining that @emuvente

BoulderBrains:
Ready for what I hope is the last round of reviews, finally managed to get the cy.mock data working for these cypress tests.

src/pages/Autolending/LendTimingDropdown.vue
emuvente: Using numeral's `format` function still gives you the same problem though, as that's returning string and not numbers.

emuvente: The problem is that you want to use `userBalance` and `loanAndDonationAmount` as numbers when you are comparing them (so that you don't end up with '26' being considered smaller than '5') but you also want to use them as formatted strings in the messages that you print out so that they look good. You can make a test that reveals this problem by setting an non-null cIdleStartTime and a balance that is greater than 3 and less than 25. In those conditions, you would just expect the message to contain 'Your current balance is lower', but when you run the test you'll find that it's actually saying that the balance is eligible

emuvente: One way you could fix it is by using different variables for the numbers and the formatted strings
```
const userBalance = numeral(this.userBalance).value();
const userBalanceFormatted = numeral(this.userBalance).format('0,0.00');
const loanAndDonationAmount = (1 + this.donationPercentage / 100) * 25;
const loanAndDonationAmountFormatted = numeral(loanAndDonationAmount).format('0,0.00');
```
That allows `userBalance` and `loanAndDonationAmount` to be safely used in comparisons (like `if(userBalance < loanAndDonationAmount)`) and then `userBalanceFormatted` and `loanAndDonationAmountFormatted` can be used in the template strings for the message

BoulderBrains:
@mcstover @eddieferrer This is ready for a re-review when you two have time, thanks 🙏

PR #1423, opened 2020-01-23
CASH-1695 CASH-1696 CASH-1697 thanks page share buttons

emuvente:


src/components/Checkout/SocialShare.vue
```diff
@@ -50,18 +50,31 @@
...
+			>
 				<icon-facebook class="social__icon" />
 				<span>Share</span>
-			</button>
-			<button class="social__btn social__btn--twitter">
+			</a>
+			<a
+				class="social__btn social__btn--twitter"
+				:href="twitterShareUrl"
+				target="_blank"
```
ryan-ludwig: Probably a good idea to add rel="noopener" to all of these target="_blank" links.
https://developers.google.com/web/tools/lighthouse/audits/noopener

emuvente: 👍  yeah that looks like a good idea

PR #1426, opened 2020-01-24
New Monthly Good Page Content

eddieferrer:
Apologies for the large amount of changes in this PR. Here is a summary of the changes:

1. Minor fixes to storybook stories. 
2. Hero asset downloaded from fastly and added as responsive image component. 
3. svgs used in the old page moved over to be inline svgs. 
4. 10 years one billion image cleaned up, and fixed here and in the /protocol page (the previous image was squished and not in the appropriate ratio. (thanks @ryan-ludwig for helping with this)
5. form validation css styles copied over from auth0 and applied to form on this page, removed other duplicate css related to this. 
6. improved KvHero by removing the fixed height, now it can handle carousels, and image heros of varying heights. 
7. Added content slot to kv hero, this allows content to be overlayed on top of the image, like the form/cta text on this page. 
8. Added content to page (all static for now, will move to contentful in the next phase)
9. fixed bug for kvdropdown bleeding outside of page
10. I made some very slight visual modifications to the page look. (Made 100% guarantee icon a little bigger/sharper, some spacing issues with faqs on mobile, resolution of 10 billion image, form element layouts, error messages, etc) These are all very minor and I do not believe they will be an issue

**Page Status:** 

- The page has all the content.
- The form has validation
- Page will check if user is a monthly good subscriber and display alternate notice if they are
- Form does nothing on button click. 

Along the way I discovered several issues with the current monthlygood page and tried to not replicate those. :)

_* remember to view /monthlygood-ui you have to modify your .htaccess_


src/components/Kv/KvResponsiveImage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<img :src="imageUrl" v-bind="$attrs">
+	<img :src="imageUrl" v-show="isRetina != null">
```
ryan-ludwig: In order to pass alt text through, we should either still bind the $attrs, or make an alt prop on KvResponsiveImage and pass it down to the image.

src/components/Kv/KvResponsiveImage.vue
```diff
@@ -9,14 +9,16 @@ import { isHighDensity, isRetina } from '@/util/checkScreenDensity';
 import screenSizeMixin from '@/plugins/screen-size-mixin';
 
 export default {
+	// initializing isRetina as null and using the v-show above prevents flashing of a
+	// low res image before isRetina and screenWidth are set
```
ryan-ludwig: 👀😃

src/components/Kv/KvResponsiveImage.vue
eddieferrer: There is no need to do `v-bind="$attrs"` on root component elements, as Vue already does this automatically. `v-bind="$attrs"` should only be used when you want to bind the attribute to non root elements.

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+	text-shadow: 1px 1px 3px #333;
+	margin: 0;
+}
+
+.mg-headline {
+	font-weight: bold;
+	font-size: 2rem;
+	line-height: 2rem;
+
+	@media only screen and (min-width: 800px) {
```
ryan-ludwig: Possibly use the `@include breakpoint(xlarge)` mixin to keep a consistent set of breakpoints

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+	}
+}
+
+.mg-subhead {
+	padding: 0;
+	font-size: 1.2rem;
+	line-height: 1.35rem;
+	margin-bottom: 0.65rem;
+
+	@media only screen and (min-width: 800px) {
```
ryan-ludwig: Possibly use the `@include breakpoint(xlarge)` mixin to keep a consistent set of breakpoints

src/pages/Protocol.vue
```diff
@@ -116,7 +116,9 @@
 					<kv-custom-donation :default-value="100" />
 				</div>
 				<div class="show-for-large medium-8 medium-offset-2 large-5 large-offset-1 columns">
-					<img src="~@/assets/images/10-years-billion-impact.jpg">
+					<kv-responsive-image
+						:images="billionImpactImages"
```
ryan-ludwig: Needs alt text or an alt=""

src/components/Kv/KvResponsiveImage.vue
ryan-ludwig: ah right! thanks

src/pages/MonthlyGood/KivaAsExpert.vue
```diff
@@ -0,0 +1,84 @@
+<template>
+	<div class="kiva-as-expert-section-wrapper">
```
emuvente: I think this wrapper & background color should be moved to the monthly good page component, as it only matters in the context of that page.

src/pages/MonthlyGood/HowItWorks.vue
```diff
@@ -0,0 +1,103 @@
+<template>
+	<div class="how-works-section-wrapper">
```
emuvente: This wrapping div is unnecessary and the `.row` div can be used as the root element instead.

src/pages/MonthlyGood/FrequentlyAskedQuestions.vue
```diff
@@ -0,0 +1,84 @@
+<template>
+	<div class="frequently-asked-questions-section-wrapper">
```
emuvente: This wrapping div is unnecessary and the `.row` div can be used as the root element instead.

src/pages/MonthlyGood/EmailPreview.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+	<div class="email-preview-section-wrapper">
```
emuvente: This wrapping div is unnecessary and the `.row` div can be used as the root element instead.

src/pages/MonthlyGood/ExpandableQuestion.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+	<div class="small-12 columns expandable-faq">
+		<div class="row expandable-faq__header"
```
ryan-ludwig: This would be good as a <button> el to get keyboard focus.

src/pages/MonthlyGood/ExpandableQuestion.vue
```diff
@@ -0,0 +1,89 @@
+<template>
```
ryan-ludwig: What do you think about making this into a re-usable KvAccordian component?

src/pages/MonthlyGood/HowItWorks.vue
eddieferrer: This one is necessary because of the full width background color

src/pages/MonthlyGood/KivaAsExpert.vue
eddieferrer: For optimization reasons, the image on this component has a background color and the component is going to look strange if the component background does not match the image background.

src/pages/MonthlyGood/HowItWorks.vue
emuvente: You're right, sorry, I missed that one. Confused it with `.learn-more-wrapper`

src/pages/MonthlyGood/MoreAboutKiva.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+	<div class="more-about-kiva-section-wrapper">
```
emuvente: This one looks like it could be combined with the `.row` div

src/pages/MonthlyGood/MoreAboutKiva.vue
```diff
@@ -0,0 +1,61 @@
...
+		<div class="row">
+			<div class="small-12 columns">
+				<h2 class="impact-text">
+					More about Kiva
+				</h2>
+			</div>
+			<div class="small-12 large-7 columns">
+				<p>
```
emuvente: It might be possible to put the column classes on this paragraph, and the following paragraph, and then get rid of the wrapping div

src/pages/MonthlyGood/KivaAsExpert.vue
emuvente: 👍

src/pages/MonthlyGood/ExpandableQuestion.vue
```diff
@@ -0,0 +1,89 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import 'settings';
+
+.expandable-faq {
+	border-top: 1px solid $charcoal;
+
+	.toggle-arrow {
+		height: 16px;
```
BoulderBrains: height: 1rem;

src/pages/MonthlyGood/ExpandableQuestion.vue
```diff
@@ -0,0 +1,89 @@
...
+		width: 25px;
```
BoulderBrains: width: 1.563rem;

src/pages/MonthlyGood/KivaAsExpert.vue
```diff
@@ -0,0 +1,84 @@
...
+		color: $kiva-green;
+		margin-bottom: 1.85rem;
+	}
+
+	.community-image {
+		padding: 1.25rem;
+	}
+
+	.hide-for-large.community-image {
+		max-width: 400px;
```
BoulderBrains: max-width: 25rem;

src/pages/MonthlyGood/MoreAboutKiva.vue
eddieferrer: I can go either way on this. I think the template looks a little cleaner with the grid classes on the divs.

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+			border: 1px solid $charcoal;
+			background-color: rgba(255, 255, 255, 0.7);
+			padding: 0.15rem 0 0 0.45rem;
+			line-height: 1.25rem;
+			width: 100%;
+		}
+	}
+
+	//set min height to improve sizing when image has not loaded yet
+	min-height: 100px;
```
BoulderBrains: min-height: 6.25rem;

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+	@media only screen and (min-width: 761px) {
```
BoulderBrains: We should use the x-large break point here.

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+		min-height: 355px;
```
BoulderBrains: min-height: 22.19rem;

src/pages/MonthlyGood/MoreAboutKiva.vue
emuvente: It does look a little cleaner, but mobile performance is greatly dependent on the number of dom elements in the page, so we should take every chance we can to use as few elements as possible

src/pages/MonthlyGood/MonthlyGoodPage.vue
```diff
@@ -1,47 +1,199 @@
...
+	}
+	@media only screen and (min-width: 1025px) {
+		min-height: 440px;
```
BoulderBrains: min-height: 27.5rem;

src/pages/MonthlyGood/MoreAboutKiva.vue
```diff
@@ -0,0 +1,61 @@
...
+
+<style lang="scss" scoped>
+@import 'settings';
+
+.icon-kiva-promise {
+	fill: transparent;
+	color: $charcoal;
+	display: block;
+	margin: 0 auto 2rem;
+	max-height: 130px;
```
BoulderBrains: max-height:  8.125rem;

src/pages/MonthlyGood/MonthlyGoodPage.vue
emuvente: For these more complex ones you can also use rem-calc e.g. `rem-calc(355)`

eddieferrer:
> Added a couple comments about changing px definitions to rems, as that's our standard practice unless we need 1-3px for a border. Also called out a spot where we could used a defined breakpoint instead of a px defined px screen size. Other than that, looks good, thanks for building all this out @eddieferrer 👍

Thanks for doing all the px -> rem conversions :)

src/pages/MonthlyGood/ExpandableQuestion.vue
eddieferrer: Lets hold off on this for now

mcstover: Agree. we should build at some point. @ryan-ludwig could you please create a ticket in the vue queue for this new component... you could ref this file as an example of what should be supported.

PR #1435, opened 2020-01-30
CASH-1754 Add transaction time to receipt

ryan-ludwig:


src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -244,6 +245,9 @@ export default {
 		},
 	},
 	computed: {
+		formattedTransactionTime() {
+			return `${format(this.receipt.transactionTime, 'MMMM DD, YYYY h:mm A')} PDT`;
```
emuvente: `format` requires a `Date` object. You'll have to create that with `Date.parse(this.receipt.transactionTime)`. Also, you'll want to be sure that you're correctly accounting for the timezone (The formatted time may not be in PDT)

ryan-ludwig: This appears to be currently working because we're still on v1.x of the date-fns library which can take a String instead of a Date object. It's a good plan to convert it to a Date though in prep for v2.x of date-fns.

ryan-ludwig: Oh, and yep timezones are broken. Good catch!

PR #1436, opened 2020-01-30
Add transaction tracking after successful completion of the transaction.

mcstover:


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -428,15 +432,29 @@ export default {
...
+				}),
+				donations: _map(this.donations, donation => {
+					return _pick(donation, ['__typename', 'id', 'price']);
+				}),
+			};
+			// fire transaction events
+			this.$kvTrackTransaction(transactionData);
+			// redirect to thanks
+			window.setTimeout(
+				this.redirectToThanks(transactionId),
```
remoun: IIUC this calls redirectToThanks and passes its result to setTimeout, rather than enqueueing a call. Is that right?

mcstover: Good call @remoun Thanks!

PR #1439, opened 2020-01-31
If using the suggested message, update it when a new loan is selected

ryan-ludwig:


src/components/Checkout/SocialShare.vue
```diff
@@ -245,7 +245,16 @@ export default {
...
 		},
 		setSelectedLoanIndex(index) {
+			// are we currently using the suggested message?
+			const isUsingSuggestedMessage = this.isSuggestedMessage;
+
+			// update the selected loan
 			this.selectedLoanIndex = index;
+
+			// if we were using the suggested message for the previous loan, use the suggested message for the new one.
+			if (isUsingSuggestedMessage) {
```
eddieferrer: Do we need to create `isUsingSuggestedMessage` ? why dont you just check for this.isSuggestedMessage on this line?

PR #1445, opened 2020-02-03
VUE-210 Document KvLightbox

ryan-ludwig:


.storybook/stories/KvFormKitchenSink.stories.js
```diff
@@ -166,40 +166,6 @@ export const KitchenSink = () => ({
 
 				</fieldset>
 
-				<fieldset>
```
ryan-ludwig: Sneaking this in as a cleanup, since going forward we'll just being using `<kv-dropdown>` rather than `<select>`

PR #1446, opened 2020-02-04
Improve cypress mocking

emuvente:
This changes the serialization method for the mocks so that simple primitive values and variable values can be used, which makes mocking simple responses much easier. This also includes an example test involving a mutation, where the mock returns a different value based on the request arguments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
 		"report": "webpack-bundle-analyzer dist/client-bundle-stats.json",
 		"unit": "jest --config test/unit/jest.conf.js --coverage",
 		"test": "npm run lint && npm run unit && npm run build && npm run test:e2e",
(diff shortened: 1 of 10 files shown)


ryan-ludwig:
> Awesome work.
> 
> I am a little concerned about the new classes that you added that are solely being used for cypress tests. For the qa automation, we went with 'qa-*' classes...
> 
> Cypress recommends data-cy attributes:
> https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements
> 
> Should we take this opportunity to replace our cypress tests with data attributes while we still have a low number of tests?

 +1 on using data-attributes as a way to clearly separate these selectors from styling. Maybe one data attribute for both qa and cypress would good? data-test, data-qa, data-automation... something like that?

BoulderBrains:
🙋🏼‍♂️'data-test' feels right to me.

PR #1471, opened 2020-02-15
Show Autolend timing messaging even if cIdleStartTime is null.

mcstover:


mcstover:
Closing for now until further details are sorted out. I'll fix the test then too!

PR #1482, opened 2020-02-21
CASH-1860 header theming

ryan-ludwig:
* Allows[ theming of the header](https://github.com/kiva/ui/pull/1482/commits/cc8ca5bcbf7c5e81a6776f67d67a5f3e1e20a1a2) using any colors you like. Browsers who don't support CSS custom properties (IE11 and below) get the default colors
* Wraps much of the header contents in a <nav> element as an accessibility landmark
* Storybook it up

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -1,290 +1,288 @@
 <template>
 	<header class="top-nav" :style="cssVars">
-		<nav aria-label="Primary navigation">
-			<div class="header-row row">
-				<router-link class="header-logo header-button" to="/" v-kv-track-event="['TopNav','click-Logo']">
-					<kv-icon name="new-kiva-logo" />
-				</router-link>
-				<router-link
-					:id="lendMenuId"
-					to="/lend-by-category"
-					class="header-button"
-					v-kv-track-event="['TopNav','click-Lend']"
+		<div class="header-row row">
+			<router-link class="header-logo header-button" to="/" v-kv-track-event="['TopNav','click-Logo']">
+				<kv-icon name="new-kiva-logo" />
+			</router-link>
+			<router-link
+				:id="lendMenuId"
+				to="/lend-by-category"
+				class="header-button"
+				v-kv-track-event="['TopNav','click-Lend']"
+			>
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -1,275 +1,290 @@
...
-			<div class="flexible-center-area">
-				<div id="top-nav-search-area" v-if="!hideSearchInHeader" :aria-hidden="searchOpen ? 'false' : 'true'">
-					<button class="close-search hide-for-large"
-						:aria-expanded="searchOpen ? 'true' : 'false'"
-						:aria-pressed="searchOpen ? 'true' : 'false'"
-						aria-controls="top-nav-search-area"
-						@click="toggleSearch"
-						v-kv-track-event="['TopNav','click-search-close-mobile']"
+	<header class="top-nav" :style="cssVars">
+		<nav aria-label="Primary navigation">
```
ryan-ludwig: This diff of the template seems like a lot, but the change is wrapping everything in the header with a <nav> element

src/components/WwwFrame/TheHeader.vue
```diff
@@ -396,6 +425,11 @@ export default {
 			}
 		}
 	},
+	mounted() {
+		if ('requestIdleCallback' in window) {
+			requestIdleCallback(this.loadLendInfo); // preload the Lend dropdown if we're idle
```
ryan-ludwig: This prevents the loading spinner on the Lend dropdown if your browser supports requestIdleCallback (no Safari or IE11).

mcstover: This could be a nice enhancement for those staying within the Ui app (the data would be in the apollo cache) but will likely result in a pretty big request for unnecessary data coming into the page with every refresh.

I would prefer this feature be worked on in a separate PR.

Once it's separated out we'll also need to add a mechanism to ensure this only get's called 1 time. They use a task queue here (https://developer.mozilla.org/en-US/docs/Web/API/Background_Tasks_API) but that is prob overkill for this...

mcstover: Thank you let's pick this up at another time...and maybe do a little checking on the performance of this request.

PR #1491, opened 2020-02-24
Fix for Link in Global Promo Banner

eddieferrer:
* Made link in global promo banner not be required, if link is missing, banner will render correctly
* Also fixed banner icon shrinking

VUE-216


eddieferrer:
CSS changes besides 'flex-shrink: 0;' are just cleanup and organization, since some of those styled were copied from legacy.

PR #1492, opened 2020-02-24
update memcached urls for prod and qa

mcstover:


brainbugged:
@mcstover : dev.js needs two nodes too.

mcstover:
@brainbugged This branch won't deploy to dev and you've already added the 2 nodes for that into master so we're good.

PR #1495, opened 2020-02-25
CASH-1861 Themable footer

ryan-ludwig:


src/components/WwwFrame/TheFooter.vue
```diff
@@ -1,212 +1,214 @@
 <template>
-	<footer class="www-footer">
-		<nav class="small-footer hide-for-large">
-			<ul class="hide-for-print">
-				<li>
-					<router-link :to="applyUrl">
-						Borrow
+	<footer class="www-footer" :style="cssVars">
```
ryan-ludwig: The diff looks big, but it's just having one outer `nav` element instead of two and adding the cssVars

emuvente: Why 1 nav and 2 divs instead of just 2 navs?

ryan-ludwig: Initially I thought it would be reading as 2 navigation landmarks for screen readers, and since it's really just one footer nav area I thought there should only be one. 

Now I see that one of the navs would always be display:none based on screen width. So likely only one nav would be announced.

I could go either way!

emuvente: We should take every chance we can to use fewer dom nodes, since mobile performance is significantly impacted by the number of dom nodes, so I think we should use 2 nav elements here.

I'm pretty sure you're right that only one of them will be announced since one will always have `display: none`. That's what I recall happening when I tested it before.

PR #1496, opened 2020-02-25
CASH-1830 IWD Homepage Takeover

ryan-ludwig:
CASH-1862,
CASH-1863,
CASH-1864,
CASH-1865,
CASH-1866,
CASH-1867,
CASH-1879,
CASH-1880

TODO: 
- Probably break up some areas into separate components. 
- Add flourishes, flags, etc. 
- Get category urls and images. 
- Wrap up the hero image area, ensure IE11 is handled. 
- Probably turn all sponsor logos greyscale. 
- Performance testing, accessibility testing.
- Get feedback, rinse and repeat 🛀.
- Conditional text / CTAs depending on date? Or maybe contentful?


ryan-ludwig:
Depends on #1495

ryan-ludwig:
Still some work and updates to go, but the meat of it is here. Feedback welcome!

src/pages/Homepage/iwd/IWDHomepage.vue
```diff
@@ -0,0 +1,813 @@
+<template>
```
ryan-ludwig: Majority of work is in this file

src/components/Kv/KvResponsiveImage2.vue
```diff
@@ -0,0 +1,136 @@
+<template>
```
ryan-ludwig: Goal of this component is to use the same api as KvResponsiveImage but output a picture element with <source> and srcset tags in order to take advantage of the look-ahead browser rendering of images. 
With the exististing KvResponsiveImage, images fetching doesn't start until after domcontentloaded because we need to wait for JS to query the screenwidth and density before deciding setting the image src. 

Browsers that don't understand the picture element (IE11) use the same logic that we had before. Browsers that do understand picture get significantly earlier image downloads.

src/components/Kv/KvResponsiveImage2.vue
```diff
@@ -0,0 +1,136 @@
...
+export default {
+	props: {
+		// An array of image sizes and urls, e.g. [['small', 'img-sm.jpg'], ['small retina', 'img-sm-retina.jpg']]
+		images: {
+			type: Array,
+			required: true
+		}
+	},
+	data() {
+	// initializing isRetina as null and using the v-show above prevents flashing of a
```
ryan-ludwig: screen-size-mixin.js is merged into this component since that's the only place it's currently used.

src/router/routes.js
```diff
@@ -1,5 +1,5 @@
 export default [
-	{ path: '/', component: () => import('@/pages/Homepage/Homepage') },
+	{ path: '/', component: () => import('@/pages/Homepage/iwd/IWDHomepage') },
```
ryan-ludwig: @mcstover Will need a game plan for this route

mcstover: @ryan-ludwig I think this will be fine for demo and testing tomorrow. We'll likely want to move back to a single page level component for the Homepage, then within that component we'll use a setting to toggle between 2 content components and the theme props.

src/pages/Homepage/iwd/IWDHomepage.vue
mcstover: As mentioned below... you'll want to remove the www-page component here and just keep the slot content. We'll have to do the same for the current homepage moving it's slot-content into it's own file (let's call it defaultHomepage).

mcstover: Again all good for now.

src/components/Kv/KvResponsiveImage2.vue
BoulderBrains: Nice! Thanks for the explanation @ryan-ludwig

PR #1511, opened 2020-03-02
Use stage memcache cname

colin-at-kiva:
Updating the memcache location in stage


colin-at-kiva:
Stage only has 1 at the moment - the `stage-memcached-01` will resolve to `kivadev-oregon-memcached.knmtma.0001.usw2.cache.amazonaws.com` instead of the `cfg` endpoint currently being used

PR #1512, opened 2020-03-03
Cash 1883

BoulderBrains:
New text on /settings/portfolio when the user's daysIdle = 0;


src/pages/Autolending/LendTimingDropdown.vue
```diff
@@ -123,6 +128,11 @@ export default {
 				return `Since you haven’t made a loan yourself in ${daysIdle} days, you will be eligible for auto-lending on May 20, 2020 when the first 90-day auto-loans are made.`;
 			}
 
+			// cash-1883
+			if (daysIdle === 0) {
+				// eslint-disable-next-line max-len
+				return `You're currently active! You'll be eligible for auto-lending if you don't make a loan yourself in the next ${this.lendAfterDaysIdle} days.`;
+			}
```
mcstover: @BoulderBrains Seeing now that the copy is the same for all cases and the lendAfterDaysIdle will apply either way, you can move this block up to line 107 making it the first statement potentially returned. Then you can delete the duplicate block that runs for daysIdle === 0 within the 90 days block.

PR #1513, opened 2020-03-03
Contentful dynamic content test page - CASH-1819

eddieferrer:
* A possible way to do very dynamic content from contentful


src/pages/MonthlyGood/EmailPreviewContentful.vue
```diff
@@ -0,0 +1,80 @@
...
+	color: $kiva-green;
+	margin-bottom: 1.85rem;
+}
+
+.email-preview-text {
+	padding: 4rem 1rem;
+}
+
+img {
+	max-width: 350px;
```
BoulderBrains: tiny catch here, max-width: 21.88rem;

PR #1518, opened 2020-03-05
Update node version used to our new version 12 and use npm prune comm…

mcstover:
…and to remove dev dependencies but preserve prod dependencies

### DIFF
--- a/Dockerfile.jenkins
+++ b/Dockerfile.jenkins
@@ -1,16 +1,16 @@
 FROM alpine
 
-RUN apk add --no-cache nodejs=10.16.0-r0
+RUN apk add --no-cache nodejs=12.15.0-r0
 RUN apk add --no-cache npm
 RUN apk add --no-cache git
 RUN apk add --no-cache bash
 RUN mkdir /ui
 COPY ./ /ui/
 WORKDIR /ui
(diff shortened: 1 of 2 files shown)


Dockerfile.jenkins
```diff
@@ -1,16 +1,16 @@
...
 
-RUN apk add --no-cache nodejs=10.16.0-r0
+RUN apk add --no-cache nodejs=12.15.0-r0
 RUN apk add --no-cache npm
 RUN apk add --no-cache git
 RUN apk add --no-cache bash
 RUN mkdir /ui
 COPY ./ /ui/
 WORKDIR /ui
+# we should switch this to npm ci (need to figure out how to tell Jenkins to maintain some cache of modules first)
```
leoperry: For Gradle caches we mount caches in a running container, build the artifact, then extract and build the final image with a Dockerfile. That makes our final image extremely minimal in terms of layers. To do this directly in a Dockerfile, one idea could be to use a multistage strategy (where the earlier stage maintains a cache of dependencies and updates as necessary).

mcstover: Does that container get re-used for the next build? This comment is around maintaining some caching of node_modules for the next build coming up vs re-downloading all of them each time...

PR #1522, opened 2020-03-06
Update Dockerfile.jenkins for build to K8. Add .dockerignore to preve…

mcstover:
…nt local node_modules from being copied to the image.

### DIFF
--- a/Dockerfile.jenkins
+++ b/Dockerfile.jenkins
@@ -34,4 +34,4 @@ RUN npm install --production
 EXPOSE 8888
 
 # Launch application
-CMD [ "npm", "start", "--", "--port=8888", "--config=local", "--prefix", "/ui" ]
+CMD [ "npm", "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui" ]
--- a/Dockerfile.jenkins
+++ b/Dockerfile.jenkins
@@ -1,16 +1,33 @@
-# includes nodejs + npm
-FROM node:12-alpine
-
+# build stage
+FROM node:12-alpine as build-stage
 RUN apk add --no-cache git
 RUN apk add --no-cache bash
+
 RUN mkdir /ui
 COPY ./ /ui/
 WORKDIR /ui
-# we should switch this to npm ci (need to figure out how to tell Jenkins to maintain some cache of modules first)
+
 RUN npm i --no-save
 RUN npm run build
-RUN npm prune --production
-WORKDIR /ui/ui-build
-RUN npm install cross-env
-CMD [ "/usr/bin/npm", "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui/ui-build" ]
 
+# run stage
+FROM node:12-alpine as run-stage
+# Change working directory
+RUN mkdir /ui
+WORKDIR /ui
+# Bring over app files
+COPY --from=build-stage /ui/build ./build
+COPY --from=build-stage /ui/config ./config
+COPY --from=build-stage /ui/deploy ./deploy
+COPY --from=build-stage /ui/dist ./dist
+COPY --from=build-stage /ui/server ./server
+COPY --from=build-stage /ui/src ./src
+COPY --from=build-stage /ui/package.json .
+COPY --from=build-stage /ui/package-lock.json .
+# Install PROD dependencies only
+RUN npm install --production
+# Expose API port to the outside
+EXPOSE 8888
+# Launch application
+# CMD ["npm","run","start","--","--config=local"]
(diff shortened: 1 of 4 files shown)


Dockerfile.jenkins
```diff
@@ -1,16 +1,37 @@
...
+FROM node:12-alpine as run-stage
+
+# Change working directory
+RUN mkdir /ui
+WORKDIR /ui
+
+# Bring over app files
+COPY --from=build-stage /ui/build ./build
+COPY --from=build-stage /ui/config ./config
+COPY --from=build-stage /ui/deploy ./deploy
```
JoeTravisKiva: I don't think you should need to include the deploy directory in the runtime image

Dockerfile.jenkins
```diff
@@ -1,16 +1,37 @@
...
 
+# run stage
+FROM node:12-alpine as run-stage
+
+# Change working directory
+RUN mkdir /ui
+WORKDIR /ui
+
+# Bring over app files
+COPY --from=build-stage /ui/build ./build
```
JoeTravisKiva: Is the build directory necessary for this stage?

mcstover: Haven't fully verified but I think we may reference. Luckily not much for combined file size here so we can dig deeper later.

deploy/kiva-ui.yml
```diff
@@ -18,7 +18,7 @@ spec:
       - name: kiva-ui
         image: kiva/ui:master
         command: [ "npm" ]
-        args: [ "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui/ui-build" ]
+        args: [ "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui" ]
```
leoperry: I'd recommend we delete `kiva-ui.yml` as it's not actually getting deployed by Helm (it lives outside the `charts` directory and otherwise isn't referenced anywhere I can see). If we want to do local k8s deployments we should set up our chart + values file to handle that scenario for us.

deploy/charts/templates/deployment.yaml
```diff
@@ -29,8 +29,8 @@ spec:
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}" # In Jenkins, need to call helm with --set image.tag=master-XY
-        command: ["/usr/bin/npm"]
-        args: ["start", "--", "--port=8888", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui/ui-build"]
+        command: ["npm"]
+        args: ["start", "--", "--port=8888", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui"]
```
leoperry: Likewise i recommend a little streamlining here. We should decided whether we want this to live in the deployment spec or Dockerfile -- but I'd strongly recommend against having this in both places.

The Dockerfile is OK if we don't need anything dynamic or can rely on environment variables.

If we have a more complex dynamic scenario then doing this via Helm + values in the deployment spec might be easier.

Dockerfile.jenkins
```diff
@@ -1,16 +1,36 @@
...
+COPY --from=build-stage /ui/config ./config
+COPY --from=build-stage /ui/dist ./dist
+COPY --from=build-stage /ui/server ./server
+COPY --from=build-stage /ui/src ./src
+COPY --from=build-stage /ui/package.json .
+COPY --from=build-stage /ui/package-lock.json .
+
+# Install PROD dependencies only
+RUN npm install --production
```
leoperry: So I think this can work for now, but if we want to slim down the final image, we should really examine what is needed at container runtime for the app. So *after* npm install, what is really needed to run the app? And then that set of assets is what we'd want to copy in.

The answer may just be "everything copied over above"... in which case, there may not be a lot of savings here.

However, we may still be able to optimize builds a bit by reordering. So e.g. if we could run npm install first, then copy over source files and run npm build (might not work that way... I fully admit I don't know anything about the UI app build process), then we might be able to leverage local docker image layer caching to speed up builds.

All this is just contextualizing some of the conversation we've had elsewhere closer to the Dockerfile, and I think could just be things to consider in future revs rather than needing to be addressed right now.

deploy/charts/templates/deployment.yaml
mcstover: Agree, I'm totally fine with moving the command from here if it gets handled by the other dynamic config...

Dockerfile.jenkins
mcstover: Also agree. Just whittled these down with a rough pass. We may be able to drop the `src` folder but I'll to run some local tests with the app to determine exactly what is needed for a deployed instance.

PR #1523, opened 2020-03-06
Ingress config

JoeTravisKiva:
- Consolidating port configurations to a single value
- remove unnecessary port 443 inclusion
- reduce duplication in ingress configuration, so common values are in main chart, overrides at environment level

### DIFF
--- a/deploy/charts/templates/deployment.yaml
+++ b/deploy/charts/templates/deployment.yaml
@@ -30,11 +30,11 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}" # In Jenkins, need to call helm with --set image.tag=master-XY
         command: ["/usr/bin/npm"]
-        args: ["start", "--", "--port=8888", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui/ui-build"]
+        args: ["start", "--", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui/ui-build"]
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         ports:
          - name: {{ .Values.service.portName }}
-           containerPort: {{ .Values.service.containerPort }}
+           containerPort: {{ .Values.service.port }}
 #       env:
 #         - name: foo
 #           valueFrom:
--- a/deploy/charts/templates/service.yaml
+++ b/deploy/charts/templates/service.yaml
@@ -12,7 +12,6 @@ spec:
   type: {{ .Values.service.type }}
   ports:
   - port: {{ .Values.service.port }}
-    targetPort: {{ .Values.service.targetPort }}
     name: {{ .Values.service.portName }}
   selector:
     app.kubernetes.io/name: {{ include "ui-helm.name" . }}
--- a/deploy/charts/templates/service.yaml
+++ b/deploy/charts/templates/service.yaml
@@ -12,6 +12,7 @@ spec:
   type: {{ .Values.service.type }}
   ports:
   - port: {{ .Values.service.port }}
+  	targetPort: {{ .Values.service.portName }}
     name: {{ .Values.service.portName }}
   selector:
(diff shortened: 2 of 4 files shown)


deploy/charts/templates/service.yaml
```diff
@@ -12,7 +12,6 @@ spec:
   type: {{ .Values.service.type }}
   ports:
   - port: {{ .Values.service.port }}
-    targetPort: {{ .Values.service.targetPort }}
```
leoperry: I agree with removing the `service.targetPort` value, but I think we should still specify `targetPort` in service specs as otherwise we're relying on the Kubernetes default (which defaults to be equal to `port`. I prefer to use port aliases for this. So, because we named the port to be `service.portName` in the deployment spec, I would recommend specifying the `targetPort` to also be `service.portName` in the service spec.

deploy/charts/values.yaml
```diff
@@ -15,35 +15,11 @@ replicaCount: 2
...
-
-
 ingress:
   enabled: true
   name: "kiva-ui-ingress"
   annotations:
-    certmanager.k8s.io/acme-challenge-type: dns01
-    certmanager.k8s.io/acme-dns01-provider: route53
-    certmanager.k8s.io/cluster-issuer: letsencrypt-prod
+    cert-manager.io/cluster-issuer: letsencrypt-prod
```
leoperry: 👍 These ingress annotations are looking right now.

deploy/dev/values.yaml
```diff
@@ -37,7 +27,7 @@ ingress:
           paths:
           - backend:
               serviceName: "kiva-ui"
-              servicePort: 443
+              servicePort: *servicePort
```
leoperry: Likewise here I prefer to use port aliases, so if we named the service port as `service.portName` we could likewise use that alias here.

deploy/charts/templates/deployment.yaml
```diff
@@ -30,11 +30,11 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}" # In Jenkins, need to call helm with --set image.tag=master-XY
         command: ["/usr/bin/npm"]
-        args: ["start", "--", "--port=8888", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui/ui-build"]
+        args: ["start", "--", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui/ui-build"]
```
leoperry: One of my concerns with all this (and I'm fine rolling it out and seeing what happens at this stage) is whether the UI server is actually serving https. Our Kubernetes ingress model terminates ssl at the ingress controller, and cluster-internal traffic can be unencrypted. Thus if UI is trying to serve https itself, I think with the current configuration we're going to have some ssl issues. Might be nothing more than a certificate warning right now, and, as I mentioned, we can roll with this and see what happens. @mcstover Any thoughts on UI serving https by default?

JoeTravisKiva: Is the UI trying to serve https?  My assumption was that it was not.

deploy/charts/templates/service.yaml
JoeTravisKiva: sure. can do

deploy/charts/templates/deployment.yaml
leoperry: I am basing this on vague recollection, but I think Matt will know immediately. In any event, the concern came up mainly while thinking about these ports, but it's not really germane to this PR. I would recommend we merge this in and continue revving on the build / deploy.

mcstover: All of the SSL resolution must be handled before entry into app.

leoperry: Ok, great. This is exactly what we want.

leoperry:
@mcstover @JoeTravisKiva Taking the liberty to merge this in to get the combined changes from this and #1522 deployed. We can take stock of where we are and move forward from there.

PR #1529, opened 2020-03-09
fix researcher's name

go-bears:
make typo fix change


go-bears:
thanks for extra eyes!

PR #1532, opened 2020-03-10
CASH-1807 add recommended loans row

emuvente:


emuvente:
Ran into complications with mocking, so there isn't any test. I also ran into issues with the basket id when trying to use the federation endpoint from my vm, so merging this to dev will allow us to actually test things out

go-bears:
🙌  🎉 ,  Woo!

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -459,23 +522,45 @@ export default {
...
+						query: loanChannelQuery,
+						variables: {
+							// exclude custom rows + limit for ssr
+							ids: _take(_without(ids, ...customCategoryIds), ssrRowLimiter),
+							imgDefaultSize,
+							imgRetinaSize,
+							// @todo variables for fetching data for custom channels
+						},
+					}),
+					recommendedRow ? client.query({
```
mcstover: fun!

src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -307,7 +308,7 @@ export default {
 			handler(channel) {
 				this.name = _get(channel, 'name');
 				this.url = _get(channel, 'url');
-				this.loans = _get(channel, 'loans.values');
+				this.loans = _filter(_get(channel, 'loans.values'));
```
ryan-ludwig: what does this _filter do?

emuvente: Handles cases where `loans` is something like `[null, null]` (I only encountered this in testing)

PR #1533, opened 2020-03-10
Use new kv-responsive-image with picture element throughout the site

ryan-ludwig:
* Moves functionality from kv-responsive-image2 into kv-responsive-image.
* Tested in IE11, Safari, Firefox, Chrome
* Touchups to homepage category grid

VUE-226


src/components/Homepage/CategoryGrid.vue
```diff
@@ -77,7 +80,7 @@ import getCacheKey from '@/util/getCacheKey';
 
 export default {
 	name: 'CategoryGrid',
-	serverCacheKey: () => getCacheKey('CategoryGrid'),
+	serverCacheKey: () => getCacheKey('CategoryGrid_2020-03-10'),
```
ryan-ludwig: Is there a convention for changing the serverCacheKey?

emuvente: You don't need to do this because you're changing the component entirely, not setting something up where it changes from props. The getCacheKey method appends the current git commit hash to the cache key, so you don't need to ever worry about previous versions of the component being used.

ryan-ludwig: @emuvente cool, thank you!

eddieferrer:
> Interested to see this in action. I think we need to develop some ability within this component to set a default height rather than using auto. See current the MG landing page at a small or tablet size.
@ryan-ludwig 
Agreed. See the MG Landing page "Monthly Good Delivered Image"  @ about 1015 px wide. It is horribly pixelated. 

Can we just remove the height/width styles? 

Also see: src/pages/MonthlyGood/EmailPreviewContentful.vue and the css I am using there for the img.

ryan-ludwig:
Following up on this, figured out the issue on Monthly Good Delivered page [here](https://github.com/kiva/ui/pull/1534). Since img is no longer the first element of kv-responsive-image the CSS wasn't targeting it anymore. Does not appear to happen elsewhere.

PR #1537, opened 2020-03-11
Inline the Kiva logo on the header

ryan-ludwig:
Ensures the Kiva logo is visible immediately instead of waiting for the icon sprite to be ajaxed in


test/unit/jest.conf.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
 		'^.+\\.(gql|graphql)$': 'jest-transform-graphql',
 		'^.+\\.vue$': 'vue-jest',
 		'^.+\\.js$': 'babel-jest',
+		'^.+\\.svg$': '<rootDir>/test/unit/specs/svgTransform.js',
```
ryan-ludwig: @emuvente Needed to add in this transform [to make Jest happy](https://vue-svg-loader.js.org/faq.html#how-to-use-this-loader-with-jest ). Wasn't sure where to put this svgTransform.js file though. Thoughts?

emuvente: I think making a new folder `/test/unit/transforms/` and putting it there would make sense

src/components/WwwFrame/TheHeader.vue
```diff
@@ -3,7 +3,7 @@
 		<nav aria-label="Primary navigation">
 			<div class="header-row row">
 				<router-link class="header-logo header-button" to="/" v-kv-track-event="['TopNav','click-Logo']">
-					<kv-icon name="new-kiva-logo" />
```
emuvente: I don't see `new-kiva-logo` being used anywhere else besides styleguide, so let's go ahead and remove that icon file too so that it doesn't get bundled into the icon sprite svg anymore

ryan-ludwig: 👍 Do you know if the old stack will still use it in some way?

emuvente: The old stack uses a different sprite bundle, so it won't be affected. We could consider inlining the svg for the old stack too, actually.

mcstover:
Thanks for working this out guys! 👍

PR #1538, opened 2020-03-12
Contentful Schema Change

eddieferrer:
* Changes contentful local resolver schema to match federation contentful
* Changes response handling to match federation responses

VUE-227

### DIFF
--- a/src/pages/PageTwo.vue
+++ b/src/pages/PageTwo.vue
@@ -1,5 +1,5 @@
 <template>
-	<!-- This is a sample page which recreates the monthly good landing page but with content from contentfulCMS -->
+	<!-- This is a sample page which recreates the monthly good landing page but with content from contentful -->
 	<www-page>
 		<kv-hero class="mg-hero" :class="{'experiment':isExperimentActive}">
 			<template v-slot:images>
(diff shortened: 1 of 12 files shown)


src/pages/PageTwo.vue
```diff
@@ -92,8 +92,8 @@ import MoreAboutKiva from '@/pages/MonthlyGood/MoreAboutKiva';
 import KivaAsExpertContentful from '@/pages/MonthlyGood/KivaAsExpertContentful';
 import FrequentlyAskedQuestions from '@/pages/MonthlyGood/FrequentlyAskedQuestions';
 
-const pageQuery = gql`
-	query ($contentType: String!, $contentKey: String) {
+// ! TODO this query is not working
```
eddieferrer: @emuvente 

This is the query that does not work. It seems that I can't combine `my` and `contentful` in the same query. The result is a `Network error: Cannot read property 'my' of undefined` Which make no sense to me. This error is the same if this query is executed on prefetch/in methods, server side, client side, etc.

mcstover: @eddieferrer Is this error new or was it failing prior to these changes?

emuvente: When I try to load this page locally, I'm getting the following error for that query, as well as no data returned:
```
Variable "$contentType" is never used
```
That error is happening because Apollo processes the @client query using the local resolvers and then sends the rest of the query to the api, meaning this is what the api sees:
```
query ($contentType: String!, $contentKey: String) {
  my {
    autoDeposit {
      isSubscriber
      __typename
    }
    __typename
  }
  general {
    uiExperimentSetting(key: "mg_hero") {
      key
      value
      __typename
    }
    __typename
  }
}
```
Since only the error and no data is returned for that query, any attempts to read `my` will fail because `data` is `null`. So I think we just aren't able to mix local queries and api queries when using variables in the local queries.

eddieferrer: thanks @emuvente

eddieferrer:
I gave up on trying to make the contentful query with the new schema work in preFetch. Will revisit later, but we might not even need that, since we'll be dumping the localResolver in the future. I imagine once we do that it will be a minor change to get it working in the UI

PR #1553, opened 2020-03-18
Push loan info into pageview analytics object. This will reflect the …

mcstover:
…loans presented for the recommended loans row.

### DIFF
--- a/src/pages/Lend/LendByCategoryPage.vue
+++ b/src/pages/Lend/LendByCategoryPage.vue
@@ -246,7 +246,8 @@ export default {
 			// track recommended loans row
 			if (this.showRecommendedLoansRow) {
 				if (this.recommendedLoans !== undefined && this.recommendedLoans.length > 0) {
-					_each(this.recommendedLoans, (loan, loanIndex) => {
+					const recommendedLoanValues = _get(this.recommendedLoans, 'values');
+					_each(recommendedLoanValues, (loan, loanIndex) => {
 						loanIds.push({
 							r: 95,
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -242,6 +243,20 @@ export default {
 				}
 			}
 
+			// track recommended loans row
+			if (this.showRecommendedLoansRow) {
+				if (this.recommendedLoans !== undefined && this.recommendedLoans.length > 0) {
+					_each(this.recommendedLoans, (loan, loanIndex) => {
+						loanIds.push({
+							r: 95,
```
rschoenbeck: If we can't get the exact row position I would recommend putting in impossible value here, such as the #2 that is suggested -- we should leave 95 just for the loan channel.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -431,11 +446,14 @@ export default {
...
 					});
-					const loans = _get(data, 'ml.recommendationChannel.loans');
-					if (loans && loans.values && loans.values.length > 0) {
+					// share out results of loan query for row analytics
+					this.recommendedLoans = _get(data, 'ml.recommendationChannel.loans');
+					if (this.recommendedLoans
+						&& this.recommendedLoans.values
+						&& this.recommendedLoans.values.length > 0) {
 						const channel = {
 							id: -2,
```
rschoenbeck: Should this match the channel # ("c" above) and be 95?

PR #1554, opened 2020-03-18
Fixes window undefined in server side catch for redirect

eddieferrer:
* /funded/ page will sometimes error, this changes the redirect to use the router instead of window.location

VUE-174


eddieferrer:
Even though ``/lend/${loanIdFromRoute}?minimal=false` ` is not a route in the UI, pushing to the router still works, tested in my VM

PR #1555, opened 2020-03-18
VUE-223 KvIcon: dynamically inline most icons, sprite common icons

ryan-ludwig:
Follow up from https://github.com/kiva/ui/pull/1541

This PR moves the majority of icon files out of the global icon sprite and into a folder where we dynamically import them only in the components which use them. This dramatically reduces the DOM nodes on all pages from ~465 to ~25 from the sprite.

- Icons which are used on most pages are kept in the sprite.
- Icons which I couldn't find a reference to in the UI repo have been removed.
- Some folder reorganization
- While testing icons if I noticed a <span or similar with an @click event on it I changed it to a <button for accessibility purposes
- After struggling with several icons not working correctly with webpack-svgstore-plugin which makes the sprite I ended up forking it and updating the SVGO dependency. That seems to have solved the issues. I have a PR up for them, hopefully it'll be resolved soon.

To test:
pull down, 
run npm install,
browse to some pages you've worked on, make sure icons look correct.
npm run storybook to see the available icons


.storybook/stories/GenericPromoBanner.stories.js
```diff
@@ -70,7 +26,7 @@ export const Default = () => ({
 	props: {
 		iconKey: {
 			type: String,
-			default: select('Icon Key', iconKeySelect, 'present')
+			default: 'present'
```
ryan-ludwig: Sorry @eddieferrer, since these icons are imported directly now I couldn't get them to swap out dynamically. I added a story to loop through all the available contentful icons instead.

package.json
```diff
@@ -180,7 +180,7 @@
 		"webpack-filter-warnings-plugin": "^1.2.1",
 		"webpack-hot-middleware": "^2.25.0",
 		"webpack-node-externals": "^1.7.2",
-		"webpack-svgstore-plugin": "^4.1.0"
+		"webpack-svgstore-plugin": "ryan-ludwig/webpack-svgstore-plugin#develop"
```
ryan-ludwig: This is unfortunate, but the only way I could get the sprite plugin to work with our magnifying glass icon was to fork it and update it's SVGO dependency. I have a PR up with them. 

I believe @BoulderBrains has had problems adding SVGs to the sprite in the past, I believe this is the same issue.

eddieferrer: Can you link to that PR?

eddieferrer:
Lets also please make sure to squash these commits when we merge this in. Thats a lot of commits...

package.json
ryan-ludwig: https://github.com/mrsum/webpack-svgstore-plugin/pull/180

ryan-ludwig:
> Lets also please make sure to squash these commits when we merge this in. Thats a lot of commits...

Good call, I'll rebase after reviews.

package.json
eddieferrer: I think webpack-svgstore-plugin is practically deprecated. We should try to replace it with something else

.storybook/stories/GenericPromoBanner.stories.js
eddieferrer: Unfortunate, but should be fine. 👍

package.json
ryan-ludwig: Yeah, I've wondered if we shouldn't just do the sprite manually since it has so few items now.

mcstover:
@ryan-ludwig @emuvente Please hold on merging this until next week.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -736,7 +749,7 @@ $close-search-button-size: 2.5rem;
 	.icon {
 		width: $top-nav-font-size;
 		height: $header-height;
-		stroke: $kiva-darkgreen;
+		fill: $kiva-darkgreen;
```
emuvente: I think the 'x' icon that this is for uses stroke not fill

ryan-ludwig: Oh, I converted all strokes to fills

emuvente: Could you expand on what you did for that?

ryan-ludwig: Mainly to make them consistent. You don't need to know whether an icon is built with a stroke or a fill, it's always just a fill. 
The other reason is that I don't see other icon packages using stroke.

mcstover: Def a nice simplification... @ryan-ludwig I think you updated the svg code right? or maybe regenerated after changing in Sketch...

ryan-ludwig: @matt, yes outlined the strokes in illustrator and re-exported/Svgo'd

emuvente: ☝️ That's what I was wondering, thank you

package.json
ryan-ludwig: I have my doubts that webpack-svgstore-plugin is going to merge my PR. Don't really love that this is pointing to my personal github for something we're wanting to get into prod.

2 options I think:
1) Fork webpack-svgstore-plugin to the kiva github, make the same change, point package.json to that.  
2) Ditch webpack-svgstore-plugin, roll the sprite by hand (easy), and ajax it in from TheHeader.vue or something similar.

Both have upsides and downsides. @emuvente @mcstover thoughts?

mcstover: I'm ok with #1.

Alternatively, if we take the #2 approach could we just import and render the image in App.vue to ensure it's always present?

ryan-ludwig: Going to run with #1 for today to get this big PR into dev.

With #2, yeah I think putting the sprite directly into App.vue makes sense to me 👍.

PR #1560, opened 2020-03-20
Updated date-fns library to latest version

eddieferrer:
* Updated all uses of date-library to fix breaking changes in move from v1 to v2

VUE-204


eddieferrer:
These are the breaking changes:

https://github.com/date-fns/date-fns/blob/master/CHANGELOG.md#200---2019-08-20

Notes: 

- parseISO is the date-fns recommended way to convert strings to dates.
- All date-fns functions require dates and not strings. 
- This is a pretty high risk PR, it touches a lot of areas. 
- I'm 95% confident `src/components/LoanCards/GridMicroLoanCard.vue` was not being used, hence the deletion

PR #1563, opened 2020-03-25
Dsci 447 funded loan filtering

mcstover:


mcstover:
FYI @rschoenbeck @ferrous777

mcstover:
We should be able to test this in Dev by funding one of the loans that show in the recommended loan list and ensuring it is filtered out on the next reload. To take it further we can compare the query in graphiql to see which loans are funded or not.

src/plugins/funded-loans-mixin.js
```diff
@@ -0,0 +1,39 @@
+import _filter from 'lodash/filter';
+import _get from 'lodash/get';
+import numeral from 'numeral';
+
+export default {
```
emuvente: I think this would make more sense as just a util file that exports 2 functions rather than as a mixin, since there's nothing vue related in this code, but this is fine for getting filtering working.

PR #1565, opened 2020-03-26
Implements Contenful Service from GQL Federation

eddieferrer:
* In all environments, except VM use GQL Federation for contentful
* In the VM use a local resolver which accesses Federation with apollo-fetch. This allows the VM to get contentful from `marketplace-api.dk1.kiva.org/graphql` and everything else from `api-vm.kiva.org/graphql` 

VUE-248


eddieferrer:
WIP don't review yet

PR #1571, opened 2020-03-31
add Cynthia Cryder paper

go-bears:
updated Research Page with new link


go-bears:
thanks @BoulderBrains !

PR #1574, opened 2020-04-01
Feature grow 57 app icons in footer

mlibazisi:
Ticket: https://kiva.atlassian.net/browse/GROW-57

**Desktop Before:**
desk-top-before

**Desktop After:**
desk-top-after

**Mobile Before:**
mobile-before

**Mobile After:**
mobile-after


src/components/WwwFrame/TheFooter.vue
```diff
@@ -36,6 +36,18 @@
 					Terms of use
 				</router-link>
 			</div>
+			<ul class="hide-for-print download-app-icons">
+				<li class="app-store-icon">
+					<a :href="appStoreUrl" target="_blank">
+						<img src="@/assets/images/app-store.svg">
```
ryan-ludwig, line 42: It would be good to add alt text to these images since there's no text inside the link, maybe `alt="Download Kiva's App on the App Store"`

Also, since these are on the footer we can add [`loading="lazy"` ](loading="lazy" )attribute to prevent them from being downloaded on Chrome unless they're in view.

src/components/WwwFrame/TheFooter.vue
```diff
@@ -36,6 +36,18 @@
 					Terms of use
 				</router-link>
 			</div>
+			<ul class="hide-for-print download-app-icons">
+				<li class="app-store-icon">
+					<a :href="appStoreUrl" target="_blank">
```
mcstover, line 41: @mlibazisi please follow up with @brandonatkiva to see if he wants tracking params on these links...

prob something like:
`v-kv-track-event="['Footer', App-Store-Link', 'iOS']"`

src/components/WwwFrame/TheFooter.vue
```diff
@@ -316,6 +344,20 @@ $footer-separator-color: $light-green;
...
+		display: inline-block;
+		height: 60px;
+	}
+
+	ul.download-app-icons img {
+		height: 100%;
+	}
+
+	ul.download-app-icons li.app-store-icon {
+		padding: 10px;
```
ryan-ludwig, line 357: In general the UI repo does most things in rem units (aside from percents). There's a handy sass function, rem-calc, that will convert px to rem for you though.
For this it would be: 
`padding: rem-calc(10);`

src/components/WwwFrame/TheFooter.vue
ryan-ludwig, line 41: Pretty minor, but [target="_blank" links ](https://developers.google.com/web/tools/lighthouse/audits/noopener) are best with a `rel="noopener"` attribute.

src/components/WwwFrame/TheFooter.vue
mlibazisi: Thanks for the headsup @ryan-ludwig

PR #1580, opened 2020-04-03
Only show app banners on certain routes

ryan-ludwig:
Also continue loading webmanifest on all pages, but move app banner logic into TheAppInstallBanner.vue.

GROW58, GROW-59

### DIFF
--- a/src/components/WwwFrame/TheAppInstallBanner.vue
+++ b/src/components/WwwFrame/TheAppInstallBanner.vue
@@ -8,6 +8,47 @@
 
 <script>
 export default {
+	metaInfo() {
+		return {
+			meta: this.isAppInstallPage ? [
+				// Apple specific meta tag to show native app install banner
+				// This is all that we need to do for iOS
+				{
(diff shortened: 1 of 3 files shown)


src/components/WwwFrame/TheAppInstallBanner.vue
```diff
@@ -8,6 +8,47 @@
...
+				}
+			] : []
+		};
+	},
+	computed: {
+		isAppInstallPage() {
+			const route = this.$route;
+			const baseUrlsToInclude = [
+				'start',
+				'portfolio',
```
ryan-ludwig, line 28: Realizing that several of these urls (and their children) are in the old stack.

mcstover: yes and login + register in a different repo as well. I know this component will have to be replicated somewhat in the old stack but may or may not be reusable in the Auth0 repo. Not sure if we're using vue-meta there.

PR #1583, opened 2020-04-03
Check for fields in the Contentful response before using them.

mcstover:


mcstover:
@eddieferrer This is actually an issue with no content being returned which I'm working through in the VM now.

PR #1591, opened 2020-04-07
Include wrapper in size declaration for icon on Loan card drawer

mcstover:


mcstover:
@ryan-ludwig I kinda feel like that injected `wrapper` class should be `icon-wrapper` instead...

ryan-ludwig:
> @ryan-ludwig I kinda feel like that injected `wrapper` class should be `icon-wrapper` instead...

Ahh, I don't disagree, however KvIcon's styles are now scoped, so we shouldn't run into any collisions.

For this PR, I think you can add just add class "icon" to kv-icon on line 12 and leave the css as is. I must have lost that class during the conversion.

mcstover:
@ryan-ludwig  adding `icon` alongside `wrapper` leaves the width on the direct icon unset and applies styles meant just for the svg element to the wrapper. This results in a variety of different broken positioning problems.

PR #1592, opened 2020-04-07
Fix giant icon on mobile info-panel 

ryan-ludwig:


ryan-ludwig:
@mcstover Sorry for missing this the first time around. I assume scoped styles would prevent problems, but the outer div can still bite us I guess?

PR #1593, opened 2020-04-08
Contentful powered lightbox for thanks page

eddieferrer:
VUE-258

This is a contentful lightbox wrapper which takes content from contentful, and puts it in a kv-lightbox component.

Does not yet support embedding of videos in the lightbox. 

Sample lightbox is on contentful dev.

### DIFF
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -103,6 +103,9 @@ export default {
 			}
 		}
 	},
+	mounted() {
+		this.isShown = this.visible;
+	},
 	methods: {
 		closeLightbox() {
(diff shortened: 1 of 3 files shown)


src/components/Kv/KvLightbox.vue
```diff
@@ -103,6 +103,9 @@ export default {
 			}
 		}
 	},
+	mounted() {
```
eddieferrer, line 106: This just makes it so the lightbox can be immediately activated by passing true to the prop, since watch won't activate if initial value is true

mcstover:
I think this is a good simple start. Still versatile and doesn't tie use to content ordering!

PR #1596, opened 2020-04-09
Apollo contentful refactor

eddieferrer:
- Creates a new apollo client for contentful requests 
- Client is provided to components through 'contentful' provider
- Client has a different uri only in the dev-vm, based on dev-vm config. 
- Client shares cache with other apollo client. 
- Removed contentful local resolver, contentful from local schema, GraphQL JSONObject package

This setup allows contentful requests to hit federation server from the dev-vm, it also allows them to be prefetched and put in the cache on the server. 

This code:
```	
      mounted() {
		this.contentful.query({
			query: contentful,
			variables: {
				contentType: 'uiSetting',
				contentKey: 'ui-global-promo',
			}
		}).then(({ data }) => {
			// process the contentful response here
		});
	},
	serverPrefetch() {
		return this.contentful.query({
			query: contentful,
			variables: {
				contentType: 'uiSetting',
				contentKey: 'ui-global-promo',
			}
		});
	}
```

Works for getting the content server side and reading it from the cache on the client. It is repetitive, and I only implemented it in `src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue`

In a follow up PR, I plan to extract this code out into a plugin, similar to `apollo-plugin` that would allow us to do this with less code repetition, by adding a $options.contentful object to components


eddieferrer:
> 👍 nice work! I can see this kind of pattern being useful for other federation api requests. Maybe we should consider calling it a 'federation' client rather than just a 'contentful' client

Good idea 👍

config/local.js
```diff
@@ -9,6 +9,9 @@ module.exports = merge(base, {
 		auth0: {
 			enable: false,
 		},
+		contentfulService: {
+			uri: 'https://api.kivaws.org/graphql',
```
mcstover, line 13: `https://marketplace-api.k1.kiva.org/graphql`

config/local.vm.js
```diff
@@ -8,6 +8,9 @@ module.exports = merge(base, {
 		auth0: {
 			enable: false,
 		},
+		contentfulService: {
+			uri: 'https://api.kivaws.org/graphql',
```
mcstover, line 12: `https://marketplace-api.k1.kiva.org/graphql`

config/local.js
eddieferrer: 👍

PR #1599, opened 2020-04-12
Variation on our KvPillToggle that inclues an in place custom input f…

mcstover:
…orm if one of passed in options has a key of 'custom'.

### DIFF
--- a/.storybook/stories/MultiAmountSelector.stories.js
+++ b/.storybook/stories/MultiAmountSelector.stories.js
@@ -1,0 +1,70 @@
+import MultiAmountSelector from '@/components/Forms/MultiAmountSelector';
+
+export default {
+	title: 'Forms/MultiAmountSelector',
+	component: MultiAmountSelector,
+};
+
+export const multiAmountSelector = () => ({
+	components: {
+		MultiAmountSelector
+	},
+	data() {
+		return {
+			amountOptions: [
+				{
+					title: '$5',
+					key: '5',
+				},
+				{
+					title: '$10',
+					key: '10',
+				},
+				{
+					title: '$25',
+					key: '25',
+				},
+				{
+					title: '$50',
+					key: '50',
+				},
+				{
+					title: 'Other',
+					key: 'custom'
+				}
+			],
+			amountSelected: '25',
+			customAmount: 5,
+		};
+	},
+	template: `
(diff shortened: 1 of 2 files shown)


src/components/Forms/MultiAmountSelector.vue
```diff
@@ -0,0 +1,289 @@
...
+				// reset to valid default value
+				this.customAmountModel = this.minCustomAmount;
+			}
+		},
+	},
+	methods: {
+		onChange(key) {
+			// Implementation Required in Parent Component
+			// Enables tracking of selected option/key
+			this.$emit('pill-toggled', key);
```
ryan-ludwig, line 137: We can get Storybook to document our events if we comment with JsDoc. There's an example [here](https://github.com/kiva/ui/blob/master/src/components/Kv/KvCheckbox.vue#L53)

.storybook/stories/MultiAmountSelector.stories.js
```diff
@@ -0,0 +1,81 @@
...
+			:selected="amountSelected"
+			@pill-toggled="pillToggled"
+			:min-custom-amount="minCustomAmount"
+			:max-custom-amount="maxCustomAmount"
+			@custom-amount-updated="customAmountUpdated"
+			:split-pills="true"
+		/>
+	`,
+	methods: {
+		// Implementation Required in Parent Component
```
ryan-ludwig, line 62: Nice callout

ryan-ludwig:
From a little higher level, if this is the only implementation of the split pill / pill-style radio, do we need the other variations of our KvRadio and KvPillToggle?

ryan-ludwig:
I think this is really good progress. There might be some tweaks to make the UX a little more clear. When I hit "custom" and it turned white with $25, I didn't understand that I was in an input field. Maybe we clear the input, and the placeholder text says "Enter custom amount" or "$ ..." or something. And we focus() on the input so the user can start typing and they see the blinking cursor. Another approach might be to keep the value at $25, but select() the text with javascript to make it more obvious it's an input, and also let them type right away.

mcstover:
@ryan-ludwig I can select the text in the Custom field. The minimum amounts have to be pre-filled.

As for the other implementations usages, I suspect we'll be seeing them in due time.

PR #1600, opened 2020-04-12
Cov 37 implement recurring onetime options

mcstover:


mcstover:
This will be minimized once https://github.com/kiva/ui/pull/1599 is merged.

src/pages/LandingPages/MGCovid19/CovidLandingForm.vue
```diff
@@ -2,61 +2,42 @@
...
-						Field is required
-					</li>
-					<li v-if="!$v.amount.minValue || !$v.amount.maxValue">
-						Enter an amount of $5-$10,000
-					</li>
-				</ul>
-			</fieldset> -->
 		</div>
 
 		<kv-button class="smaller" type="submit" :disabled="$v.$invalid">
```
eddieferrer, line 43: This line won't let the page load since $v is undefined since the validation mixin was removed. There is also no more validation on the page. This allows a user to enter an invalid amount in "Other" and proceed with the take action button. Which seems to cause some weird behavior, if the amount was too low it gets rounded up, if it was too high it seems to get truncated.

I would suggest adding validation back to this component based on the selected amount, and then you can disable the button based on that

mcstover: Yep good call. Def want to be able to disable the button.

src/components/Forms/MultiAmountSelector.vue
```diff
@@ -115,14 +117,21 @@ export default {
 	},
 	watch: {
 		customAmountModel(newVal) {
-			// emit a custom amount updated event only if valid
```
mcstover, line 118: removed to allow validation in parent to operate on current values/state

src/pages/LandingPages/MGCovid19/CovidLandingForm.vue
mcstover: It's back!

src/components/Forms/MultiAmountSelector.vue
```diff
@@ -115,14 +117,21 @@ export default {
...
+			// Implementation Required in Parent Component
+			// Enables tracking of custom option/key associated value
+			this.$emit('custom-amount-updated', newVal);
 		},
 		selected(newVal) {
+			// to focus custom amount input element
+			if (newVal === 'custom') {
+				this.$nextTick(() => {
+					try {
+						this.$refs.kvCurrencyRef[0].$refs.kvCurrencyInputRef.focus();
```
mcstover, line 129: @ryan-ludwig This has the affect of focusing the the input and placing the blinking cursor alongside the number. The focus event to `select` the text is applied to the parent component above and only works when you click into the input field AND if the input is already present. I tried adding an extra `select()` event here too but it had no effect.

mcstover: An overall improvement but it got really messy to add that focus event :(

ryan-ludwig: Yeah, this select is acting finicky. I was able to make the focus happen (but not select oddly) using
```
				this.$nextTick(() => {
					try {
						this.$refs.kvCurrencyRef[0].$el.select();
					} catch (e) { ...
```

ryan-ludwig: @mcstover When I select "custom", the input is populating with $5 for me. I don't think we want to anchor the user at the smallest amount possible. Maybe we should anchor it to $25 and let them modify from there, or set it to whatever was selected before they hit custom? I think best case is to have it empty, but I'm not sure how we get around the validation rules. @eddieferrer do you have any thoughts?

eddieferrer: I don't like the idea of hacking around the validation rules.. I'm fine with having custom start at whatever is selected before they hit custom

mcstover: Focus was already working fine when clicking the custom option  though I see now that it's not setting focus when you switch between onetime/recurring if you previously had the custom option selected. Getting it to select the text isn't required if the focus stuff is working ok.

Tiff and Olly will instruct on what the default should be when the custom option is chosen. I don't think we should leave it empty but if they wanted to do that it would be fine to just let the validation rule show it's error as useful instruction for the valid range that needs to be input.

mcstover: Ok latest push adds focus functionality when switching between the onetime/recurring components if the custom option was selected prior. Focus was already in place when click the "Other" box within the components.

I'll follow up during testing to see what if Olly + Tiff want default amounts and what they should be for each deposit type if so.

src/components/Forms/MultiAmountSelector.vue
```diff
@@ -27,11 +27,13 @@
...
 					<kv-currency-input
 						id="custom-amount-input"
+						ref="kvCurrencyRef"
 						class="input-element"
-						:class="{'custom-input-element': true, 'error': $v.$invalid }"
+						:class="{'custom-input-element': true, 'error': $v.$invalid}"
 						v-model="customAmountModel"
 						:aria-disabled="!selected === 'custom'"
 						:disabled="!selected === 'custom'"
+						@focus.native="$event.target.select()"
```
ryan-ludwig, line 36: Is this guy doing anything?

mcstover: If you have 1 instance of this component that just sits on the page it will select the text once you click into the input (after you've given focus to something else that is).

My implementation using focus between instances in the parent completely overrides it though.

I can remove it.

mcstover: gone

PR #1602, opened 2020-04-14
Add guards for missing country information in algolia facets.

mcstover:


mcstover:
This deals with the recurring failures (and massive number of sentry errors) that occur when a loan is missing country data in Algolia.

src/pages/Lend/Filter/FilterSections/LocationMulti/FilterSectionLocationMulti.vue
```diff
@@ -54,14 +55,20 @@ export default {
 	},
 	methods: {
 		transformItems(items) {
+			// filter out any undefined or malformed facets ref. VUE-271
+			const filteredItems = _filter(items, item => {
+				if (typeof item !== 'undefined' && item.value !== '>') {
```
emuvente, line 60: The function passed to _filter just needs to return a boolean, so it's possible for this to be shortened to something like
```
_filter(items, item => {
    return typeof item !== 'undefined' && item.value !== '>';
});
```
And you could do the same with the facet filter in algolia-init-mixin.js

PR #1607, opened 2020-04-14
Update default amounts and options on landing page form along with st…

mcstover:
…eps content.


mcstover:
FYI @ryan-ludwig The amount updates from Tiff and Henrique are now set in this component. They chose to increase the default custom amounts.

PR #1612, opened 2020-04-16
Bugfix grow 11 hero alignment

mlibazisi:
Ticket: https://kiva.atlassian.net/browse/GROW-11

Adjusted alignment of "amount selector" so it displays without cutting-off on different sized screens

large
medium
small


src/pages/MonthlyGood/HowItWorksExperiment.vue
```diff
@@ -0,0 +1,111 @@
+<template>
```
mcstover, line 1: Please remove this file and make prop based updates to the existing HowItWorks component.

src/pages/MonthlyGood/LandingFormExperiment.vue
```diff
@@ -11,10 +11,10 @@
 				@custom-amount-updated="customAmountUpdated"
 				:split-pills="true"
 			/>
-			<span class="large-4 medium-4 small-12 columns select-category-label">
+			<span class="xlarge-5 columns select-category-label">
```
emuvente, line 14: Using the row/column classes isn't necessary for this component. You can remove the wrapping `<div class="row">` and remove the `xlarge-5`, `xlarge-7`, and `columns` classes from these spans.
Additionally, you can remove the
```
max-width: 8.25rem;
padding-right: 0;
```
from the css for `.select-category-label`.
Since these elements are spans, they will default to `display: inline` which will give you the appearance you're currently going for.

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -314,14 +321,47 @@ export default {
...
+			margin-left: 1rem;
+		}
+
+		@include breakpoint(medium only) {
+			top: 17rem;
+		}
+
+		@include breakpoint(small only) {
+			top: 17rem;
+		}
```
emuvente, line 337: I recommend defining these rules like this:
```
...
::v-deep .overlay-content {
    top: 17rem;
    bottom: auto;
    transform: none;
    @include breakpoint(large) {
        margin-left: 1rem;
        top: 2rem;
    }
}
...
```
You can also do a similar thing for the `.overlay-column` rules below.

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -314,14 +321,47 @@ export default {
...
+			}
+
+			@include breakpoint(small only) {
+				margin: 0 auto;
 			}
 		}
 	}
 
+	::v-deep .images > div {
+		height: 32rem;
```
emuvente, line 360: Which div is this? It's probably better to let the height of the div come from the size of the elements within it, rather than using a fixed size.

src/pages/MonthlyGood/HowItWorksExperiment.vue
```diff
@@ -0,0 +1,111 @@
...
+@import 'settings';
+
+.how-works-section-wrapper {
+	@include breakpoint(medium only) {
+		padding-top: 12rem;
+	}
+
+	@include breakpoint(small only) {
+		padding-top: 14rem;
+	}
```
emuvente, line 65: When using breakpoints to change the padding, it's better to use a small-screen first approach, in this case that could look like:
```
.how-works-section-wrapper {
    padding-top: 14rem;
    @include breakpoint(medium) {
        padding-top: 12rem;
    }
    @include breakpoint(large) {
        padding-top: 0; // or whatever you want this to be
    }
```

mlibazisi: @emuvente Thanks for the comments. I'm new to front-end dev so i'm figuring it out as I go 😄

PR #1619, opened 2020-04-21
Experiment grow 74 covid 19 text to tip line

mlibazisi:
Change the tip line item in /checkout to include copy tied to our COVID-19 relief efforts

### DIFF
--- a/src/components/Checkout/DonationItem.vue
+++ b/src/components/Checkout/DonationItem.vue
@@ -198,7 +198,7 @@ export default {
 						// eslint-disable-next-line max-len
 						client.query({ query: experimentAssignmentQuery, variables: { id: 'specific_donation_use_text' } }),
 						// Get the assigned experiment version for GROW-74
-						client.query({ query: experimentAssignmentQuery, variables: { id: 'checkout_donation_tag_line' } }),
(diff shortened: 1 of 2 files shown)


src/components/Checkout/DonationItem.vue
```diff
@@ -315,6 +320,17 @@ export default {
 				}) || {};
 				this.specificDonationUseTextExperiment = specificDonationUseTextExperiment.version === 'shown';
 			}
+			// GROW-74: Donation tag line
+			if (this.hasLoans) {
+				const donationTagLineExperiment = this.apollo.readFragment({
+					id: 'Experiment:checkout_donation_tag_line',
+					fragment: experimentVersionFragment,
+				}) || {};
+				if (donationTagLineExperiment.version === 'shown') {
```
mcstover, line 329: Please fire an event for the control case as well.

mlibazisi:
> Thanks for the clean up. You'll need to update the query for your experiment setting here: https://github.com/kiva/ui/blob/experiment-GROW-74-covid-19-text-to-tip-line/src/graphql/query/checkout/donationData.graphql
> 
> Otherwise you will never have the setting to operate on...

@mcstover Oh, now I understand what you meant! Thanks!

PR #1620, opened 2020-04-21
Deploy Storybook to Github Pages

mcstover:


.github/workflows/storybook-to-gh-pages.yml
```diff
@@ -0,0 +1,37 @@
...
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+#   push:
+#     branches: [ master ]
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '1 7 * * *'
```
mcstover, line 12: Currently set to a once daily with the assumption that we can go in a manually run the action if we need to.

ryan-ludwig:
Sweet!

PR #1621, opened 2020-04-21
Fix storybook+apollo bug

ryan-ludwig:


ryan-ludwig:
Verified locally, and Eddie gave me the fix, so merging.

eddieferrer:
Thank you

PR #1629, opened 2020-04-22
Covid Response updated design pass

ryan-ludwig:
COV-46, COV-48, COV-49


src/assets/scss/global/typography.scss
```diff
@@ -29,6 +29,14 @@ h1.huge {
 	@include overlay-text();
 }
 
+.medium-text {
```
ryan-ludwig, line 32: Started seeing patterns with the text sizes in Olly's new comp. Added these two mixins here and to the typography styleguide

ryan-ludwig:
Once we're ready to merge this we should be able to get feedback outside of the VPN via the storybook story.

src/assets/scss/_settings.scss
```diff
@@ -249,11 +249,14 @@ $small-text-font-size: 0.875rem;
 $small-text-line-height: 1.357rem;
 
 $medium-text-font-size: 1.25rem;
-$medium-text-line-height: 1.125;
+$medium-text-line-height: 1.35;
```
mcstover, line 252: This change looks like it might affect the text layout or size of the KvDropDownRounded component...did you notice any visual changes there?

mcstover: Outside of this, I didn't see any explicit usages of the medium-text-line-height though it may cascade to the medium breakpoint by default.

ryan-ludwig: Good catch, set KvDropdownRounded explicitly to 1.125 to make sure it isn't effected.

PR #1633, opened 2020-04-23
Shrink size and spacing for CookieBanner to reduce usage of screen re…

mcstover:
…al estate.


mcstover:
To be replicated in legacy at a later time.

PR #1636, opened 2020-04-24
Autolending When

eddieferrer:
* Refactored/organized components
* Styled autolending settings when section

AUTO-130

Left the autolending messaging unchanged in a new component.


src/pages/Autolending/AutolendingWhen.vue
```diff
@@ -0,0 +1,248 @@
...
+		preFetch: true,
+		result({ data }) {
+			this.isEnabled = !!_get(data, 'autolending.currentProfile.isEnabled');
+			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
+			this.lendAfterDaysIdle = _get(data, 'autolending.currentProfile.lendAfterDaysIdle');
+			this.isChanged = !!_get(data, 'autolending.profileChanged');
+			this.donation = _isFinite(donationPercentage) ? donationPercentage : 15;
+			this.donationToggle = this.donation !== 0 ? 'on' : 'off';
+			// After initial value is loaded, setup watch
+			this.$watch('donation', (donation, previousDonation) => {
```
mcstover, line 161: What's the affect of enabling these watcher in the ssr context? Seems like we could be setting up some type of reactive code that can never be executed? Does it just hang around in the node app instance memory?

mcstover: I think we usually activate watchers once we know we're in the client...

eddieferrer: Yeah thats fair, I was just trying to avoid the initial firing of the watcher on setting the initial value. I'll look into it

PR #1646, opened 2020-04-28
GROW-79 Replace home page hero with MG sign-up form

mlibazisi:

### DIFF
--- a/src/pages/LandingPages/MGCovid19/MGCovidHero.vue
+++ b/src/pages/LandingPages/MGCovid19/MGCovidHero.vue
@@ -84,10 +84,6 @@ export default {
 		},
 		result({ data }) {
 			this.covidLandingActive = _get(data, 'general.mg_covid_active.value') === 'true' || false;
-			// Force the covid landing form to show when running the GROW-79 experiment
-			if (!this.covidLandingActive && this.isExperimentActive) {
-				this.covidLandingActive = true;
-			}
 			this.isMonthlyGoodSubscriber = _get(
 				data,
 				'my.autoDeposit.isSubscriber',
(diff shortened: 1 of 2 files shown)


src/pages/LandingPages/MGCovid19/MGCovidHero.vue
```diff
@@ -84,6 +84,10 @@ export default {
 		},
 		result({ data }) {
 			this.covidLandingActive = _get(data, 'general.mg_covid_active.value') === 'true' || false;
+			// Force the covid landing form to show when running the GROW-79 experiment
+			if (!this.covidLandingActive && this.isExperimentActive) {
```
mcstover, line 88: We should not be overriding this feature flag.

It's purpose is two-fold; first as a gate in production today which you don't have to worry about and second as a kill switch for preventing entry into MG using the Covid Category and/or using the onetime deposit option.

Should there be a need to turn this off quickly we want to have a way that isn't subject to stale cookies in the browser as an experiment might be. That said, we would most likely want to turn off the experiment too.

mlibazisi: Got it. Thanks!

src/pages/LandingPages/MGCovid19/MGCovidHero.vue
```diff
@@ -95,6 +95,9 @@ export default {
 		category: {
 			type: String,
 			default: 'covid-relief'
+		},
+		isExperimentActive: {
```
emuvente, line 99: It doesn't look like there are any usages of this currently. What is this going to be used for? Is that code still in progress?

mlibazisi: @emuvente good catch. will remove it. thanks

PR #1653, opened 2020-04-30
hide promo banner on covid landing and mg setup and thanks pages alon…

mcstover:
…g with on checkout.


mcstover:
This is already in the current release branch to go out with the hotifx.

PR #1656, opened 2020-05-01
Auto 132 cleanup and copy changes

eddieferrer:
Copy changes. Cleaned up the "countries" copy since the new copy just says '# countries' and we dont need to know the names of the countries

Confirmed with design that the mobile counter should be removed. Removed that component


src/pages/Autolending/AutolendingSettingsPage.vue
```diff
@@ -9,13 +9,7 @@
 		<autolending-who />
 
 		<div class="row column save-button-area">
-			<save-button class="show-for-large" />
```
mcstover, line 12: Why is the count and warning less important b/c you're using a mobile device?

eddieferrer: The count is now displayed in the "Who options" modal and also in the who you'll support section on the settings page. It is displayed the same on mobile and desktop. 

The warning is now also displayed the same on mobile and desktop (pop up modal message)

PR #1657, opened 2020-05-01
Update images to support new campaign

mcstover:


ryan-ludwig:
Cool images, do you have the sources available? I think these might be smaller and sharper as PNGs or possibly even SVG if they provided vector images.

PR #1661, opened 2020-05-01
GROW-93 Add matched donation language

mlibazisi:


src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -392,6 +399,7 @@ export default {
 				});
 		},
 		result({ data }) {
+			this.isMGTaglineActive = _get(data, 'general.mg_tagline_active.value') === 'true' || false;
```
mcstover, line 402: this should start with `ui` not `general`
we have special handling for graphql and the `uiConfigSetting` that expects a setting to be named like `ui.mg_tagline_active` 

We should also add the word "donation" somewhere in your setting key so we have some idea where the setting applies in the mg page...

src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -262,6 +262,12 @@ import userInfoQuery from '@/graphql/query/userInfo.graphql';
 import loanGroupCategoriesMixin from '@/plugins/loan-group-categories';
 
 const pageQuery = gql`{
+    ui {
+      mg_donationtagline_active: uiConfigSetting(key: "mg_donationtagline_active") {
```
mcstover, line 266: make this line...
`mgDonationTaglineActive: uiConfigSetting(key: "ui.mg_donationtagline_active") {`

src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -392,6 +399,7 @@ export default {
 				});
 		},
 		result({ data }) {
+			this.isMGTaglineActive = _get(data, 'ui.mg_donationtagline_active.value') === 'true' || false;
```
mcstover, line 402: and make this line:
`this.isMGTaglineActive = _get(data, 'general.mgDonationTaglineActive.value') === 'true' || false;`

PR #1664, opened 2020-05-02
Homepage slideshow tweaks

ryan-ludwig:
Prevent a padded green box if there's nothing in contentful for the headline or subhead slot.

If no CTA text is provided, make the whole image a link.
This will allow marketing to upload images with CTAs embedded in the image (for COV campaign)


ryan-ludwig:
Awesome, thanks @mcstover !

PR #1665, opened 2020-05-02
FOUC: try moving inline styles off of the root element

ryan-ludwig:
It's a shot in the dark, but since the delayed styles are for TheHeader and TheFooter I thought I'd try moving the inline styles which we use for theming off of their root elements. Maybe webpack/vue is getting hung up on that, and it's a recent change related to those elements.

Thought we could see if it helped in dev. If not, will roll it back.

### DIFF
--- a/src/components/WwwFrame/TheFooter.vue
+++ b/src/components/WwwFrame/TheFooter.vue
@@ -1,276 +1,278 @@
 <template>
-	<footer class="www-footer" :style="cssVars">
-		<nav class="small-footer hide-for-large" aria-label="Footer navigation">
-			<ul class="hide-for-print">
-				<li>
-					<router-link :to="applyUrl">
-						Borrow
+	<footer class="www-footer">
+		<div :style="cssVars">
+			<nav class="small-footer hide-for-large" aria-label="Footer navigation">
+				<ul class="hide-for-print">
+					<li>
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheFooter.vue
```diff
@@ -1,276 +1,278 @@
 <template>
-	<footer class="www-footer" :style="cssVars">
-		<nav class="small-footer hide-for-large" aria-label="Footer navigation">
-			<ul class="hide-for-print">
-				<li>
-					<router-link :to="applyUrl">
-						Borrow
+	<footer class="www-footer">
+		<div :style="cssVars">
```
ryan-ludwig, line 3: this is the change

ryan-ludwig:
Another way to test this theory would be to remove the CSS custom properties all together and see if that has an impact.

ryan-ludwig:
@mcstover Let me know how to test from my VM :) Will test things from there.

ryan-ludwig:
Following up on this, @emuvente @mcstover  removing the CSS custom properties on TheFooter and TheHeader and even the serverCacheKey had no effect. The header and footer styles are still being injected into the html after some of the JS which I believe is why we see the FOUC. Still no idea why those styles are loaded later though. Not sure if it's something related to those particular components, or something else.

image

PR #1667, opened 2020-05-04
Feature grow 94 matching tip text

mlibazisi:


src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -92,9 +92,13 @@
 											v-model="donationCheckbox"
 											@change="donationCheckboxChange()"
 										/>
-										<span
-											class="additional-left-pad-spans display-inline-block"
-										>{{ donationTagLine }}</span>
+										<span class="additional-left-pad-spans display-inline-block">
```
ryan-ludwig, line 95: Because you have a block level element (a div) inside of this inline element, I'd recommend changing this span to a div.

PR #1682, opened 2020-05-07
GROW-95 New covid response icons / background

ryan-ludwig:
Change out icons, add background colors/images, rework paddings/margins because of background color. Swap dash for a shorter dash.


src/pages/LandingPages/MGCovid19/MGCovidHero.vue
```diff
@@ -119,8 +119,8 @@ export default {
 	computed: {
 		pageCopy() {
 			return {
-				headline: 'Join Kiva\'s Global <span class="no-wrap">COVID—19</span> Response', // eslint-disable-line max-len
-				subhead: 'Lend a hand to provide relief to entrepreneurs and small businesses impacted <span class="no-wrap">by COVID—19.</span>', // eslint-disable-line max-len
+				headline: 'Join Kiva\'s Global <span class="no-wrap">COVID‑19</span> Response', // eslint-disable-line max-len
+				subhead: 'Lend a hand to provide relief to entrepreneurs and small businesses impacted <span class="no-wrap">by COVID‑19.</span>', // eslint-disable-line max-len
```
mcstover, line 123: Please note the original `dash` character would prevent the break of the Covid-19 on it's own without the no-wrap span. Previously we had other words in side the no-wrap span which I had to move out to prevent them word "Response" from falling off of the page to the right in mobile.

mcstover: https://www.toptal.com/designers/htmlarrows/punctuation/non-breaking-hyphen/

ryan-ludwig:
> Please verify the spacing between the hero and explained sections on the landing page after the code is compiled. I previously had to re-introduce the section class to the landing page layout to fix this. Dev mode did not show any issue but once compiled the was no top padding due to hero styles taking precedence.

Gave it a test, should be good!

PR #1692, opened 2020-05-12
Check for requestAnimationFrame before using in a loopig animation. F…

mcstover:
…all back to scrollTo.

Check ticket for details from Sentry.


mcstover:
🥳

PR #1708, opened 2020-05-12
GROW-96 Test different default contribution amounts

mlibazisi:


src/pages/LandingPages/MGCovid19/CovidLandingForm.vue
```diff
@@ -163,8 +176,43 @@ export default {
...
+			if (this.isExperimentActive) {
+				this.recurringAmountSelection = '25';
+				this.recurringCustomAmount = 25;
+				this.recurringAmount = 25;
+			}
+		},
+	},
+	mounted() {
+		// Fire Event for GROW-96
+		this.$kvTrackEvent(
```
mcstover, line 210: @mlibazisi Overall looks good. For this analytics call we should wrap it in a check that looks at the experiment setting to see if it is enabled or not. We don't want to fire this event if we haven't yet started the experiment.

This update will allow us to deploy the code then turn it on at a later date.

mlibazisi: @mcstover Should I set it as a separate admin setting that i will need to query for?

mcstover: It will be present on the result of the `pageQuery` variable above so you could do a readQuery with query set to `pageQuery` and get the full experiment setting.

PR #1709, opened 2020-05-12
GROW-84 design/text updates

ryan-ludwig:
Remove header/footer theme for continuity in the site.
Remove kv-checkout-steps in favor of text stating you're on step1.
Add button text to alert the user when we're moving to step2.
Query for trusteeId and append that to the link if we have one.


src/pages/Borrow/BorrowHowLong.vue
```diff
@@ -18,31 +20,55 @@
...
 	components: {
 		IntroGraphic,
 		KvButton,
 	},
+	data() {
+		return {
+			trusteeId: ''
+		};
+	},
+	apollo: {
```
ryan-ludwig, line 58: @emuvente Will this work instead of you passing in a trusteeId from the old stack as a query param? Also wasn't sure of a good way to test this.

emuvente: No, because this is a person who hasn't signed up yet. We're passing along the id of the trustee that referred this user to the /borrow page

ryan-ludwig: > No, because this is a person who hasn't signed up yet. We're passing along the id of the trustee that referred this user to the /borrow page

ahh, thanks for the explanation.

emuvente: You should just use the page query variable in the created hook:
```
created() {
    this.trusteeId = _get(this, '$route.query.refid', '');
}
```

src/pages/Borrow/BorrowIntro.vue
```diff
@@ -69,7 +69,15 @@
 				</kv-button>
 			</div>
 			<div class="small-6 columns">
-				<kv-button to="/borrow/pre-application/community" class="smaller cta-btn">
+				<kv-button
+					:to="{
+						path: '/borrow/pre-application/community',
+						query: {
+							refid: this.$route.query.refid
```
emuvente, line 76: `this` is implicit in templates, so this line should be `refid: $route.query.refid` (same for the button in `BorrowCommunity.vue`)

PR #1710, opened 2020-05-13
Add check for null on category_row_hillclimb experiment version.

mcstover:


emuvente:
This should fix the invariant violation

mcstover:
@emuvente indeed it does!

PR #1721, opened 2020-05-13
Import common version of fuse build for now until https://github.com/…

mcstover:
…krisk/Fuse/pull/383/files is merged to allow webpack to transpile the esm build.


mcstover:
:(

PR #1722, opened 2020-05-14
Prefetch some items in the lend menu

eddieferrer:
* Display 'all' categories in the lend menu
* Prefetch my info and regions info in the lend menu
* Show loader for categories query which takes a long time

VUE-222

There was some pretty extensive discussion on VUE-222, I gave this a shot, looks good to me, but interested in everyones thoughts.

### DIFF
--- a/src/components/WwwFrame/LendMenu/LendMegaMenu.vue
+++ b/src/components/WwwFrame/LendMenu/LendMegaMenu.vue
@@ -2,7 +2,7 @@
 	<div class="lend-mega-menu">
 		<div class="categories-section" :style="{ marginLeft: categoriesMargin }">
 			<h2>Categories</h2>
-			<kv-loading-spinner v-if="isLoading" />
+			<kv-loading-spinner v-if="isChannelsLoading" />
 			<ul :style="categoriesStyle">
 				<li
 					v-for="(category, index) in categories"
@@ -44,6 +44,7 @@
 		</kv-expandable>
 		<div class="middle-section">
 			<h2>Regions</h2>
+			<kv-loading-spinner v-if="isRegionsLoading" />
 			<ul>
 				<li v-for="region in regions" :key="region.name">
 					<button
@@ -131,7 +132,11 @@ export default {
 			type: Number,
 			default: 0,
 		},
-		isLoading: {
+		isRegionsLoading: {
+			type: Boolean,
+			default: true,
+		},
+		isChannelsLoading: {
 			type: Boolean,
 			default: true,
 		},
--- a/src/components/WwwFrame/LendMenu/LendMegaMenu.vue
+++ b/src/components/WwwFrame/LendMenu/LendMegaMenu.vue
@@ -2,6 +2,7 @@
 	<div class="lend-mega-menu">
 		<div class="categories-section" :style="{ marginLeft: categoriesMargin }">
 			<h2>Categories</h2>
+			<kv-loading-spinner v-if="isLoading" />
 			<ul :style="categoriesStyle">
 				<li
 					v-for="(category, index) in categories"
@@ -88,14 +89,11 @@
 			</ul>
 		</div>
 		<kv-expandable property="width" :skip-leave="true">
-			<!-- eslint-disable -->
 			<div
-				v-for="region in regions"
+				v-for="region in openRegions"
 				:key="region.name"
-				v-if="isOpenSection(region.name)"
 				class="right-section"
 			>
-			<!-- eslint-enable -->
 				<h2>{{ region.name }}</h2>
 				<country-list :countries="region.countries" />
 			</div>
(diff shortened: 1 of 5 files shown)


src/graphql/query/categoryRows.graphql
```diff
@@ -13,5 +13,13 @@ query categoryRows {
 				}
 			}
 		}
+		countryFacets {
```
eddieferrer, line 16: Adding this was the only way I could resolve an invariant error. 😕

src/components/WwwFrame/LendMenu/LendMegaMenu.vue
```diff
@@ -166,6 +170,9 @@ export default {
 		},
 		sectionOpen() {
 			return this.openedSection !== '';
+		},
+		openRegions() {
```
eddieferrer, line 174: Appeases eslint. No combining v-for with v-if

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -107,48 +130,30 @@ export default {
...
-		stopLoading() {
-			if (this.loadingSemaphore > 0) {
-				this.loadingSemaphore -= 1;
-			} else {
-				this.loadingSemaphore = 0;
-			}
-		},
 	},
-	watch: {
-		hasUserId(hasUserId) {
```
eddieferrer, line 128: I assume this was maybe left over from before auth0? The userId seems to be reliably available at prefetch, so not sure why we need the watch and the loadingSemaphore...

mcstover:
oooh fun!

src/components/WwwFrame/LendMenu/TheLendMenu.vue
emuvente: The watch is necessary to handle the user becoming logged in while still in the page, so that when they open the menu again they get the personalization

src/graphql/query/categoryRows.graphql
emuvente: This is an expensive thing to be adding to a query that doesn't appear to be using it. Where were you getting an invariant violation?

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -107,48 +130,30 @@ export default {
 		onLoad() {
 			this.apollo.watchQuery({ query: publicLendMenuQuery }).subscribe({
 				next: ({ data }) => {
-					this.userId = _get(data, 'my.userAccount.id');
```
emuvente, line 110: This still needs to be queried for as the user may become logged by using the pop-up login form the sign-in button in the header

src/graphql/query/categoryRows.graphql
eddieferrer: On the homepage.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
eddieferrer: Can you give me an example of where this isn't working? I tested this code using this flow: 

1. Have something in basket
2. be logged out
3. go to `/checkout`
4. open lend menu
5. click 'Sign In' to trigger log in pop up
6. sign in
7. open lend menu again

src/components/WwwFrame/LendMenu/TheLendMenu.vue
eddieferrer: See my comment above

emuvente: Turns out you're already still querying for this because you've moved it into the apollo.result function, which has a watchQuery automatically setup for it and thus is responding to any changes in the logged-in state, my mistake.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
emuvente: Yeah that makes sense since using the apollo options sets up the watch for you.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -61,8 +78,17 @@ export default {
 				'Oceania'
 			],
 			loadingSemaphore: 0,
+			isApolloLoading: true
 		};
 	},
+	apollo: {
+		query: pageQuery,
+		preFetch: true,
```
mcstover, line 86: Prefetch to be removed which allows us to not add the `countryFacets` to the categoryRows query. We only want this in the client.

emuvente: 👍

eddieferrer:
Prefetch only user id and added independent loaders to menu.

PR #1728, opened 2020-05-18
Added checking of global promo banner field to determine visibility, …

eddieferrer:
…removed css visibility, always hide on checkout route. VUE-278

Note: I have already migrated the contentful contentType modifications and the list type field hiddenUrls is on global promo content types in dev and master. I have also modified the current global promo banner (covid19) to add the excluded Urls to it.

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
+++ b/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
@@ -32,7 +32,7 @@ export default {
 			specialConditions: null
 		};
 	},
-	mounted() {
+	created() {
 		this.federation.query({
 			query: contentful,
 			variables: {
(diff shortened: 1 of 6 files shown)


src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -66,12 +66,25 @@ export default {
 						'endDate'
 					);
 				});
+
 				if (activePromoBanner) {
+					// always hide the promo banner on the checkout page.
+					if (this.$route.path === '/checkout') {
```
eddieferrer, line 72: Keeps all the hiding logic in this component, and not css rules.

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -91,15 +104,6 @@ export default {
 			}
 		});
 	},
-	serverPrefetch() {
```
eddieferrer, line 94: This was some experimental code of mine, it wasn't working as intended

PR #1732, opened 2020-05-18
Check on created to see if logged in user is part of this team alread…

mcstover:
…y. If so redirect to authenticate.


src/components/Forms/JoinTeamForm.vue
```diff
@@ -164,9 +164,34 @@ export default {
...
+				this.apollo.query({
+					query: myTeamsQuery,
+					variables: {
+						teamIds: [teamId]
+					}
+				}).then(({ data }) => {
+					this.isMember = _get(data, 'my.teams.values', []).length;
+					// if lender is a member proceed
+					if (this.isMember) {
+						window.location = '/authenticate/redirect&team_id=278842&promo_id=286';
```
emuvente, line 181: Probably want to use the query parameters here...

src/components/Forms/JoinTeamForm.vue
```diff
@@ -164,9 +164,34 @@ export default {
...
+						this.loading = false;
+					}
+				});
+			}
 		}
 	},
 	created() {
+		// in the browser check if the lender has already joined the team
+		if (typeof window !== 'undefined') {
+			this.handleAlreadyJoined();
```
emuvente, line 193: What if we did this in the preFetch instead? That way we could redirect back without having to load the page in the browser (by rejecting the prefetch promise with the url to redirect to)

PR #1740, opened 2020-05-19
Add K8 configs for test env. Clean up general test env configs. Add U…

mcstover:
…i deployent to k8 test from jenkins. Remove initial root level kiva-ui.yml file which should now be driven by specific env configs + templates.


Jenkinsfile
```diff
@@ -53,8 +53,22 @@ pipeline {
...
+
+    stage('Kubernetes_Test_Deployment') {
+      when {
+        branch 'master'
       }
+      steps {
+        echo "Deploying to development Kubernetes cluster..."
+        withKubeConfig([credentialsId: "jenkins-ci-kubeconfig"]) {
+          withAWS([credentials: "jenkins-ci"]) {
+            sh "helm3 upgrade --install ${K8S_RELEASE_NAME} ./deploy/charts --namespace ${K8S_NAMESPACE} --values ./deploy/test/values.yaml --set image.tag=${TAG_NAME} --set deployenv.environment=k8stest"
```
JoeTravisKiva, line 67: Are you planning on deploying this to the test namespace?

Dockerfile.jenkins
```diff
@@ -33,4 +33,4 @@ RUN npm install --production
 EXPOSE 8888
 
 # Launch application
-CMD [ "npm", "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui" ]
+# CMD [ "npm", "start", "--", "--port=8888", "--config=k8sdev", "--prefix", "/ui" ]
```
JoeTravisKiva, line 36: Is this meant to be commented?

mcstover: Yes, the --config value must be dynamic and the rest should be as well. I'd rather control this through the vars in the deploy configs.

Jenkinsfile
mcstover: Not sure what you mean by that.

The configs referenced here are to deploy a new instance of the container to ui-test.dk1.kiva.org with specific UI configuration which will eventually be served in the context of https://www.test.kiva.org.

leoperry: I think it's that `--namespace ${K8S_NAMESPACE}` will end up deploying this to the (dev) namespace -- kiva-marketplace. So in the context of this pipeline we'd end up deploying over our dev deployment.

mcstover: Oh I see now... can update

Dockerfile.jenkins
mcstover: Looking at this more and I may need switch to directly supporting ENV vars instead of passing these flags through the npm start command. Then we can set this to `CMD [ "npm", "start"]`

FYI @emuvente

mcstover: Then again...I did this hoping this would take care of starting the container:
https://github.com/kiva/ui/blob/master/deploy/charts/templates/deployment.yaml#L33

Is that a correct assumption? @JoeTravisKiva @leoperry

JoeTravisKiva: So I think that's a correct assumption, but I believe that it would be better to apply the environment specific config through ENV variables as you mention above. That's how we do it in the graphql federation service, and it also gives more flexibility in being able to override the values at runtime.

Also, not entirely on the same subject, but we also use https://github.com/krallin/tini as an entry point, which we got from https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md

mcstover: @JoeTravisKiva I'll look into adding Tini at a later time

I'm going to try to start the app using the templated npm start script for now as changing to ENV variables will require refactoring portions of the Ui app.

PR #1741, opened 2020-05-19
Hotfix lw 1912 jointeam member check

mcstover:


mcstover:
Verified these updates in Dev and Automation tests have also verified the redirect working.

PR #1743, opened 2020-05-20
Working POC for vue-apollo

eddieferrer:
* Initial POC for having vue-apollo installed along our current setup and demonstrating what a component would look like.

Theres a lot of changes here, but most of them are just find/replace. 

Files of Interest: 

server/index.template.html
src/api/apollo.js
src/client-entry.js
src/server-entry.js
src/main.js
src/pages/PageTwo.vue

### DIFF
--- a/src/api/apollo.js
+++ b/src/api/apollo.js
@@ -18,35 +18,36 @@ export default function createApolloClient({
 	types = [],
 	uri,
 	appConfig,
-	existingCache
+	ssr = false,
+	clientId,
 }) {
-	let cache;
-	if (!existingCache) {
-		cache = new InMemoryCache({
-			fragmentMatcher: new IntrospectionFragmentMatcher({
-				introspectionQueryResultData: {
-					__schema: { types }
-				}
-			}),
-			// Return a custom cache id for types that don't have an id field
-			dataIdFromObject: object => {
-				if (object.__typename === 'Setting' && object.key) return `Setting:${object.key}`;
-				if (object.__typename === 'Shop') return 'Shop';
-				if (object.__typename === 'ShopMutation') return 'ShopMutation';
-				return defaultDataIdFromObject(object);
-			},
-			// Use a simpler underlying cache for server renders
-			resultCaching: typeof window !== 'undefined',
-			// Block modifying cache results outside of normal operations
-			// see https://github.com/apollographql/apollo-client/pull/4543
-			freezeResults: true,
-		});
-	} else {
-		cache = existingCache;
-	}
+	const cache = new InMemoryCache({
+		fragmentMatcher: new IntrospectionFragmentMatcher({
+			introspectionQueryResultData: {
+				__schema: { types }
+			}
+		}),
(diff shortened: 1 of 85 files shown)


src/main.js
```diff
@@ -61,19 +69,29 @@ export default function createApp({
 	// Provide application config to all components
 	Vue.prototype.$appConfig = appConfig;
 
+	// Apollo
+	const apolloProvider = new VueApollo({
```
eddieferrer, line 73: Vue-apollo supports multiple clients.

src/pages/Homepage/Homepage.vue
```diff
@@ -4,6 +4,11 @@
 		:footer-theme="footerTheme"
 	>
 		<component :is="activeHomepage" />
+		<router-link
```
eddieferrer, line 7: Just a temporary link to test prefetching from the homepage to page-two

src/pages/PageTwo.vue
```diff
@@ -161,51 +126,73 @@ export default {
 	},
 	inject: ['apollo', 'federation'],
 	apollo: {
```
eddieferrer, line 128: This is what the vue-apollo API would look like in a component.

eddieferrer:
This is a working POC of migrating state management to use vue-apollo. 

See page-two for what the API would look like in a component. 

Some benefits:
* 'loading' state of queries easily read
* Everything is basically a query, vue-apolllo discourages the use of readFragment, readQuery, etc. You make a query and it gets the data. 
* I believe this will improve our prefetch/cache setup and hopefully lead to less invariant errors and faster loads. 

Migration path: 
* Clear $options.apollo name space (seen in this PR)
* Refactor all uses of queries in methods/computeds, etc. - This is just another find and replace since vue-apollo exposes $apollo. So all this.apollo.query in methods in components become this.$apollo. 
* Refactor all of the new $options.kvapollo to be vue-apollo queries. 
* Remove our custom apollo plugin and preFetch functions
* profit.

src/api/apollo.js
```diff
@@ -71,12 +72,25 @@ export default function createApolloClient({
 		// Allow optimizations that are only possible because we have freezeResults=true
 		// see https://github.com/apollographql/apollo-client/pull/4543
 		assumeImmutableResults: true,
+		...(ssr ? {
+		// Set this on the server to optimize queries when SSR
+			ssrMode: true,
+		} : {
+		// This will temporary disable query force-fetching
+			ssrForceFetchDelay: 100,
```
emuvente, line 80: This is an interesting alternative to using readQuery to avoid refetching hydrated data, but it sounds like there may be challenges with setting an appropriate delay time. This pr suggests a way to accomplish the same thing without having to rely on timing: https://github.com/apollographql/apollo-client/issues/4814

src/pages/Homepage/Homepage.vue
mcstover: I think you can just use the link from `ui-site-map` right?

PR #1745, opened 2020-05-21
Removed some unused dependencies. Added node env to client.prod so it…

eddieferrer:
… can build

'apollo-fetch' and 'contentful' were left by me. They were used for the contentful local resolver at one time.

'form-data-entries' is left over from before auth0 days


package.json
```diff
@@ -79,7 +76,6 @@
 		"js-cookie": "^2.2.1",
 		"jsonwebtoken": "^8.5.1",
 		"lodash": "^4.17.15",
-		"lodash-es": "^4.17.15",
```
eddieferrer, line 82: I compared using 'lodash-es' imports vs 'lodash' and it actually increased the bundle size

mcstover: 😲

emuvente: Curious. I had assumed using es modules would give us tree-shaking and therefore less lodash in the final bundle, but I guess not

PR #1748, opened 2020-05-21
VUE-219 KvLightbox a11y improvements

ryan-ludwig:
Adds accessibility improvements following the recommendations for a dialog
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

- [x] Adds a focus trap,
- [x] Sets the role to dialog, sets the body role to document per this bootstrap issue https://github.com/twbs/bootstrap/issues/15875

Remaining items from the ARIA dialog recommendations
- [ ] When the dialog appears on the screen, keyboard focus (whose control depends upon the dialogs purpose) should be moved to the default focusable control inside the dialog. For dialogs that only provide a basic message, it could be an "OK" button. For dialogs containing a form it could be the first field in the form.
- [ ] After the dialog is dismissed, keyboard focus should be moved back to where it was before it moved into the dialog. Otherwise the focus can be dropped to the beginning of the page.
- [ ] Add an aria-labelledby attribute if we have a title

### DIFF
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -10,7 +10,7 @@
 			role="dialog"
 		>
 			<focus-lock>
-				<div class="kv-lightbox">
+				<div class="kv-lightbox" role="document">
 					<div class="row lightbox-row">
 						<div class="columns lightbox-columns">
 							<!-- eslint-disable max-len -->
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -112,11 +112,6 @@ export default {
 	},
 	mounted() {
 		this.isShown = this.visible;
-
-		if (this.$slots.title) {
-			this.$slots.title[0].elm.id = 'title';
-			this.$refs.kvlightbox.setAttribute('aria-labelledby', 'title');
-		}
 	},
 	beforeDestroy() {
 		this.closeLightbox();
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -115,7 +115,7 @@ export default {
 
 		if (this.$slots.title) {
 			this.$slots.title[0].elm.id = 'title';
-			this.$refs.kvlightbox.ariaLabelledby = 'title';
+			this.$refs.kvlightbox.setAttribute('aria-labelledby', 'title');
 		}
 	},
 	beforeDestroy() {
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -112,6 +112,11 @@ export default {
 	},
 	mounted() {
 		this.isShown = this.visible;
+
+		if (this.$slots.title) {
+			this.$slots.title[0].elm.id = 'title';
+			this.$refs.kvlightbox.ariaLabelledby = 'title';
+		}
 	},
 	beforeDestroy() {
 		this.closeLightbox();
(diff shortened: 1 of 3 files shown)


eddieferrer:
Better than it was. 👍 

I was still able to tab out of the modal and focus background elements in on page in the background by messing around with the autolend who you'll lend to settings.

src/components/Kv/KvLightbox.vue
```diff
@@ -105,6 +112,11 @@ export default {
 	},
 	mounted() {
 		this.isShown = this.visible;
+
+		if (this.$slots.title) {
+			this.$slots.title[0].elm.id = 'title';
```
eddieferrer, line 117: vendors~app.js:90792 [Vue warn]: Error in mounted hook: "TypeError: Cannot set property 'id' of undefined"

ryan-ludwig: Good catch. Removing this block completely still causes the focus lock to escape (though no console error). There's something more going on I think.

eddieferrer: I believe this console error has to do with some deprecated slot syntax we are using. It points to src/pages/Autolending/SaveButton.vue which is using this slot syntax: `#title` Which has been deprecated in 2.6 https://vuejs.org/v2/guide/components-slots.html#Named-Slots-with-the-slot-Attribute

We should make a new item to refactor all slot use to be consistant with the new syntax.

eddieferrer: I believe the best approach here is to make sure you are setting the id="title" in the components using kv-lightbox. In kv-lightbox you can check for the existence of the title slot and add the aria-labelled by property. Reference: https://github.com/vuejs/vue/issues/9580 this.$slots.title[0].elm.id is a hack I don't think we should use it

ryan-ludwig: Adding the ID in the parent component would be less hacky, but every time someone makes a lightbox they'd have to know to add the ID of "title" to what they put in the header slot, so it probably won't happen much. Then the aria-labelledby could be wrong which is probably worse than not being there at all.

A different approach could be to add a prop to kv-lightbox called "title" which takes a string. If the prop is there we always render it inside an h2 with the id, and can then attach the aria-labelledby. An upside is more consistent looking dialogs, a downside is less visual flexibility, though I suppose you could ::v-deep the styling if necessary. And less flexibility to put things beyond strings in the title. I think there's only one instance of that in [AutolendingWho.vue](https://github.com/kiva/ui/blob/2dbcbcc8a2fd6f2b3b0b3d9e44da0c62325dd869/src/pages/Autolending/AutolendingWho.vue#L26)

ryan-ludwig: I'm starting to like the prop idea more. If you want something custom you not use the title prop and build a special title in the body, but the majority of the dialogs would be consistent and handle the labelled-by out of the gate. @mcstover @emuvente thoughts?

emuvente: Yeah I think a prop is a good way to go about it, especially since, as you point out, ::v-deep can be used if alternative styling is necessary.

mcstover: I think the prop approach is fine and then for other cases where we need to have a more fully featured header we can just add and style it through the main slot...

ryan-ludwig: Cool, removed the arial-labelledby block for this PR. Will open a different with the prop change

PR #1766, opened 2020-05-27
VUE-318 lightbox improvements

ryan-ludwig:
- Use prop to set title for consistency. 
- Add aria-labelledby and fix focus-lock for a11y. 
- Remove fixed width from lightboxes to allow narrower content without overrides. 
- Add a css custom property for overriding the title color.
- Remove unused "showCloseButton" property
- Scope lightbox styles

### DIFF
--- a/src/pages/Autolending/AutolendingStatus.vue
+++ b/src/pages/Autolending/AutolendingStatus.vue
@@ -37,17 +37,10 @@
 
 				<kv-lightbox
 					class="autolending-status-lightbox"
-					:no-padding-bottom="true"
-					:no-padding-top="true"
 					:visible="showLightbox"
+					title="Change your auto-lending status"
 					@lightbox-closed="showLightbox = false"
 				>
-					<template slot="title">
-						<h3>
-							Change your auto-lending status
-						</h3>
-						<hr>
-					</template>
 					<div class="status-radio-wrapper">
 						<kv-radio
 							data-test="is-autolending-on"
@@ -86,7 +79,6 @@
 						</kv-radio>
 					</div>
 					<template slot="controls">
-						<hr>
 						<kv-button
 							data-test="status-save-button"
 							class="smaller button"
@@ -266,22 +258,8 @@ export default {
 @import 'settings';
 
 .autolending-status-lightbox {
-	::v-deep .kv-lightbox {
-		max-width: 520px;
-	}
-
-	h3 {
-		margin-top: 1.75rem;
-		display: inline-block;
-		font-weight: $global-weight-bold;
-	}
-
 	.status-radio-wrapper {
 		padding: 1.5rem 0;
-	}
-
-	.button {
-		margin-bottom: 1.75rem;
 	}
 
 	.dropdown-wrapper {
--- a/src/pages/Autolending/AutolendingWho.vue
+++ b/src/pages/Autolending/AutolendingWho.vue
@@ -17,30 +17,22 @@
 
 				<kv-lightbox
 					class="autolending-who-lightbox"
-					:no-padding-bottom="true"
-					:no-padding-top="true"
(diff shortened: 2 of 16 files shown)


src/pages/Autolending/AutolendingWho.vue
```diff
@@ -17,30 +17,22 @@
```
ryan-ludwig, line 17: @eddieferrer Would appreciate if you could double check the autolending settings lightboxes to make sure I didn't mess anything up here.

src/pages/Autolending/AutolendingWho.vue
```diff
@@ -17,30 +17,22 @@
...
-									@click.prevent="backToAllOptions"
-								>
-									<kv-icon class="arrow back-arrow" name="small-chevron" :from-sprite="true" />
-									Back to all options</a>
-							</transition>
-							<hr>
-						</div>
-					</template>
 					<div class="who-inputs-wrapper">
+						<transition :name="slideTransition" mode="out-in">
```
eddieferrer, line 25: This is not the same as the lines removed in the slot. When the transition happens, the title should be replaced with the "back to all options" with a slide transition

src/pages/Autolending/AutolendingStatus.vue
```diff
@@ -37,17 +37,10 @@
...
 				<kv-lightbox
 					class="autolending-status-lightbox"
-					:no-padding-bottom="true"
-					:no-padding-top="true"
 					:visible="showLightbox"
+					title="Change your auto-lending status"
 					@lightbox-closed="showLightbox = false"
 				>
-					<template slot="title">
-						<h3>
```
eddieferrer, line 46: The autolend h3 titles have different styling than the styling of the lightbox titles provided in this PR. Take a look at the dev env for reference.

src/pages/Autolending/AutolendingWho.vue
```diff
@@ -123,7 +115,6 @@
 						</transition>
 					</div>
 					<template slot="controls">
-						<hr>
```
eddieferrer, line 126: The autolend settings lightboxes should have an hr line separating the title and the button area...

src/pages/Autolending/SaveButton.vue
```diff
@@ -14,27 +14,23 @@
 		</kv-button>
 		<kv-lightbox
 			:visible="warningVisible"
-			:no-padding-top="true"
```
eddieferrer, line 17: These no-padding-* props are being used, not sure they should be removed.

ryan-ludwig:
@eddieferrer There are visual changes to the lightboxes in autolending settings but I feel that they still convey the same hierarchy and are now consistent with the lightboxes throughout the site.

Before:
image
image
image

After:
image
image
image

The final image with the "choose categories" behaves a little differently by not swapping out the title, but in the same vein, I feel the user experience is equivalent and we gain consistency through the site.

eddieferrer:
> @eddieferrer There are visual changes to the lightboxes in autolending settings but I feel that they still convey the same hierarchy and are now consistent with the lightboxes throughout the site.
> 
> Before:
> image
> image
> image
> 
> After:
> image
> image
> image
> 
> The final image with the "choose categories" behaves a little differently by not swapping out the title, but in the same vein, I feel the user experience is equivalent and we gain consistency through the site.

I understand, it wasn't clear to me that the visual changes were intentional in this PR.

I would suggest then that we run this change by design.

ryan-ludwig:
> 
> I understand, it wasn't clear to me that the visual changes were intentional in this PR.
> 
> I would suggest then that we run this change by design.

👍  will do

ryan-ludwig:
Got the go-ahead from design to make these visual changes to autolending settings dialogs. Thanks @eddieferrer for reaching out as well.

PR #1778, opened 2020-06-02
GROW-111, GROW-110, ML loan recommendations in the interstitial & funded pages

ryan-ludwig:


go-bears:
ooh! 😍

src/components/LoansYouMightLike/lymlContainer.vue
```diff
@@ -257,6 +313,33 @@ export default {
...
+				id: 'Experiment:ml_loan_to_loan',
+				fragment: experimentVersionFragment,
+			}) || {};
+
+			if (localExperiment.version === 'control' || localExperiment.version === 'shown') {
+				this.expMlLoanToLoan = localExperiment.version === 'shown';
+			}
+		},
+		fireExperimentTracking() {
+			// track loans shown
```
ryan-ludwig, line 329: Will remove these console logs after I've had some additional testing in dev

mcstover: cool totally understand the interest in testing this where that query can run properly

src/components/Lightboxes/AddToBasketInterstitial.vue
```diff
@@ -93,50 +96,70 @@ import cookieStore from '@/util/cookieStore';
...
 	},
 	computed: {
 		basketInterstitialActive() {
 			return this.basketInterstitialState.active || false;
 		}
 	},
+	watch: {
+		userPrefHideInterstitial() {
+			try {
+				localStorage.setItem('userPrefHideInterstitial', this.userPrefHideInterstitial);
```
emuvente, line 136: Use the `store2` package rather than using local storage directly. When you do that, there's no need for the try/catch

src/components/Lightboxes/AddToBasketInterstitial.vue
```diff
@@ -93,50 +96,70 @@ import cookieStore from '@/util/cookieStore';
...
-				this.showInterstitial = interstitialState.active ? interstitialState.visible : false;
-
 				this.basketInterstitialState = {
 					...this.basketInterstitialState,
 					active: interstitialState.active,
 					visible: interstitialState.visible,
 					loanId: interstitialState.loanId,
 				};
 
+				this.userPrefHideInterstitial = localStorage.getItem('userPrefHideInterstitial') === 'true';
```
emuvente, line 160: Because of the `watch` on this property, setting this here will cause the tracking event to fire. I would recommend tracking/setting the value in storage using a click handler, rather than a watch.

src/components/Lightboxes/AddToBasketInterstitial.vue
ryan-ludwig: Ooh, it actually stores the value as a Boolean instead of String. 😲

PR #1781, opened 2020-06-02
VUE-308 Subscription Settings Mutations

eddieferrer:
* Adds mutations for modifying MG parameters in lightbox and updating status card on subscription settings page
* Adds mutations for cancelling subscriptions
* Adds lightbox to prompt for confirmation of cancelling subscription

VUE-308

### DIFF
--- a/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
+++ b/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
@@ -27,7 +27,7 @@
 						>{{ dayOfMonth | numeral('Oo') }}</a> of each month <a
 							role="button"
 							@click.prevent="showLightbox = true;"
-						>${{ autoDepositAmount }}</a> will be
+						>{{ totalCombinedDeposit | numeral('$0,0.00') }}</a> will be
 						transferred from PayPal <a
 							role="button"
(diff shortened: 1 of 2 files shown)


src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -333,6 +326,12 @@ export default {
 		});
 	},
 	computed: {
+		selectedGroupDescriptor() {
+			// eslint-disable-next-line max-len
+			const selectedCategory = this.lendingCategories.find(category => category.value === this.category) || {};
+			// Sanitize and set initial form values.
+			return selectedCategory ? selectedCategory.shortName : 'all borrowers';
```
mcstover, line 333: @eddieferrer Careful with this one. I think there is an additional condition in the legacy page that just skips setting this value (and saving this value) if the category in the lenders account doesn't match one of the provided enums. This is setup in the `_serializeMGCategoryOptions` method of AutoDepositView.js and accounts for lenders that have specialized criteria from outside the normal flow.

mcstover: We have a `Preserve existing settings` option for those lenders with an empty value ''.

PR #1784, opened 2020-06-02
GROW-109 Add population property handling and tests for experiment resolvers

emuvente:


src/api/localResolvers/experiment.js
```diff
@@ -25,24 +26,40 @@ export default () => {
...
+								population !== currentAssignment.population
+								&& currentAssignment.version === 'unassigned'
+							)
+						)
+					) {
 						// assign the version using the experiment data (undefined if experiment disabled)
 						currentAssignment = {
 							id,
-							version: assignVersion(experiment || {}),
+							version: assignVersion(experiment),
```
mcstover, line 60: I think we need to ensure an object is passed here if the experiment value is null. Otherwise we get breakage when trying to pull items off the object inside `assignVersion`

emuvente: Since the if statement checks `experiment !== null && experiment.enabled`, there's no way to get a null/undefined experiment value in here, and it was coming up in code coverage as an uncovered branch.

mcstover: 👍

PR #1786, opened 2020-06-03
VUE-308 Subscription Settings Additional UI Behavior

eddieferrer:
* Cleaned up CSS
* Added 'are you sure you' warnings
* Added save button on subscriptions page to save settings when changes are made in the modal and the modal is closed.

VUE-308

### DIFF
--- a/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
+++ b/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
@@ -42,7 +42,7 @@
 						class="monthly-good-settings-lightbox"
 						:visible="showLightbox"
 						title="Change your monthly good"
-						@lightbox-closed="showLightbox = false"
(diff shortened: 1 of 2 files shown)


src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -348,6 +348,12 @@ export default {
 		}
 	},
 	methods: {
+		closeLightbox() {
+			// this will trigger only when the lightbox is closed manually
```
ryan-ludwig, line 352: Looking at KvLightbox, @lightbox-closed should fire on outside click, esc, or close button.

eddieferrer: Good catch @ryan-ludwig this is working as intended, but the comment is incorrect. I will update this comment in the next PR

PR #1790, opened 2020-06-04
VUE-308 Subscriptions Settings Navigation Guard and isAutoSubscriber …

eddieferrer:
…check

* Added login redirect for subscription settings if not logged in
* Added more accurate check for isAutoDepositSubscriber

VUE-308

### DIFF
--- a/src/pages/Subscriptions/SubscriptionsSettingsPage.vue
+++ b/src/pages/Subscriptions/SubscriptionsSettingsPage.vue
@@ -89,6 +89,13 @@ const pageQuery = gql`{
 			isSubscriber
 			isOnetime
 		}
+		lastLoginTimestamp @client
+	}
+	general {
+		activeLoginDuration: configSetting(key: "login_timeouts.www.active_login") {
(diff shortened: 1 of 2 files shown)


src/pages/Subscriptions/SubscriptionsSettingsPage.vue
```diff
@@ -113,9 +120,35 @@ export default {
...
 	apollo: {
 		query: pageQuery,
-		preFetch(config, client) {
-			return client.query({
-				query: pageQuery
+		preFetch(config, client, { route, kvAuth0 }) {
+			return new Promise((resolve, reject) => {
+				client.query({
+					query: pageQuery
+				}).then(({ data }) => {
```
eddieferrer, line 127: This is copied over from the autolending settings page. As an aside, I really dislike this method of checking for authentication in components. Authentication checking should be handled by the router in navigation guards. I believe we are only doing this in 3 places at the moment, but I might tackle this as a refactor in the near future

PR #1808, opened 2020-06-09
moving data call to prefetch to prevent awkward loading state

BoulderBrains:


src/pages/Donate/DonateFromMacro.vue
```diff
@@ -53,29 +59,26 @@ export default {
...
-
-	created() {
-		this.federation.query({
-			query: contentful,
-			variables: {
-				contentType: 'page',
-				contentKey: 'support-kiva',
-			}
-		}).then(({ data }) => {
+	inject: ['apollo', 'federation'],
```
mcstover, line 62: don't need `federation` any longer

PR #1829, opened 2020-06-17
Lint inline graphql queries

emuvente:


emuvente:
All the linting error and warnings addressed in this pr can be seen here: https://travis-ci.org/github/kiva/ui/builds/699468074

PR #1830, opened 2020-06-17
WRD Homepage Takeover Driven by contentful setting

eddieferrer:
* Uses a contentful setting to drive homepage content

LW-1956

### DIFF
--- a/src/pages/Homepage/DefaultHomepage.vue
+++ b/src/pages/Homepage/DefaultHomepage.vue
@@ -1,6 +1,5 @@
 <template>
 	<div>
-		<top-message-contentful />
 		<hero-slideshow
 			v-if="showSlideShow && !isExperimentActive"
 			:promo-enabled="promoEnabled"
@@ -39,7 +38,6 @@ import MonthlyGoodExplained from '@/components/Homepage/MonthlyGoodExplained';
 import MGCovidExplained from '@/pages/LandingPages/MGCovid19/MGCovidExplained';
 import MGCovidHero from '@/pages/LandingPages/MGCovid19/MGCovidHero';
 import CategoryGrid from '@/components/Homepage/CategoryGrid';
-import TopMessageContentful from './TopMessageContentful';
 import { processContent } from '@/util/contentfulUtils';
 
 const pageQuery = gql`{
@@ -59,7 +57,6 @@ export default {
 		CategoryGrid,
 		MGCovidExplained,
 		MGCovidHero,
-		TopMessageContentful,
 	},
 	data() {
 		return {
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -24,7 +24,7 @@ import TopMessageContentful from './TopMessageContentful';
 
 const activePageQuery = gql`query homepageFrame {
 	contentful {
-		entries(contentType: "uiSetting", contentKey: "ui-homepage-top")
+		entries(contentType: "uiSetting", contentKey: "ui-homepage-takeover")
 	}
 }`;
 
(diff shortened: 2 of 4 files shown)


eddieferrer:
In my dev vm I get: 
vendors~app.js:99356 [Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content. This is likely caused by incorrect HTML markup, for example nesting block-level elements inside <p>, or missing <tbody>. Bailing hydration and performing full client-side render.

I burned some time trying to fix this issue and got nowhere. Everything seems to be working, the issue seems to point to the kvResponsiveImage component. I tried setting cache keys on it, and modifying the selection of the homepage component in homepage.vue several ways, and nothing would make it go away, any suggestions there are appreciated. 

I dont think this warning is an issue though, and the homepage takeover will still work.

src/pages/Homepage/DefaultHomepage.vue
```diff
@@ -1,6 +1,5 @@
 <template>
 	<div>
-		<top-message-contentful />
```
eddieferrer, line 3: I don't believe this should be here. This component doesnt even render without content prop

src/pages/Homepage/Homepage.vue
```diff
@@ -72,26 +69,33 @@ export default {
...
+			const contentSetting = _get(data, 'contentful.entries.items', []).find(item => item.fields.key === 'ui-homepage-takeover'); // eslint-disable-line max-len
+			if (_get(contentSetting, 'fields')) {
 				const isContentEnabled = settingEnabled(
 					contentSetting.fields,
 					'active',
 					'startDate',
 					'endDate'
 				);
 				if (isContentEnabled) {
+					const takeOverKey = _get(contentSetting, 'fields.dataObject.takeOverKey');
```
eddieferrer, line 82: I think this is a cleaner way to pick 1 homepage takeover design based on a single setting in contentful

src/pages/Homepage/iwd/IWDHomepage.vue
```diff
@@ -342,7 +341,6 @@ export default {
 		NumberCountries,
 		WavyDivider,
 	},
-	inject: ['apollo'],
```
eddieferrer, line 345: Not needed

src/pages/Homepage/iwd/IWDHomepage.vue
```diff
@@ -11,7 +11,6 @@
 		>
 			<kv-responsive-image
 				:images="heroImageSet"
-				alt=""
```
eddieferrer, line 14: Not needed, kv responsive wont even pass attributes down.

mcstover:
@eddieferrer Turning of JS reveals an interesting comparison of rendered html. Without JS, the clickable `.hero` anchor tag is a sibling to the `.hero__text-wrapper` and `svg`... However, the Vue rendered html places `.hero__text-wrapper` and `svg` within the anchor...

The mismatch still exists even with the responsive image component removed...

Certainly not ideal but also not a blocker for such a tight timeline...

mcstover:
👍 +1

src/pages/Homepage/iwd/IWDHomepage.vue
ryan-ludwig: @mcstover @eddieferrer A small thing, but we should always have some alt tag on images. `alt=""` tells a screenreader that this is a decorative image and to ignore it. Having no alt attribute [makes some screenreaders announce the image URL](https://www.w3.org/WAI/tutorials/images/decorative/). 

kv-responsive-image v-binds $attrs to the `img` inside. We could add a non-required alt prop, and if there's no prop set alt to "" on the img inside. Though I kindof like reinforcing the always alt pattern so it's remembered for regular images.

mcstover: This is a good callout. I think we just got caught up in the rush.

eddieferrer: @ryan good point, I was incorrect about my comment above. Ill revert this

PR #1836, opened 2020-06-18
new donate hero images

BoulderBrains:
These are the changes that will be needed to replace the hero images, however there are issues with small image sizes and a missing image size. Going to track down those.


BoulderBrains:
- New images
- Styles finessed a little bit
- Verified locally across all screen sizes.

PR #1837, opened 2020-06-18
Hotfix wrd 2020

mcstover:


mcstover:
FYI @eddieferrer Verified with a compiled run in the VM.

src/pages/Homepage/iwd/IWDHomepage.vue
```diff
@@ -11,7 +11,6 @@
 		>
 			<kv-responsive-image
 				:images="heroImageSet"
-				alt=""
```
ryan-ludwig, line 14: A small thing, but we should always have some alt tag on images. `alt=""` tells a screenreader that this is a decorative image and to ignore it. Having no alt attribute [makes some screenreaders announce the image URL](https://www.w3.org/WAI/tutorials/images/decorative/). 

For kv-responsive-image we could add a non-required alt prop, and if there's no prop set alt to "" on the img inside. Though I kindof like reinforcing the always alt pattern so it's remembered for regular images. Don't feel strongly.

PR #1843, opened 2020-06-22
VUE-336 Promo Banner Border Bug

eddieferrer:
* Moved CSS for promo banner small border so that border is not visible if banner is not visible. (It creates a problem for homepage takeovers with white headers)

VUE-336


ryan-ludwig:
👍

PR #1845, opened 2020-06-22
Added missing alt tags to kv responsive components

eddieferrer:


src/pages/Possibility/KivaContentBlock.vue
```diff
@@ -28,7 +28,7 @@
 					</div>
 				</div>
 				<div class="columns large-5 show-for-large">
-					<kv-responsive-image :images="images" />
+					<kv-responsive-image :images="images" alt="Images of possibility" />
```
ryan-ludwig, line 31: This can be alt="" since "images of possibility" doesn't provide much useful context to a user. We'd have to pass it in with the images array if there was something we really wanted to call out.

src/pages/Possibility/KivanSlider.vue
```diff
@@ -28,10 +28,10 @@
 		>
 			<div class="row">
 				<div class="small-12 large-6">
-					<kv-responsive-image :images="slide.left" />
+					<kv-responsive-image :images="slide.left" alt="Images of possibility" />
```
ryan-ludwig, line 31: Same as comment above

PR #1846, opened 2020-06-22
Added missing alt tags to kv reponsive components

eddieferrer:


eddieferrer:
This is just a fix on the previous commit.

PR #1852, opened 2020-06-23
K8 setup test 1

mcstover:
This is the first of several transfer PRs to cover incremental config updates.

### DIFF
--- a/Dockerfile.jenkins
+++ b/Dockerfile.jenkins
@@ -2,6 +2,10 @@
 FROM node:12-alpine as build-stage
 RUN apk add --no-cache git
 RUN apk add --no-cache bash
+RUN apk add --no-cache tini
+
+# Use Tini for entrypoint, available at /sbin/tini
+ENTRYPOINT ["/sbin/tini", "--"]
 
 RUN mkdir /ui
 COPY ./ /ui/
--- a/Dockerfile.jenkins
+++ b/Dockerfile.jenkins
@@ -2,10 +2,6 @@
 FROM node:12-alpine as build-stage
 RUN apk add --no-cache git
 RUN apk add --no-cache bash
-RUN apk add --no-cache tini
-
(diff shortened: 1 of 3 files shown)


deploy/test/values.yaml
```diff
@@ -6,11 +6,17 @@ image:
   pullPolicy: Always
 
 deployenv:
-  environment: "" # set either by jenkins or manually
+  environment: "k8stest" # injected into deployment template
 
 secretName: kiva-ui
 
 replicaCount: 2
+# minReplicas: 2
```
mcstover, line 14: the future!

deploy/charts/templates/deployment.yaml
```diff
@@ -29,8 +29,8 @@ spec:
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}" # In Jenkins, need to call helm with --set image.tag=master-XY
-        command: ["npm"]
-        args: ["start", "--", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}", "--prefix", "/ui"]
+        command: ["tini"]
+        args: ["--", "npm", "start", "--", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}"]
```
mcstover, line 33: K8s calls these when starting the container. `command` here maps to and overrides `ENTRYPOINT` in Dockerfile.jenkins. Likewise, `args` here overrides `CMD` in the Dockerfile.

Dockerfile.jenkins
```diff
@@ -12,6 +12,11 @@ RUN npm run build
 
 # run stage
 FROM node:12-alpine as run-stage
+RUN apk add --no-cache tini
```
mcstover, line 15: Now we have `tini` available in the built container

PR #1854, opened 2020-06-23
Add cpu and memory limits and resource requests. Add autoscaler and c…

mcstover:
…onfig to operate based on average memory usage. yaml file editorconfig.

Fix template name to be ui-helm instead of helm-config

use proper resource check for memory in autoscaler

Fix value suffix for Utilization averageValue

tuning resource request/limit settings and autoscale averageValue

bump average auto scaling memory utilization

Fix autoscale type to match value settings. Add requests/limits to deployment

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,7 @@ indent_size = 4
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2
--- a/deploy/charts/templates/autoscale.yaml
+++ b/deploy/charts/templates/autoscale.yaml
@@ -1,0 +1,25 @@
+{{- $fullName := include "ui-helm.fullname" . -}}
+apiVersion: autoscaling/v2beta2
+kind: HorizontalPodAutoscaler
+metadata:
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -7,3 +7,7 @@ indent_size = 4
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*.{yml,yaml}]
+indent_style = space
+indent_size = 2
```
mcstover, line 13: So much easier to deal with yaml now!

deploy/charts/templates/autoscale.yaml
```diff
@@ -0,0 +1,25 @@
...
+    name: {{ include "ui-helm.fullname" . }}
+  minReplicas: {{ .Values.minReplicas }}
+  maxReplicas: {{ .Values.maxReplicas }}
+  metrics:
+    - type: Resource
+      resource:
+        name: memory
+        target:
+          type: AverageValue
+          averageValue: 320Mi
```
mcstover, line 25: When the average memory consumption reaches this point, scale up. _This is base don our current app running in cluster mode which has 2 internal node processes. Cluster will soon be removed._

PR #1855, opened 2020-06-23
K8 setup test 4

mcstover:


deploy/charts/values.yaml
```diff
@@ -29,6 +29,8 @@ ingress:
 configmap:
   name: kiva-ui
   all:
+  - key: NODE_ENV
+    value: "production"
```
mcstover, line 33: All deployments now get this environment variable by default

deploy/charts/templates/deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}" # In Jenkins, need to call helm with --set image.tag=master-XY
         command: ["tini"]
-        args: ["--", "npm", "start", "--", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}"]
+        args: ["--", "node", "server/index.js", "--port={{ .Values.service.port }}", "--config={{ .Values.deployenv.environment }}"]
```
mcstover, line 33: NODE_ENV is now provided globally and we can start the app directly. This removes the intermediary `npm` process and allows Terminus to exit the node process directly. This prevents a lingering `npm` process which would fail when it lost track of the app it "remembered" starting.

server/index.js
```diff
@@ -82,29 +83,11 @@ app.use(logger.errorLogger);
 // Final Error Handler
 app.use(logger.fallbackErrorHandler);
 
-// Cluster Activation
-// See: https://nodejs.org/docs/latest-v8.x/api/cluster.html
+// initialize http server instance
+const server = http.createServer(app);
```
mcstover, line 87: We have to pass a server instance to terminus rather than the express app instance.

server/index.js
```diff
@@ -82,29 +83,11 @@ app.use(logger.errorLogger);
...
-	});
-} else {
-	// Start the worker processes
-	// - these can share any TCP connection
-	console.log(`Worker ${process.pid} started`); // eslint-disable-line
-
-	app.listen(port, () => console.log(`server started at localhost:${port}`));
-}
+// listen for requests
+server.listen(port, () => console.log(`server (pid: ${process.pid}) started at localhost:${port}`));
```
mcstover, line 93: The internal cluster flow is no more.

server/util/terminusConfig.js
```diff
@@ -0,0 +1,53 @@
...
+		}
+		resolve(true);
+	});
+}
+
+function onSignal() {
+	console.log('Server is starting cleanup');
+	return Promise.all([
+		// TODO: add any additional shutdown methods
+		disconnectCache()
```
mcstover, line 18: Anything else we should handle here?

server/index.js
```diff
@@ -82,29 +84,40 @@ app.use(logger.errorLogger);
 // Final Error Handler
 app.use(logger.fallbackErrorHandler);
 
-// Cluster Activation
-// See: https://nodejs.org/docs/latest-v8.x/api/cluster.html
+if (config.server.disableCluster) {
```
mcstover, line 87: Temporarily gate cluster mode with environment var.

deploy/charts/templates/deployment.yaml
eddieferrer: I wonder if this is the same issue that causes a failed 'npm run test:e2e' to leave ghost npm processes as well.

deploy/charts/templates/deployment.yaml
```diff
@@ -42,3 +42,15 @@ spec:
               name: {{ .Values.configmap.name }}
         resources:
           {{- toYaml .Values.resources | nindent 12 }}
+        readinessProbe:
+          httpGet:
+            path: /healthcheck
```
leoperry, line 47: I'm not 100% aware of what the Terminus healtcheck actually does, so I'll just provide some [generic feedback](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) here.

Readiness probes are used to determine when a pod is ready to accept traffic (meaning, capable of actually serving the request in full). It's checked throughout the life of the pod (so that pods could temporarily stop accepting traffic without needing a restart), and is allowed to fail `failureThreshold` (default 3) times in a row before being restarted. Generally you want to set your initial delay conservatively. Readiness probes often check dependencies, so you might want to extend the `periodSeconds` timeout longer if a dependency check might take longer at times due to temporary latency.

Liveness probes are used to determine whether to restart a pod after everything starts. Generally the probe determines whether the pod can be considered responsive, and has a similar `failureThreshold` mechanism as readiness probes. Often liveness probes don't check dependencies, instead preferring to have a lightweight and reliable check. Again, we may want to consider setting delays and timeouts conservatively to avoid excessive restarts if something is temporarily off.

The dependencies of the app may not imply a significant difference between "service is ready to accept traffic" and "service is alive" so the same health check may be entirely appropriate.

deploy/charts/templates/deployment.yaml
leoperry: This seemed to be a key to managing node app shutdown from various resources I saw... but it also seemed nontrivial to implement, so 👍.

deploy/charts/templates/autoscale.yaml
```diff
@@ -22,4 +22,4 @@ spec:
         name: memory
         target:
           type: AverageValue
-          averageValue: 320Mi
+          averageValue: 150Mi
```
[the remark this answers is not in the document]
mcstover: Good call I'll work on that in a future PR. This would make it easier to experiment with different values across environments or even choose to run fewer/more instances...

So far the instances are pretty stable in their resource usage which is good to see!

deploy/charts/templates/deployment.yaml
mcstover: I've def read through the docs on this and we could consider adding more nuance at some point.  For now I decided to just reuse a single endpoint that represents a minimal request/response cycle. 

The terminus lib allows us to create our own methods (aka. complexity) for these checks. However, the recurrence of these checks in such short succession means we need to make sure they are super efficient.

As for what either of these mean in the Ui app, we are generally "ready" when we can serve requests. That said, a couple of places we could get more diligent about would be to check if the graphql schema has been fetched and  check to verify if the cache connection is active. Now, what value do we get from these checks? If we're checking/hitting connection to cache every 5-10 seconds? If the cache is not available for 1 instance it's likely not available for any of them...

The settings here are conservative based on observation in start up and readiness of the pods.

deploy/charts/templates/autoscale.yaml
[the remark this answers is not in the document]
mcstover: ref. https://github.com/kiva/ui/pull/1864

PR #1864, opened 2020-06-24
make autoscale, readiness and liveliness probes configurable by envir…

mcstover:
…onment. test cpu based scaling in the test env.

adjust cpu utilization

Minor tweaks to resource configs, lots of commentary on default values.


mcstover:
@leoperry @emuvente Made various items configurable per environment and added commentary about how these configurable values are used.

PR #1869, opened 2020-06-25
Add Select/Deselect All in country selection UI

remoun:
https://kiva.atlassian.net/browse/AUTO-43

country-select-all

### DIFF
--- a/src/pages/Autolending/CheckList.vue
+++ b/src/pages/Autolending/CheckList.vue
@@ -1,18 +1,32 @@
 <template>
-	<ul class="check-list" :class="{ 'use-columns': useColumns }">
-		<li
-			:key="id"
-			v-for="{ id, name, selected } in items"
+	<div>
+		<button
+			v-if="items.some(x => !x.selected)"
+			@click="emitAll(true)"
 		>
-			<kv-checkbox
-				:id="`${name}-${id}` | changeCase('paramCase')"
-				:checked="selected"
-				@change="$emit('change', $event, id);"
+			Select All
+		</button>
+		<button
+			v-if="items.some(x => x.selected)"
+			@click="emitAll(false)"
+		>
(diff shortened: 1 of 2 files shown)


src/pages/Autolending/CheckList.vue
```diff
@@ -1,18 +1,32 @@
...
+			@click="emitAll(true)"
 		>
-			<kv-checkbox
-				:id="`${name}-${id}` | changeCase('paramCase')"
-				:checked="selected"
-				@change="$emit('change', $event, id);"
+			Select All
+		</button>
+		<button
+			v-if="items.some(x => x.selected)"
```
eddieferrer, line 10: This works, but it would be better style to put the array method in a computed property and check for that value in the v-if.

remoun:
> Per the design in AUTO-43, these buttons should be blue to indicate that they are clickable.

Done.


> The state where 1 option is checked is confusing. It reads as "Select All Deselect All". My preference here would be to simplify the UI and not include this state.

After playing with it some more, I feel it's more confusing to have the buttons appear and disappear, so I'm inclined to keep them unconditionally visible. I proposed that in [AUTO-43](https://kiva.atlassian.net/browse/AUTO-43). How does that sound to you?

eddieferrer:
> > Per the design in AUTO-43, these buttons should be blue to indicate that they are clickable.
> 
> Done.
> 
> > The state where 1 option is checked is confusing. It reads as "Select All Deselect All". My preference here would be to simplify the UI and not include this state.
> 
> After playing with it some more, I feel it's more confusing to have the buttons appear and disappear, so I'm inclined to keep them unconditionally visible. I proposed that in [AUTO-43](https://kiva.atlassian.net/browse/AUTO-43). How does that sound to you?

Unconditionally visible is fine with me. 👍

PR #1872, opened 2020-06-27
Simple log formatter and well known log excludes.

mcstover:


mcstover:
@emuvente @eddieferrer this includes the basic formatter and 2 samples. The api.auth exclusion will clean up our server logs a ton!

once this is in I'll go through and start converting all uses.

PR #1878, opened 2020-07-01
Update log formatting for server and util files.

mcstover:


mcstover:
FYI @eddieferrer @ryan-ludwig Doing some log cleanup to match the format of logs collected by winston and to prevent multiline logs for various errors.

PR #1880, opened 2020-07-01
Fixed issue with auth pop up login not redirecting on preFetch rejection

eddieferrer:
When preFetchAll is used outside of apollo mixin (by component methods) we need to send the rejection to the router.

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -427,7 +427,9 @@ export default {
 							route: this.$route,
 							kvAuth0: this.kvAuth0,
 						}).catch(error => {
-							this.$router.push(error);
+							if (error.path) {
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -426,6 +426,8 @@ export default {
 						return preFetchAll(matched, this.apollo, {
 							route: this.$route,
 							kvAuth0: this.kvAuth0,
+						}).catch(error => {
+							this.$router.push(error);
```
emuvente, line 430: We should check that the error is actually a route before sending it to router.push, either by checking `error instanceof Error` or `error.url`

eddieferrer: Added a check for error.path.

PR #1881, opened 2020-07-01
GROW-118 Dynamic loan urls and images from Ml recommender

ryan-ludwig:
Couple of things to iron out still, but ready for some review.

To demo:
- checkout this branch, 
- change your config/dev-vm.js graphql endpoint in config to point to federation.
- restart your UI server,
- Open your browser and hit
https://dev-vm-01.kiva.org/live-loan/u/234/img/1 
https://dev-vm-01.kiva.org/live-loan/u/234/url/1 

A few comments on the files below


server/live-loan-router.js
```diff
@@ -0,0 +1,125 @@
...
+	return new Promise((resolve, reject) => {
+		const endpoint = config.app.graphqlUri;
+		let query;
+		if (type === 'url') {
+			query = `{
+				ml {
+					recommendationsByLoginId(
+						segment: all
+						  loginId: ${loginId}
+						  offset: ${offset}
```
ryan-ludwig, line 19: Note, I have not been able to make any offset other than 0 work in the graphQl query. Will check in with the Ml team.

server/live-loan-router.js
```diff
@@ -0,0 +1,125 @@
...
+	userRouter.use('/:userId/img', imgRouter);
+
+	// URL Router
+	urlRouter.get('/', (req, res) => { res.sendStatus(500); });
+	urlRouter.get('/:offset', async (req, res) => {
+		const { userId, offset } = req.params;
+		try {
+			const loanData = await getRecommendedLoans('url', userId/* , offset */);
+			const offsetLoanId = loanData[0].id;
+			res.redirect(302, `/lend/${offsetLoanId}`);
```
ryan-ludwig, line 100: Do we want to do any caching with the Url redirect?

server/live-loan-router.js
```diff
@@ -0,0 +1,125 @@
...
+		} catch (err) {
+			console.error(err);
+			res.redirect(302, '/lend-by-category/');
+			// res.sendStatus(500); // TODO, do we want 500 error here instead?
```
ryan-ludwig, line 104: Do we want 500 error here instead?

server/live-loan-router.js
```diff
@@ -0,0 +1,125 @@
...
+	});
+
+	// IMG Router
+	imgRouter.get('/', (req, res) => { res.sendStatus(500); });
+	imgRouter.get('/:offset', async (req, res) => {
+		const { userId, offset } = req.params;
+		try {
+			const loanData = await getRecommendedLoans('img', userId/* , offset */);
+			const loanCardImgBuffer = await drawLoanCard(loanData[0]); // TODO: offset?
+			// TODO: Add loanCard img to memcache?
```
ryan-ludwig, line 115: Add loanCard img to memcache here?

src/plugins/loan-use-filter.common.js
```diff
@@ -0,0 +1,69 @@
+const numeral = require('numeral');
```
ryan-ludwig, line 1: This file is a duplicate of src/plugins/loan-use-filter.js but with module.exports to make node 12 happy.

server/util/live-loan/live-loan-draw.js
```diff
@@ -0,0 +1,130 @@
...
+				strokeGrey: '#d8d8d8',
+				textDark: '#484848',
+				white: '#ffffff'
+			};
+
+			function fontFile(name) {
+				return path.join(__dirname, '../../../src/assets/fonts', name);
+			}
+			registerFont(fontFile('PostGrotesk-Medium.ttf'), { family: 'Kiva Post Grot', weight: '400' });
+			registerFont(fontFile('PostGrotesk-MediumItalic.ttf'), { family: 'Kiva Post Grot', weight: '400', style: 'italic' });
```
ryan-ludwig, line 28: TODO: Having issues with node-canvas and registering fonts. So far only the first font is able to be used. Still working on this but I don't think it'll be a dealbreaker.

server/live-loan-router.js
mcstover: You could also consider just fetching like 3 loans, then filtering and serving accordingly

server/live-loan-router.js
mcstover: I don't think so... that would risk reusing loans that may have been removed from the getRecommendedLoans query.

server/live-loan-router.js
mcstover: I think you'll need to throw and error before the res.redirect in the try block in order to get down to this catch...maybe if 0 loans are returned... I don't see how you could get a loan back and not have an id to trigger this fail...

server/live-loan-router.js
mcstover: We need to come up with a system to reduce the number of images that might be cached. As I mentioned before, rounding the currently funded about to the nearest 20th percent would be a good start. Then you could have cache keys for images like so... `live-loan-##loanid##-0`, `live-loan-##loanid##-20`, `live-loan-##loanid##-40` etc...

This would at least give us a baseline of measurement to ensure we have caching instances that can hold enough entries.

mcstover: Then your process will be more like...
`
const loanCardCacheKey = live-loan-##loanid##-0
let cachedLoanCardImage = retrieveFromCache(live-loan-##loanid##-0)
if (!cachedLoanCardImage) {
  cachedLoanCardImage = await drawLoanCard()
}
... res.send(cachedLoanCardImage);
`

server/live-loan-router.js
ryan-ludwig: Ml team is working on offset. Faking it for now using by setting limit to the offset number, then using the last loan from the result.

server/live-loan-router.js
```diff
@@ -0,0 +1,125 @@
...
+	});
+}
+
+module.exports = function liveLoanRouter() {
+	const router = express.Router();
+	const userRouter = express.Router({ mergeParams: true });
+	const urlRouter = express.Router({ mergeParams: true });
+	const imgRouter = express.Router({ mergeParams: true });
+
+	router.get('/', (req, res) => { res.sendStatus(500); });
```
emuvente, line 85: 500 should only be sent in genuine server failure cases. For these routes that you aren't doing anything except sending the 500 status, I would recommend just not including a handler at all. That will let the route fall through to the generic 404.

emuvente:
To ensure that the data is accurate, I would recommend against using different graphql queries for the url and img handling. There should be a single get recommended loans call that is cached (old dynamic loan code would cache this data for 30 minutes, but I could see 10 minutes or less being ok too) and used by both the url and img handlers, as that prevents the situation of the url query getting a different loan than the image query.

server/live-loan-router.js
emuvente: We could also just cache one image for a loan, and use a ttl rather than a cache key based on the loan progress. (The cached value could be json, allowing you to store and check the time it was cached)

src/plugins/loan-use-filter.common.js
emuvente: You could also just change loan-use-filter.js. It should still work when using `module.exports` instead of `export`.

server/live-loan-router.js
emuvente: No caching here, but see my comment about caching the getRecommendedLoans query itself

server/live-loan-router.js
emuvente: Do we even need to use the offset at all in this query? Couldn't we always fetch 4 loans, cache that result, and then use the offset argument to pick which of those loans in the cache to return? Fundraising loans are already going to be filtered out by the ML endpoint, so we shouldn't be in a situation where we need to fetch more than the initial amount of loans.

server/live-loan-router.js
mcstover: @emuvente I'd be fine with that if we can build in a "delay" of sorts so that we're not re-generating the image for small changes in the funded amount. my interest is reduce re-generating the image repeatedly (aka load on the express object) and to prevent excessive cache entries for each loan...

emuvente: The ttl would handle that delay, as the image would not be regenerated while the cached image is young enough. It would also handle the excessive entries for each loan as only 1 image would ever be cached for a loan.

mcstover:
> To ensure that the data is accurate, I would recommend against using different graphql queries for the url and img handling. There should be a single get recommended loans call that is cached (old dynamic loan code would cache this data for 30 minutes, but I could see 10 minutes or less being ok too) and used by both the url and img handlers, as that prevents the situation of the url query getting a different loan than the image query.

With regard to caching the loan query...at first I was thinking we wouldn't want to...and that may still be the case... This is a user level query so what's the likely hood of the cache result ever getting used?

emuvente:
> This is a user level query so what's the likely hood of the cache result ever getting used?
It's going to be used at least once, because the query is called once for the image, and then a second time for the url when the user clicks on the image and needs to be redirected.

server/live-loan-router.js
ryan-ludwig: > Do we even need to use the offset at all in this query? Couldn't we always fetch 4 loans, cache that result, and then use the offset argument to pick which of those loans in the cache to return? Fundraising loans are already going to be filtered out by the ML endpoint, so we shouldn't be in a situation where we need to fetch more than the initial amount of loans.

Love it, now fetching 4 loans and caching the result to memcache for 30 minutes, then using offset to grab the right entry.

server/live-loan-router.js
ryan-ludwig: > 500 should only be sent in genuine server failure cases. For these routes that you aren't doing anything except sending the 500 status, I would recommend just not including a handler at all. That will let the route fall through to the generic 404.

Nice, router section got a lot simpler from this feedback.

server/live-loan-router.js
```diff
@@ -0,0 +1,150 @@
...
+				parsedData = JSON.parse(data);
+			}
+			resolve(parsedData);
+		});
+	});
+}
+
+function setLoansToCache(loginId, loans, cache) {
+	return new Promise((resolve, reject) => {
+		const cacheTime = 30 * 60; // 30 minutes
```
ryan-ludwig, line 30: @emuvente In my testing I tried setting this to 1 minute and did some additional logging. I was still getting cached results several minutes later. Ever tinkered with expiration times?

emuvente: it looks like you're supposed to pass the time as the expires property in an object for the third argument of `set()`, rather than just the time https://github.com/memcachier/memjs#settings-values

server/live-loan-router.js
```diff
@@ -0,0 +1,150 @@
...
+			res.redirect(302, '/lend-by-category/');
+		}
+	});
+
+	// IMG Router
+	router.use('/u/:userId/img/:offset', async (req, res) => {
+		const { userId, offset } = req.params;
+		try {
+			const loanData = await fetchRecommendedLoans(userId, cache);
+			const loanCardImgBuffer = await drawLoanCard(loanData[offset - 1]);
```
emuvente, line 135: Let's be sure to use the cache here as well for the buffer value so that we don't re-draw the same loan too often

server/util/live-loan/live-loan-draw.js
emuvente: What if you used a different font family name for each one?

ryan-ludwig: No luck there, I think I may need to patch the fonts or wait for a new release of node-canvas - https://github.com/Automattic/node-canvas/issues/1572

server/live-loan-router.js
ryan-ludwig: ahh, thank you!

server/live-loan-router.js
ryan-ludwig: @emuvente @mcstover To reiterate, we'll cache the generated image buffer to memcache for say 10(?) minutes, and serve that to anyone else who would see the same loan. We'll also add a max-age header for also 10(?) minutes to the image response.

ryan-ludwig:
Going to merge this and work on cache improvements next week as hotfixes.

PR #1887, opened 2020-07-08
Hide auto-lending donation controls and zero out donation if there is an auto-deposit donation

remoun:
Resolves https://kiva.atlassian.net/browse/AUTO-207

I couldn't tell why this component makes a direct mutation to update the `donationPercentage` (rather than go through the local resolver), but IIUC this change will pre-empt that code path, as the controls that modify `this.donation` are not rendered.

Screen Shot 2020-07-07 at 11 06 03 PM

### DIFF
--- a/src/pages/Autolending/AutolendingWhen.vue
+++ b/src/pages/Autolending/AutolendingWhen.vue
@@ -36,39 +36,44 @@
 				>
 					<div class="when-inputs-wrapper">
 						<lend-timing-dropdown />
-						<kv-radio
-							data-test="is-autolending-donation-on"
-							id="is-autolending-donation-on"
-							radio-value="on"
-							v-model="donationToggle"
-						>
-							Include a donation to Kiva of
-							<kv-dropdown-rounded v-model="donation">
-								<option :value="0">
-									0%
-								</option>
-								<option :value="5">
-									5%
-								</option>
-								<option :value="10">
-									10%
-								</option>
-								<option :value="15">
-									15%
-								</option>
-								<option :value="20">
-									20%
-								</option>
-							</kv-dropdown-rounded>
-						</kv-radio>
-						<kv-radio
-							data-test="is-autolending-donation-off"
-							id="is-autolending-donation-off"
-							radio-value="off"
-							v-model="donationToggle"
-						>
-							No donation to Kiva
-						</kv-radio>
+						<p v-if="hasAutoDepositDonation && donation === 0">
+							Your auto-deposit includes a donation, so auto-lending donations are disabled.
+						</p>
+						<div v-else>
+							<kv-radio
+								data-test="is-autolending-donation-on"
+								id="is-autolending-donation-on"
+								radio-value="on"
+								v-model="donationToggle"
+							>
+								Include a donation to Kiva of
+								<kv-dropdown-rounded v-model="donation">
+									<option :value="0">
+										0%
+									</option>
+									<option :value="5">
+										5%
+									</option>
+									<option :value="10">
+										10%
+									</option>
+									<option :value="15">
+										15%
+									</option>
+									<option :value="20">
+										20%
+									</option>
+								</kv-dropdown-rounded>
+							</kv-radio>
+							<kv-radio
+								data-test="is-autolending-donation-off"
+								id="is-autolending-donation-off"
+								radio-value="off"
+								v-model="donationToggle"
+							>
+								No donation to Kiva.
+							</kv-radio>
+						</div>
 					</div>
 					<template slot="controls">
 						<kv-button
@@ -128,6 +133,7 @@ export default {
 			lendAfterDaysIdle: 0,
 			donation: 15,
 			donationToggle: 'on',
+			hasAutoDepositDonation: false,
 		};
 	},
 	apollo: {
@@ -140,15 +146,21 @@ export default {
 					lendAfterDaysIdle
 				}
 			}
+			my {
+				autoDeposit {
+					donateAmount
+				}
+			}
 		}`,
 		preFetch: true,
 		result({ data }) {
 			this.isEnabled = !!_get(data, 'autolending.currentProfile.isEnabled');
-			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
 			this.lendAfterDaysIdle = _get(data, 'autolending.currentProfile.lendAfterDaysIdle');
 			this.isChanged = !!_get(data, 'autolending.profileChanged');
+			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
 			this.donation = _isFinite(donationPercentage) ? donationPercentage : 15;
 			this.donationToggle = this.donation !== 0 ? 'on' : 'off';
+			this.hasAutoDepositDonation = _get(data, 'my.autoDeposit.donateAmount') > 0;
 		},
 	},
 	mounted() {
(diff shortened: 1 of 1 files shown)


src/pages/Autolending/AutolendingWhen.vue
```diff
@@ -140,20 +146,30 @@ export default {
...
+			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
 			this.donation = _isFinite(donationPercentage) ? donationPercentage : 15;
 			this.donationToggle = this.donation !== 0 ? 'on' : 'off';
+			this.hasAutoDepositDonation = _get(data, 'my.autoDeposit.donateAmount') > 0;
 		},
 	},
 	mounted() {
 		// After initial value is loaded, setup watch
 		this.$watch('donation', (donation, previousDonation) => {
+			if (this.hasAutoDepositDonation) {
```
eddieferrer, line 169: If I understand correctly, the donation is getting zeroed out on the backend? If that is the case, we shouldn't need this code here. I would also prefer to avoid the param reassign.

src/pages/Autolending/AutolendingWhen.vue
```diff
@@ -176,7 +192,7 @@ export default {
 			if (donationToggle !== previousDonationToggle) {
 				if (donationToggle === 'on' && this.donation === 0) {
 					// set default donation if 0
-					this.donation = 15;
+					this.donation = this.hasAutoDepositDonation ? 0 : 15;
```
eddieferrer, line 195: Same comment as above. If the donation is getting zeroed out on the backend, then modifying these watches seems unnecessary since they won't even be shown.

src/pages/Autolending/AutolendingWhen.vue
```diff
@@ -140,20 +146,30 @@ export default {
...
+			}
 		}`,
 		preFetch: true,
 		result({ data }) {
 			this.isEnabled = !!_get(data, 'autolending.currentProfile.isEnabled');
-			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
 			this.lendAfterDaysIdle = _get(data, 'autolending.currentProfile.lendAfterDaysIdle');
 			this.isChanged = !!_get(data, 'autolending.profileChanged');
+			const donationPercentage = _get(data, 'autolending.currentProfile.donationPercentage');
 			this.donation = _isFinite(donationPercentage) ? donationPercentage : 15;
```
eddieferrer, line 161: A cleaner solution instead of modifying the watches, we can set this.donation to be 0 in the results block if hasAutoDepositDonation

eddieferrer:
I was also able to get into this situation: 

Screen Shot 2020-07-08 at 7 58 31 AM
Screen Shot 2020-07-08 at 7 58 13 AM

Where the UI is telling me I have a 10% donation, but it is impossible to change

src/pages/Autolending/AutolendingWhen.vue
remoun: Fair point. Reverted.

src/pages/Autolending/AutolendingWhen.vue
remoun: Reverted.

src/pages/Autolending/AutolendingWhen.vue
remoun: I did that initially, then realized that would make the UI misleading; I think we should render whatever the server returns.

remoun:
> I was also able to get into this situation:
> Where the UI is telling me I have a 10% donation, but it is impossible to change

Good catch! I updated the condition to `hasAutoDepositDonation && deposit === 0` in case the donation percentage is set. 

We might need to do some cleanup here. [AUTO-206](https://kiva.atlassian.net/browse/AUTO-206) tracks that; it seemed we didn't have any people at that intersection but I'll double-check.

mcstover:
@remoun I think the recent updates are on the right track. @eddieferrer has some comments and ideas coming your way to ensure new visitors – who may have the auto deposit donation – get their autolend donation zeroed out on save.

Once that is in place, I think we will have generally covered the majority of cases and get merge this. Then we'll want to test around other scenarios and existing flow.

src/pages/Autolending/AutolendingWhen.vue
```diff
@@ -36,39 +36,44 @@
...
-						</kv-radio>
-						<kv-radio
-							data-test="is-autolending-donation-off"
-							id="is-autolending-donation-off"
-							radio-value="off"
-							v-model="donationToggle"
-						>
-							No donation to Kiva
-						</kv-radio>
+						<p v-if="hasAutoDepositDonation && donation === 0">
```
eddieferrer, line 39: @remoun we should remove this `donation === 0` as new users who have AutoDepositDonation will not have a donation of 0

remoun: It seems better to tweak our initialization code to not set auto-lending donations when there's an auto-deposit donation -- in that indeterminate state, incorrectly hiding the edit UI seems worse; plus when users actually attempt to set the donationPercentage it would get reset to 0. I believe I've handled all the call sites for https://kiva.atlassian.net/browse/AUTO-44.

mcstover: Hiding the additional donation Ui if they have a donation set elsewhere is the correct path. It's highly likely that the new lender would have > 0 default percentage set so they will see a Ui that they are ineligible with the code above.

I think you could potentially solve this in the initialization phase by running an edit mutation in the mounted hook to zero out the donation in client state. If the visitor chooses to complete set up and "save" their auto-lend profile, their donation would then be stored as 0. This would allow the above code to work for both scenarios.

remoun: Note that [Monthly Good setup explicitly sets `donationPercentage=0`](https://github.com/kiva/kiva/blob/c33c23a493ab1921c0b18184c0256a5fca975d50/libs/kv/lib/monthly_good.php#L74) and [so does `setAutolendingEnabled`](https://github.com/kiva/kiva/blob/c33c23a493ab1921c0b18184c0256a5fca975d50/httpdocs/classes/API/FundAccount.php#L559)

I started adding the mutation, but realized that that would just update the client-side state, and create a consistently misleading UI. So I'm inclined to keep it with the `hasAutoDepositDonation && donation === 0` to handle the default case and degrade gracefully for other combinations.

remoun: Added a comment summarizing this rationale.

mcstover: Ok I see more clearly where you are coming from now. Any case where this still occurs should be handled with additional messaging and state handling in a new ticket. We'll likely want to create a new ticket for how to handle lenders that arrive to this page who have autolending off but an amount already set on their autolend_profile too.

remoun: I'm pretty sure we're good in that case too. We reset the AL donation when we turn on auto-lending if they have an auto-deposit donation, as [that flow goes `setAutolendingEnabled`](https://github.com/kiva/kiva/blob/c33c23a493ab1921c0b18184c0256a5fca975d50/httpdocs/classes/API/FundAccount.php#L559).

While trying this out in my VM as an MG subscriber, `/settings/autolending` redirected to `/settings/credit`, and the latter told me auto-lending is not available for MG subscribers. So I think that covers all the paths.

PR #1888, opened 2020-07-08
GROW-127 Experiment Redirect To Basket

eddieferrer:
* Experiment to redirect users straight to basket when adding a first loan.

GROW-127

### DIFF
--- a/src/components/LoanCards/Buttons/ActionButton.vue
+++ b/src/components/LoanCards/Buttons/ActionButton.vue
@@ -11,7 +11,10 @@
 </template>
 
 <script>
+import gql from 'graphql-tag';
+import store2 from 'store2';
 import _includes from 'lodash/includes';
+import _get from 'lodash/get';
 import addToBasketInsterstitial from '@/plugins/add-to-basket-show-interstitial';
 import Lend25Button from './Lend25Button';
 import LendIncrementButton from './LendIncrementButton';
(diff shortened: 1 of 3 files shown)


src/components/LoanCards/Buttons/ActionButton.vue
```diff
@@ -88,14 +108,76 @@ export default {
...
+							);
+
+							if (this.addToBasketRedirectExperimentShown) {
+								// User is part of shown group
+								this.$router.push({
+									path: '/basket',
+								});
+							} else {
+								// User is part of control group
+								this.triggerAddToBasketInterstitial(payload.loanId);
```
eddieferrer, line 154: This repetition of `triggerAddToBasketInterstitial` is ugly, but I was having trouble finding a better way to rework this.

src/components/LoanCards/Buttons/ActionButton.vue
```diff
@@ -68,6 +81,13 @@ export default {
 			default: false,
 		},
 	},
+	data() {
+		return {
+			addToBasketRedirectExperimentShown: false,
+			hasFreeCredits: false,
+			userPrefContinueBrowsing: false,
```
emuvente, line 88: I don't think we should store/check this. We should just be using whether they have something already in their basket or if they're using promo credits.

eddieferrer: @emuvente I initially thought this as well, the intended behavior is: 
1. Be part of experiment, have no promo credits, get redirected to basket. 
2. Click "Continue Browsing"
3. Add another loan, or don't, check out. 
4. Come back next week, add something to basket, they should NOT be redirected to basket under any circumstance, because they previously clicked "Continue Browsing"

emuvente: Where is that step 4 requirement? I don't see that in the ticket anywhere.

eddieferrer: I verified with Brandon that "_didn’t previously click “continue browsing loans”_" means, we should save it in localStorage and should last as long as possible

PR #1889, opened 2020-07-08
SUBS-314 Minor Style Changes to Drop in UI

eddieferrer:


src/components/Checkout/DropInPaymentWrapper.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="row align-right">
-		<div class="payment-holder small-12 medium-10 medium-offset-1 large-offset-5 large-7 columns">
+		<div class="payment-holder small-12 medium-8 large-7 columns">
```
eddieferrer, line 3: I decided on this approach instead of max-width. The offset classes were unnecessary with the align-right on the row above. 

If we make it any narrower on large the CVV field end up wrapping and looks too big. I went from medium-10 to medium-8 for the same reasons. It is now slightly wider than the non BT UI, but not as wide as it was before this PR

PR #1900, opened 2020-07-14
SUBS-9 Add BT Drop In Payment options to MG Setup Page

eddieferrer:
* Adds BT Drop In interface to MG setup page.
* Gated by exp and feature flag
* Rename exp to differentiate between checkout and mg and allow independent control of each

SUBS-9

### DIFF
--- a/src/components/MonthlyGood/DropInPaymentWrapperMG.vue
+++ b/src/components/MonthlyGood/DropInPaymentWrapperMG.vue
@@ -1,0 +1,460 @@
+<template>
+	<div class="row align-center">
+		<div class="payment-holder small-12 medium-8 large-7 columns">
+			<div id="dropin-container"></div>
+			<div id="dropin-button">
+				<kv-button
+					value="submit"
+					id="dropin-submit"
+					class="button"
+					:disabled="!enableCheckoutButton"
+					@click.native="submitDropInPayment"
+				>
+					<kv-icon name="lock" />
+					Confirm
+				</kv-button>
+			</div>
+
+			<loading-overlay
+				v-if="updatingPaymentWrapper"
+				id="payment-updating-overlay"
+			/>
+			<p class="attribution-text">
+				Thanks to PayPal powered by Braintree,
+				Kiva receives free payment processing for all loans.
+			</p>
+		</div>
+	</div>
+</template>
+
+<script>
+import _get from 'lodash/get';
+import numeral from 'numeral';
+import * as Sentry from '@sentry/browser';
+import Dropin from 'braintree-web-drop-in';
+import checkoutUtils from '@/plugins/checkout-utils-mixin';
+import getClientToken from '@/graphql/query/checkout/getClientToken.graphql';
+import braintreeCreateMonthlyGoodSubscription from '@/graphql/mutation/braintreeCreateMonthlyGoodSubscription.graphql';
+import KvButton from '@/components/Kv/KvButton';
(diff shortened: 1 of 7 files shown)


src/graphql/query/checkout/initializeCheckout.graphql
```diff
@@ -27,7 +27,7 @@ query initializeCheckout($basketId: String) {
 			value
 			key
 		}
-		braintreeDropInExperiment: uiExperimentSetting(key: "braintree_dropin") {
+		braintreeDropInExperimentCheckout: uiExperimentSetting(key: "braintree_dropin_checkout") {
```
eddieferrer, line 30: I want to rename this exp to differentiate from the mg one.

src/pages/MonthlyGood/MonthlyGoodSetupPageControl.vue
```diff
@@ -220,17 +220,25 @@
...
-									<div class="paypal-cover" v-if="$v.$invalid"></div>
-									<pay-pal-mg
-										:amount="totalCombinedDeposit"
-										@complete-transaction="submitMonthlyGood()"
-									/>
-									<p class="small-text">
-										Thanks to PayPal, Kiva receives free payment processing.
-									</p>
-								</div>
+							<div class="payment-dropin-wrapper" v-if="showDropInPayments">
```
eddieferrer, line 223: This will show if exp for bt drop in is active. I am assuming that there is no change in the drop in UI if there is a billing agreement or not

src/pages/MonthlyGood/MonthlyGoodSetupPageControl.vue
```diff
@@ -259,24 +267,36 @@ import { required, minValue, maxValue } from 'vuelidate/lib/validators';
 
 import AlreadySubscribedNotice from './AlreadySubscribedNotice';
 import LegacySubscriberNotice from './LegacySubscriberNotice';
-import PayPalMg from './PayPalMG';
 
+import PayPalMg from '@/components/MonthlyGood/PayPalMG';
```
eddieferrer, line 271: Moved this component

eddieferrer:
This PR is working, I tested with both CC and paypal. I think we can reuse a lot of the code in the drop in UI between MG and checkout, I will have a follow up PR for that

src/components/MonthlyGood/DropInPaymentWrapperMG.vue
```diff
@@ -0,0 +1,460 @@
...
+		},
+		submitDropInPayment() {
+			this.$kvTrackEvent('Registration', 'click-confirm-monthly-good', 'register-monthly-good');
+
+			// request payment method
+			this.btDropinInstance.requestPaymentMethod()
+				.then(btSubmitResponse => {
+					const transactionNonce = _get(btSubmitResponse, 'nonce');
+					const deviceData = _get(btSubmitResponse, 'deviceData');
+					const paymentType = _get(btSubmitResponse, 'type');
```
mcstover, line 155: not sure we need to pass deviceData any longer but we should push through the paymentType and tie it into the analytics event.

src/pages/MonthlyGood/MonthlyGoodSetupPageControl.vue
mcstover: That's correct we'll show the dropin ui if they have an active billing agreement which will allow a choice of vaulted payment methods for this subscription.

PR #1903, opened 2020-07-16
Multiple setuiab query params

eddieferrer:
* Added ability to chain multiple setuiab query params, for example:

https://dev-vm-01.kiva.org/lend-by-category/women?setuiab=add_to_basket_redirect.shown&setuiab=add_to_basket_v2.control


eddieferrer:
I needed to do some testing with multiple experiments, and was disappointed we didnt support this.

PR #1911, opened 2020-07-17
Lw 1972

BoulderBrains:
Created this as draft until it's refined. 

Remaining pieces: 
- [x]  Get the line-height set on the .message text, won't apply, need to a more specific selector

- [x] Amount buttons need to be pushed left on large screens

- [ ]  Thermometer animation down, when accordion is closed

- [x] Hook up button amounts to contentful

- [x] loading state?


src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -66,6 +95,12 @@ export default {
 				});
 
 				if (activePromoBanner) {
+					// Check banner type
+					if (activePromoBanner.fields.bannerType === 'Appeal Banner') {
+						this.appealEnabled = true;
+						this.appealBannerContent = activePromoBanner;
+					}
```
mcstover, line 102: This whole block should move to the bottom and have an `else` statement that now contains `this.isPromoEnabled = true;` You could also move the assignment of `this.promoBannerContent = {...` into the else block before setting isPromoEnabled to true.

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -1,9 +1,15 @@
 <template>
-	<generic-promo-banner
-		v-if="isPromoEnabled"
-		:icon-key="promoBannerContent.iconKey"
-		:promo-banner-content="promoBannerContent"
-	/>
+	<div>
+		<generic-promo-banner
+			v-if="isPromoEnabled && !showAppeal"
```
mcstover, line 4: can drop the `!showAppeal` check here

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -1,9 +1,15 @@
...
+			:icon-key="promoBannerContent.iconKey"
+			:promo-banner-content="promoBannerContent"
+		/>
+		<appeal-banner
+			v-if="showAppeal && !hasPromoSession"
```
mcstover, line 9: can drop the `!hasPromoSession` check here

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner.vue
```diff
@@ -0,0 +1,344 @@
...
+			const appealHeadline = props.appealBannerContent.richText;
+			return documentToHtmlString(appealHeadline);
+		},
+		bannerBody(props) {
+			const appealBody = props.appealBannerContent.additionalContent[0].fields.richText;
+			return documentToHtmlString(appealBody);
+		},
+		buttonAmounts(props) {
+			const { donationAmounts } = props.appealBannerContent.dataObject;
+			return donationAmounts;
```
mcstover, line 133: Please check for donation amounts and provide defaults if the donationAmounts array is empty. You can set the defaults to the current 3 options.

BoulderBrains: Put in an attempt to fix this, though I'm having issues figuring out the right check to put in place. Going to eat some food and take a break before revisiting.

PR #1913, opened 2020-07-17
GROW-138 lend-by-category as homepage experiment

mlibazisi:
Ticket: GROW-138

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -11,6 +11,7 @@
 import _get from 'lodash/get';
 import gql from 'graphql-tag';
 import { settingEnabled } from '@/util/settingsUtils';
+import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import {
 	lightHeader, lightFooter, iwdHeaderTheme, iwdFooterTheme, wrdHeaderTheme, wrdFooterTheme
 } from '@/util/siteThemes';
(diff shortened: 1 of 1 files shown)


eddieferrer:
I think we need to prefetch this experiment in `src/util/experimentPreFetch.js` to avoid the flash of the non experiment homepage.

ryan-ludwig:
> I think we need to prefetch this experiment in `src/util/experimentPreFetch.js` to avoid the flash of the non experiment homepage.

@eddieferrer Gave it a test in my VM. Because the query is run in prefetch there's no flash that I could tell.

src/pages/Homepage/Homepage.vue
```diff
@@ -74,34 +81,51 @@ export default {
...
-			// determine if take-over setting is active
-			const contentSetting = _get(data, 'contentful.entries.items', []).find(item => item.fields.key === 'ui-homepage-takeover'); // eslint-disable-line max-len
-			if (_get(contentSetting, 'fields')) {
-				const isContentEnabled = settingEnabled(
-					contentSetting.fields,
-					'active',
-					'startDate',
-					'endDate'
+			// lend-by-category as homepage experiment - EXP-GROW-138-Jul2020
+			const lendByCategoryExperiment = this.apollo.readFragment({
```
mcstover, line 85: @mlibazisi @ryan-ludwig @eddieferrer 

I just noticed that this probably needs a prior step to work. Looking into a couple of other experiment setups that were not initializing properly I found they were based solely on the settings and had not been assigned yet.

While this approach will work if you are testing via the force url it will not work if you are just relying on the setting distribution to initialize the experiment.

You'll need to create a prefetch function that chains the experiment assignment after the initial page query. See https://github.com/kiva/ui/pull/1920 for how that should be setup.

mlibazisi: @mcstover Thanks for the feedback. Will do this as a new PR

PR #1916, opened 2020-07-19
Limit for loan query must be positive rather than 0.

mcstover:
Fixes minor error due to validation in GraphQL API
`query: "query ($userId: Int) {↵  lend {↵    loans(limit: 0, filters: {lenderFavorite: $userId}) {↵      totalCount↵      __typename↵    }↵    __typename↵  }↵  my {↵    savedSearches(limit: 10) {↵      values {↵        id↵        name↵        url↵        __typename↵      }↵      __typename↵    }↵    __typename↵  }↵}"
`

`0: "GraphQLError: Limit must be positive."`

### DIFF
--- a/src/graphql/query/lendMenuPrivateData.graphql
+++ b/src/graphql/query/lendMenuPrivateData.graphql
@@ -1,6 +1,6 @@
 query lendMenuPrivateData($userId: Int) {
 	lend {
-		loans(limit: 0, filters: {
+		loans(limit: 1, filters: {
 			lenderFavorite: $userId
 		}) {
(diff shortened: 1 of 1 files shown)


emuvente:
Did this only start throwing errors recently with the backend max limit changes? Because limits of 0 should work, (for the purposes of only getting totalCount and not actually fetching any objects) and if they aren't being accepted we need to fix the backend to accept them again.

mcstover:
I'm not sure when this started...seems to start May 26th in woodshack. Will look deeper this week to find the change that started this.

PR #1917, opened 2020-07-19
Log network errors and their corresponding query.

mcstover:
Open to other ideas here. With regard to the ECONNRESET errors, there is a chance that our queries are malformed and this will help us inspect those and to determine if any particular query is the culprit.


mcstover:
@emuvente going to throw this into test and maybe stage and see if it yields anything useful

src/api/NetworkErrorLink.js
```diff
@@ -0,0 +1,16 @@
...
+	return onError(({
+		networkError,
+		operation,
+		forward
+	}) => {
+		if (networkError) {
+			logFormatter(operation, 'error');
+			logFormatter(networkError, 'error');
+		}
+		forward();
```
emuvente, line 14: Looking at examples, calling this doesn't appear necessary

mcstover: I thought so too! Especially looking at the code for the error link. However, when I initially tried without it it that made this a terminating link and halted the flow to other links...

emuvente: aha

PR #1921, opened 2020-07-21
GROW-138 Add prefetch function

mlibazisi:
See https://github.com/kiva/ui/pull/1913


mlibazisi:
> This should do it.
> 
> I forgot to mention but one way to verify this is to use the force url to activate a specific version on page load. Then remove the force url and refresh the page to verify that your activated version is still served appropriately.

@mcstover Thanks a lot! Thats helpful to know!

PR #1926, opened 2020-07-22
Dynamically import homepage component

ryan-ludwig:
Prevents loading in JS and SVGs that aren't being used depending on the experiment. Good results:

image

image

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -18,10 +18,6 @@ import {
 } from '@/util/siteThemes';
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
 import WwwPage from '@/components/WwwFrame/WwwPage';
-import DefaultHomePage from '@/pages/Homepage/DefaultHomepage';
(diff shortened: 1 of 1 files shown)


eddieferrer:
@ryan-ludwig @emuvente 

It seems to me that this is breaking things. It's causing invariant errors in subcomponents of the homepage components, preventing the preFetch function from running, and SSR from taking place in those components. When apolloPreFetch happens server side, and it tries to get the component tree on Homepage.vue, it is unaware of DefaultHomePage, LendByCategoryHomepage, IWDHomePage, WRDHomePage. 

I believe we have to revert this entirely. 

Can you guys confirm?

emuvente:
Yeah that's true, prefetch isn't going to work for it since they are dynamically loaded instead of being included in the component tree. In the short term we'll probably have to revert to resolve this, but we could keep the gains from this change by having the prefetch function for Homepage.vue call the prefetch functions for the component tree of the selected homepage.

ryan-ludwig:
Dang :(  I suppose at minimum we can comment out IWDHomepage and WRDHomepage.

PR #1929, opened 2020-07-24
Grow 134

BoulderBrains:
Statistics section for the Lend By Category Homepage: 
Mobile: 
Screen Shot 2020-07-23 at 6 48 08 PM

Desktop: 
Screen Shot 2020-07-23 at 6 46 54 PM


BoulderBrains:
@mcstover couple things I'd like to connect with you about: 
1) Get clarity around the apollo/federation usage and correct it
2) Correctly catching the error on line 79 of HomepageStatistics.vue

src/pages/Homepage/HomepageStatictics.vue
```diff
@@ -0,0 +1,128 @@
...
+				console.log('data', data);
+			});
+		},
+	},
+	created() {
+		try {
+			const statisticData = this.apollo.readQuery({
+				query: homepageStatistics,
+			});
+			this.amountFunded = _get(statisticData, 'general.kivaStats.amountFunded');
```
mcstover, line 74: @BoulderBrains 
Seeing that you have no variables in your prefetch you could simplify your prefetch setup based on the structure in BraintreeCheckout.vue. https://github.com/kiva/ui/blob/master/src/components/Checkout/BraintreeCheckout.vue#L248

The main takeaway is that you can use a `result` block as the second part of the prefetch function and assign all of this data there, similar to CheckoutPage.vue prefetch setup. Then you can get rid of this entire created block.

BoulderBrains: Thank you @mcstover, I fixed that up in my latest commit.

src/pages/Homepage/HomepageStatictics.vue
```diff
@@ -0,0 +1,116 @@
...
+import { differenceInYears } from 'date-fns';
+import homepageStatistics from '@/graphql/query/homepage/homepageStatistics.graphql';
+
+export default {
+	inject: ['apollo'],
+	apollo: {
+		query: homepageStatistics,
+		preFetch: true,
+		result({ data }) {
+			this.amountFunded = _get(data, 'general.kivaStats.amountFunded');
```
mcstover, line 64: You'll need to create the reactive data object (`data() { return { your: defaults },`) and instantiate each of these data points there. Give them each a default value based on what is shown in prod currently... Then you won't have to worry about any unforeseen failure in the prefetch.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -124,16 +124,32 @@
...
 			<div class="row">
-				<h2 class="statistics__header small-12 columns">
-					Statistics
-				</h2>
+				<div class="small-12 columns">
+					<kv-responsive-image
+						class="statistics__img"
+						:images="statistics"
+						loading="lazy"
+						alt="statistics image"
```
ryan-ludwig, line 132: Since this image is just for decoration (it doesn't add meaning for a user), and empty alt tag, `alt=""`, will keep the image from announcing to screenreaders.

src/pages/Homepage/HomepageStatictics.vue
```diff
@@ -0,0 +1,124 @@
...
+<style lang="scss" scoped>
+@import 'settings';
+
+.stats-box {
+	border-radius: rem-calc(25);
+	background-color: $kiva-accent-blue;
+	color: $white;
+	padding: 3.5rem 0;
+
+	&__group {
```
ryan-ludwig, line 98: BEMin it up, woo! :)

BoulderBrains: I've enjoyed working in BEM. It structures the code in a way that is quickly readable, I like it. 📖

src/pages/Homepage/HomepageStatictics.vue
```diff
@@ -0,0 +1,124 @@
...
+		margin-bottom: rem-calc(40);
+
+		&.even {
```
ryan-ludwig, line 101: For true BEMminess, these would be written in the modifier syntax. It seems a little verbose, but that's kinda part of BEM 🤷‍♂️

So even and odd would be written as 
`&--even { special overridey styles here }` 
`&--odd { special overridey styles here }`

When sass compiles that, the classname end up as `.stats-box__group--even` and `.stats-box__group--odd`

Some nice examples here [](https://seesparkbox.com/foundry/bem_by_example)

BoulderBrains: Thanks for this @ryan

eddieferrer:
@BoulderBrains I noticed a couple of things with this PR. 

"Statictics" should be spelled correctly in filenames/references

Also seeing an invariant error:

`"ApolloMixin homepageStatictics: Invariant Violation: Can't find field amountFunded on object`

eddieferrer:
@BoulderBrains Also take a look at: src/graphql/query/whyKivaData.graphql I think we can reuse that query instead of creating a new one

BoulderBrains:
Thanks for the follow up here @eddieferrer, sorry for not tagging you originally. 
Corrections made in this new PR: 
https://github.com/kiva/ui/pull/1937

PR #1930, opened 2020-07-24
Update loan related query fragments to only count those that lent to …

mcstover:
…the loan.


src/graphql/fragments/loanCardFields.graphql
```diff
@@ -38,7 +38,7 @@ fragment loanCardFields on LoanBasic {
 		lentTo
 	}
 	# lenders added for DetailedLoanCard Query Cache
-	lenders {
+	lenders(limit: 0) {
```
mcstover, line 41: @colin-at-kiva I believe the sections below here for `... on LoanPartner` and `... on LoanDirect` get expanded into what look like mostly identical queries as part of the federation process. So it's pretty dry at the code level but must be expanded to properly resolve.

PR #1932, opened 2020-07-24
GROW-131 Loan Categories Section Component

eddieferrer:
* Styled component
* Reworked to fix invariant error
* Added touch scrolling and related
* Added cleaning up category names

GROW-131

### DIFF
--- a/src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
+++ b/src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
@@ -6,12 +6,13 @@
 			ref="categoryOptions"
 			:style="{ left: scrollPos + 'px' }"
 		>
+			<kv-loading-spinner v-if="!categoriesLoaded" />
 			<a
 				href="#"
 				role="button"
 				class="category-options__link"
 				:class="{'active': category.id === activeCategory}"
-				v-for="(category) in prefetchedCategoryInfo"
+				v-for="category in prefetchedCategoryInfo"
 				:key="category.id + '-link'"
 				@click.prevent="setActiveCategory(category.id)"
 			>
@@ -20,6 +21,7 @@
 			<router-link
 				class="category-options__link"
 				:to="`/lend-by-category`"
+				v-if="prefetchedCategoryInfo.length > 0"
 			>
 				More
 			</router-link>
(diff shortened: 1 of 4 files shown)


eddieferrer:
This PR is for this component: `src/components/Homepage/LendByCategory/LoanCategoriesSection.vue`

On activating a category, the component fetches 8 loans, then filters them based on funding status, and passes the filtered subset to the loan-category component. 

Because we keep track of the loans and pass them to the query as excluded ids, when clicking on a following category, the loans that have already been fetched for previous categories are excluded. 

We're only fetching 8 loans at the moment, so in theory if those 8 came back and were filtered out we would not have a full category. One approach to solve this would be to implement a queue and get more category loans, similar to the featuredLoans functionality, or an easier solution would be to get more loans in the initial query, lets say 16, and slice the top 8 after filtering and hope for the best.

src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
```diff
@@ -138,41 +164,74 @@ export default {
...
-		// get category loans for a given category id
+		// get filtered loans for a given category id
 		getCategoryLoans(categoryId) {
-			return this.categoriesWithLoans.filter(category => category.id === categoryId)[0] || {};
+			let filteredLoansArray = [];
+
+			const allLoansForCategory = this.categoriesWithLoans
+				.filter(category => category.id === categoryId)[0] || {};
+
+			const testFilter = loan => loan.loanAmount.includes('2');
```
mcstover, line 190: What are you testing for here, was this literally a "test" filter? Seems like loanAmount is a string that can include all numbers and a period.

eddieferrer: I forgot to remove that test filter 🤦

src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
```diff
@@ -60,74 +68,92 @@ export default {
...
-			} catch (e) {
-				logReadQueryError(e, 'LoanCategoriesSection loanChannelInfoQuery');
-			}
-		}
-	},
 	created() {
-		// set initial active category
-		this.setActiveCategory(this.categoryIds[0]);
+		// TODO
+		// Get these queries in preFetch without causing an invariant error.
```
mcstover, line 76: I think this should either be guarded to only run in the browser, either with a window check or moved to mounted. I don't think there's anything in the created hook that requires this promise to complete on the server like there would be with a prefetch.

Maybe it's ok if the page only prefetches the featured loan section and this one loads up once in the browser. That would lighten the initial load.

mcstover: If we went this route of not prefetching the first set, I would consider implementing a more graceful loading state for area of the page. Something that could reveal and animate both the Names slider and the Loan rows as they come into view and transition between sets.

Requesting a larger amount of loans isn't out of the question... maybe only 12. Though having the dynamic need based loader would be better at some point.

eddieferrer: I moved the initialization code to mounted and added a loader. I think this is good enough to merge and continue progress. I am aware that in the case where we fetch 8 loans for a category, and some of those loans are filtered out the category row slider will have an insufficient amount of loans

PR #1934, opened 2020-07-25
grow-135 assests, structure, styles.

BoulderBrains:
Got most of this in lender quotes section for the lend by category homepage built today. 
It currently looks like this: 
**Mobile**
Screen Shot 2020-07-24 at 5 15 02 PM

**Desktop**
Screen Shot 2020-07-24 at 5 14 46 PM

The main piece left is getting the card background image into place. 

Just a draft PR currently feel free to make any comments you've got.


ryan-ludwig:
We can improve the file size of the flourish assets. They should end up under 10k for sure. With the weathered texture look they have it makes SVGs really huge. I had good luck exporting them in Photoshop with PNG-8 at 16 colors and then running those PNGs through [ImageOptim](https://imageoptim.com/).

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -157,6 +157,47 @@
 				<h2 class="lender-quotes__header text-center small-12 columns">
 					What our lending <br class="so mo"> community thinks
 				</h2>
+				<div
+					class="lender-quotes__lender-quote--card small-10 medium-8 large-5 xlarge-4 columns"
```
ryan-ludwig, line 161: I would recommend making the cards themselves their own BEM "block" since they could stand alone easily without the lender-quotes section.. Maybe .quote-card or something.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -157,6 +157,47 @@
...
+						class="lender-quotes__lender-quote--attribution"
+					>
+					</div>
+					<div
+						v-html="lenderQuote.title"
+						class="lender-quotes__lender-quote--title"
+					>
+					</div>
+					<!-- TODO:
+					This is a little tricky, need to put more thought in here
```
ryan-ludwig, line 189: Oh interesting, you could add a modifier class with the loop index on the quote-card. And then in CSS use the index to position them.

Sortof like 
```
<div :class="quote-card quote-card--{index}">
  <div :class="quote-card__flourish">
  ...

.quote-card__flourish {
  position: absolute;
}

.quote-card--1 .quote-card__flourish {
  top: 1rem;
  left: 1rem;
  width: 3rem;
}
.quote-card--2 quote-card__flourish {
  top: 2rem;
  right: 0;
  width: 2.5rem;
}
```

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -157,6 +157,47 @@
...
+				>
+					<kv-responsive-image
+						class="lender-quotes__lender-quote--lender-img"
+						:images="[
+							['small', lenderQuote.image],
+						]"
+						loading="lazy"
+						alt=""
+					/>
+					<div
```
eddieferrer, line 173: I'm not sure that these should be divs. To keep it semantic, I think they should be a combination of p's and spans. Is there any reason to use v-html here instead of just template interpolation {{ }} ?

src/pages/Homepage/LendByCategoryHomepage.vue
eddieferrer: ☝️ Like @ryan-ludwig  said. You could also target with `:nth-child(odd)` `:nth-child(even)` if you wanted to support an arbitrary number of quotes,

src/pages/Homepage/LendByCategoryHomepage.vue
BoulderBrains: Nope no specific reason Eddie, thanks for bringing this up, I've made that change after reading a little on this. (https://forum.vuejs.org/t/performance-v-html-vs-interpolation-aka-braces/29559/3)

eddieferrer: @BoulderBrains 👍  theres also some XSS security issues with v-html, they wouldnt apply in this case, but something to be aware of: https://vuejs.org/v2/guide/syntax.html#Raw-HTML

BoulderBrains:
Adjustments mentioned above have been addressed. 

:
Screen Shot 2020-07-27 at 4 38 43 PM

:
Screen Shot 2020-07-27 at 4 38 01 PM

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -459,6 +506,80 @@ export default {
...
+		}
+
+		&__flourish {
+			position: absolute;
+			z-index: 0;
+		}
+	}
+
+	// Couldn't get BEM working in here
+	.quote-card:nth-child(even) {
```
ryan-ludwig, line 560: Yeah, this is a sass thing I'm afraid. What you've got is good.

In sass the` &` really just means append that to the parent selector in SASS. It makes it easy to write BEM, but it's not really BEM specific.
```
.hi {
  &-there {
```

would output `.hi-there` 

```
.hi:nth-child(even) {
  &-there {
```
would try to output `.hi:nth-child(even)-there` but fail because it's an invalid selector.


So in your case you could do
```
.quote-card:nth-child(even) {
  .quote-card {
    &__flourish { ...
```

But that seems excessive. What you've got is readable to me.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -157,6 +157,33 @@
...
+						class="quote-card__flourish"
+						:images="[
+							['small', lenderQuote.background],
+						]"
+						loading="lazy"
+						alt=""
+					/>
+					<img
+						:src="lenderQuote.image"
+						class="quote-card__lender-img"
```
ryan-ludwig, line 175: adding `loading="lazy"` and `:alt="lenderQuote.attribution"` would be good here.

src/pages/Homepage/LendByCategoryHomepage.vue
BoulderBrains: Thanks for this explanation @ryan-ludwig

ryan-ludwig:
@BoulderBrains I think adding a `z-index: -1;` to all the flourishes will safeguard overlaps.

eddieferrer:
@BoulderBrains yellow flourish still visible under card at ~999px. I think you can add a background of white to `.quote-card`  to take care of the issue

BoulderBrains:
Thanks for the review @eddieferrer, I believe that all of your raised issues have been addressed in the last few commits. 
- Took out the xlarge-4 class, so quote-cards remain at the large-5 size all the way up. 
- Fixed the flourish overlap issue, good catch on that one, thank you. 
**Questions:**
- I fixed the margin between the two card by manually setting the margin-right and margin-left on the odd/even card definitions: 
```.quote-card:nth-child(even) {
		@include breakpoint(large) {
			margin-right: 0.9rem;  <--- HERE
		}

		.quote-card__flourish {
			top: 0;
			left: 0;
			width: 10rem;

			@include breakpoint(large) {
				width: 8rem;
			}
		}
	}

	.quote-card:nth-child(odd) {
		@include breakpoint(large) {
			margin-left: 0.9rem; <--- HERE
		}

		.quote-card__flourish {
			bottom: 4rem;
			right: 0;
			width: 7.5rem;

			@include breakpoint(large) {
				width: 5.5rem;
			}
		}
	}```
Would it be prefered to try and mess around with the offset and incomplete row options in foundation? 

- Changing the width of the cards makes the left card and the .lender-quotes__flourish to overlap. I should be able to fix this with z-index as both the .quote-card(position: relative) and .lender-quotes__flourish(position: absolute) are position with a definition that meets the z-index requirements, however this doesn't seem to be working. Seeking advice there. 

Let me know what you think, thanks again for the review.

File: src/pages/Homepage/LendByCategoryHomepage.vue [hunk 6]
```diff
@@ -439,7 +487,9 @@ export default {
 
 .lender-quotes {
 	&__header {
+		margin-bottom: 2rem;
 		font-weight: bold;
+		z-index: 1;
```
ryan-ludwig, line 492: not needed

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -459,6 +509,89 @@ export default {
 			pointer-events: none;
```
ryan-ludwig, line 509: needs a z-index: -1

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -459,6 +509,89 @@ export default {
...
+		&__quote {
+			line-height: 1.3;
+			margin-bottom: rem-calc(30);
+		}
+
+		&__attribution {
+			line-height: 1.3;
+			font-weight: bold;
+			margin-bottom: 0;
+			z-index: 100;
```
ryan-ludwig, line 548: not needed

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -459,6 +509,89 @@ export default {
...
+			position: relative;
+		}
+
+		&__title {
+			z-index: 100;
```
ryan-ludwig, line 553: not needed

BoulderBrains:
Screen Shot 2020-07-28 at 2 06 17 PM

Screen Shot 2020-07-28 at 2 05 48 PM

BoulderBrains:
Thank you both 🙏

PR #1938, opened 2020-07-29
GROW-159 KvCarousel

ryan-ludwig:
Most of the work done to make KvCarousel able to handle what's needed for the homepage redesign as well as other carousels Kiva has done in the past (homepage hero images, KivanSlider on holiday promo).

Still have some styling todos around colors and the arrow btns but don't want to hold up Eddie. I also don't love the way the transitions are handled right with scoped slots, may work on that more tomorrow. Lotta position absolutes, wish I could use grid.

Right now none of the other implementations of KvCarousel are live so this shouldn't break anything.

@eddieferrer There's a storybook story "loanCardExample", with an idea around how to show the loans. The text above the slider isn't part of this component, but you should be able to build that using the change event on the carousel. 
image

### DIFF
--- a/.storybook/stories/KvCarousel.stories.js
+++ b/.storybook/stories/KvCarousel.stories.js
@@ -1,0 +1,395 @@
+import Vue from 'vue'
+import Vue2TouchEvents from 'vue2-touch-events';
+
+import KvCarousel from '@/components/Kv/KvCarousel';
+import KvCarouselSlide from '@/components/Kv/KvCarouselSlide';
+import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
+import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
+
+Vue.use(Vue2TouchEvents);
+
+const defaultCarouselSlides = `
+	<template slot="default" slot-scope="props">
+		<kv-carousel-slide :transition-name="props.transitionName">
+			<img src="https://via.placeholder.com/300x220/e8f0f6/000000" style="width: 100%;">
+		</kv-carousel-slide>
+		<kv-carousel-slide :transition-name="props.transitionName">
+			<img src="https://via.placeholder.com/300x220/a87c7c/000000" style="width: 100%;">
+		</kv-carousel-slide>
+		<kv-carousel-slide :transition-name="props.transitionName">
+			<img src="https://via.placeholder.com/300x220/f6dbb8/000000" style="width: 100%;">
+		</kv-carousel-slide>
+		<kv-carousel-slide :transition-name="props.transitionName">
+			<img src="https://via.placeholder.com/300x220/b39696/000000" style="width: 100%;">
+		</kv-carousel-slide>
+	</template>
+`;
+
+export default {
+	title: 'Kv/KvCarousel',
+	component: KvCarousel,
+ };
+
+export const Default = () => ({
+	components: {
+		KvCarousel,
+		KvCarouselSlide
+	},
+	template: `
+		<kv-carousel style="width: 400px; height: 250px;">
+			${defaultCarouselSlides}
+		</kv-carousel>
+	`,
+});
+
+export const PauseOnHoverFalse = () => ({
+	components: {
+		KvCarousel,
+		KvCarouselSlide
+	},
+	template: `
+		<kv-carousel
+			:pause-on-hover="false"
+			style="width: 400px; height: 250px;"
+		>
+			${defaultCarouselSlides}
+		</kv-carousel>
+	`,
+});
+
+export const AutoplayFalse = () => ({
+	components: {
+		KvCarousel,
+		KvCarouselSlide
+	},
+	template: `
+		<kv-carousel
+			:autoplay="false"
+			style="width: 400px; height: 250px;"
+		>
+			${defaultCarouselSlides}
+		</kv-carousel>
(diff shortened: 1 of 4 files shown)


.storybook/stories/KvCarousel.stories.js
```diff
@@ -0,0 +1,395 @@
...
+		<kv-carousel
+			indicator-style="none"
+			style="width: 400px; height: 250px;"
+		>
+			${defaultCarouselSlides}
+		</kv-carousel>
+	`,
+});
+
+export const HideArrowsTrueInidicatorNone = () => ({
```
eddieferrer, line 105: misspelling

src/components/Kv/KvCarousel.vue
```diff
@@ -29,53 +166,288 @@ export default {
...
-		clearInterval(this.interval);
+		if (this.autoplay) {
+			clearInterval(this.intervalTimer);
+		}
 	},
 	methods: {
+		/**
+		 * Move forward one slide
+		 *
+		 * @public This is a public method
```
eddieferrer, line 197: These @public decorations should be removed since these methods will not be called by reference anymore (they are only called by reference from KivanSlider component)

src/components/Kv/KvCarousel.vue
```diff
@@ -1,24 +1,161 @@
...
+	<div
+		class="kv-carousel"
+		:class="carouselModifierClasses"
+		ref="KvCarousel"
+	>
+		<div class="kv-carousel__slides"
+			@mouseover="pauseOnHover ? paused = true : paused = false"
+			@touchstart="pauseOnHover ? paused = true : paused = false"
+			@mouseleave="pauseOnHover ? paused = false : paused = false"
+			@touchend="pauseOnHover ? paused = false : paused = false"
```
emuvente, line 11: For `@mouseleave` and `@touchend` you don't have to check `pauseOnHover` because you're setting `paused` to false either way.

src/components/Kv/KvCarousel.vue
```diff
@@ -29,53 +166,288 @@ export default {
...
+		height: 1rem;
+		border-radius: 50%;
+		background: $kiva-text-dark;
+		pointer-events: auto;
+		position: relative;
+		overflow: hidden;
+
+		&:hover,
+		&:focus {
+			background: yellow; // TODO: colors
```
eddieferrer, line 338: How do you propose we handle the colors?

src/components/Kv/KvCarousel.vue
```diff
@@ -1,24 +1,161 @@
...
+					name="small-chevron"
+					:from-sprite="true"
+					title="Show previous slide"
+				/>
+			</button>
+			<button
+				class="kv-carousel__arrows-btn kv-carousel__arrows-btn--right"
+				@click="advance"
+			>
+				<kv-icon
```
eddieferrer, line 34: We might need to adjust these icons. The small chevron is a little thin and could possibly be bolder? There is also the slightest CSS bug: When the right arrow is clicked, the chevron shifts in position slightly then shifts back as the slide transitions.

.storybook/stories/KvCarousel.stories.js
```diff
@@ -0,0 +1,395 @@
...
+import KvCarouselSlide from '@/components/Kv/KvCarouselSlide';
+import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
+import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
+
+Vue.use(Vue2TouchEvents);
+
+const defaultCarouselSlides = `
+	<template slot="default" slot-scope="props">
+		<kv-carousel-slide :transition-name="props.transitionName">
+			<img src="https://placekitten.com/300/290" style="width: 100%;">
```
eddieferrer, line 14: Let's change placekitten to https://via.placeholder.com/{width}x{height}/{background-color-hexcode} (I know, I'm no fun)

src/components/Kv/KvCarousel.vue
ryan-ludwig: Yeah I'm planning on making a fatter chevron for the KvIcon sprite sheet today. I noticed that right arrow click shift too! so strange 😕

src/components/Kv/KvCarousel.vue
ryan-ludwig: Good question, I'd like to get this merged (after squash) before the branch cut. Then today work with Aaron to get the hover/focus/progressbars etc colors sorted out.

.storybook/stories/KvCarousel.stories.js
ryan-ludwig: 😿

src/components/Kv/KvCarousel.vue
ryan-ludwig: My thinking was they could be useful in the future if other components on a page want to interact with this slider. Don't feel strongly though. Also I was trying to get Storybook to auto-document the methods but no luck :/

ryan-ludwig:
> * I think we can delete the `KivanSlider.vue` component. To avoid any confusion. This PR breaks it's use in `/possibilities`, which isn't active anyway, and you've created a story here of a viable way to reimplement it. It's only used in `YearEnd.vue` anyway.
Sounds good, will do that in a separate PR.

PR #1942, opened 2020-07-30
GROW-131 Loan Card

eddieferrer:
* Styling and cleanup for the loan card on the lend by category homepage.

### DIFF
--- a/src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
+++ b/src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
@@ -12,7 +12,7 @@
 				:class="{'active': category.id === activeCategory}"
 				v-for="category in prefetchedCategoryInfo"
 				:key="category.id + '-link'"
-				@click.prevent="setActiveCategory(category.id)"
+				@click.prevent.native="setActiveCategory(category.id)"
 			>
 				{{ cleanCategoryName(category) }}
 			</kv-button>
@@ -262,6 +262,8 @@ export default {
 	top: -4.5rem;
 	position: absolute;
 	white-space: nowrap;
+	text-align: center;
+	padding: 0 2rem;
 
 	&__link {
 		color: $charcoal;
@@ -305,9 +307,9 @@ export default {
 		justify-content: space-between;
 		min-width: 100%;
 	}
-}
 
-.loading-spinner {
-	margin: 0 auto;
+	.loading-spinner {
+		margin: 0 auto;
+	}
 }
 </style>
--- a/src/components/Homepage/LendByCategory/LoanCategory.vue
+++ b/src/components/Homepage/LendByCategory/LoanCategory.vue
@@ -22,13 +22,12 @@
 				>
 					<loan-card-controller
 						loan-card-type="LendHomepageLoanCard"
-						v-for="(loan, index) in loans"
+						v-for="loan in loans"
 						:key="loan.id"
 						:loan="loan"
 						:items-in-basket="itemsInBasket"
 						:category-id="loanChannel.id"
 						:category-set-id="setId"
-						:card-number="index + 1"
 						:enable-tracking="true"
 						:is-visitor="!isLoggedIn"
 					/>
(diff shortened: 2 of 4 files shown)


src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
```diff
@@ -12,7 +12,7 @@
 				:class="{'active': category.id === activeCategory}"
 				v-for="category in prefetchedCategoryInfo"
 				:key="category.id + '-link'"
-				@click.prevent="setActiveCategory(category.id)"
+				@click.prevent.native="setActiveCategory(category.id)"
```
eddieferrer, line 15: I've forgotten this small detail several times now. KvButton needs the .native decorator on the click event to work when it is triggering a component function.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -359,6 +359,10 @@ export default {
 }
 
 .loan-categories {
+	& .row {
+		max-width: 69.15rem;
```
eddieferrer, line 363: This feels hacky, but the design given is wider than our container and I don't see any other way to get 3 cards in the row

eddieferrer:
Still left to do:
- Add story in storybook for this card. 
- Sync up with Brandon about tracking events and update those
- Change out the arrow for the ones @ryan-ludwig recently made.

src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
ryan-ludwig: It'd be nice to not need that at all and just use @click. Pretty sure we can make changes in KvButton to facilitate that. Will make a ticket in the Vue backlog.

src/components/Homepage/LendByCategory/LoanCategory.vue
```diff
@@ -72,22 +71,18 @@
...
 import _throttle from 'lodash/throttle';
 import LoanCardController from '@/components/LoanCards/LoanCardController';
-import categoryRowArrowsVisibleMixin from '@/plugins/category-row-arrows-visible-mixin';
 import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
 
-// TODO change card component used in this component and these values to fit 3 cards in row.
-const cardWidth = 320;
-const cardRightMargin = 10;
+// These values have to be the same as the values in src/components/LoanCards/LendHomepageLoanCard.vue
+const cardWidth = 305;
```
ryan-ludwig, line 77: I don't think it's worth going down a rabbit hole on this right now, but something to chew on. From what I see here, javascript is using px values, and CSS is using rem values. If the user has their browser font-size (not browser zoom) set to a non-default, the box will take up a different number of px due to remminess, but the JS will still be running on the original px. One thought is to get the width of the card on the page using JS instead of hardcoding it. 

Again, not something worth doing now, I'd like to see this kindof multi-slide functionality added as an improvement to KvCarousel in the future instead.

src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -212,173 +167,94 @@ export default {
...
-	.hover-loan-card-image {
-		@extend .base-portrait-hover-loan-card-image;
 
-		height: $large-hover-card-image-height;
+// These values have to be the same as the values in src/components/Homepage/LendByCategory/LoanCategory.vue
+$card-width: rem-calc(305);
+$card-margin: rem-calc(14);
+$card-half-space: rem-calc(14/2);
+$card-height: rem-calc(388);
+
```
BoulderBrains, line 176: Should we define these in _settings.scss, then have the variable to use here and in LendHomepageLoanCards.vue? There's a card definition grouping on line 557 of _settings.scss

BoulderBrains:
Sorry @ryan-ludwig, I got too clicking and requested a re-review from you by accident. My bad.

src/components/Homepage/LendByCategory/LoanCategory.vue
eddieferrer: This is a good point @ryan-ludwig

PR #1946, opened 2020-07-30
GROW-131 Story & Small Fixes

eddieferrer:
* Added loan card story.
* Fixed a couple of issues discovered using the storybook story

### DIFF
--- a/.storybook/mixins/loan-data-prop-story-mixin.js
+++ b/.storybook/mixins/loan-data-prop-story-mixin.js
@@ -1,0 +1,40 @@
+import { select, text, number, boolean } from '@storybook/addon-knobs';
+
+export default {
+	props: {
+		loan: {
+			type: Object,
+			default: () => {
+				return {
+					id: 1,
+					name: `${text('Name', 'Loan Name', 'Loan')}`,
+					image: {
+						retina: `${text('Image Retina', 'https://via.placeholder.com/956x720', 'Loan')}`,
+						default: `${text('Image Default', 'https://via.placeholder.com/478x360', 'Loan')}`,
+						hash: ''
+					},
+					loanAmount: `${text('Amount', '1250', 'Loan')}`,
+					geocode: {
+						country: {
+							name: `${text('Country', 'United States', 'Loan')}`,
+							isoCode: `${text('ISOCode', 'us', 'Loan')}`
+						},
+					},
+					use: `${text('Use', 'to buy more fire wood to sell at local market to meet customer demand.', 'Loan')}`,
+					status: `${select('Status', ['funded', 'expired', ''], '', 'Loan')}`,
+					borrowerCount: number('Borrower Count', 2, {}, 'Loan'),
+					lenderRepaymentTerm: number('Lender Repayment Term', 24, {}, 'Loan'),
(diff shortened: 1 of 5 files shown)


src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -39,7 +39,7 @@
 				:status="loan.status"
 				:use="loan.use"
 				:loan-id="loan.id"
-				:max-use-length="65"
+				:max-use-length="62"
```
eddieferrer, line 42: This was overflowing in some weird cases

src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -102,33 +102,33 @@ export default {
 		BorrowerInfoName,
 	},
 	props: {
+		amountLeft: {
```
eddieferrer, line 105: No changes here, just alphabetizing props

src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -239,7 +236,7 @@ $card-height: rem-calc(388);
 	}
 
 	&__action-button-container {
-		width: rem-calc(137);
+		width: rem-calc(150);
```
eddieferrer, line 239: This is to accomodate the "Lend again" button state, which could happen

src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -248,13 +245,16 @@ $card-height: rem-calc(388);
 
 		.action-button {
 			margin: 0;
-			font-size: rem-calc(20);
 			padding: 0.75rem 1rem;
 		}
+
+		.action-button:not(.loan-funded-text):not(.loan-expired-text):not(.loan-selected-text) {
```
eddieferrer, line 251: Technically, these states should never happen, but removed the font size for them in case they do

.storybook/stories/LendHomepageLoanCard.stories.js
```diff
@@ -0,0 +1,77 @@
...
+
+export const Default = () => ({
+	mixins: [apolloStoryMixin, loanDataPropStoryMixin],
+	components: {
+		LendHomepageLoanCard,
+	},
+	props: {
+		amountLeft: {
+			type: Number,
+			default: number("Amount Left", 250),
```
ryan-ludwig, line 28: Could these props be part of loan-data-prop-story-mixin.js?

.storybook/mixins/loan-data-prop-story-mixin.js
```diff
@@ -0,0 +1,40 @@
+import { select, text, number, boolean } from '@storybook/addon-knobs';
+
+export default {
+	props: {
+		loan: {
+			type: Object,
```
ryan-ludwig, line 6: 🔥

.storybook/stories/LendHomepageLoanCard.stories.js
eddieferrer: Maybe, I think this gives us a little more flexibility. For example, grid loan card takes a title prop, the lend homepage card doesn't. While its a safe bet that they will probably all have props like "amountLeft" maybe they won't

PR #1948, opened 2020-07-30
adding Get started link

BoulderBrains:
Adding get started link at the top of the LBChomepage. Put some dummy kvtracking data in there until I hear back from brandon about new tracking variables.

### DIFF
--- a/src/pages/Homepage/LendByCategoryHomepage.vue
+++ b/src/pages/Homepage/LendByCategoryHomepage.vue
@@ -12,6 +12,18 @@
 					<p class="featured-loans__body">
 						Kiva empowers underserved people to achieve their
 						dreams by crowdfunding loans and unlocking capital.
+						<a
+							class="show-for-large featured-loans__body--link"
+							href="/lend-by-category"
+							v-kv-track-event="[
+								'homepage',
+								'click-Get started',
+								'lend-by-category-homepage-redirect',
+								'true'
+							]"
+						>
+							Get started
+						</a>
 					</p>
 				</div>
 			</div>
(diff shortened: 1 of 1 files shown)


src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -12,6 +12,18 @@
 					<p class="featured-loans__body">
 						Kiva empowers underserved people to achieve their
 						dreams by crowdfunding loans and unlocking capital.
+						<a
+							class="show-for-large featured-loans__body--link"
```
ryan-ludwig, line 16: Here the link isn't a modifier but instead another element of the block featured-loans. So it shouldn't have two dashes which are reserved for modifiers. A class like `featured-loans__cta`, or `featured-loans__body-link` would be more apt.

ryan-ludwig: I believe this should be a <router-link> instead of <a>

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -12,6 +12,18 @@
 					<p class="featured-loans__body">
 						Kiva empowers underserved people to achieve their
 						dreams by crowdfunding loans and unlocking capital.
+						<router-link
+							class="show-for-large featured-loans__body-link"
+							href="/lend-by-category"
```
eddieferrer, line 17: @BoulderBrains this is causing an error. Sorry I didn't catch this in my initial review. Should be:
`:to="'/lend-by-category'"`

eddieferrer: @BoulderBrains Also, while you're in this code....  The design calls for this text to be vertically centered in the section.

BoulderBrains: Will have these changes up for review shortly.

PR #1953, opened 2020-07-31
GROW-130 Featured Loans

eddieferrer:
* Implemented features loans on the lend by category homepage

### DIFF
--- a/src/components/LoanCards/LendHomepageLoanCard.vue
+++ b/src/components/LoanCards/LendHomepageLoanCard.vue
@@ -194,7 +194,7 @@ $card-half-space: rem-calc(14/2);
 	box-shadow: 0 0.65rem $card-margin $card-half-space rgb(153, 153, 153, 0.1);
 
 	&__image-wrapper {
-		height: rem-calc(168);
+		height: rem-calc(165);
 		border-radius: 0.65rem 0.65rem 0 0;
 		overflow: hidden;
 		flex-shrink: 0;
@@ -202,7 +202,7 @@ $card-half-space: rem-calc(14/2);
 		::v-deep a.borrower-image-link {
 			position: relative;
 			display: inline-block;
-			height: rem-calc(168);
+			height: 100%;
 		}
 
 		::v-deep button.favorite-star {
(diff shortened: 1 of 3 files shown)


src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -194,15 +194,15 @@ $card-half-space: rem-calc(14/2);
 	box-shadow: 0 0.65rem $card-margin $card-half-space rgb(153, 153, 153, 0.1);
 
 	&__image-wrapper {
-		height: rem-calc(168);
+		height: rem-calc(165);
```
eddieferrer, line 197: I made some very minor style tweaks here.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -335,11 +337,27 @@ export default {
...
+		padding: 6rem 0;
+	}
+}
+
 .lend-by-category-homepage {
 	overflow: hidden;
 }
 
 .featured-loans {
+	padding: 2rem 0 1rem;
```
eddieferrer, line 355: This section and `.loan-categories` dont have the same padding as per the design

eddieferrer:
This PR has the basic functionality of the featured loans carousel. 

Now that we have preFetch working again on homepage subcomponents. We can prefetch the categories setting, then prefetch the category information, and the featured loan for the first category. 

Issues I'm aware of and will followup on. 

- Not currently filtering funded loans based on fundedAmount/reservedAmount, I suppose if a funded loan comes back as the featured loan, we should fetch another loan. I wonder if this is worth doing, as it would be 2 sequential API requests, and that could take longer than the 5 seconds the slide takes to auto slide. Maybe we just show a funded loan? I think we should consider filtering funded loans on the backend and passing that as a param so that they are not returned. 
- Currently the featured loans carousel and the first loan on the category carousel will be the same. I will fix this in a followup PR.

emuvente:
> Not currently filtering funded loans based on fundedAmount/reservedAmount

This could also be handled by requesting 3-4 loans in the initial query and displaying the first one that isn't funded, which should work for most cases. For the rare case where all of the loans from the initial query are funded, a second query could be done at that point to get another 3-4 loans.

Handling it on the backend is an option too, but we would have to figure out what to do with the limit and offset parameters.

PR #1954, opened 2020-07-31
get started redirect link fix and message centering using foundation align-self-middle class

BoulderBrains:
PR addressing these concerns: 
- this is causing an error. Should be:
:to="'/lend-by-category'" 
- The design calls for this text to be vertically centered in the section.
     - I've added the [foundation class](https://get.foundation/sites/docs/flex-grid.html#vertical-alignment-of-child-columns-individually-) .align-self-middle, which once the carousel has a height, it should correctly vertically center the text as the design shows.


eddieferrer:
I did not test this. I have a feeling that the vertical alignment might not work because the carousel doesn't just have a straightforward height value.

PR #1956, opened 2020-07-31
Fixed some issues with queries in this component

eddieferrer:
Guarding some of these queries to avoid apollo error. Also forgot to include the category id in the prefetch for the first loan 🤦


eddieferrer:
I'm just gonna merge this in so I can verify on dev

PR #1958, opened 2020-08-03
GROW-130 Fixed CSS issue for featured loans carousel

eddieferrer:


eddieferrer:
Just nesting these CSS styles, should fix the issue in dev

src/components/Homepage/LendByCategory/FeaturedLoansCarousel.vue
```diff
@@ -276,6 +276,32 @@ h3 {
 .featured-loans-carousel {
 	height: 0;
 	padding-bottom: 110%;
+
+	// Overwrite styles for loan card to make it responsive.
+	::v-deep .lend-homepage-loan-card {
+		width: 93%;
+	}
+
+	::v-deep .lend-homepage-loan-card__image-wrapper {
```
ryan-ludwig, line 285: Rather than v-deeping here, what do you think about moving these styles into LendHomepageLoanCard, and setting a property like "expanded = true", or "compact = true", something like that to enable them. That would make this component more re-usable in the future.

eddieferrer: @ryan You'll have to do some tweaking with the width to make these cards essentially responsive. I think if we see they get reused again at this size, this would be a good refactor.

ryan-ludwig: @eddieferrer Sounds good. In the semi-near term I'd like to get all these loan cards catalogged so we can start consolidating components, and design can see what sizes they have to work with from Storybook.

PR #1959, opened 2020-08-03
GROW-131 Prefetch Categories And Small CSS fix

eddieferrer:
* Fixed CSS of category loan carousel so that loans with 2 line info align with loans with 3 line borrower info
* Prefetch categories and category infor for category loan carousel
* Unrelated fix that was throwing a vue error

GROW-131

### DIFF
--- a/src/pages/LandingPages/MGCovid19/MGCovidHero.vue
+++ b/src/pages/LandingPages/MGCovid19/MGCovidHero.vue
@@ -152,6 +152,8 @@ export default {
 				['wga retina', heroImagesRequire('./hero_3840x1480.jpg')],
 			],
 			expVideoActive: true,
+			isMonthlyGoodSubscriber: false,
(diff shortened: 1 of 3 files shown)


src/pages/LandingPages/MGCovid19/MGCovidHero.vue
```diff
@@ -152,6 +152,8 @@ export default {
 				['wga retina', heroImagesRequire('./hero_3840x1480.jpg')],
 			],
 			expVideoActive: true,
+			isMonthlyGoodSubscriber: false,
```
eddieferrer, line 155: This isn't related to the LBC homepage but I saw this error popping up.

mcstover: should be fine as it just defines these in data.

PR #1965, opened 2020-08-04
GROW-189 Focus Styles Fix on Homepage Links

eddieferrer:
* Fixed focus style for get started link and category links


src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -374,10 +375,6 @@ export default {
 		}
 	}
 
-	&__body-link {
```
eddieferrer, line 377: `display: block` was making this have a really wide box when focused, switched to using a `br` instead

PR #1970, opened 2020-08-06
Travis config allow nodejsv14

mcstover:


mcstover:
I noticed that too @emuvente ! will switch after this build runs...

PR #1974, opened 2020-08-06
Analytics for selected payment type when signing up for Monthly Good.

mcstover:


src/components/MonthlyGood/MonthlyGoodDropInPaymentWrapper.vue
```diff
@@ -78,14 +78,13 @@ export default {
 		submitDropInMonthlyGood() {
 			this.submitting = true;
 
-			this.$kvTrackEvent('Registration', 'click-confirm-monthly-good', 'register-monthly-good');
```
mcstover, line 81: this will also get called once the `complete-transaction` event bubbles up.

PR #1975, opened 2020-08-06
Fixed image sizes for loan card for IE

eddieferrer:


src/components/LoanCards/LendHomepageLoanCard.vue
```diff
@@ -194,6 +194,12 @@ export default {
 			position: relative;
 			display: inline-block;
 			height: 100%;
+
+			.borrower-image {
+				width: 100%;
```
mcstover, line 199: 👍

PR #1981, opened 2020-08-06
GROW-174 minimal frame

ryan-ludwig:
It seemed that since we still probably need most things in WwwPage, a simpler solution was to add a new prop "minimal" to TheHeader instead.


emuvente:
I think we probably don't want the banner area component.

.storybook/mixins/apollo-story-mixin.js
```diff
@@ -1,13 +1,35 @@
 export default {
 	provide: {
-		apollo: {
+		federation: {
```
mcstover, line 3: Why is there a split on federation. Seems like this should just always use the public/prod Federation Graphql endpoint...

ryan-ludwig: We don't currently reach out to any endpoints in storybook. Just mocks. 

GlobalPromotionBannerContentful referenced federation and was causing storybook to barf. 
https://github.com/kiva/ui/blob/master/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue#L41

ryan-ludwig: Wait that's not right...

emuvente:
This frame should also include the progress bar that's just below the header indicating the progress in the onboarding flow.

.storybook/mixins/apollo-story-mixin.js
ryan-ludwig: @mcstover Would it be better to change GlobalPromotionBannerContentful.vue to reference apollo instead of federation?

ryan-ludwig: cc: @eddieferrer

mcstover: ahh ok that makes sense. a little future clean up then... ty.

mcstover: No big rush to change that atm.

ryan-ludwig:
> This frame should also include the progress bar that's just below the header indicating the progress in the onboarding flow.

I was thinking the progress bar would be inside onboarding page allowing us to use the minimal header in other contexts

.storybook/mixins/apollo-story-mixin.js
eddieferrer: @ryan-ludwig I agree with this approach for the stories. the federation provider might be able to be refactored or removed at this point, but that is outside the scope of this work. On the other hand, it is nice to be able to hit the federation server separately while in dev...

ryan-ludwig:
> Any sign of changes in rendering of the header now that it's checking the prop and contained in the template?
Nothing I noticed

emuvente:
> I was thinking the progress bar would be inside the onboarding page(s) allowing us to use the minimal header in other contexts

Since we don't have any other projects yet that are calling for a minimal header, having that flexibility isn't as important for this project. The point of the page frame ticket was to have a single frame containing all the shared components for the onboarding pages.

There's a way to do that that would allow the minimal header to be available to other future pages, which is to create an alternative www-page just for these onboarding pages that includes only the header with the minimal prop set to true, the progress bar, a single slot for content, the basket bar, the cookie banner, and the footer.

PR #1982, opened 2020-08-06
Settings Card Component

eddieferrer:
* Refactor card used on settings pages into a new component
* Add component story
* Remove duplicate CSS and component imports

### DIFF
--- a/src/components/Kv/KvSettingsCard.vue
+++ b/src/components/Kv/KvSettingsCard.vue
@@ -1,24 +1,41 @@
 <template>
-	<div class="settings-card">
-		<div class="icon-wrapper">
+	<div class="settings-card" :class="{ obscure: disabled }">
+		<div class="settings-card__icon-wrapper">
+			<!-- @slot Icon of the settings card. Should not be empty. -->
 			<slot name="icon"></slot>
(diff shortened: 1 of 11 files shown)


src/components/Kv/KvSettingsCard.vue
```diff
@@ -0,0 +1,58 @@
...
+};
+</script>
+
+<style lang="scss" scoped>
+@import 'settings';
+
+.settings-card {
+	background: $white;
+	padding: 1.95rem;
+	display: grid;
```
ryan-ludwig, line 33: Does this component still render semi-ok in IE?

eddieferrer: @ryan-ludwig It doesnt look too bad without the grid support. 

Screen Shot 2020-08-06 at 4 12 15 PM

ryan-ludwig: Yeah, not bad at all, nice!

eddieferrer: @ryan-ludwig I decided to refactor this to use flex instead for better IE support. There really was no reason to use grid here.

PR #1983, opened 2020-08-07
Grow 190 async loans displayed events

mcstover:


src/components/Homepage/LendByCategory/LoanCategoriesSection.vue
```diff
@@ -48,6 +48,7 @@
 				:is-logged-in="isLoggedIn"
 				:is-visible="category.id === activeCategory"
 				:key="category.id + '-category'"
+				:row-number="categoryIds.indexOf(category.id) + 1 || null"
```
eddieferrer, line 51: @mcstover indexOf will return '-1' if not found, which will evaluate to 0 here and never null

 I don't think this matters.  because we have the array of backup categories so I'm not sure in what situation indexOf wouldn't find the category id.

mcstover: Thanks @eddieferrer updates forthcoming

PR #1990, opened 2020-08-10
Additional Storybook enhancements

eddieferrer:
Some functional improvements to mocking data. 

My thinking is that we can mock objects/arrays of data in the mock-data folder. Instead of mocking entire API responses.  We can then reuse these across different stories for different needs. 

For example, we can use these objects to mock apollo response. Like `.storybook/stories/LendByCategoryHomepage.stories.js`

Or we can use them to mock props, like in: `.storybook/stories/CheckoutReceipt.stories.js`

We can also then write any logic to process the data in the stories themselves or in utility functions such as the `arrayOfLoans` functions or the apollo mocks in `.storybook/stories/LendByCategoryHomepage.stories.js`

### DIFF
--- a/.storybook/stories/LendHomepageLoanCard.stories.js
+++ b/.storybook/stories/LendHomepageLoanCard.stories.js
@@ -69,7 +69,7 @@ export const Default = () => ({
 			return {
 				maxWidth: '476px',
 			}
-		}
+		},
 	},
 	template: `
 		<div :style="styling">
@@ -85,6 +85,12 @@ export const Default = () => ({
 				:loan="loan"
 				:percent-raised="percentRaised"
 			/>
+			<component is="style">
+				.lend-homepage-loan-card {
+					border-radius: 0.65rem;
+					box-shadow: 0 0.65rem 0.875rem 0.4375rem hsla(0,0%,60%,.1);
+				}
+   			 </component>
 		</div>
 	`,
 });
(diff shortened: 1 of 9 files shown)


.storybook/stories/LendHomepageLoanCard.stories.js
```diff
@@ -85,6 +85,12 @@ export const Default = () => ({
 				:loan="loan"
 				:percent-raised="percentRaised"
 			/>
+			<component is="style">
```
eddieferrer, line 88: This is a hack that basically inserts a style tag and allows for styling without the need to add a :style attribute to the component in the story

ryan-ludwig: oh weird! but neat :)

I think the part I don't like here is that it's opaque where .lend-homepage-loan-card class resides. I'm guessing it's the first node inside lend-homepage-loan-card component. One idea is to add a class in this story on the component, maybe `<lend-homepage-loan-card class="loan-card" ...>` and then inside your style tag `.loan-card { ... }`. This would tell the whole story of how you need to have this CSS applied to outer node of the component to get the border radius/box shadow effect.

eddieferrer: @ryan-ludwig good suggestion, updated

PR #1997, opened 2020-08-11
GROW-168 Google Pay

eddieferrer:
* Adds Google Pay checkout option for Braintree Drop In UI  based on experiment setting.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -55,6 +55,9 @@ module.exports = {
 			url: 'www.paypal.com',
 			environment: 'production'
 		},
+		googlePay: {
+			merchantId: 'f5swrg87wym8bx37'
+		}
 	},
 	server: {
 		port: 8888,
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -5,6 +5,7 @@
 				ref="braintreeDropInInterface"
 				:amount="amount"
 				flow="checkout"
+				:payment-types="paymentTypes"
 				@transactions-enabled="enableCheckoutButton = $event"
 			/>
 			<div id="dropin-button">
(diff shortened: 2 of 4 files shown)


config/index.js
```diff
@@ -55,6 +55,9 @@ module.exports = {
 			url: 'www.paypal.com',
 			environment: 'production'
 		},
+		googlePay: {
+			merchantId: 'f5swrg87wym8bx37'
```
mcstover, line 59: do we need to blank this out for sandbox?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -53,9 +68,40 @@ export default {
...
+			// applePay & googlePay payment methods experiment - EXP-GROW-168-Aug2020
+			const paymentMethodsExperiment = this.apollo.readFragment({
+				id: 'Experiment:braintree_dropin_apple_google',
+				fragment: experimentVersionFragment,
+			}) || {};
+			if (paymentMethodsExperiment.version === 'shown') {
+				this.paymentTypes = ['paypal', 'card', 'googlePay'];
+			}
+			// Fire Event for EXP-GROW-168-Aug2020
+			if (paymentMethodsExperiment.version && paymentMethodsExperiment.version !== 'unassigned') {
```
mcstover, line 92: Once deployed, please verify this is firing correctly. We normally want to reserve analytics events until mounted or after...

config/index.js
eddieferrer: @mcstover according to the docs and my testing, this is fine. It is required for production, but won't hurt anything if it's there for sandbox

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
eddieferrer: @mcstover good catch, fixed this

PR #2004, opened 2020-08-12
GROW-173 updated KvFlag component which can display SVG or sprited flags

ryan-ludwig:
Context:
* We need to be able to show circular flags for the GROW-165 epic. In the designs there is a list of small flags for all countries, and then a block which shows a handful of large flags.

Goals:
* Allow inlined SVG of flags for the handful of large flags.
* Provide a sprite for when you have many small flags on one page.
  * Note, I did try out an SVG sprite and the filesize ended up being over 4MB (not to mention the huge number of DOM nodes). There's a lot of complexity in some flags. PNG is a better choice here.
  * Sprited images should handle retina displays, but without downloading both normal and 2x files.
* Should handle square (or circle) flag aspect ratios
* Component should be able to be sized only using CSS.

Notes
* Since we needed a library of SVG flags, I audited several flag repos on Github. Several countries use different aspect ratios in their flags which makes consistent usage and spriting awkward. The most popular and maintained repo (https://github.com/lipis/flag-icon-css) took care of this, although sometimes their crop choices aren't my favorite (see USA flag in particular). They also have a set of cropped square flags which is ideal for our circle usage.
* I wrote a node script that generates the sprite for us based on the SVGs. Fun! Not sure where to put it. Maybe I should make a new repo/npm module that we add as a dependency which will call to generate the sprite for us at build time? 
* Another thought is to not add the flag SVGs to our assets/images/ folder and instead add https://github.com/lipis/flag-icon-css as a dependency, then inline them directly from node_modules. Or do the same with the new repo/npm module mentioned above.


ryan-ludwig:
Would love to get your initial thoughts on this component. Once we feel good about it I'll replace KvFlag with this and update it's usage in the site.

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,118 @@
+<template>
```
ryan-ludwig, line 1: Here is the KvFlag component

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,118 @@
...
+			default: '4x3', // 4x3 or 1x1
+		},
+		fromSprite: {
+			type: Boolean,
+			default: false
+		}
+	},
+	computed: {
+		flagSVG() {
+			return () => import(`~/flag-icon-css/flags/${this.aspectRatio}/${this.country}.svg`);
```
ryan-ludwig, line 42: This is how the inlined svgs are handled. Same ideas as KvIcon.

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,118 @@
...
... 'es-ca', 'es-ga', 'es', 'et', 'eu', 'fi', 'fj', 'fk', 'fm', 'fo', 'fr', 'ga', 'gb-eng', 'gb-nir', 'gb-sct', 'gb-wls', 'gb', 'gd', 'ge', 'gf', 'gg', 'gh', 'gi', 'gl', 'gm', 'gn', 'gp', 'gq', 'gr', 'gs', 'gt', 'gu', 'gw', 'gy', 'hk', 'hm', 'hn', 'hr', 'ht', 'hu', 'id', 'ie', 'il', 'im', 'in', 'io', 'iq', 'ir', 'is', 'it', 'je', 'jm', 'jo', 'jp', 'ke', 'kg', 'kh', 'ki', 'km', 'kn', 'kp', 'kr', 'kw', 'ky', 'kz', 'la', 'lb', 'lc', 'li', 'lk', 'lr', 'ls', 'lt', 'lu', 'lv', 'ly', 'ma', 'mc', 'md', 'me', 'mf', 'mg', 'mh', 'mk', 'ml', 'mm', 'mn', 'mo', 'mp', 'mq', 'mr', 'ms', 'mt', 'mu', 'mv', 'mw', 'mx', 'my', 'mz', 'na', 'nc', 'ne', 'nf', 'ng', 'ni', 'nl', 'no', 'np', 'nr', 'nu', 'nz', 'om', 'pa', 'pe', 'pf', 'pg', 'ph', 'pk', 'pl', 'pm', 'pn', 'pr', 'ps', 'pt', 'pw', 'py', 'qa', 're', 'ro', 'rs', 'ru', 'rw', 'sa', 'sb', 'sc', 'sd', 'se', 'sg', 'sh', 'si', 'sj', 'sk', 'sl', 'sm', 'sn', 'so', 'sr', 'ss', 'st', 'sv', 'sx', 'sy', 'sz', 'tc', 'td', 'tf', 'tg', 'th', 'tj', 'tk', 'tl', 'tm', 'tn', 'to', 'tr', 'tt', 'tv', 'tw', 'tz', 'ua', 'ug', 'um', 'un', 'us', 'uy', 'uz', 'va', 'vc', 've', 'vg', 'vi', 'vn', 'vu', 'wf', 'ws', 'xk', 'ye', 'yt', 'za', 'zm', 'zw']; // eslint-disable-line max-len
```
ryan-ludwig, line 22: It would be nice to import this from our custom node_module that also houses the SVGs and sprite.

ryan-ludwig:
Chatted with @mcstover, going to move the flag SVGs out of our repo, and just reference them from `node_modules/flag-icon-css`

src/components/Kv/KvFlag2.vue
mcstover: Or maybe a generic country util file that can hold the array or a map...

Also just for stating it, we found that the geocode>country>isoCode comes back as capital letters from the graphql api.

src/components/Kv/KvFlag2.vue
mcstover: I initially thought we might want a fallback image incase this fails for some reason. However, my main concern is that the layout isn't altered if this ends up blank and there is no type of "missing image" icon. A simple grey bg would be fine too.

ryan-ludwig: Added a background color. If the sprite or svg fails to load, KvFlag will still take up the space allowed but show a grey box.

emuvente:
@ryan-ludwig Why not use the flag svgs that we already have here? https://github.com/kiva/styleguide/tree/master/source/images/flags

ryan-ludwig:
Ooof, I did not know they existed.

mcstover:
@ryan-ludwig @emuvente May want to do a size comparison of those files vs using the new include.

I also very much like that we don't have to house these in our own repo as we will certainly need them elsewhere in the future and would much rather just add an npm dep instead of a couple hundred files...

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,119 @@
...
+					class="kv-flag__img"
+				/>
+			</template>
+			<span class="show-for-sr">{{ countryName }}</span>
+		</div>
+	</div>
+</template>
+
+<script>
+import { getCodes, getName } from 'country-list';
```
emuvente, line 22: Already noted this in the other PR, but I'll mention it here as well; the country-list package has slightly different country codes than we use, and it's also missing Kosovo, so we should account for that or use something else.

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,115 @@
...
+
+	&--4x3 {
+		.kv-flag__wrapper {
+			padding-bottom: 75%;
+		}
+
+		.kv-flag__img-sprite {
+			background-image: url('~flag-icon-css/flags/sprite/4x3/flag-sprite-32.png');
+
+			@include breakpoint(retina) {
```
ryan-ludwig, line 95: Did not previously know about this mixin from foundation. Handy :)

ryan-ludwig:
@eddieferrer @emuvente @mcstover This should be ready for review. To verify, pull down branch, `npm i`, `npm run storybook`, check KvFlag2 stories.

If this looks good, I'll do a follow up PR that renames KvFlag2 to KvFlag and updates usage in throughout the site.

src/components/Kv/KvFlag2.vue
```diff
@@ -0,0 +1,127 @@
...
+		},
+		/**
+		 * Inline the flag SVG into the markup instead of using a background image sprite
+		 * Use when showing a small amount of flags per page or if you need a larger flag than 32px wide.
+		* */
+		inlineSvg: {
+			type: Boolean,
+			default: false
+		}
+	},
```
emuvente, line 51: I think it would be good to add a cache key to this component to help this render really fast on the server. Something like:
```
import getCacheKey from '@/util/getCacheKey';
...
serverCacheKey: props => getCacheKey(`KvFlag-${props.country}-${props.aspectRatio}-${props.inlineSvg}`),
```

ryan-ludwig: Ooh, great idea! 

> I think it would be good to add a cache key to this component to help this render really fast on the server. Something like:
> 
> ```
> import getCacheKey from '@/util/getCacheKey';
> ...
> serverCacheKey: props => getCacheKey(`KvFlag-${props.country}-${props.aspectRatio}-${props.inlineSvg}`),
> ```

PR #2006, opened 2020-08-13
use apollo preFetch for contentful global promo to prevent full page …

mcstover:
…repaint after async load.

### DIFF
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -65,7 +65,7 @@ module.exports = {
 			},
 			{
 				// Inject styles from the /pages/ directory as <style> tags
-				test: /\/pages\/.+\.scss$|\/banners\/GenericPromoBanner\.scss$/,
+				test: /\/pages\/.+\.scss$/,
 				use: [ "thread-loader", "vue-style-loader"].concat(styleLoaders)
 			},
 			{
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -65,7 +65,7 @@ module.exports = {
 			},
 			{
 				// Inject styles from the /pages/ directory as <style> tags
-				test: /\/pages\/.+\.scss$/,
+				test: /\/pages\/.+\.scss$|\/banners\/GenericPromoBanner\.scss$/,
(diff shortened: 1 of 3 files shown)


mcstover:
This is a good start and server renders the content properly. However, the optionality in GlobalPromotionalBannerContentful.vue with regard to which component is rendered prevents the css from being inlined.

One option I'm testing is selectively added the GenericPromoBanner to our vue-style-loader config.
(https://github.com/kiva/ui/blob/master/build/webpack.base.conf.js#L68)
If this works it could be useful for some other items as well...

mcstover:
Ok so this worked and ads a relatively small file. This case is a bit special b/c we almost always have content in this Global Promo space so in my opinion it's an acceptable addition.

image

Server rendered header (js disabled)
image

mcstover:
> This is a good start and server renders the content properly. However, the optionality in GlobalPromotionalBannerContentful.vue with regard to which component is rendered prevents the css from being inlined.
> 
> One option I'm testing is selectively added the GenericPromoBanner to our vue-style-loader config.
> (https://github.com/kiva/ui/blob/master/build/webpack.base.conf.js#L68)
> If this works it could be useful for some other items as well...

Alternatively, we remove the v-if on the components in GlobalPromotionalBannerContentful.vue and push a prop so visibility is handled internally, within GenericPromo...

mcstover:
@emuvente @eddieferrer This seems to produce the correct behavior on the homepage, lend-by-category, and category pages.

eddieferrer:
> > This is a good start and server renders the content properly. However, the optionality in GlobalPromotionalBannerContentful.vue with regard to which component is rendered prevents the css from being inlined.
> > One option I'm testing is selectively added the GenericPromoBanner to our vue-style-loader config.
> > (https://github.com/kiva/ui/blob/master/build/webpack.base.conf.js#L68)
> > If this works it could be useful for some other items as well...
> 
> Alternatively, we remove the v-if on the components in GlobalPromotionalBannerContentful.vue and push a prop so visibility is handled internally, within GenericPromo...
> 
> 🤦 Deleted my last comment b/c I didn't have js turned off and that file was still not inlined... will test some other options...

I believe replacing the v-if with v-show might have worked as well.

PR #2016, opened 2020-08-17
Updated graphql tools dependency, modified fetchSchema function for n…

eddieferrer:
…ew graphql tools api


eddieferrer:
Should take care of: https://github.com/kiva/ui/pull/2007

PR #2017, opened 2020-08-17
Implements KvFlag2 throughout site. Renames KvFlag2 to KvFlag.

ryan-ludwig:
GROW-173


src/pages/LendingStats/ItemList.vue
```diff
@@ -68,10 +76,15 @@ export default {
 		text-overflow: ellipsis;
 	}
 
+	.item-flag,
 	.item-icon {
-		width: 1.5rem;
-		height: 0.7777rem;
+		width: rem-calc(16);
```
emuvente, line 81: That looks like a smaller width than what was there before (1.5rem would be be rem-calc(24))

ryan-ludwig: These were kindof funny. Because of their fixed height, it was making the width smaller than 1.5rem. 16px seemed about the equivalent.

PR #2025, opened 2020-08-18
VUE-372 Fix for 500 error on receipt page

eddieferrer:
* Prevents 500 error if receipt query fails for some reason

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -116,11 +116,18 @@ export default {
 			// The default empty object and the v-if will prevent the
 			// receipt from rendering in the rare cases this query fails.
 			// But it will not throw a server error.
-			this.receipt = _get(data, 'shop.receipt', {});
+			this.receipt = _get(data, 'shop.receipt');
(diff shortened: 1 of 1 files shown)


eddieferrer:
If the receipt query fails, assuming the lender query does not, the user will see the message saying confirmation sent to email

src/pages/Thanks/ThanksPage.vue
```diff
@@ -111,8 +112,13 @@ export default {
...
+
+			// The default empty object and the v-if will prevent the
+			// receipt from rendering in the rare cases this query fails.
+			// But it will not throw a server error.
+			this.receipt = _get(data, 'shop.receipt', {});
+			const loansResponse = _get(this.receipt, 'items.values', []);
+			this.loans = loansResponse
 				.filter(item => item.basketItemType === 'loan_reservation')
 				.map(item => item.loan);
 		}
```
mcstover, line 124: @eddieferrer could we please console an error if the receipt and/or the lender is missing at the end of this result. For the lender we can prob just key off the presence or lack thereof `userAccount`.

PR #2051, opened 2020-08-27
VUE-362 VUE-396 Remove lodash omit and pick

eddieferrer:
* Removed lodash methods omit and pick in favor of native JS

VUE-362
VUE-396

### DIFF
--- a/src/api/fixtures/LoanSearchCriteria.js
+++ b/src/api/fixtures/LoanSearchCriteria.js
@@ -1,4 +1,3 @@
-import _pick from 'lodash/pick';
 import LoanSearchFilters, {
 	filtersAreEqual,
 	getCacheableFilters,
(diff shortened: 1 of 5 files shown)


src/api/fixtures/LoanSearchCriteria.js
```diff
@@ -29,22 +28,15 @@ export function getCacheableCriteria(criteria) {
...
 		filters: getInputFilters(filters),
 	};
 }
 
 // Return criteria that can be used in a loan search
-export function getSearchableCriteria({ filters, ...criteria }) {
-	const cleanCriteria = _pick(criteria, [
-		'queryString'
-	]);
+export function getSearchableCriteria({ filters }) {
```
eddieferrer, line 38: This function is referenced in a single place, and there it is destructured to only use filters

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -547,11 +545,13 @@ export default {
 			const transactionData = {
 				transactionId: numeral(transactionId).value(),
 				itemTotal: this.totals.itemTotal,
-				loans: _map(this.loans, loan => {
-					return _pick(loan, ['__typename', 'id', 'price']);
+				loans: this.loans.map(loan => {
+					const { __typename, id, price } = loan;
+					return { __typename, id, price };
```
mcstover, line 550: did you see any difference in the return value if there are no loans present?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -547,11 +545,13 @@ export default {
...
 				}),
-				donations: _map(this.donations, donation => {
-					return _pick(donation, ['__typename', 'id', 'price']);
+				donations: this.donations.map(donation => {
+					const { __typename, id, price } = donation;
+					return { __typename, id, price };
```
mcstover, line 554: Same here...So long as they are gracefully handled when missing these should be fine.

src/pages/Checkout/CheckoutPage.vue
eddieferrer: These will both return an empty array

PR #2055, opened 2020-08-27
GROW-170 Cause selector component

ryan-ludwig:
*Screenshots from Aaron's designs below*

Component needs to be able to be used as a radio:
image

A checkbox:
image

And an icon only
image


ryan-ludwig:
I went back and forth a lot over how to structure this component. Should it just be the icon image? Maybe we have a few components instead of 1? Should it be a KvComponent? I'm still open to ways to structure this better.

mcstover:
@ryan-ludwig I think it's fine for this to be specialized as you've done it. It will most likely only live in 1-2 specialized contexts anyhow.

.storybook/stories/KvCauseSelector.stories.js
```diff
@@ -0,0 +1,125 @@
+import KvCauseSelector from '@/components/Kv/KvCauseSelector';
+
+export const causeList = [
```
eddieferrer, line 3: If you don't export this, it won't be picked up as a story and you don't have to exclude it below. It should just be a global variable in this file.

src/components/Kv/KvCauseSelector.vue
```diff
@@ -0,0 +1,250 @@
...
+import { paramCase } from 'change-case';
+import KvIcon from '@/components/Kv/KvIcon';
+
+const imageRequire = require.context('@/assets/images/cause-selector/', true);
+
+export default {
+	components: {
+		KvIcon
+	},
+	model: {
```
eddieferrer, line 64: Why do we need this here? I think this only applies if you are using v-model directly

.storybook/stories/KvCauseSelector.stories.js
ryan-ludwig: Oh, I'm using it in a WIP story for KvCarousel right now

eddieferrer: I see, if you're going to be using it in multiple stories, then maybe lets add it as a separate file in `mock-data`

src/components/Kv/KvCauseSelector.vue
ryan-ludwig: Good catch, left over from an earlier iteration.

PR #2061, opened 2020-08-31
VUE-389 Auto Deposit What To Expect

eddieferrer:
* AutoDeposit what to expect, general page layout and CSS


eddieferrer:
screencapture-dev-vm-01-kiva-org-auto-deposit-new-2020-08-31-10_50_05

PR #2082, opened 2020-09-03
GROW-175, GROW-176 Getting started - causes and places

ryan-ludwig:


src/pages/GetStarted/GetStartedCauses.vue
```diff
@@ -1,31 +1,191 @@
...
+					checked: false,
+				},
+				{
+					id: 4,
+					name: 'education',
+					checked: false,
+				},
+				{
+					id: 0, // TODO: Couldn't find an id for technology
+					name: 'technology',
```
ryan-ludwig, line 90: @emuvente Do you know what the id is for "technology" category?

emuvente: I can't find one for htat either, so I think we'll have to make it.

mcstover: I think we need to figure out what the loan search criteria is supposed to be for this. Then if you do create a new Loan Channel, please add it to the query map too (https://github.com/kiva/ui/blob/master/src/plugins/loan-channel-query-map.js)

emuvente: I believe for the technology category it's just the "Technology" tag

src/components/Kv/KvCauseSelector.vue
```diff
@@ -45,7 +45,7 @@
 						alt=""
 					>
 				</div>
-				<div class="kv-cause-selector__text">{{ cause }}</div>
+				<div class="kv-cause-selector__text">{{ capitalizeFirstLetter(cause) }}</div>
```
eddieferrer, line 48: I believe you could also use: `{{ cause|changeCase('titleCase') }}`

ryan-ludwig: > I believe you could also use: `{{ cause|changeCase('titleCase') }}`

I think that would change 'COVID-19' to 'Covid-19' unfortunately

eddieferrer: Sign, I was afraid of that.

eddieferrer: @ryan-ludwig since all the causes are one word, we can just use CSS `text-transform: capitalize;`

mcstover (approved):

emuvente (approved):

PR #2083, opened 2020-09-03
VUE-390 Auto Deposit Sign Up Form

eddieferrer:
* Implemented auto deposit sign up form.
* Fix other forms with day of month to avoid GQL validation errors in federation

### DIFF
--- a/src/components/AutoDeposit/AutoDepositSignUpForm.vue
+++ b/src/components/AutoDeposit/AutoDepositSignUpForm.vue
@@ -1,0 +1,475 @@
+<template>
+	<form
+		class="ad-sign-up-form"
+		@submit.prevent
+		novalidate
+	>
+		<!-- Amount and Day Input -->
+		<div class="row column align-middle">
+			<span class="ad-sign-up-form__deposit-text">
+				Deposit
+			</span>
+			<label
+				class="show-for-sr"
+				:class="{ 'error': $v.adAmount.$invalid }"
+				for="amount"
+			>
+				Amount
+			</label>
+			<kv-currency-input class="text-input" id="amount" v-model="adAmount" />
+			<span class="ad-sign-up-form__deposit-text">
+				each month on the
+			</span>
+			<label class="show-for-sr" :class="{ 'error': $v.$invalid }" :for="dayOfMonth">
+				Day of the Month
+			</label>
+			<input v-if="isDayInputShown"
+				@blur="hideDayInput()"
+				class="text-input"
+				id="dayOfMonth"
+				type="number"
+				placeholder=""
+				required
+				min="1"
+				max="31"
+				v-model.number="dayOfMonth"
+			>
+			<button
+				class="ad-sign-up-form__button--ordinal-day"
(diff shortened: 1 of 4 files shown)


eddieferrer:
This is what it looks like so far (this PR doesnt include the drop in, pictured here)
screencapture-dev-vm-01-kiva-org-auto-deposit-new-2020-09-03-15_39_21

src/components/AutoDeposit/AutoDepositSignUpForm.vue
```diff
@@ -0,0 +1,475 @@
...
+			donationOptionSelected: '15',
+			isDonationOptionsDirty: false,
+			showDropInPayments: true,
+			isLoggedIn: false,
+		};
+	},
+	mixins: [
+		validationMixin,
+	],
+	validations: {
```
eddieferrer, line 208: I double checked these validation with Leo

src/components/AutoDeposit/AutoDepositSignUpForm.vue
```diff
@@ -0,0 +1,475 @@
...
+
+// import AutoDepositDropInPaymentWrapper from '@/components/AutoDeposit/AutoDepositDropInPaymentWrapper';
+import IconPencil from '@/assets/icons/inline/pencil.svg';
+import KvCheckbox from '@/components/Kv/KvCheckbox';
+import KvCurrencyInput from '@/components/Kv/KvCurrencyInput';
+import KvDropdownRounded from '@/components/Kv/KvDropdownRounded';
+
+const pageQuery = gql`query monthlyGoodSetupPageControl {
+    general {
+		braintreeDropInFeature: uiConfigSetting(key: "feature.braintree_dropin") {
```
mcstover, line 170: I don't think we need this seeing as there wouldn't be any type of fallback...also I wouldn't want it to collide with other drop-in usage elsewhere. We would need to do a series of convoluted steps to be able to redirect back to the old page which I don't think are worth it. Let's just roll with the new new.

src/components/AutoDeposit/AutoDepositSignUpForm.vue
```diff
@@ -0,0 +1,457 @@
...
+				max="31"
+				v-model.number="dayOfMonth"
+			>
+			<button
+				class="ad-sign-up-form__button--ordinal-day"
+				@click="isDayInputShown = true"
+				v-if="!isDayInputShown"
+			>
+				<strong>{{ dayOfMonth | numeral('Oo') }}</strong>
+				<icon-pencil class="icon-pencil" />
```
ryan-ludwig, line 43: The end result will be pretty much the same but I think you can use` <kv-icon name="pencil">` for this

mcstover (approved):

src/components/AutoDeposit/AutoDepositSignUpForm.vue
eddieferrer: Good catch, I used this in a couple of places. I'll fix in a follow up PR.

PR #2085, opened 2020-09-04
Drop in pencil icon improvement

eddieferrer:
* Replace IconPencil for KvIcon usage. 
* Fixed bug where if form was invalid, day of month input would have red outline, it should only outline if dayOfMonth itself is invalid.


eddieferrer:
Pending rebase

ryan-ludwig (approved):
Thanks for doing that!

mcstover (approved):
nice clean up and move to the icon component

PR #2087, opened 2020-09-07
VUE-405 giant flag fix

ryan-ludwig:
On the funded loan page KvFlag's default 100% width was overriding the 1rem width set on the fundedBorrowerProfile.vue page.
The order the CSS is bundled can cause this. I feel like we don't usually see it in prod bundles though :/
image

I also noticed the country names in assistive text were coming in as undefined so fixed that in the flag-icon-css repo and bumped the package-lock here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19109,7 +19109,7 @@
 			}
 		},
 		"flag-icon-css": {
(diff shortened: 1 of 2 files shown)


mcstover (commented):
👍

PR #2092, opened 2020-09-09
Enable usage of serverPrefetch

emuvente:
I wanted to use Vue's `serverPrefetch` method to get some data for a component after the component instance was created. That allows me to use the instance's prop values as variables for queries, something we can't do with our home-grown apollo prefetch method.

I think of the `serverPrefetch` method the same as being in the browser and showing a loading state while fetching data to render a component (e.g. how classic /lend behaves with loading the page first and then showing a loading spinner while the loans are fetched). The difference is that the data fetching happens server-side and the final html isn't created until after that completes, allowing the fully-loaded and rendered page to be sent from the server.

To enable the usage of this in our app, we only needed to change when we set the context state using the `context.rendered` callback (see https://ssr.vuejs.org/guide/data.html#final-state-injection).

### DIFF
--- a/src/server-entry.js
+++ b/src/server-entry.js
@@ -108,19 +108,27 @@ export default context => {
 					kvAuth0,
 				});
 			}).then(() => {
-				if (isDev) logFormatter(`data pre-fetch: ${Date.now() - s}ms`);
-				// After all preFetch hooks are resolved, our store is now
-				// filled with the state needed to render the app.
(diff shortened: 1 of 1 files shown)


mcstover (approved):

eddieferrer:
👍  this is great, I was messing around with this a while back, and couldn't get it to work properly, I missed the detail of `context.rendered` in the server entry

eddieferrer (approved):

PR #2103, opened 2020-09-11
GROW-178 Get Started loan results

emuvente:
Just getting a draft pr up of this for initial review because it's turning pretty big


src/components/LoanCards/RecommendedLoanCard.vue
```diff
@@ -0,0 +1,486 @@
...
+
+$largest: rem-calc(1250);
+
+.rec-loan-card {
+	@extend %nested-column-flex;
+
+	flex-shrink: 0;
+	min-width: rem-calc(230);
+	width: 230/320 * 100%;
+	max-width: rem-calc(280);
```
ryan-ludwig, line 256: One thought on these loan cards is let the parent handle the widths (remove max-width, set width to always be 100%). We could then drop them into our standard grid use our `small-12 large-6 columns` type syntax do the work.

src/components/Kv/KvFlag.vue
```diff
@@ -56,7 +57,12 @@ export default {
 			return `Flag of ${getNameByCode(this.country)}`;
 		},
 		flagSVG() {
-			return () => import(`~/flag-icon-css/flags/${this.aspectRatio}/${this.country.toLowerCase()}.svg`);
+			// Pulling these out here so that Vue registers them as reactive.
```
ryan-ludwig, line 60: 👍

.storybook/stories/RecommendedLoanCard.stories.js
```diff
@@ -0,0 +1,82 @@
...
+export const Default = () => ({
+	mixins: [apolloMockMixin],
+	components: { RecommendedLoanCard },
+	template: '<recommended-loan-card :loan-id="1998250" />'
+});
+
+export const Loading = () => ({
+	mixins: [apolloLoadingMixin],
+	components: { RecommendedLoanCard },
+	template: '<recommended-loan-card :loan-id="1998250" />'
```
ryan-ludwig, line 81: The loading state on this looks awesome!

src/components/Kv/KvLoadingParagraph.vue
```diff
@@ -0,0 +1,39 @@
...
+		},
+	},
+	serverCacheKey: ({ numLines }) => getCacheKey(`KvLoadingParagraph-${numLines}`),
+	components: {
+		KvLoadingPlaceholder,
+	},
+};
+</script>
+
+<style lang="scss">
```
ryan-ludwig, line 26: Can we keep this scoped as well as the styles in KvLoadingPlaceholder? In this template you should be able to add a class like `<kv-loading-placeholder class="placeholder"> `and set the same styles on `.placeholder` directly.

src/components/Kv/KvLoadingPlaceholder.vue
```diff
@@ -0,0 +1,37 @@
+<template functional>
```
ryan-ludwig, line 1: These KvLoading components are sweet. When time can you add some .stories please.

ryan-ludwig:
Lookin great!

.storybook/stories/KvButton.stories.js
```diff
@@ -20,6 +20,11 @@ export const DefaultSmallest = () => ({
 	template: '<kv-button class="smallest">Button Smallest</kv-button>'
 });
 
+export const Rounded = () => ({
+	components: { KvButton },
+	template: '<kv-button class="rounded">Button</kv-button>'
```
ryan-ludwig, line 25: #sadpanda  I wish design would choose whether our buttons are rounded or our buttons are not rounded. I don't see a compelling reason to have 2 options.

mcstover: It is a little frustrating but overall I think what @emuvente has done here and has started could help us chart a way forward for adding in new styles alongside our existing styles too.

mcstover:
@emuvente I like the way this is going! Thanks for charting a way forward with a mixin driven approach to the loan cards.

.storybook/stories/RecommendedLoanCard.stories.js
```diff
@@ -0,0 +1,82 @@
...
+					reservedAmount: "0.00",
+					isExpiringSoon: false
+				},
+				plannedExpirationDate: "2020-09-10T19:30:13Z"
+			}
+		}
+	}
+};
+
+const apolloMockMixin = {
```
eddieferrer, line 42: Instead of defining this, we can just reuse: `.storybook/mixins/apollo-story-mixin.js`

src/pages/GetStarted/GetStartedResults.vue
```diff
@@ -118,14 +152,113 @@ export default {
...
+				loans(limit: $limit) {
+					values {
+						id
+					}
+				}
+			}
+		}`,
+		result(result) {
+			// TODO: error handling
+			let loanValues = result.data?.general?.lendingPreferences?.loans?.values || [];
```
eddieferrer, line 212: Have we confirmed that our babel setup transpiles this correctly for IE11?

.storybook/stories/RecommendedLoanCard.stories.js
emuvente: I didn't use that initially because it doesn't mock any data, but now that I know what my needs are I can adapt that mixin to include mocked data.

src/components/Kv/KvLoadingParagraph.vue
emuvente: 👍

src/components/Kv/KvLoadingPlaceholder.vue
emuvente: 👍

src/components/LoanCards/RecommendedLoanCard.vue
emuvente: 👍 

I initially started with that, but the styles were different enough that I stopped using the row/column classes and just defined my own, but now that I've got it this far I think there's a different way I can achieve the results I'm looking for using the row/column classes.

src/pages/GetStarted/GetStartedResults.vue
emuvente: Not yet, but I'll check that before moving this out of draft

.storybook/stories/RecommendedLoanCard.stories.js
emuvente: Thanks!

src/pages/GetStarted/GetStartedResults.vue
emuvente: Confirmed, babel is transpiling this! This particular line is transformed into this:
```
var g=(null===(e=t.data)||void 0===e||null===(n=e.general)||void 0===n||null===(s=n.lendingPreferences)||void 0===s||null===(a=s.loans)||void 0===a?void 0:a.values)||[];
```

mcstover: 💯

src/components/Kv/KvButton.vue
```diff
@@ -57,6 +57,10 @@ export default {
 	font-weight: $button-font-weight;
 	box-shadow: $button-primary-shadow;
 
+	&.rounded {
+		border-radius: rem-calc(10);
+	}
```
mcstover, line 62: @ryan-ludwig At some point in the near future we may want to collect the various "rounded" settings we have around Ui and try to establish a couple of prescribed options...

mcstover (approved):
🚢

There are somethings we'll need to add to this loan card in the future to make it useful in other contexts...

eddieferrer (approved):

src/components/Kv/KvLoadingParagraph.vue
ryan-ludwig: I think we should figure out how we'd like to handle managing style scope as a team at some point soon.

ryan-ludwig (approved):

PR #2104, opened 2020-09-13
GROW-223: basket verification in new UI

remoun:
https://kiva.atlassian.net/browse/GROW-223

* Used the design and copy linked from the ticket.
* One thing I added is some text on the checkout page when the basket is verified. I'm not attached to this, but I found it helpful to differentiate the states.
* Another thing I implemented, which is not specified in the design, is hiding the checkout buttons (but not the login button) when verification is required, especially since the verification lightbox is dismissable. [GROW-243](https://kiva.atlassian.net/browse/GROW-243) tracks making verification block checkout on the backend.

verification-required
verification-pending
verification-verified

### DIFF
--- a/src/components/Checkout/BasketVerification.vue
+++ b/src/components/Checkout/BasketVerification.vue
@@ -1,0 +1,16 @@
+<template>
+	<div>
+	</div>
+</template>
+
+<script>
+export default {
+	inject: ['apollo'],
+	data() {
+		return {
+		};
+	},
+	methods: {
+	},
+};
+</script>
--- a/src/components/Checkout/BasketVerification.vue
+++ b/src/components/Checkout/BasketVerification.vue
@@ -64,6 +64,7 @@ export default {
 		query: getVerificationState,
 		result({ data }) {
 			this.setState(data?.shop?.basketVerificationState || this.verificationState);
+			this.email = data?.my?.userAccount?.email || this.email || '';
 		},
 	},
 	computed: {
@@ -79,7 +80,6 @@ export default {
 		},
 		send() {
 			this.setState('pending');
-			// TODO: Set this.email!!!
 			this.apollo.mutate({ mutation: startBasketVerificationMutation });
 		},
 	},
--- a/src/components/Checkout/BasketVerification.vue
+++ b/src/components/Checkout/BasketVerification.vue
@@ -4,13 +4,33 @@
 </template>
 
 <script>
+import getVerificationState from '@/graphql/query/checkout/basketVerificationState.graphql';
+
 export default {
 	inject: ['apollo'],
 	data() {
 		return {
+			verificationState: '',
 		};
 	},
+	apollo: {
+		preFetch: true,
+		query: getVerificationState,
+		result({ data }) {
+			this.setState(data?.shop?.basketVerificationState || this.verificationState);
+		},
+	},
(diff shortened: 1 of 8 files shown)


mcstover (approved):
Overall I think this looks good @remoun 

Please work with Ruby and @emuvente as I believe there is actually a slightly different design to be applied in this scenario. Ruby can help guide on what the interface should show.

src/components/Checkout/BasketVerification.vue
```diff
@@ -0,0 +1,104 @@
...
+	created() {
+		// Create an observer for changes to the basket verification states
+		this.apollo.watchQuery({ query: getVerificationState })
+			.subscribe({
+				next: ({ data }) => {
+					const state = _get(data, 'shop.basketVerificationState', this.verificationState);
+					this.setState(state);
+				},
+				error: logReadQueryError,
+			});
```
emuvente, line 78: You can use the apollo option block for this:
```
apollo: {
	preFetch: true,
	query: getVerificationState,
	result({ data }) {
		const state = _get(data, 'shop.basketVerificationState', this.verificationState);
		this.setState(state);
	},
}
```

src/components/Checkout/BasketVerification.vue
```diff
@@ -0,0 +1,104 @@
...
+	},
+	methods: {
+		setState(state) {
+			this.verificationState = state;
+			this.$emit('verification-state', this.verificationState);
+			this.$emit('verification-required', this.verificationRequired);
```
emuvente, line 84: We usually have parent components feed state to child components rather than the other way around. I'm curious how this will work with the server-side render, because that's not clear to me. I think this should be ok, but I want to test some things out with it.

src/components/Checkout/BasketVerification.vue
remoun: Nice, thanks!

src/components/Checkout/BasketVerification.vue
mcstover: I'm guessing this will just fail silently on the server, or maybe work intermittently, depending on timing of other operations... It's not promised based and has no mechanism to force the server renderer to wait on it.

I don't think there is a need to force that as it should run in the client as well...

src/components/Checkout/BasketVerification.vue
```diff
@@ -0,0 +1,99 @@
...
+		},
+		send() {
+			this.setState('pending');
+			// TODO: Set this.email!!!
```
emuvente, line 83: The email can be fetched in graphql from `my.userAccount.email`

emuvente:
This also needs the changes to the kiva credit line item that should now have a "Apply Kiva Credit" button.
Additionally, It looks like this is showing the lightbox immediately when the page opens, but it shouldn't be showing it until the user clicks that apply button.

src/components/Checkout/BasketVerification.vue
remoun: Added that. Thanks!

remoun:
I couldn't get the Kiva Credit to be not applied yet visible on the new Checkout page, so I'm punting on that; filed [GROW-262](https://kiva.atlassian.net/browse/GROW-262) to track that.

emuvente (approved):
👍

PR #2105, opened 2020-09-13
GROW-240 banner

ryan-ludwig:
Bump storybook version
Build and animate Swashie
Starting point of KvProgressCircle
Allow theming for KvProgressBar
Mostly duplicate AppealBanner component as AppealBanner15 and add to 15year story.


src/pages/15Years/15Years.vue
```diff
@@ -1,33 +1,36 @@
...
-					<fifteen-years-sponsors />
-				</section>
-				<section>
-					<fifteen-years-how-kiva-works />
-				</section>
-			</fifteen-years-styles>
-		</div>
-	</www-page>
+	<div>
+		<appeal-banner-15 :appeal-banner-content="{}" />
```
ryan-ludwig, line 3: I don't really understand how appeal-banner works, but i'm pretty sure this isn't where it should go.

mcstover: Correct. The global promo banner and any type of appeal is selected in the GlobalPromotionalBannerContentful.vue file. Selection is based on the first found Banner from the contentful setting AND the type of the Banner.

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
```diff
@@ -0,0 +1,562 @@
...
+	</div>
+</template>
+
+<script>
+import _forEach from 'lodash/forEach';
+import _get from 'lodash/get';
+import numeral from 'numeral';
+import store2 from 'store2';
+
+import appealBannerQuery from '@/graphql/query/appealBanner.graphql';
```
mcstover, line 105: The query in the global promo banner will get the setting and banner from contentful. We'll need to create a new query for this file that uses the new graphql query we've yet to build.

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
```diff
@@ -0,0 +1,562 @@
+<template>
```
mcstover, line 1: Lots of cleanup can be done in this file.

src/pages/15Years/15Years.vue
mcstover: So you can just remove it here as in will be included with www-page

mcstover (approved):
Thanks for pushing it forward some more!

PR #2107, opened 2020-09-14
GROW-176 Get Started Places

eddieferrer:
* Implemented get started places page


eddieferrer:
screencapture-dev-vm-01-kiva-org-get-started-places-2020-09-14-16_32_23

src/pages/GetStarted/GetStartedPlaces.vue
```diff
@@ -214,9 +240,9 @@ export default {
 		clearFilter() {
 			this.filterTerm = '';
 		},
-		selectAllCountries() {
-			this.selectAll = !this.selectAll;
-			this.countryList.forEach(country => {
+		toggleAllCountries(state) {
+			this.selectAll = state;
+			this.filteredCountryList.forEach(country => {
```
ryan-ludwig, line 245: I believe this causes a bug in the select all functionality. If you filter for "Asia", hit "select all", then clear the Asia filter the asia countries are no longer selected (since you're now viewing this.countryList). If you type "as" you'll see the asia countries selected again (since you're viewing this.filteredCountryList where those items are selected).

One idea is to 
this.filteredCountryList.forEach(country => {
 

 
I believe this causes a bug in the select all functionality. If you filter for "Asia", hit "select all", then clear the Asia filter the asia countries are no longer selected (since you're now viewing this.countryList). If you type "as" you'll see the asia countries selected again (since you're viewing this.filteredCountryList where those items are selected).

ryan-ludwig:
Looks good! Just one hitch on the select all functionality.

src/pages/GetStarted/GetStartedPlaces.vue
eddieferrer: Good catch 👍

eddieferrer:
I've updated the selection behavior, I believe it should be working as intended. (Select all selects all the visible/filtered countries)

emuvente (approved):
👍

PR #2113, opened 2020-09-16
GROW-177 Edit Preferences Results

eddieferrer:
* Edit preferences block on results page.
* Capitalize causes on cause selection page

Screen Shot 2020-09-16 at 12 16 56 PM
Screen Shot 2020-09-16 at 12 16 41 PM

There is a chance that this page can be reached without causes and/or places, see screenshot ^

### DIFF
--- a/src/components/GetStarted/EditPreferences.vue
+++ b/src/components/GetStarted/EditPreferences.vue
@@ -1,0 +1,295 @@
+<template>
+	<div class="edit-preferences">
+		<h2 class="edit-preferences__header">
+			Looking for different loans?
+		</h2>
+		<p class="edit-preferences__sub-header">
+			Change your preferences and find more borrowers to support.
+		</p>
+		<div class="row">
+			<div class="small-12 large-6 columns">
+				<!-- Causes Card -->
+				<div class="edit-preferences__card" v-if="causes.length > 0">
+					<div class="edit-preferences__icon-wrapper">
+						<span class="edit-preferences__icon-background">
+							<kv-icon
+								class="edit-preferences__check-icon"
+								name="checkmark"
+							/>
+						</span>
+					</div>
+					<div class="edit-preferences__content-wrapper">
+						<div class="row collapse text-left align-middle">
+							<div class="small-6 large-7 xlarge-6 columns">
+								<h2 class="edit-preferences__content-wrapper--title">
+									Your {{ causes.length }} causes
+								</h2>
+								<router-link
+									class="smallest"
+									to="/get-started"
+								>
+									Change causes &#8594;
+								</router-link>
+							</div>
+							<div class="small-6 large-5 xlarge-6 columns">
+								<ul class="edit-preferences__list">
+									<li
+										class="edit-preferences__list-item"
+										v-for="(cause, index) in causesTrimmed"
+										:key="index"
+									>
+										<kv-cause-selector
+											:cause="cause.name"
+											:as-icon="true"
+										/>
+									</li>
+									<li v-if="causes.length > 3" class="edit-preferences__list-item--counter">
+										<strong>+{{ causes.length - 2 }}</strong>
+									</li>
+								</ul>
+							</div>
+						</div>
+					</div>
+				</div>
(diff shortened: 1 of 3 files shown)


src/pages/GetStarted/GetStartedCauses.vue
```diff
@@ -263,6 +263,10 @@ export default {
 	&__summary {
 		max-width: 30.5rem;
 		margin: 0 0 2.5rem auto;
+
+		::v-deep b {
```
eddieferrer, line 267: TIL that you have to use v-deep to target selectors in v-html blocks

src/pages/GetStarted/GetStartedResults.vue
```diff
@@ -1,16 +1,17 @@
...
+
 		<div class="loan-results">
 			<template v-if="totalCount > 0">
 				<h1 class="loan-results__headline">
 					Here {{ countVerb }} {{ countNumber }} {{ countPeople }} you can support!
 				</h1>
 				<p class="loan-results__tagline">
 					In fact, we found {{ totalCount }} {{ countPeople }} that we think you’ll love to lend to.
-					<router-link>
+					<!-- <router-link>
```
eddieferrer, line 12: This throws an error for missing required 'to' prop. Commented out until we figure out where this link should go to.

src/components/GetStarted/EditPreferences.vue
```diff
@@ -0,0 +1,295 @@
...
+							<kv-icon
+								class="edit-preferences__check-icon"
+								name="checkmark"
+							/>
+						</span>
+					</div>
+					<div class="edit-preferences__content-wrapper">
+						<div class="row collapse text-left align-middle">
+							<div class="small-6 large-7 xlarge-6 columns">
+								<h2 class="edit-preferences__content-wrapper--title">
```
ryan-ludwig, line 24: Minor, but in BEM-land the '--' in a classname indicates a "modifier" class. So this would be saying you have a special version of the element, "content-wrapper" (that maybe you want to style differently than the normal content-wrapper).

A single hyphen would be a way to keep regular "element" status, like "content-wrapper-title". Or maybe just "content-title".

http://getbem.com/naming/
http://getbem.com/faq/#css-nested-elements

src/components/GetStarted/EditPreferences.vue
```diff
@@ -0,0 +1,295 @@
...
+										class="edit-preferences__list-item"
+										v-for="(cause, index) in causesTrimmed"
+										:key="index"
+									>
+										<kv-cause-selector
+											:cause="cause.name"
+											:as-icon="true"
+										/>
+									</li>
+									<li v-if="causes.length > 3" class="edit-preferences__list-item--counter">
```
ryan-ludwig, line 46: Same BEM comment as above

ryan-ludwig (approved):
Looks good!

src/components/GetStarted/EditPreferences.vue
eddieferrer: 👍

src/components/GetStarted/EditPreferences.vue
eddieferrer: I think this one is ok, it is a modifier on the list item.

src/pages/GetStarted/GetStartedResults.vue
emuvente: 👍

src/components/GetStarted/EditPreferences.vue
ryan-ludwig: OK gotcha. Modifier classes are also always in addition to the normal class. So it'd end up being 
`class="edit-preferences__list-item edit-preferences__list-item--counter"`
Where your --counter CSS would be anything that overrides or is in addition to list-item.

It does feel a little annoying to double up on the classes. But also kinda nice once you're used to it.
The "Under the Hood" section shows a good example: http://getbem.com/introduction/

emuvente (approved):
👍

PR #2114, opened 2020-09-16
Grow 171

BoulderBrains:
Start of grow-171


eddieferrer:
Screen Shot 2020-09-16 at 4 05 12 PM

eddieferrer:
I kept both the right and left arrows, the design seems to imply that only 1 arrow is shown at a time, but I dont think we should make that change...

@ryan-ludwig thanks for your work on the carousel, and the cause carousel example, made this very straight forward. Also makes the component a lot cleaner without having to use our custom touch/scroll methods. 

Theres some weird scrolling behavior where its not really snapping to the cause circles. I tried both embla `containScroll` options but it didnt make a difference. I think it may have to do with there not being enough causes to slide to? 

The design has all 9 cause categories displayed. Is this a change we're making? I just left it at the original 5 categories we had as it was not clear to me if we are adding categories. I know `refugees` was removed because it didnt have enough loans at the time.

emuvente:
> The design has all 9 cause categories displayed. Is this a change we're making? I just left it at the original 5 categories we had as it was not clear to me if we are adding categories. I know refugees was removed because it didnt have enough loans at the time.

Good catch! I'll as product + design about that

emuvente (approved):
👍

BoulderBrains:
👀 👍 
Thanks for taking this @eddieferrer

PR #2115, opened 2020-09-16
Stop passing route hashes as selectors unconditionally

remoun:
Given our use of Auth0 and this scrollBehavior, we end up getting errors like:
```
	SyntaxError: Document.querySelector: '#access_token=...'
	SyntaxError: Document.querySelector: '#_=_' is not a valid selector
	SyntaxError: Document.querySelector: '#/login' is not a valid selector
	SyntaxError: Failed to execute 'querySelector' on 'Document': '#/' is not a valid selector.
```
This should fix those errors.

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -4,12 +4,14 @@ import routes from './routes';
 
 Vue.use(Router);
 
+const DISALLOW_SELECTOR_REGEX = /[;=/]/;
+
 export default function createRouter() {
 	return new Router({
 		mode: 'history',
 		routes,
 		scrollBehavior(to, from, savedPosition) {
-			if (to.hash) {
+			if (to.hash && !DISALLOW_SELECTOR_REGEX.test(to.hash)) {
 				return { selector: to.hash };
 			}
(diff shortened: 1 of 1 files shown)


src/router/index.js
```diff
@@ -4,13 +4,18 @@ import routes from './routes';
...
 	return new Router({
 		mode: 'history',
 		routes,
 		scrollBehavior(to, from, savedPosition) {
-			if (to.hash) {
-				return { selector: to.hash };
+			if (to.hash && !DISALLOW_SELECTOR_REGEX.test(to.hash)) {
+				return {
+					selector: to.hash,
+					offset: savedPosition || { x: 0, y: 0 },
```
emuvente, line 17: I don't think we want this offset defined at all, as it's the offset relative to the position of the element represented by `selector` whereas `savedPosition` is the offset relative to the window

remoun: Ah great catch

emuvente (approved):
👍  thanks for figuring this out!

mcstover (approved):

PR #2119, opened 2020-09-17
Add anayltics events for get started flow

mcstover:


src/pages/GetStarted/GetStartedPlaces.vue
```diff
@@ -95,6 +96,11 @@
 						<button
 							class="get-started__select-all-btn"
 							@click.prevent="toggleAllCountries(!selectAll)"
+							v-kv-track-event="[
+								'Lending',
+								'click-country-all',
+								'select all'
```
eddieferrer, line 102: This will track on `select all` and `deselect all`, maybe change this last param on the event to: `selectAll ? 'select all' : 'deselect all`

eddieferrer (approved):

src/pages/GetStarted/GetStartedPlaces.vue
mcstover: That didn't worked tried several ways but couldn't get the switch due to timing issues.

mcstover: Would have used the toggleAllCountries but it's reused and I just didn't want to refactor much for a such a minor element in the interface.

PR #2122, opened 2020-09-18
KvFlag expansion / specificity fix

ryan-ludwig:
Width is set on the .wrapper one div deep, so we don't need it on the top level where it can cause problems.


emuvente (approved):
👍

PR #2124, opened 2020-09-18
GROW-245 timeline carousel first pass

ryan-ludwig:
Also a few tweaks to global 15 year styles per design and tweak KvProgressCircle default stroke width to match KvProgressBar


mcstover (approved):
Stickers!

PR #2130, opened 2020-09-21
Add more clean category names

emuvente:


mcstover (approved):
These would ensure proper the presentation IF they are presented based on the external setting right?

emuvente:
> These would ensure proper the presentation IF they are presented based on the external setting right?

Yes that's right

PR #2134, opened 2020-09-21
Fix string passed to Edit Causes instance of KvCauseSelector.

mcstover:
Fixes `vue.runtime.esm.js:1888 Error: Cannot find module './refugees-and-id-ps_2x.png'` wherein the LoanChannel name was being used that doesn't match the image names.

### DIFF
--- a/src/components/GetStarted/EditPreferences.vue
+++ b/src/components/GetStarted/EditPreferences.vue
@@ -181,12 +181,67 @@ export default {
 			default: 0,
 		},
 	},
+	data() {
+		return {
+			causeList: [
+				{
+					id: 5,
+					name: 'women',
+					checked: false,
+				},
+				{
+					id: 6,
+					name: 'shelter',
+					checked: false,
+				},
+				{
+					id: 4,
+					name: 'education',
+					checked: false,
+				},
+				{
+					id: 102,
+					name: 'technology',
+					checked: false,
+				},
+				{
+					id: 8,
+					name: 'agriculture',
+					checked: false,
+				},
+				{
+					id: 96,
+					name: 'COVID-19',
+					checked: false,
+				},
+				{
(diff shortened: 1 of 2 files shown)


mcstover:
Verified working in the VM

src/components/GetStarted/EditPreferences.vue
```diff
@@ -181,12 +181,67 @@ export default {
 			default: 0,
 		},
 	},
+	data() {
+		return {
+			causeList: [
```
eddieferrer, line 186: This is repeated in `src/pages/GetStarted/GetStartedCauses.vue` and could be moved to a data mixin like the one we have for `src/plugins/loan-channel-query-map.js`

mcstover: True, just not a priority at this time.

src/components/GetStarted/EditPreferences.vue
```diff
@@ -181,12 +181,67 @@ export default {
...
+					checked: false
+				}
+			]
+		};
+	},
 	computed: {
 		causesTrimmed() {
-			if (this.causes.length <= 3) {
-				return this.causes;
+			const causeIds = this.causes.map(cause => cause.id);
```
eddieferrer, line 237: Alternatively, we can just pass the id's into this component from `GetResults` since that is all this component needs and we could remove this map.

eddieferrer (approved):

emuvente (approved):

src/components/GetStarted/EditPreferences.vue
mcstover: We need the shortened names from the map, I just enabled the id's so i could easily map from the LoanChannel based values (names are different here) to the causeList values which dictate the image paths.

PR #2136, opened 2020-09-21
Break header styling into it's own component

ryan-ludwig:


ryan-ludwig:
Just trying to not step on your toes in sponsors

PR #2153, opened 2020-09-25
allow null for riskRating if that's what the api initially returns.

mcstover:


mcstover:
@emuvente @eddieferrer This allowed me to come into the settings page with an empty risk rating and proceed to activate and set my custom risk rating values.

eddieferrer (approved):

emuvente (approved):

PR #2161, opened 2020-09-28
RALLY-001 Initial layout, animation, and GK on 15YearHeader

rally-adam:


.storybook/preview.js
```diff
@@ -43,6 +56,12 @@ addParameters({
 				value: '#000000'
 			},
 		],
-	}
+	},
+	viewport: {
+    viewports: {
```
ryan-ludwig, line 61: Looks like a double-up here

ryan-ludwig: oh, my mistake! looks good.

src/components/15Years/15YearsStyles.vue
```diff
@@ -31,23 +31,32 @@
 		@include body-text();
 	}
```
ryan-ludwig, line 33: Since the CSS in this file isn't scoped these changes are effecting sections lower on the page. @rally-adam Can you please move these changes into 15YearsHeader.vue

src/assets/scss/components/15-years/15-years.scss
```diff
@@ -59,25 +70,29 @@ $font-family-monospace: 'Roboto Mono', monospace;
...
 	line-height: 1.35;
 
-	@include breakpoint('large') {
-		font-size: rem-calc(16);
+	// @include breakpoint('large') {
+	// 	font-size: rem-calc(16);
+	// }
+
+	@include breakpoint(small) {
+		font-size: calc(18px / 320 * 100vw);
```
ryan-ludwig, line 82: Browsers seem to be ignoring these calced values because of the vw unit.

src/pages/15Years/15Years.vue
```diff
@@ -50,6 +54,20 @@ export default {
 	metaInfo: {
 		title: '15 Years'
 	},
+	props: {
```
emuvente, line 57: Since this is a page component, I think these would make more sense as data fields and be set in this component instead of the story. Thoughts on that, anyone?

src/assets/scss/components/15-years/15-years.scss
emuvente: Yeah, stylelint is calling these out as invalid calc values because of the mixed px and vw units:
```
function-calc-no-invalid: Expected to be compatible with the left and right argument types of "*" operation.
```

emuvente: I think something like `calc(18 / 320 * 100vw)` (dropping the px) might work?

src/pages/15Years/15Years.vue
emuvente: I'll make that change

src/components/15Years/15YearsStyles.vue
emuvente: @ryan-ludwig Have you done this yet? I was going to fix some stylelint warnings from this file

src/assets/scss/components/15-years/15-years.scss
```diff
@@ -59,25 +71,30 @@ $font-family-monospace: 'Roboto Mono', monospace;
...
 	line-height: 1.35;
 
-	@include breakpoint('large') {
-		font-size: rem-calc(16);
+	// @include breakpoint('large') {
+	// 	font-size: rem-calc(16);
+	// }
+
+	@include breakpoint(small) {
+		font-size: calc(18 / 320 * 100vw);
```
ryan-ludwig, line 83: We can't do these vw multipliers for accessibility reasons. They prevent a user from increasing their font size. Will push up a change.

src/assets/scss/components/15-years/15-years.scss
rally-adam: Yeah, dropping px should work. Would it be more appropriate to move these rules into the Header component? Was just speaking w/ the other dev about this

src/components/15Years/15YearsStyles.vue
ryan-ludwig: @ryan-ludwig done

emuvente:
@rally-adam are all of those `bin` files being used? I only see a reference to `19-15.bin` in the code

rally-adam:
@emuvente only one will be used; the others are currently present because I haven't had a chance to run the options past the designers in order to choose one. We can either removed unused ones now, or remove the unused ones after reviewing with design.

emuvente:
@rally-adam waiting until after design reviews it works for me

emuvente (approved):
I think this is good to merge! We can work on this more from master

ryan-ludwig (approved):

mcstover (approved):
Thank you all for getting this first stage in order for merge.

PR #2162, opened 2020-09-28
GROW-256 Banner Transition

eddieferrer:
Added smoother transition for kiva 15 banner

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
@@ -1,6 +1,11 @@
 <template>
 	<div class="appeal-15-wrapper">
-		<transition name="kvfastfade">
+		<transition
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
```diff
@@ -196,21 +195,31 @@ a {
 
 .appeal-15-row {
 	&--open {
-		padding: rem-calc(16) 0;
+		& > div {
```
eddieferrer, line 198: I had to manually add these padding top and padding bottom here to get the transition to work. With padding on the `appeal-15-row` the animation to height: 0 would leave the padding visible. I'm not sure why this is the case, I even double checked the box-sizing and this seems strange to me. I gave up on that issue, to get around it, the row has 0 padding, and I added padding-top and bottom to the divs within it. I specificied *-bottom *-top so that the grid left/right padding on the div's wouldn't be affected

PR #2173, opened 2020-09-30
GROW-232 a11y changes, scroll to section functionality

ryan-ludwig:
- a11y: Removed vw units applied to typography since they prevent the user from changing font size
- a11y: Wrap the contents of the header links in the a tag so the contents will announce on screenreaders
- perf: Sized sticker images to the width they're used at
- Add smooth scroll from the header links to the sections below


ryan-ludwig:
@rally-adam Sorry for so much refactoring in the header card CSS. Once I removed the vw units things got a little crazy. Hopefully this matches your design intent.

mcstover (approved):
@ryan-ludwig I think this all makes sense.

@rally-adam Would you like to pull this down and verify before we merge?

emuvente (approved):
👍

PR #2175, opened 2020-09-30
links gathered and updated

BoulderBrains:
Country wide links gathered and updated. Went with links to the old stack page so we could include countries that don't currently have fundraising loans, but may in the future.

### DIFF
--- a/src/components/15Years/15YearsIndividualsData.js
+++ b/src/components/15Years/15YearsIndividualsData.js
@@ -15,7 +15,7 @@ export default [
 		videoId: 462763475,
 		cta1: 'Lend to a South American borrower',
 		cta2: 'Lend monthly to U.S. small businesses',
-		link1: '/lend?country=bo,br,co,ec,py,pe', // all countries in S. America
+		link1: '/lend/filter?countries=South%20America%20%3E%20Bolivia~South%20America%20%3E%20Brazil~South%20America%20%3E%20Colombia~South%20America%20%3E%20Ecuador~South%20America%20%3E%20Paraguay~South%20America%20%3E%20Peru&sortBy=popularity', // all countries in S. America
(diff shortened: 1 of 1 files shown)


mcstover:
We should really be linking to lend/filter for these. You can include countries that may not be available at this moment there as well. Just need to use the country name from our DB (Geocode Table).

mcstover:
Verified that “unknown” or "absent" countries don’t break the sort on lend/filter…
https://www.kiva.org/lend/filter?countries=Eastern%20Europe~Eastern%20Europe%20%3E%20Albania~Eastern%20Europe%20%3E%20Kosovo~Eastern%20Europe%20%3E%20Moldova~Eastern%20Europe%20%3E%20Matthew&sortBy=popularity
Check that last country 😉

BoulderBrains:
Thanks @mcstover, Getting these links fixed up this morning.

BoulderBrains:
One odd thing with these links is that if a user goes into the country section of the filter menu and clicks "deselect all", it only removes the countries that currently have fundraising loans. 

All Eastern Europe countries with fundraising loans, plus I manually added Albina
Screen Shot 2020-09-30 at 9 52 34 AM

I click inside the location region menu and "deselect all" and I'm left with the filter I manually added (because there were no fundraising loans currently in that country)
Screen Shot 2020-09-30 at 9 52 12 AM

On work around is the user can click the "Reset all" option when all the filters are applied, but I thought it would be worth bringing up this possible hole in the flow. 
Screen Shot 2020-09-30 at 9 52 02 AM

mcstover (approved):
Thanks for the updates and callouts on the quirks.

something to explore in another PR would be to make that deselect operate from the top level...

ryan-ludwig (approved):
Thank you!

PR #2181, opened 2020-09-30
Use numeral for number formatting and rearrange combined calc/rem-calc

emuvente:
Small change so that we don't have to use custom code for number formatting, and a fix for this warning:
```
WARNING: calc(100vw - 72px) was passed to rem-calc(), which is not a number.
```


mcstover (approved):
looks good though I wonder if that last rem-calc item has been altered..

ryan-ludwig (approved):
👍

PR #2187, opened 2020-09-30
GROW-256 Appeal 15 Banner Changes

eddieferrer:
* Removed animation on load, animated % fill, made % query only happen on client, switched from store2 to cookies to prevent unwanted flash of banner state

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
@@ -153,7 +153,7 @@ export default {
 	},
 	methods: {
 		shrinkAppeal() {
-			cookieStore.set('appeal_banner_15_shrunk', true);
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBanner15.vue
```diff
@@ -138,18 +119,41 @@ export default {
...
+						vm.fifteenYearGoalPercent = percentFull.val;
+					}
+				});
+			}
+		});
+	},
 	methods: {
 		shrinkAppeal() {
-			store2.session.set('appeal_banner_15_shrunk', true);
+			cookieStore.set('appeal_banner_15_shrunk', true);
```
emuvente, line 156: I would recommend adding a third argument here with `{ path: '/' }` to ensure that only one cookie is created for the whole site instead of a cookie for each page

mcstover (approved):
Thank you!

PR #2200, opened 2020-10-01
Fixed jank of partner text expansion

eddieferrer:
* Switch to css animations for the stats coming in, with a delay to allow the text to free up some space before they come in. Prevents the layout shift of the page.


mcstover (approved):
nice

ryan-ludwig (approved):
smooooth

PR #2202, opened 2020-10-01
Fixes header positioning on wide resolutions

eddieferrer:


ryan-ludwig:
I think this should get fixed by our usual .row .column containers.

ryan-ludwig:
Though I realize there's CSS in the header module that overrides our normal stuff. Probably will take a little wrangling.

rally-adam:
Yeah - I was noticing this morning that there was an issue at wide resolutions and started looking into it, but I see you're already on it! This seems like a pragmatic immediate fix, but if it's possible to achieve this properly within your .row .column system, maybe cleaner in the long-term.

mcstover (approved):
I'm ok with this approach if the foundation classes can't accommodate...

xga-offset-2 or something like that...

ryan-ludwig:
I'm good with running with this for now.

ryan-ludwig (approved):

eddieferrer:
> I think this should get fixed by our usual .row .column containers.

I tried that and it turned into a bit of a mess, since it is absolutely positioned. I think this fix has the most minimal impact and least amount of code

PR #2204, opened 2020-10-01
Kiva15 link fix

BoulderBrains:
if the link contains 'http' we render an <a> with an href
Otherwise we render the <router-link> component how it was previously.


emuvente (approved):
👍

mcstover (commented):
👍 ty

PR #2211, opened 2020-10-02
VUE-397 New Auto Deposit Routing

eddieferrer:
* Changed AD route from temp route to final


mcstover:
Let's until later in the sprint and make sure to merge before branch cut on Wed. In the meantime, we can do some additional verification at the `-new` url in dev

mcstover (approved):

PR #2224, opened 2020-10-07
VUE-264 experiments by route

ryan-ludwig:
This is an idea to simplify experiment code in Vue components so adding and removing them becomes easier.

main.js checks the route, if an experiment should be present on that route, prefetch it.
In the page you want to do something with the experiment, import a util and check the status ideally directly in the data object, then fire your tracking event in created or wherever it makes sense. 

To make cleaning up easier, I think some naming conventions could be used both in the admin and frontend repos to help search across the codebase.
Possibly something like `'EXP-GROW-254-Oct2020'` would be the experiment key in admin, and we camel-case it on the front: `'expGrow254Oct2020'`? Any other ideas on naming so it's clear where experiment code lies for easy cleanup later?

Any other thoughts or ideas on simplifying experiment code?


src/pages/15Years/15Years.vue
```diff
@@ -31,6 +31,7 @@
 <script>
 
 import { fifteenYearHeaderTheme, fifteenYearFooterTheme } from '@/util/siteThemes';
```
ryan-ludwig, line 33: This is an example of what reading the experiment in a component would look like.

src/pages/15Years/15Years.vue
```diff
@@ -52,12 +53,22 @@ export default {
 	metaInfo: {
 		title: '15 Years'
 	},
+	inject: ['apollo'],
 	data() {
 		return {
 			headerTheme: fifteenYearHeaderTheme,
 			footerTheme: fifteenYearFooterTheme,
+			uiExpGROW166Shown: getExperimentVersion('intercom_messenger', this.apollo) === 'shown',
```
ryan-ludwig: Admin, this key could be 'EXP-GROW-166-Aug2020' instead of 'intercom_messenger'

src/pages/15Years/15Years.vue
```diff
@@ -52,12 +53,22 @@ export default {
 	metaInfo: {
 		title: '15 Years'
 	},
+	inject: ['apollo'],
 	data() {
 		return {
 			headerTheme: fifteenYearHeaderTheme,
 			footerTheme: fifteenYearFooterTheme,
+			expGrow166Aug2020: getExperimentVersion('intercom_messenger', this.apollo) === 'shown',
```
ryan-ludwig, line 61: In admin, the experiment key could be 'EXP-GROW-166-Aug2020' instead of 'intercom_messenger'

ryan-ludwig:
Implementing this site-wide would be a few days task, but also be an opportunity to clean up a lot of dead experiment code. Experiments that are still running would be moved into main.js and loaded via the util.

mcstover:
This is great, Let's do some local/vm verification this week and work on introducing into dev at the end of the week or early next week.

BoulderBrains:
👀

src/util/experimentPreFetch.js
```diff
@@ -147,8 +195,19 @@ export function fetchAllExpSettings(config, apolloClient, route) {
 			}
 			return true;
 		})
-		// prefetch all active experiment settings and assignments
+		// prefetch all active experiment settings and assignments if the current route matches their glob
 		.then(() => {
-			return Promise.all(_map(activeExperiments, settingId => fetchExperimentSettings(settingId, apolloClient)));
+			const currentRouteExperiments = activeExperiments.filter(experiment => {
+				let expInRoute = false;
+				experiment.routes.forEach(expRoute => {
```
eddieferrer, line 202: You can cleanup some of this code by using array.some()

eddieferrer:
This looks good to me, it's nice to remove all those readFragment from the components.

emuvente:
This looks great! While we're here I would recommend we also push to use the version names in the experiment setting that we want to use to track the experiment instead of 'control' and 'shown' like we've done for many experiments.

That way we can change things like this:
```
this.$kvTrackEvent(
	'homepage',
	'EXP-GROW-166-Aug2020',
	this.expGrow166Aug2020 === 'shown' ? 'b' : 'a',
);
```
to just this:
```
this.$kvTrackEvent(
	'homepage',
	'EXP-GROW-166-Aug2020',
	this.expGrow166Aug2020,
);
```

mcstover:
> This looks great! While we're here I would recommend we also push to use the version names in the experiment setting that we want to use to track the experiment instead of 'control' and 'shown' like we've done for many experiments.
> 
> That way we can change things like this:
> 
> ```
> this.$kvTrackEvent(
> 	'homepage',
> 	'EXP-GROW-166-Aug2020',
> 	this.expGrow166Aug2020 === 'shown' ? 'b' : 'a',
> );
> ```
> 
> to just this:
> 
> ```
> this.$kvTrackEvent(
> 	'homepage',
> 	'EXP-GROW-166-Aug2020',
> 	this.expGrow166Aug2020,
> );
> ```

This would be a good process update on our part but we must make sure that "version names" in the settings are always set to "a", "b", and possibly "c" as product requested.

PR #2228, opened 2020-10-07
GROW-290 globe flag markers

ryan-ludwig:
Show flag instead of pin marker. Enlarge flag dots and make them red.

### DIFF
--- a/src/components/15Years/15YearsGlobe.vue
+++ b/src/components/15Years/15YearsGlobe.vue
@@ -268,6 +268,7 @@ export default {
 
 <style lang="scss">
 @import 'settings';
+@import 'components/15-years/15-years';
 
 .callout {
 	display: inline-block;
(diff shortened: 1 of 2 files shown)


src/components/15Years/15YearsGlobe.vue
```diff
@@ -277,16 +278,12 @@ export default {
...
+	border: 2px solid white;
 	border-radius: 50%;
 	box-shadow: 4px 4px 8px rgba(0, 0, 0, 0.25);
-
-	@include breakpoint(large) {
-		width: 7px;
-		height: 7px;
-	}
+	width: 16px;
+	height: 16px;
```
emuvente, line 286: Can this be rems instead or does it have to be px for globekit?

ryan-ludwig: It can be rems. I used px because everything else in this file was px. Will do a follow-up PR with px2rem conversions for this component and 15YearsHeader.vue

mcstover (approved):
pending minor tweaks

emuvente (approved):
👍

PR #2234, opened 2020-10-09
GROW-262 Delay showing basket verification lightbox until credit is applied and verification is required

emuvente:
This depends on kiva/kiva#9342, and so this shouldn't be merged until that is in development. This should also probably not be merged until after any expected ui hotfixes have been made.

### DIFF
--- a/src/components/Checkout/BasketVerification.vue
+++ b/src/components/Checkout/BasketVerification.vue
@@ -11,27 +11,31 @@
 		</div>
 
 		<kv-lightbox
-			:visible="verificationState === 'required'"
+			:visible="visible && verificationState === 'required'"
 			class="verification-required"
 			title="Verify your email"
+			@lightbox-closed="close"
 		>
 			<p>To ensure your safety, we added an extra layer of security.</p>
 			<p>Once we verify your account, you can continue checking out!</p>
-			<kv-button @click.native="send">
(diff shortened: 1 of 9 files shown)


src/plugins/checkout-utils-mixin.js
```diff
@@ -94,6 +95,12 @@ export default {
 				// Log validation errors
 				Sentry.captureException(`${error}:${value}`);
 
+				// Show the verificatoin lightbox if basket is not verified, and don't show a tip message
```
remoun, line 98: nit: typo in verification

src/components/Checkout/BasketVerification.vue
```diff
@@ -42,46 +46,69 @@
...
 		},
 		send() {
-			this.setState('pending');
-			this.apollo.mutate({ mutation: startBasketVerificationMutation });
+			this.sending = true;
+			this.apollo.mutate({
+				mutation: startBasketVerificationMutation,
+				awaitRefetchQueries: true,
+				refetchQueries: [
+					{ query: getVerificationState },
```
remoun, line 91: Nice!

remoun (approved):

src/plugins/checkout-utils-mixin.js
emuvente: 👍

PR #2244, opened 2020-10-12
Fold in basic flow for login/reg and minimal checkout.

mcstover:


mcstover:
@emuvente @eddieferrer Please review at your leisure. These are all isolated changes in prep for some demo recordings were doing today. The concept of InContextCheckout that covers a minimal basket and its required validations might be of interest.

src/pages/LandingPages/CorporateCampaign/InContextCheckout.vue
```diff
@@ -0,0 +1,129 @@
...
+
+		};
+	},
+	computed: {
+		registerOrLoginHref() {
+			if (typeof window !== 'undefined') {
+				return `/ui-login?force=true&doneUrl=${window.location.href}`;
+			}
+			// TODO: map using route for ssr
+			return '/ui-login?force=true';
```
emuvente, line 93: For both client and server you can use
```
return `/ui-login?force=true&doneUrl=${this.$route.path}`
```

mcstover: Cool thanks. I need to make sure all query params will flow through as well and will work that out with the router data.

emuvente: Good point, you might need to do `encodeURIComponent(this.$route.fullPath)`

emuvente (commented):
👍

PR #2258, opened 2020-10-14
Moved custom join array function into utils, added unit test

eddieferrer:
We do this type of thing in several places: get-started causes list, get-started country lists, autolending filters list come to mind.

I did not refactor the old use cases, but going forward we can use this util function


emuvente (approved):
Nice!

mcstover (approved):

BoulderBrains (approved):

PR #2259, opened 2020-10-14
grow-274 page with header and sections

BoulderBrains:
MFA page frame with sections. 

Ticket mentions to show/hide the MFA section based on a feature flag that's yet to be built out. (https://kiva.atlassian.net/browse/GROW-276), so that small portion was left out.

Screen Shot 2020-10-14 at 3 25 30 PM


src/pages/Settings/MFASettings.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+	<div>
```
emuvente, line 2: This div can be dropped so that `<www-page>` is the root element

emuvente (approved):
one nitpick but overall looks great! 👍

eddieferrer (approved):
Looks good. 

Up for discussion, but I wonder since these designs are so old if we should take the liberty of making these new setting pages be more stylistically in line with the new settings pages: settings/autolending, or settings/subscriptions.

 I'm thinking specifically of some of the common elements like the styling on the page header and the idea of settings cards used on those pages to divide the sections 🤔

BoulderBrains:
@eddieferrer I think that's a good question to raise. 
Looking at the designs, they appear to be last updated "Sept.14.2020", which still feel pretty current. Maybe we can raise the question with the right person from design tomorrow and ask if the style on /settings/autolending would be prefered? If that's the case, we could spin up another ticket to make that update and capture the work (though it's a small amount) to update from the styles from the mocks we started with. @emuvente What are your thoughts? and if we reach out to design, who would be the right person? Olly? Ruby? (looks like the last edits to the figma doc I'm going off were made by Ruby.) 

Anyways, here's what the /settings/autolending page looks like:
Screen Shot 2020-10-14 at 5 54 46 PM

I'm going to merge this branch into master and leave the branch open until we close the loop on this discussion tomorrow.

PR #2260, opened 2020-10-14
VUE-431 Add ID field to every query that can have one

emuvente:
This adds a linting rule to require that the `id` and `key` fields are always requested in graphql queries when they are available on the type. This also adds the id field to all the queries that did not have one. And finally, this removes the custom Shop and ShopMutation id definitions. This makes it easier for Apollo to cache everything that it needs to.

The hope is that it resolves the various Invariant Violations that we've seen, especially around using the Shop basket. This solution was inspired by the fix in #2141. 

We could also remove the extra `basketId` variable definitions, as those should be set by `BasketLink`, but I think it makes sense to test that this isn't causing problems first.


eddieferrer (approved):
This is a nice improvement if it clears up those invariant errors

mcstover (approved):
💯

PR #2262, opened 2020-10-16
SUBS-320 MG Thanks Page Upsell

eddieferrer:
* Implements MG Thanks page upsell based on experiment

See SUBS-320 for specs. 

Still TODO:
* Amount drop down should have "other" option that enables user to input custom amount. 
* For should populate to start on today minus 1, so that MG won't kick in until approximately 30 days from now

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -20,7 +20,13 @@
 						</p>
 					</div>
 				</template>
+			</div>
+		</div>
 
+		<monthly-good-c-t-a v-if="showMonthlyGoodCTA" />
+
+		<div class="row page-content">
+			<div class="small-12 columns thanks">
 				<checkout-receipt
 					v-if="receipt"
 					class="thanks__receipt"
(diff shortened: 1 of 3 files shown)


eddieferrer:
screencapture-dev-vm-01-kiva-org-checkout-thanks-2020-10-15-10_02_22

src/pages/Thanks/ThanksPage.vue
```diff
@@ -105,8 +117,19 @@ export default {
...
+				variables: {
+					checkoutId: numeral(route.query.kiva_transaction_id).value()
+				}
+			}).then(() => {
+				return client.query({
+					query: experimentAssignmentQuery, variables: { id: 'mg_thanks_cta' }
+				});
+			});
+		},
 		preFetchVariables({ route }) {
```
eddieferrer, line 133: I don't understand why `preFetchVariables` and `variables` is needed in this case, since I'm specifying the variable in the query above...

emuvente: Because `query` is defined in the apollo options, our apollo plugin is calling `preFetchVariables` to use for calling `apollo.readQuery`, and on the client it's calling `variables` to use for calling `apollo.watchQuery` (https://github.com/kiva/ui/blob/master/src/plugins/apollo-plugin.js#L22). If you don't define `query`, those functions won't be needed, but then you also need to setup your own calls to `apollo.readQuery` and `apollo.watchQuery`.

emuvente (approved):
👍

mcstover (approved):

PR #2263, opened 2020-10-17
grow-296 two step verification page frame reworked to more current style

BoulderBrains:
Moved the Two Step verify pages over to the style of /settings/autolend

Now it looks like this: 
Screen Shot 2020-10-17 at 9 38 45 AM


emuvente:
This looks great! I just have a few suggestions about the routing:

I think the TwoStepVerifyMethod.vue component should be included in TwoStepVerifyPage.vue directly instead of as a router-view. In addition to that, I think the `/settings/security` route should point to MFASettings.vue, and a `/settings/security/mfa` route should point to TwoStepVerifyPage.vue.

Side note: sorry for not noticing this in the previous PR, but MFASettings.vue should probably be renamed to something like SecuritySettings.vue, since it's not meant to be just MFA.

eddieferrer (approved):

emuvente (approved):
👍

PR #2267, opened 2020-10-19
Grow 274

BoulderBrains:


BoulderBrains:
Changed route and reworked components after the discussion with design last week. Page frame now looks like this: 
Screen Shot 2020-10-19 at 3 08 05 PM

src/router/routes.js
```diff
@@ -184,6 +184,19 @@ export default [
 		],
 
 	},
+	{
+		path: '/settings/security',
+		component: () => import('@/pages/Settings/SecuritySettings'),
+		children: [{
+			path: '',
+			name: 'password',
```
eddieferrer, line 192: Are we sure about the name on this route? We don't typically have names on routes...

src/router/routes.js
```diff
@@ -184,6 +184,19 @@ export default [
...
+			component: () => import('@/pages/Settings/SecuritySettingsPage'),
+			meta: {
+				activeLoginRequired: true,
+			}
+		},
+		]
+	},
 	{ path: '/settings/security/mfa', component: () => import('@/pages/Settings/TwoStepVerifyPage') },
```
eddieferrer, line 200: You can make this a child of the settings/security route above

BoulderBrains: The TwoStepVerifyPage has it's own header and footer and putting this as a child of settings/security results in: 
Screen Shot 2020-10-19 at 3 50 10 PM

src/router/routes.js
BoulderBrains: removed, thanks for calling it out

emuvente:
I think SecuritySettings.vue should be combined with SecuritySettingsPage.vue into a single component that doesn't use a router-view component. I think that will allow the 2-step verification page to be used as that child route without causing problems 🤔

src/router/routes.js
eddieferrer: Yeah you're right, I'm fine with leaving it as is then

src/pages/Settings/SecuritySettings.vue
```diff
@@ -0,0 +1,35 @@
...
+import WwwPage from '@/components/WwwFrame/WwwPage';
+
+export default {
+	components: {
+		WwwPage,
+	},
+};
+</script>
+
+<style lang="scss">
```
ryan-ludwig, line 24: Please make sure the styles stay "scoped"

ryan-ludwig: On the other pages as well

BoulderBrains:
@ryan-ludwig I've scoped the css of both files
@emuvente
- I've combine SecuritySettings & SecuritySettingsPage into one component that doesn't use the <router-view>
- In order for both routes to load with the correct components, I had to use two different routes, each with a paths and component.

BoulderBrains:
I would like to get this merged before branch cut today, any other thoughts suggestions? @emuvente @eddieferrer @ryan-ludwig

src/pages/Settings/Password.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<div class="row">
+		<div class="column small-12 large-8">
```
ryan-ludwig, line 3: I believe this class should be "columns"

src/pages/Settings/TwoStepVerification.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+	<div class="row">
+		<div class="column small-12 large-8">
```
ryan-ludwig, line 3: same "columns"

ryan-ludwig (approved):
Minor class name thing, but otherwise looks good!

eddieferrer (approved):

src/pages/Settings/Password.vue
eddieferrer: You can use either column or columns, since there is only 1, in this case 'column' makes grammatical sense

PR #2270, opened 2020-10-20
Fixed link on borrower pre-application/how-long page

SqueakAttack:
Removed s from check out this useful guide link at /borrow/pre-applic…ation/how-long.
Tested locally and this appears to work as expected.


mcstover (approved):
-s

PR #2271, opened 2020-10-20
VUE-432 Move the query into created

ryan-ludwig:


mcstover (approved):
This should work fine. I just pushed something similar to the previous branch that i was able to verify the data flow for in my vm.

emuvente (approved):
I'm surprised this would work, but I guess I just learned something new about apollo!

ryan-ludwig:
🤞

mcstover:
Verified in dev via VPN. Cherry-picked to release.

PR #2274, opened 2020-10-20
SUBS-320 Thanks Page MG CTA Start Date

eddieferrer:
* Implemented start date minus 1 for MG setup form for users coming from thanks page CTA

This was part of the spec for SUBS-320. The idea being that the user will start MG about next month.

### DIFF
--- a/src/pages/MonthlyGood/MonthlyGoodSetupPageVariant.vue
+++ b/src/pages/MonthlyGood/MonthlyGoodSetupPageVariant.vue
@@ -203,6 +203,12 @@
 </template>
 
 <script>
+/**
+ * This template is not currently in use.
+ * It only exists for reference.
+ * There have been modifications to  MonthlyGoodSetupPageControl
+ * that have not been made here
+ */
 import numeral from 'numeral';
 import _get from 'lodash/get';
 import gql from 'graphql-tag';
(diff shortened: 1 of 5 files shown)


src/pages/MonthlyGood/MonthlyGoodSetupPageVariant.vue
```diff
@@ -203,6 +203,12 @@
 </template>
 
 <script>
+/**
```
eddieferrer, line 206: Should we delete this template and the wrapper template that decides which setup form?

mcstover: hmmm. I know the payment section of the variant is now out of date so we def want to prevent this from being turned on at all. That said, I don't know if we want to delete it outright just yet as it may have some useful code for the new designs...but we can always look back at a release branch to anything useful... so yea I'm ok removing it and cleaning up the exp codes.

mcstover (approved):

src/pages/MonthlyGood/MonthlyGoodSetupPageVariant.vue
eddieferrer: I'll remove it in a followup PR. In the SUBS backlog, there are items for a MG Landing page redesign, as well as exploring making monthly good signup a single page vs landing > setup.

emuvente (commented):
👍

PR #2287, opened 2020-10-21
Storybook stories for the two new /settings page urls

BoulderBrains:
Here are two new storybook stories to capture the
- /settings/security
- /settings/security/mfa
pages.


mcstover:
@BoulderBrains I think at the least we should try to create a section for `/settings/security` that has both of these pages as children components if possible.

emuvente (approved):
This also works as-is if that section suggestion doesn't pan out

BoulderBrains:
@mcstover @emuvente Something like this? 
Screen Shot 2020-10-22 at 9 38 44 AM

mcstover (approved):
Thank you!

ryan-ludwig (approved):
👍

emuvente (approved):
👍

PR #2291, opened 2020-10-23
Comment out dynamic swashie, Add party swashie

ryan-ludwig:


ryan-ludwig:
Do we want to try to get this into Monday's release?

mcstover (approved):
totally fine for a quick fix. We def don't want to delete the rest of this as we'll be basing a new donation banner on it in the coming sprints!

ryan-ludwig:
Cherry picked these commits into release-20201026 as well

PR #2293, opened 2020-10-26
phrasing changes

ryan-ludwig:
Replace white/blacklist with allow/deny

### DIFF
--- a/server/util/errorLogger.js
+++ b/server/util/errorLogger.js
@@ -10,7 +10,7 @@ module.exports = {
 			})
 		],
 		meta: false,
-		allowedRequestList: ['url', 'responseTime', 'level', 'message', 'headers.host', 'headers.method'],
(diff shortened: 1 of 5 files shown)


eddieferrer:
@ryan-ludwig there are several additional uses, if you want to bundle them all up in one PR...

server/util/errorLogger.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
 			})
 		],
 		meta: false,
-		requestWhitelist: ['url', 'responseTime', 'level', 'message', 'headers.host', 'headers.method'],
```
eddieferrer, line 13: This is part of the winston API. They are working on changing these terms for the upcoming 5.0 release, but we cannot make this modification just yet....

BoulderBrains (approved):
👍

emuvente (approved):
👍

server/util/errorLogger.js
mcstover: Thanks @eddieferrer for catching this!

PR #2306, opened 2020-10-26
Settings pages: add secondary nav and tertiary nav as a dropdown

ryan-ludwig:
VUE-437

### DIFF
--- a/src/components/WwwFrame/Menus/TheSettingsTertiaryMenu.vue
+++ b/src/components/WwwFrame/Menus/TheSettingsTertiaryMenu.vue
@@ -1,0 +1,54 @@
+<template>
+	<tertiary-menu>
+		<ul>
+			<li>
+				<router-link
+					to="/settings/account"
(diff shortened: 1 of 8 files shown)


src/components/WwwFrame/Menus/TheSettingsTertiaryMenu.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+	<tertiary-menu>
+		<ul>
+			<li>
+				<router-link
```
ryan-ludwig, line 5: Not sure if these are exactly the links we want

mcstover: With the exception of Security settings, which may need a feature flag, the other links show in the legacy settings sub nav too.

We could add Saved searches to fully match the two apps, though this will require a bit more setup.

mcstover (approved):

BoulderBrains (approved):

PR #2308, opened 2020-10-27
Settings area semantic tweaks

ryan-ludwig:
Sets page titles, use h1s.

This changes the appearance of the headings on these pages. If we want to keep the old appearance, we can do something like `<h1 class="h2">` but we should keep the `<h1>` for semantics and screenreaders. 

My personal opinion is to use a raw h1 here, and really on nearly all pages for consistency. If design wants to change the typography of what a page header look like, we change it in our global styles and all pages with a raw h1 will get the update, rather than more and more one-off designs that we'll have to go through when design updates to a new look.

Raw h1 on the left, h2 on the right:
image

### DIFF
--- a/src/pages/Autolending/AutolendingPage.vue
+++ b/src/pages/Autolending/AutolendingPage.vue
@@ -2,7 +2,7 @@
 	<www-page class="autolending" :gray-background="true">
 		<div class="title-area">
 			<div class="row column">
-				<h2>Auto-lending settings</h2>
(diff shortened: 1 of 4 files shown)


mcstover (approved):
Thanks, we should always maintain the semantic hierarchy of the page using html. To prevent the visual change in this case you could add a utility/modifier class that sets the prescribed style on the headers. `h2-header`, `settings-header`, etc...

eddieferrer (approved):

PR #2311, opened 2020-10-28
Bump isomorphic-fetch to v3.0.0

eddieferrer:
* Bump dependency
* Fix deprecated headers.getAll()

### DIFF
--- a/server/util/getSessionCookies.js
+++ b/server/util/getSessionCookies.js
@@ -23,15 +23,16 @@ module.exports = function getSessionCookies(url = '', requestCookies = {}) {
 					Cookie: getCookieString(requestCookies),
 				},
 			}).then(res => {
-				const setCookies = res.headers.getAll('set-cookie');
-				const parsed = setCookieParser(setCookies, { decodeValues: false });
+				const combinedCookieHeader = res.headers.get('set-cookie');
+				const splitCookieHeaders = setCookieParser.splitCookiesString(combinedCookieHeader);
+				const parsed = setCookieParser.parse(splitCookieHeaders, { decodeValues: false });
 				const cookies = parsed.reduce((cookieObject, cookie) => {
 					if (cookie.value !== 'deleted') {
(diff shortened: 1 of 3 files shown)


server/util/getSessionCookies.js
```diff
@@ -23,15 +23,16 @@ module.exports = function getSessionCookies(url = '', requestCookies = {}) {
 					Cookie: getCookieString(requestCookies),
 				},
 			}).then(res => {
-				const setCookies = res.headers.getAll('set-cookie');
-				const parsed = setCookieParser(setCookies, { decodeValues: false });
+				const combinedCookieHeader = res.headers.get('set-cookie');
```
eddieferrer, line 26: This returns a comma separated list, where getAll() returned an array

server/util/getSessionCookies.js
```diff
@@ -23,15 +23,16 @@ module.exports = function getSessionCookies(url = '', requestCookies = {}) {
...
+				const splitCookieHeaders = setCookieParser.splitCookiesString(combinedCookieHeader);
```
eddieferrer, line 27: Turns comma list into array

eddieferrer:
Used this documentation example: https://www.npmjs.com/package/set-cookie-parser#usage-in-react-native
Seems to be working for me, definitely not crashing with an error anymore. That being said, I couldn't produce a test case locally where `set-cookie` was actually part of the response header

server/util/getSessionCookies.js
eddieferrer, line 26: I couldn't reproduce a case locally where these headers are present

emuvente (approved):
This looks correct to me. I'll try it out locally to see if I can get the expected headers to show up.

emuvente:
Testing locally worked as expected. Also it looks like the failing travis test is just a transient error. You could push to this branch again to restart the build to confirm, but I think this is good to merge

mcstover (approved):

PR #2314, opened 2020-10-29
Attempt to fix flaky Travis test

eddieferrer:
AutolendingPage spec fails periodically on Travis. This might be able to fix it, as cypress will retry the assertion before click.

Reference:
https://docs.cypress.io/guides/core-concepts/retry-ability.html#Commands-vs-assertions

`If the assertion that follows the cy.get() command fails, then the cy.get() command will requery the application’s DOM again. Then Cypress will try the assertion against the elements yielded from cy.get(). If the assertion still fails, cy.get() will try requerying the DOM again, and so on until the cy.get() command timeout is reached.`


emuvente (approved):
👍

PR #2315, opened 2020-10-29
VUE-314 settings index page

ryan-ludwig:


src/router/routes.js
```diff
@@ -162,6 +162,13 @@ export default [
 	{ path: '/process-browser-auth', component: () => import('@/pages/ProcessBrowserAuth') },
 	{ path: '/protocol', component: () => import('@/pages/Protocol') },
 	{ path: '/register/social', component: () => import('@/pages/LoginAndRegister/RegisterSocial') },
+	{
+		path: '/settings',
+		component: () => import('@/pages/Settings/SettingsPage'),
+		meta: {
+			activeLoginRequired: true,
```
emuvente, line 169: The top-level settings page only requires a passive login, so this should be `authenticationRequired`

src/pages/Settings/SettingsPage.vue
```diff
@@ -0,0 +1,224 @@
...
+			isMfaActive: false,
+			isSubscriber: false,
+		};
+	},
+	apollo: {
+		query: pageQuery,
+		prefetch: true,
+		result({ data }) {
+			this.isSubscriber = data?.my?.autoDeposit?.isSubscriber ?? false;
+			this.isMfaActive = data?.general?.featureSetting?.value ?? false;
```
emuvente, line 188: I think you'll need to use `settingsUtils.readBoolSetting` for this

src/pages/Settings/SettingsPage.vue
```diff
@@ -0,0 +1,224 @@
...
+						<li>
+							<h3>Repayment settings</h3>
+							<p>Where you want your money to go when it's repaid</p>
+						</li>
+						<li>
+							<h3>Inactivity settings</h3>
+							<p>Where you want your money to go if your account becomes inactive</p>
+						</li>
+						<li
+							v-if="isSubscriber"
```
eddieferrer, line 71: I don't think we need this conditional here for auto lend settings

mcstover: Well if we follow our other paradigms for handling AL when someone is subscribing to MG, this check should be the inverse so that the section doesn't show...

That said, leaving it off completely would allow someone to try and click into it which would then be subject to the redirect if they have and active MG.

emuvente: To match what we had in the monolith for this page, this should be `v-if="!isSubscriber"` so that it doesn't show when the user is signed up for monthly good.

src/pages/Settings/SettingsPage.vue
ryan-ludwig: @emuvente Made a tweak to check against 'true' as a string, does that look like it'll work to you? Seems ok in my testing.

eddieferrer (approved):

src/pages/Settings/SettingsPage.vue
emuvente: 👍

emuvente (approved):
👍

PR #2316, opened 2020-10-29
grow-295 mostly in place, waiting on user mfa gql endpoint, foundatio…

BoulderBrains:
FAQ for grow-295 is contained within this work

Mobile: 
Screen Shot 2020-11-03 at 11 12 43 AM

Desktop:
Screen Shot 2020-11-03 at 11 13 03 AM

TODO: 
- MFA flag for individual user's doesn't yet exist, once it does, fold it into TwoStepVerification.vue
- ~~For some reason the Foundation Source Ordering isn't working for me in SecuritySettings.vue 🤔 
https://get.foundation/sites/docs/grid.html#source-ordering~~


mcstover (approved):

src/pages/Settings/TwoStepVerification.vue
```diff
@@ -1,38 +1,73 @@
...
+			return this.isMFAActive ? '<span style="color:#4faf4e;font-weight:500;">On</span>' : '<span style="color:#d74937;font-weight:500;">Off</span>';
+		}
+	},
+	inject: ['apollo'],
+	// apollo: {
+	// 	query: pageQuery,
+	// 	preFetch: true,
+	// 	result({ data }) {
+	// 		console.log('data', data);
+	// 		// this.isMFAActive = data.whatever
```
ryan-ludwig, line 69: There's some code for this feature flag ready to snag from the settings page: https://github.com/kiva/ui/blob/master/src/pages/Settings/SettingsPage.vue#L188

BoulderBrains: The end point I need to use here is created in grow-273, once that ticket is wrapped up, I'll get this hooked up correctly. The setting you pointed to is the mfa setting in general, I need an individual user's mfa enabled/disabled state.

src/pages/Settings/TwoStepFaq.vue
```diff
@@ -0,0 +1,96 @@
...
+					Level up your security with 2-step verfication
+				</div>
+				<p>
+					2-step verification protects your Kiva account by requiring an
+					additional code when you log in on a device we don't recognize.
+				</p>
+				<div class="strong">
+					FAQ
+				</div>
+				<div>
```
ryan-ludwig, line 15: I'd recommend setting these as an unordered list (`<ul>`), since it's a list of links

src/pages/Settings/TwoStepFaq.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+	<div>
+		<kv-settings-card>
+			<template v-slot:content>
+				<div class="strong">
```
ryan-ludwig, line 5: Semantically this header is probably an <h2>

src/pages/Settings/TwoStepFaq.vue
```diff
@@ -0,0 +1,96 @@
...
+					Level up your security with 2-step verfication
+				</div>
+				<p>
+					2-step verification protects your Kiva account by requiring an
+					additional code when you log in on a device we don't recognize.
+				</p>
+				<div class="strong">
```
ryan-ludwig, line 12: Semantically this header is probably an `<h3>`

src/pages/Settings/TwoStepFaq.vue
```diff
@@ -0,0 +1,96 @@
...
+					FAQ
+				</div>
+				<div>
+					<a @click="triggerWhatIsLightbox">What is 2-step Verification?</a>
```
ryan-ludwig, line 16: Since these do not change URLs, they should be set as button elements, rather than anchors.
You can match the styling using `<kv-button class="text-link" @click="...`

eddieferrer, line 16: These would be more semantic as buttons with the modifier class `text-link`

src/pages/Settings/TwoStepVerification.vue
```diff
@@ -1,38 +1,73 @@
...
-		</div>
-	</div>
+			<kv-icon
+				class="icon"
+				title="Auto-lending On"
+				name="auto-icon-on"
+			/>
+		</template>
+		<template v-slot:content>
+			<div>Status: <span v-html="MFAStatus"></span></div>
```
eddieferrer, line 13: This works, but I would prefer to not use v-html. A span with a conditional class and the styles in css I think would be better, the content can be a computed

ryan-ludwig, line 13: Semantically this header is an `<h3>`

eddieferrer:
As a general comment I would prefer we try to move towards a nuxt like folder structure, I think this will make things easier if we ever do decide to move to Nuxt. This means that the components in `/pages/` are referenced in the router, and components like the `TwoStepFaq` should not be in pages, it should be in the components directory.

src/pages/Settings/TwoStepFaq.vue
BoulderBrains: I've updated these to h2, worth noting that it looks different from the mock because of this update.

ryan-ludwig: No problem there, we can always change the appearance with CSS to match designs if needed. It's important to use the right tags so that machines without eyes (like search engines and screen readers) can understand what the page _means_. Visuals gets layered on after that.

BoulderBrains:
@eddie
- I've removed the v-html and replaced it with a conditional class.
- I've moved the 3 components out of the pages directory,  created a Settings folder in components and moved them into it accordingly.

eddieferrer (approved):

src/components/Settings/TwoStepFaq.vue
```diff
@@ -0,0 +1,122 @@
...
+			this.howItProtectsLBVisible = false;
+		},
+	}
+};
+</script>
+
+<style lang="scss" scoped>
+@import 'settings';
+
+ul {
```
ryan-ludwig, line 127: Minor thing, but when using scoped styles, there's a small performance hit if you use non-class selectors. 
From https://vue-loader.vuejs.org/guide/scoped-css.html 

> Scoped styles do not eliminate the need for classes. Due to the way browsers render various CSS selectors, p { color: red } will be many times slower when scoped (i.e. when combined with an attribute selector). If you use classes or ids instead, such as in .example { color: red }, then you virtually eliminate that performance hit.

So ideally this would be `.faq-list` or something instead of `ul`

ryan-ludwig: Same with the h2, h3

ryan-ludwig (approved):

mcstover (approved):
🕺

PR #2322, opened 2020-11-02
Update borrower application link to new link

coderintherye:
Updates to the new borrower application link which has moved out of Salesforce's portal and into FormAssembly forms which are embedded on kiva.org
In the main Kiva code we have this as a settings manager setting, not sure if that is possible here so just using the actual relative URL
This would ideally go out with next Kiva production release (which is when the code changes are going out in the PHP code for the new borrower application)

### DIFF
--- a/src/pages/Borrow/BorrowHowLong.vue
+++ b/src/pages/Borrow/BorrowHowLong.vue
@@ -27,7 +27,7 @@
 		</div>
 		<!-- eslint-disable max-len -->
 		<kv-button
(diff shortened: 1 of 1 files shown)


coderintherye:
@mcstover Are UI releases done in coordination with Kiva releases, or is there a process for doing a release more like how we did with styleguide? Asking only in the sense of wanting to tie this change to the Kiva prod release next week

mcstover:
> @mcstover Are UI releases done in coordination with Kiva releases, or is there a process for doing a release more like how we did with styleguide? Asking only in the sense of wanting to tie this change to the Kiva prod release next week

We're currently in sync with Kiva/main releases so they'll go out together for the next release

mcstover (approved):
This is fine for now and if it starts to change a lot we can get the setting value via the api and make it always dynamic.

coderintherye:
thanks @mcstover !

PR #2326, opened 2020-11-04
Kv Analytics Refactor

eddieferrer:
* Removed unncessary 5th param for kvTrackEvent
* Removed support for string type with | for kvTrackEvent
* Refactors trackEvent, add support for snowplow value

### DIFF
--- a/src/components/WwwFrame/Menus/CustomSecondaryMenu.vue
+++ b/src/components/WwwFrame/Menus/CustomSecondaryMenu.vue
@@ -2,15 +2,14 @@
 	<div>
 		<secondary-menu class="show-for-large">
 			<ul class="row">
-				<li v-for="{name, routerLink, url, eventTracking} in menuItems" :key="name">
+				<li v-for="{name, routerLink, url} in menuItems" :key="name">
 					<router-link
 						:to="routerLink"
-						v-kv-track-event="eventTracking"
 						v-if="routerLink"
 					>
 						{{ name }}
 					</router-link>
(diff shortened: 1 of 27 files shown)


eddieferrer:
Adds support for snowplow values. 

- `value` in trackEvent param (the 4th param) was being sent to GA as value, but was actually property in snowplow. 
- Added support for the actual value property to be sent to snowplow. 
- In order to keep legacy events reporting the same events, where the 4th param is GA Value, but Snowplow Property, I duplicated the 4th param in all those events. Going forward we can either keep the duplication for events that have a 4th param (most only have 3 params anyways) or we can use snowplow value more and pass the 4th param as null. 
- Double checked a lot of these events in looker, because of this mix up, many events have se_property set as NaN
- This should not break any legacy events because there will be no change in looker, except for the addition of value in cases where it was missing before.

src/components/WwwFrame/Menus/CustomSecondaryMenu.vue
```diff
@@ -2,15 +2,14 @@
 	<div>
 		<secondary-menu class="show-for-large">
 			<ul class="row">
-				<li v-for="{name, routerLink, url, eventTracking} in menuItems" :key="name">
```
eddieferrer, line 5: This eventTracking is never used

mcstover (approved):

BoulderBrains (approved):

emuvente (approved):
👍

ryan-ludwig (approved):

PR #2328, opened 2020-11-05
Enhancements to layout and styles of campaign mvp. Started moving por…

mcstover:
…tions of the page out into independent components and creating stories for them.


.storybook/stories/CorporateCampaign.stories.js
```diff
@@ -0,0 +1,58 @@
+import Vue from 'vue';
+import StoryRouter from 'storybook-vue-router';
+import { withKnobs, text, object } from '@storybook/addon-knobs';
```
ryan-ludwig, line 3: Minor, these aren't needed since you're using the `args`, and `argTypes` syntax

mcstover: Good to know...

I need to figure out how to segment those automatic options/args so that only the applicable ones show up in each story...

.storybook/stories/CorporateCampaign.stories.js
```diff
@@ -0,0 +1,58 @@
...
+
+import CorporateCampaignLanding from '@/pages/LandingPages/CorporateCampaign/CCLandingPage';
+import CampaignHeader from '@/components/ContentGroups/CampaignHeader';
+import CampaignStatus from '@/components/ContentGroups/CampaignStatus';
+
+export default {
+	title: 'Pages/CorporateCampaignLanding',
+	component: CorporateCampaignLanding,
+	args: {
+		headerAreaContent: {
```
ryan-ludwig, line 20: Since headerAreaContent only applies to CampaignHeader, and the other props apply to StatusRow, I recommend creating two separate story files.

PR #2329, opened 2020-11-05
SUBS-542 DropIn Loading & LoadingOverlay

eddieferrer:
* Tweaks interaction between our loader and the dropIn loader slightly
* Move LoadingOverlay out of lend into Kv folder, fix imports
* Apply KvLoading overlay to AD and MG signup drop ins to display a loading state blocking the UI while the page prefetches on success redirect


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -147,7 +151,7 @@
 					@updating-totals="setUpdatingTotals"
 					@refreshtotals="refreshTotals"
 				/>
-				<loading-overlay v-show="updatingTotals" id="updating-overlay" class="updating-totals-overlay" />
+				<kv-loading-overlay v-if="updatingTotals" id="updating-overlay" class="updating-totals-overlay" />
```
eddieferrer, line 154: There was a small bug here with the way the mounted hook was being fired on v-show which led to the loader sometimes not being in the correct place

mcstover (approved):
Nice cleanup and improvements to sequencing

emuvente (approved):
Nice! 👍

PR #2332, opened 2020-11-06
Wire up in context thanks and improve style of in context checkout.

mcstover:
Relevant updates:
- Moved Campaign Loan Display and pagination handling into it's own component
- Checkout is now embedded in the page after a loan is added to basket rather than showing in a lightbox
- InContext Checkout component has some new options to control showing or hiding the donation row and whether to redirect to thanks automatically.
- In context Thanks is shown after completion of checkout

### DIFF
--- a/src/components/Checkout/InContext/InContextCheckout.vue
+++ b/src/components/Checkout/InContext/InContextCheckout.vue
@@ -1,37 +1,29 @@
 <template>
-	<div>
-		<kv-lightbox
-			:visible="lightboxClosed"
-			@lightbox-closed="exitLightBox"
-			title="Complete your loan"
+	<div class="in-context-checkout">
+		<basket-items-list
+			:class="hideDonationClass"
+			:loans="loans"
+			:donations="donations"
+			:kiva-cards="kivaCards"
+			:loan-reservation-total="parseInt(totals.loanReservationTotal)"
+		/>
+		<hr>
+		<kv-button
+			v-if="!isActivelyLoggedIn"
+			class="smaller checkout-button"
+			id="Continue-to-legacy-button"
+			v-kv-track-event="['basket', 'Redirect Continue Button', 'exit to legacy']"
+			:href="registerOrLoginHref"
 		>
-			<!-- :teams="teams" -->
-			<basket-items-list
-				:loans="loans"
-				:donations="donations"
-				:kiva-cards="kivaCards"
-				:loan-reservation-total="parseInt(totals.loanReservationTotal)"
-			/>
-			<hr>
-			<template v-slot:controls>
-				<kv-button
-					v-if="!isActivelyLoggedIn"
-					class="smaller checkout-button"
-					id="Continue-to-legacy-button"
-					v-kv-track-event="['basket', 'Redirect Continue Button', 'exit to legacy']"
-					:href="registerOrLoginHref"
-				>
-					Continue
-				</kv-button>
+			Continue
+		</kv-button>
 
-				<kiva-credit-payment
-					v-else
-					@complete-transaction="completeTransaction"
-					class=" checkout-button"
-					id="kiva-credit-payment-button"
-				/>
-			</template>
-		</kv-lightbox>
+		<kiva-credit-payment
+			v-else
+			@complete-transaction="completeTransaction"
+			class=" checkout-button"
+			id="kiva-credit-payment-button"
+		/>
 	</div>
 </template>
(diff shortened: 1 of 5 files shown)


src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -111,16 +111,17 @@ export default {
...
-			window.setTimeout(
-				() => {
-					this.redirectToThanks(transactionId);
-				},
-				800
-			);
+			if (this.autoRedirectToThanks) {
+				window.setTimeout(
+					() => {
+						this.redirectToThanks(transactionId);
```
eddieferrer, line 117: Is this defined somewhere?

src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -111,16 +111,17 @@ export default {
...
+					},
+					800
+				);
+			}
+
+			this.$emit('transaciton-complete', transactionData);
```
eddieferrer, line 123: This is misspelled, but consistently misspelled in the parent as well :)

src/components/Checkout/InContext/InContextCheckout.vue
mcstover: lives in checkout-utils-mixin.js

src/components/ContentGroups/CampaignLoanDisplay.vue
```diff
@@ -0,0 +1,236 @@
...
+				this.$router.push({
+					query: {
+						...this.$route.query,
+						...this.urlParams
+					}
+				});
+			}
+		},
+	},
+	// TODO: determine if this logic will work at this level
```
eddieferrer, line 191: It won't work since those guard are only available to the components passed to the router in routes.js. You can possibly do all query/urlToQuery logic and transformations in CCLanding page and pass it to this component as a prop?

src/components/ContentGroups/CampaignLoanDisplay.vue
```diff
@@ -0,0 +1,236 @@
+<template>
+	<section
+		v-if="!checkoutVisible"
+		id="#campaign-loans"
```
eddieferrer, line 4: should be `campaign-loans`

src/components/ContentGroups/CampaignThanks.vue
```diff
@@ -0,0 +1,167 @@
...
+			type: Number,
+			default: null,
+		},
+	},
+	data() {
+		return {
+			lender: {},
+			loans: [],
+			receipt: {},
+			showReciept: false,
```
eddieferrer, line 69: misspelled

emuvente (approved):
nice 👍

src/components/ContentGroups/CampaignThanks.vue
```diff
@@ -0,0 +1,167 @@
...
+		fetchReceipt() {
+			this.apollo.query({
+				query: thanksPageQuery,
+				variables: {
+					checkoutId: this.transactionId
+				}
+			}).then(({ data }) => {
+				this.lender = {
+					...data.my.userAccount,
+					teams: data.my.teams.values.map(value => value.team)
```
eddieferrer, line 96: can we guarantee this will always exist here? Maybe change to optional chaining?

src/components/ContentGroups/CampaignLoanDisplay.vue
mcstover: Yea that's what I thought I remembered too. The route logic only works on the top level component served to the route.

I'm still working out the isolation of pagination logic which this is related to. I can probably do what you've suggested once I create a utils file for pagination, which uses some additional functions to format the `offset`. This is no concern of the page level component so I'd rather not require it at this level. _It's still defined in the data object here but is not used, still a bit more clean up here..._

src/components/ContentGroups/CampaignThanks.vue
mcstover: I could probably add some more checks... This came right out of the thanks page. I just moved it out of the prefetch structure.

src/components/ContentGroups/CampaignLoanDisplay.vue
mcstover: noted. Currently unused.

mcstover:
verified green https://travis-ci.org/github/kiva/ui/builds/741975357

PR #2347, opened 2020-11-11
SUBS-490 Remove Legacy Payment Wrappers from Checkout

eddieferrer:
* Removes legacy payment wrappers from checkout page


mcstover (approved):
!!!

PR #2355, opened 2020-11-13
Grow 297

BoulderBrains:
The api.authenticatiion fails intermittently, but commenting out lines 145 -147 of KvAuth0.js gets around that failure.


src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -65,6 +210,18 @@ export default {
...
+	}
+
+	.section {
+		margin-bottom: rem-calc(50);
+	}
+
+	.sub-section {
+		margin-bottom: 2rem;
+	}
 }
```
BoulderBrains, line 225: I'll fix this to be BEM style

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,21 +88,116 @@
...
 
 <script>
+import gql from 'graphql-tag';
+
 import KvIcon from '@/components/Kv/KvIcon';
 import KvSettingsCard from '@/components/Kv/KvSettingsCard';
+import KvButton from '@/components/Kv/KvButton';
 import TheMyKivaSecondaryMenu from '@/components/WwwFrame/Menus/TheMyKivaSecondaryMenu';
 import WwwPage from '@/components/WwwFrame/WwwPage';
+import removeMfa from '@/graphql/mutation/removeMfa.graphql';
```
emuvente, line 98: Where's this file? I don't see it in the repo or in this pr

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,21 +88,116 @@
...
+		KvButton,
 		TheMyKivaSecondaryMenu,
 		WwwPage,
 	},
 	metaInfo: {
 		title: '2-step verification',
 	},
+	mounted() {
+		if (this.kvAuth0.enabled) {
+			this.kvAuth0.getMfaManagementToken()
```
emuvente, line 130: I think you might have to call `this.kvAuth0.checkSession()` before this to ensure the user is logged in

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,21 +88,116 @@
...
+				otherwise update the URL to include ?mfa=off, which triggers
+				a condition in the mounted hook */
+			if (timeSinceLastLogin >= 5) {
+				window.location = `/ui-login?force=true&doneUrl=${encodedDoneUrl}`;
+			} else {
+				window.location = doneUrl;
+			}
+		},
+		turnOffMfa() {
+			this.apollo.mutate({
```
emuvente, line 194: This will also need the mfa_token variable

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,21 +88,116 @@
...
+	computed: {
+		cardTitle() {
+			if (this.isMfaActive) {
+				return 'Add a backup method';
+			}
+			return 'How would you like your verification code?';
+		},
+		cardSubhead() {
+			if (this.isMfaActive) {
+				return "Set up additional backup steps so you can log in even if your other options aren't available";
```
ryan-ludwig, line 173: minor, should use single quotes here and the line below

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -20,16 +20,66 @@
 							name="auto-icon-on"
 						/>
 					</template>
+
 					<template v-slot:content>
-						<p>You'll be asked for a verification code when accessing you Kiva account.</p>
+						<div v-if="!isMfaActive">
+							<h2>{{ cardTitle }}</h2>
```
ryan-ludwig, line 26: What is the h2 if `isMfaActive` is true?

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -20,16 +20,66 @@
...
+						<div v-if="isMfaActive" class="section">
+							<h3 class="strong">
+								2-step verification is turned on
+							</h3>
+							<p>
+								We'll ask you for your password and verification code from your
+								mobile phone in order to log into your Kiva account.
+							</p>
+							<a
+								@click="checkLastLoginTime"
```
ryan-ludwig, line 39: This feels funny without an `href`. I'd recommend making it a` <kv-button class="text-link" ...`

src/pages/Settings/TwoStepVerificationPage.vue
BoulderBrains: I had that handled on line 54.
But your comment called out the duplication and I've simplified it now.

src/components/Settings/TwoStepVerification.vue
```diff
@@ -20,7 +20,10 @@
 				<span class="strong">both your password and an authenication code from your mobile phone</span>
 				in order to access your account.
 			</p>
-			<kv-button class="smallest">
+			<kv-button
+				href="/settings/security/mfa"
```
ryan-ludwig, line 24: I think this can be a `to` instead of `href` since they're both new stack pages

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,33 +93,143 @@
...
 		TheMyKivaSecondaryMenu,
 		WwwPage,
 	},
 	metaInfo: {
 		title: '2-step verification',
 	},
+	mounted() {
+		if (this.kvAuth0.enabled) {
+			this.kvAuth0.checkSession();
+			this.kvAuth0.getMfaManagementToken()
```
emuvente, line 135: You need to wait for the `checkSession` promise to complete before calling `getMfaManagementToken`. You can do that by changing these lines to
```
this.kvAuth0.checkSession()
	.then(() => this.kvAuth0.getMfaManagementToken())
	.then(token => {
		...
```

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -38,33 +93,143 @@
...
+				otherwise update the URL to include ?mfa=off, which triggers
+				a condition in the mounted hook */
+			if (timeSinceLastLogin >= 5) {
+				window.location = `/ui-login?force=true&doneUrl=${encodedDoneUrl}`;
+			} else {
+				window.location = doneUrl;
+			}
+		},
+		turnOffMfa(token) {
+			this.apollo.mutate({
```
emuvente, line 199: Instead of taking a token as an argument, you'll need to fetch it again, like in my comment above
```
this.kvAuth0.checkSession()
	.then(() => this.kvAuth0.getMfaManagementToken())
	.then(token => {
		this.apollo.mutate({
```

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -10,7 +10,7 @@
 		</div>
 		<div class="row">
 			<div class="column small-12 large-8">
-				<kv-settings-card title="How would you like your verification code?">
```
ryan-ludwig, line 13: Would recommend setting this to `:title="cardTitle"` and removing the h3 with cardTitle below. I may be mistaken, but we should have an overarching `<h2>` for this card which this title prop provides us.

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -20,16 +20,71 @@
...
+							</h3>
+							<p>{{ cardSubhead }}</p>
 
-						<h3>
-							Text message or phone call
-						</h3>
+							<div class="two-step-verification__section-sub-section">
+								<h3 class="strong">
+									Authentication app
+									<span class="two-step-verification__section-sub-section-green">(Recommended)</span>
```
ryan-ludwig, line 65: I'm getting fussy, but this could just be a utility class of `class="green"`

BoulderBrains: Not fussy, that's much cleaner. Thanks

src/pages/Settings/TwoStepVerificationPage.vue
BoulderBrains: So the issue with using the card title, is the location of the card title when a user has MFA enabled. In that case, the card title needs to move below the the newly shown "2-step verification is turned on" and "security method(s)" pieces of the page. 
MFA_Step_2_v1_Sept_14_2020_RubyPark__–_Figma-3

emuvente (approved):
👍

eddieferrer (approved):

BoulderBrains:
@ryan-ludwig here are the Kv-settings-card changes we discussed: 

MFA OFF: 
Screen Shot 2020-11-19 at 1 13 46 PM


MFA ON:
Screen Shot 2020-11-19 at 1 00 59 PM

ryan-ludwig (approved):
Nice!

PR #2360, opened 2020-11-17
GROW-313 swashie appeal banner

ryan-ludwig:


src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -5,13 +5,8 @@
...
 		/>
-		<appeal-banner
-			v-if="showAppeal"
-			:appeal-banner-content="appealBannerContent.fields"
-		/>
-		<appeal-banner-15
-			v-if="show15YearAppeal"
-			:cta-link="appeal15link"
+		<appeal-banner-swashie-container
+			v-if="customAppealEnabled"
```
mcstover, line 9: Please promote this new swashie face banner to be the defacto "Appeal Banner" both here and in Contentful. The "Custom Appeal" type should be for more one-off endeavors such as the Kiva 15.

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBannerSwashieContainer.vue
```diff
@@ -0,0 +1,115 @@
...
+			this.isOpen = bannerIsOpen === 'true';
+		}
+	},
+	methods: {
+		onToggleBanner(isOpen) {
+			this.isOpen = isOpen;
+			cookieStore.set('appeal_banner_is_open', isOpen, { path: '/' });
+		},
+		onAmountSelected(amount) {
+			console.log(amount);
```
eddieferrer, line 84: extra log here :)

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -76,10 +60,6 @@ export default {
 		preFetch(config, client) {
```
eddieferrer, line 60: Without the experiment query, we can just switch this back to the default preFetch: 
`query: bannerQuery,
preFetch: true`

src/components/WwwFrame/TheHeader.vue
```diff
@@ -121,7 +121,10 @@
 					>
 						<span>
 							<span class="amount">{{ balance | numeral('$0') }}</span>
-							<img :src="profilePic">
```
ryan-ludwig, line 124: sneaking in an a11y fix

src/util/expander.js
```diff
@@ -31,16 +31,17 @@ export function expand(el, {
 	// set the property to the 'from' value
 	el.style[property] = from;
 
+	el.addEventListener('transitionend', function listener() {
```
ryan-ludwig, line 34: This prevents an occasional timing issue where the transition isn't quite finished, but the previous setTimeout is reached.

ryan-ludwig:
Would love a little extra QA on this if anyone has some extra time.

src/util/expander.js
```diff
@@ -31,16 +31,17 @@ export function expand(el, {
...
+		unsetStyles(el, { property });
+		el.removeEventListener('transitionend', listener, true);
+		// finally, call the done callback after the transition
+		done();
+	}, true);
+
 	// ...and set the property to the measured value on the next tick so it animates w/ css
 	window.setTimeout(() => {
 		el.style[property] = propValue;
-	}, 1);
```
emuvente, line 37: The delay for this timeout still needs to be 1 so that this value is set on the next rendering loop instead of the current rendering loop (ditto for the same timeout in the collapse function)

src/util/expander.js
emuvente: Nice!

mcstover (approved):
So excite!

BoulderBrains (approved):

src/util/expander.js
ryan-ludwig: @emuvente a setTimeout of 0 should run in the next tick which works here (otherwise we wouldn't see the transition at all). Another way to init the transition would be to force a reflow calculation by asking for the offsetWidth or something like it before setting the property again. A more detailed explanation here https://stackoverflow.com/a/60686560

ryan-ludwig:
Merging to get this in before branch cut

emuvente (commented):
👍

PR #2363, opened 2020-11-20
Instead of settimeout, force a reflow to avoid timing bugs

ryan-ludwig:
Seeing the banner sometimes get stuck in a closed position. After this change haven't seen it totally bork yet. Would like to do more browser testing once it's in dev.


ryan-ludwig:
Sorry for the churn on this

emuvente (approved):
Nice! I had to read up on that hack; it's an interesting one!

PR #2367, opened 2020-11-21
grow-298 mfaMethods and data gathered, v-for in place to create list …

BoulderBrains:
…of mfa methods for user. mutation getting basket id instead of this.authId

### DIFF
--- a/src/pages/Settings/TwoStepVerificationPage.vue
+++ b/src/pages/Settings/TwoStepVerificationPage.vue
@@ -61,10 +61,21 @@
 					</template>
 
 					<template v-slot:content>
-						<div>
-							<h4>Authenticator app</h4>
-							<a>REMOVE LINK</a>
-						</div>
+						<ul class="methodsList">
+							<li
+								class="method"
+								v-for="(mfaMethod, index) in mfaMethods" :key="index"
+							>
+								<h4>{{ mfaMethod.authTypeName }}</h4>
+								<p>{{ mfaMethod.authNumber }}</p>
+								<kv-button
+									class="text-link"
+									@click.native.prevent="removeMfaMethod"
+								>
+									Remove
+								</kv-button>
+							</li>
+						</ul>
 					</template>
 				</kv-settings-card>
 
@@ -126,6 +137,7 @@ import KvButton from '@/components/Kv/KvButton';
 import TheMyKivaSecondaryMenu from '@/components/WwwFrame/Menus/TheMyKivaSecondaryMenu';
 import WwwPage from '@/components/WwwFrame/WwwPage';
 import removeMfa from '@/graphql/mutation/removeMfa.graphql';
+import removeOneMfaMethod from '@/graphql/mutation/removeOneMfaMethod.graphql';
 
 const mfaQuery = gql`query mfaQuery($mfa_token: String!) {
 	my {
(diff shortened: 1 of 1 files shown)


src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -205,8 +201,57 @@ export default {
...
 			return 'You\'ll be asked for a verification code when accessing you Kiva account.';
 		}
+
 	},
 	methods: {
+		gatherMfaEnrollments() {
+			this.kvAuth0.checkSession()
+				.then(() => this.kvAuth0.getMfaManagementToken())
+				.then(token => {
+					return this.apollo.query({
```
emuvente, line 211: In my testing this method was correctly only called after `removeMfaMethod` had completed, however instead of making a new request to the network it just read the values from the cache, which is probably why it seemed like it was being called before `removeMfaMethod` had completed. To fix that, you can add `fetchPolicy: 'network-only',` to the options for this query.

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -290,14 +260,57 @@ export default {
...
+			console.log('formatMfaMethods(authEnrollments)', authEnrollments);
+
+			// When a user adds a text based recovery there are 2 new methods added to their authEnrollments
+			// 1) 1: {id: "sms|dev_tRkQzShZEIDAD0Q9", active: true, authenticator_type: "oob", oob_channel: "sms", name: "XXXXXXXX0367", …}
+			// 2) 2: {id: "voice|dev_tRkQzShZEIDAD0Q9", active: true, authenticator_type: "oob", oob_channel: "voice", name: "XXXXXXXX0367", …}
+			// Since the user only added 1 authentication method we only want to display 1 of these
+			// in their list of authentication methods
+
+			// My first thought was to remove one of the values based on the ID, since they come through with a similar ID
+			// ie. "sms|dev_tRkQzShZEIDAD0Q9" & "voice|dev_tRkQzShZEIDAD0Q9"
```
mcstover, line 279: you could split the ID on the `|` character then use the 2nd item in the resulting array.

How about for the the authenticator_type are there other `oob` options beyond voice/sms?

Also, I'm curious if we know which of these options the lender actually "selected" when setting this up? We should show the type they selected during setup.

If they add a phone number does it create both entries or is it only when they select "sms/text" recovery that 2 similar entries are created? If the duplicate is only created when selecting "sms/text" then we can always show that one if it's present.

emuvente (approved):
Looks great! 👍

src/pages/Settings/TwoStepVerificationPage.vue
BoulderBrains: - Thanks for the .split("|") suggestion for the ID @mcstover, that's what I've implemented in my latest commit.
- I'm not aware of other oob options beyond voice/sms
- We don't have a way of knowing if a user chooses text or voice when setting up the MFA method.
- The duplicate method is shown when the user adds a phone number. During the recovery process I believe the user can choose to receive a text or voice call for recovery, leading to the duplicate entry in authEnrollments. For the front end it doesn't make sense to display both due to them being tied together (if one is deleted, the matching one is too.) That's why I wanted it filtered out for display.

mcstover (approved):

ryan-ludwig (approved):
👍

PR #2368, opened 2020-11-23
Add facebook events for ftd and kiva cards during transaction

mcstover:
Use new isFirstTimeDepositor flag in transaction analytics and add new analytics events if a transaction includes a Kiva card.

Depends on https://github.com/kiva/kiva/pull/9458


emuvente (approved):
👍

PR #2371, opened 2020-11-24
SUBS-582: Use a configurable note size rather than a hardcoded share price

remoun:
Uses the GraphQL field defined in https://github.com/kiva/kiva/pull/9463 but defaults to $25 without it. Also adds a couple of `25_hard_coded` tags.

Screenshot below. See [SUBS-582](https://kiva.atlassian.net/browse/SUBS-582) for before picture.

$5-in-new-UI

### DIFF
--- a/src/components/Checkout/LoanPrice.vue
+++ b/src/components/Checkout/LoanPrice.vue
@@ -29,6 +29,7 @@ import _forEach from 'lodash/forEach';
 import updateLoanReservation from '@/graphql/mutation/updateLoanReservation.graphql';
 import updateKivaCardAmount from '@/graphql/mutation/updateKivaCardAmount.graphql';
 import KvIcon from '@/components/Kv/KvIcon';
+import { buildPriceArray } from '@/util/loanUtils';
 
 export default {
 	components: {
@@ -97,16 +98,8 @@ export default {
 				// if we've met reserve ensure at least this loan share is set
 				remainingAmount = Math.max(remainingAmount, parseFloat(this.price, 10));
 
-				// convert this to formatted array for our select element
 				const minAmount = parseFloat(this.minAmount || 25, 10); // 25_hard_coded
-				const numShares = remainingAmount / minAmount;
-				const priceArray = [];	// ex. priceArray = ['25', '50', '75']
-				// loop and build formatted array
-				// eslint-disable-next-line no-plusplus
-				for (let i = 1; i <= numShares; i++) {
-					priceArray.push(numeral(minAmount * i).format('0,0'));
-				}
-				return priceArray;
+				return buildPriceArray(remainingAmount, minAmount);
 			}
 			if (this.type === 'kivaCard') {
 				// convert this to formatted array for our select element
--- a/src/components/Checkout/LoanPrice.vue
+++ b/src/components/Checkout/LoanPrice.vue
@@ -52,6 +52,10 @@ export default {
 			type: String,
 			default: ''
 		},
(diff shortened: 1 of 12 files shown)


src/components/Checkout/LoanPrice.vue
```diff
@@ -81,24 +85,28 @@ export default {
...
 				// convert this to formatted array for our select element
-				const reservePriceArray = this.buildShareArray(sharesBelowReserve);
-
-				return reservePriceArray;
+				const minAmount = parseFloat(this.minAmount || 25, 10); // 25_hard_coded
+				const numShares = remainingAmount / minAmount;
+				const priceArray = [];	// ex. priceArray = ['25', '50', '75']
+				// loop and build formatted array
+				// eslint-disable-next-line no-plusplus
+				for (let i = 1; i <= numShares; i++) {
```
eddieferrer, line 106: We can change this line to: 
`for (let i = 1; i <= numShares; i+=1)`
And remove the no-plusplus exception
Reference: https://eslint.org/docs/rules/no-plusplus

src/graphql/query/algoliaLoanStatus.graphql
```diff
@@ -9,6 +9,7 @@ query algoliaLoanStatus(
 				id
 				status
 				matchingText
+				minNoteSize
```
eddieferrer, line 12: I'm _ok_ with adding minNoteSize to all these queries, even though some of them are not using it.

eddieferrer:
I think we'll also possibly to take a look at the drop down share price builder in `src/components/LoanCards/Buttons/LendIncrementButton.vue` ?

mcstover (approved):

src/graphql/query/algoliaLoanStatus.graphql
remoun: Good to hear, thank you. I'm admittedly not sure which ones really need it, particularly this one.

src/components/Checkout/LoanPrice.vue
remoun: Switched to `Array(N).map((_, i) => ...)`

src/graphql/query/algoliaLoanStatus.graphql
remoun: If there are queries you're sure don't need this addition, I'll gladly remove them.

src/components/Checkout/LoanPrice.vue
```diff
@@ -80,25 +85,19 @@ export default {
...
-				let remainingShares = parseFloat(this.loanAmount) - parseFloat(this.fundedAmount);
-
-				// subtract reservedAmount shares (minus our own reserved shares)
+				let remainingAmount = parseFloat(this.loanAmount) - parseFloat(this.fundedAmount);
+				// subtract reservedAmount (minus our own reserved amount)
 				// - only do this for loans that are not ending soon
 				// - for loans ending soon we just show remaining shares which are all un-reserved
 				if (!this.isExpiringSoon) {
-					remainingShares -= (parseFloat(this.reservedAmount) - parseInt(this.price, 10));
+					remainingAmount -= (parseFloat(this.reservedAmount) - parseFloat(this.price, 10));
```
emuvente, line 93: `parseFloat` doesn't take a second argument, so in this case it would just be `parseFloat(this.price)` (same comment for the other `parseFloat` usages below)

emuvente (approved):
👍

mcstover (approved):

PR #2374, opened 2020-11-24
GROW-353 Homepage styling changes

emuvente:


src/components/Kv/KvCarousel.vue
```diff
@@ -195,6 +195,7 @@ export default {
 		this.embla = EmblaCarousel(this.$refs.KvCarousel, {
 			loop: true,
 			containScroll: 'trimSnaps',
+			inViewThreshold: 0.90,
```
emuvente, line 198: I found that the first slide on page load was being pushed to the left, causing the text inside it to be cut-off. However, when I resized the window, the slide moved to the right and was no longer cutting off the text. I noticed that this `inViewThreshold` option is set when resizing, so I tried including it here in the initial options and the problem went away (i.e. the first slide was no longer being cutoff). My guess is that we should actually be setting this option to 1, so that the full slide always appears, but didn't want to change things too much without checking the other usages of the carousel.

ryan-ludwig (approved):
👍

mcstover (approved):

PR #2376, opened 2020-11-24
GROW-354 Change homepage loan cards to have "View loan" button instead of lend button

emuvente:


mcstover (approved):
Very direct! Was thinking this might actually be a good option to live within the Action Button at some point...

PR #2377, opened 2020-11-24
VUE-456 update storybook to work with node15

eddieferrer:
This is for node 15. 

Initial install requires `--legacy-peer-deps` because of an issue with `storybook-vue-router`


src/components/CorporateCampaign/CampaignVerificationForm.vue
```diff
@@ -9,7 +9,7 @@
 					:width="iFrameWidth"
 					frameborder="0"
 				></iframe>
-				<script src="//kiva.tfaforms.net/js/iframe_resize_helper.js"></script>
```
ryan-ludwig, line 12: oh nice, to get my local working today I removed this from the html and added it to metaInfo
```
	metaInfo: {
		script: [
			{ src: '//kiva.tfaforms.net/js/iframe_resize_helper.js', async: true }
		],
	},
```

Not sure if there's a preferred way or not

ryan-ludwig (approved):

src/components/CorporateCampaign/CampaignVerificationForm.vue
eddieferrer: @ryan-ludwig can you verify if my fix works for you (I'm assuming you're on node 12, npm 6, storybook 6.x)

mcstover: I'll test this change out locally and update the PR I have just pushed that uses this if all is well.

eddieferrer:
I'm closing this PR as to not muck up the Node 15 stuff. I have confirmed that with the fixes to `CampaignVerificationForm.vue` Storybook 6.0.28 works on Node 15/NPM 7

PR #2356, opened 2020-11-13
Run travis on node-v15 branch

mcstover:

PR #2380, opened 2020-11-25
VUE-455 npm run test for node15

eddieferrer:
Fixes linting and unit testing. 

Note, cypress still fails because of the `--http-server-default-timeout=240000` in start script. That change is in https://github.com/kiva/ui/pull/2356/files


eddieferrer:
This isn't needed

PR #2381, opened 2020-11-25
GROW-352 banner changes

ryan-ludwig:
It's not 100% perfect, but will be enough to review in the morning with stakeholders


emuvente (approved):
👍

mcstover (approved):

PR #2386, opened 2020-11-25
Render newlines as line breaks in the appeal banner body

emuvente:
I was looking into how to get line breaks created in contentful rich text (using `Shift + Enter`) to be rendered as `<br />`, and found [a comment](https://github.com/contentful/rich-text/issues/69#issuecomment-479790401) pointing me in the direction of this whitespace value `pre-wrap` that works like this: "Sequences of white space are preserved. Lines are broken at newline characters, at `<br>`, and as necessary to fill line boxes."

So I added a line break into the rich text for the appeal banner and tried `pre-wrap` and it worked:
Screen Shot 2020-11-25 at 12 40 26 PM


mcstover (approved):
Nice find. Prob something we'll want to use in several places that load up rich text entries

ryan-ludwig:
Nice find!

ryan-ludwig (approved):

PR #2389, opened 2020-11-25
GROW-313 smoother transitions

ryan-ludwig:
Since we were transitioning on height on the appeal banner, we ended up with a lot of jank. This package does some interesting stuff behind the scenes to use transforms for everything instead. Ends up smooth.

Don't love throwing this in at the last second without much testing though. Adds 10k too.
https://bundlephobia.com/result?p=vue-smooth-reflow@0.1.12

Thoughts?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33329,6 +33329,11 @@
 				}
 			}
 		},
+		"vue-smooth-reflow": {
+			"version": "0.1.12",
+			"resolved": "https://registry.npmjs.org/vue-smooth-reflow/-/vue-smooth-reflow-0.1.12.tgz",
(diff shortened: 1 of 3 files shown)


emuvente:
I'm not thrilled about the 10k, but the code looks like it doesn't have any extra fluff, and it looks like we could use this in place of anything that we're currently using `expander` for, so eventually we'll be able to drop that which makes this not as big of an add.

emuvente (approved):

ryan-ludwig:
This could be componentized to behave more like Vue's `<transition>` el.
https://www.npmjs.com/package/vue-smooth-reflow#smooth-component

mcstover:
Does this ensure we pre-render the initial open or closed state from the server in order to prevent the full re-paint due to immediate change in promo state?

ryan-ludwig:
@mcstover No changes in the initial open/closed state from the server.

mcstover (approved):

ryan-ludwig:
Going to merge this so we can start playing with it in dev.

PR #2394, opened 2020-11-30
grow-363 adding tracking links into footer

BoulderBrains:
Tracking events have been added to all clicks within TheFooter.vue based on the structure laid out in grow-363

### DIFF
--- a/src/components/WwwFrame/TheFooter.vue
+++ b/src/components/WwwFrame/TheFooter.vue
@@ -3,36 +3,58 @@
 		<nav class="small-footer hide-for-large" aria-label="Footer navigation">
 			<ul class="hide-for-print">
 				<li>
-					<router-link :to="applyUrl">
+					<router-link
+						:to="applyUrl"
+						v-kv-track-event="['Footer', 'click-Borrow-mobile view']"
+					>
 						Borrow
 					</router-link>
 				</li>
 				<li>
-					<router-link :to="aboutUrl">
+					<router-link
+						:to="aboutUrl"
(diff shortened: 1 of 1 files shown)


emuvente (approved):
👍

src/components/WwwFrame/TheFooter.vue
```diff
@@ -3,36 +3,58 @@
 		<nav class="small-footer hide-for-large" aria-label="Footer navigation">
 			<ul class="hide-for-print">
 				<li>
-					<router-link :to="applyUrl">
+					<router-link
+						:to="applyUrl"
+						v-kv-track-event="['Footer', 'click-Borrow-mobile view']"
```
ryan-ludwig, line 8: My instinct is that we could have the same tracking events for mobile and desktop

ryan-ludwig: Maybe double check with @brandonatkiva

BoulderBrains: You're instincts were right @ryan-ludwig. I connected with Brandon and he requested that the mobile links be mapped to analogous options from the desktop version of the footer. Changes made in latest commit.

ryan-ludwig (approved):

mcstover:
@BoulderBrains @emuvente These should probably added to styleguide as well so the same events are fired across the entire site.

emuvente:
@mcstover Yeah, I asked Brandon about that. This change is mainly being requested for being able to track interactions on the homepage, so for the sake of reducing the time to make this hotfix we're doing it in the new stack first and in the old stack in a follow-up ticket.

PR #2397, opened 2020-11-30
grow-358 subhead copy update

BoulderBrains:
Small copy update for the sub-heading on the homepage

Copy now reads: 
With Kiva you can lend as little as $25 and make a big change in someone's life.

Screen Shot 2020-11-30 at 3 56 15 PM


emuvente (approved):
👍

ryan-ludwig (approved):

PR #2400, opened 2020-12-01
include ability to create queries with Loan Tags from an LSC. Minor t…

mcstover:
…weaks and clean up to campaign page and components

The update to include loanTags should start fixing up ML2-108

### DIFF
--- a/src/api/fixtures/LoanSearchFilters.js
+++ b/src/api/fixtures/LoanSearchFilters.js
@@ -24,6 +24,7 @@ export default function LoanSearchFilters() {
 		partner: null,
 		riskRating,
 		sector: null,
+		loanTags: null,
 	};
 }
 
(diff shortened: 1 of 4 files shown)


src/components/CorporateCampaign/CampaignLoanDisplay.vue
```diff
@@ -11,14 +11,15 @@
 		</div> -->
 		<div class="columns small-12 large-8 align-self-middle" v-if="showLoans && loans.length > 0">
 			<div class="loan-card-group row small-up-1 large-up-2 xxlarge-up-3">
+				<!-- LendHomepageLoanCard -->
```
mcstover, line 14: temporary switch due to removal of action button

src/api/fixtures/LoanSearchFilters.js
```diff
@@ -68,10 +69,12 @@ export function getInputFilters(filters) {
 // Return filters that can be used in a loan search
 export function getSearchableFilters(filters) {
```
mcstover, line 70: This is the only method I'm currently using to create a loanQuery for the campaign pages.

mcstover: Addition of `loanTags` in the other methods don't appear to have any affect on the functionality of auto lending so I think they are ok.

emuvente (approved):
👍  it's cool to see the fixtures being expanded upon!

PR #2409, opened 2020-12-03
GROW-316 international phone number input

ryan-ludwig:
If you have a moment, please pull down, run a local storybook and take a look at the stories.

### DIFF
--- a/src/components/Kv/KvPhoneInput.vue
+++ b/src/components/Kv/KvPhoneInput.vue
@@ -13,6 +13,7 @@
 			>
 				Select your country
 			</label>
+
 			<kv-dropdown-rounded
 				class="kv-phone-input__country-select"
 				:id="`country_select_${id}`"
@@ -91,24 +92,28 @@ export default {
 		};
 	},
 	computed: {
+		countryCallingCode() {
+			return this.asYouTypeFormatter?.getNumber()?.countryCallingCode
+			|| getCountryCallingCode(this.selectedCountryCode);
+		},
+		e164Value() { // E.164 formatted phone number
+			return this.asYouTypeFormatter?.getNumber()?.number || '';
+		},
 		isEmpty() {
 			return this.displayValue.length === 0;
 		},
 		isValid() {
 			return this.asYouTypeFormatter?.isValid();
 		},
-		countryCallingCode() {
-			return this.asYouTypeFormatter?.getNumber()?.countryCallingCode
-			|| getCountryCallingCode(this.selectedCountryCode);
-		},
 		placeholderNumber() {
 			return getExampleNumber(this.selectedCountryCode, exampleNumbers)?.formatNational() || '';
 		},
-		e164Value() { // E.164 formatted phone number
-			return this.asYouTypeFormatter?.getNumber()?.number || '';
-		}
 	},
 	methods: {
+		formatNumber(val) {
+			this.asYouTypeFormatter = new AsYouType(this.selectedCountryCode);
+			return this.asYouTypeFormatter.input(val);
+		},
 		onInputCountry() {
 			this.displayValue = this.formatNumber(this.displayValue);
 		},
@@ -121,10 +126,6 @@ export default {
 			 * @type {Event}
 			 */
 			this.$emit('input', this.e164Value);
-		},
-		formatNumber(val) {
-			this.asYouTypeFormatter = new AsYouType(this.selectedCountryCode);
-			return this.asYouTypeFormatter.input(val);
 		},
 	},
 };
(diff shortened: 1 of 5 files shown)


src/components/Kv/KvPhoneInput.vue
```diff
@@ -0,0 +1,176 @@
...
+			return this.asYouTypeFormatter?.isValid();
+		},
+		placeholderNumber() {
+			return getExampleNumber(this.selectedCountryCode, exampleNumbers)?.formatNational() || '';
+		},
+	},
+	methods: {
+		async init() {
+			// If we're passed in a non US number from the start, set the country and reformat the number.
+			this.asYouTypeFormatter.input(this.value);
```
ryan-ludwig, line 114: This method feels a little funny and could prob be written better.

src/components/Kv/KvPhoneInput.vue
```diff
@@ -0,0 +1,176 @@
...
+		isValid() {
+			return this.asYouTypeFormatter?.isValid();
+		},
+		placeholderNumber() {
+			return getExampleNumber(this.selectedCountryCode, exampleNumbers)?.formatNational() || '';
+		},
+	},
+	methods: {
+		async init() {
+			// If we're passed in a non US number from the start, set the country and reformat the number.
```
ryan-ludwig, line 113: This method feels a little funny and could prob be written better.

emuvente:
Playing around with it I noticed that I can't backspace to delete the whole number. It gets stuck on the close parenthesis. 
Screen Shot 2020-12-04 at 12 26 21 PM

emuvente:
Other than that odd issue, everything was working great!

ryan-ludwig:
I think this is ready for review now @emuvente @eddieferrer . I'm assuming we'll handle validation in the parent page/component.

src/components/Kv/KvPhoneInput.vue
```diff
@@ -0,0 +1,201 @@
...
+			this.displayNumber = this.formatPhoneNumber(e.target.value);
+			this.emitUpdatedNumber();
+		},
+		emitUpdatedNumber() {
+			/**
+			 * The value of the input in E.164 formatting
+			 * @event input
+			 * @type {Event}
+			 */
+			this.$emit('input', this.e164Number);
```
ryan-ludwig, line 140: I think to do validation in a parent, we'll need need to emit the selected country as well.

src/components/Kv/KvPhoneInput.vue
```diff
@@ -0,0 +1,201 @@
...
+		},
+		e164Number() { // E.164 formatted phone number
+			const phoneNumber = parsePhoneNumberFromString(this.displayNumber, this.selectedCountryCode);
+			return phoneNumber?.number || '';
+		},
+		isEmpty() {
+			return this.displayNumber.length === 0;
+		},
+		isValid() {
+			const phoneNumber = parsePhoneNumberFromString(this.displayNumber, this.selectedCountryCode);
```
ryan-ludwig, line 109: And move this code to the parent validator

src/components/Kv/KvPhoneInput.vue
mcstover: Yea I think this could be an object that always includes the `valid` flag as well...

src/components/Kv/KvPhoneInput.vue
mcstover: Well probably not as that method comes out of the lib imported here...

Seems like this component should house the basic validation based on the lib you're using. Then if we need to do additional validation based on our own data or some other criteria that would be either plugged-in here or dealt with in a/the wrapping component.

eddieferrer:
This looks really good.

I noticed in the two examples on storybook. Default vs Non-US country the value prop is different. For the default, the value prop is: `value:"828-479-5482"` where for non-us value prop is: `value:"+58 0412-1234567"` So one includes the country code but the other doesnt. Will this cause issues?

src/components/Kv/KvPhoneInput.vue
ryan-ludwig: If we emit an object, I dont' think we'd be able to do the traditional v-model bindings like 
`<kv-phone-input v-model="myCoolPhoneNumber" />`
We could emit a separate event like 'validity-changed' or something like that though

ryan-ludwig:
@eddieferrer If no country code is included in the prop it will default to US.

src/components/Kv/KvPhoneInput.vue
ryan-ludwig: @mcstover good call 👍

eddieferrer (approved):

emuvente (approved):
👍

PR #2410, opened 2020-12-03
GROW-279 Create page to allow users to initiate email verification

emuvente:
This page is used to initiate email verification when MFA is not available. Various other pages (like /withdraw and /settings/security) will redirect here if the user has not authenticated with MFA.

### DIFF
--- a/src/pages/StartVerification.vue
+++ b/src/pages/StartVerification.vue
@@ -1,0 +1,159 @@
+<template>
+	<www-page class="start-verification-page">
+		<div class="page-content row align-center">
+			<div class="columns shrink">
+				<template v-if="!sent">
+					<h1>Email verification required</h1>
+					<p>To ensure your safety, we added an extra layer of security.</p>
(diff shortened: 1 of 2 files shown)


emuvente:
This depends on https://github.com/kiva/kiva/pull/9483

ryan-ludwig:
Looks good, my only question is about having this as a top-level page. Would it make sense as a child of `/settings/security/start-verification`? or even `/settings/security/mfa/start-verification` ?

ryan-ludwig (approved):

emuvente:
> Would it make sense as /settings/security/start-verification? or even /settings/security/mfa/start-verification ?

This isn't really a settings page, so I'm not sure if putting it under settings would make sense. I put it at the top level because that we're we've placed some other auth-related pages so far (Error.vue, ProcessBrowserAuth.vue). I'd be down for putting it somewhere else though.

PR #2415, opened 2020-12-04
Drop sanitize html

emuvente:


.stylelintrc
```diff
@@ -8,7 +8,7 @@
     "scss/at-rule-no-unknown": true,
     "at-rule-empty-line-before": null,
     "no-empty-source": null,
-    "unit-whitelist": ["em", "rem", "%", "px", "deg", "vh", "vw", "ms", "s", "fr"],
+    "unit-allowed-list": ["em", "rem", "%", "px", "deg", "vh", "vw", "ms", "s", "fr"],
```
emuvente, line 11: Stylelint was giving a deprecation warning about this renamed config value.

package.json
```diff
@@ -170,6 +169,7 @@
 		"mini-css-extract-plugin": "^0.10.0",
 		"null-loader": "^4.0.0",
 		"ora": "^4.0.2",
+		"postcss": "^7.0.35",
```
emuvente, line 172: Postcss v8 was being included by sanitize-html, and that was satisfying the peer dependency requirement that stylelint has for postcss >=7. Without sanitize-html, we need to include postcss ourselves.

.stylelintrc
eddieferrer: These were added in: https://github.com/stylelint/stylelint/releases/tag/13.7.0 So I think we need to also update package.json to `"stylelint": "^13.7.0"`

mcstover (approved):

eddieferrer (approved):

PR #2418, opened 2020-12-09
GROW-358 Use a contentful image for the appeal banner

ryan-ludwig:
image


src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBannerCircular.vue
```diff
@@ -159,7 +148,7 @@ export default {
 			type: String,
 			default: '',
 		},
-		image: {
+		imageUrl: {
 			type: String,
 			default: ''
```
mcstover, line 153: I think we need a default here that's not controlled in contentful...

mcstover (approved):
Looks good.

We should probably consider adding an id field to the Responsive Image Content Type that is more code friendly...but in the meantime just adding documentation for how this should be used is prob fine.

eddieferrer:
Looks good. I think we're getting a little hacky with the Responsive Image Set on contentful, and should maybe consider adding a different content type that just links to a single item of media for cases like this...

eddieferrer (approved):

mcstover:
> Looks good. I think we're getting a little hacky with the Responsive Image Set on contentful, and should maybe consider adding a different content type that just links to a single item of media for cases like this...

We could make the Media type available so we can include the media directly when it doesn't require any responsive versions.

ryan-ludwig:
I agree, 

> Looks good. I think we're getting a little hacky with the Responsive Image Set on contentful, and should maybe consider adding a different content type that just links to a single item of media for cases like this...

ryan-ludwig:
To follow up, we'll need to come back around to changing the media type in the future.

PR #2419, opened 2020-12-09
GROW-275 Require mfa or email verification to access security settings

emuvente:


config/index.js
```diff
@@ -44,9 +44,7 @@ module.exports = {
 			browserCallbackUri: 'https://www.kiva.org/process-browser-auth',
 			serverCallbackUri: 'https://www.kiva.org/process-ssr-auth',
 			domain: 'login.kiva.org',
-			scope: 'https://www.kiva.org/last_login ' +
-				'https://www.kiva.org/kiva_id ' +
-				'openid email profile',
+			scope: 'openid mfa',
```
emuvente, line 47: The `email` and `profile` scopes add PII to the id token, and it isn't being used, so I removed those scopes.
The `last_login` and `kiva_id` scopes aren't needed because those claims are always added to the tokens no matter what scope is requested.

mcstover (approved):

ryan-ludwig (approved):

mcstover (approved):

PR #2423, opened 2020-12-11
GROW-355 Initial pass at verification code input

ryan-ludwig:
image

### DIFF
--- a/src/components/Kv/KvVerificationCodeInput.vue
+++ b/src/components/Kv/KvVerificationCodeInput.vue
@@ -1,0 +1,95 @@
+<template>
+	<div
+		class="kv-verification-code-input"
+		:style="cssVars"
+	>
+		<!--
+			How I imagine this to work with multiple inputs:
+			multiple <inputs hidden from screenreaders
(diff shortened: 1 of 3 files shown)


src/components/Kv/KvVerificationCodeInput.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+	<div
+		class="kv-verification-code-input"
+		:style="cssVars"
+	>
+		<!--
+			How I imagine this to work with multiple inputs:
+			multiple <inputs hidden from screenreaders
+			1 input hidden from sighted with the autocomplete tag.
+			the values get updated in sync.
```
emuvente, line 10: That makes sense! I think the tricky part will be focusing on the correct input, specifically when deleting/backspacing.

emuvente (approved):
This looks great!

src/components/Kv/KvVerificationCodeInput.vue
ryan-ludwig: oh good call, I hadn't thought about backspacing

PR #2429, opened 2020-12-14
GROW-303 phone setup UI

ryan-ludwig:
Part 1 of GROW-303. 
There's work to be done to make the GraphQL mutations/queries still. Would like to handle that in a separate PR


.babelrc
```diff
@@ -16,7 +16,14 @@
 	],
 	"plugins": [
 		"@babel/plugin-syntax-dynamic-import",
-		"@babel/plugin-syntax-import-meta"
+		"@babel/plugin-syntax-import-meta",
+		"@babel/plugin-transform-modules-commonjs",
+		[
+			"@babel/plugin-transform-runtime",
+			{
+				 "regenerator": true
```
ryan-ludwig, line 24: This is in response to an error I was getting from the AsYouTypeFormatter in libphonenumber-js (in KvPhoneInput.vue). 

I'm really not sure what the implications of this are in prod. Are there performance concerns?

image

src/components/Settings/PhoneAuthentication.vue
```diff
@@ -0,0 +1,178 @@
...
+				>
+					Enter your phone number here:
+				</label>
+				<kv-phone-input
+					class="phone-authentication__phone-input"
+					:class="{ 'phone-authentication__phone-input--error' : $v.phoneNumber.$error}"
+					id="phone_input"
+					ref="phoneInput"
+					v-model="phoneNumber"
+					@blur="$v.phoneNumber.$touch"
```
ryan-ludwig, line 21: @eddieferrer I think I finally figured out how to get Vuelidate errors to only show when you blur the input! So you don't see "this field is required" before you've ever typed anything.

src/pages/Settings/SettingsPage.vue
```diff
@@ -182,7 +182,11 @@ export default {
 	},
 	apollo: {
 		query: pageQuery,
-		prefetch: true,
+		preFetch(config, client) {
```
eddieferrer, line 185: I don't think this is necessary and can be `preFetch: true,`

src/components/WwwFrame/Menus/TheSettingsTertiaryMenu.vue
```diff
@@ -76,7 +76,11 @@ export default {
 	},
 	apollo: {
 		query: pageQuery,
-		prefetch: true,
+		preFetch(config, client) {
```
eddieferrer, line 79: I don't think this is necessary and can be `preFetch: true,`

eddieferrer (commented):
I think we should add an eslint exception for this line: 
`const mfaOffConfirm = window.confirm('Are you sure you want to turn off 2-step verification?');` To avoid the lint warning every time.

.babelrc
emuvente: This should be fine. I think it will actually reduce the bundle size a little bit, if I'm correctly understanding what it does.

eddieferrer: Ive never run into this. Looks ok 👍

src/pages/Settings/SettingsPage.vue
ryan-ludwig: When I do `preFetch: true` the content isn't there on the initial render. A few seconds later it jumps in causing a page reflow. Is there something else I should be doing to prevent that reflow?

ryan-ludwig: I stand corrected. Seems fine now. Not sure what I was seeing before.

emuvente (approved):
👍

eddieferrer (approved):

PR #2430, opened 2020-12-15
SUBS-595 Donation Item

eddieferrer:
* Added contentful driven donation item text with variable that gets replaced on the UI


mcstover (approved):
Please verify in dev that that we fallback to the hard-coded versions if this entry is missing or blank.

eddieferrer:
> Please verify in dev that that we fallback to the hard-coded versions if this entry is missing or blank.

Verified in dev that this is the case

mcstover:
👍

PR #2436, opened 2020-12-15
Grow 301

BoulderBrains:
- I've added a new preventClose option on KvLightbox. Storybook addition, incoming.
- For the small screen version we going with something closer to the default KvLightbox styles vs the custom mobile view shown in the Figma docs. 
- Only once the user has selected the checkbox indicating they have recorded the recovery code are they able to close the lightbox with a click of the "Done/Continue" button.
- RecoveryCode.vue take one prop which is the user's recovery code.

Desktop:
Screen Shot 2020-12-16 at 2 10 41 PM

Mobile: 
Screen Shot 2020-12-16 at 2 11 06 PM


emuvente (approved):
This looks good! @ryan-ludwig I'm thinking that we can change this as needed to incorporate it into the phone and otp setup flows

src/pages/Settings/RecoveryCodeConfirm.vue
```diff
@@ -0,0 +1,138 @@
...
+		title="2-step verification is on"
+		@lightbox-closed="closeLightbox"
+		prevent-close="true"
+	>
+		<p>
+			Copy this recovery code and keep it somewhere safe. You'll
+			need it if you want to log in without your device.
+		</p>
+		<label>
+			<input
```
emuvente, line 13: Actually, let's make this a div instead of a label/input, since we just want to display the recovery code that we already have.

BoulderBrains: updated in latest commit

emuvente (approved):

ryan-ludwig (approved):

src/pages/Settings/RecoveryCodeConfirm.vue
```diff
@@ -0,0 +1,135 @@
...
+			>
+				I have safely recorded this code
+			</kv-checkbox>
+		</div>
+		<div>
+			<kv-button
+				class="expanded smaller show-for-small-only"
+				:disabled="isDisabled"
+				@click.native.prevent="closeLightbox"
+				aria-label="Close small"
```
ryan-ludwig, line 38: We shouldn't need this aria label

src/pages/Settings/RecoveryCodeConfirm.vue
```diff
@@ -0,0 +1,135 @@
...
+			>
+				Continue
+			</kv-button>
+
+			<kv-button
+				class="expanded smaller show-for-medium"
+				:disabled="isDisabled"
+				@click.native.prevent="closeLightbox"
+				aria-label="Close large"
```
ryan-ludwig, line 47: Same, no need for an aria label because we have readable text inside the button. The label would be useful if you had an icon inside but no text.

BoulderBrains: At some point during development, storybook was throwing an accessibility error asking for these fields. Thanks for noticing and explaining why they are not needed. They've been removed in the latest commit.

PR #2442, opened 2020-12-17
GROW-303 full-screen mobile lightbox

ryan-ludwig:
Refactors KvLightbox to fill the screen on mobile, while keeping the title and controls always visible.

As part of GROW-303, we've changed course a little - instead of doing special pages for mobile, we're making the lightbox work better on phones. 

I'll do some more testing on my vm tomorrow morning since KvLightbox is used in many places, but initial tests look good.

Component keeps the same props/events/slots as before.

image
image

### DIFF
--- a/src/components/Kv/KvLightbox.vue
+++ b/src/components/Kv/KvLightbox.vue
@@ -1,58 +1,57 @@
 <template>
 	<transition name="kvfade">
-		<div
-			v-show="isShown"
-			class="kv-lightbox-wrap"
-			:class="{'inverted': inverted}"
-			ref="kvlightbox"
-			@keyup.esc="closeLightbox"
-			@click.stop.prevent="closeLightbox"
-			role="dialog"
-			:aria-labelledby="title ? 'lightbox-title' : null"
-		>
-			<focus-lock :disabled="!isShown">
+		<focus-lock :disabled="!isShown">
+			<div
+				v-show="isShown"
+				class="kv-lightbox"
+				:class="{
+					'kv-lightbox--inverted': inverted,
+					'kv-lightbox--no-padding-top': noPaddingTop,
+					'kv-lightbox--no-padding-bottom': noPaddingBottom,
+					'kv-lightbox--no-padding-sides': noPaddingSides,
+					'kv-lightbox--full-width': fullWidth,
+				}"
+				ref="kvlightbox"
+				@keyup.esc="closeLightbox"
+				@click.stop.prevent="closeLightbox"
+				role="dialog"
+				:aria-labelledby="title ? 'lightbox-title' : null"
+			>
 				<div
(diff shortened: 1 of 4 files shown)


ryan-ludwig:
For a simpler API, it'd be nice to remove or consolidate the `noPaddingTop, noPaddingBottom, noPaddingSides` props into just `noPadding`

mcstover (approved):
Nice!

And I agree we could consolidate the padding arg, so long as we can override with external styles when necessary...

src/components/Kv/KvLightbox.vue
```diff
@@ -130,12 +129,16 @@ export default {
 	methods: {
 		closeLightbox() {
 			if (!this.preventClose) {
-				this.isShown = false;
+				// scroll any body content back to top
+				this.$refs.kvLightboxBody.scrollTop = 0;
```
emuvente, line 133: I don't think we want this. If someone has scrolled far down the page and then opens a lightbox, we wouldn't want them to be placed back on the top of the page after they close the lightbox.

emuvente: Nevermind, I see that this is the lightbox body, not the page body, sorry for my mix up!

emuvente (approved):
nice! 👍

BoulderBrains (approved):
👍  nice solution for grow-303

src/components/Kv/KvLightbox.vue
ryan-ludwig: This will only scroll the contents inside the lightbox back up. Their page scroll position won't change.

ryan-ludwig: I'll fix that comment

mcstover:
👍

eddieferrer (approved):

eddieferrer:
I was doing some testing and it seems I am unable to close the lightbox by hitting the escape key in some situations.

Steps to reproduce:
1 - Open up light box
2 - Click somewhere to lose keyboard focus (for example next to a button in the lightbox_controls) so that there is no keyboard focus
3 - Hit Esc. Nothing happens

ryan-ludwig:
Thanks for testing @eddieferrer . Latest commit has a fix for the escape key. Any ideas why the e2e autolending tests are failing? Testing manually I was able to see the dialogs.

eddieferrer:
@ryan-ludwig The tests are failing because the wrapper element on the lightbox has a height of 0 pixels. This is the element that gets the class that cypress is targeting. 

Screen Shot 2020-12-18 at 8 59 47 AM

ryan-ludwig:
Thanks @eddieferrer, I added a [data-test=lightbox] attribute to kv-lightbox for easier selector grabbing.

PR #2444, opened 2020-12-17
I've added an example of the preventClose lightbox to storybook.

BoulderBrains:
During my work on grow-301 I added the preventClose prop to KvLightbox.vue. This gets an example of the prevent close lightbox into storybook. 

The preventClose lightbox: 
- Hides the x in the upper right corner of the lightbox
- Disables closing the lightbox on a click of the opaque background

In this version of the lightbox we're only giving the user 1 method of closure.

For grow-301 this closure method is disabled until a user clicks a KvCheckbox, at which point, the close button become enabled and user is able to close the lightbox.

### DIFF
--- a/.storybook/stories/KvLightbox.stories.js
+++ b/.storybook/stories/KvLightbox.stories.js
@@ -195,8 +195,53 @@ export const FullWidth = () => ({
 });
 
 
-export const NoPadding = () => ({
+export const preventClose = () => ({
 	components: { KvLightbox, KvButton },
+	props: {
+		preventClose: {
+			default: boolean('preventClose', true)
+		},
+	},
+	data: () => ({
+		lightboxVisible: false,
+	}),
+	methods: {
+		showLightbox() {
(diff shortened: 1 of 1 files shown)


mcstover:
Is it possible to also ensure the preventClose option disables the `esc` key too?

emuvente:
> Is it possible to also ensure the preventClose option disables the `esc` key too?

It does. The escape key closes the lightbox because of this line `@keyup.esc="closeLightbox"` and `preventClose` is checked from within that `closeLightbox` method.

ryan-ludwig:
Oh shoot, we doubled up work on this a little. I also have a story for preventClose in [this PR](https://github.com/kiva/ui/pull/2442/files#diff-3fbf144af2bc733a346892b2cc5bd0b4d7b5da81a6d304d666b20cd43e41da58R77). Since the other PR uses the newer Storybook "args" syntax, I'd recommend we use that one. Thanks for making a story though and sorry for the duplication!

PR #2447, opened 2020-12-18
GROW-303 refinements

ryan-ludwig:
Show PhoneAuthentication as part of a route, use some naming conventions from GROW-302 PR. 

@emuvente  Will likely have some minor merge conflicts after GROW-302 is merged. I'm happy to handle those.


emuvente (approved):
👍  nice! 

PhoneAuthentication.vue still needs the recovery code step, but that doesn't have to be in this PR.

PR #2454, opened 2020-12-23
Corporate campaign header

ryan-ludwig:
ML2-136, ML2-137
I think there may be some more styling to do on mobile, but wanted to get this pushed up before the break. Open to taking some new directions with this also.

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -28,7 +28,7 @@
 					<div class="flexible-center-area"></div>
 					<router-link
 						v-show="showBasket"
-						to="/basket"
+						:to="addHashToRoute('show-basket')"
 						class="header-button show-for-large"
 						v-kv-track-event="['TopNav','click-Basket']"
 					>
@@ -53,7 +53,7 @@
 					<router-link
 						v-show="!isVisitor"
 						:id="myKivaMenuId"
-						to="/portfolio"
+						:to="addHashToRoute('show-portfolio')"
 						class="header-button my-kiva"
 						v-kv-track-event="['TopNav','click-Portfolio']"
 					>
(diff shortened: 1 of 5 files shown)


ryan-ludwig:
TODO: Sticky that nav

mcstover (approved):
Def the quickest route tp our goal!

src/components/WwwFrame/TheHeader.vue
```diff
@@ -9,6 +16,57 @@
...
+						<kiva-logo class="icon" />
+						<span class="show-for-sr">Kiva</span>
+						<div class="header-corporate-logo" v-if="!!this.$slots.corporateLogo">
+							<slot name="corporateLogo"></slot>
+						</div>
+					</div>
+					<div class="flexible-center-area"></div>
+					<router-link
+						v-show="showBasket"
+						to="/basket"
```
mcstover, line 31: This will need to emit an event as we cannot navigate away from this page.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -9,6 +16,57 @@
...
+							['TopNav','click-Sign-in'],
+							['TopNav','EXP-GROW-282-Oct2020',redirectToLoginExperimentVersion]
+						]"
+					>
+						<span>Sign in</span>
+					</router-link>
+					<router-link
+						v-show="!isVisitor"
+						:id="myKivaMenuId"
+						to="/portfolio"
```
mcstover, line 56: Same here, no navigating away from this page. If the hash thing works we can do similar here. if not we can just prevent navigation.

mcstover:
👍  We can iterate will small fixes later if you want to merge this soon. I did run this against other UI pages and didn't see any immediate issues with the standard header.

PR #2471, opened 2021-01-05
VUE-430 Cypress coverage

emuvente:


test/mergeCoverage.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Create combined coverage report
+console.log('Creating combined coverage report...');
+const context = istanbulReport.createContext({
+	coverageMap: map,
+	dir: 'test/coverage',
+});
+istanbulReports.create('lcov', {}).execute(context);
+
+console.log('Coverage report created at test/coverage/lcov.info');
```
emuvente, line 52: This script turned out to be necessary because jest and cypress use slightly different versions of Istanbul that generate incompatible reports, preventing `nyc merge` from working.

This is loosely based on https://github.com/facebook/jest/blob/master/scripts/mapCoverage.js.

The empty coverage file handling is a workaround for this issue https://github.com/istanbuljs/istanbuljs/issues/531.

emuvente: In the future this can also be used for merging together coverage reports from separate packages in a monorepo.

mcstover (approved):
Rad!

eddieferrer (approved):

PR #2475, opened 2021-01-06
SUBS-609 Monthly Good Login Experiment

eddieferrer:
* On entering `/monthlygood/setup` check for experiment value. 
* If value is control, redirect to login page. 
* If value is shown, display setup page, replacing braintree drop in with "Continue" button
* When clicking continue button, save all form values as query params, redirect to login page

### DIFF
--- a/src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
+++ b/src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
@@ -213,7 +213,7 @@
 									<em>* {{ isOnetime ? 'This contribution' : 'Enrolling in Monthly Good' }} will also disable your current auto lending settings.</em>
 								</p>
 
-								<div class="payment-dropin-wrapper">
+								<div class="payment-dropin-wrapper" v-if="hasActiveLogin">
 									<div class="payment-dropin-invalid-cover" v-if="$v.$invalid"></div>
 									<monthly-good-drop-in-payment-wrapper
 										:amount="totalCombinedDeposit"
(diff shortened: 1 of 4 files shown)


src/util/authenticationGuard.js
```diff
@@ -5,14 +5,16 @@ import authenticationQuery from '@/graphql/query/authenticationQuery.graphql';
 const isServer = typeof window === 'undefined';
 
 // Return true if user logged in recently enough and false otherwise
-const checkLastLoginTime = (data, durationKey, defaultDuration) => {
-	const lastLogin = _get(data, 'my.lastLoginTimestamp', 0);
-	const duration = 1000 * (parseInt(_get(data, `general.${durationKey}.value`), 10) || defaultDuration);
+export function checkLastLoginTime(data, durationKey, defaultDuration) {
```
eddieferrer, line 8: Export this function so we can import it into monthly good setup page and also refactor `_get` from it.

mcstover (approved):
Right on!

src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -393,27 +426,107 @@ export default {
...
+			logReadQueryError(e, 'MonthlyGoodSetupPage pageQuery');
+		}
+
 		// Sanitize and set initial form values.
+		// Initial group from prop
 		if (this.lendingCategories.find(category => category.value === this.category)) {
 			this.selectedGroup = this.category;
 		}
 
-		if (!Number.isNaN(Number(this.amount))) {
```
emuvente, line 413: Removing the `Number()` call means that things like `undefined` or `'blabla'` will be allowed, since those are not `NaN`. `isNaN` returns true only for values that are `NaN` and are type `Number` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#examples)

eddieferrer: I believe this call is redundant. The prop is coming from the route, and in the route definition:
`amount: Number(route.query.amount)`

src/router/routes.js:105

emuvente: Aha, yeah I think that should take care of that!

emuvente (approved):

PR #2477, opened 2021-01-07
Grow 376

BoulderBrains:
Added loading placeholders for a MFA on/off <span> in the TwoStepVerification kvSettingsCard on the /settings/security page.

Screen Shot 2021-01-07 at 10 23 21 AM

and for all 3 of the kv-settings-cards on the /settings/security/mfa page:
Screen Shot 2021-01-07 at 10 28 23 AM

### DIFF
--- a/src/components/Settings/TwoStepVerification.vue
+++ b/src/components/Settings/TwoStepVerification.vue
@@ -1,12 +1,13 @@
 <template>
-	<div>
+	<div class="two-step-card-wrapper">
 		<kv-loading-placeholder
-			class="loading"
+			class="two-step-card-wrapper--loading"
 			v-if="isLoading"
 		/>
 		<kv-settings-card
 			v-if="!isLoading"
 			title="2-Step verification"
+			class="two-step-card-wrapper__two-step-card"
 		>
 			<template v-slot:icon>
 				<!-- TODO: THIS ICON IS A PLACEHOLDER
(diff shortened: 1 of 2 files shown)


src/components/Settings/TwoStepVerification.vue
```diff
@@ -1,33 +1,43 @@
...
+				<kv-icon
+					class="icon"
+					title="Auto-lending On"
+					name="auto-icon-on"
+				/>
+			</template>
+			<template v-slot:content>
+				<h3 class="strong">
+					Status:
+					<span :class="`two-step-card-wrapper__two-step-card--mfa-${MFAStatus}`">{{ MFAStatus }}</span>
```
emuvente, line 24: I think it would be better to have the loading placeholder be just for this MFA status span instead of the entire settings card, that way the user won't have to wait to be able to click the "Manage 2-step verification" button.

src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -100,7 +108,7 @@
 							<p>{{ cardSubhead }}</p>
```
emuvente, line 108: I think it would be good to have a loading placeholder for this sub-section as well, since `cardSubhead` might change after the loading completes

emuvente (commented):
Nice!

src/pages/Settings/TwoStepVerificationPage.vue
BoulderBrains: Screen Shot 2021-01-07 at 10 28 23 AM
3rd loading placeholder added, thanks for this suggestion.

src/components/Settings/TwoStepVerification.vue
BoulderBrains: Moved the loading placeholder to work with on/off span instead of entire card.
Screen Shot 2021-01-07 at 10 23 21 AM

emuvente (approved):
👍  Looks great!

ryan-ludwig:
👍 Nice!

PR #2480, opened 2021-01-07
Ensure we can prevent redirects from clicking the loan card image or …

mcstover:
…borrower name. Also bubble up events we can act on to provide alternate interactions.

### DIFF
--- a/src/components/LoanCards/LoanCardImage.vue
+++ b/src/components/LoanCards/LoanCardImage.vue
@@ -90,14 +90,15 @@ export default {
 	methods: {
 		handleImageClick(event) {
 			this.$emit('image-click');
-			if (this.disableLink) {
-				event.preventDefault();
-				return;
-			}
+
 			this.$emit('track-loan-card-interaction', {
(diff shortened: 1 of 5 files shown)


mcstover:
I verified that the homepage usage of this loan card is fine and the lbc usage of the name is good too.

FYI @emuvente @eddieferrer @BoulderBrains regarding some of the new props used on elements of this particular loan card. Props would need to be passed to other loan cards to use these features.

src/components/LoanCards/LoanCardImage.vue
```diff
@@ -90,14 +90,15 @@ export default {
 	methods: {
 		handleImageClick(event) {
 			this.$emit('image-click');
-			if (this.disableLink) {
-				event.preventDefault();
-				return;
-			}
+
 			this.$emit('track-loan-card-interaction', {
```
mcstover, line 94: This was previously not getting fired with the disableLink case. It never affected anything in prod.

ryan-ludwig (approved):

BoulderBrains (approved):
👀

PR #2481, opened 2021-01-07
VUE-467 KvLightBox Focus

eddieferrer:
* Fixes issue with mg settings lightbox not firing watcher
* Fixes issue with all lightboxes autofocusing close button on load

VUE-467

The focus issue was only happening on the site, not storybook. Verified that the lightbox still pops up and traps focus, but focus isn't initialized.

### DIFF
--- a/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
+++ b/src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
@@ -146,7 +146,6 @@
 					<subscriptions-monthly-good-cancellation-flow
 						:show-cancel-lightbox="showCancelLightbox"
 						:subscription-id="subscriptionId"
-						v-if="showCancelLightbox"
 						@confirm-cancel="$emit('cancel-subscription'); showCancelLightbox = false;"
 						@abort-cancel="showCancelLightbox = false"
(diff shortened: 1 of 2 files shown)


src/pages/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -146,7 +146,6 @@
 					<subscriptions-monthly-good-cancellation-flow
 						:show-cancel-lightbox="showCancelLightbox"
 						:subscription-id="subscriptionId"
-						v-if="showCancelLightbox"
```
eddieferrer, line 149: This was causing a weird issue where the lightbox 'visible' watcher was not firing. I suppose this v-if takes precedence and the lightbox child component gets removed from the dom before the 'visible' prop is changed

ryan-ludwig:
Added a few comments to the VUE-467 ticket. There are 2 issues, the first is that it is correct for focus to go to the first focusable element in the dialog (but maybe we can do something about styling). The other is that when the dialog is closed, focus should return to the previously focused element. By blurring that is lost.

eddieferrer:
Superseded by https://github.com/kiva/ui/pull/2485

PR #2483, opened 2021-01-07
added .native.prevent to the click handler to open FAQ lightboxes

BoulderBrains:
added .native.prevent in order to get the mfa FAQ lightboxes working


emuvente (approved):
👍

PR #2485, opened 2021-01-08
VUE-467 lightbox focus

ryan-ludwig:


src/components/Kv/KvLightbox.vue
```diff
@@ -1,25 +1,26 @@
...
+				'kv-lightbox--no-padding-top': noPaddingTop,
+				'kv-lightbox--no-padding-bottom': noPaddingBottom,
+				'kv-lightbox--no-padding-sides': noPaddingSides,
+				'kv-lightbox--full-width': fullWidth,
+			}"
+			data-test="kv-lightbox"
+			@click.stop.prevent="closeLightbox"
+			v-show="isShown"
+		>
+			<focus-lock v-if="isShown" :return-focus="true">
```
ryan-ludwig, line 16: the `return-focus` prop fixes a bug we have in prod where the focus doesn't return to the original button when the dialog closes

ryan-ludwig: Also had to change from :disabled to v-if in order to get the return focus piece working. (The element needs to leave the DOM for it to work)

ryan-ludwig:
Verified using voiceover that our focus moves around correctly, and it announces as a dialog element correctly.

ryan-ludwig:
https://user-images.githubusercontent.com/187937/103963190-843dcf00-510d-11eb-9b9e-be701d153e46.mp4

eddieferrer (approved):
Looks great! thanks @ryan-ludwig for taking this over

src/components/Kv/KvLightbox.vue
```diff
@@ -180,6 +183,10 @@ export default {
 		background: $white;
 		position: relative;
 
+		&:focus {
+			outline: 0; // safe(?), since JavaScript is the only way to focus on this el.
```
mcstover, line 187: 👍

eddieferrer:
@ryan-ludwig fixed tests. It seems that our tests where hitting the save button and then making sure the radio was selected, but I believe this was a bug, as hitting save closes the modal. I don't know how it was working before, but the order of the assertions is as intended now.

ryan-ludwig:
Nice, thanks for digging into that @eddieferrer

PR #2488, opened 2021-01-08
GROW-382 use mfaMethod.id instead of mfaMethod.authId

emuvente:
I haven't actually tested this on my vm, but I came across this as the possible problem after I noticed in the network inspector that the removeOneMfaMethod query didn't have the `id` variable. I don't see `authId` used anywhere else, so I think this didn't get updated when `formatMfaMethods()` changed.

### DIFF
--- a/src/pages/Settings/TwoStepVerificationPage.vue
+++ b/src/pages/Settings/TwoStepVerificationPage.vue
@@ -268,7 +268,7 @@ export default {
 						mutation: removeOneMfaMethod,
 						variables: {
 							mfa_token: token,
-							id: mfaMethod.authId
+							id: mfaMethod.id
 						}
 					})
 						.then(() => {


src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -268,7 +268,7 @@ export default {
 						mutation: removeOneMfaMethod,
 						variables: {
 							mfa_token: token,
-							id: mfaMethod.authId
+							id: mfaMethod.id
```
BoulderBrains, line 271: 💯

BoulderBrains (approved):

BoulderBrains:
Good catch @emuvente I'm gonna go forward and merge this small change into master. 👍

PR #2489, opened 2021-01-11
Initial filter implementation test including the gender filter and a …

mcstover:
…preselected set of sectors.

### DIFF
--- a/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
+++ b/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
@@ -73,18 +73,18 @@ export default {
 		showFilters() {
 			this.filtersVisible = true;
 		},
-		updateFilter(filter, value) {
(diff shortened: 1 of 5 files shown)


mcstover:
@ryan-ludwig I've added a query param to show and hide the filters as they progress. They'll be hidden by default but easily activated.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -0,0 +1,199 @@
...
+			filterKeys.forEach(key => {
+				this.modifiedFilters[key] = payload[key];
+			});
+		},
+		applyFilters() {
+			console.log(this.modifiedFilters);
+			this.$emit('updated-filters', this.modifiedFilters);
+			this.filtersVisible = false;
+		},
+		copyFilters(initialFilters) {
```
mcstover, line 101: We'll prob move this to a util file later. It makes a plain object copy of the initial reactive filters.

PR #2491, opened 2021-01-11
Lightbox fixes

ryan-ludwig:
Removes "fullWidth" prop, since the only place it was used was on the interstitial, and it wasn't really helping anything there.
Fixes max-width problem that came in during re-ordering of nodes last week.


mcstover (approved):
Please have a look at that interstitial across viewport sizes

PR #2493, opened 2021-01-11
removing settings card's padding on small screens

BoulderBrains:
Removed the padding from the settings-cards on small screens for /settings/security & setting/security/mfa. 
This change makes these two pages look visually consistent with the /settings/autolend page when on a small screen.

Cards now have no padding while on small screen sizes:
Screen Shot 2021-01-11 at 2 00 31 PM


emuvente (approved):
Nice! 👍

PR #2494, opened 2021-01-12
SUBS-616 BT DropIn Fatal Error

eddieferrer:
* Display an error in tip message when BT Drop In has a fatal error


mcstover (approved):
A good start and we should get that Sentry error too...

PR #2495, opened 2021-01-12
More lightbox fixes

ryan-ludwig:


src/components/Kv/KvLightbox.vue
```diff
@@ -173,13 +175,17 @@ export default {
 	bottom: 0;
 	z-index: 1500;
 	background: rgba(72, 72, 72, 0.6);
+	font-size: 1rem;
```
ryan-ludwig, line 178: This bug showed up in the checkout page. Prevents the following:
```
<div style="font-size: .5rem;">
<p> Lightbox text is also small</p>
<kv-lightbox>
My text is also small
</kv-lightbox>
</div>
```

mcstover (approved):

PR #2501, opened 2021-01-12
VUE-461 Removed Items from thanks page for print

eddieferrer:
* Fixed padding and removed promo banner and monthly good upsell from thanks page


emuvente (approved):
Thanks! :+1:

PR #2506, opened 2021-01-13
Lightbox fixes

ryan-ludwig:
Cleanup more unused full-widths.
Fixes a bug where the watcher doesn't always fire on the first initial load, so focus lock, scroll lock, etc. doesn't get set.


src/components/Kv/KvLightbox.vue
```diff
@@ -117,6 +112,18 @@ export default {
 		// Create and set our internal visibility property
 		visible() {
 			this.isShown = this.visible;
+			this.init();
+		}
+	},
+	mounted() {
+		this.isShown = this.visible;
+		this.init();
```
ryan-ludwig, line 120: I couldn't think of a good method name. Init function sets scroll lock, focus, focus lock, esc key listeners, but also tears them down.

eddieferrer (approved):

PR #2512, opened 2021-01-13
grow-388 adding beforeRouteUpdate() to allow MFA methods list to upda…

BoulderBrains:
…te without page refresh

Grow-388: https://kiva.atlassian.net/browse/GROW-388

Adding in a navigation guard which detects the change and updates the page with the changes without the user having to refresh. 

Vue navigation guards:
https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

Coincides with these /kiva/main changes: 
https://github.com/kiva/kiva/pull/9559

### DIFF
--- a/src/pages/Settings/TwoStepVerificationPage.vue
+++ b/src/pages/Settings/TwoStepVerificationPage.vue
@@ -219,6 +219,12 @@ export default {
 			return this.mfaMethods.length > 0;
 		},
 	},
+	beforeRouteUpdate(to, from, next) {
+		if (to.path === '/settings/security/mfa') {
+			this.gatherMfaEnrollments();
+		}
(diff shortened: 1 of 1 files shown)


mcstover (approved):
Code wise this is correct. Have you been able to verify that the action does in fact trigger this hook? I think the path segments will need to change to trigger this.

BoulderBrains:
@mcstover I tested this locally and it was working, adding MFA methods correctly updated the user's list of MFA methods without refreshing the page.

mcstover (approved):

BoulderBrains:
Verified and working ✔️

emuvente (approved):
👍

PR #2520, opened 2021-01-14
Ensure lightbox is full width on mobile.

ryan-ludwig:


src/components/Kv/KvLightbox.vue
```diff
@@ -16,43 +16,45 @@
...
-							<kv-icon
-								class="kv-lightbox__close-btn-icon"
-								name="small-x"
-								:from-sprite="true"
-							/>
-						</button>
-					</div>
-					<div class="kv-lightbox__body"
-						ref="kvLightboxBody"
+				<div class="kv-lightbox__container">
```
ryan-ludwig, line 19: This is the main change. Adds a wrapper div to hook into because <focus-lock> was adding a div outside of normal control.

mcstover (approved):

PR #2521, opened 2021-01-14
grow-387 adjustments to KvVerficationCodeInput for small screens.

BoulderBrains:
On small screen, the Auth App Code, and the 6 digit input didn't fit on our smallest support screen size 320px
TheIssue

Just for reference, this is the same lightbox on large screens (looking good)
Lg screen before my changes

I made adjustments to:
AppAuthentication.vue - Updated the font size of the returned code
-- The one thing I'm worried about with my fix, is this font-size being too small, setting it to 12px

KvVerificationCodeInput.vue - added a few style adjustments that I was hoping to get your eyes on @ryan-ludwig, I just made a couple style adjustments for small screens and reimplement most of your original styles at the medium breakpoint.

**My Fix** 
_small screen_
fixed small screen

_large screen_ 
Lfg screen after my changes

I think this looks good across screen sizes, but my one concern is that the App Auth code that you manually input into your auth app is too small (renders as 12px bolded). What do you two think?

### DIFF
--- a/src/components/Kv/KvVerificationCodeInput.vue
+++ b/src/components/Kv/KvVerificationCodeInput.vue
@@ -89,15 +89,9 @@ export default {
 	&__input {
 		font-size: 1em;
 		font-variant-numeric: tabular-nums;
-		letter-spacing: 0.2em;
-		padding: 0.7em 0 0.7em 0;
-		text-align: center;
-
-		@include breakpoint(medium) {
-			letter-spacing: 0.5em;
-			padding: 1em 0 1em 0;
-			width: calc(var(--kv-verification-code-input-maxlength) * 1.1125em + 1.25em);
-		}
+		letter-spacing: 0.5em;
+		padding: 1em 0 1em 1em;
+		width: calc(var(--kv-verification-code-input-maxlength) * 1.1125em + 1.25em);
 	}
 }
 </style>
--- a/src/components/Kv/KvVerificationCodeInput.vue
+++ b/src/components/Kv/KvVerificationCodeInput.vue
@@ -89,9 +89,15 @@ export default {
 	&__input {
 		font-size: 1em;
 		font-variant-numeric: tabular-nums;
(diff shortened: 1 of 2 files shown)


emuvente (approved):
12px is the minimum I would feel ok putting anything at, so this works 👍

src/components/Kv/KvVerificationCodeInput.vue
```diff
@@ -89,9 +89,15 @@ export default {
 	&__input {
```
ryan-ludwig, line 89: Because this component is set in `ems`, you can resize the whole deally by setting the font size.  That way the padding, letter-spacing etc. don't need to change.

In your AppAuthentication.vue it'd look something like
```
<kv-verification-code-input class="myclass">

.myclass {
  font-size: .75rem;

  @include breakpoint(large) {
    font-size: 1rem;
  }
}
```

BoulderBrains: @ryan In this same PR, I'm making those changes inside AppAuthentication.vue. It's my understanding that the additional changes were needed inside KvVerificationCodeInput.vue because it doesn't fit on small screen. This is a screen shot of how the lightbox loads with just your suggested changes and KvVerificationCodeInput.vue reverted to how it was before. 
Screen Shot 2021-01-15 at 9 38 42 AM
(*The input flows off the screen and the padding is too large to work on small screens)

emuvente (approved):
👍

BoulderBrains:
I connected with Ryan this morning and made some adjustments to these style changes. Set KvVerificationCodeInput.vue back to what it was and am now overriding the styles that I need to in order to make this look good on small screens.
Small:
Screen Shot 2021-01-15 at 12 37 37 PM

Larger:
Screen Shot 2021-01-15 at 12 37 45 PM

ryan-ludwig (approved):

PR #2531, opened 2021-01-16
Fix issue where lightbox didn't contain wide content correctly

ryan-ludwig:


mcstover (approved):
I think this will do what we need!

PR #2539, opened 2021-01-20
ML2-122 loan details

ryan-ludwig:
Not perfect but gets us something to discuss.


mcstover (approved):
Pretty concise for what will now be displayed!

PR #2542, opened 2021-01-21
ML2-179 Close basket if all loans are removed

ryan-ludwig:


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -764,6 +764,11 @@ export default {
 			// eslint-disable-next-line no-underscore-dangle
 			this.kivaCards = basketItems.filter(item => item.__typename === 'KivaCard');
 
+			// If the checkout lightbox is currently open and user removes all loans, close it.
+			if (!this.basketLoans.length) {
+				this.checkoutVisible = false;
+			}
```
mcstover, line 770: What if we create an `else` statement in the `showCheckout` method below...

PR #2547, opened 2021-01-22
GROW-419 GROW-420 MG Selector Interactions

eddieferrer:
* Close MG Selector pop ups on pressing escape
* CLose MG pop ups when clicking outside


mcstover (approved):
Cool

PR #2552, opened 2021-01-25
Update BorrowHowLong.vue

coderintherye:
Move to /my/borrower/application instead of /my/borrower/minimum-eligiblity


coderintherye:
@mcstover this puts the link to the new URL. It doesn't need a hotfix, but can we put this in for next release?

mcstover (approved):

mcstover:
@coderintherye this will go out on Monday if that's ok with you.

PR #2553, opened 2021-01-25
Remove home_lenderpreferences exp which is not used anymore

eddieferrer:
Remove stale commented out experiment


mcstover (approved):
Should be good, we can peak back if we need this later for v2

PR #2559, opened 2021-01-26
ML2-167 improve checkout styles

ryan-ludwig:
Broadly, reworks checkout CSS to use standard Foundation grid with less overrides (mainly by adding .columns class and removing padding/margin css). Aimed to prevent overrides on overrides for the in-context checkout and make working in checkout more predictable. Visually the existing checkout should not have changed (within a few pixels).

* Replace some span > divs
* Fix scoping, removed forced-width class.

* Fix issue where lightbox contents with a foundation grid were indented too much, add a story for that.
* Min-width for the in-context checkout

### DIFF
--- a/src/components/Checkout/KivaCardRedemption.vue
+++ b/src/components/Checkout/KivaCardRedemption.vue
@@ -1,9 +1,9 @@
 <template>
-	<div class="columns">
+	<div>
 		<div
 			class="kiva-card-entry-wrapper row"
 		>
-			<span class="small-3 large-2">
+			<div class="small-3 large-2 columns">
 				<button @click="toggleAccordion">
 					<kv-icon
 						:class="{ flipped: open }"
@@ -12,9 +12,9 @@
 						:from-sprite="true"
 					/>
 				</button>
-			</span>
+			</div>
 			<button
-				class="featured-text accordion-title small-9 large-10"
+				class="featured-text accordion-title small-9 large-10 columns"
 				@click="toggleAccordion"
 			>
 				Have a Kiva Card?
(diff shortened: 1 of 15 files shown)


src/components/Checkout/KivaCardRedemption.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-	<div class="columns">
+	<div>
```
emuvente, line 2: Since there's only one child in this div we can probably remove it and just have the child be the root element of this component

mcstover (approved):
Overall happy to see a more standard approach here. Please cover your bases with some visual verification across checkout and all the containers and their content that this touched.

src/components/Kv/KvLightbox.vue
```diff
@@ -49,7 +49,12 @@
 					<div class="kv-lightbox__body"
 						ref="kvLightboxBody"
 					>
-						<slot>Lightbox body</slot>
+						<!-- gives lightbox content foundation grids the correct margins -->
+						<div class="row small-collapse">
+							<div class="small-12 columns">
```
emuvente, line 54: Since there's only a single column in this row, the column and row [can be combined](https://get.foundation/sites/docs/grid.html#combined-column-row), i.e.
```
<div class="row column collapse">
	<slot>Lightbox body</slot>
</div>
```

emuvente (approved):
I love this refactor! Good job cleaning up the usage of the grid classes. I left a couple comments but overall it's great!

src/components/Kv/KvLightbox.vue
ryan-ludwig: Nice, thanks!

ryan-ludwig: Rats, for some reason the combination of `row column collapse` on the same div don't play well together.

PR #2564, opened 2021-01-27
SUBS-614 Add A Card Interface

eddieferrer:
* Adds ability to add a card in payment settings

~Basically Done. Waiting for Production btTokenKey - See ITS-6279~

~- We have to use a token key to bypass the vault and only show an add card form in UI~
~- Token key is fine to expose publicly: https://developers.braintreepayments.com/guides/authorization/tokenization-key/javascript/v3#obtaining-a-tokenization-key~

Screen Shot 2021-01-27 at 11 29 54 AM

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -28,6 +28,7 @@ module.exports = {
 			apiKey: '82ec72aa3177a6f4fc47b7103e6db786',
 			defaultIndex: 'prod_fundraising_popularity',
 		},
+		btTokenKey: 'production_x6dwd3k4_f5swrg87wym8bx37',
 		auth0: {
 			loginRedirectUrls: {
 				X6gQsD1f3Y4dvCQK8LGxYPc84UZ9Svts: 'https://admin.kiva.org/admin/login?force=1',
--- a/src/components/Payment/BraintreeDropInInterface.vue
+++ b/src/components/Payment/BraintreeDropInInterface.vue
@@ -34,7 +34,11 @@ export default {
 		* */
 		flow: {
 			type: String,
-			default: 'checkout'
+			default: 'checkout',
+			validator(value) {
+				// The value must match one of these strings
+				return ['checkout', 'vault'].indexOf(value) !== -1;
+			}
 		},
 		/**
 		 * Preselect Vaulted Payment Method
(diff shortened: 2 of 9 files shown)


config/index.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
 			apiKey: '82ec72aa3177a6f4fc47b7103e6db786',
 			defaultIndex: 'prod_fundraising_popularity',
 		},
+		btTokenKey: 'sandbox_q7kbtvzy_vynd473f79sfskz3', // !TODO
```
eddieferrer, line 31: Production key goes here

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -34,7 +34,11 @@ export default {
 		* */
 		flow: {
 			type: String,
-			default: 'checkout'
+			default: 'checkout',
+			validator(value) {
```
eddieferrer, line 38: We should probably make use of prop validators like this in more places

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -53,53 +57,75 @@ export default {
...
 			clientToken: null,
-			updatingPaymentWrapper: true,
+			updatingPaymentWrapper: false,
 		};
 	},
 	methods: {
 		setUpdatingPaymentWrapper(state) {
 			this.updatingPaymentWrapper = state;
 		},
-		initializePaymentOptions() {
```
eddieferrer, line 69: Renamed this method to be a little more descriptive of what it is doing.

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -429,6 +455,8 @@ $border-width: 1px;
 
 		// Credit Card icons in form header
 		[data-braintree-id="card-view-icons"] {
+			padding-bottom: 0;
```
eddieferrer, line 458: Makes the braintree loading in drop in look a little better

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -53,53 +57,75 @@ export default {
...
+				return ['token-key', 'client-token'].indexOf(value) !== -1;
+			}
 		},
 	},
 	data() {
 		return {
 			btDropinInstance: () => {},
 			clientToken: null,
-			updatingPaymentWrapper: true,
+			updatingPaymentWrapper: false,
```
mcstover, line 82: Won't this turn the initial loading state off everywhere?

mcstover (approved):

src/components/Payment/BraintreeDropInInterface.vue
eddieferrer: If it actually does need to load -- when (this.auth === 'client-token') and we need a token, then we set loading state before we fetch the token in getDropInToken()

PR #2565, opened 2021-01-27
GROW-401 Reset MFA if there is an inactive recovery code on page load

emuvente:
If a user cancels their App Authentication setup (by closing the lightbox or closing the page) we can end up in a situation where a recovery code was generated but it was not shown to the user. We can identify those situations by the user having a recovery code method that is not active (there is only ever 1 recovery code method, and it only becomes active once the user confirms their first MFA enrollment). We can fix this situation by reseting the user's MFA so that they get a new recovery code on their next setup.

### DIFF
--- a/src/components/Settings/AppAuthentication.vue
+++ b/src/components/Settings/AppAuthentication.vue
@@ -156,6 +156,9 @@ export default {
 	mounted() {
 		this.startEnrollment();
 	},
+	beforeDestroy() {
+		this.lightboxVisible = false;
+	},
 	methods: {
 		afterScan() {
 			this.step = 1;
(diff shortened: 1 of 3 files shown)


src/pages/Settings/TwoStepVerificationPage.vue
```diff
@@ -168,7 +168,10 @@ export default {
 			const mfaOffConfirm = window.confirm('Are you sure you want to turn off 2-step verification?'); // eslint-disable-line no-alert, max-len
 			if (mfaOffConfirm) {
 				// Upon confirm trigger mutation to turn off mfa
-				this.turnOffMfa();
+				this.turnOffMfa().then(() => {
+					// Upon completion return to the security settings page
+					this.$router.push('/settings/security');
```
emuvente, line 173: This is for GROW-426, which I decided to include here because it was quick.

src/components/Settings/AppAuthentication.vue
```diff
@@ -156,6 +156,9 @@ export default {
 	mounted() {
 		this.startEnrollment();
 	},
+	beforeDestroy() {
+		this.lightboxVisible = false;
```
emuvente, line 160: This is necessary because KvLightbox emits a close event when it's destroyed, which triggers the completeSetup method in this component. By setting lightbotVisible to false before this component is destroyed, we prevent it from causing a double-navigation error.

mcstover (approved):

BoulderBrains (approved):

PR #2571, opened 2021-01-28
ML2-196 Add IE notice for managed lending 

ryan-ludwig:


mcstover (approved):
Def a candidate for a global component...

emuvente (commented):
If their browser in unsupported, this message might not be shown. If there's a javascript feature we're using that isn't supported by their version of IE, then the javascript won't parse and this won't get executed.
Alternatively, we could check the user agent string from the request on the server, allowing the message to be rendered on the server, and then it doesn't matter if the javascript works in their browser or not.

ryan-ludwig:
That's a really good point @emuvente

src/main.js
```diff
@@ -65,6 +66,7 @@ export default function createApp({
 	// Provide locale to all components
 	// TODO: use this to set locale in VueI18n
 	Vue.prototype.$locale = locale;
+	Vue.prototype.$browser = browser;
```
ryan-ludwig, line 69: Adds this object globally
Screen Shot 2021-02-01 at 4 26 01 PM

ryan-ludwig:
@emuvente @mcstover This pass uses Bowser to parse the user agent on the server and client.
Verified the banner shows on Windows 7 + IE11 with javascript turned on and off. Does not show on Windows 7 + Edge.

src/main.js
ryan-ludwig: Perhaps a different name than $browser? Any thoughts?

emuvente: I'm ok with `$browser`, but if we were to change it I think `$device` might be a good option as it's an object with all the info about the device being used to view the site, not just the browser

emuvente (approved):
👍  Nice!

BoulderBrains (approved):

mcstover (approved):

PR #2576, opened 2021-01-29
VUE-383 Replace cookieStore with cookie-universal

emuvente:
I haven't fully QA'd this yet, but the unit tests are passing. Please take a look and try it out yourselves!


mcstover:
This is awesome, I'll give it a spin tomorrow sometime!

emuvente:
Adding a note here that we identified that the issue still happens because the Vue prototype can be shared between renders! To make this work properly, we would need to pass the cookies via provide/inject instead of through the Vue prototype (same for $locale).

As an alternative, we want to try https://github.com/kiva/ui/pull/2585 which has minimal changes to our production code and also makes it possible to use the Vue prototype for other things, the downside being that it probably uses more memory than our current deployment.

emuvente:
Closing in favor of #2687

PR #2582, opened 2021-02-01
Fix availableTotal check in OrderTotals. Prevent resetting loan carou…

mcstover:
…sel position when a loan has been added to basket. Exit the promo code check quickly if there isn't one to process. Add fallback handling to fetch a new basket cookie and refresh the page after the thanks lightbox has closed.


mcstover:
Minor fixes to deploy with the release today

ryan-ludwig (approved):

PR #2585, opened 2021-02-01
VUE-383 Run renders in new context

emuvente:
Here's a simpler solution to the problem if we want to explore this route. I didn't get any invariant violations when using locust against this code.

### DIFF
--- a/src/util/apolloPreFetch.js
+++ b/src/util/apolloPreFetch.js
@@ -41,7 +41,7 @@ export function handleApolloErrors(handlers = {}, errors, args) {
 		// Call the error handler with the errors and any additional args passed in from the top function
 		const handlerPromise = handler({ graphQLErrors, ...args });
 		// Throw error for malformed error handler functions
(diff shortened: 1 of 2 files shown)


src/util/apolloPreFetch.js
```diff
@@ -41,7 +41,7 @@ export function handleApolloErrors(handlers = {}, errors, args) {
 		// Call the error handler with the errors and any additional args passed in from the top function
 		const handlerPromise = handler({ graphQLErrors, ...args });
 		// Throw error for malformed error handler functions
-		if (!(handlerPromise instanceof Promise)) {
+		if (typeof handlerPromise.then !== 'function') {
```
emuvente, line 44: It turns out that `Promise` is different in each context, so instanceof will not always return true. Testing that `.then` is a function is also a valid way of identifying promises, so I switched to using that.

emuvente:
Closing in favor of #2687

PR #2587, opened 2021-02-02
grow-465 switched order of auth methods, recommended -> most secure

BoulderBrains:
Flipped auth method order and updated Recommended -> Most Secure

Screen Shot 2021-02-01 at 5 35 39 PM


emuvente (approved):
👍 💨

mcstover (approved):

PR #2591, opened 2021-02-02
ML2-222 Don't send the raw contentful image

ryan-ludwig:
The images uploaded for Visa's campaign were pretty giant for what they were.
Contentful lets you add query params to dynamically set the height/width/format/quality. Saves a lot of bytes.

Original
image

Using picture
image


eddieferrer:
Excellent 👍  We should probably be using this in all cases that we get images from contentful? Maybe a nice wrapper component in the future?

eddieferrer (approved):

mcstover (approved):
Very cool!

ryan-ludwig:
> We should probably be using this in all cases that we get images from contentful? Maybe a nice wrapper component in the future?
@eddieferrer Yeah, I think it's a good idea to do this for all images from contentful, a wrapper could be handy :)

ryan-ludwig:
Created VUE-483

PR #2596, opened 2021-02-03
Ensure loans are attributed to the newly joined or previously joined …

mcstover:
…associated promo team. Add additional authStatus checks to ensure user state is up to date.

### DIFF
--- a/src/components/CorporateCampaign/CampaignJoinTeamForm.vue
+++ b/src/components/CorporateCampaign/CampaignJoinTeamForm.vue
@@ -13,7 +13,8 @@
 					<p>
 						Lending Teams are self-organized groups on Kiva.org where you can
 						connect and rally around shared lending goals. By joining the
-						{{ teamName }}, you can easily see our collective impact across countries and sectors.
+						{{ teamName }} Lending Team, you can easily see our collective impact
+						across countries and sectors.
 					</p>
 					<p v-if="showError" class="error">
 						Oh no! Something went wrong! Please try to join this team later.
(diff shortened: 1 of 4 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -909,16 +926,34 @@ export default {
...
 				query: myTeamsQuery
 			}).then(({ data }) => {
 				this.myTeams = data.my?.lender?.teams?.values ?? [];
+				this.addTeamToLoans();
 			});
 		},
+		addTeamToLoans() {
+			if (this.basketLoans.length && this.teamId) {
+				const loans = [];
+				// TODO Collect these promises and refresh basket once complete
```
mcstover, line 942: will remove TODO later as the code below does just what it says ;)

src/components/CorporateCampaign/CampaignJoinTeamForm.vue
```diff
@@ -203,6 +204,12 @@ export default {
 <style lang="scss" scoped>
 @import 'settings';
 
+.join-team-lightbox {
```
ryan-ludwig, line 207: To avoid v-deeping and overriding the lightbox container, another approach is to put the class on the contents rather than the lightbox el. 

```
<kv-lightbox>
  <div class="join-team-lightbox">lightbox contents...
</kv-lightbox>

.join-team-lightbox { max-width: 30rem; }
```

ryan-ludwig (approved):

src/components/CorporateCampaign/CampaignJoinTeamForm.vue
mcstover: oh good call, I'll try that in a follow up.

PR #2597, opened 2021-02-03
ML2-178 Only show the contents of the lightbox when printing

ryan-ludwig:


ryan-ludwig:
This applies to printing anytime a lightbox is open, not just managed lending.

ryan-ludwig:
Tried a few ways to go about this. This one seems fairly clean, but I'm def open to other approaches. The main problem is that most printing defaults don't print background colors (including white), so the lightbox is hard to keep looking like a lightbox.

emuvente:
If you want to force a background color to print, you can use this mixin that we had in styleguide:
https://github.com/kiva/styleguide/blob/master/source/css/scss/app/print.scss#L3-L13

mcstover (approved):

ryan-ludwig:
@emuvente that's amazing, is there anything box-shadow can't do?

ryan-ludwig:
I think just printing the lightbox contents is a good solution. Took a look at how Jira handles printing with lightboxes and it seems similar to this.

emuvente (approved):
👍

PR #2598, opened 2021-02-03
grow-470 hiding the monthly good selector portion of the monthlygoodh…

BoulderBrains:
…omepage

R2.2 of GROW-470 requests that the "monthly good selector sections" of the monthlygoodhomepage be removed.
- I've added a data point and conditionally hid it for now. 

(In the near future)
We have an experiment that will test showing/hiding this MG selector module (grow-427). Once that ticket is sorted out, the data point, can been hooked up to the experiment setting.

### DIFF
--- a/src/pages/Homepage/MonthlyGoodHomepage.vue
+++ b/src/pages/Homepage/MonthlyGoodHomepage.vue
@@ -58,7 +58,10 @@
 			</div>
 		</section>
 
-		<section class="monthly-good-info section">
+		<section
+			v-if="showMonthlyGoodSection"
+			class="monthly-good-info section"
+		>
 			<div class="row align-center">
 				<div class="small-10 large-12 columns">
 					<h1 class="monthly-good-info__header" v-html="infoHeadline">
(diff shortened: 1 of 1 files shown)


emuvente (approved):
👍

src/pages/Homepage/MonthlyGoodHomepage.vue
```diff
@@ -363,6 +370,7 @@ export default {
 			isSticky: false,
 			initialBottomPosition: 0,
 			mgStickBarOffset: 0,
+			showMonthlyGoodSelectorSection: false,
```
mcstover, line 373: Please put this in a Data object coming out of Contentful (`ui-homepage-monthly-good`. There is a JSON object there you can use which will be available to pass down from Homepage > MG Homepage as a prop.

This way the configuration won't be hard-coded and require a release.

mcstover: please make this configurable externally with the next phase.

mcstover (approved):

PR #2607, opened 2021-02-04
SUBS-629 Icon Button

eddieferrer:
* New button style
* New themeable component for wrapping new button styles with 2 icons
* Made changes to share icon to allow themeing
* Removed unused FB button

-----
Screen Shot 2021-02-04 at 8 34 38 AM

Another new button style. I know I know.... 

I named it Ghost - https://uxplanet.org/ghost-buttons-in-ux-design-4cf3717334f8 but I'm open to suggestions an I know we already have a ghost color 

These will be used for upcoming Thanks Page redesign and Guest Checkout


src/components/Kv/KvIconButton.vue
```diff
@@ -0,0 +1,124 @@
...
+		height: 2rem;
+		width: 2rem;
+		max-width: 2rem;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		margin-right: 1rem;
+	}
+
+	.icon-wrapper ::v-deep div {
```
mcstover, line 75: Is this "deep" div within the KvIcon or a markup element that could change?

ryan-ludwig:
There's some really good stuff here, the piece I'm having trouble with is that we are adding yet another button style to KvButton without having a birds-eye view that I'm aware of for consistency. This new button has **a different font size, color, border-radius, border-thickness, and different padding than all of our others.**

image

On it's face it seems closest to our like "secondary". style (though we really should make secondary be the same height as the others). 

I'd like to see a birds eye view of the button variations before adding a new style. In the immediate term, using "secondary" plus the icon slots seems like it could work, which would allow us to fix all "secondary" buttons to a new classic/ghost etc when that's specced out.

src/assets/icons/inline/share.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 20 20"><defs/><path stroke="#484848" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M15 6.667a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM5 12.5a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM15 18.333a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM7.158 11.258l5.692 3.317M12.842 5.425L7.158 8.742"/></svg>
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 20 20"><defs/><path d="M15 6.667a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM5 12.5a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM15 18.333a2.5 2.5 0 100-5 2.5 2.5 0 000 5zM7.158 11.258l5.692 3.317M12.842 5.425L7.158 8.742"/></svg>
```
ryan-ludwig, line 1: nice

src/components/Checkout/SocialShare.vue
```diff
@@ -649,5 +649,9 @@ $loan-triangle-size: rem-calc(12);
 	height: 1rem;
 	width: 1rem;
 	margin-right: 1rem;
+	color: $umbreon;
```
ryan-ludwig, line 652: How does $umbreon fit in with the rest of our color scheme? If it's something design is planning to use regularly we should add it to StyleguideColors.vue

eddieferrer:
@ryan-ludwig I agree. I did not add a screenshot of the hover state, but the "ghost" button, while similar to secondary (you are correct though, different font, spacing, radius, border etc). Actually has a totally different hover style. So it's like secondary, but on hover its a primary button, and also has a third "active" state

src/components/Kv/KvIconButton.vue
eddieferrer: Its the KvIcon wrapper.

src/components/Checkout/SocialShare.vue
mcstover: @eddieferrer I think we should avoid applying the umbreon color at this point as it looks like it will become basically a global override for everything in the content area of this page.

src/components/Kv/KvIconButton.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+	<kv-button class="icon-btn"
```
ryan-ludwig, line 2: I like the theme idea here, but it doesn't allow us use our other existing button appearances easily without passing most of their styling in via CSS custom properties.

Another way to go about this would be to add 2 more slots to KvButton.vue. Or potentially "icon", and "icon-right" props . That would require rewriting KvButton some to not use the render function. 

I can imagine `<kv-button class="ghost" icon="info" icon-right="chevron">` or something like that.

ryan-ludwig:
Now that I've seen the designs with this in context I'm wondering if this is just not a KvButton.  It's left-aligned, but 100% width, new style, uses 2 icons - one with a background circle. White background when active which won't translate when the background of the page is white...  Maybe instead of forcing KvButton to work, this is just a new button component.

src/components/Checkout/SocialShare.vue
eddieferrer: Done, changed it back to kiva-text-dark

BoulderBrains (approved):

mcstover (approved):
We might want to notate that this is only approved for use on Thanks at this time. Maybe in the Story...

ryan-ludwig (approved):

emuvente (approved):
👍

eddieferrer:
Sorry for the back and forth on this. I'm building out the new Thanks page at the same time and after trying to use this component there, I decided against modifying the share component or the share icon. I found a way to style svg icons with stroke and updated this accordingly. I also added all 3 new icons for the Thanks page here and in the story ^

mcstover (approved):

PR #2616, opened 2021-02-05
Grow 330

BoulderBrains:
Almost there, but I could use some help debugging at this point. 

The majority of this work has been due to that LendByCategoryPage.vue had been coded to only handle 1 recommended row. With the addition of this new experiment row we now have 2 recommended loan rows. 

Currently: 

Logged in: 
- Invariant Violation console error thrown
✔️ - Both the Recommended loan row & Spotlighted by Kiva rows are preset and displaying loans 

Logged out: 
- Invariant Violation console error thrown
- Just the Spotlighted by kiva row is shown on client load, when serve loads, recommended row appears with 1 loan. ????

Invariant Violation (thrown when server loading finishes)
Screen Shot 2021-02-08 at 6 47 11 PM


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -398,30 +445,53 @@ export default {
...
 					const data = this.apollo.readQuery({
 						query: recommendedLoansQuery,
 						variables,
 					});
+					console.log('data after read query', data);
+
+					const allRecLoanChannels = data.ml?.getOrderedChannelsByIds ?? [];
+
+					const allRecLoans = allRecLoanChannels.map(channel => {
+						const channelLoans = channel.loans?.values ?? [];
```
mcstover, line 473: You'll want to create another const here for to extract data from the rec channel stored in categorySetting.

`const channelData = this.categorySetting.filter(cat => id === channel.id);`

Something like this should give you an object of the original channel data including it's name and description. You can then use these in the object you're build below where appropriate.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -398,30 +445,53 @@ export default {
...
+						if (channelLoans.length) {
+							return {
+								values: this.filterFundedLoans(channelLoans),
+								id: channel.id,
+								name: channel.name,
+								description: channel.description,
```
mcstover, line 479: Then you'd use `channelData` instead of `channel` on the name and description fields.

morganlutz:
@BoulderBrains row 103 title/description is available in DEV now

image

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -413,70 +435,104 @@ export default {
...
 		},
-		fetchRecommendedLoans() {
-			// once the initial loan set is filtered and loaded, we query for more loans if there are too few
-			// if there are 6 or fewer recommended loans we'll fetch 20 more
-			const recLoanChannel = _find(this.categorySetting, { __typename: 'RecLoanChannel' });
-			if (recLoanChannel && _get(this, 'recommendedLoans.values.length') < 6) {
+		fetchRecommendedLoans(offset = 0) {
+			const lengthCheck = this.recommendedLoans.map(channel => {
+				const loanCount = channel.values?.length ?? 0;
+				return loanCount > 25;
```
mcstover, line 461: please back this limit down to 6

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -586,12 +658,17 @@ export default {
...
+		console.log('Take and without', _take(_without(this.realCategoryIds, this.customCategoryIds), ssrRowLimiter));
+		// RETURNING: Take and without [60, 103]
 		// Read the SSR ready loan channels from the cache
 		try {
 			const categoryData = this.apollo.readQuery({
 				query: loanChannelQuery,
 				variables: {
-					ids: _take(this.realCategoryIds, ssrRowLimiter),
+					// ERROR IS BEING THROWN HERE or withing realCategoryIds or customCategoryIds
+					ids: _take(_without(this.realCategoryIds, this.customCategoryIds), ssrRowLimiter),
```
BoulderBrains, line 671: To me it looks like the error looks to be thrown here, while we're joining these two arrays

BoulderBrains:
The latest commit fixes the Invariant Violation. I've still got a little clean up and to test that everything is working correctly.

src/pages/Lend/LendByCategoryPage.vue
emuvente: catrgorySetting only has channel ids, so I don't think it can be used like that.

emuvente (approved):
:+1:  This looks good to me! Just remove that one extra _without call

src/pages/Lend/LendByCategoryPage.vue
mcstover: You're right! we worked through that directly and used data from recommended channels object

mcstover (approved):

PR #2622, opened 2021-02-09
Update team attribution for a loan on mounted.

mcstover:


src/components/Checkout/TeamAttribution.vue
```diff
@@ -55,6 +55,12 @@ export default {
 			return _orderBy(this.teams, 'name');
 		}
 	},
+	mounted() {
+		// update initial team id
+		this.$nextTick(() => {
+			this.selectedId = this.teamId || 0;
```
ryan-ludwig, line 61: Curious why teamId needs a delay

ryan-ludwig (approved):

src/components/Checkout/TeamAttribution.vue
mcstover: We try to init the `selectedId` in the data object in the same way though it doesn't get picked up in that cycle. I verified that that teamId is present and set on the loan object and wanted to make sure it got picked up and applied. It may be fine without the nextTick 🤷

mcstover: Oh I think this will be fine without the nextTick will handle in different PR

PR #2632, opened 2021-02-11
Remove promoBaner query from checkout. Previously this was used for H…

mcstover:
…oliday mode which should be migrated to contentful when we need it again.


mcstover:
Now that this query isn't used in the header, there is no query to read from. There were no other usages of this query so I think I could remove the file as well.

emuvente (approved):

eddieferrer (approved):

PR #2634, opened 2021-02-11
Grow 458

BoulderBrains:
https://kiva.atlassian.net/browse/GROW-458
- Created the guest checkout exp in admin: uiexp.guest_checkout
- EXP Fires for non-registrants users when hitting the /checkout page

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -257,6 +257,7 @@ export default {
 			addToBasketRedirectExperimentShown: false,
 			loginButtonExperimentVersion: null,
 			redirectToLoginExperimentVersion: null,
+			guestCheckoutExperimentVersion: null,
 		};
 	},
 	apollo: {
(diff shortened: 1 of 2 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -373,6 +374,21 @@ export default {
 			fragment: experimentVersionFragment,
 		}) || {};
 		this.redirectToLoginExperimentVersion = redirectToLoginExperiment.version;
+
+		// GROW-458 Guest Checkout Experiment
+		const guestCheckoutExperiment = this.apollo.readFragment({
+			id: 'Experiment:guest_checkout',
+			fragment: experimentVersionFragment,
+		}) || {};
+		this.guestCheckoutExperimentVersion = guestCheckoutExperiment.version;
```
emuvente, line 383: This should probably be behind the `!cookieStore.get('kvu')` check as well, so that we aren't showing the experiment to existing users

emuvente (approved):
👍

PR #2635, opened 2021-02-11
SUBS-630 Tracking Events for MG Cancel Flow

eddieferrer:
* Added tracking for starting MG cancel flow
* Added tracking event for cancelling MG cancel flow


mcstover (approved):
Seems fine, maybe check in on if the Labels should maintain the same upper/lower casing as the presented text on the button...

emuvente (approved):

PR #2636, opened 2021-02-12
VUE-484 KvCarousel for homepage loans

ryan-ludwig:
Refactors loan carousel on the homepage to use KvCarousel

Ensures a little more than 1 loan card is always visible on mobile to let the user know they can swipe
Fixes a minor bug where the carousel size would increase when the end of the loans was reached
Prevents page height jumps when loans load in
Gets those buttery swipes


ryan-ludwig:
Tested in FF and Chrome. Unable to test in IE11 yet, but should be able to once this is in dev.

mcstover (approved):
These improvements came out of feedback when using this setup on the MVP pages so it's nice to move them back into the homepage as well!

emuvente (approved):
Awesome!

eddieferrer (approved):
Woohoo no more window math!

PR #2642, opened 2021-02-12
Fix issue where embla was crashing with 'slidesToScroll = visible' if…

ryan-ludwig:
… slides were added dynamically.

### DIFF
--- a/src/components/CorporateCampaign/CampaignLoanRow.vue
+++ b/src/components/CorporateCampaign/CampaignLoanRow.vue
@@ -18,7 +18,6 @@
 			</p>
 		</div>
 		<kv-carousel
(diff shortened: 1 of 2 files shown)


src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -18,7 +18,7 @@
 			</p>
 		</div>
 		<kv-carousel
-			v-if="!loadingLoans && !zeroLoans"
+			v-show="!zeroLoans"
```
ryan-ludwig, line 21: no more tearing down the carousel

mcstover (approved):

PR #2647, opened 2021-02-13
Make the promoOnly query variable dynamic so it's not universally app…

mcstover:
…lied which breaks visits with no promo applied. Update loan query to remove watch query which fixes the same results hangup on the loading state and ensure we can react when the promise completes. watchQuery .subscribe will never call it's next function if the cache results do not change.

### DIFF
--- a/src/components/CorporateCampaign/CampaignLoanRow.vue
+++ b/src/components/CorporateCampaign/CampaignLoanRow.vue
@@ -76,7 +76,6 @@
 
 <script>
 import basicLoanQuery from '@/graphql/query/basicLoanData.graphql';
-import cookieStore from '@/util/cookieStore';
 import KvCarousel from '@/components/Kv/KvCarousel';
 import KvCarouselSlide from '@/components/Kv/KvCarouselSlide';
 import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
@@ -106,6 +105,10 @@ export default {
 		isVisible: {
 			type: Boolean,
 			default: false
+		},
+		promoOnly: {
+			type: Object,
+			default: null
 		},
 		rowNumber: {
 			type: Number,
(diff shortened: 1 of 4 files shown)


src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -153,7 +156,17 @@ export default {
 		loans() {
 			if (this.loans.length) {
 				this.$nextTick(() => {
-					this.$refs.campaignLoanCarousel.reInit();
+					if (this.$refs.campaignLoanCarousel) {
+						// re-init carousel
+						this.$refs.campaignLoanCarousel.reInit();
+						// shake the carousel to re-init controls
```
mcstover, line 162: I admit this is a janky way to handle things but it works well enough for now. With the new loading sequence, it seems that the carousel controls weren't re-initialized properly. Specifically, the "next" arrow remained disabled after re-init. It would not activate on resizing the page but would re-activate if you swiped to scroll one loan over... I verified all state before and after the re-init but it seemed like the internal checks on the button weren't being updated.

src/components/LoanCards/HoverLoanCard/DetailedLoanCard.vue
```diff
@@ -534,6 +534,7 @@ $row-arrow-width: 2.5rem;
 		width: rem-calc(100);
 		height: rem-calc(100);
 		pointer-events: none;
+		z-index: 1;
```
mcstover, line 537: This was showing up in the VM...maybe not an issue once compiled fully. Just a safeguard

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1129,6 +1137,15 @@ export default {
 	}
 }
 
+.loan-details-lightbox {
```
mcstover, line 1140: Button was often not visible over the loan image. We can refine further later.

PR #2648, opened 2021-02-13
Ensure ordered rows were assigned to rowData before the loanChannelQu…

mcstover:
…ery is run. Also ensure recommended rows are filtered out of the loanChannelQuery readQuery operation.

### DIFF
--- a/src/graphql/query/lendByCategory/lendByCategory.graphql
+++ b/src/graphql/query/lendByCategory/lendByCategory.graphql
@@ -16,10 +16,6 @@ query lendByCategory($basketId: String) {
 			key
 			value
 		}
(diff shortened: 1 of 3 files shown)


emuvente (approved):

mcstover:
@emuvente Just a heads up, It seems there's another issue with the experiment assignment order and maybe the inclusion of the X-Experiments header values. If the new exp assignment occurs as part of this page load then setRows won't have the experiment when prepping the ordered channels.

Making this exp globally assigned will help in some cases but not the direct visit scenario. I also started looking controlling the flow of steps that lead up to setRows to try and ensure the exp version is set by then... more to come later.

BoulderBrains (commented):
👀

PR #2649, opened 2021-02-14
Isolate watcher for loanQueryVars and scope certain operations to onl…

mcstover:
…y occur in the loan display type that is visible.

### DIFF
--- a/src/components/CorporateCampaign/CampaignLoanGridDisplay.vue
+++ b/src/components/CorporateCampaign/CampaignLoanGridDisplay.vue
@@ -162,9 +162,6 @@ export default {
 		},
 	},
 	watch: {
-		loanQueryVars(next, prev) {
-			this.loanQueryVarsStack.push(prev);
-		},
 		filters(next, prev) {
 			if (next !== prev) {
 				this.loanQueryFilters = next;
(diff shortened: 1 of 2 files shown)


src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -154,7 +154,7 @@ export default {
 	},
 	watch: {
 		loans() {
-			if (this.loans.length) {
+			if (this.loans.length && this.isVisible) {
```
mcstover, line 157: re-initializing when not in the view created a crash in the carousel code :(

src/components/CorporateCampaign/CampaignLoanGridDisplay.vue
```diff
@@ -204,10 +207,12 @@ export default {
 				this.loans = data.lend?.loans?.values ?? [];
 				this.totalCount = data.lend?.loans?.totalCount ?? 0;
 
-				this.$emit('update-total-count', this.totalCount);
-				this.checkIfPageIsOutOfRange(this.loans.length, this.pageQuery.page);
+				if (this.isVisible) {
+					this.$emit('update-total-count', this.totalCount);
+					this.checkIfPageIsOutOfRange(this.loans.length, this.pageQuery.page);
```
mcstover, line 212: This notice was showing up while on the carousel interface if previous paging has been used in the grid interface.

I think we could eventually sync the paging across both views...

PR #2656, opened 2021-02-16
Feature grow 486 guest acc claim

mlibazisi:
Ticket: Grow-486

### DIFF
--- a/src/pages/LoginAndRegister/GuestAccountClaim.vue
+++ b/src/pages/LoginAndRegister/GuestAccountClaim.vue
@@ -7,15 +7,15 @@
 			<p>
 				To finish creating your account, please enter your first and last name below.
 			</p>
-			<form id="guestAccountClaimForm" action="." @submit.prevent.stop="postGuestAccountClaimForm">
+			<form id="guestAccountClaimForm" action="." @submit.prevent.stop="submitForm">
 				<label class="input-label" for="firstName">
 					First name
 				</label>
-				<input id="firstName" type="text" required @valid="firstName = $event">
+				<input id="firstName" v-model="firstName" type="text" required @valid="firstName = $event">
 				<label class="input-label" for="lastName">
 					Last name
 				</label>
-				<input id="lastName" type="text" required @valid="lastName = $event">
+				<input id="lastName" v-model="lastName" type="text" required @valid="lastName = $event">
 				<KvButton class="claim-button" type="submit">
 					Done
 				</KvButton>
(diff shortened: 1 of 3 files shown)


src/pages/LoginAndRegister/GuestAccountClaim.vue
```diff
@@ -0,0 +1,90 @@
...
+		formData() {
+			return [
+				`firstName=${encodeURIComponent(this.firstName)}`,
+				`lastName=${encodeURIComponent(this.lastName)}`,
+			].join('&');
+		}
+	},
+	methods: {
+		claimGuestAccount() {
+			this.apollo.mutate({
```
emuvente, line 55: We won't have an authentication token at this point, so we won't be able to make this graphql call. What we'll have to do is make a separate page that makes this graphql call and then redirects the user to their final destination, but let's do that in a separate ticket. For this ticket, remove the `this.apollo.mutate` call.

src/pages/LoginAndRegister/GuestAccountClaim.vue
```diff
@@ -0,0 +1,90 @@
...
+				One last thing!
+			</h1>
+			<p>
+				To finish creating your account, please enter your first and last name below.
+			</p>
+			<form id="guestAccountClaimForm" action="." @submit.prevent.stop="claimGuestAccount">
+				<label class="input-label" for="firstName">
+					First name
+				</label>
+				<input id="firstName" v-model="firstName" type="text" required @valid="firstName = $event">
```
emuvente, line 14: The `@valid="firstName = $event"` isn't needed because the name is already being updated by the `v-model` propertry.

emuvente (commented):
Thanks for pulling this together!

I think you should add some validation to the form:

1. Add a data property called `showValidationError` that is initially `false`
2. Add an error message to the page just before the submit button that is only visible when `showValidationError` is true:
```
<ul v-show="showValidationError" class="validation-errors">
	<li>
		You must provide your first and last name
	</li>
</ul>
```
3. Create a computed property `isValid` that only returns true when both names have been provided:
```
isValid() {
	return this.firstName.trim() && this.lastName.trim();
}
```
4. Check isValid when submitting the form:
```
claimGuestAccount() {
	if (!this.isValid) {
		this.showValidationError = true;
	} else {
		window.location = ...
```

src/router/routes.js
```diff
@@ -158,6 +158,13 @@ export default [
 			authenticationRequired: true,
 		}
 	},
+	{
+		path: '/register/guest',
+		component: () => import('@/pages/LoginAndRegister/GuestAccountClaim'),
+		meta: {
+			activeLoginRequired: true,
```
emuvente, line 165: The user won't be logged in at this point, so we shouldn't require an active login

emuvente (approved):
One last comment, but otherwise it looks great! 👍

PR #2665, opened 2021-02-17
VUE-326 Refactor donate from macro site page

eddieferrer:
* Get images from contentful (VUE-326)
* Add contentful util helper which converts contentful responsive image into kv-responsive-image array
* Remove _get from page
* Remove usage of 'content' from contentful use, contents instead.

This is both the work for VUE-326 and VUE-463 (partial). And necessary so we can remove the content group 'content' field

TODO: 
There are corresponding changes in contentful, I've already made them in the dev environment, related to the support-kiva page nested content. 

Once this gets to prod, we can delete the 'support-kiva' page from prod, and copy the updated one from dev.

### DIFF
--- a/src/pages/Donate/DonateFromMacro.vue
+++ b/src/pages/Donate/DonateFromMacro.vue
@@ -5,7 +5,7 @@
 		:footer-theme="footerTheme"
 	>
 		<donate-from-macro-hero
-			:data="donateFromMacroContent"
+			:data="heroContentGroup"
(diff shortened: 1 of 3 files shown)


mcstover (approved):
May want to test some of these computed fields to see what happens if there is no content for that field. Primary want to make sure it doesn't break the server render of the page due to an error from v-html trying to render 'undefined'.

PR #2674, opened 2021-02-18
Grow 435

BoulderBrains:
This code has been cleaned up and is now ready for review. 

The one issue I had:
inside of stepsText() when I try to use documentToHtmlString for the Headline and Subheadline it removes the values pulled from contentful and returns and empty string. It seems to be working fine without them in there and one of the steps has a break in the middle of the subHeading. You think I'm safe to just remove the mention of documentToHtmlString?

* I can also update the location of the src/components/IWD/2021/HowItWorks.vue file if we want to house these components elsewhere?

Desktop: 
Screen Shot 2021-02-18 at 7 01 19 PM

Mobile:
Screen Shot 2021-02-18 at 7 01 29 PM

### DIFF
--- a/src/components/IWD/2021/HowItWorks.vue
+++ b/src/components/IWD/2021/HowItWorks.vue
@@ -1,0 +1,178 @@
+<template>
+	<section class="how-it-works section text-center">
+		<div class="row">
+			<p
+				v-html="howItWorksHeadline"
+				class="how-it-works__header large-text small-12 columns"
+			>
+			</p>
+			<p
+				v-html="howItWorksSubhead"
+				class="large-6 large-offset-3 columns"
+			>
+			</p>
+		</div>
+		<ol class="how-it-works__list row">
+			<li class="how-it-works__li small-12 xxlarge-3 columns">
+				<img
+					v-if="step1Image.url"
+					class="how-it-works__img"
+					:src="step1Image.url"
+					:alt="step1Image.description"
+				>
+				<p
+					v-html="step1Text[0]"
+					class="how-it-works__subtitle"
+				>
+				</p>
+				<p v-html="step1Text[1]"></p>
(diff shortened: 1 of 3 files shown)


emuvente:
> You think I'm safe to just remove the mention of documentToHtmlString

Yes, `documentToHtmlString` is only needed for Rich Text fields like bodyCopy. Since headline and subhead are just simple text fields, it's not needed.

src/components/IWD/2021/HowItWorks.vue
```diff
@@ -0,0 +1,107 @@
...
+export default {
+	props: {
+		content: {
+			type: Object,
+			default: () => {},
+		},
+	},
+	computed: {
+		howItworksText() {
+			return this.content?.contents?.find(({ key }) => key.indexOf('iwd-2021-homepage-how-it-works-text') > -1);
```
emuvente, line 49: I would recommend using 'homepage-how-it-works-text' as the string that you're checking for, so that this component can be used for other Contentful page layouts that have different prefixes

emuvente (commented):
I think it would make sense to put the component in `src/components/Homepage/` so that it can be used as a generic component for other homepage layouts.

src/components/IWD/2021/HowItWorks.vue
mcstover: good call

mcstover (approved):
Nice cleanup @BoulderBrains from the the first stage to the second in the PR!

BoulderBrains:
@emuvente Thanks for tagging the right Matt. My latest commit should fix up everything you've mentioned above. Thanks for the reveiw.

emuvente (approved):
This looks great! 👍

mlibazisi (approved):
👍

PR #2675, opened 2021-02-19
ML2-233 Filter out any contentful excluded tag names

ryan-ludwig:
Added an excludedTags array on the stover-ct page for testing


mcstover (commented):
Whoa!

PR #2685, opened 2021-02-20
VUE-221 Starting point for lighthouse CI

ryan-ludwig:
Was able to lighthouse CI running with UI locally. Will have to see how things go as a github action.


mcstover (approved):

emuvente (approved):
Nice! 👍

PR #2687, opened 2021-02-22
VUE-488 Stop using Vue prototype for request-specific variables

emuvente:
This is an alternative solution for #2576 and #2585 which renders in the same context (good for performance) and also isolates cookies for each request (fixing basket-related invariant violations).

I did a test using locust, adding a loan to my basket and reloading checkout. There were no invariant violations and I only had one basket id in my Apollo cache.

I also moved $device and $locale out of the Vue prototype in case there would be any issues with them being overwritten by simultaneous requests. 

Finally, I added a check to the Vue plugin installation so that it only happens once on the server, allowing us to use the cached Vue module instead of deleting it every time we render.


src/components/LoansYouMightLike/lymlContainer.vue
```diff
@@ -142,7 +142,7 @@ export default {
 			expMlLoanToLoanVersion: null,
 		};
 	},
-	inject: ['apollo'],
+	inject: ['apollo', 'cookieStore'],
```
mcstover, line 145: Is cookieStore needed in these components that don't use it?

Ok for now but maybe we can clean up later if it's not required in this context

mcstover:
@emuvente This is great. I'll run some local tests this evening as well and hopefully we can merge tmrw so it can bake a couple days.

src/components/LoansYouMightLike/lymlContainer.vue
emuvente: It's needed here because the apollo mixin uses it (to get the basket id).

Initially I tried to put the `inject` option in the mixin itself so that this line wouldn't be needed at all, but I ran into a bunch of warnings from Vue that apollo and cookieStore were not defined (I think it's related to the order we load the plugins, specifically vue-progressbar, which appears to create a Vue instance inside its install function). I decided not to focus on resolving that, as this can be cleaned up later like you said.

ryan-ludwig (approved):

BoulderBrains (approved):

src/components/LoansYouMightLike/lymlContainer.vue
mcstover: Ahh I see. Thanks for the added info!

mcstover (approved):

PR #2692, opened 2021-02-23
ML2-241 Add ability to hideStatusBar and statusMessageOverride from contentfu…

ryan-ludwig:
…l page setting

### DIFF
--- a/src/components/CorporateCampaign/CampaignStatus.vue
+++ b/src/components/CorporateCampaign/CampaignStatus.vue
@@ -10,7 +10,7 @@
 		<div class="small-12 large-8 columns">
 			<div class="campaign-status__message">
 				<template v-if="statusMessageOverride">
-					<span v-html="statusMessageOverride"></span>
+					<span>{{ statusMessageOverride }}</span>
 				</template>
 				<template v-else>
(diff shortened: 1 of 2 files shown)


src/components/CorporateCampaign/CampaignStatus.vue
```diff
@@ -4,33 +4,38 @@
...
+			'campaign-status--success': (promoApplied && !promoErrorMessage) || statusMessageOverride,
 		}"
 	>
 		<div class="small-12 large-8 columns">
 			<div class="campaign-status__message">
-				<template v-if="loadingPromotion && !promoApplied && !promoErrorMessage">
-					<kv-loading-spinner />
-					<span>Validating Promotion</span>
+				<template v-if="statusMessageOverride">
+					<span v-html="statusMessageOverride"></span>
```
mcstover, line 13: could this just been a template string instead of v-html. Seeing as it will come out of a text setting and has no processing? Or do you think we could support basic html in that text string that would make this necessary?

ryan-ludwig: Good call, I don't think we need to have html in there.

mcstover (approved):

PR #2694, opened 2021-02-23
Grow 438

BoulderBrains:
Content served from [Contentful](https://app.contentful.com/spaces/j0p9a6ql0rn7/environments/development/entries/6z2xDCkFRQ0qr9g74EoFNQ?previousEntries=7oYYUSWjeQdlnSUuuOHx70)

Desktop:
Screen Shot 2021-02-23 at 9 35 35 PM

Tablet: 
Screen Shot 2021-02-23 at 9 35 21 PM

Mobile: 
Screen Shot 2021-02-23 at 9 34 57 PM

**FYI**
1) How I have the flourishes positioned will work for the review tomorrow, but if someone updates the text in contentful and the length changes the spacing get messed up. Will have to rework how this is positioned.

2) There's an issue with the spacing at the bottom of the section. This is where i've included the flourish images to then reposition, though they are still taking up space in the dom. I think this is the issue. 

Result: 
(too much space at the bottom of the section)
Screen Shot 2021-02-23 at 8 44 20 PM


emuvente (commented):
To deal with the extra white space, you'll probably need to use `position: absolute` for the flourish images.

To make that easier to work with, I would recommend putting the supporter image and the flourish images inside of a wrapper div. That wrapper `div` should have `position: relative`, and the flourish images should be placed using percent values for top/left/bottom/right.

That should also make it so that you don't need as many breakpoint styles.

emuvente (approved):
This looks good! We can update it if needed after the design review

BoulderBrains:
Merging for now, Cleanup to come.

BoulderBrains:
@mlibazisi I went ahead and merged this because Shua and I have a review meeting with stakeholders this morning. If you see anything during your review, let me know and I'll address it.

PR #2695, opened 2021-02-24
ML2-232 Allow toggling between showing all chips or a subset

ryan-ludwig:


src/components/CorporateCampaign/CampaignPartner.vue
```diff
@@ -96,6 +96,7 @@ export default {
 
 	&__img {
 		display: block;
+		width: 100%;
```
ryan-ludwig, line 99: Sneaking this in. Unrelated to chips.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -37,13 +37,31 @@
...
+					/>
+				</div>
+				<div class="small-12 large-3 xxlarge-2 columns">
+					<kv-button
+						class="loan-filters__chips-toggle text-link"
+						@click.native="chipsCollapsed = !chipsCollapsed"
+					>
+						{{ chipsCollapsed ? `Show all ${filterChips.length} filters` : 'Hide filters' }}
+					</kv-button>
+				</div>
```
mcstover, line 63: Could you please drop in a "Reset All" button here too.

The action needs to emit `this.$emit('reset-loan-filters');` and then the parent will need to listen for `@reset-loan-filters="handleResetLoanFilters"`

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -488,6 +534,24 @@ export default {
 			if (this.selectedSort !== sortBy) {
 				this.selectedSort = sortBy;
 			}
+		},
+		handleResetFilters() {
+			this.modifiedFilters = this.copyFilters(this.initialFilters);
```
mcstover, line 539: Why was this needed, were the filters set hanging around with just the emit? The emitted event should handle resetting things...

mcstover: Ok. I see this is exactly the case so good deal

mcstover (approved):

PR #2704, opened 2021-02-25
Prevent thrash with setting useMatcherAccountIds after adding to bask…

mcstover:
…et by converting initialFilters into a standard data object instead of a computed property.

### DIFF
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -803,7 +803,7 @@ export default {
 				this.updateBasketState();
 			});
 		},
-		async setInitialFilters() {
+		setInitialFilters() {
 			// initialize filter object
 			let filters = LoanSearchFilters();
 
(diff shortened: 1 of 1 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -812,26 +784,54 @@ export default {
...
+				this.setInitialFilters();
+
 				this.showLoans = true;
 
 				this.setAuthStatus(this.kvAuth0?.user);
+
 				this.updateBasketState();
 			});
 		},
+		async setInitialFilters() {
```
ryan-ludwig, line 806: Is there something async in here?

mcstover: oh not any longer. I previous was putting `this.showLoans` above in a then block from setInitialFilters... turns out it wasn't needed in my testing.

ryan-ludwig (approved):

PR #2705, opened 2021-02-25
GDPR-27 OneTrust in Dev

eddieferrer:
* Adds the oneTrust tool for dev environments

### DIFF
--- a/config/dev-vm.js
+++ b/config/dev-vm.js
@@ -56,9 +56,14 @@ module.exports = merge(base, {
 		federationService: {
 			uri: 'https://marketplace-api.dk1.kiva.org/graphql'
 		},
-		paypal : {
+		paypal: {
 			url: 'www.sandbox.paypal.com',
 			environment: 'sandbox'
+		},
+		oneTrust: {
+			enable: true,
+			key: 'db9dcf94-1c32-40fb-9a57-cefafea1088d',
+			domainSuffix: '-test',
 		},
 	},
 	server: {
--- a/src/head/script.js
+++ b/src/head/script.js
@@ -1,114 +1,178 @@
 export default config => {
-	// check for opt out of 3rd party scripts + cookies
-	const cookies = typeof document !== 'undefined' ? document.cookie.split(';') : [];
-	let optout = false;
-	for (let i = 0; i < cookies.length; i++) { // eslint-disable-line
-		if (cookies[i].indexOf('kvgdpr') !== -1 && cookies[i].indexOf('opted_out=true') !== -1) {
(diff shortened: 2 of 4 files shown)


config/dev-vm.js
```diff
@@ -56,10 +56,15 @@ module.exports = merge(base, {
...
 		},
-		paypal : {
+		paypal: {
 			url: 'www.sandbox.paypal.com',
 			environment: 'sandbox'
 		},
+		oneTrust: {
+			enable: true,
+			key: 'db9dcf94-1c32-40fb-9a57-cefafea1088d',
+			domainSuffix: '-test',
```
eddieferrer, line 66: The only difference between the oneTrust test & production scripts is the domainSuffix - for production this block will look like: 
```
			enable: true,
			key: 'db9dcf94-1c32-40fb-9a57-cefafea1088d',
			domainSuffix: '',
```

src/head/script.js
```diff
@@ -1,95 +1,178 @@
...
+			/** Category 'C0002'
+			* Performance Cookies
+			* These cookies allow us to count visits and traffic sources so we can measure and improve the performance
+			* of our site. They help us to know which pages are the most and least popular and see how visitors move
+			* around the site. All information these cookies collect is aggregated and therefore anonymous. If you
+			* do not allow these cookies we will not know when you have visited our site, and will not be able to
+			* monitor its performance.
+			* */
+			if (config.enableAnalytics && config.enableGA) {
+				OneTrust.InsertHtml('', 'head', insertGoogleAnalytics, null, 'C0002');
```
eddieferrer, line 62: This is kind of a hack because the OneTrust API is pretty limited. You only get InsertScript, which doesn't work unless the script is in an external file, and would require some refactoring of the way we load the external libraries. The InsertHtml method here, inserts an empty string into the head of the document, and then calls our callback function

src/head/script.js
```diff
@@ -92,4 +96,109 @@ export default config => {
...
+			* monitor its performance.
+			* */
+			if (config.enableAnalytics) {
+				if (config.enableGA && !optout) {
+					OneTrust.InsertHtml('', 'head', insertGoogleAnalytics, null, 'C0002');
+				}
+				if (config.enableSnowplow) {
+					OneTrust.InsertHtml('', 'head', insertSnowplow, null, 'C0002');
+				}
+				if (config.enableGTM && !optout) {
```
mcstover, line 145: GTM needs to be set as a Targeting tool. It's primary usage is to load up marketing tags.

mcstover (approved):

PR #2707, opened 2021-02-25
IWD grid fixes for bottom cta, sponsors sections

BoulderBrains:
After completing the previous grid fixes, I went back into the pieces of the IWD page I built and implemented the grid consistently. Visually it looks the same, but this is better and more consistent. 🧹


emuvente (approved):
👍

PR #2708, opened 2021-02-25
VUE-491 Push results to our LHCI server on Heroku

ryan-ludwig:


lighthouserc.js
```diff
@@ -14,7 +14,9 @@ module.exports = {
 			]
 		},
 		upload: {
-			target: 'temporary-public-storage',
+			target: 'lhci',
+			serverBaseUrl: 'https://lighthouse-ci-kiva-dev.herokuapp.com/',
+			token: 'LHCI_DEV_BUILD_TOKEN goes here',
```
ryan-ludwig, line 19: If we pass a secret here, LHCI won't be able to run from PRs coming outside of our organization. Adding it in in plaintext here is probably OK but wanted to double check @mcstover @emuvente 

From https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/server.md#security - 
"The build token allows a user to upload new data to a particular project, but does not allow the destruction or editing of any historical data. If your project is open source and you want to collect Lighthouse data on pull requests from external contributors then you should consider the build token to be public knowledge."

ryan-ludwig: I have the build and admin tokens generated, just wasn't sure if we wanted to keep this build token secret or not.

emuvente: If someone uploads junk data then we will be able to delete it, correct? If so I think it's ok to include the build token. If it's abused in someway we can regenerate the build token and keep it secret after that.

lighthouserc.js
```diff
@@ -14,7 +14,9 @@ module.exports = {
 			]
 		},
 		upload: {
-			target: 'temporary-public-storage',
+			target: 'lhci',
+			serverBaseUrl: 'https://lighthouse-ci-kiva-dev.herokuapp.com/',
+			token: process.env.LHCI_DEV_BUILD_TOKEN,
```
mcstover, line 19: can we move this up into the workflow yml

ryan-ludwig: I don't believe so.

mcstover (approved):

PR #2712, opened 2021-03-01
Update handling for setting contentful content links to have target _…

mcstover:
…blank

### DIFF
--- a/src/components/CorporateCampaign/CampaignPartner.vue
+++ b/src/components/CorporateCampaign/CampaignPartner.vue
@@ -70,7 +70,11 @@ export default {
 		// make sure all partner content links open externally
 		if (this.$refs.partnerBodyCopy) {
 			const links = this.$refs.partnerBodyCopy.querySelectorAll('a');
-			links.forEach(link => { link.target = '_blank'; }); // eslint-disable-line no-param-reassign
+			if (links.length > 0) {
+				Array.prototype.forEach.call(links, link => {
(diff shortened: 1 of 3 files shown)


src/components/CorporateCampaign/CampaignPartner.vue
```diff
@@ -70,7 +70,11 @@ export default {
 		// make sure all partner content links open externally
 		if (this.$refs.partnerBodyCopy) {
 			const links = this.$refs.partnerBodyCopy.querySelectorAll('a');
-			links.forEach(link => { link.target = '_blank'; }); // eslint-disable-line no-param-reassign
```
ryan-ludwig, line 73: Haven't tested in IE, but I think Array.from would do the trick too
`Array.from(links).forEach(link => { link.target = '_blank'; }); // eslint-disable-line no-param-reassign)`

ryan-ludwig: Oh, nevermind. https://gomakethings.com/converting-a-nodelist-to-an-array-with-vanilla-javascript/. Unless maybe babel covers that by default.

ryan-ludwig (approved):

src/components/CorporateCampaign/CampaignPartner.vue
mcstover: yep big ol 🤦  on my part. @emuvente helped me sort that out.

emuvente (approved):
👍

PR #2715, opened 2021-03-02
VUE-479 Removes Federation Apollo Client

eddieferrer:
* Removes secondary apollo client

This was the second apolloClient that I created to be able to run contentful queries from the federation service and queries to the vm. This is no longer needed with our new docker setup.


mcstover (approved):

emuvente (approved):
:+1:

PR #2720, opened 2021-03-04
VUE-494 Make a separate LHCI action and settings file for running against prod

ryan-ludwig:
Creates a separate action to run LHCI against prod and send the results to our prod LHCI server on heroku. We can call this action via a webhook after a Jenkins deploy, although I think a better approach would be a nightly github cron since our prod site can change significantly due to Contentful changes and experiments.

Edit: Set as a cron for now to see how that goes


.github/workflows/lighthouse-prod.yml
```diff
@@ -0,0 +1,25 @@
+name: Lighthouse CI PROD
+
+on: workflow_dispatch
```
ryan-ludwig, line 3: setup as manual dispatch for testing, but can change to a CRON https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

mcstover (approved):

emuvente (approved):
👍

PR #2722, opened 2021-03-05
Grow 478

BoulderBrains:
Team, here's my first attempt at merging editable text strings from contentful and data points from graphql. Let me know if you see any issues or areas of improvement.
I've used the new kv-contentful-img component and followed a similar pattern to the one we use to get contentful data for IWD2021Homepage.

[Grow-478](https://kiva.atlassian.net/browse/GROW-478)
[Figma file](https://www.figma.com/file/bmkLizsksdVmUfqHGLxVLt/Kiva-Site-Mock---Current-Home-Page?node-id=0%3A1)

[Contentful (Home - Default) page layout](https://app.contentful.com/spaces/j0p9a6ql0rn7/environments/development/entries/4PDLHnwGuinST6AXBvre5O)

### Mobile: 
Screen Shot 2021-03-04 at 4 07 12 PM

### Tablet: 
Screen Shot 2021-03-04 at 4 07 41 PM

### Desktop:
Screen Shot 2021-03-04 at 4 06 38 PM


src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,216 @@
...
+				:key="statBlock.key"
+				class="small-12 medium-4 columns"
+			>
+				<kv-contentful-img
+					v-if="statBlock.image.url"
+					:class="'statistics__stat-block--icon statistics__stat-block--icon-' + index"
+					:contentful-src="statBlock.image.url"
+					:alt="statBlock.image.description"
+					fallback-format="png"
+					:width="50"
```
BoulderBrains, line 32: @ryan-ludwig How does the width I'm passing through here translate to the size of the img since I'm setting the width of each icon individually? We have a v-if="(width || height)" check within kv-contentful-img, so I've included this width in order to get the icons to display. 50 is about the average width of the icons I'm displaying.

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,216 @@
+<template>
+	<section class="statistics section">
+		<div class="row">
+			<h1
```
ryan-ludwig, line 4: This should probably be an h2 since we should only have one h1 on the page

src/components/Homepage/HomepageStatistics.vue
ryan-ludwig: You're right that you'll always need either a width or a height set. It looks like all of those icons in figma are 48 x 48 if you're exporting it from that layer.
image

So I'd double check that the PNGs you are uploading to contentful are 96x96 (48x48 * 2), and then set width=48 and height=48

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	&__stat-block {
+		margin-top: rem-calc(40);
+
+		&--icon {
+			margin-bottom: rem-calc(20);
+		}
+
+		&--icon-0 {
```
ryan-ludwig, line 194: I think you should be able to get rid of these different icon sizes if they're all exported at 48x48 from Figma

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	&__video-wrapper {
+		&--video {
+			width: 100%;
+			border-radius: rem-calc(10);
+		}
+	}
+
+	&__stat-block {
```
ryan-ludwig, line 187: I'd recommend breaking stat-block into it's own BEM "block". Something like 
```
.stat-block {
  &__icon {...}
  &__stat {...}
}

```

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,216 @@
...
+			return icons.map(image => ({
+				description: image?.description ?? '',
+				title: image?.title ?? '',
+				url: image?.file?.url ?? ''
+			}));
+		},
+		statsVideo() {
+			// Grabbing the 2x video from contentful.
+			// this.content?.media[3] is the 1x video
+			const video = this.content?.media[4] ?? [];
```
ryan-ludwig, line 107: It would be safer to grab the video by using its contentful title instead of its position in the array. Right now if the  
 order where changed in Contentful this could break.

ryan-ludwig, line 107: Have you tried using the 1x video? Does it look good enough?

BoulderBrains: Also added more specificity to the statsIcons. Thanks.

src/pages/Homepage/LendByCategoryHomepage.vue
```diff
@@ -156,27 +156,29 @@
...
-					<h2 class="statistics__header">
-						<span>100%</span>
-						<span class="statistics__header-small">of your loan goes to the field.</span>
-					</h2>
-					<p class="statistics__body">
-						We don't take a penny.
-					</p>
-				</div>
-				<homepage-statistics />
+		<section class="statistics section">
```
mcstover, line 159: Is this block meant to show always? Seems like it may be duplicated content if `statisticsContentfulContentGroup` exists.

BoulderBrains: There is some additional work here, but replacing this 100% loan image and moving it into contentful is handled in this ticket: [Grow-432](https://kiva.atlassian.net/browse/GROW-432)

Once this is brought over to contentful, this top section will be removed and data will be coming through statisticsContentfulContentGroup.

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,211 @@
...
+			const contentfulTextString = this.statsBlockText[2].copy;
+			const numberOfLendersFormatted = numeral(this.numLenders).format('0.0b').slice(0, -1);
+			// Splitting the string returned from contentful, where it finds '{value}'
+			const stringSplit = contentfulTextString.split('{value}');
+			// Pushing the three pieces of the split string the respected arrays
+			part1.push(stringSplit[0]);
+			part2.push(stringSplit[1]);
+			part3.push(stringSplit[2]);
+			// Joining all the pieces of the string and the data from graphql
+			const finalString = part1 + this.numCountries + part2 + numberOfLendersFormatted + part3;
```
mcstover, line 157: Please work on a utility function to add to contentfulUtils to handle this type of string parsing and concatenation. I think you could automate this with a function that accepts your sourceString, splitKey and an array of injected values.

Something like:
`
buildDynamicString(this.statsBlockText[2].copy, '{value}', [this.numCountries, numberOfLendersFormatted]);

buildDynamicString(sourceString, splitKey dynamicValues) {
  // let newString = '';
  // split the sourceString
  // forEach with index the sourceString
  //   > concat the item to the newString
  //   > if dynamicValue[i] exist concat dynamicValue[i] onto the newString
  }
`

emuvente: +1

src/pages/Homepage/Homepage.vue
```diff
@@ -24,7 +24,7 @@ const TopMessageContentful = () => import('./TopMessageContentful');
 
 const activePageQuery = gql`query homepageFrame {
 	contentful {
-		entries(contentType: "page", contentKey: "home")
+		entries(contentType: "page", contentKey: "home-default")
```
emuvente, line 27: I think we'll want to keep the page as "home", but this is fine for now to test out the default page layout. I can handle updating this as a part of GROW-432.

src/components/Homepage/HomepageStatistics.vue
BoulderBrains: Thanks for that Matt, I've added this in the latest commit.

BoulderBrains:
@emuvente & @mcstover 
I created the contentfulUtils function mentioned in your comment above. 

While this is in working order currently. One issue I was unable to resolve before signing off, was grabbing the contentful StatsVideo and StatsIcons using the contentful title name. You see my commented out attempts on lines 101 & 114 of HomepageStatistics.vue

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,205 @@
...
+			type: Object,
+			default: () => {},
+		},
+	},
+	computed: {
+		statsText() {
+			return this.content?.contents?.find(({ key }) => key.indexOf('homepage-statistics-headline-text') > -1);
+		},
+		statsHeadline() {
+			return this.statsText.headline;
```
emuvente, line 87: this.statsText might be undefined, so I would recommend using `this.statsText?.headline ?? ''`

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,205 @@
...
+			return allStatsText?.map((block, index) => ({
+				key: block.key || index,
+				copy: documentToHtmlString(block.bodyCopy ?? ''),
+				image: this.statsIcons[index],
+			}));
+		},
+		statsIcons() {
+			// Attempted to be more specific by using the title to pull in the content,
+			// but wasn't able to get this working
+			// const icons = this.content?.media?.filter(({ key }) => key.indexOf('homepage-stats-icon') > -1);
```
emuvente, line 101: This should work if you use `title` instead of `key`

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,205 @@
...
+				description: image?.description ?? '',
+				title: image?.title ?? '',
+				url: image?.file?.url ?? ''
+			}));
+		},
+		statsVideo() {
+			// Attempted to be more specific by using the title to pull in the content,
+			// but wasn't able to get this working
+			// eslint-disable-next-line max-len
+			// const video = this.content?.media?.find(({ key }) => key.indexOf('homepage-statistics-video-1x') > -1);
```
emuvente, line 114: This should work if you use `title` instead of `key`

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,205 @@
...
+		return {
+			repaymentRate: '',
+			totalLoansInDollars: '',
+			numCountries: '',
+			numLenders: '',
+		};
+	},
+	apollo: {
+		query: whyKivaQuery,
+		preFetch: false,
```
emuvente, line 65: Use `preFetch: true` so that server side rendering works.

src/components/Homepage/HomepageStatistics.vue
```diff
@@ -0,0 +1,205 @@
...
+		&__icon {
+			margin-bottom: rem-calc(20);
+		}
+
+		&__stat {
+			font-size: rem-calc(21);
+
+			&::v-deep i {
+				color: $kiva-green;
+				font-weight: bold;
```
emuvente, line 200: Add `font-style: normal` so that the text isn't italicized.

emuvente (approved):
👍

mcstover (approved):

PR #2730, opened 2021-03-09
Occasional carousel item width bug

ryan-ludwig:
Fix specificity issue depending on what order the CSS bundles are served. Possible it only happen on VMs, but still annoying.
Fixes this:
image


mcstover (approved):
Thank you, I saw this too but couldn't figure what I had done to cause!

PR #2737, opened 2021-03-09
Feature grow 441 guest email input

mlibazisi:
Ticket: GROW-441

Screen Shot 2021-03-09 at 10 52 55 AM

------

Screen Shot 2021-03-09 at 10 53 04 AM

------

Screen Shot 2021-03-09 at 10 53 15 AM

------

Screen Shot 2021-03-09 at 10 53 30 AM


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -8,6 +8,44 @@
...
 			/>
+			<div v-if="isGuestCheckout" id="guest-checkout">
+				<label class="input-label" for="emailReceipt">
+					Where should we email your receipt?
+					<input
+						type="email"
+						name="emailReceipt"
+						v-model="email"
+						id="emailReceipt"
+					>
```
mcstover, line 19: Please add `class="fs-exclude"` to any input element where an email is entered.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -57,59 +98,69 @@ export default {
...
 			this.$emit('updating-totals', true);
 			// Validate Basket prior to starting
 			this.validateBasket()
-				.then(validationStatus => {
-					if (validationStatus === true) {
-						// request payment method
-						this.submitDropInPayment();
-					} else {
-						// validation failed
+					.then(validationStatus => {
```
mcstover, line 120: please clean up all the extra tabs through these files, there was no code change so these tab indexes shouldn't need to change.

mlibazisi: @mcstover I agree. Sorry the PR was still in draft mode. Will have it ready in a few minutes for review. Thanks

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -57,10 +93,20 @@ export default {
 	},
 	data() {
 		return {
+			email: null,
+			isGuestCheckout: false,
```
mcstover, line 97: I think this needs to be a prop...don't see anything internal to this component that would alter this value.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -38,12 +71,15 @@ import braintreeDepositAndCheckout from '@/graphql/mutation/braintreeDepositAndC
 import KvButton from '@/components/Kv/KvButton';
 import KvIcon from '@/components/Kv/KvIcon';
 import BraintreeDropInInterface from '@/components/Payment/BraintreeDropInInterface';
+import KvCheckbox from '@/components/Kv/KvCheckbox';
+import { email, getFailures } from '@/util/validators';
```
emuvente, line 75: Please use [vuelidate](https://vuelidate.js.org/) for validating the form inputs. You can see example usage in the [Monthly Good Setup page](https://github.com/kiva/ui/blob/master/src/pages/MonthlyGood/MonthlyGoodSetupPage.vue)

mlibazisi: 👍

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -8,6 +8,39 @@
...
+						target="_blank"
+						title="Open Privacy Policy in a new window">Privacy Policy</a>.
+				</kv-checkbox>
+				<kv-checkbox
+					id="emailUpdates"
+					class="checkbox"
+					name="emailUpdates"
+					:checked="true"
+				>Receive email updates from Kiva (including borrower updates and promos). You can unsubscribe anytime.
+				</kv-checkbox>
```
emuvente, line 42: These checkboxes are green, but the mocks have them as blue to match with the checkbox in the Braintree dropin. I think we should discuss what we want to do about that in our team meeting and decide if we want to change that as a part of this ticket or as part of another.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -38,29 +71,42 @@ import braintreeDepositAndCheckout from '@/graphql/mutation/braintreeDepositAndC
...
 	data() {
 		return {
+			email: null,
 			enableCheckoutButton: false,
 			paymentTypes: ['paypal', 'card', 'applePay', 'googlePay'],
 		};
 	},
+	validations: {
+		email: { email }
+	},
```
emuvente, line 109: I think you should also add the "required" validator for the email input and the TOU checkbox.

emuvente (commented):
There's some additional work to be done here with providing the email address to the validatePreCheckout mutation and setting the isGuestCheckout prop, but we can talk about whether to do that as a part of this ticket or as part of another in our team meeting.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -38,30 +78,68 @@ import braintreeDepositAndCheckout from '@/graphql/mutation/braintreeDepositAndC
...
+			default: false
 		}
 	},
 	data() {
 		return {
+			email: null,
+			termsAgreement: false,
 			enableCheckoutButton: false,
+			displayEmailErrors: false,
+			displayTermsAgreementErrors: false,
```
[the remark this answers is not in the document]
mlibazisi: @emuvente I had tried that before, but the problem is that it was showing the error message as soon as the page loaded. This started to happen when i added the "required" validation. Any suggestions on how to resolve that? Thanks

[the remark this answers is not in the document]
emuvente: Were you using `$invalid` or `$error`? Because those behave slightly differently. `$error` will be false until the input is interacted with or `$touch()` is called, whereas `$invalid` can be true immediately. That's why I recommend using `$error` for the `v-if`.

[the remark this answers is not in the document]
mlibazisi: @emuvente Ok, i was finally able to get it to work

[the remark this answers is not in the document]
mlibazisi: I was using $invalid. It works now. thanks 👍

emuvente (approved):
👍 Looks good!

PR #2748, opened 2021-03-17
LW-2063 Add FullStory exclude/mask classes to checkout areas

ryan-ludwig:


mcstover (approved):
These look good. Thanks!

src/pages/LoginAndRegister/GuestAccountClaim.vue
```diff
@@ -9,6 +9,7 @@
 			</p>
 			<form id="guestAccountClaimForm" action="." @submit.prevent.stop="claimGuestAccount">
 				<kv-base-input name="firstName"
```
ryan-ludwig, line 11: FYI @emuvente These are the guest checkout component I found. If there's others in UI let me know, or could you add .fs-exclude classes to those inputs when you create them please

emuvente (approved):
👍

PR #2749, opened 2021-03-17
LW-2067 Fullstory mask and excludes on thanks page

ryan-ludwig:


src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -34,7 +38,7 @@
 							data-test="loan"
 							class="loan"
 						>
-							<h3 class="loan__name">
+							<h3 class="loan__name fs-exclude">
```
mcstover, line 41: I don't think we need to mask this one right? This is the public loan name...

mcstover, line 41: 👍  verified ticket spec

mcstover (approved):

PR #2750, opened 2021-03-17
LW-2062 full story lend by category

ryan-ludwig:


ryan-ludwig:
@mcstover This one is clunky since it will mask _all_ category names, not just the personalized one. The special category name is set [here](https://github.com/kiva/ui/blob/2bab9ec53683fa2a5dc6c06eb548e6b4fb6a13e9/src/pages/Lend/LendByCategoryPage.vue#L190). 

Another approach would be to check the loan channel id within these components, and if it's 91, add the fs-mask class. Thoughts?

mcstover:
@ryan-ludwig I think it would be good to use a conditional class on the h2 "if" the loanChannel id is 95. Hopefully, we'll just remove that sometime in the near future...but for now it would at least allow most category row headings to remain visible.

mcstover (approved):

PR #2751, opened 2021-03-17
GROW-454 handle guest account claim completion redirection

emuvente:
Handles redirecting guest users who have just created their passwords through login and then on to either /portfolio or /checkout, and shows a tip message confirming the creation of the account.


BoulderBrains (approved):

mlibazisi (approved):
👍

PR #2752, opened 2021-03-17
LW-2056 full story payments

ryan-ludwig:
image

### DIFF
--- a/src/pages/Settings/PaymentSettings.vue
+++ b/src/pages/Settings/PaymentSettings.vue
@@ -24,7 +24,7 @@
 								class="payment-settings-default-form__cc-wrapper"
 							>
 								<kv-radio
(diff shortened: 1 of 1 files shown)


src/pages/Settings/PaymentSettings.vue
```diff
@@ -24,7 +24,7 @@
 								class="payment-settings-default-form__cc-wrapper"
 							>
 								<kv-radio
-									class="payment-settings-default-form__radio"
+									class="payment-settings-default-form__radio fs-mask"
```
mcstover, line 27: I think we should either exclude all of this or wrap `paymentMethod.description` in a span and exclude just that if we want to see the payment type icon.

mcstover (approved):

PR #2755, opened 2021-03-18
ML donation and loading state hotfixes

ryan-ludwig:
* Show donation field for matching campaigns. Show cart lightbox after logging in for matching campaigns.
* Showing a loading state on page load if loans are in cart


mcstover (approved):

mcstover:
Verified in Dev

PR #2760, opened 2021-03-22
Gdpr 38 asyn analytics setup

mcstover:
Creates a polling mechanism and promise based setup for analytics initialization before attempting to fire the page view events.


mcstover:
ref https://github.com/kiva/ui/pull/2759

PR #2761, opened 2021-03-22
VUE-510 Update Husky and fix linting hook

eddieferrer:
* Fixes CheckoutDropIn formatting
* Update husky to v5
* fix husky hooks


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -27,23 +27,32 @@
 					name="termsAgreement"
```
eddieferrer, line 27: Only auto formatting changes for this file. Not sure why the formatting was so whacky.

eddieferrer:
`npm run lint` is a tad slower than I would like. We could alternatively only lint staged files on commit instead of pre-push. This would lint significantly faster, but it would not allow committing files with lint errors, which could be an inconvenience with work in progress commits.

ryan-ludwig (approved):

BoulderBrains (approved):

mcstover (approved):
We can test out the time difference for a little while then look into tuning the linting a bit.

emuvente (commented):
👍

PR #2762, opened 2021-03-22
SUBS-282 Added MG Landing Page Hero Image to Contentful

eddieferrer:
Get KvHero image for MG landing page from contentful


eddieferrer:
This approach isn't ideal

PR #2764, opened 2021-03-23
GROW-512 Change guest checkout checkbox colors

mlibazisi:
Ticket: GROW-512

Screen Shot 2021-03-22 at 4 57 37 PM

### DIFF
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -106,7 +106,7 @@ export default {
 		},
 		isGuestCheckout: {
 			type: Boolean,
-			default: true,
+			default: false,
 		},
 	},
 	data() {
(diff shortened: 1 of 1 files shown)


emuvente (approved):
👍

BoulderBrains (approved):

BoulderBrains:
I just noticed in the figma spec the the size of these checkboxes appears to be bigger than what's implemented here. I know the ticket only mentions a color change, so maybe that work doesn't belong in here. Either way, I messed around with the css and something like this should work:

`#guest-checkout {
	.kv-checkbox {
		.input {
			&:checked + .label,
			&:active + .label {
				.square {
                                         width: 1.5rem;
                                         height: 1.5rem;
					background-color: $blue;
					border-color: $blue;
				}
                                  .square::after {
                                       top: 1px;
                                       font-size: rem-calc(27);
                                       left: rem-calc(7);
                                   }
			}
		}
	}
}`

mlibazisi:
> #guest-checkout { .kv-checkbox { .input { &:checked + .label, &:active + .label { .square { width: 1.5rem; height: 1.5rem; background-color: $blue; border-color: $blue; } .square::after { top: 1px; font-size: rem-calc(27); left: rem-calc(7); } } } } }

@BoulderBrains when I tested the code above, it gives 3 different sizes. So if we wanted to go that direction, a little more work will have to be done, maybe in a separate ticket.
Screen Shot 2021-03-23 at 10 31 50 AM

PR #2765, opened 2021-03-23
VUE-511, lighthouse, external links need rel=noopener for security

ryan-ludwig:


src/components/WwwFrame/TheFooter.vue
```diff
@@ -218,6 +218,7 @@
 								<a
 									:href="careersUrl"
 									target="_blank"
+									rel="noopener"
```
mcstover, line 221: seems kinda odd that we would set this for our own domain with a subdomain that we control.

mcstover: Also found this note about `_blank` providing the same functionality pretty interesting...
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

So I'm guessing lighthouse is saying we should set noopener b/c we are using target _blank...but not considering that the intended affect is already in place for some browsers.

mcstover (approved):

src/components/WwwFrame/TheFooter.vue
ryan-ludwig: Oh nice, I'd heard browsers were working on a better fix than having to add the noopeners all the time. Cool to see, and pretty good support!
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility

I'm mainly trying to clean up quick lighthouse errors so there's a better signal to noise ratio in the reports

ryan-ludwig:
Looks the audit will dissappear in the next lighthouse version - https://github.com/GoogleChrome/lighthouse/issues/9531

PR #2775, opened 2021-03-24
VUE-498 Refactor DonateFromMacro Hero Image

eddieferrer:
* Get responsive sources using single image from contentful
* Uses new KvContentfulImg capabilities to create a set of responsive image sources using the image from contentful
* Allows us to have a single asset on contentful as opposed to 9


BoulderBrains (approved):
👍

PR #2776, opened 2021-03-24
Skip husky install in CI environments

eddieferrer:


package.json
```diff
@@ -34,7 +34,7 @@
 		"storybook": "start-storybook -p 6006",
 		"build-storybook": "build-storybook",
 		"deploy-storybook": "storybook-to-ghpages",
-		"prepare": "husky install"
+		"prepare": "if [[ !${CI} != *true ]]; then husky install; fi"
```
mcstover, line 37: Do we need to set this somewhere?

package.json
```diff
@@ -167,6 +166,7 @@
 		"git-revision-webpack-plugin": "^3.0.6",
 		"hard-source-webpack-plugin-fixed-hashbug": "^0.13.2",
 		"html-minifier": "^4.0.0",
+		"husky": "^5.2.0",
```
mcstover, line 169: did you run npm i again after moving this to update package-lock

eddieferrer: Whoops! Will do

package.json
eddieferrer: It's set in `Jenkinsfile`

mcstover: oh indeed!

mcstover (approved):

PR #2779, opened 2021-03-24
GROW-519 show braintree drop-in for guest checkout

emuvente:
This will show the braintree drop-in and guest checkout fields when a non-logged in user selects "Checkout as guest". This does not remember that the user selected that option, so if they refresh the page before completing the checkout they will need to select "Checkout as guest" again. We can change that behavior in another ticket if needed.

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -83,6 +83,7 @@
 									<checkout-drop-in-payment-wrapper
 										v-else
 										:amount="creditNeeded"
+										:is-guest-checkout="checkingOutAsGuest"
 										@refreshtotals="refreshTotals"
 										@updating-totals="setUpdatingTotals"
 										@complete-transaction="completeTransaction"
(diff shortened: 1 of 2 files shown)


BoulderBrains (approved):

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -506,7 +514,7 @@ export default {
 			// Doing nothing here allows the normal link handling to happen, which will send the user to /ui-login
 		},
 		guestCheckout() {
-			console.log('Guest checkout method triggered.');
+			this.checkingOutAsGuest = true;
```
mcstover, line 517: should this be a togglable method? meaning can they switch away from guestCheckout after selecting it?

mlibazisi (approved):
👍

src/pages/Checkout/CheckoutPage.vue
emuvente: There isn't a button or link to back out of the guest checkout process in any part of the flow I can see, but refreshing the page would reset them.

PR #2784, opened 2021-03-24
GROW-509 thanks tracking

ryan-ludwig:


emuvente (approved):

src/components/Checkout/GuestUpsell.vue
```diff
@@ -86,12 +81,25 @@ export default {
...
+			if (this.borrowerName) {
+				return `Get updates from ${this.borrowerName} and choose
+					how to re-lend your money when they pay you back.`;
+			}
+			return `Get updates from the borrowers you support and choose
+				how to re-lend your money when they pay you back.`;
+		}
+	},
+	mounted() {
+		this.$kvTrackEvent('Thanks', 'view-register-upsell', this.borrowerUpdateText);
```
ryan-ludwig, line 94: @emuvente Do you think doing this on mounted might be problematic? Trying to fire the event anytime the registration upsell is shown.

emuvente: I think that should be fine, but we can play around on the thanks page a bit and see if it behaves in an unexpected or non-intuitive way

PR #2788, opened 2021-03-25
VUE-513 Remove processPageContent

eddieferrer:
* processPageContent was only being used on 2 pages, removed in favor of processPageContentFlat.
* This will make it easier to improve the processPageContentFlat function to process pages with multiple layout references

### DIFF
--- a/src/pages/Donate/DonateFromMacro.vue
+++ b/src/pages/Donate/DonateFromMacro.vue
@@ -40,7 +40,7 @@ import gql from 'graphql-tag';
 import { lightHeader, lightFooter } from '@/util/siteThemes';
 import WwwPage from '@/components/WwwFrame/WwwPage';
 import DonateFromMacroHero from '@/pages/Donate/DonateFromMacroHero';
-import { processPageContent } from '@/util/contentfulUtils';
+import { processPageContentFlat } from '@/util/contentfulUtils';
 import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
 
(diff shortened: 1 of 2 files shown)


emuvente:
I'm actually working on a generic Contentful page component that uses `processPageContent`, because `processPageContentFlat` doesn't let you know the order of the items in `contentGroups`, and it also doesn't let you have more than one content group of a given type. I would really want `processPageContentFlat` to support those things if we're going to get rid of `processPageContent`

mcstover:
@eddieferrer Let's do please keep the `processPageContent` method for now so we can work out some of the ideas of how best to access a dynamic list of ordered content groups in isolation.

eddieferrer:
> I'm actually working on a generic Contentful page component that uses `processPageContent`, because `processPageContentFlat` doesn't let you know the order of the items in `contentGroups`, and it also doesn't let you have more than one content group of a given type. I would really want `processPageContentFlat` to support those things if we're going to get rid of `processPageContent`

@emuvente sounds great, would it be helpful then to merge the refactor of the Thanks/Donate component here? Then at least all the components are using the same current function at the moment. Or should I just trash this PR?

mcstover:
I'm to see the refactor of the other pages if you can keep them that's good progress.

emuvente:
@eddieferrer yeah the refactor of the thanks/donate components would be helpful and it looks good!

eddieferrer:
@mcstover @emuvente Kept the refactor of the 2 components. Left contentful utils as they are. Just FYI, I tested this, and it will not require any changes on contentful.

emuvente (approved):
👍

mcstover (approved):

PR #2789, opened 2021-03-26
VUE-512 live loan images, loan-to-loan

ryan-ludwig:
Adds ability to hit loan-to-loan recommendation urls like 
`/live-loan/l/*loanid*/img/1` &
`/live-loan/l/*loanid*/url/1`

Retains loginid recommendation urls like
 `/live-loan/u/*loginid*/img/1` &
`/live-loan/u/*loginid*/url/1`
 
Refactor to reduce duplicate code.

### DIFF
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -76,12 +76,15 @@ function fetchRecommendedLoans(type, id, cache) {
 	let query;
 	let queryResultPath;
 	if (type === 'user') {
-		query = recommendationsByLoginIdQuery;
+		query = recommendationsByLoginIdQuery(id);
 		queryResultPath = 'data.ml.recommendationsByLoginId.values';
 	} else if (type === 'loan') {
-		query = recommendationsByLoanIdQuery;
+		query = recommendationsByLoanIdQuery(id);
 		queryResultPath = 'data.ml.relatedLoansByTopics[0].values';
+	} else {
+		throw new Error('type must be user or loan');
 	}
+
 	return new Promise((resolve, reject) => {
 		memJsUtils.getFromCache(`recommendations-by-${type}-id-${id}`, cache).then(data => {
 			if (data) {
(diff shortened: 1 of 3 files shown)


mcstover:
Could you please create a util method (could be in this file for now) to format all console logs or errors with this structure? 

Just switch between info and error where appropriate.

`console.info/error/warning(JSON.stringify({
    meta: {},
    level: 'info', // info or error or warning
    message: ###message###
})));`

server/live-loan-router.js
```diff
@@ -3,170 +3,201 @@ const express = require('express');
...
+			await memJsUtils.setToCache(
+				`recommendations-by-${type}-id-${id}`,
+				JSON.stringify(loanData),
+				expires,
+				cache
+			);
+			return loanData;
+		} catch (err) {
+			// throw if we can't save to memcache since not using caching would be problematic during an email campaign
+			throw new Error(`Error setting loan data to cache: ${err}`);
```
ryan-ludwig, line 131: Should we error out if we can't save to memcache here?

ryan-ludwig:
Would love to get a few eyes on this since quite a bit was refactored.

server/live-loan-router.js
mcstover: We should initially format and log the error for now...(on line 115 too)

A future improvement would be to return the results of a generic loan set for both the image and url states.

server/live-loan-router.js
```diff
@@ -3,170 +3,203 @@ const express = require('express');
...
+				expires,
+				cache
+			);
+			return loanData;
+		} catch (err) {
+			log(`Error setting loan data to cache, ${err}`, 'error');
+		}
+	} else {
+		// future improvement, return a default set of 4 loans here
+		throw new Error('No loans returned');
```
ryan-ludwig, line 134: @mcstover I'm throwing here to reject out of the promise, which gets caught and logged in `redirectToUrl` and `serveImg`. An alternative is to resolve as a blank array, and then check that we have loans before processing on lines 143 and 159.

mcstover (approved):

PR #2795, opened 2021-03-27
Extract and preserve utm params in the url as we manage filter/facet …

mcstover:
…state.

### DIFF
--- a/src/util/algoliaUtils.js
+++ b/src/util/algoliaUtils.js
@@ -1,6 +1,9 @@
 import _filter from 'lodash/filter';
 import _find from 'lodash/find';
 
+// temporary store for utm query params
+let utmParams = [];
+
 // dev_fundraising_loans is the base index, all others are replicas
 // INFO: This list powers the sort order component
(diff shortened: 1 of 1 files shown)


src/util/algoliaUtils.js
```diff
@@ -76,9 +79,25 @@ export function rebuildSortByIndexFromRoute(routeStateSortBy) {
...
+		return utm;
+	});
+	return Object.fromEntries(utmArray);
+}
+
 // Build Algolia specific map of current search state to url query parameters
 export function stateToRoute(uiState) {
-	// console.log(`uiState: ${JSON.stringify(uiState)}`);
+	// extract and assign utm params if present
+	if (window.location.search.indexOf('utm') !== -1 && !utmParams.length) {
```
mcstover, line 97: I verified that both stateToRoute and routeToState "know" about the windon.location.search value during page load and throughout their first processing cycle. routeToState runs first and will establish the cached set of utm params if they are present, then stateToRoute will use them.

PR #2804, opened 2021-03-31
VUE-315 Email Settings

eddieferrer:
* Initial new stack email settings page

Working great, all functionality duplicated from legacy page. 

**STILL TODO:**
* Add trustee email settings section
* Add borrower email settings section
* Add service worker for lending notifications on browser
* Set up this new route in kiva/main and puppet repos
* Delete legacy code in kiva/main and styleguide repos

screencapture-dev-vm-01-kiva-org-settings-email-2021-03-31-09_42_11

### DIFF
--- a/src/pages/Settings/EmailSettings.vue
+++ b/src/pages/Settings/EmailSettings.vue
@@ -1,0 +1,497 @@
+<template>
+	<www-page class="email-settings" :gray-background="true">
+		<the-my-kiva-secondary-menu slot="secondary" />
+		<div class="email-settings__title-area">
+			<div class="row column">
+				<h1>Email Settings</h1>
+			</div>
+		</div>
+		<form @submit.prevent>
+			<div class="row">
+				<kv-settings-card class="columns small-12 large-8" title="">
+					<template #content>
+						<p>
+							To customize the communications you receive select specific emails
+							from the list below. You can also disable most email communication
+							but
+							<strong>Kiva is still legally required to send a few emails</strong>
+							about your account status and activity.
+						</p>
+						<label for="global-unsubscribed-input">
+							<strong>Do you want to receive emails from Kiva?</strong>
+						</label>
+						<kv-dropdown-rounded
+							id="global-unsubscribed-input"
+							v-model="form.globalUnsubscribed"
+						>
+							<option :value="false">
+								Yes, send me emails
+							</option>
+							<option :value="true">
+								Send only legally required emails
+							</option>
+						</kv-dropdown-rounded>
+					</template>
+				</kv-settings-card>
+			</div>
+			<div class="row">
+				<!-- General Settings -->
+				<kv-settings-card
+					class="columns small-12 large-8"
+					title="General Kiva / account updates"
+					:disabled="form.globalUnsubscribed"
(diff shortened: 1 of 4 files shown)


src/pages/Settings/PaymentSettings.vue
```diff
@@ -196,6 +196,9 @@ export default {
 		TheMyKivaSecondaryMenu,
 		WwwPage,
 	},
+	metaInfo: {
```
eddieferrer, line 199: Unrelated to email settings, but I noticed payment settings was missing a title

src/pages/Settings/SettingsPage.vue
```diff
@@ -92,7 +92,7 @@
 					class="section__link"
 					href="/settings/payments"
 				>
-					<h2 class="section__title">Payment Settings</h2>
+					<h2 class="section__title">Payment settings</h2>
```
eddieferrer, line 95: Capitalization to match other sections

src/pages/Settings/EmailSettings.vue
```diff
@@ -0,0 +1,518 @@
...
+		<div class="row">
+			<!-- Loan Settings -->
+			<kv-settings-card
+				class="columns small-12 large-8"
+				title="Lending activity updates"
+				:disabled="form.globalUnsubscribed"
+			>
+				<template #content>
+					<form @submit.prevent>
+						<fieldset>
```
ryan-ludwig, line 131: There's a lot of fieldsets here. They're usually for grouping a set of related inputs. So for this area, I think you'd have the Select All button and the two checkboxes below in the same fieldset.

ryan-ludwig: Or maybe everything in this settings-card is in the same fieldset

src/pages/Settings/EmailSettings.vue
```diff
@@ -0,0 +1,518 @@
...
+					</form>
+				</template>
+			</kv-settings-card>
+		</div>
+
+		<!-- !TODO Add Trustee email settings section  -->
+		<!-- !TODO Add Borrower email settings section  -->
+		<div class="row">
+			<div class="columns small-12 large-8">
+				<kv-button
```
ryan-ludwig, line 185: I think semantically it would be make sense to have all of these settings cards wrapped in one single form. Then this button would be a type="submit" and wouldn't need @click.native. This would also allow you to hit "enter" while you're editing the form to submit it.

Roughly
`<form @submit.prevent="saveSettings" :disabled="!isChanged || isProcessing">` 
and 	
`<kv-button class="smallest" type="submit" :disabled="!isChanged || isProcessing">`

eddieferrer: Wrapped it in a single form, I agree... I tried making the button a submit button, but ran into an unfortunate issue - Changing the first drop down, focus seems to move to the "Select All" button, hit enter, the form then both activates "Select All" and submits. 
So I left the default form submit null and kept the manually firing of the submit function on the button

ryan-ludwig (approved):

BoulderBrains (approved):
Nice port of the email settings.

PR #2806, opened 2021-03-31
grow-545 removing authentication requirement for /register/guest-redi…

BoulderBrains:
…rect, adding check within GuestAccountRedirect.vue

Testing this flow out locally was a bit problematic. 

I was able to verify that going to /register/guest-redirect while: 
logged out with loan in basket: 
- Resulted in a doneUrl of /checkout

logged in: 
- Resulted in a done url of /portfolio.

### DIFF
--- a/src/pages/LoginAndRegister/GuestAccountRedirect.vue
+++ b/src/pages/LoginAndRegister/GuestAccountRedirect.vue
@@ -31,6 +31,7 @@ export default {
 				// Check to see if user is authenticated
 				if (!data?.my?.userAccount?.id) {
 					redirectPath = '/ui-login';
+					query.login_hint = 'login';
 					query.doneUrl = finalUrl;
 				}
 				return Promise.reject({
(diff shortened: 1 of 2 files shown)


src/pages/LoginAndRegister/GuestAccountRedirect.vue
```diff
@@ -14,13 +14,28 @@ export default {
...
+				const finalUrl = data?.shop?.nonTrivialItemCount > 0 ? '/checkout' : '/portfolio';
 				// Add claimed=1 to the url to show a confirmation tip message on the page
+				const query = { claimed: 1 };
+
+				let redirectPath = finalUrl;
+
+				// Check to see if user is authenticated
+				if (!data?.my?.userAccount?.id) {
+					redirectPath = '/ui-login';
+					query.doneUrl = finalUrl;
```
emuvente, line 34: I think you'll need to also set `query.login_hint = 'login';` so that the login page is redirected to instead of register.

emuvente (approved):
👍

src/pages/LoginAndRegister/GuestAccountRedirect.vue
mcstover: should we add the login_hint for both scenarios?

PR #2808, opened 2021-04-01
VUE-524 VUE-525 Additional Email Settings

eddieferrer:
* Email settings for borrowers and trustees


mcstover (approved):
so many settings!

PR #2812, opened 2021-04-01
GROW-556 Add additional registration paths to login sync bypass list

emuvente:
I'm not certain that this will fox the GROW-556 error, but I highly suspect that it will, and it's worth having anyway even if it doesn't fix the problem.


mcstover (approved):

BoulderBrains (commented):
👀 👍

PR #2819, opened 2021-04-06
SUBS-499 Friendly BT Errors

eddieferrer:
* Extracted BT DropIn Wrapper error processing to a plugin
* The UI has to use extensions.code instead of errors.code to display friendly errors

The API returns a response that looks like:
```
{
  "errors": [
    {
      "message": "Some error",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "shop",
        "doNoncePaymentDepositAndCheckout"
      ],
      "code": "braintree.transactionRequestRejected",
      "extensions": {
        "code": "braintree.transactionRequestRejected"
      }
    }
  ],
  "data": {
    "shop": {
      "doNoncePaymentDepositAndCheckout": null
    }
  }
}
```
This can be replicated in the GQL playground, but on the UI side, `errors[0].code` is always undefined. I'm not sure where that is happening. Maybe `src/util/apolloPreFetch.js` But these queries aren't in preFetch so I'm not sure. 

`errors[0].extensions.code` contains the same error as `errors[0].code` that we want to log and display, so modified mutation error handling to use that, also moved the error handling into a plugin since all these components use the exact same error handling

Looker showing the BT error code always undefined: https://kiva.looker.com/sql/hnkrkc4dnmgvst

While working on this, I noticed that there are a couple of other mutation handlers that are looking for `errors[0].code`, are these working correctly? -- For example: `validateGuestBasketAndCheckout`

### DIFF
--- a/src/plugins/braintree-dropin-error-mixin.js
+++ b/src/plugins/braintree-dropin-error-mixin.js
@@ -7,8 +7,6 @@ export default {
 			// eslint-disable-next-line max-len
 			const standardError = `There was an error processing your payment. Please try again. ${standardErrorCode}`;
 
-			// Payment method failed, unselect attempted payment method
-			this.$refs.braintreeDropInInterface.btDropinInstance.clearSelectedPaymentMethod();
 			// Potential error message: 'Transaction failed. Please select a different payment method.';
 			this.$showTipMsg(standardError, 'error');
 
--- a/src/plugins/braintree-dropin-error-mixin.js
+++ b/src/plugins/braintree-dropin-error-mixin.js
@@ -1,0 +1,19 @@
+export default {
+	methods: {
+		processBraintreeDropInError(trackCategory, kivaBraintreeResponse) {
+			const errorCode = kivaBraintreeResponse.errors?.[0]?.extensions?.code;
+			const errorMessage = kivaBraintreeResponse.errors?.[0]?.message;
+			const standardErrorCode = `(Braintree error: ${errorCode})`;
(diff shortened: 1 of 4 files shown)


src/plugins/braintree-dropin-error-mixin.js
```diff
@@ -0,0 +1,19 @@
...
+	methods: {
+		processBraintreeDropInError(trackCategory, kivaBraintreeResponse) {
+			const errorCode = kivaBraintreeResponse.errors?.[0]?.extensions?.code;
+			const errorMessage = kivaBraintreeResponse.errors?.[0]?.message;
+			const standardErrorCode = `(Braintree error: ${errorCode})`;
+			// eslint-disable-next-line max-len
+			const standardError = `There was an error processing your payment. Please try again. ${standardErrorCode}`;
+
+			// Payment method failed, unselect attempted payment method
+			this.$refs.braintreeDropInInterface.btDropinInstance.clearSelectedPaymentMethod();
```
mcstover, line 11: this seems a little out of place... but if it's global and should be executed for "any" error, I suppose it's ok.

mcstover (approved):
Nice update!

src/plugins/braintree-dropin-error-mixin.js
eddieferrer: I agree with this actually, moved it back to the payment wrappers so the ref is at least in the same file

mcstover:
👍

PR #2820, opened 2021-04-06
Grow 469

BoulderBrains:
Grow-469: https://kiva.atlassian.net/browse/GROW-469

Added a guest: true flag to be read off the URL so alternative messaging can be shown.

### DIFF
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -171,7 +171,6 @@ export default {
 							const loginHint = encodeURIComponent(
 								`login|${JSON.stringify({
 									msg: 're-auth-acc-exists',
-									guest: true,
 								})}`
 							);
 							window.location = `/ui-login?force=true&doneUrl=/checkout&login_hint=${loginHint}`;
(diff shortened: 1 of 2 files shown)


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -171,6 +171,7 @@ export default {
 							const loginHint = encodeURIComponent(
 								`login|${JSON.stringify({
 									msg: 're-auth-acc-exists',
+									guest: true,
```
emuvente, line 174: This shouldn't be used here because this is a case of an account already existing with the email address they tried to use, meaning that they aren't a guest.

The `guest: true` should be added to the login redirect in GuestAccountRedirect.vue instead

BoulderBrains: Thanks for the review. I've moved this to the correct location in my latest commit.

src/pages/LoginAndRegister/GuestAccountRedirect.vue
```diff
@@ -34,8 +34,10 @@ export default {
 					return Promise.reject({
 						path: '/ui-login',
 						query: {
-							login_hint: 'login',
-							doneUrl: `${path}?${queryString}`,
+							loginHint: `login|${JSON.stringify({
+								guest: true,
+							})}`,
+							doneUrl: `${path}?${queryString}&login_hint=${query.loginHint}`,
```
emuvente, line 40: The login hint doesn't need to be included in the done url

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
emuvente: 👍

emuvente (approved):
👍

PR #2823, opened 2021-04-08
Use exclude wherever the description can be used/shown as it can cont…

mcstover:
…ain an email address in the case of vaulted PayPal methods.

### DIFF
--- a/src/pages/Settings/PaymentSettings.vue
+++ b/src/pages/Settings/PaymentSettings.vue
@@ -33,7 +33,7 @@
 										:src="paymentMethod.imageUrl"
 										alt="credit card"
 									>
-									<span class="fs-mask">{{ paymentMethod.description }}</span>
+									<span class="fs-exclude">{{ paymentMethod.description }}</span>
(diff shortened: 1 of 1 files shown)


src/pages/Settings/PaymentSettings.vue
```diff
@@ -131,9 +133,11 @@
 		<kv-lightbox
 			class="active-card-lightbox"
 			:visible="showActiveLightbox"
-			:title="`Unable to remove card ${lowerCaseDescription}`"
```
ryan-ludwig, line 134: Not a blocker, but another approach would be to allow HTML in the title prop in KvLightbox so we can add a span here instead of bypassing the standard title.

ryan-ludwig (approved):

src/pages/Settings/PaymentSettings.vue
mcstover: That's certainly one approach. I also considered passing a prop with a css class for the title area...

PR #2825, opened 2021-04-08
GROW-574 Fixes ThanksPage V2 experiment when user is not a guest

eddieferrer:


src/pages/Thanks/ThanksPage.vue
```diff
@@ -188,7 +188,16 @@ export default {
 			// But it will not throw a server error.
 			this.receipt = data?.shop?.receipt;
 			this.isGuest = this.receipt && !data?.my?.userAccount;
-			this.thanksPageVersion = this.isGuest ? 'b' : 'a';
```
eddieferrer, line 191: This line seemed to be overwriting the change in the mounted block for some reason, in my testing result sometimes finished after the mounting code... kind of puzzling.

mcstover (approved):

src/pages/Thanks/ThanksPage.vue
emuvente: result is also called everytime the watchQuery call has a new result, so it could be that

emuvente (approved):
👍

src/pages/Thanks/ThanksPage.vue
eddieferrer: @emuvente good call, I didn't think about that!

PR #2829, opened 2021-04-09
GROW-573  Hide guest checkout for kiva cards

mlibazisi:
Ticket: GROW-573

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -316,7 +316,7 @@ export default {
 				{ __typename: 'Credit', creditType: 'redemption_code' }
 			);
 			this.hasFreeCredits = _get(data, 'shop.basket.hasFreeCredits');
-			if (this.redemption_credits || this.hasFreeCredits) {
+			if (this.redemption_credits.length || this.hasFreeCredits !== false) {
 				this.disableGuestCheckout();
 			}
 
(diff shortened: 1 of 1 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -317,6 +316,9 @@ export default {
 				{ __typename: 'Credit', creditType: 'redemption_code' }
 			);
 			this.hasFreeCredits = _get(data, 'shop.basket.hasFreeCredits');
+			if (this.kivaCards || this.redemption_credits || this.hasFreeCredits) {
```
emuvente, line 319: We should only check for `if (this.redemption_credits || this.hasFreeCredits) {` because we still want to allow guest users to purchase kiva cards. It's using kiva cards (redemption credit) that we want to block.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -572,6 +581,9 @@ export default {
 				if (hasFreeCredits) {
 					if (refreshEvent === 'kiva-card-applied') {
 						this.$kvTrackEvent('basket', 'free credits applied', 'exit to legacy');
+						if (this.checkingOutAsGuest) {
+							this.disableGuestCheckout();
+						}
```
emuvente, line 586: I don't think we need the `if` here, I think we can just call `this.disableGuestCheckout()` because we've already checked that free credits were being applied to the basket.

emuvente (approved):
👍

BoulderBrains (approved):
💯

PR #2833, opened 2021-04-12
VUE-526 Push notifications for new email settings page

eddieferrer:
* Added push notifications util and component
* Push notificication will still use legacy service worker


eddieferrer:
Possibly working? We can push to dev and verify there since testing service workers on the VM seems pretty hard. src/util/pushNotificationsManager.js references and installs the service worker: `/js/kv-notifications/kvPushNotificationsServiceWorker.js` which is the legacy service worker and lives on the old stack and is unchanged

emuvente (approved):
👍  looks great!

mcstover (approved):

PR #2834, opened 2021-04-12
SUBS-663 Added Route for dynamic landing page

eddieferrer:
Added dynamic route for `/lp/support-causes-monthly` 
Added content on contentful, for now just a copy of the `Home` page content. 
I made the page key `lp/support-causes-monthly` which I know doesnt match the format of the keys for the cc landing pages, but thought this would be more descriptive

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -98,6 +98,13 @@ export default [
 	{ path: '/lend-by-category/:category', component: () => import('@/pages/Lend/LoanChannelCategoryPage') },
 	{ path: '/lend/filter', component: () => import('@/pages/Lend/Filter/LendFilterPage') },
 	{
+		path: '/lp/:dynamicRoute',
+		component: () => import('@/pages/ContentfulPage'),
+		meta: route => ({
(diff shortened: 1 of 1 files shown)


src/router/routes.js
```diff
@@ -97,6 +97,13 @@ export default [
 	{ path: '/lend-by-category', component: () => import('@/pages/Lend/LendByCategoryPage') },
 	{ path: '/lend-by-category/:category', component: () => import('@/pages/Lend/LoanChannelCategoryPage') },
 	{ path: '/lend/filter', component: () => import('@/pages/Lend/Filter/LendFilterPage') },
+	{
+		path: '/lp/:dynamicRoute',
+		component: () => import('@/pages/ContentfulPage'),
+		meta: route => ({
+			contentfulPage: `lp/${route.params.dynamicRoute}`,
```
mcstover, line 104: if the `/` is ok in the queries I don't see an issue with this and think it's useful to have some namespacing built into those keys seeing as they must be unique.

mcstover (approved):

PR #2838, opened 2021-04-13
VUE-551 Use header & footer themes for the page frame in ContentfulPage

emuvente:
The `headerTheme` and `footerTheme` fields on the `PageLayout` content type in Contentful are setup as a dropdown with limited options, similar to the `type` field on `ContentGroup`. The options for each one match up with the names of the exported header and footer themes from `src/util/siteThemes.js`.

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -1,5 +1,8 @@
 <template>
-	<component :is="pageFrame">
+	<component :is="pageFrame"
+		:header-theme="headerTheme"
+		:footer-theme="footerTheme"
+	>
 		<template v-if="!pageError">
 			<component
 				v-for="({ component, content }) in contentGroups"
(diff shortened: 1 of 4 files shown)


src/pages/ContentfulPage.vue
```diff
@@ -209,6 +215,8 @@ export default {
 				this.pageError = true;
 			} else {
 				this.title = (pageData?.page?.pageLayout?.pageTitle || pageData?.page?.pageTitle) ?? undefined;
+				this.headerTheme = siteThemes[pageData?.page?.pageLayout?.headerTheme] || {};
+				this.footerTheme = siteThemes[pageData?.page?.pageLayout?.footerTheme] || {};
```
mcstover, line 219: ⭐

mcstover (approved):

mcstover:
@emuvente My initial thought on this is that it needs to be at the "Page Layout" section – or should be available there too so it overrides the "Page" level declaration.

@ejensen @bryfox Wonder if there is a way to sync up the light/dark theme needs with this paradigm

eddieferrer (approved):

emuvente:
> My initial thought on this is that it needs to be at the "Page Layout" section

@mcstover Agreed! That's why I've only defined it at the "Page Layout" level

BoulderBrains (approved):

PR #2839, opened 2021-04-13
SUBS-669 Monthly Good Landing Page Hero

eddieferrer:
* Adds a wrapper component to the landing page route that passes attributes to the ContentfulPage component
* Adds ability for ContentfulPage component to pass these attributes down to all components generated from contentful
* Read attribute in HomepageHero component to determine if CTA button should go to a link or fire a function
______

**The problem:**
The new dynamic monthly good landing page, has basically the same hero as the homepage except the button should fire a function instead of go to a link. 

**The setup**
We can use a wrapper component, referenced from the route, that passes options to ContentfulPage, these options are then passed from ContentfulPage to all child components and can be used there. 

In this PR, the dynamic landing page and the homepage both have a hero component that is `homepageHero` type in contentful, and both use the same `HomepageHero` component in the code, but the action on clicking the button is different. 

Contentful page layout for reference (work in progress) - https://app.contentful.com/spaces/j0p9a6ql0rn7/environments/development/entries/3z8Yn1Inpy8iCZPY72pAS5

We'll also need to expand the options object to support 1 more modification (green styling on the button) but could support more options in the future. 

----

For the actual function for the landing page CTA, I imagine this can either go through an event bus, since it needs to communicate with the MonthlyGoodSelector component

----

I'm open to suggestions, this is a weird approach to a unique problem


eddieferrer:
I realize this limits how dynamic the route can be since it will always pull in the attrs in Monthly Good Page. An alternative approach could be to just hard code a dictionary of settings based on component key in ContentfulPage -- `If component key is homepage-hero-monthly-good, pass in these additional props or attrs to component`

emuvente:
> If component key is homepage-hero-monthly-good, pass in these additional props or attrs to component

Building on that idea, we could use a prop instead of hard-coding it. Since it would be a prop it could be defined in a wrapper component as well as in the router config.

For instance, the prop value could be:
```
{
    homepageHero: {
        customCtaFunction: () => alert('cta!'),
    },
    someOtherComponent: {
        propName: 'propValue',
    },
    ...
}
```
And then that could be used in a ContentfulPage method to get the options by type:
```
getComponentOptions(type) {
    return this.componentOptions[type] || {};
}
```
And then the template could bind those props to the component:
```
<component
    v-for="({ component, content }) in contentGroups"
    :key="content.key"
    :is="component"
    :content="content"
    v-bind="getComponentOptions(content.type)"
/>
```

eddieferrer:
> > If component key is homepage-hero-monthly-good, pass in these additional props or attrs to component
> 
> Building on that idea, we could use a prop instead of hard-coding it. Since it would be a prop it could be defined in a wrapper component as well as in the router config.
> 
> For instance, the prop value could be:
> 
> ```
> {
>     homepageHero: {
>         customCtaFunction: () => alert('cta!'),
>     },
>     someOtherComponent: {
>         propName: 'propValue',
>     },
>     ...
> }
> ```
> 
> And then that could be used in a ContentfulPage method to get the options by type:
> 
> ```
> getComponentOptions(type) {
>     return this.componentOptions[type] || {};
> }
> ```
> 
> And then the template could bind those options to the component as props:
> 
> ```
> <component
>     v-for="({ component, content }) in contentGroups"
>     :key="content.key"
>     :is="component"
>     :content="content"
>     v-bind="getComponentOptions(content.type)"
> />
> ```

I think this is a better approach, I'll clean this PR up

eddieferrer:
I made a cleaner version of this PR here: https://github.com/kiva/ui/pull/2842

PR #2842, opened 2021-04-14
SUBS-669 Added MG Dynamic Landing Page Hero

eddieferrer:
* Adds ability for ContentfulPage to pass down props into components
* Reads key for MG Hero and passes down options for cta function and button class
* Fix some event handling on MG Selector. Open selector when root event is fired from MG Hero

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -9,6 +9,7 @@
 				:key="content.key"
 				:is="component"
 				:content="content"
+				v-bind="getComponentOptions(content.key)"
 			/>
 		</template>
 		<template v-else>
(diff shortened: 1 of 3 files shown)


eddieferrer:
This is a cleaner PR of the things discussed here: https://github.com/kiva/ui/pull/2839

src/pages/ContentfulPage.vue
```diff
@@ -160,6 +161,21 @@ const getContentGroups = pageData => {
 	})).filter(group => typeof group.component === 'function');
 };
 
+const componentOptions = {
```
mcstover, line 164: While this is fine to start with, we might consider moving this to a utility file or mixin at some point...

mcstover: Potentially, even component specific...

mcstover (approved):
Very neat solution @eddieferrer + @emuvente

src/pages/ContentfulPage.vue
emuvente: Yeah I think this could also be passed as a prop to ContentfulPage, and then it can be defined by whatever page is actually using it.

emuvente (approved):
👍

PR #2843, opened 2021-04-14
Ml2-128 Opt out of UPC campaign

ryan-ludwig:


src/components/Checkout/VerifyRemovePromoCredit.vue
```diff
@@ -0,0 +1,112 @@
...
+	},
+	methods: {
+		removePromoCredit() {
+			if (this.activeCreditType) {
+				this.removeCredit(this.activeCreditType);
+			}
+			this.closeLightbox();
+		},
+		removeCredit(type) {
+			// TODO: Setup removing "promo_credit" type
```
mcstover, line 73: this is done and covered by the active promo type prop.

src/components/Checkout/VerifyRemovePromoCredit.vue
```diff
@@ -0,0 +1,112 @@
...
+				mutation: removeCreditByType,
+				variables: {
+					creditType: type
+				}
+			}).then(() => {
+				this.setUpdating(false);
+				this.$kvTrackEvent('basket', 'Kiva Credit', 'Remove Credit Success');
+				this.$emit('credit-removed');
+			}).catch(error => {
+				console.error(error);
```
mcstover, line 85: prob should convert to use logFormatter

mcstover (approved):

src/components/Checkout/VerifyRemovePromoCredit.vue
ryan-ludwig: Are you wanting to have all console.errors use the util in server/util/log.js ?  If so I'd like to move that to a different location and add a couple tests around it.

src/components/Checkout/OrderTotals.vue
```diff
@@ -276,44 +254,18 @@ export default {
 				this.setUpdating(false);
 			});
 		},
-		removeCredit(type) {
```
mcstover, line 279: @ryan-ludwig We need to add this method back in so that standard credit removal continues to function.

PR #2844, opened 2021-04-15
SUBS-670 Adds MG Info section to dynamic LP and selector fixes

eddieferrer:


src/components/MonthlyGood/MonthlyGoodSelector.vue
```diff
@@ -159,18 +159,11 @@ export default {
...
-			 */
-			document.getElementsByClassName('monthly-selector__button')[0].focus();
-			this.isCauseOpen = true;
-			this.isAmountOpen = false;
-		});
+		this.$root.$on('openMonthlyGoodSelector', this.showCausesRootEvent);
 	},
 	beforeDestroy() {
 		document.removeEventListener('keyup', this.onKeyUp);
+		this.$root.$off('openMonthlyGoodSelector', this.showCausesRootEvent);
```
eddieferrer, line 166: We should remove this listener on destroy

src/components/Homepage/HomepageMonthlyGoodInfo.vue
```diff
@@ -0,0 +1,155 @@
+<template>
```
eddieferrer, line 1: New component with new type. We should generalize this, so the icons can be modified via contentful.

src/components/MonthlyGood/MonthlyGoodSelectorWrapper.vue
```diff
@@ -74,12 +74,16 @@ export default {
...
 		},
 		initStickyBehavior() {
-			// Initial position down the page of element before mg selector element
-			// const { bottom } = this.$el.getElementsByClassName('monthly-good-info')[0].getBoundingClientRect();
+			// temporary set to not sticky to calculate these values based on original element position on page
+			this.isSticky = false;
+			this.mgStickBarOffset = 0;
+
+			// Initial position down the page of selector
+			const { top } = this.$el.getElementsByClassName('monthly-good-selector')[0].getBoundingClientRect();
```
eddieferrer, line 82: This used to be based on the positioning of the previous element. Made this so it doesnt depend on other elements. It's a bit tricky because on resize the sticky calculations should be based on where the non-sticky version of the element is on the page, not where it is currently (since it could be sticky when the window is resized)

src/components/MonthlyGood/MonthlyGoodSelectorWrapper.vue
```diff
@@ -145,275 +149,6 @@ export default {
 	}
 }
 
-.featured-loans,
```
eddieferrer, line 148: Forgot to remove all this CSS when I made this component. 🤦

src/components/Homepage/HomepageMonthlyGoodInfo.vue
mcstover: Yea we can do some moving around and renaming later to start isolating these"Content Block" type components.

mcstover (approved):

PR #2850, opened 2021-04-19
SUBS-671-694 Monthly Good LP Mobile MG Selector

eddieferrer:
* Implements new MG Cause/Amount selector for mobile web for MG LP
* Fix issue with cause/amount selection firing event listener for click outside

### DIFF
--- a/src/components/MonthlyGood/MonthlyGoodSelector.vue
+++ b/src/components/MonthlyGood/MonthlyGoodSelector.vue
@@ -30,7 +30,7 @@
 							'click-mgpromo-cause',
 							option.marketingName
 						]"
-						@click="selectCause(option)"
+						@click.stop="selectCause(option)"
 					>
 						<img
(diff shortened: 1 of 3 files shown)


src/components/MonthlyGood/MonthlyGoodSelector.vue
```diff
@@ -30,7 +30,7 @@
 							'click-mgpromo-cause',
 							option.marketingName
 						]"
-						@click="selectCause(option)"
+						@click.stop="selectCause(option)"
```
eddieferrer, line 33: Stops propagation and prevents the click-outside handler for firing since these elements are absolutely positioned and are technically 'outside'

mcstover (approved):
Look good, thank for hooking up a mobile view!
Def would be good to abstract out some of those option selections and data handling at a later date.

PR #2852, opened 2021-04-19
VUE-317 Deprecated Slot Syntax Migration

eddieferrer:
* Finished migrating all instances of deprecated slot syntax to new 2.6+ slot syntax

____

Finishes migration of old slot syntax. Three types of changes in all these files: 

Simple slot syntax change: 
```
-                               <template slot="controls">
+                               <template #controls>
```

In cases where a custom element is inserted in a slot, the custom element must be wrapped in a template tag:
```
-               <the-my-kiva-secondary-menu slot="secondary" />
+               <template #secondary>
+                       <the-my-kiva-secondary-menu />
+               </template>
```

For slots with scope, the replacement looks like
```
-                                                                                       <template slot="default" slot-scope="{ items }">
+                                                                                       <template #default="{ items }">
```

### DIFF
--- a/src/pages/LendingStats/LendingStatsPage.vue
+++ b/src/pages/LendingStats/LendingStatsPage.vue
@@ -1,8 +1,10 @@
 <template>
 	<www-page>
-		<the-my-kiva-secondary-menu slot="secondary" />
+		<template #secondary>
+			<the-my-kiva-secondary-menu />
+		</template>
 		<div class="row page-content">
-			<the-portfolio-tertiary-menu slot="tertiary" class="show-for-large" />
+			<the-portfolio-tertiary-menu class="show-for-large" />
 			<div class="small-12 large-9 columns">
 				<h1>Lending stats</h1>
 				<p>
--- a/.storybook/stories/KvCarousel.stories.js
+++ b/.storybook/stories/KvCarousel.stories.js
@@ -7,7 +7,7 @@ import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
 import causeList from '../mock-data/cause-selector-data-mock';
 
 const defaultCarouselSlides = `
-	<template slot="default">
+	<template #default>
 		<kv-carousel-slide style="background-color: green;">
 			<img src="https://via.placeholder.com/300x220/e8f0f6/000000" style="width: 100%;">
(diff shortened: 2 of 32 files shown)


src/pages/LendingStats/LendingStatsPage.vue
```diff
@@ -1,8 +1,10 @@
 <template>
 	<www-page>
-		<the-my-kiva-secondary-menu slot="secondary" />
+		<template #secondary>
+			<the-my-kiva-secondary-menu />
+		</template>
 		<div class="row page-content">
-			<the-portfolio-tertiary-menu slot="tertiary" class="show-for-large" />
+			<the-portfolio-tertiary-menu class="show-for-large" />
```
emuvente, line 7: Should this be
```
<template #tertiary>
	<the-portfolio-tertiary-menu class="show-for-large" />
</template>
```
?

eddieferrer: Technically yes. But that breaks the page. I don't believe that this page was actually using the tertiary slot on www-page. Probably because it was being rendered inside of the default slot. I don't think any component actually is, and it can actually be removed. 

The problem is that the slot on www-page is right above the default slot, so adding the menu as the first thing on the default slot displays it correctly. 

The correct syntax for lending stats should be:
```
		<template #secondary>
			...
		</template>
		<template #tertiary>
			...
		</template>
                <div class="row page-content">
                ...
                </div>
```

But what happens in that case is the tertiary slot, which has grid classes, gets displayed outside of page-content and breaks the page. 

The original markup:
```
	<the-my-kiva-secondary-menu slot="secondary" />
		<div class="row page-content">
			<the-portfolio-tertiary-menu slot="tertiary" class="show-for-large" />
                ...
                </div>

```
Was just rendering the secondary slot and the default slot. You can verify this  by seeing how the "row page-content" div is interweaved with the slots. 

You can see similar behavior in other pages that have a tertiary menu, they are just added to the default slot so they would render correctly (legal sub pages)

emuvente (approved):
👍 Looks great!

mcstover (approved):

BoulderBrains (approved):

ryan-ludwig (approved):

PR #2853, opened 2021-04-19
PSD-137: create test page for displaying flss queries

go-bears:
Jira Ticket:
https://kiva.atlassian.net/browse/PSD-137

What does it do:
sets up route, page and independent graphql queries for flss test demo page. 

How to test:
spinning up ui server should be able to navigate to 
`localhost:8888/flss`
then the page should render with loans from trending now sort from dev graphql endpoint

### DIFF
--- a/config/local.js
+++ b/config/local.js
@@ -5,13 +5,13 @@ module.exports = merge(base, {
 	app: {
 		host: 'localhost',
 		publicPath: '/',
-		graphqlUri: 'https://marketplace-api.dk1.kiva.org/graphql',
+		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
 		auth0: {
 			enable: false,
 		},
 	},
 	server: {
-		graphqlUri: 'https://marketplace-api.dk1.kiva.org/graphql',
+		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
 		sessionUri: '',
 		memcachedEnabled: false,
 		memcachedServers: '',
--- a/config/local.js
+++ b/config/local.js
@@ -5,13 +5,13 @@ module.exports = merge(base, {
 	app: {
 		host: 'localhost',
 		publicPath: '/',
-		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
+		graphqlUri: 'https://marketplace-api.dk1.kiva.org/graphql',
 		auth0: {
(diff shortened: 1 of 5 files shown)


src/pages/FlssPrototypes/queries/fragments/loanCardFields.graphql
```diff
@@ -0,0 +1,51 @@
+fragment loanCardFields on LoanBasic {
```
mcstover, line 1: No need duplicate this file.

src/pages/FlssPrototypes/queries/testLoanData.graphql
```diff
@@ -0,0 +1,34 @@
+#import './fragments/loanCardFields.graphql'
```
mcstover, line 1: Please move this file to `/src/graphql/query/` and rename to something like `flssQuery.graphql` and then it can just maintain the default path to the loanCardFields fragment.

go-bears: thanks @mcstover ! I was trying to keep the page queries localized and separate to flss project. thanks for letting know where the files can live.

go-bears:
btw, I think we're okay just merging into an `flss` branch that we keep updated to master. But if y'alls are good with merging into master, I can close this pr and open a new pr to merge into master as well.

config/local.js
```diff
@@ -5,13 +5,13 @@ module.exports = merge(base, {
...
 		host: 'localhost',
 		publicPath: '/',
-		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
+		graphqlUri: 'https://marketplace-api.dk1.kiva.org/graphql',
 		auth0: {
 			enable: false,
 		},
 	},
 	server: {
-		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
```
mcstover, line 14: please revert to prod as dev isn't accessible outside the vpn.

config/local.js
```diff
@@ -5,13 +5,13 @@ module.exports = merge(base, {
 	app: {
 		host: 'localhost',
 		publicPath: '/',
-		graphqlUri: 'https://marketplace-api.k1.kiva.org/graphql',
```
mcstover, line 8: same as below

BoulderBrains:
👀

PR #2855, opened 2021-04-19
SUBS-671 Bug Fix

eddieferrer:
* Fixes an issue where the hidden MG mobile selector, is opening a modal, which is scroll locking the body on desktop

### DIFF
--- a/src/components/MonthlyGood/MonthlyGoodSelectorWrapper.vue
+++ b/src/components/MonthlyGood/MonthlyGoodSelectorWrapper.vue
@@ -12,6 +12,7 @@
 		<!-- MG Selector Mobile -->
 		<section
 			class="monthly-good-selector section show-for-small-only"
+			v-if="isMobile"
 		>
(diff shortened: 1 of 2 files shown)


src/components/MonthlyGood/MonthlyGoodSelectorWrapper.vue
```diff
@@ -80,19 +82,26 @@ export default {
 			// set offset
 			this.mgStickBarOffset = offsetHeight;
 		},
+		determineIfMobile() {
+			this.isMobile = document.documentElement.clientWidth < 480;
```
eddieferrer, line 86: This  feels kind of hacky, but since both MG selectors, desktop and mobile need to listen to the root event from contentful, I need vue to actually remove it from the DOM with v-if instead of just hiding it with CSS utility classes. 

Possible opportunity for a mixin or utility function?

mcstover (approved):

PR #2858, opened 2021-04-20
SUBS-663 Dynamic MG LP route

eddieferrer:
* Fix for MG LP Landing page dynamic route and meta tag

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -4,7 +4,7 @@ export default [
 		name: 'homepage',
 		component: () => import('@/pages/Homepage/Homepage'),
 		meta: {
-			contentfulPage: 'home',
+			contentfulPage: () => 'home',
 		},
 	},
 	{ path: '/15', component: () => import('@/pages/15Years/15Years') },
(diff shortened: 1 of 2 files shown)


src/router/routes.js
```diff
@@ -101,9 +101,9 @@ export default [
 	{
 		path: '/lp/:dynamicRoute',
 		component: () => import('@/pages/ContentfulPage'),
-		meta: route => ({
```
eddieferrer, line 104: I guess you can't do this. 

This was casually working because if meta is undefined,  `contentfulPage` will actually make the pageQuery with an empty key, which will return all routes from contentful sorted by date added, and because this was the latest page on contentful it was being  rendered. But when a new route was added this morning (by mobile), it completely broke the landing page. 

Not sure if we should validate the meta param in contentfulPage to handle this...

mcstover: I think we should either provide a default OR probably just redirect to home if the key is empty.

mcstover (approved):

emuvente (approved):
👍 Looks good! Could you also please update the route example in the doc block on line 46 of ContentfulPage? thanks

PR #2859, opened 2021-04-20
GROW-586 GROW-585

emuvente:
This adds a new local resolver field to handle the kvu cookie being server-only. It works by using the kvu cookie to set a "hasEverLoggedIn" field in the apollo cache on the server, and then forces the browser to use the cached value for that field.

That new field is used to fix the issue in GROW-586 that was sending experiment events for users that were ineligible for guest checkout.

This also includes the requested change in GROW-585, "Checkout as guest" -> "Continue as guest".

### DIFF
--- a/src/api/localResolvers/my.js
+++ b/src/api/localResolvers/my.js
@@ -1,4 +1,6 @@
-export default ({ kvAuth0 }) => {
+import hasEverLoggedInQuery from '@/graphql/query/shared/hasEverLoggedIn.graphql';
+
+export default ({ cookieStore, kvAuth0 }) => {
 	return {
 		resolvers: {
 			My: {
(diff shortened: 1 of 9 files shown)


src/api/localResolvers/my.js
```diff
@@ -1,11 +1,34 @@
...
+export default ({ cookieStore, kvAuth0 }) => {
 	return {
 		resolvers: {
 			My: {
 				lastLoginTimestamp() {
 					return kvAuth0.getLastLogin();
 				}
 			},
+			Query: {
+				hasEverLoggedIn(obj, args, { cache }) {
```
mcstover, line 12: We should look out for any impact to page render speed with these new checks.

mcstover (approved):

BoulderBrains (approved):
Nice solution Shua.

PR #2863, opened 2021-04-22
SUBS-696 MG Menu Exp Mobile

eddieferrer:
* Adds monthly good menu exp to mobile
* Change menu exp button capitalization


src/components/WwwFrame/TheHeader.vue
```diff
@@ -409,7 +411,7 @@ export default {
 		PromoBannerLarge,
 		PromoBannerSmall,
 		SearchBar,
-		TheLendMenu: () => import('@/components/WwwFrame/LendMenu/TheLendMenu'),
```
eddieferrer, line 412: This dynamic import was breaking the exp wrapper component, which has a duplicate slot with the lend menu

mcstover: So now that will get bundled into the header bundle instead of being an on demand import?

eddieferrer: I believe thats true, but I am not sure what repercussions that will have. @emuvente might know more?

emuvente: Yeah, that's all that should happen. We previously used that dynamic import to delay the fetching of lend menu data, but that's handled by kv-dropdown events now

emuvente (approved):

mcstover (approved):

PR #2864, opened 2021-04-23
Grow 466

BoulderBrains:
[Grow-466](https://kiva.atlassian.net/browse/GROW-466) 

I've just added a tip message that reads "Recovery code copied" when a user clicks the Copy code button on during the MFA flow.

### DIFF
--- a/src/pages/Settings/RecoveryCodeConfirm.vue
+++ b/src/pages/Settings/RecoveryCodeConfirm.vue
@@ -69,6 +69,8 @@ export default {
 		copyToClipboard(string) {
 			try {
 				navigator.clipboard.writeText(string);
+				// Tip message triggered letting user know code successfully copied to clipboard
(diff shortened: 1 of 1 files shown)


src/pages/Settings/RecoveryCodeConfirm.vue
```diff
@@ -69,6 +69,8 @@ export default {
 		copyToClipboard(string) {
 			try {
 				navigator.clipboard.writeText(string);
+				// Tip message letting user know code successfully copied to clipboard
+				this.$showTipMsg('Recovery code copied.', 'confirm');
```
mcstover, line 73: This should work fine but you don't need the `confirm` param as it's not a valid type. You can just leave it off for the default style and icon to show.

ryan-ludwig (approved):

mcstover (approved):

emuvente (approved):
👍

PR #2866, opened 2021-04-23
SUBS-679 Rework

eddieferrer:
* Fixes issues with the lend menu experiment caused by duplicate refs.

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -188,17 +188,20 @@
 				</div>
 				<promo-banner-small :basket-state="basketState" />
 				<kv-dropdown
+					v-if="mgHighlightInNavVersion === 'shown'"
+					:controller="lendMenuId"
+					@show.once="loadLendInfo"
+				>
+					<monthly-good-exp-menu-wrapper ref="mgExpWrapper" />
+				</kv-dropdown>
+				<kv-dropdown
+					v-if="mgHighlightInNavVersion !== 'shown'"
 					:controller="lendMenuId"
 					@show.once="loadLendInfo"
 					@show="onLendMenuShow"
 					@hide="onLendMenuHide"
 				>
-					<monthly-good-exp-menu-wrapper v-if="mgHighlightInNavVersion === 'shown'">
-						<template #lendmenu>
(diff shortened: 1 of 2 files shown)


eddieferrer:
I had to basically completely refactor this. After some research it turns out Vue does pretty weird things when there are multiple elements with the same ref. I've been testing this approach and looks like it is working really well. 

I have also discovered a separate bug, verified that it exists in prod, and logged it as VUE-566

src/components/WwwFrame/TheHeader.vue
```diff
@@ -576,7 +575,11 @@ export default {
 			return route;
 		},
 		loadLendInfo() {
-			this.$refs.lendMenu.onLoad();
+			if (this.mgHighlightInNavVersion === 'shown') {
```
eddieferrer, line 578: This starts loading the exp "Borrower" menu when the first lend dropdown is shown, which leads to nice UI behavior since it starts loading the categories before the user has decided to hover over "Find a Borrower"

src/components/WwwFrame/TheHeader.vue
```diff
@@ -188,17 +188,20 @@
...
+				<kv-dropdown
+					v-if="mgHighlightInNavVersion !== 'shown'"
 					:controller="lendMenuId"
 					@show.once="loadLendInfo"
 					@show="onLendMenuShow"
 					@hide="onLendMenuHide"
 				>
-					<monthly-good-exp-menu-wrapper v-if="mgHighlightInNavVersion === 'shown'">
-						<template #lendmenu>
-							<the-lend-menu ref="lendMenu" />
```
eddieferrer, line 198: You can't have multiple refs "lendMenu" :(

src/components/WwwFrame/TheHeader.vue
```diff
@@ -411,7 +413,7 @@ export default {
 		PromoBannerLarge,
 		PromoBannerSmall,
 		SearchBar,
-		TheLendMenu,
+		TheLendMenu: () => import('@/components/WwwFrame/LendMenu/TheLendMenu'),
```
eddieferrer, line 416: I added this back in, it was not the cause of the issues as I had initially thought

mcstover (approved):

PR #2871, opened 2021-04-27
New frame and re-use of donation form for Kiva lender web donation pa…

mcstover:
…ge. Stubs out locations for FAQ + Donation status column.

### DIFF
--- a/src/pages/Donate/DonateForm.vue
+++ b/src/pages/Donate/DonateForm.vue
@@ -78,7 +78,6 @@ export default {
 				return {
 					category: 'Donate Form',
 					action: 'click-donate-support-us-form',
-					label: 'Donate',
 				};
 			},
 		},
--- a/src/pages/Donate/DonateForm.vue
+++ b/src/pages/Donate/DonateForm.vue
@@ -19,7 +19,12 @@
 				<kv-button class="smaller submit-btn" type="submit" :disabled="$v.$invalid">
 					{{ buttonText }}
 				</kv-button>
-				<div class="attribution-text text-center" v-html="formDisclaimer"></div>
+				<!-- Donation Disclaimer should always be present if we have a payment option active -->
+				<div
+					class="attribution-text text-center"
+					v-if="showDisclaimer || activateMonthlyOption"
+					v-html="formDisclaimerCopy"
+				></div>
 			</div>
 		</div>
 	</form>
@@ -51,13 +56,13 @@ export default {
 		};
 	},
 	props: {
+		activateMonthlyOption: {
+			type: Boolean,
+			default: false
+		},
 		buttonText: {
 			type: String,
-			default: 'Contribute monthly'
-		},
-		id: { // used when you have multiple instances of this form on one page.
-			type: String,
-			default: 'instance1',
+			default: 'Donate'
 		},
 		data: {
 			type: Array,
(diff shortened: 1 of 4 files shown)


src/router/routes.js
```diff
@@ -59,6 +59,7 @@ export default [
 	},
 	{ path: '/covid19response', component: () => import('@/pages/LandingPages/MGCovid19/MGCovid19') },
 	{ path: '/donate/support-kiva', component: () => import('@/pages/Donate/DonateFromMacro') },
+	{ path: '/donate/supportus-ui', component: () => import('@/pages/Donate/DonateSupportUs') },
```
mcstover, line 62: temp path until we reach parity, then we'll replace the existing route.

src/pages/Donate/DonateSupportUs.vue
```diff
@@ -0,0 +1,136 @@
...
+import gql from 'graphql-tag';
+
+import WwwPage from '@/components/WwwFrame/WwwPage';
+import { processPageContentFlat } from '@/util/contentfulUtils';
+import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
+import DonateForm from './DonateForm';
+
+const pageQuery = gql`query donateContent {
+	contentful {
+		entries (contentType: "page", contentKey: "web-donate-support-us")
```
eddieferrer, line 39: for almost all the other pages we're mapping these keys to the URL, any reason why this isn't `donate/supportus`  or `supportus` ?

src/pages/Donate/DonateForm.vue
```diff
@@ -66,11 +71,30 @@ export default {
...
 			default: '',
-		}
+		},
+		formSubmitAnalytics: {
+			type: Object,
+			default: () => {
+				return {
+					category: 'Donate Form',
+					action: 'click-donate-support-us-form',
+					label: 'Donate',
```
eddieferrer, line 81: I think you can remove the label property from here and the other components, since the analytics should always use buttonText for the analytics event

src/pages/Donate/DonateSupportUs.vue
mcstover: Over this is using a derivative of the url with a prefix for the intended platform. Generally, I don't think we should be using `/` in our keys or string identifiers.

This also aligns with the system we are using for mobile to prefix content for the apps.

eddieferrer: Ok sounds good, I wasn't aware we were using prefixes like that

src/pages/Donate/DonateForm.vue
mcstover: yea, I agree and waffled on that a bit... will add a default for the buttonText.

src/pages/Donate/DonateSupportUs.vue
mcstover: I'll admit the "app-" prefix is new as well but certainly a welcome addition now that Mobile is moving fully into setting up their own pages with their own content entries. You'll notice in contentful that ALL of the entries for this page are prefixed in a similar manner which is likely the proper paradigm to start following across all new content.

eddieferrer (approved):

BoulderBrains (approved):

emuvente (approved):
👍

PR #2877, opened 2021-04-28
SUBS-659 Monthly Good Page Login Flow

eddieferrer:
* Use hasEverLoggedIn to determine if the user should login before or after the MG setup form.


mcstover (approved):

emuvente (approved):
:+1: looks good!

PR #2882, opened 2021-04-29
SUBS-700 video in hero

eddieferrer:
I copied the styling (rounded corners, full width of containing column) from the "100% of your loan" video. Surprisingly, this looks great, specially if the video is in a vertical orientation. Also, contentful conveniently passes us the media file type so we can diferentiate videos and images. I made a test lp on dev at: `/lp/eddie-test` to see it in action. It does use the horizontal video from "100% of your loan"

--------

I moved the component out of Homepage, and changed some of the keys. My idea is to change the type on contentful and the component file name and keys to not reference homepage, to try to create a more generic component that isn't conceptually tied to the homepage. Does this idea make sense?


src/components/Contentful/ContentfulDynamicHero.vue
```diff
@@ -0,0 +1,229 @@
+<template>
```
eddieferrer, line 1: This is basically a copy of the current HomepageHero, with minor modifications

src/components/Contentful/ContentfulDynamicHero.vue
```diff
@@ -0,0 +1,229 @@
...
+		},
+		heroMedia() {
+			return {
+				description: this.mediaObject?.description ?? '',
+				title: this.mediaObject?.title ?? '',
+				url: this.mediaObject?.file?.url ?? ''
+			};
+		},
+		heroText() {
+			return this.content?.contents?.find(({ key }) => key.indexOf('homepage-hero-text') > -1);
```
eddieferrer, line 126: Should we change this to be more generic?

src/pages/ContentfulPage.vue
```diff
@@ -114,7 +116,8 @@ const getComponentFromType = type => {
 		case 'homepageBottomCTA':
 			return HomepageBottomCTA;
 		case 'homepageHero':
```
eddieferrer, line 118: Should we move this key to something more generic?

mcstover: I think so, now or later is fine though...

src/components/Contentful/ContentfulDynamicHero.vue
mcstover: Seems like we'd want this to remain as homepage specific....

src/components/Contentful/ContentfulDynamicHero.vue
```diff
@@ -0,0 +1,229 @@
...
+		<div class="row align-center">
+			<!-- hero media -->
+			<div
+				class="small-12 medium-10 large-6 xlarge-5 columns"
+			>
+				<template v-if="isHeroImage">
+					<router-link
+						:to="heroButton.link"
+						v-kv-track-event="[
+							'homepage',
```
mcstover, line 12: we'll need to dynamically derive these contexts...it may be helpful to have a util we can import that contains a map of urls to values we want to use in analytics events so we can make sure to match our event standards.

_There are some other contexts that might be better served with a prop of analytics data..._

src/components/Contentful/ContentfulDynamicHero.vue
emuvente: This is a Generic Content Block, not a Content Group, so we don't have a type field that we can key off of.

I think it would be fine to change this to be checking the key for just 'hero-text'.

src/pages/ContentfulPage.vue
emuvente: Yeah, something more generic for this would be good. Maybe 'dynamicHero'? And it might be better to make the change sooner while there aren't many existing Content Groups with that type to update.

ryan-ludwig (approved):
Nice!

BoulderBrains (approved):

emuvente (approved):
:+1:

mcstover (approved):

PR #2890, opened 2021-05-03
Add TheHeader styles to each pages css file.

mcstover:
This is a "blunt" solution to the fouc when our pages server render. It adds the header css to every page's generated `.css` file so keep an eye on the file size and Lighthouse reports.

PR #2892, opened 2021-05-03
VUE-577 Use vue-style-loader for all styles

emuvente:
Alternative to https://github.com/kiva/ui/pull/2890. This is also a "blunt" solution to fouc, though this sends all styles to vue-style-loader.

### DIFF
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -72,8 +72,8 @@ module.exports = {
 				],
 			},
 			{
-				// Inject styles from the /pages/ directory as <style> tags
-				test: /\/pages\/.+\.scss$/,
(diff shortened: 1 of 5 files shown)


BoulderBrains (approved):
🧹  👍

mcstover (approved):
Let's go with this. I like the other additional improvements it brings!

ryan-ludwig (approved):

eddieferrer (commented):
I believe this is the only use of the mini-css-extract plugin. We can remove it from package.json and also remove the cache exclusion from HardSourceWebpackPlugin `build/webpack.client.base.conf.js:66`

mcstover (approved):

PR #2896, opened 2021-05-04
SUBS-709 Adds ability for dynamic carousel in dynamic contentful hero

eddieferrer:


eddieferrer:
https://user-images.githubusercontent.com/4371888/117069430-4f1f4300-ace1-11eb-979b-1a915b28fb8f.mov

eddieferrer:
I wasn't sure which style of carousel, we wanted here, so went with mostly default options as seen in video

mcstover (approved):
Looks good!

Did this require any contentful updates or were you able to handle that in the code?

eddieferrer:
> Looks good!
> 
> Did this require any contentful updates or were you able to handle that in the code?

@mcstover no contentful updates needed

src/components/Contentful/DynamicHero.vue
```diff
@@ -79,28 +97,49 @@
 <script>
 import KvButton from '@/components/Kv/KvButton';
 import KvContentfulImg from '@/components/Kv/KvContentfulImg';
+import KvCarousel from '@/components/Kv/KvCarousel';
```
ryan-ludwig, line 100: Could these be dynamically imported depending on `isHeroCarousel` ?

eddieferrer: Yeah possibly? I can  do that in a followup PR. Are you concerned about loading embla and carousel when it's not needed?

ryan-ludwig: Yeah, exactly. Embla has some weight. More broadly working towards dynamically importing only the components in use when we have conditional logic like experiments and Contentful driven stuff.

ryan-ludwig: I also realize that Embla is currently loaded everywhere in our vendors bundle right now, but we should be able to sort that out separately.

PR #2898, opened 2021-05-05
grow-601 fixing image size,  and button styles

BoulderBrains:
[Grow-601](https://kiva.atlassian.net/browse/GROW-601)

Styles looked good on my VM previously, but once compiled on dev we had two small issues. 
Screen Shot 2021-05-04 at 2 50 22 PM

1) The appeal banner image wasn't filling the space provided for it. Display: flex; fixed that. 
2) Button styles were not being respected. Removing the smallest class on dev and moving that padding into the __btn definition seemed to do the trick. 

Here's how it looks fixed: 
Screen Shot 2021-05-04 at 5 32 25 PM


emuvente (approved):
:+1:

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBannerCircular.vue
```diff
@@ -344,6 +345,7 @@ export default {
 
 	&__progress-circle,
 	&__image {
+		display: flex;
```
ryan-ludwig, line 348: I believe we can remove this display:flex, if the image width and height are set correctly on line 29. I was seeing a width=164 and height=164 should work.

BoulderBrains: Updated in the latest commit, thanks for flagging this @ryan-ludwig.

ryan-ludwig (approved):
👍

BoulderBrains:
I keep missing my window to be allowed to merge this, before I have to re-merge master into it and wait for all the tests to pass again.

PR #2899, opened 2021-05-05
Setup new right column content driven by contentful. Includes re-use …

mcstover:
…of meter from the appeal banner.

### DIFF
--- a/src/pages/Donate/DonateSupportUs.vue
+++ b/src/pages/Donate/DonateSupportUs.vue
@@ -3,7 +3,7 @@
 		id="support-us-landing"
 	>
 		<div class="row page-content">
-			<div class="small-12 large-7 columns">
+			<div class="small-12 large-7 columns donation-form-holder">
 				<h1 class="donate-headline" v-html="headlineCopy"></h1>
 				<div class="donate-subhead" v-html="subHeadCopy"></div>
 				<donate-form
@@ -15,11 +15,10 @@
 					:show-disclaimer="false"
 				/>
 			</div>
-			<div class="small-12 large-5 columns">
-				<h4>Loans vs. Donations</h4>
-				<p>Kiva does not take a cut of loans, optional donations fund our operations.</p>
+			<div class="small-12 large-5 columns donation-meter-holder">
+				<donate-support-us-right-rail :content="donationCalloutsCG" />
 			</div>
-			<div class="small-12 columns donation-faq">
+			<div class="small-12 columns donation-faq donation-faq-holder">
 				<kv-frequently-asked-questions
 					:faqs-contentful="frequentlyAskedQuestions"
 					:headline="frequentlyAskedQuestionsHeadline"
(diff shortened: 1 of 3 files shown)


eddieferrer (commented):
The 'globe' image on contentful is 48x48, it should be uploaded at at least double that resolution to support the 2x needed for high density screens - As it is now it looks blurry

src/pages/Donate/DonateSupportUsRightRail.vue
```diff
@@ -0,0 +1,151 @@
...
+	},
+	data() {
+		return {};
+	},
+	computed: {
+		donationCallouts() {
+			const callouts = [];
+
+			const makeLoansHappenCopy = this.content?.contents?.find(
+				item => item.key === 'web-donate-support-us-make-loans-happen'
```
eddieferrer, line 56: I think I would prefer this key and the next key to be something a little more generic maybe `web-donate-support-us-callout-1` -- My thinking is that when this content changes the key will not be tied to the content

src/pages/Donate/DonateSupportUs.vue
```diff
@@ -15,11 +15,10 @@
...
 				/>
 			</div>
-			<div class="small-12 large-5 columns">
-				<h4>Loans vs. Donations</h4>
-				<p>Kiva does not take a cut of loans, optional donations fund our operations.</p>
+			<div class="small-12 large-5 columns donation-meter-holder">
+				<donate-support-us-right-rail :content="donationCalloutsCG" />
 			</div>
-			<div class="small-12 columns donation-faq">
+			<div class="small-12 columns donation-faq donation-faq-holder">
```
eddieferrer, line 21: This `donation-faq` class isn't used

src/pages/Donate/DonateSupportUsRightRail.vue
mcstover: Good idea!

src/pages/Donate/DonateSupportUs.vue
mcstover: > The 'globe' image on contentful is 48x48, it should be uploaded at at least double that resolution to support the 2x needed for high density screens - As it is now it looks blurry

Totally agree! Where did you get the larger versions? They weren't in the Figma files I have for this page. Either way it can just be swapped in contentful when we get the better version.

mcstover: Exported at 2x from Figma!

src/pages/Donate/DonateSupportUsRightRail.vue
mcstover: Updated

src/pages/Donate/DonateSupportUs.vue
mcstover: also removed unused class

eddieferrer (approved):

PR #2900, opened 2021-05-05
Dynamically import KvCarousel component

eddieferrer:
I think this will keep the Carousel component and its dependencies from loading for non Carousel use cases?

### DIFF
--- a/src/components/Contentful/DynamicHero.vue
+++ b/src/components/Contentful/DynamicHero.vue
@@ -6,7 +6,7 @@
 				class="small-12 medium-10 large-6 xlarge-5 columns"
 			>
 				<template v-if="isHeroCarousel">
-					<component :is="kvCarousel"
+					<kv-carousel
 						indicator-style="bar"
 					>
 						<kv-carousel-slide v-for="(image, index) in heroMedia" :key="index">
@@ -19,7 +19,7 @@
 								:alt="image.description"
 							/>
 						</kv-carousel-slide>
-					</component>
+					</kv-carousel>
 				</template>
 				<template v-if="isHeroImage">
 					<router-link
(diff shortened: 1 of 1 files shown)


src/components/Contentful/DynamicHero.vue
```diff
@@ -97,9 +95,7 @@
 <script>
 import KvButton from '@/components/Kv/KvButton';
 import KvContentfulImg from '@/components/Kv/KvContentfulImg';
-import KvCarousel from '@/components/Kv/KvCarousel';
```
ryan-ludwig, line 100: I think all of these imports can be removed, and you can set them as dynamic imports directly in the components object like
```
  components: {
    KvCarousel: () => import('@/components/Kv/KvCarousel')
    KvCarouselSlide: () => import('@/components/Kv/KvCarouselSlide')
    ...
  }
```

And then just use them regularly in the template without needing the data or `<component is="...>` stuff

https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components

I think we'd have to run the webpack bundle analyzer thing to verify. Or maybe we can just watch the network tab to see JS chunks come in?

ryan-ludwig: @emuvente Did you have a good testing method you used when doing the homepage dynamic import stuff?

eddieferrer: Yeah I wasn't unsure if these 2 approaches are the same

eddieferrer: Changed it to @ryan-ludwig suggestion above, since it's definitely more readable than `component is`

src/components/Contentful/DynamicHero.vue
```diff
@@ -5,22 +5,20 @@
 			<div
 				class="small-12 medium-10 large-6 xlarge-5 columns"
 			>
-				<template v-if="isHeroCarousel">
+				<template v-if="!isHeroCarousel">
```
ryan-ludwig, line 8: Oh, should this ! be here?

eddieferrer: 🤦  testing gone wrong

mcstover (approved):

mcstover:
Will be good to initially just verify that current functionality is unchanged, then we can move on to more rigorous testing of the new carousel.

PR #2901, opened 2021-05-05
GROW-577 Fix flag sprite loading on server by allowing sprite to be added to server bundle

emuvente:
By default, our server webpack config file "externalizes" all files in node_modules, meaning that instead of being included directly in the server bundle they are included as a `require` statement. That keeps the size of the server bundle low by allowing node to get the file from node_modules instead of the bundle. That's great for javascript, but it doesn't work for static files being included in css, because node does not know how to handle those static files.

This PR addresses the issue of the flag-icon-css module images being externalized, and we will need to make a similar change if we ever include other images from node_modules in our css in the future.


mcstover (approved):

ryan-ludwig (approved):

ryan-ludwig:
There are a couple of JavaScript helper functions in flag-icon-css that are used in a couple of places (getCodes, getNameByCode, getCountryList). Will these change deoptimize that JS?

ryan-ludwig:
If so, this could be set to `/flag-icon-css/flags/`to only externalize the image files

mcstover:
@ryan-ludwig @emuvente It may be the case that we can optimize further and we should def verify that flags are being generated properly by the scripts you mentioned.

Merged so we can get the dev env back in order and get on to branch cut.

PR #2906, opened 2021-05-06
SUBS-708 Background options for dynamicHero

eddieferrer:
* Adds background content type to contentful, which has an image and a color as content
* Use background content to change background on dynamic hero

### DIFF
--- a/src/components/Contentful/DynamicHero.vue
+++ b/src/components/Contentful/DynamicHero.vue
@@ -1,5 +1,5 @@
 <template>
-	<section class="contentful-hero">
+	<section class="contentful-hero" :style="heroBackgroundStyle">
 		<div class="row align-center">
 			<!-- hero media -->
 			<div
@@ -173,6 +173,24 @@ export default {
 				link: this.heroText?.primaryCtaLink ?? '',
 			};
 		},
+		heroBackground() {
+			return this.content?.contents?.find(({ key }) => key.indexOf('background') > -1);
+		},
+		heroBackgroundStyle() {
+			// Creates an object with conditional keys if heroBackground contains those elements
+			const styleObject = {
+				...(this.heroBackground?.backgroundColor && {
+					'background-color': this.heroBackground?.backgroundColor
+				}),
+				...(this.heroBackground?.backgroundImage?.file?.url && {
+					'background-image': `url("${this.heroBackground?.backgroundImage?.file?.url}")`
+				}),
+				...(this.heroBackground?.backgroundImage?.file?.url && {
+					'background-size': 'cover'
+				})
+			};
+			return styleObject;
+		}
 	},
 };
 </script>
--- a/src/util/contentfulUtils.js
+++ b/src/util/contentfulUtils.js
@@ -145,6 +145,22 @@ export function formatUiSetting(contentfulContent) {
 }
 
 /**
+ * Format Background (contentful type id: background)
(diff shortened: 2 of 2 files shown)


src/components/Contentful/DynamicHero.vue
```diff
@@ -173,6 +173,24 @@ export default {
...
+			// Creates an object with conditional keys if heroBackground contains those elements
+			const styleObject = {
+				...(this.heroBackground?.backgroundColor && {
+					'background-color': this.heroBackground?.backgroundColor
+				}),
+				...(this.heroBackground?.backgroundImage?.file?.url && {
+					'background-image': `url("${this.heroBackground?.backgroundImage?.file?.url}")`
+				}),
+				...(this.heroBackground?.backgroundImage?.file?.url && {
+					'background-size': 'cover'
```
eddieferrer, line 189: I think this is the default we want. It will span the background full width.

src/components/Contentful/DynamicHero.vue
```diff
@@ -173,6 +173,24 @@ export default {
...
+			return this.content?.contents?.find(({ key }) => key.indexOf('background') > -1);
+		},
+		heroBackgroundStyle() {
+			// Creates an object with conditional keys if heroBackground contains those elements
+			const styleObject = {
+				...(this.heroBackground?.backgroundColor && {
+					'background-color': this.heroBackground?.backgroundColor
+				}),
+				...(this.heroBackground?.backgroundImage?.file?.url && {
+					'background-image': `url("${this.heroBackground?.backgroundImage?.file?.url}")`
```
eddieferrer, line 186: The caveat here is that this is getting the full size image from contentful. Not sure if theres a way to optimize this.

src/components/Contentful/DynamicHero.vue
```diff
@@ -173,6 +173,24 @@ export default {
 				link: this.heroText?.primaryCtaLink ?? '',
 			};
 		},
+		heroBackground() {
+			return this.content?.contents?.find(({ key }) => key.indexOf('background') > -1);
+		},
+		heroBackgroundStyle() {
```
mcstover, line 179: I think we should make these 2 properties utility methods that can be called from computed properties in other components as well.

Something along these lines...

contentGroupBackground(contents)
// `contents` can be termed differently across contexts outside of these landing pages.
+
contentGroupBackgroundStyles(backgroundContent)

src/components/Contentful/DynamicHero.vue
mcstover: I think this is a good first version. Ultimately, this may be a use case for our KvContentfulImage or event a responsive image set...

src/components/Contentful/DynamicHero.vue
mcstover: I think this is good for the css background image initial version

ryan-ludwig:
An alternative to using CSS background images would be to use an `` with [object-fit: 'cover'](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) applied. You could set that image to a lower z-index than your content.  This would let you use kv-contentful-img with some reasonable sizing. I really don't trust people to upload sanely sized images to contentful.

ryan-ludwig:
`object-fit: 'cover'` is likely what you'll need for your video element

eddieferrer:
I agree with @ryan-ludwig above, I'll rework this to use `object-fit` instead of inline css

PR #2907, opened 2021-05-06
grow-601 appeal closed button state fixed by removing smallest class,…

BoulderBrains:
… fixing hover state of button when appeal is open

[Grow-601](https://kiva.atlassian.net/browse/GROW-601)

Found two more css errors in dev. 

1) open appeal button hover state. Text should remain kiva-green, it’s white in dev.
Screen Shot 2021-05-06 at 12 58 50 PM

2) Closed appeal - button had smallest class which was causing additional styles applied (ie the box-shadow here) 
Screen Shot 2021-05-06 at 12 58 36 PM


emuvente (approved):
👍

mcstover (approved):
good catch

ryan-ludwig (approved):

PR #2912, opened 2021-05-10
SUBS-708 Background options for dynamicHero

eddieferrer:
* Adds background content type to contentful, which has media and a color as content
* Use background content to change background on dynamic hero to be either a video or an image or a color


src/components/Contentful/DynamicHeroBackground.vue
```diff
@@ -0,0 +1,116 @@
+<template>
```
mcstover, line 1: Please rename so this is generic and doesn't allude to some dependency on the Hero component.

eddieferrer:
Ive tested this and it works in any full width "section" including multiple sections on the same page. Looks like: 

https://user-images.githubusercontent.com/4371888/117851087-d1a48700-b23a-11eb-8f6c-02bfd65659e0.mov

Screen Shot 2021-05-10 at 12 44 11 PM
Screen Shot 2021-05-10 at 12 44 50 PM

See my comments

src/components/Contentful/DynamicHero.vue
```diff
@@ -217,9 +232,9 @@ export default {
 		line-height: rem-calc(54);
 		font-weight: 500;
 
-		a,
-		em,
-		i {
+		::v-deep a,
```
eddieferrer, line 235: This v-deep is for styling the v-html from contentful

src/components/Contentful/SectionWithBackground.vue
```diff
@@ -0,0 +1,56 @@
+<template>
```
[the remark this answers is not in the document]
eddieferrer: > I could see them becoming 1 component as well.

👍

PR #2931, opened 2021-05-14
Update github actions to use node 15

eddieferrer:


mcstover (approved):
Thanks, I'm starting to test v16 locally now so we'll likely move to that pretty soon when it hits LTS.

PR #2933, opened 2021-05-14
Update helmet v4 CSP policy

eddieferrer:


mcstover (approved):
Let's dig in a little more and see what it would take to put this back in place.

PR #2950, opened 2021-05-25
VUE-519 VUE-567 Setup tailwind and kv-components

emuvente:
This adds tailwind and kv-components. This also sets up the purge css tailwind feature, to minimize the included classes. Even with that purging, I see an increase in css of ~10kb, which is included in each page load, although with gzip it's more like an increase of only 2.4kb.

To limit the effect of that, we could possibly make a separate page frame for tailwind-based styling, and move the tailwind include to that page frame instead of App.vue.


package.json
```diff
@@ -175,10 +176,11 @@
 		"jest-transform-graphql": "^2.1.0",
 		"memory-fs": "^0.5.0",
 		"ora": "^4.0.2",
-		"postcss": "^7.0.35",
-		"postcss-loader": "^3.0.0",
+		"postcss": "^8.3.0",
+		"postcss-loader": "^4.3.0",
 		"rimraf": "^3.0.2",
-		"sass-loader": "^8.0.2",
+		"sass": "^1.32.13",
```
emuvente, line 182: There's a more recent version of this, but the more recent version has deprecated using `/` for division, in favor of `math.div()`, and it generates several hundred deprecation warnings in the build log. I tried using the options to suppress those warnings, but the options did not seem to work.

src/components/Kv/KvCarousel.vue
```diff
@@ -434,7 +434,7 @@ $bar-indicator-margin: rem-calc(4);
 			&::after {
 				display: block;
 				content: '';
-				background: hsla(0, 0, 100, 0.5);
+				background: hsla(0, 0%, 100%, 0.5);
```
emuvente, line 437: Sass has deprecated using unit-less values for saturation and lightness

BoulderBrains:
👀

mcstover (approved):
Thanks for hooking this up and the extra style-loader cleanup too!

BoulderBrains (approved):

eddieferrer (approved):
👍  great to see the dependency updates too

ryan-ludwig (commented):
Nice! 👍

PR #2953, opened 2021-05-28
Grow-673 borrower profile frame

BoulderBrains:
[Grow-673](https://kiva.atlassian.net/browse/GROW-673)

Borrower profile page frame with tailwinds grid, margin & max-width across screen sizes. 

Storybook story included as well.


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+	<!-- Header goes here -->
```
mcstover, line 2: We just need to add the www-page component added to get the header and footer and all other global elements of the Ui app.

mcstover:
@emuvente @BoulderBrains I keep getting the same updates for package-lock.json so it may be fine to let them through here.

BoulderBrains:
@mcstover I reached out to @emuvente and cleared the package-lock.json changes earlier. They were a result of running 'npm ci', which I had to run in order to get storybook running again after merging the recent changes.

src/pages/BorrowerProfile/BorrowerProfile.vue
BoulderBrains: I removed  header and footer comments

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -0,0 +1,35 @@
+<template>
```
emuvente, line 1: Please add `www-page` as the root element (e.g. https://github.com/kiva/ui/blob/master/src/pages/BorrowerProfile/fundedBorrowerProfile.vue#L2). That will add the header and footer to the page.

eddieferrer:
> @emuvente @BoulderBrains I keep getting the same updates for package-lock.json so it may be fine to let them through here.

Same for me

src/pages/BorrowerProfile/BorrowerProfile.vue
BoulderBrains: Just added.

mcstover (approved):

BoulderBrains:
@mcstover, @emuvente & @eddieferrer, the package-lock.json file changes have been reverted from this PR.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -0,0 +1,47 @@
...
+	<www-page
+		id="borrower-profile"
+		:header-theme="headerTheme"
+		:footer-theme="footerTheme"
```
emuvente, line 5: Since we are using the default themes for this page, you don't need to include the header-theme and footer-theme props. Removing them will also take care of these warnings in the console:
```
[Vue warn]: Invalid prop: type check failed for prop "headerTheme". Expected Object, got String with value "".
[Vue warn]: Invalid prop: type check failed for prop "footerTheme". Expected Object, got String with value "".
```

emuvente (approved):
👍 looks great!

BoulderBrains:
One odd thing I've noticed in storybook for this new page frame is that there's an additional padding added to the 'sb-show-main sb-main-padded' which results in padding going around the header and footer. This same class is on other page files, but not respected the same way. 

Screen Shot 2021-06-01 at 10 21 33 AM

mcstover (approved):

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -0,0 +1,42 @@
...
+
+		Tailwinds classes used:
+		Medium screen:
+		max-w-4xl = max-width: 56rem;
+
+		Large screen:
+		max-w-6xl = max-width: 72rem;
+
+		-->
+		<div class="grid-cols-12 mx-5 md:mx-8 lg:mx-16 md:max-w-4xl lg:max-w-6xl flex-auto">
```
ryan-ludwig, line 23: I believe a wrapper `<div class="container">` will remove the need for these classes.

BoulderBrains: @ryan-ludwig Does the container class have the tailwinds grid, max-widths and margins set on it? Trying to better understand how this is all coming together.

ryan-ludwig (approved):
👍

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+	<www-page id="borrower-profile">
+		<div class="container">
+			<div class="grid grid-cols-12 gap-x-2">
+				<div class="col-span-12 md:col-start-2 md:col-span-10 lg:col-span-7"
+					style="background-color:lightgreen;"
```
mcstover, line 6: are these just for debugging + temporary grid visibility?

BoulderBrains:
I've reworked how the grid is applied. Adding the .container div gave me max-width and margin definitions. Within that container i've added three divs with tailwinds grids defined. To the two inner divs, there's an inline style so you can see the grid breaking across screen sizes, will remove before merging.

src/pages/BorrowerProfile/BorrowerProfile.vue
BoulderBrains: I put them there, so you can see the grid breaking across screen sizes if you pulled it up in storybook. I'll remove the inline style before merging.

BoulderBrains:
> One odd thing I've noticed in storybook for this new page frame is that there's an additional padding added to the 'sb-show-main sb-main-padded' which results in padding going around the header and footer. This same class is on other page files, but not respected the same way.
> 
> **New page frame**
> Screen Shot 2021-06-01 at 10 21 33 AM
> 
> **Another page with same elements, css is there but not respected the same way**
> Screen Shot 2021-06-01 at 10 37 35 AM
> 
> * There is a margin: -1rem; on the < div > directly inside the < div id="root" > which corrects this padding for this page.

@ryan-ludwig what do you think about this situation I'm still seeing in storybook?

ryan-ludwig:
@BoulderBrains For the margin/fullscreen issue, you can add a layout parameter of 'fullscreen' to the story to get it to go to the edges. https://storybook.js.org/docs/react/configure/story-layout

Something like
```
// BorrowerProfile.stories.js

export default {
	title: '/Pages/Borrower Profile',
	component: BorrowerProfile,
	parameters: {
		layout: 'fullscreen',
	},
...
```

It wouldn't be a good idea to do that in all of the "Pages" stories, but we can come back around to it after this PR.

mcstover (approved):

PR #2956, opened 2021-05-28
Turn on production mode but exclude bundle analyzer.

mcstover:


mcstover:
This looks like it will help ease the memory consumption during the e2e runs.

emuvente (approved):
👍

PR #2958, opened 2021-06-01
build: package-lock updates that we removed from borrower profile frame build

BoulderBrains:
These are the package-lock.json updates that I removed from [this PR.](https://github.com/kiva/ui/pull/2953)

These updates were a result of running npm install in the /kiva/ui repo.


mcstover (approved):
looks familiar ;)

emuvente (approved):

PR #2963, opened 2021-06-03
GROW-659 Borrower name component

mlibazisi:
Ticket: GROW-659

### DIFF
--- a/src/components/BorrowerProfile/BorrowerName.vue
+++ b/src/components/BorrowerProfile/BorrowerName.vue
@@ -5,40 +5,13 @@
 </template>
 
 <script>
-import gql from 'graphql-tag';
 
 export default {
-	inject: ['apollo', 'cookieStore'],
-	data() {
-		return {
-			name: '',
-		};
-	},
-	apollo: {
-		query: gql`
-			query borrowerName($loanId: Int!) {
-				lend {
-					loan(id: $loanId) {
-						id
-						name
-					}
-				}
-			}
-		`,
-		preFetch: true,
-		preFetchVariables({ route }) {
-			return {
-				loanId: route?.params?.id ?? 0,
-			};
-		},
-		variables() {
-			return {
-				loanId: this.$route?.params?.id ?? 0,
-			};
-		},
-		result({ data }) {
-			this.name = data?.lend?.loan?.name ?? '';
-		},
-	},
+	props: {
+		name: {
+			type: String,
+			default: ''
+		}
+	}
 };
 </script>
--- a/src/components/BorrowerProfile/BorrowerName.vue
+++ b/src/components/BorrowerProfile/BorrowerName.vue
@@ -1,5 +1,7 @@
 <template>
-	<h1 class="text-h2">{{ name }}</h1>
+	<h1 class="text-h2">
+		{{ name }}
+	</h1>
 </template>
 
 <script>
--- a/src/components/BorrowerProfile/BorrowerName.vue
+++ b/src/components/BorrowerProfile/BorrowerName.vue
@@ -1,5 +1,5 @@
 <template>
-	<div class="borrower-name-wrapper">
+	<div class="borrower-name-wrapper text-gray-800">
 		<h2>{{ name }}</h2>
 	</div>
 </template>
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/BorrowerName.vue
```diff
@@ -0,0 +1,38 @@
...
+@import 'settings';
+
+.borrower-name-wrapper {
+	position: absolute;
+	width: 160px;
+	height: 48px;
+	left: 336px;
+	top: 188px;
+	color: $umbreon;
+
```
BoulderBrains, line 28: @mlibazisi for these classes we need to use the [Tailwinds css classes](https://tailwindcss.com/docs). We don't want any scss in this file. 

for example the color css above would be handled with this tailwind class "text-gray-800"

BoulderBrains: The same is true for all of these other css definitions, you need to find tailwind specific classes and delete all css definitions in this file.

mlibazisi: 👍

ryan-ludwig (approved):

src/components/BorrowerProfile/BorrowerName.vue
```diff
@@ -0,0 +1,44 @@
...
+import gql from 'graphql-tag';
+
+export default {
+	inject: ['apollo', 'cookieStore'],
+	data() {
+		return {
+			name: '',
+		};
+	},
+	apollo: {
```
mcstover, line 17: I don't think this should be in this component. The name shouldn't need it's own direct query.

emuvente: Could you talk through more of your reasoning for that? My thinking is that we have each component responsible for its own data. That way the components can be loosely coupled, we can selectively control what is and isn't prefetched, and if a component needs to be changed to include extra data that change only needs to happen in the component itself and no changes are needed in the wrapping component. I could see it maybe being overkill for the borrower name,  as it's a single data point, but I think in general we should have each component fetch its own data (especially things like the fundraising progress component, which will need several data points).

emuvente: @mcstover How about this:

- in a separate PR we make a "SummaryCard" component for the borrower profile that handles the data fetching and display of the avatar image, loan use, name, fundraising progress, and location/sector
- if we make child components for the SummaryCard, they just use props
- for this PR we change this component to accept name as a prop

I think that will allow us to not need this apollo boilerplate code for just the name, while also maintaining a loose coupling between most of the components on the page.

My main goal is to avoid a situation where we are passing around a single large loan object to every component on the page that only need one piece of data from the loan.

mcstover: I think the strategy you described above works for some collections of components/data but we should strike a better balance between significantly increasing the number of queries needed to bootstrap the page vs the independent data fetch.

Generally, the BP card section should prob have 1 query that gets the name, image, location, description + tag then a second query for the loan fundraising info + meter section.

This is effectively 7 vs 2 queries and we should consider the additional overhead that comes with the prefetch subscription which will run and update the data 2x for each query.

As for moving toward the more independently modular approach, I'd like to see some ways to remove the level of boilerplate required to bootstrap this. In this case that could mean passing the query and param values which currently depend on a specific id in the url which reduces the reusability.

mcstover: That sounds good @emuvente !!

emuvente: @mcstover Makes sense, and sounds good!
@mlibazisi sorry for the churn on this, could you please change this component back to accepting the name as a prop?

mlibazisi: 👍

emuvente (approved):
👍 looks good!

src/components/BorrowerProfile/BorrowerName.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+	<h1 class="text-h2">
+		{{ name }}
```
BoulderBrains, line 3: Should this just be an <h2> ?

mlibazisi: @BoulderBrains In an earlier discussion with @emuvente and @ryan-ludwig, it was recommended that i do an h1 with a class of h2

mcstover (approved):

PR #2966, opened 2021-06-04
build: Loan use component for borrower profile

BoulderBrains:
[grow-619](https://kiva.atlassian.net/browse/GROW-619) Loan use statement for borrower profile. 

Component that accepts some props and returns the formatted loan use (up to 100 characters length) based on the status of the loan.

### DIFF
--- a/src/components/BorrowerProfile/LoanUse.vue
+++ b/src/components/BorrowerProfile/LoanUse.vue
@@ -33,7 +33,7 @@ export default {
 	computed: {
 		loanUseFiltered() {
 			// eslint-disable-next-line max-len
-			return loanUseFilter(this.use, this.name, this.status, this.loanAmount, this.borrowerCount, 100);
+			return loanUseFilter(this.use, this.name, this.status, this.loanAmount, this.borrowerCount);
 		},
 	}
 };
--- a/src/components/BorrowerProfile/LoanUse.vue
+++ b/src/components/BorrowerProfile/LoanUse.vue
@@ -1,0 +1,43 @@
+<template>
+	<div>
+		<span class="text-h2 text-gray-800">
+			{{ filteredLoanUse }}
+		</span>
+	</div>
+</template>
+
+<script>
+const loanUseFilter = require('../../plugins/loan-use-filter');
+
+export default {
+	props: {
+		loanAmount: {
+			type: String,
+			default: '',
+		},
+		use: {
+			type: String,
+			default: '',
+		},
+		status: {
+			type: String,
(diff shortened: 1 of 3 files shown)


src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -0,0 +1,41 @@
...
+		},
+		name: {
+			type: String,
+			default: '',
+		},
+		borrowerCount: {
+			type: Number,
+			default: 1,
+		}
+	},
```
mlibazisi, line 32: Just wondering, is there a reason you decided to passing in multiple props, instead of just the loan object?

src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+	<span class="text-h2 text-gray-800">
```
ryan-ludwig, line 2: I'd recommend a <p> tag here as this is looks to be a.. paragraph, while <span> has no semantic meaning.

Minor, but .text-gray-800 is not needed in this case as that is the default color for all of our text.

src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -0,0 +1,41 @@
...
+		},
+		borrowerCount: {
+			type: Number,
+			default: 1,
+		}
+	},
+	computed: {
+		loanUseFiltered() {
+			// eslint-disable-next-line max-len
+			return loanUseFilter(this.use, this.name, this.status, this.loanAmount, this.borrowerCount, 100);
```
emuvente, line 36: Please use a max length of 1000, so that the loan use statement does not get trimmed. Alternatively, you could not give a max length value and update the loan use filter to not trim the loan use when `loanUseMaxLength` is 0.

BoulderBrains: Thanks Shua, I've made this update in my latest commit.

emuvente: Looks good! You'll also need to update the test for loanUseFilter, specifically [this line](https://github.com/kiva/ui/blob/master/test/unit/specs/plugins/loanUseFilter.spec.js#L47), so that it expects the full loan use message when a max length is not provided.

mcstover (approved):

emuvente (approved):
👍 looks great!

mlibazisi (approved):
👍

ryan-ludwig (approved):

PR #2967, opened 2021-06-04
Update webpack-cli and any cli usage of webpack to use new params and…

mcstover:
… flags.


emuvente (approved):
Those different options for webpack all look good to me!

mcstover:
I verified all of utils that use the cli as well as our more commonly used scripts (build, start, dev, storybook, lint, test, e2e etc) even though they don't use the cli and didn't see any issues. Of course running through actual ci will be the real test ;)

PR #2972, opened 2021-06-08
Vue 749 responsive adaptive dynamic hero

mcstover:
Depends on (includes: https://github.com/kiva/ui/pull/2970)

### DIFF
--- a/src/components/Contentful/DynamicHero.vue
+++ b/src/components/Contentful/DynamicHero.vue
@@ -32,7 +32,7 @@
 								heroMedia[0].description,
 							]"
 						>
-							<kv-contentful-img-new
+							<kv-contentful-img
 								v-if="heroMedia[0].url"
 								class="contentful-hero__img"
 								:contentful-src="heroMedia[0].url"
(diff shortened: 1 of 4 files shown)


emuvente (approved):

BoulderBrains (approved):

src/components/Contentful/DynamicHero.vue
```diff
@@ -32,7 +32,7 @@
 								heroMedia[0].description,
 							]"
 						>
-							<kv-contentful-img
+							<kv-contentful-img-new
```
ryan-ludwig, line 35: Should this be `kv-contentful-img` ?

ryan-ludwig: (throughout)

ryan-ludwig (approved):

src/components/Contentful/DynamicHero.vue
mcstover: Well the `-new` represents sections that are using kv-components instead of our existing Ui based version.

I think I can go ahead with a full migration to use the kv-components version. It worked fine for the single image case which I've tested and I'll test the carousel too and get that using it.

PR #2977, opened 2021-06-09
Use contentType to find background and check for contentType before u…

mcstover:
…sing it for comparison.

### DIFF
--- a/src/components/Contentful/DynamicHero.vue
+++ b/src/components/Contentful/DynamicHero.vue
@@ -235,8 +235,9 @@ export default {
 			};
 		},
 		heroBackground() {
-			// TODO: use the contentType param to target this content instead of a string fragment in the key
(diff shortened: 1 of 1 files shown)


src/components/Contentful/DynamicHero.vue
```diff
@@ -235,8 +235,9 @@ export default {
 			};
 		},
 		heroBackground() {
-			// TODO: use the contentType param to target this content instead of a string fragment in the key
-			return this.content?.contents?.find(({ key }) => key.indexOf('background') > -1);
+			return this.content?.contents?.find(({ contentType }) => {
```
mcstover, line 238: There's currently an issue in dev wherein we're not checking for `key` as it's not initially present so it's breaking the page.

I don't think the contentType can be undefined but am checking anyway...

PR #2980, opened 2021-06-09
Wrap how it works component with section background and tweak content…

mcstover:
… layout when there are varied numbers of items.

### DIFF
--- a/src/components/Homepage/HomepageHowItWorks.vue
+++ b/src/components/Homepage/HomepageHowItWorks.vue
@@ -1,49 +1,55 @@
 <template>
-	<section class="how-it-works section text-center">
-		<div class="row">
-			<p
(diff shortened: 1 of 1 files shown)


src/components/Homepage/HomepageHowItWorks.vue
```diff
@@ -86,10 +97,9 @@ export default {
 @import 'settings';
 
 .how-it-works {
-	padding: 2rem 0;
-
 	&__list {
 		list-style: none;
+		justify-content: space-around;
```
mcstover, line 102: This ensures that if we have fewer than 4 list items they distribute nicely when not stacked.

ryan-ludwig (approved):

BoulderBrains (approved):

PR #2981, opened 2021-06-09
GROW-629 LoanProgress component

emuvente:


src/api/localResolvers/loan.js
```diff
@@ -0,0 +1,166 @@
...
+			LoanPartner: {
+				fundraisingPercent,
+				fundraisingTimeLeft,
+				unreservedAmount,
+			},
+			LoanDirect: {
+				fundraisingPercent,
+				fundraisingTimeLeft,
+				unreservedAmount,
+			},
```
mcstover, line 163: This is coming together nicely. I'm assuming these local resolver fields will expand with the page so it would be nice  move the plain functions that power them out into a loan util file at some point.

mcstover (approved):

mlibazisi (approved):
👍

BoulderBrains (approved):

PR #2988, opened 2021-06-10
Feature grow 634 loan description

mlibazisi:
Ticket: GROW-634


src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -0,0 +1,118 @@
+<template>
+	<section>
+		<p
+			class="tw-text-h2"
+		>
+			More about this loan
```
BoulderBrains, line 6: Should this be "{{ borrowerName }}'s story"?

mlibazisi: You're right, I was looking at the older sigma file. Will update!

src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -0,0 +1,122 @@
...
+		loanAlertTextParagraphs() {
+			return this.toParagraphs(this.loanAlertText);
+		},
+		businessDescriptionParagraphs() {
+			return this.toParagraphs(this.businessDescription);
+		},
+		purposeParagraphs() {
+			return this.toParagraphs(this.purpose);
+		},
+		toParagraphs(text) {
```
emuvente, line 116: This should go in `methods` instead of `computed`

emuvente (commented):
I think the `<p>` tags that have `text-h2` and `text-h3` classes could actually be changed into `<h2>` and `<h3>` tags, as they are actually headings instead of paragraphs.

Other than that, this looks good!

BoulderBrains (approved):

PR #3006, opened 2021-06-16
Configurations to make vm to mac dev and option.

mcstover:


config/dev-vm-mac.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = merge(base, devVm, {
+	app: {
+		graphqlUri: 'https://api-vm.kiva.org/fed/graphql',
+	},
+	server: {
+		graphqlUri: 'https://api-vm.kiva.org/fed/graphql',
+		// memcachedEnabled: false,
+		memcachedEnabled: true,
+		memcachedServers: 'localhost:11211',
```
mcstover, line 13: This resolves to memcached running in docker. Verify from with `nc localhost 11211`. `stats` to inspect. `quit` to exit.

server/dev-server.js
```diff
@@ -3,6 +3,7 @@ require('../build/check-versions')();
 
 // dependencies
 require('dotenv').config({ path: '/etc/kiva-ui-server/config.env' });
+require('dotenv').config({ path: './config.env' });
```
mcstover, line 6: Multiple requires here are fine, `/etc...` is protected on the mac so the app couldn't reach it without a lot of mucking about with permissions that might affect other things.

.gitignore
```diff
@@ -15,3 +15,5 @@ build/schema.graphql
 *.njsproj
 *.sln
 .vscode
+.env
+.config.env
```
mcstover, line 19: So we don't commit them 🤦‍♂️

ryan-ludwig (approved):

emuvente (approved):

PR #3011, opened 2021-06-16
GROW-639 Add loan details component

mlibazisi:


emuvente (approved):
This looks great! One improvement I would suggest is making a generic component for a single description item, so that it can be used for the field partner and trustee components without having to copy all the classes.

Something like:
```
<template>
	<div class="tw-flex tw-mb-2">
		<dt class="tw-flex-1">
			<button class="tw-underline">
				{{ term }}
			</button>
		</dt>
		<dd>{{ details }}</dd>
	</div>
</template>
```

emuvente (approved):
:+1:

PR #3013, opened 2021-06-17
Feature grow 636 more about this loan

mlibazisi:
covers grow 636 and grow 634

### DIFF
--- a/src/components/BorrowerProfile/LoanDescription.vue
+++ b/src/components/BorrowerProfile/LoanDescription.vue
@@ -18,7 +18,7 @@
 			<section v-if="storyDescription">
 				<p
 					v-for="(paragraph, index) in storyDescriptionParagraphs"
-					:key="index"
+					:key="`storyDescription-${index}`"
 					class="tw-text-base"
 				>
 					{{ paragraph }}
@@ -57,7 +57,7 @@
 				<div>
 					<p
 						v-for="(paragraph, index) in descriptionInOriginalLanguageParagraphs"
-						:key="index"
+						:key="`originalLanguageParagraph-${index}`"
 						class="tw-text-base"
 					>
 						{{ paragraph }}
--- a/src/components/BorrowerProfile/LoanDescription.vue
+++ b/src/components/BorrowerProfile/LoanDescription.vue
@@ -38,7 +38,12 @@
 			</section>
 
 			<section v-if="storyTranslation">
-				<img v-if="reviewerImageLink" src="{{ reviewerImageLink }}" alt="Translator profile picture">
+				<img
+					loading="lazy"
+					v-if="reviewerImageLink"
+					:src="reviewerImageLink"
+					alt="Translator profile picture"
+				>
 				<p class="tw-text-base">
 					Translated from {{ language }}
 					<span v-if="reviewerName">
--- a/src/components/BorrowerProfile/LoanDescription.vue
+++ b/src/components/BorrowerProfile/LoanDescription.vue
@@ -33,14 +33,23 @@
 
 			<section>
 				<!--
-				Previous Loan sections omitted. Needs separate ticket
+				"Previous Loan sections" omitted. Needs separate ticket
 				-->
 			</section>
 
 			<section v-if="storyTranslation">
-				<img v-if="reviewerImageLink" src="{{ reviewerImageLink }}" alt="Translator profile picture"/>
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -0,0 +1,120 @@
...
+		businessDescriptionParagraphs() {
+			return this.toParagraphs(this.businessDescription);
+		},
+		purposeParagraphs() {
+			return this.toParagraphs(this.purpose);
+		}
+	},
+	methods: {
+		toParagraphs(text) {
+			return String(text).split(/(\r\n\r\n|\n\n|\n \n|<br>)/);
```
emuvente, line 115: I would recommend using `/(\r|\n|<br\s*\/?>)+/` for the regex pattern (here and in LoanDescription), as that will match any series of returns, new lines, and line break elements.

src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -1,118 +1,149 @@
...
-			v-if="partnerName"
-		>
-			<div>
-				{{ moreInfoAboutLoan }}
-			</div>
-			<div
-				v-if="loanAlertText"
+		<div>
+			<h2
+				class="tw-text-h2"
```
emuvente, line 5: Since this is already an h2, I think you don't need to add the `tw-text-h2` class. The same should go for the other h2 and h3 elements in this pr, and the `tw-text-base` class on the p elements.

emuvente (approved):
A couple small comments, but overall looks great!

src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -1,118 +1,149 @@
...
+			>
+				{{ borrowerOrGroupName }}'s story
+			</h2>
+		</div>
+		<div>
+			<section v-if="storyDescription">
 				<p
-					v-for="(paragraph, index) in loanAlertTextParagraphs"
+					v-for="(paragraph, index) in storyDescriptionParagraphs"
 					:key="index"
```
ryan-ludwig, line 21: There's a possibility the key of `index` will conflict on other loops within this page if they also use `index` for their key.
I'd recommend prefixing the index with a string.
Something like
```:key="`storyDescription-${index}`"```

File: src/components/BorrowerProfile/LoanDescription.vue [hunk 4]
```diff
@@ -1,118 +1,149 @@
...
+						</a> {{ reviewerName }}
+					</span>.
+					<a>
+						View original language description.
+					</a>
+				</p>
+				<div>
+					<p
+						v-for="(paragraph, index) in descriptionInOriginalLanguageParagraphs"
+						:key="index"
```
ryan-ludwig, line 60: Similar thing here as the index issue above.
maybe ```:key="`originalLanguageParagraph-${index}`"```

src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -0,0 +1,120 @@
...
+				v-if="loanAlertText"
+			>
+				<h3
+					class="tw-text-h3"
+				>
+					About {{ partnerName }}:
+				</h3>
+				<p
+					v-for="(paragraph, index) in loanAlertTextParagraphs"
+					:key="index"
```
ryan-ludwig, line 24: Same index stuff throughout this file

ryan-ludwig:
I did some more reading on the `:key="index"` thing. It is probably fine because those loops have different parent elements, so Vue figures it out.  Still a good practice to prefix those names, but should be all good here. 👍

src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -1,118 +1,149 @@
...
+			</section>
+
+			<section>
+				<!--
+				"Previous Loan sections" omitted. Needs separate ticket
+				-->
+			</section>
+
+			<section v-if="storyTranslation">
+				<img v-if="reviewerImageLink" src="{{ reviewerImageLink }}" alt="Translator profile picture">
```
ryan-ludwig, line 41: I believe the src attribute will need to be changed to
`<img :src="reviewerImageLink" ...`

ryan-ludwig, line 41: Suggest adding a lazy loading attribute
`<img loading="lazy" ...`
since this image is further down the page

ryan-ludwig (approved):
Looks great!

PR #3017, opened 2021-06-18
Homepage contentful block - image scaling

ryan-ludwig:
Currently the adaptive images on the homepage block are being served at the dimensions they're uploaded at in contentful. This PR scales them down to their container size at the different breakpoints.


mcstover (approved):
Good catch!

PR #3018, opened 2021-06-18
GROW-674 Borrower loan description image

emuvente:


src/components/BorrowerProfile/BorrowerImage.vue
```diff
@@ -0,0 +1,136 @@
+<template>
+	<picture
+		class="
+			tw-inline-block
+			tw-relative
+			tw-overflow-hidden
+		"
+		:style="`padding-bottom: ${aspectRatio * 100}%;`"
```
mcstover, line 8: Does this manifest as actual padding below the image?

mcstover (approved):

src/components/BorrowerProfile/BorrowerImage.vue
mcstover: Seems like more of an offset technique. I was just wondering about reusability on something like the loan card...

mlibazisi (approved):
👍

src/components/BorrowerProfile/BorrowerImage.vue
emuvente: Yeah, it's something we've used on loan cards before (see https://github.com/kiva/ui/blob/master/src/components/LoanCards/LoanCardImage.vue#L112-L117). The element with the padding is position relative, and the image it contains is position absolute. The padding value itself is the aspect ratio of the image, as a percent, and what that gives you is a responsive image placeholder (for while the image is loading) that automatically matches the expected height of the image no matter what width it is displayed at.

src/components/BorrowerProfile/BorrowerImage.vue
```diff
@@ -0,0 +1,136 @@
...
+		<img
+			v-if="hash"
+			class="
+				tw-absolute
+				tw-w-full
+				tw-h-full
+				tw-object-contain
+			"
+			:src="defaultUrl"
+			:alt="alt"
```
ryan-ludwig, line 20: a `loading="lazy"` attribute or `:loading="loading"` and using a prop would be good here.

ryan-ludwig (approved):
Really nice work!

PR #3019, opened 2021-06-18
use ignoreDuplicateBuildFailure in dev too

mcstover:


mcstover:
Prod runs were failing with the same OOM issue.

ryan-ludwig (approved):

PR #3020, opened 2021-06-18
Add loan description to full page & related minor fixes

emuvente:


mcstover (approved):

mlibazisi (approved):
👍

src/components/BorrowerProfile/LoanDescription.vue
```diff
@@ -42,7 +42,7 @@
 					loading="lazy"
 					v-if="reviewerImageLink"
 					:src="reviewerImageLink"
-					alt="Translator profile picture"
+					:alt="reviewerName"
```
ryan-ludwig, line 45: Is the width/height of this image constant? If it's always square, having width=100 height=100 to indicate the aspect ratio will prevent layout shift

emuvente: That's something we'll need to ask design about, as I'm not sure if the image is actually desired here. My plan is to find a loan with a reviewer that has an image, and then share the new profile page for that loan with the team to see what changes are desired. After that we can adjust this as needed.

PR #3027, opened 2021-06-25
Grow 688

BoulderBrains:
[Grow-688](https://kiva.atlassian.net/browse/GROW-688) 

- Added tracking to various versions of the CTA/action button shown in the LendCta. 
- Added a new non actionable button to handle when loan is funded, refunded, expired or all shares of the loan are reserved.
- Tested locally.

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -75,7 +75,7 @@
 							v-kv-track-event="[
 								'Lending',
 								'Add to basket',
-								'Lend now'
+								ctaButtonText
 							]"
 						>
 							{{ ctaButtonText }}
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -462,7 +462,7 @@ export default {
 		},
 		lendButtonVisibility() {
 			// eslint-disable-next-line max-len
-			if (!this.isAdding && this.amountInBasket === '' && !this.lentPreviously && this.status !== 'funded' && this.status !== 'refunded' && this.status !== 'expired' && !this.allSharesReserved) {
+			if (this.state !== 'adding' && this.state === 'basketed' && this.state !== 'lent-to' && this.status !== 'funded' && this.status !== 'refunded' && this.status !== 'expired' && !this.allSharesReserved) {
 				return true;
 			}
 			return false;
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -59,6 +59,7 @@
 							@change="$kvTrackEvent(
 								'Lending',
 								'Modify loan amount',
+								'priceOption from below?'
 							)"
 						>
 							<option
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -59,6 +59,7 @@
 							@change="$kvTrackEvent(
 								'Lending',
 								'Modify loan amount',
+								'priceOption from below?'
```
mcstover, line 62: We normally collect the amount in the 4th (value) + 5th (property) params as whole numbers.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -69,19 +70,24 @@
 								${{ priceOption }}
 							</option>
 						</kv-ui-select>
+
 						<!-- Lend button -->
+						<!-- eslint-disable max-len -->
 						<kv-ui-button
-							v-if="!isAdding && this.amountInBasket === '' && !this.lentPreviously"
+							v-if="!isAdding && this.amountInBasket === '' && !this.lentPreviously && this.status !== 'funded' && this.status !== 'refunded' && this.status !== 'expired' && !this.allSharesReserved"
```
mcstover, line 77: Can we create a couple computed properties to help with these multi-state checks for visibility? The inverse of this is reused below so a single computed property could power both cases.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -60,25 +70,72 @@
...
+							]"
 						>
 							Continue to checkout
 						</kv-ui-button>
+
+						<!-- Lend again/lent previously button -->
+						<kv-ui-button
+							v-if="this.lentPreviously"
+							class="tw-inline-flex tw-flex-1"
+							@click.native="addToBasket"
```
ryan-ludwig, line 109: No need for `@click.native` on the new buttons, just `@click` will work now :)

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -60,25 +66,67 @@
...
 						<!-- Lend button -->
 						<kv-ui-button
-							v-if="!showAdding && !inBasket"
+							v-if="lendButtonVisibility"
 							class="tw-inline-flex tw-flex-1"
 							@click="addToBasket"
+							v-kv-track-event="[
+								'Lending',
+								'Add to basket',
+								'Lend now'
```
emuvente, line 78: You could use the `ctaButtonText` variable here instead of `'Lend now'` so that the tracking event label always matches the actual text displayed on the button.

emuvente (approved):
This looks great! One small comment, but otherwise it's good to go

src/components/BorrowerProfile/LendCta.vue
mcstover: Good idea!

BoulderBrains: Good idea indeed, change complete.

emuvente (approved):

mcstover (approved):

ryan-ludwig (approved):

mlibazisi (approved):

PR #3037, opened 2021-06-28
GROW-691 Fix lend cta not updating after adding to basket

emuvente:
The userProperties.amountInBasket field only works when the query is made with a logged in user. To deal with that, this changes the "in basket" check to look for the loan id in the basket items array.

While I was testing that, I noticed that the "non actionable loan" button was being displayed but with "Lend now" as the button text, and that when adding the loan to the basket there were 2 buttons, one saying 'lend now' and the other saying 'adding to basket...'. Those were making it difficult to test my changes, so I refactored the state checks for the buttons to address those issues.

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -111,7 +111,7 @@
 
 						<!-- Funded, refunded, expired/ allSharesReserved button -->
 						<kv-ui-button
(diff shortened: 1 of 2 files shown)


mcstover (approved):

mlibazisi (approved):
👍

BoulderBrains (approved):
Nice refactors in there, the switch cases read more cleanly.

PR #3040, opened 2021-06-29
GROW-648 Country info

mlibazisi:


BoulderBrains (approved):
👍

src/components/BorrowerProfile/CountryInfo.vue
```diff
@@ -0,0 +1,84 @@
+<template>
+	<section>
+		<h2 class="tw-text-h2">
+			{{ countryName }} at a glance
+		</h2>
+		<div class="tw-flex lg:tw-mb-3">
+			<div class="tw-flex-auto">
+				<h2 class="tw-text-h2 tw-block tw-m-0">
```
ryan-ludwig, line 8: Minor, but I think this and the "loans currently fundraising" below would be better marked up as a paragraph tag, with 2 spans inside for styling. This would let a screenreader announce them together instead of as separate sentences.

Something like
```
<div class="tw-flex-auto">
		<p>
			<span class="tw-block tw-text-h2">$2500</span>
			<span class="tw-block tw-text-h4 tw-bg-gray-500">average annual income (USD)</span>
		</p>
</div>
```

emuvente: +1

src/components/BorrowerProfile/CountryInfo.vue
```diff
@@ -0,0 +1,118 @@
...
+				countryFacets {
+				  country {
+					isoCode
+					region
+				  }
+				}
+			  }
+			}
+		`,
+		preFetch: true,
```
emuvente, line 102: Querying countryFacets can take a while, so let's not prefetch this.

mlibazisi: @emuvente The query seems to only work with prefetch. Not sure if it's because its timing out

emuvente (approved):

PR #3042, opened 2021-06-29
Setup base components for map options. Leaflet for broader browser su…

mcstover:
…pport and MapLibreGL for nicer animations but slightly lower support


src/components/Kv/KvMapLibre.vue
```diff
@@ -0,0 +1,150 @@
...
+			zoomActive: false,
+		};
+	},
+	computed: {
+		refString() {
+			return `mapholder${this.mapId}`;
+		},
+	},
+	mounted() {
+		testDelayedGlobalLibrary('maplibregl').then(response => {
```
ryan-ludwig, line 75: I think vue-meta can handle all of this using a callback function https://vue-meta.nuxtjs.org/api/#callback

src/components/Kv/KvMapLibre.vue
```diff
@@ -0,0 +1,150 @@
+<template>
+	<div class="tw-w-full tw-relative">
+		<div :id="`kv-map-holder-${this.mapId}`" :ref="refString" style="min-height: 320px; min-width: 320px;">
+			<h1>a map</h1>
```
ryan-ludwig, line 4: Since these maps are decorative, can we remove this h1 and text?

mcstover: def will remove.

src/components/Kv/KvMapLibre.vue
mcstover: This appears to work nicely! Thanks!

mcstover: This only works with the server injected library so wasn't feasible in a component that provides either library script.

src/components/Kv/KvMapLibre.vue
mcstover: Replaced with default black image panel...it's kinda harsh though...

src/components/Kv/KvMap.vue
```diff
@@ -0,0 +1,297 @@
+<template>
+	<div class="tw-relative tw-block tw-w-full" :style="mapDimensions">
```
emuvente, line 2: Since you have the aspect ratio provided as a prop, you could use the padding-bottom trick to automatically set the height `padding-bottom: ${100 / aspectRatio}%;`. You would also need to add the `tw-absolute` class to the map element so that it would cover up the padding. That way this component will always take up the right amount of space on the page, even before it loads.

src/components/Kv/KvMap.vue
```diff
@@ -0,0 +1,297 @@
...
+		 * Control how quickly the autoZoom occurs
+		 */
+		autoZoomDelay: {
+			type: Number,
+			default: 1500
+		},
+		/**
+		 * These maps require a height to be set in order to render
+		 */
+		height: {
```
emuvente, line 72: Is this prop still needed?

src/components/Kv/KvMap.vue
```diff
@@ -0,0 +1,297 @@
...
+			const container = this.$el?.getBoundingClientRect();
+			const height = container ? `${container.width / this.aspectRatio}px` : '300px';
+			const width = container ? `${container.width}px` : '100%';
+			return {
+				height,
+				width,
+			};
+		},
+		refString() {
+			return `mapholder${this.mapId}`;
```
emuvente, line 140: I think the wrapperObserver might need to be destroyed if this changes, though this probably won't change in any of our use cases.

src/util/timeoutUtils.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Checks for global object on window from provided 3rd party library
+ * @param String library Variable name for eventual global library
+ * @param Number timeout Millisecond duration we should poll for presence of library
+ * @returns Promise
+ */
+export default function testDelayedGlobalLibrary(library, timeout = 3000) {
```
emuvente, line 7: Nice!

src/components/Kv/KvMap.vue
mcstover: It's not currently used. I may add some logic to override the aspectRatio driven values if the height and/or width are provided...

src/components/Kv/KvMap.vue
mcstover: Good idea, I'll give this a try!

src/components/Kv/KvMap.vue
mcstover: We would want to refresh the observer if this changed, but probably also re-render the map...Shouldn't be an issue in the near term.

src/components/Kv/KvMap.vue
mcstover: Ok I've added that in the computed props along with override handling for specified height and width values in via props.

src/components/Kv/KvMap.vue
```diff
@@ -0,0 +1,310 @@
+<template>
+	<div class="tw-relative tw-block tw-w-full" :style="mapDimensions">
+		<div
+			:id="`kv-map-holder-${this.mapId}`"
+			:ref="refString"
+			class="tw-w-full tw-h-full tw-bg-black"
+			:style="{ position: 'absolute' }"
```
mcstover, line 7: This was required to override the gl lib css :(, the leaflet based lib obeyed tailwind...

src/components/Kv/KvMap.vue
mcstover: Added width and height overrides in latest update.

emuvente (approved):
👍

mlibazisi (approved):

BoulderBrains (approved):

ryan-ludwig (approved):

PR #3048, opened 2021-06-30
general flss loans query

go-bears:
Jira Ticket:
https://kiva.atlassian.net/browse/PSD-265

What does this do:
creates graphql loan query to flss service with filtering ability. the query should return all information necessary to populate a loan card. 

how to test:
The full expanded version of the query with a country default of ["US"] should return loans from the flss endpoint:
`https://marketplace-api.k1.kiva.org/graphql`

```graphql
fragment loanCardFields on LoanBasic {
	id
	status
	name
	borrowerCount
	geocode {
		# city added for DetailedLoanCard Query Cache
		city
		# state is used for US loans on LoanChannelCategoryPage
		state
		country {
			isoCode
			name
			region
		}
	}
	use
	activity {
		id
		name
	}
	sector {
		id
		name
	}
	whySpecial
	lenderRepaymentTerm
	loanAmount
	minNoteSize
	loanFundraisingInfo {
		fundedAmount
		reservedAmount
		isExpiringSoon
	}
	plannedExpirationDate
	matchingText
	userProperties {
		favorited
		lentTo
	}
	# lenders added for DetailedLoanCard Query Cache
	lenders(limit: 0) {
		totalCount
	}
	... on LoanPartner {
		partnerName
	}
	...on LoanDirect {
		trusteeName
	}
}

query flssLoanData(
  $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
  $imgDefaultSize: String = "w480h360"
) {
  fundraisingLoans(filters: $filterObject) {
    totalCount
    values {
      id
      ...loanCardFields
      image {
        id
        default: url(customSize: $imgDefaultSize)
        retina: url(customSize: $imgRetinaSize)
        hash
      }
    }
  }
}
```

### DIFF
--- a/src/graphql/query/flssLoansQuery.graphql
+++ b/src/graphql/query/flssLoansQuery.graphql
@@ -4,11 +4,11 @@ query flssLoanData(
   $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
-  $limit: Int
-  $offest: Int
+  $limit: Int = 20
+  $offset: Int = 0
 ) {
   fundraisingLoans(filters: $filterObject,
-  							limit:  $limit,
+  							limit: $limit,
   							pageNumber: $offset) {
     totalCount
     values {
--- a/src/graphql/query/flssLoansQuery.graphql
+++ b/src/graphql/query/flssLoansQuery.graphql
@@ -1,11 +1,15 @@
 #import '../fragments/loanCardFields.graphql'
 
 query flssLoanData(
-  $filterObject: [FundraisingLoanSearchFilterInput!]
+  $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
+  $limit: Int = 50
+  $pageNumber: Int
 ) {
-  fundraisingLoans(filters: $filterObject) {
+  fundraisingLoans(filters: $filterObject,
+  							limit:  $limit,
+  							pageNumber: $pageNumber) {
     totalCount
     values {
       id
--- a/src/graphql/query/flssLoansQuery.graphql
+++ b/src/graphql/query/flssLoansQuery.graphql
@@ -4,12 +4,12 @@ query flssLoanData(
   $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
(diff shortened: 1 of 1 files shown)


src/graphql/query/flssLoansQuery.graphql
```diff
@@ -0,0 +1,21 @@
+#import '../fragments/loanCardFields.graphql'
+
+query flssLoanData(
+  $filterObject: [FundraisingLoanSearchFilterInput!]
+  $imgDefaultSize: String = "w480h360"
+  $imgRetinaSize: String = "w960h720"
+) {
```
mcstover, line 7: Looks good so far. Please add in variable handling for the `limit` and `pageNumber` so we can query for a specific number of loans to show and handle paging too.

go-bears: cool. thanks for 👀

go-bears: so add
$limit: int = (is there a default value you like?)
$pageNumber: int (no particular default)?

mcstover: We generally don't set defaults for these as the api should be defaulting to 20 for limit and 0 for pageNumber. Offset is how we usually term what is "pageNumber" in this case.

If there are no defaults in flss we can use 20 + 0

src/graphql/query/flssLoansQuery.graphql
```diff
@@ -1,11 +1,15 @@
 #import '../fragments/loanCardFields.graphql'
 
 query flssLoanData(
-  $filterObject: [FundraisingLoanSearchFilterInput!]
+  $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
+  $limit: Int = 50
```
mcstover, line 7: This is too high... Two reasons, 1. we couldn't show that many loans on a single screen ( currenly we might show 9 on a big screen), 2. we would never want to try and ingest that many loans into a client at one time as all the extra data in loanCardFields has to resolve from graphql too which would create a massive data payload.

go-bears: thanks for the insight, Matt! You got to the comment box before my next push went through 😄 
you're 💨  fast!

go-bears: appreciate the eyes and the explanation!

src/graphql/query/flssLoansQuery.graphql
mcstover: Thank you!

mcstover (approved):

src/graphql/query/flssLoansQuery.graphql
go-bears: Whoop! thank YOU! 🎉

PR #3049, opened 2021-07-01
AUTH-309 Use login sync cookie to limit checkSession calls

emuvente:


src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -149,23 +149,25 @@ export default {
...
-				this.savedSearches = _get(data, 'my.savedSearches.values');
-			} else {
-				this.favoritesCount = 0;
-				this.savedSearches = [];
-			}
-		}).finally(() => {
-			// data might have changed since the initial render, so trigger any needed updates
-			this.onOpen();
-		});
+		if (this.hasUserId) {
```
emuvente, line 152: This query was being called every time the page loaded, even when the user was not logged in and `this.userId` was undefined/null. On the vm this query is fairly slow, even without a logged in user, so hopefully skipping this when it doesn't have a user id will speed things up while developing and slightly speed things up in prod.

mcstover (approved):

PR #3052, opened 2021-07-01
GROW-710 BorrowerCountry wrapper component

mlibazisi:


src/components/BorrowerProfile/BorrowerCountry.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+	<article>
+		<kv-map
+			class="
+				tw-rounded
+				tw-overflow-hidden
+				tw-my-5
```
mcstover, line 7: Great job getting this component in place @mlibazisi. The only thing I updated in this file was to change the margin to only vertical and bump it up once more.

.storybook/stories/KvMap.stories.js
```diff
@@ -5,19 +5,32 @@ export default {
 	component: KvMap,
 	args: {
 		autoZoomDelay: 1000,
+		height: null,
 		initialZoom: null,
 		lat: 37.700091,
 		long: -123.013243,
 		mapId: 0,
 		useLeaflet: false,
+		width: null,
```
mcstover, line 14: Added the width and height props here and in the template to fix up the stories!

src/components/Kv/KvMap.vue
```diff
@@ -86,14 +86,14 @@ export default {
 		 */
 		lat: {
 			type: Number,
-			default: 37.700091
+			default: null
```
mcstover, line 89: Having defaults in lat + long was causing them to get set during the component creation and not updated when the props came in. Added the watcher and ability to initialize using those or on mounted.

emuvente (approved):
👍

PR #3059, opened 2021-07-07
Grow 688 fix

BoulderBrains:
While QAing grow-688 I found a couple issues. 

1) The lend again button was showing up when the loan had been basketed. 
Screen Shot 2021-07-07 at 8 00 29 AM
- Changing the v-if condition on the "lend again" button fixed this issue

2) Tracking was not working correctly for the lend again button, upon click I was getting the tracking info of the Lend button. I added keys to both buttons, which tells vue that the buttons are in fact different and passes through all the data for each button. They are so similar, it seems that vue was only updating some of the fields (not tracking fields) which was causing the issue.

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -382,8 +382,6 @@ export default {
 			switch (this.state) {
 				case 'loading':
 					return 'Loading...';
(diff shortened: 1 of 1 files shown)


emuvente (approved):
👍

mcstover (approved):

PR #3064, opened 2021-07-09
Wrap Lend CTA addToBasket buttons in a form element and fire action o…

mcstover:
…n submit instead of click

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -45,42 +45,88 @@
 						{{ lgScreenheadline }}
 					</p>
 					<span class="tw-flex tw-pb-1 lg:tw-pb-3">
-						<label for="LoanAmountDropdown" class="tw-sr-only">Lend amount</label>
-						<kv-ui-select
-							v-if="hideShowLendDropdown"
-							id="LoanAmountDropdown"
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -45,42 +45,88 @@
...
-							<option
-								v-for="priceOption in prices"
-								:key="priceOption"
-								:value="priceOption"
-							>
-								${{ priceOption }}
-							</option>
-						</kv-ui-select>
+						<form @submit.prevent="addToBasket" class="tw-w-full tw-flex">
+							<fieldset class="tw-w-full tw-flex">
```
mcstover, line 49: repeating these width and flex classes was required to ensure the new containing elements stretched to the full width of the span...

ryan-ludwig, line 49: Nice! Would recommend adding `:disabled="isAdding"` to this fieldset to prevent users who hit enter twice rapidly from submitting twice.

ryan-ludwig (approved):

BoulderBrains (approved):

src/components/BorrowerProfile/LendCta.vue
mcstover: Good call. Added.

PR #3066, opened 2021-07-10
Vue 625 braintree client init

mcstover:
The braintree-dropin-lib doesn't play nice with NodeJS 16. This ensures that all is well with the impending update.


emuvente (approved):
:+1:

BoulderBrains (approved):

PR #3068, opened 2021-07-12
Grow 630

BoulderBrains:
[Grow-630](https://kiva.atlassian.net/browse/GROW-630)

Here's the first crack at the the slide in animation and slot machine style animation for the Number of Lenders and Matching Text data.

**Loading animation:**
_Small screen_ -> 
small loading animation

_Medium screen_ -> 
https://user-images.githubusercontent.com/1521381/125360682-7ae62700-e329-11eb-85b9-b77480bfcb1d.mp4

_Large screen_ -> 
https://user-images.githubusercontent.com/1521381/125360780-99e4b900-e329-11eb-82fe-268d2d39ee01.mp4

The **slot machine style animation** is consistent across screen sizes and here's an example of that -> 
https://user-images.githubusercontent.com/1521381/125361029-f9db5f80-e329-11eb-91ea-71a658e0f593.mp4

There are **3 issues I'd like feedback**/suggestions with
**RESOLVED** 1) At the end of the slot machine animation, you can see the matching text jumping down instead of having a smooth transition. Code wise, the same animations are being run on the Number of lenders stat and the Matching Text, I believe the Matching text jumping down is a result of it being the second in the list of animated item, the hiding of the matching text timing doesn't seem to be working out with the show timing of the number of lenders stat resulting in the visual jump. Check out the video,  I'm open suggestions/discussion around this. 
https://user-images.githubusercontent.com/1521381/125362921-f39ab280-e32c-11eb-9534-99574823996b.mp4

**RESOLVED** 2) This is an edge case. The loading animation for when a user is in a medium screen size with the LendCta out of view, ie "isSticky" doesn't work. I attempted to adding bindings to my transition-group, similar to the other class bindings in the file, but it won't compile after doing so. ie ->

Screen Shot 2021-07-12 at 4 12 40 PM

```
<transition-group
     enter-active-class="tw-transition-transform tw-duration-700 tw-delay-300"
     :enter-class="[
					'tw-transform',
					'tw-translate-y-7',
					'md:tw--translate-y-7',
					{
						'md:tw-translate-y-7': !isSticky
					}
					'lg:tw--translate-y-7',
				]"
				:enter-to-class="[
					'tw-transform',
					'tw-translate-y-0',
					{
						'md:tw-translate-y-0': !isSticky
					}
					'md:tw-translate-y-0',
					'lg:tw-translate-y-0'
				]"
			>
```
Can we add the ability to bind these animation classes? I think we'll need to in order to get this edge case animation working correctly. As always, open to discussion and other ideas.

**RESOLVED**  3) The figma mocks show the handshake emoji (Which has now been updated to the party emoji) for the matching text. I've implemented an emoji option as well at the material design party-popper icon option. Which do you like visually? And which should we use? (i know we've been trying to do all material design icons) (*I've also reached out to design to get their opinion) - 
**UPDATE**
- Design prefers the full color emoji for visibility reasons. It's more recognizable and multi-colored.

Emoji option: 
Screen Shot 2021-07-13 at 2 53 05 PM

Material Design Icon: 
Screen Shot 2021-07-13 at 2 53 36 PM


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -158,50 +159,107 @@
...
+					</div>
+				</kv-grid>
+			</transition-group>
 		</div>
 	</div>
 </template>
 
 <script>
 import { mdiLightningBolt } from '@mdi/js';
+import { mdiHandshakeOutline } from '@mdi/js';
```
mcstover, line 262: Create a single import like so: `{ mdiLightningBolt, mdiHandshakeOutline }`

mcstover (approved):

ryan-ludwig (approved):

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -452,9 +531,21 @@ export default {
...
 		},
+		transitionEnterClasses() {
+			if (this.isSticky) {
+				return 'tw-transform tw-translate-y-7 md:tw-translate-y-7 lg:tw--translate-y-7';
+			}
+			return 'tw-transform tw-translate-y-7 md:tw--translate-y-7 lg:tw--translate-y-7';
+		},
+		transitionEnterToClasses() {
+			// eslint-disable-next-line max-len
+			return "[ 'tw-transform','tw-translate-y-0', { 'md:tw-translate-y-0': !isSticky } 'md:tw-translate-y-0','lg:tw-translate-y-0' ]";
```
ryan-ludwig, line 542: This will need to compare against `this.isSticky` like you have in transitionEnterClasses above

BoulderBrains: Better yet, I just noticed that the md:tw-translate-y-0 is the same regardless of isSticky, so I removed this computed function and moved this back up above to the enter-to-class="".

mcstover (approved):
Let's get this merged in before branch cut please.

PR #3072, opened 2021-07-13
LendCta: Smooth out the animation

ryan-ludwig:
I'm really not sure why the transition-group+delay wasn't working. It seems like it should.
Anyhow, after futzing around enough this seems to smooth things out.

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -208,10 +208,11 @@
 							'lg:tw-col-span-12'
 						]"
 					>
-						<transition-group
+						<transition
+							mode="out-in"
 							key="transition"
 							class="tw-flex tw-flex-col"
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -221,7 +222,7 @@
 							<span
 								class="tw-inline-block tw-align-middle"
 								key="numLendersStat"
-								v-show="statScrollAnimation"
+								v-if="statScrollAnimation"
```
mcstover, line 225: 👍  Didn't realize we still had `show` instead of `if` here which is required to fully remove the markup from the DOM to isolate the transitions.

mcstover (approved):

BoulderBrains (approved):

BoulderBrains:
Thanks for taking a look at this one @ryan-ludwig

PR #3074, opened 2021-07-14
PSD-280: add data call with Graphql flss query to filter-alpha page

go-bears:
Jira ticket:
https://kiva.atlassian.net/browse/PSD-280

What does it do:
adds flss query to filter-alpha page to provide access to live loans and data for component presentation. 

How to test:
download the branch and run the dev server
when the `filter-alpha` page renders open the developer tools console and look for `console.log` outputs for loans from the US and the totalCount number of available loans.

### DIFF
--- a/src/graphql/query/flssLoansQuery.graphql
+++ b/src/graphql/query/flssLoansQuery.graphql
@@ -4,7 +4,7 @@ query flssLoanData(
   $filterObject: [FundraisingLoanSearchFilterInput!]
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
-  $limit: Int = 50
+  $limit: Int = 0
   $offset: Int = 0
 ) {
   fundraisingLoans(filters: $filterObject,
--- a/src/graphql/query/flssLoansQuery.graphql
+++ b/src/graphql/query/flssLoansQuery.graphql
@@ -4,7 +4,7 @@ query flssLoanData(
   $filterObject: [FundraisingLoanSearchFilterInput!]
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
-  $limit: Int = 0
+  $limit: Int = 20
   $offset: Int = 0
 ) {
   fundraisingLoans(filters: $filterObject,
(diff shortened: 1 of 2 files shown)


src/graphql/query/flssLoansQuery.graphql
```diff
@@ -1,10 +1,10 @@
 #import '../fragments/loanCardFields.graphql'
 
 query flssLoanData(
-  $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
+  $filterObject: [FundraisingLoanSearchFilterInput!]
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
-  $limit: Int = 20
+  $limit: Int = 50
```
mcstover, line 7: Please lower this default. This file will now become shared so we shouldn't be asking for 50 loans at a time by default.

mcstover (approved):
Beyond the default fetch limit, this is a good foundation to build on!

ryan-ludwig (approved):

go-bears:
Thanks for 👍  @mcstover @ryan-ludwig ! I reset the limit to 20 but moved where that gets controlled to the vue page.

src/graphql/query/flssLoansQuery.graphql
```diff
@@ -1,10 +1,10 @@
 #import '../fragments/loanCardFields.graphql'
 
 query flssLoanData(
-  $filterObject: [FundraisingLoanSearchFilterInput!] = { countryIsoCode: {any: ["US"]}}
+  $filterObject: [FundraisingLoanSearchFilterInput!]
   $imgDefaultSize: String = "w480h360"
   $imgRetinaSize: String = "w960h720"
-  $limit: Int = 20
+  $limit: Int = 0
```
meganmcmillan, line 7: I could see this biting someone in the future if they don't set a limit in their component - "why is my query returning 0 loans???". We probably want to keep this default limit as 10-20 here, which can then be overridden in the component.

meganmcmillan (commented):
One comment. Overall this looks good!

src/graphql/query/flssLoansQuery.graphql
go-bears: thanks for 👀 !

PR #3076, opened 2021-07-14
grow-714 running content through v-html to translate HTML tags

BoulderBrains:
[Grow-714](https://kiva.atlassian.net/browse/GROW-714)

Running more about this loan content through v-html to translate html.


src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -18,8 +17,8 @@
 				<p
```
mcstover, line 17: This should be a div element instead of p.

mcstover (approved):

PR #3077, opened 2021-07-14
VUE-641 nest Tailwind classes in a container to add specificity.

ryan-ludwig:
2 main pieces to this,
* break tailwind into a static css file since they're global styles
* prepend all of the tailwind resets and utility classes with `.kv-tailwind` to give enough specificity to win over foundation resets


ryan-ludwig:
@emuvente I tried doing the alternative method of prefixing Foundation, like 

```
.kv-foundation {
	@include foundation-global-styles;
	@include foundation-flex-classes;
	@include foundation-flex-grid;
	@include foundation-visibility-classes;
	@include foundation-typography;
	@include foundation-forms;

	// import global styles
	@import 'global/page-layout';
	@import 'global/line-breaks';
	@import 'global/typography';
	@import 'global/forms';
	@import 'global/transitions';
	@import 'global/print';
	@import 'global/drop-in';
}
```

Unfortunately this meant that any tag selectors imported above (like `a {}, button{}, input{}`), now would be stronger than any single class value. 

For example
`<a class="make-text-pink">link</a>`
Would not be pink because 
`.make-text-pink { color: pink; }`
has less weight than
`.kv-foundation a { color: blue; }`

We have a lot of single class usage in the site (which is great!) so it seemed like a lot of extra work to go that route.

mcstover (approved):

src/App.vue
```diff
@@ -160,4 +160,26 @@ export default {
 #app {
 	height: 100%;
 }
+
+// Kludges to ensure Kiva Classic styles are correct when using the .kv-tailwind class
+// Remove once Kiva Classic is the default
+.kv-tailwind {
+	p,
```
ryan-ludwig, line 167: The existing reset sets line-heights on these elements to 1.75.

Tailwind's reset doesn't specify a line-height on these elements, so they were using the old system.

src/App.vue
```diff
@@ -160,4 +160,26 @@ export default {
...
+	dl,
+	ol,
+	ul { line-height: 1.5; }
+}
+
+body .kv-tailwind {
+	font-weight: 300;
```
ryan-ludwig, line 174: This took me a while to figure out. 
The new prepended-selector in this PR prefixes all tailwind stuff with .kv-tailwind. For the body reset, this looks like
`.kv-tailwind body { font-weight: 300, etc. }`
That selector will never get hit unless you were to put the .kv-tailwind class on the html element. We're not ready to do that yet.  This kludge works around the problem until we can remove the prepended selector.

ryan-ludwig:
VUE-641

mcstover (commented):
👍

PR #3082, opened 2021-07-16
Lazy load More about this loan using the introspection observer appro…

mcstover:
…ach.


BoulderBrains (approved):
"Lazy load More about this loan using the introspection observer approach"

☝️ Introspection observer sounds like a pretty good band name 🎵

ryan-ludwig (approved):

PR #3083, opened 2021-07-16
Bp minor fixes

mcstover:


src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -67,7 +67,7 @@ export default {
 		};
 	},
 	apollo: {
-		prefetch: true,
+		preFetch: true,
```
mcstover, line 70: Ensures the Loan story is actually preFetched!

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -10,41 +10,41 @@
 					<hero-background />
 				</div>
 				<content-container class="md:tw-pt-6 lg:tw-pt-8">
-					<summary-card class="tw-relative lg:tw--mb-1.5" />
+					<summary-card class="tw-relative lg:tw--mb-1.5 tw-z-1" />
```
mcstover, line 13: Application of these classes and the relative classes below did not change the layout from what I have seen. The relative classes adjusted the flow so that the content in those containing components was not obscured by the LendCTA component and it's full width content area.

BoulderBrains (approved):
🧹

ryan-ludwig (approved):

PR #3084, opened 2021-07-16
Refactor CountryInfo to fetch it's own data and add in placeholder co…

mcstover:
…ntent blocks.

### DIFF
--- a/src/components/BorrowerProfile/BorrowerCountry.vue
+++ b/src/components/BorrowerProfile/BorrowerCountry.vue
@@ -13,13 +13,7 @@
 			:zoom-level="mapZoomLevel"
 			:initial-zoom="mapInitialZoom"
 		/>
-		<country-info
-			:num-loans-fundraising="countryNumLoansFundraising"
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/BorrowerCountry.vue
```diff
@@ -48,28 +42,16 @@ export default {
...
-			countryNumLoansFundraising: 0,
-			countryAvgAnnualIncome: '',
-			countryIsoCode: '',
-			countryName: '',
-			countryRegionName: '',
 		};
 	},
 	apollo: {
-		query: gql`query borrowerCountry($loanId: Int!) {
+		query: gql`query borrowerCountryCoords($loanId: Int!) {
```
mcstover, line 48: We maintain an initial load of the map coordinates so it can be initialized from the created hook in the client and prepare for the animated map when scrolled into view.

emuvente (approved):
:+1:

BoulderBrains (approved):

PR #3087, opened 2021-07-17
Created page layout for lend filter alpha

dymonda21:
- I've created a page layout by going in and editing the lend filter alpha code and adding more towards the user interface. 

https://kiva.atlassian.net/browse/PSD-263?atlOrigin=eyJpIjoiOTNjNDNlM2I1YWFmNDc0ODg4ODQwNDhhYzRjZGJhNTciLCJwIjoiaiJ9

- The way that you can test this is by simply pulling up the lend filer alpha page and below you should see grey boxes labeled to which each object should go. Once those boxes pop up you can minimize to see the filter box move on top of the loan cards and back to the left column once maximized. 

http://localhost:8888/lend/filter-alpha


go-bears:
Congrats on getting the ticket in @dymonda21 ! 🎉 
You'll need to resolve the merge conflict but you can do that in github. If you click "resolve conflicts" it's going to open a simple text editor and show you where the conflict is. 
You'll see brackets around the code where git can't decide which change/commit it should pay attention to and keep. 

**Delete** the section that you **don't** want. and keep the section that you do want to keep. 
And delete all the parts and dashes associated with `HEAD` or   `commit` to clean up the conflict and then github will let you merge the commit.

src/pages/Lend/FilterAlpha/LendFilterAlpha.vue
```diff
@@ -11,6 +11,41 @@
...
 			</div>
+			<kv-page-container>
+				<h2 class="tw-mb-2">
+					Make a loan, change a life
+				</h2>
+				<h3 class="tw-mb-3">
+					Each Kiva loan helps people build a better future for themselves and their families.
+				</h3>
+				<kv-grid class="tw-grid-cols-2 md:tw-grid-cols-3">
+					<div class="tw-bg-gray-300 text-center sm:text-left"><p>Filters</p>
```
ryan-ludwig, line 22: This is minor and not required, but wanted to mention it. A common practice is to write responsive code "mobile first". What that means is that the default code is for mobile, and then you layer changes on top of that as the screen sizes increase.

So in this case, we want small screens to be text-left, and medium and up to be text-center. We can write that like
`text-left md:text-center` 
instead of 
`sm:text-left text-center`

Both of those will do the same thing! It's just a different way of thinking about how we layer styles.

--

Also, to get these classes to work in Kiva's application, they'll need to `tw` prefix. So `tw-text-left md:tw-text-center`

src/pages/Lend/FilterAlpha/LendFilterAlpha.vue
```diff
@@ -11,6 +11,41 @@
...
+					</div>
+
+					<div class="tw-col-span-2 tw-bg-white-300">
+						<div class="tw-bg-gray-300 tw-h-30">Search Loans</div>
+						<p>Loan Count</p>
+                        <kv-grid class="tw-grid-rows-4">
+							<div class="tw-bg-gray-300 tw-h-30">Loan Card 1</div>
```
ryan-ludwig, line 29: We use a customized spacing system at Kiva, which isn't quite like the docs in Tailwind. That's why these `tw-h-30` classes aren't making things taller right now.

You can either remove these classes, or you could use `tw-h-4` which is 32px in our system, so almost 30 :)

If you're using VsCode, I'd recommend adding the [Tailwind intellisense plugin](https://github.com/tailwindlabs/tailwindcss-intellisense). It will auto-complete classes for you so you can tell which heights are available in Kiva's system.

ryan-ludwig:
This is looking great @dymonda21, nice work! Loving how the columns change between mobile and desktop screens.

src/pages/Lend/FilterAlpha/LendFilterAlpha.vue
```diff
@@ -11,6 +11,41 @@
...
+					Make a loan, change a life
+				</h2>
+				<h3 class="tw-mb-3">
+					Each Kiva loan helps people build a better future for themselves and their families.
+				</h3>
+				<kv-grid class="tw-grid-cols-2 md:tw-grid-cols-3">
+					<div class="tw-bg-gray-300 text-center sm:text-left"><p>Filters</p>
+					</div>
+
+					<div class="tw-col-span-2 tw-bg-white-300">
```
ryan-ludwig, line 25: Minor, I don't believe the `tw-bg-white-300` class is doing anything and can be removed.

ryan-ludwig (approved):
This is looking great @dymonda21, nice work! Loving how the columns change between mobile and desktop screens.

Some very minor things that could be tweaked, but not required to get this merged.

meganmcmillan (commented):
Looking good! 💪 
Let me know if you have any questions addressing Ryan's feedback.

mcstover (approved):
Nice work @dymonda21 !

meganmcmillan (approved):

PR #3088, opened 2021-07-19
VUE-634 add caching for github action build dependencies

emuvente:
Following these guides ([github](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action), [cypress](https://docs.cypress.io/guides/continuous-integration/introduction#Caching), [stack overflow](https://stackoverflow.com/questions/55230628/is-there-a-way-to-speedup-npm-ci-using-cache)), this will make it so that the npm cache and the cypress binary cache are preserved between builds, allowing subsequent PRs to install much faster.

It's possible to preserve multiple caches, based on a key name. This uses key names based on the hash of package-lock, so anytime package-lock changes a new cache is created. The restore-keys tell github which caches to use as a fallback in case there isn't an exact match for the key (as would be the case when package-lock changes). This allows most dependencies to be installed from cache, while new/changed dependencies will be fetched from npm.

### DIFF
--- a/.github/workflows/lighthouse-dev.yml
+++ b/.github/workflows/lighthouse-dev.yml
@@ -21,9 +21,22 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 15.x
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-dependencies
+        with:
+          path: |
+            ~/.cache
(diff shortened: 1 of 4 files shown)


ryan-ludwig (approved):
Nice!

mcstover (approved):

PR #3093, opened 2021-07-20
grow-701 tracking event for borrowerProfile page load

BoulderBrains:
[Grow-701](https://kiva.atlassian.net/browse/GROW-701)

Added a tracking event for borrower profile page load, which contains the loan status.

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -89,5 +89,12 @@ export default {
 			loanId: Number(this.$route.params.id || 0),
 		};
 	},
+	created() {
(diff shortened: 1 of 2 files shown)


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -89,5 +89,12 @@ export default {
 			loanId: Number(this.$route.params.id || 0),
 		};
 	},
+	created() {
```
mcstover, line 92: I think this needs to move the `mounted` hook inside the SummaryCard.vue component.

The `status` property is populated by that query and calling in mounted ensure it's set before firing the event.

mcstover (approved):
Thank you, this should do it!

PR #3096, opened 2021-07-20
Update Ui Ingress config to use apiVersion: networking.k8s.io/v1 and …

mcstover:
…it's corresponding configuration structure.

### DIFF
--- a/deploy/charts/templates/ingress.yaml
+++ b/deploy/charts/templates/ingress.yaml
@@ -1,6 +1,6 @@
 {{- if .Values.ingress.enabled -}}
 {{- if $fullName := include "ui-helm.fullname" . -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
--- a/deploy/test/values.yaml
+++ b/deploy/test/values.yaml
@@ -41,10 +41,13 @@ ingress:
       - host: "ui-test.dk1.kiva.org"
         http:
           paths:
-          - backend:
-              serviceName: "kiva-ui"
-              servicePort: *servicePortName
-            path: /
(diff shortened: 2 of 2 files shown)


mcstover:
We were able to successfully deploy to the Test env with these configs.

emuvente (approved):

JoeTravisKiva (approved):
This looks good. Out of curiosity, did you have to manually add the `meta.helm.sh` annotations in the existing ingress resource?

mcstover:
@JoeTravisKiva I'm not sure which part of our configs you're referring to. Pretty much everything we declare is hand-hewn and not versioned by helm at this point. We can check in and discuss further if you'd like!

JoeTravisKiva:
It's these last two annotations that I had been banging my head against (our ingress doesn't have them, maybe because it's one of the oldest resources that we've got?). 
```
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-prod
    kubernetes.io/ingress.class: nginx-ingress-external
    kubernetes.io/ingress.ssl-redirect: "true"
    meta.helm.sh/release-name: kiva-ui
    meta.helm.sh/release-namespace: kiva-marketplace-qa
```
Helm didn't want to allow for an upgrade because the resource didn't have the `meta.helm.sh/release-name` and  `meta.helm.sh/release-namespace` annotations, but it appears that if we add them by hand, I can get past that.

mcstover:
@JoeTravisKiva Looks like we set those directly in the helm command in Jenkinsfile (https://github.com/kiva/ui/blob/master/Jenkinsfile#L63) The release-name gets picked up as the value after `--install` and we set the `--namespace` too...

JoeTravisKiva:
@mcstover Yeah, we use the same command. I don't know why we didn't automatically get the annotations though.

PR #3106, opened 2021-07-27
Update k8s resource usage with tierd definitions that set the max-old…

mcstover:
…-space-size lower that the requested memory limit. Updates for dev, stage and qa. Test and Prod need to be updated too.


mcstover:
FYI @emuvente @ryan-ludwig @BoulderBrains

PR #3108, opened 2021-07-28
Update DetailsTabs component to serve as controller for sourcing and …

mcstover:
…presenting definition content for links within each tab. Both legacy salesforce solutions and contentful entries are supported.


ryan-ludwig (approved):
👍

BoulderBrains (approved):

PR #3114, opened 2021-07-28
Grow 700

BoulderBrains:
[Grow-700](https://kiva.atlassian.net/browse/GROW-700)

- Anonymous user are grouped and displayed as 1 "supporter card". 
- If a user is logged in and has lent to the loan they are viewing, their "supporter card" is displayed first in the list. 
- Supporter lenders and teams urls constructed and in place, along with associated tracking.


src/components/BorrowerProfile/LendersAndTeams.vue
```diff
@@ -253,23 +306,35 @@ export default {
 				this.observer.disconnect();
 			}
 		},
+		filterAnonymousSuporters() {
+			const filterItemsList = this.truncatedItemList.filter(item => item.name === 'Anonymous');
+			if (filterItemsList) {
+				this.hasAnonymousSupporters = true;
```
ryan-ludwig, line 312: This isn't a blocker, but could `this.hasAnoymousSupporters` be a computed property instead?

It would basically return true or false using the logic you've got in this function.

src/components/BorrowerProfile/LendersAndTeams.vue
```diff
@@ -40,14 +40,28 @@
...
-					:lender-page="`${ displayType === 'lenders' ? item.lenderPage : null}`"
+					:public-id="`${ displayType === 'lenders' ? item.publicId : item.teamPublicId}`"
 					:whereabouts="`${ displayType === 'lenders' ? item.lenderPage.whereabouts : ''}`"
 				/>
+				<supporter-details
+					v-if="this.hasAnonymousSupporters && this.displayType === 'lenders'"
+					name="+ Anonymous lenders"
+					display-type="lenders"
+					:has-anonymous-supporters="this.hasAnonymousSupporters"
+				/>
```
mcstover, line 64: The supporter-details component has become quite versatile!

src/components/BorrowerProfile/SupporterDetails.vue
```diff
@@ -1,41 +1,67 @@
...
-			</span><br>
-			<span
-				v-if="lenderPage && whereabouts"
-				class="tw-text-h4"
-			>
-				{{ whereabouts }}
-			</span>
-		</div>
+		<router-link
+			class="tw-contents tw-no-underline"
```
ryan-ludwig, line 4: I believe you could remove the div that is the parent of this, and replace `tw-contents` with all of the classes from that div.

ryan-ludwig: Then if you wanted black text, you could add a class `tw-text-black` to this.

src/components/BorrowerProfile/SupporterDetails.vue
```diff
@@ -1,41 +1,67 @@
...
+						<path d="M10.1143 23.2751C21.9428 23.2751 24.6857 13.2126 24.6857 11.4626H23.6571C11.8286 11.4626 9.08569 21.5251 9.08569 23.2751H10.1143Z" fill="white" />
+						<path d="M9.08569 24.2376C9.08569 26.0751 11.1428 36.3126 23.8285 36.3126H24.8571C24.8571 34.4751 22.8 24.2376 10.1143 24.2376H9.08569Z" fill="white" />
+					</svg>
+					<!-- eslint-enable max-len -->
+				</div>
+			</div>
+			<div class="tw-inline-block">
+				<span
+					v-if="name"
+					class="tw-no-underline"
```
ryan-ludwig, line 52: I don't think these `tw-no-underline` classes are doing anything since our link tags by default have no underline

ryan-ludwig (approved):

src/components/BorrowerProfile/LendersAndTeams.vue
mcstover: I can test changing this to computed in some follow up work wherein I will be making adjustments for the full lists within the lightbox.

mcstover (approved):

PR #3123, opened 2021-07-30
VUE-642 Rich Text Renderer

eddieferrer:
Adds a dynamic rich text renderer which will render rich text fields with embedded content. 

- Will render images, video, and rich text fields embedded in rich text fields, either as entries or inline entries. 

TODO:

- Implement support for other embedded content types. (Kv-Button) 
- Possibly fine tune document nodes for our styleguide typography classes

### DIFF
--- a/build/webpack.base.conf.js
+++ b/build/webpack.base.conf.js
@@ -41,7 +41,9 @@ module.exports = {
 			'@': resolve('src'),
 			'foundation': 'foundation-sites/js',
 			// alias promise module to handle timesync calling require('promise')
-			'promise': resolve('build/promise.js')
+			'promise': resolve('build/promise.js'),
+			// required for src/components/Contentful/DynamicRichText.vue
+			'vue$': 'vue/dist/vue.esm.js'
 		}
 	},
 	module: {
--- a/src/components/Contentful/DynamicHero.vue
+++ b/src/components/Contentful/DynamicHero.vue
@@ -76,7 +76,8 @@
 				<div class="small-10 large-6 xlarge-7 align-self-middle columns contentful-hero__cta_wrapper">
 					<h1 class="contentful-hero__header" v-html="heroHeadline">
 					</h1>
-					<div class="contentful-hero__body" v-html="heroBody">
+					<div class="contentful-hero__body">
+						<dynamic-rich-text :html="heroBody" />
 					</div>
 					<kv-button
 						v-if="heroButton.text"
@@ -117,8 +118,9 @@
 <script>
 import KvButton from '@/components/Kv/KvButton';
 import SectionWithBackground from '@/components/Contentful/SectionWithBackground';
+import { richTextRenderer } from '@/util/contentful/richTextRenderer';
+import DynamicRichText from '@/components/Contentful/DynamicRichText';
 import KvContentfulImg from '~/@kiva/kv-components/vue/KvContentfulImg';
-import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
 
 /**
 * Dynamic Hero Component
@@ -132,6 +134,7 @@ export default {
 		KvButton,
 		KvCarousel: () => import('@/components/Kv/KvCarousel'),
 		KvCarouselSlide: () => import('@/components/Kv/KvCarouselSlide'),
+		DynamicRichText,
 		KvContentfulImg,
 		SectionWithBackground,
 	},
@@ -237,7 +240,7 @@ export default {
 		},
 		heroBody() {
 			const text = this.heroText?.bodyCopy ?? '';
-			return text ? documentToHtmlString(text) : '';
+			return text ? richTextRenderer(text) : '';
 		},
 		heroButton() {
 			return {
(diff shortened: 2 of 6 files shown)


go-bears:
👋  @eddieferrer ! Nice to have you back! Congrats with 🍼 👶 !

build/webpack.base.conf.js
```diff
@@ -41,7 +41,9 @@ module.exports = {
 			'@': resolve('src'),
 			'foundation': 'foundation-sites/js',
 			// alias promise module to handle timesync calling require('promise')
-			'promise': resolve('build/promise.js')
+			'promise': resolve('build/promise.js'),
+			// required for src/components/Kv/KvRichTextRenderer.vue
+			'vue$': 'vue/dist/vue.esm.js'
```
eddieferrer, line 46: The entire vue compiler is required here. Bundle comparison shows 33.27 kb for the full compiler, and 22.79 kb for our previous runtime only Vue.

src/components/Contentful/DynamicHero.vue
```diff
@@ -76,7 +76,8 @@
 				<div class="small-10 large-6 xlarge-7 align-self-middle columns contentful-hero__cta_wrapper">
 					<h1 class="contentful-hero__header" v-html="heroHeadline">
 					</h1>
-					<div class="contentful-hero__body" v-html="heroBody">
+					<div class="contentful-hero__body">
+						<dynamic-rich-text :html="heroBody" />
```
eddieferrer, line 80: Example of how to use the renderer. It works with our current rich text fields, which have no embedded content

src/components/Contentful/DynamicRichText.vue
```diff
@@ -0,0 +1,41 @@
+<script>
+import KvContentfulImg from '~/@kiva/kv-components/vue/KvContentfulImg';
```
eddieferrer, line 2: We'll have to import all possible components that can be in a rich text field here.

src/components/Contentful/DynamicRichText.vue
```diff
@@ -0,0 +1,41 @@
...
+	props: {
+		html: {
+			type: String,
+			default: ''
+		}
+	},
+	computed: {
+		dynamicComponent() {
+			return {
+				template: `<div>${this.html}</div>`,
```
eddieferrer, line 18: this.html can include an image component: "kv-contentful-img"  In order to actually render that with vue we need the full compiler build.

src/components/Contentful/DynamicRichText.vue
```diff
@@ -0,0 +1,41 @@
...
+				},
+				props: [
+					/* define props as needed */
+				]
+			};
+		}
+	},
+	render(createElement) {
+		return createElement(
+			'div', /* Wrap output in a div since Vue templates must have 1 root element */
```
eddieferrer, line 30: This wrapper this is for this `DynamicRichText.vue` component, which unfortunately is extraneous, but I couldn't figure out how to get this component to work without it

src/util/contentful/richTextRenderer.js
```diff
@@ -0,0 +1,64 @@
...
+ * Returns html string from rich text nodes
+ *
+ * @param {object} content Content of a contentful rich text field
+ * @returns {string} String of html to render
+ */
+export function richTextRenderer(content) {
+	const options = {
+		renderNode: {
+			[INLINES.EMBEDDED_ENTRY]: node => {
+				return `<div class="tw-prose">${richTextRenderer(node?.data?.target?.fields?.richText)}</div>`;
```
eddieferrer, line 53: Turtles all the way down. Unfortunately, this can create multiple nested <div class="tw-prose"> divs in the case that in contentful there is an embedded rich text field.

src/util/contentful/richTextRenderer.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+const assetRenderer = contentfulAssetObject => {
+	const isVideo = contentfulAssetObject?.file?.contentType.includes('video');
+	const isImage = contentfulAssetObject?.file?.contentType.includes('image');
+	if (isImage) {
+		return `
+			<kv-contentful-img
+				contentful-src="${encodeURI(contentfulAssetObject?.file?.url)}"
+				alt="${contentfulAssetObject?.description}"
+				height="${contentfulAssetObject?.file?.details?.image?.height}"
```
ryan-ludwig, line 26: This will be the original upload width/height of the image?

src/components/Contentful/DynamicRichText.vue
emuvente: You could use async imports, and then they won't be imported until they're used:
```
const KvContentfulImg = () => import('~/@kiva/kv-components/vue/KvContentfulImg');
```

src/util/contentful/richTextRenderer.js
eddieferrer: Yes. This is for inline images which should be reasonably sized, but I'm open to suggestions

PR #3124, opened 2021-07-30
VUE-643 remove commonjs transform

ryan-ludwig:
This seems to drastically improve tree-shaking. I'm not sure if there are improvements across the board or if it's mainly the mdi.js package getting cleaned up. 

Arrived at this direction after reading this article https://webpack.js.org/guides/tree-shaking/#conclusion

With commonjs transform: 891kb
Screen Shot 2021-07-30 at 12 52 47 PM

Without commonjs transform: 231kb 
Screen Shot 2021-07-30 at 12 47 55 PM

Oddly was having issues with the SVG loader after it was no longer being run through the commonjs transform. Rolling back a version appears to solve that. https://github.com/visualfanatic/vue-svg-loader/issues/121

Since this is kindof a big change I think it'd be good to get some testing in on dev before shipping to prod.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -17,7 +17,6 @@
 	"plugins": [
 		"@babel/plugin-syntax-dynamic-import",
 		"@babel/plugin-syntax-import-meta",
-		"@babel/plugin-transform-modules-commonjs",
 		[
 			"@babel/plugin-transform-runtime",
 			{
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,6 @@ updates:
   - dependency-name: vue-instantsearch
     versions:
     - ">= 2.7.a"
+  - dependency-name: vue-svg-loader
+    versions:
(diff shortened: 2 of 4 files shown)


ryan-ludwig:
I initially added the plugin-transform-modules-commonjs as part of https://github.com/kiva/ui/pull/2429 😰  . Tested the phone input and it appears to work correctly without it.

mcstover (approved):

eddieferrer (approved):
Looks good to me, lets give it a test. I suggest we add a line to '.github/dependabot.yml' to ignore vue-svg-loader > 0.15 for now, since this will be easy to miss and dependabot will generate a PR for the upgrade.

eddieferrer:
Alternatively. In that vue-svg-loader issue, it is suggested that we try: ``Using 0.17.0-beta.1 and swapping out babel-loader for vue-loader in my configuration also seems to work.``

emuvente (approved):

PR #3125, opened 2021-07-30
Add fallback to country level lat long if loan level lat long is miss…

mcstover:
…ing.

Add comment for fallback


mcstover:
Verified on `/lend-beta/2204683` in the vm.

emuvente (approved):
👍

BoulderBrains (commented):
👀

PR #3131, opened 2021-08-03
Use while loop to empty tracking event queue

emuvente:
The condition in the for loop checked `queue.items.length` every time the loop ran, however that length was constantly getting smaller as items are removed. So when you start with 3 items in the queue, you eventually end up with `i == 2` and `queue.items.length == 1`, so the loop is skipped and the last event in the queue does not get sent. Using a while loop that only checks if the queue is empty will ensure that all items in the queue get processed.

### DIFF
--- a/src/client-entry.js
+++ b/src/client-entry.js
@@ -157,7 +157,6 @@ router.onReady(() => {
 		app.$Progress.finish();
 		// fire pageview
 		app.$fireAsyncPageView(to, from);
-		app.$kvFireQueuedEvents();
 	});
 
 	router.onError(() => app.$Progress.fail());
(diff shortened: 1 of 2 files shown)


src/client-entry.js
```diff
@@ -157,7 +157,6 @@ router.onReady(() => {
 		app.$Progress.finish();
 		// fire pageview
 		app.$fireAsyncPageView(to, from);
-		app.$kvFireQueuedEvents();
```
emuvente, line 160: This is a typo, but also unneccessary because of the `app.$fireQueuedEvents()` call in the `app.$setKvAnalyticsData()` callback earlier in this file.

eddieferrer (approved):

BoulderBrains (approved):

mcstover (approved):
🤘🏼

PR #3133, opened 2021-08-03
PSD-264: flss data pull abstraction with tests

go-bears:
Tickets:
https://kiva.atlassian.net/browse/PSD-264

what does it do:
abstracts data fetch query from component to a reusable utils files
adds new test for utils file
refactors vue component to fetch data using new method.

How to test:
`npm run unit'` === `Test Suites: 22 passed, 22 total`


src/util/flssUtils.js
```diff
@@ -0,0 +1,30 @@
...
+import flssLoanQuery from '@/graphql/query/flssLoansQuery.graphql';
+import createApolloClient from '@/api/apollo';
+
+export function fetchData(loanQueryFilters) {
+	const zeroLoans = false;
+	const newLoans = [];
+	const totalCount = 0;
+
+	this.apollo.query({
+		query: flssLoanQuery,
```
go-bears, line 11: @ryan-ludwig I'm not sure why I'm getting an error of `"TypeError: Cannot read property 'apollo' of undefined"`
I thought apollo was provided to all the components from the `main.js` file? 

The other things I've tried was creating an apollo client within the function. But i also got an error there b/c I hadn't configured the apolloClient() correctly. The other thing I tried was writing the utils file like:
```js
export default {
inject: ['apollo'],
methods: { 
fetchData()
}
}
```
so I wasn't running into the apollo error but  lenderFilterAlpha.vue threw an error that flss.Utils.default.fetchData() was not a function.

go-bears: also @emuvente or @dymonda21  would appreciate any eyes on how to get the function imported 🙏

emuvente: Instead of using `this.apollo` I would recommend have apollo be a second argument to the fetchData function:
```
export function fetchData(loanQueryFilters, apollo) {
	...
	apollo.query ({ ...
```
and when you're calling it from a component you can pass is the instance of apollo that the component has:
```
fetchData(filters, this.apollo);
```

The way you have the import defined looks correct, but you could also try it this way:
```
export default function fetchData(loanQueryFilters, apollo) { ... }

import fetchData from '@/util/flssUtils';
```

go-bears: thanks y'alls!

test/unit/specs/util/flssUtil.spec.js
```diff
@@ -0,0 +1,30 @@
+import { fetchData } from '@/util/flssUtils';
+import flssLoanQuery from '@/graphql/query/flssLoansQuery.graphql';
+
+describe('flssUtils.js', () => {
+	describe('fetchData', () => {
```
emuvente, line 5: This is a great test!

I think you should also test the result returned by the function, which you can do with a couple additions to this test.
First, create a separate object to expect the fetchData function to return and add it to dataObj:
```
const result = { values: [], totalCount: 0 };
const dataObj = {
	data: { fundraisingLoans: result }
};
```

And then you can create another test that just checks for the result:
```
it('Returns the currently fundraising loans', done => {
	fetchData(loanQueryFilters, apollo).then(data => {
		expect(data).toBe(result);
		done();
	});
});
```

go-bears: great suggestions! will do!

emuvente (approved):
👍

go-bears:
thanks for the 👀 @emuvente @eddieferrer !

eddieferrer (commented):
👍

PR #3140, opened 2021-08-04
feat: New header frame for cause based landing pages

eddieferrer:
Adds new Content Group of type 'landingPageHero'
Creates frame and headline and subheadline for that content group
Adds tailwind wrapper class to some components from contentful page

Design reference:  https://www.figma.com/file/weQAkMDYoBA5ZnuHf4TPOA/Cause-Based-Landing-Page-Templates?node-id=136%3A331

SUBS-777

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -5,11 +5,12 @@
 	>
 		<template v-if="!pageError">
 			<component
-				v-for="({ component, content }) in contentGroups"
+				v-for="({ component, content, wrapperClass }) in contentGroups"
 				:key="content.key"
 				:is="component"
 				:content="content"
 				v-bind="getComponentOptions(content.key)"
+				:class="wrapperClass"
 			/>
 		</template>
 		<template v-else>
@@ -57,7 +58,7 @@ import * as siteThemes from '@/util/siteThemes';
 const WwwPage = () => import('@/components/WwwFrame/WwwPage');
 const WwwPageCorporate = () => import('@/components/WwwFrame/WwwPageCorporate');
 
-// Content Group components
+// Content Group Types
 // TODO: update the campaign components to accept "content" prop
 const CampaignHero = () => import('@/components/CorporateCampaign/CampaignHero');
 const CampaignLogoGroup = () => import('@/components/CorporateCampaign/CampaignLogoGroup');
@@ -67,8 +68,6 @@ const CampaignThanks = () => import('@/components/CorporateCampaign/CampaignThan
 const HomepageBottomCTA = () => import('@/components/Homepage/HomepageBottomCTA');
 const HomepageCorporateSponsors = () => import('@/components/Homepage/HomepageCorporateSponsors');
 // const HomepageGeneralStats = () => import('@/components/Homepage/HomepageGeneralStats');
-const DynamicHero = () => import('@/components/Contentful/DynamicHero');
-
 const HomepageHowItWorks = () => import('@/components/Homepage/HomepageHowItWorks');
 const HomepageLenderQuotes = () => import('@/components/Homepage/HomepageLenderQuotes');
 const HomepageLoanCategories = () => import('@/components/Homepage/HomepageLoanCategories');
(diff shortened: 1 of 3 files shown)


src/pages/ContentfulPage.vue
```diff
@@ -111,13 +113,26 @@ const getPageFrameFromType = type => {
...
 };
 
+// The components that return kv-tailwind will get wrapped in that class,
+// they have already been migrated to use tailwind CSS. Legacy components
+// should return empty string so they are not wrapped in the kv-tailwind class
+// once all components use tailwind, we can add the 'kv-tailwind' to the
+// :is="pageFrame" component on this page
+const getWrapperClassFromType = type => {
+	switch (type) {
+		case 'landingPageHero':
```
eddieferrer, line 123: Not sure what to name this component. Suggestions?

src/components/Contentful/LandingPageHero.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+	<section-with-background :background-content="heroBackground">
+		<template #content>
+			<!-- same wrapper as src/components/BorrowerProfile/ContentContainer.vue -->
```
eddieferrer, line 4: This is almost the same wrapper as `src/components/BorrowerProfile/ContentContainer.vue` Should we move ContentContainer out of BorrowerProfile, and try to make it more reusable since it might be used in other components outside of the borrower profile -- the only difference in this use case is the use of `lg:tw-col-span-6` instead of `lg:tw-col-span-7`

src/components/Contentful/LandingPageHero.vue
```diff
@@ -0,0 +1,73 @@
...
+			<kv-page-container>
+				<kv-grid class="tw-grid-cols-12">
+					<div class="tw-col-span-12 md:tw-col-start-2 md:tw-col-span-10 lg:tw-col-span-6">
+						<h1 v-html="heroHeadline">
+						</h1>
+						<h3 v-html="heroSubHeadline" class="tw-pt-2.5 lg:tw-pt-2">
```
eddieferrer, line 10: @ryan-ludwig any tailwinds best practices here on whether to use pt- class on the h3 or pb class on the h1 to add space between these 2 elements?

src/pages/ContentfulPage.vue
```diff
@@ -111,13 +113,26 @@ const getPageFrameFromType = type => {
 	}
 };
 
+// The components that return kv-tailwind will get wrapped in that class,
```
eddieferrer, line 116: I think this will be a good way to migrate the contentful components over to use tailwinds

eddieferrer:
Screen Shot 2021-08-05 at 8 23 26 AM

src/components/Contentful/LandingPageHero.vue
ryan-ludwig: I don't think there's a best practice per say. My usual system is to only use margin bottoms on everything to keep things predictable and avoid margin collapse weirdness.

src/components/Contentful/LandingPageHero.vue
ryan-ludwig: My opinion is that the important parts are housed in `kv-page-container` and `kv-grid`. How the columns are used seems more page specific so what you have here seems good to me.

ryan-ludwig (approved):

BoulderBrains (approved):

eddieferrer:
Im going to rework this so that it doesnt add a new content type but reuses the dynamic hero type

mcstover:
@eddieferrer Let's talk about this before making moves on DynamicHero. We may want to keep them separate.

src/pages/ContentfulPage.vue
mcstover: Maybe "heroWithCarousel" ...

mcstover (approved):
Pending any naming changes and I'm also ok with the existing name, good to go. The kv-tailwind class application will be very useful!

mcstover (approved):

emuvente (approved):
👍

PR #3148, opened 2021-08-06
core-14 Page frame with header, route for page build out

BoulderBrains:
[Core-14](https://kiva.atlassian.net/browse/CORE-14)
(for epic: https://kiva.atlassian.net/browse/CORE-2)

Route for the new version of the homepage we're building out. I'm using the /homepage-classic URL for the build out of this page, will be removed later.


src/router/routes.js
```diff
@@ -7,6 +7,14 @@ export default [
 			contentfulPage: () => 'home',
 		},
 	},
+	{
+		path: '/homepage-classic',
+		name: 'homepage',
+		component: () => import('@/pages/Homepage/HomepageClassic'),
+		meta: {
+			contentfulPage: () => 'homepage',
```
mcstover, line 15: The value returned here is the page key in contentful. Maybe we should make this `home-classic`  for a more clear distinction.

mcstover (approved):

PR #3163, opened 2021-08-12
feat: Adds the ability of the rich text renderer to render KvButtons

eddieferrer:
Test page: https://www.dev.kiva.org/lp/eddie-test-landing

- Renders kvButton components from embedded or inline button entries
- Wraps rich text output in kv-tailwind class to fix rendering

SUBS-789

### DIFF
--- a/src/util/contentful/richTextRenderer.js
+++ b/src/util/contentful/richTextRenderer.js
@@ -10,25 +10,32 @@ import { BLOCKS, INLINES } from '~/@contentful/rich-text-types';
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
 
 /**
- * Returns html string to render a contentful asset as a vue component
+ * Returns html string from rich text nodes
  *
- * @param {object} contentfulAssetObject Content of a contentful asset object
+ * @param {object} content Content of a contentful rich text field
  * @returns {string} String of html to render
  */
-const assetRenderer = contentfulAssetObject => {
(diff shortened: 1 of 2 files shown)


src/util/contentful/richTextRenderer.js
```diff
@@ -10,50 +10,77 @@ import { BLOCKS, INLINES } from '~/@contentful/rich-text-types';
...
+				<div class="tw-prose">${richTextRenderer(contentfulEntryNode?.data?.target?.fields?.richText)}</div>
+			`;
+		}
+		if (isButton) {
+			const analytics = contentfulEntryNode?.data?.target?.fields?.analyticsClickEvent ?? [];
+			return `
+				<kv-button
+					variant="${contentfulEntryNode?.data?.target?.fields?.style ?? 'primary'}"
+					href="${contentfulEntryNode?.data?.target?.fields?.webLink ?? '#'}"
+					v-kv-track-event="['${analytics.join("','")}']"
```
eddieferrer, line 70: This is not working, I'm looking into it

mcstover: Interesting. Do you get the correct markup here?

mcstover (approved):
Another great step forward!

src/util/contentful/richTextRenderer.js
eddieferrer: It is working, made some slight modifications so we arent firing the kv-track-event directive when the field does not exist in contentful

PR #3165, opened 2021-08-13
Adds check against hasDeposits query. If this is false we can assume …

mcstover:
…they are and FTD and in the case of a subscriber have a pending deposit.


BoulderBrains (approved):

emuvente (approved):
👍

PR #3180, opened 2021-08-24
Core 9 dynamic hero classic

mcstover:


mcstover:
Just to note, All of the image presentation and logic was directly ported from our existing DynamicHero component.

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -0,0 +1,335 @@
...
+								loop
+								muted
+								playsinline
+							></video>
+						</template>
+					</div>
+
+					<div
+						class="tw-col-span-12 md:tw-col-span-6 xl:tw-col-span-7"
+						:class="{ 'tw-order-first': swapOrder }"
```
mcstover, line 86: We can now swap the layout reversing the order of the 2 content blocks.

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -0,0 +1,335 @@
+<template>
+	<section-with-background-classic
+		:background-content="background"
+		:vertical-padding="verticalPadding"
+	>
+		<template #content>
+			<kv-page-container>
+				<kv-grid
+					class="tw-grid-cols-12 tw-gap-2 tw-items-center"
+					:class="{ 'tw-grid-cols-1 tw-mx-auto': singleColumn }"
```
mcstover, line 10: We can now specify if this should be a single column layout instead of a 2 column grid.

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -0,0 +1,335 @@
...
+					:style="customGridStyles"
```
mcstover, line 11: This is a garbage pail of styles that are easier to manage directly rather than using tailwind classes. We start with options for max width and text alignment to support the "It's a Loan not a donation" layout.

mcstover: These are NOT automatic, the will require configuration to setup and apply with any new styles we want to pass from Contentful.

BoulderBrains (approved):
🧙‍♂️ ✔️

src/components/Contentful/DynamicHeroClassic.vue
eddieferrer: Can't we just do the 'single column' layout using the centeredRichText type?

mcstover: This component supports expanded image options and specific content areas that the centeredRichText component does not. This component will for all intents and purposes remove the need for the centeredRichText version as it can do that too!

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -0,0 +1,335 @@
...
+				return contentType ? contentType === 'background' : false;
+			});
+		},
+		buttonAnalytics() {
+			const defaults = ['Hero', 'click-hero-cta', this.buttonText];
+			const contentfulAnaltyicsEvent = this.buttonContent?.analyticsClickEvent ?? null;
+			return contentfulAnaltyicsEvent || defaults;
+		},
+		// TODO: Consider Deprecating
+		buttonClass() {
```
eddieferrer, line 167: This was used to listen to the button for custom actions, like the MG interactive selector, not sure we can deprecate easily

eddieferrer:
Looks good! I have long term concerns with overloading the UI settings with all these settings attributes, but thats a conversation for another day. 

Can we possibly deprecate the DynamicHero component? Not sure if we can just swap this in.

src/components/Contentful/DynamicHeroClassic.vue
mcstover: Oh good catch that's what we needed to know!

PR #3199, opened 2021-08-27
Allow textAlign option to override default centered text alignment in…

mcstover:
… centeredRichText component


BoulderBrains (commented):
✔️

PR #3204, opened 2021-09-02
Fix memjs options argument usage

emuvente:
The third argument for `memjs.set()` is [supposed to be an options object](https://github.com/memcachier/memjs#settings-values), with a single option `expires`. Many of the current usages are passing in a number, with the assumption that the number is being used for `expires`. That's not the case. What actually happens is `memjs.set()` will ignore invalid options arguments and use 0 for `expires` (meaning never expire!). Luckily this hasn't caused a problem yet because we use the release hash in our vue component cache keys and the graphql fragment types have never changed!

### DIFF
--- a/server/util/initCache.js
+++ b/server/util/initCache.js
@@ -14,13 +14,16 @@ function FakeMemcached(options) {
 
 	// Replace 'set' with a memcached-compatible wrapper
 	const oldSet = lru.set;
-	lru.set = (key, value, age, callback) => {
+	lru.set = (key, value, { expires }, callback) => {
 		try {
-			oldSet.call(lru, key, value, age * 1000);
+			oldSet.call(lru, key, value, (expires ?? 0) * 1000);
 		} catch (error) {
 			if (callback) {
 				callback(error);
 			}
+		}
+		if (callback) {
+			callback(null, true);
 		}
 	};
 
--- a/server/util/memJsUtils.js
+++ b/server/util/memJsUtils.js
@@ -20,8 +20,7 @@ module.exports = {
 				if (error) {
 					log(`MemJS Error Setting Cache for ${key}, Error: ${error}`, 'error');
 					reject();
-				}
-				if (success) {
+				} else {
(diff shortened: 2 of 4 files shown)


server/util/initCache.js
```diff
@@ -14,14 +14,17 @@ function FakeMemcached(options) {
...
-			oldSet.call(lru, key, value, age * 1000);
+			oldSet.call(lru, key, value, (expires ?? 0) * 1000);
 		} catch (error) {
 			if (callback) {
 				callback(error);
 			}
 		}
+		if (callback) {
+			callback(null, true);
+		}
```
emuvente, line 27: This was missing the success callback! This prevented me from successfully testing live loans locally for a little while.

server/util/memJsUtils.js
```diff
@@ -20,8 +20,7 @@ module.exports = {
 				if (error) {
 					log(`MemJS Error Setting Cache for ${key}, Error: ${error}`, 'error');
 					reject();
-				}
-				if (success) {
+				} else {
```
emuvente, line 23: If `success` was false then `resolve()` would never be called

server/util/vueSsrCache.js
```diff
@@ -28,20 +26,12 @@ module.exports = function wrapper(cache) {
 			});
 		},
 		set(key, value) {
-			cache.set(key, value.html, 1 * 60 * 60, (error, success) => {
+			cache.set(key, value.html, { expires: 0 }, (error, success) => {
```
emuvente, line 29: Because the release hash is included in the component cache keys, it's ok for them to never expire (this is the same behavior as what is already happening, just explicit about it instead)

mcstover (approved):
👍

eddieferrer (approved):

ryan-ludwig (approved):

BoulderBrains:
👀

BoulderBrains (approved):

PR #3206, opened 2021-09-02
refactor: auto deposit landing page

eddieferrer:
* Refactor _get on page
* Refactor content group naming and access content groups with processPageContent

This was the only page that had ContentGroups that were camelCased. This PR just standardizes some of the naming and accessing of the content groups on the page, as well as use processPageContent and remove lodash/_get

### DIFF
--- a/src/pages/AutoDeposit/AutoDepositLandingPage.vue
+++ b/src/pages/AutoDeposit/AutoDepositLandingPage.vue
@@ -79,9 +79,8 @@
 </template>
 
 <script>
-import _get from 'lodash/get';
 import gql from 'graphql-tag';
-import { processPageContentFlat } from '@/util/contentfulUtils';
+import { processPageContent } from '@/util/contentfulUtils';
 
 import WwwPage from '@/components/WwwFrame/WwwPage';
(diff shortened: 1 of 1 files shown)


src/pages/AutoDeposit/AutoDepositLandingPage.vue
```diff
@@ -137,33 +136,49 @@ export default {
...
 			return !this.isMonthlyGoodSubscriber && !this.hasAutoDeposits && !this.hasLegacySubscription;
 		},
+		faqContentGroup() {
+			return this.contentGroups?.find(({ type }) => {
+				return type ? type === 'frequentlyAskedQuestions' : false;
+			});
+		},
+		ctaContentGroup() {
+			return this.contentGroups?.find(({ key }) => {
+				return key ? key === 'auto-deposit-cta' : false;
```
mcstover, line 162: Can we target this based on a type instead? Same with the key matching below?

eddieferrer: These contentGroups don't have a type.

mcstover: Ok I see they would need a "type" and prob some deeper content refactor to accommdate that.

mcstover (approved):

PR #3213, opened 2021-09-08
VUE-682 Use shared library for build pipeline

emuvente:
Tested using the `main` branch. The library being used has been added to the kiva-ui jenkins project configuration. For now this is using the exact same pipeline stages and steps; the only difference is that the pipeline code lives in the shared library repo. That allows using the shared libary without dealing with the risk of issues introduced by changing the actual steps taken. Future work includes making a generic pipeline that can be used by any nodejs apps and adapting the ui-server to use that generic pipeline.


mcstover (approved):
Fantastic!

PR #3214, opened 2021-09-09
feat: Render contentful preview API results when preview query param …

eddieferrer:
…is active for ContentfulPage component

VUE-671

We can add a custom URL per content type for the preview functionality in contentful, this PR only deals with the contentful page content type, allowing users to hit the preview button on contentful and preview a page -- contentful is currently set to go to `/{entry.fields.key}?preview=true`

I tried several approaches to this. My goal was to try to make it as invisible and removed from the rest of the code as possible, and to also have the capabilities to expand this in the future to other content types besides pages. 

Approaches: 
_1 - Have separate routes and components - for example `/preview/entryId` then, render that accordingly._
This was a considerable amount of work, as all the logic in contentfulPage would have to be duplicated. 

_2 - Use a custom link in apollo client to read the query and add the `preview: true` variable if the preview query param is present._
I really like this approach, as it would completely abstract the preview functionality from the vue templates and would be handled entirely by apolloClient. I could not get this to work because I couldn't get the apolloLink to know about the query param. I suppose we could possible get the query param, put that in a cookie, and then have apolloLink access that?

_3 - Current approach._
Modifies the query in all the appropriate places based on the query param. I dislike that if preFetch is a function that returns a promise, theres no easy way to modify the variables in that query (like in ContentfulPage) so we have to add the conditional variable there: `preview: args?.route?.query?.preview === 'true'`

I also don't like having to include code to deal with preview in both apolloPreFetch and apollo-plugin. 

The upside is that this code does work for both `preFetch: true` and `preFetch function` apollo configs.

Future Improvements: 
Add separate routes for previewing other content group types. These would be preview only routes for example `/preview/content-group/{key}` that would let you preview that content in isolation of the rest of the page

### DIFF
--- a/src/graphql/query/contentful.graphql
+++ b/src/graphql/query/contentful.graphql
@@ -1,6 +1,0 @@
-query contentful($contentType: String!, $contentKey: String) {
-	contentful {
-		entries(contentType: $contentType, contentKey: $contentKey)
-	}
-}
-
--- a/src/graphql/query/contentfulEntries.graphql
+++ b/src/graphql/query/contentfulEntries.graphql
@@ -1,0 +1,8 @@
+# The preview parameter should come from a route query param and
+# should not be passed in manually. It is handled by: src/util/apolloPreFetch.js on the server side and
+# src/plugins/apollo-plugin.js on the client
+query contentfulEntries($contentType: String!, $contentKey: String, $preview: Boolean) {
+	contentful {
+		entries(contentType: $contentType, contentKey: $contentKey, preview: $preview)
+	}
+}
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -49,11 +49,11 @@ To use, simply create a route that defines contentfulPage in the meta data, e.g.
 },
 */
 
-import gql from 'graphql-tag';
 import { preFetchAll } from '@/util/apolloPreFetch';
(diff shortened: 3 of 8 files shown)


emuvente (approved):
Good solution!

My thought to make approach 2 work would be to pass the entire VueRouter instance from `createRouter` to `createApolloClient` so that it could be used by a link to see what the state of the current route is when the api requests are made, but i don't know if that's an acceptable way to use VueRouter.

eddieferrer:
@emuvente Another approach could be to read the query param, put it in a cookie, then have an Apollo Link read that cookie and add the preview variable to contentful queries. Could this work?

emuvente:
@eddieferrer Yeah, using a cookie would work. Doing that would also allow making the preview mode be always-on while the cookie is set. Not sure if that would actually be desired though, but it could be useful. If not desired, you would need to make sure to delete that cookie if preview=true was not included in the route.

eddieferrer:
> @eddieferrer Yeah, using a cookie would work. Doing that would also allow making the preview mode be always-on while the cookie is set. Not sure if that would actually be desired though, but it could be useful. If not desired, you would need to make sure to delete that cookie if preview=true was not included in the route.

I will try this approach

eddieferrer:
See https://github.com/kiva/ui/pull/3220 for PR using cookies

PR #3215, opened 2021-09-09
fix: styling changes for cause landing pages and rich text support fo…

eddieferrer:
…r soft breaks

GD-113

This will need to be included in the next release. 

* Makes the heroWithCarousel component and the dynamicHeroClassic be 50%/50% column on xl screens
* Adds support to the rich text renderer for soft breaks (shift+enter) on the contentful rich text editor. See: https://github.com/kiva/ui/pull/2386

### DIFF
--- a/src/components/Contentful/DynamicHeroClassic.vue
+++ b/src/components/Contentful/DynamicHeroClassic.vue
@@ -11,7 +11,7 @@
 					:style="customGridStyles"
 				>
 					<div
-						class="tw-mx-auto tw-col-span-12 md:tw-col-span-6 xl:tw-col-span-5"
+						class="tw-mx-auto tw-col-span-12 md:tw-col-span-6"
 					>
 						<!-- eslint-enable max-len -->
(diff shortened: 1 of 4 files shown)


emuvente (approved):

mcstover (approved):

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -11,7 +11,7 @@
 					:style="customGridStyles"
 				>
 					<div
-						class="tw-mx-auto tw-col-span-12 md:tw-col-span-6 xl:tw-col-span-5"
+						class="tw-mx-auto tw-col-span-12 md:tw-col-span-6"
```
mcstover, line 14: Did you consult with Design on this change? I think this will now change the expected layout for the homepage hero instances...

ryan-ludwig (approved):

BoulderBrains:
👀

PR #3217, opened 2021-09-10
feat: change eco friendly label in UI

eddieferrer:
* Changes the 'eco friendly' label to 'climate' in the UI without touching the MG value or loan channel

GD-112

I think this is what we want. The url for the category will still be `/monthlygood?category=eco_friendly` but the drop downs and interactive selector will all display `climate` this is because the query param maps directly to the value we send to the API


mcstover (approved):
Yep I think this is what we needed

PR #3223, opened 2021-09-14
Tailwind JIT/purge fix

ryan-ludwig:
During development sometimes I'd add a new Tailwind class that hadn't been used yet and it wouldn't be added to the CSS file. Came across a few issues on Github with similar symptoms: https://github.com/tailwindlabs/tailwindcss/issues/4081. 
Then found some more info on the Tailwind JIT docs. https://tailwindcss.com/docs/just-in-time-mode#styles-don-t-update-when-saving-content-files

Bumping Tailwind and adding the `TAILWIND_MODE=watch` var seems to work for me now. Though not always with HMR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,7 +195,7 @@
 		"stylelint": "^13.7.0",
 		"stylelint-config-standard": "^20.0.0",
 		"stylelint-scss": "^3.20.1",
-		"tailwindcss": "2.2.4",
+		"tailwindcss": "2.2.15",
 		"thread-loader": "^2.1.3",
 		"url-loader": "^4.1.1",
 		"vue-jest": "^3.0.7",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"private": true,
 	"license": "UNLICENSED",
 	"scripts": {
-		"dev": "cross-env NODE_ENV=development node server/dev-server.js",
+		"dev": "cross-env NODE_ENV=development TAILWIND_MODE=watch node server/dev-server.js",
 		"start": "cross-env NODE_ENV=production node --max-old-space-size=2048 server/index.js",
 		"start-k8s": "cross-env NODE_ENV=production node --max-old-space-size=1536 server/index.js",
 		"docker-dev-vm": "rimraf node_modules/.cache/hard-source && cross-env NODE_ENV=development node server/dev-server.js",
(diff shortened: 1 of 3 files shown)


tailwind.config.js
```diff
@@ -3,13 +3,11 @@ const kvConfig = require('@kiva/kv-components/tailwind.config.js');
 module.exports = {
 	mode: 'jit',
 	presets: [kvConfig],
-	purge: {
```
mcstover, line 6: This syntax previously "fixed" the purging... is the updated syntax you have specific to the new version?

package.json
```diff
@@ -6,7 +6,7 @@
 	"private": true,
 	"license": "UNLICENSED",
 	"scripts": {
-		"dev": "cross-env NODE_ENV=development node server/dev-server.js",
+		"dev": "cross-env NODE_ENV=development TAILWIND_MODE=watch node server/dev-server.js",
```
mcstover, line 9: Happy to see this as I'd seen a similar "miss"

tailwind.config.js
mcstover: None of their examples are using this object with the `content` array now...

mcstover: I think we got this syntax from the PurgeCss docs: https://purgecss.com/configuration.html#configuration-file

ryan-ludwig: Yeah, I think it was actually the act of changing anything in the purge setting that caused Tailwind to pick up the new styles for the previous fix. Some people's workarounds to this bug were to add an empty directory to the list and save. Then remove it and save. To get Tailwind to regenerate all the classes.

mcstover (approved):
I'm cool with this if it's working well for you

tailwind.config.js
eddieferrer: > None of their examples are using this object with the `content` array now...

https://tailwindcss.com/docs/optimizing-for-production#enabling-manually

@ryan-ludwig @mcstover I think we still need the content array... I'll have to test My previous fix was specifically fixing the issue where you are using a class in UI that is not used in kv-ui-elements repo

eddieferrer:
I did some additional tests:

Adding a new class does not seem to work on HMR for me. With `TAILWIND_MODE=watch` and `tailwind.config.js` from master it seems to work on both restart and HMR.

ryan-ludwig:
@eddieferrer OK cool, thanks for testing! I'll try out reverting the changes in `tailwind.config.js` locally.

mcstover (approved):

ryan-ludwig:
I still don't get the new classes built in HMR, but I wasn't getting that before either. So if it's working better for you with the original purge settings @eddieferrer let's roll with this.

emuvente (approved):

PR #3225, opened 2021-09-15
Core 5

BoulderBrains:
[Core-5](https://kiva.atlassian.net/browse/CORE-5)

Reusable component fed by contentful data that displays centered richText content.

**Mobile:** 
Screen Shot 2021-09-15 at 5 59 41 PM

**Tablet:** 
Screen Shot 2021-09-15 at 5 59 34 PM

**Desktop:**
Screen Shot 2021-09-15 at 5 59 25 PM


src/pages/ContentfulPage.vue
```diff
@@ -89,6 +90,8 @@ const KvFrequentlyAskedQuestions = () => import('@/components/Kv/KvFrequentlyAsk
 
 const TestimonialCards = () => import('@/components/Contentful/TestimonialCards');
 
+const MultiStepCentered = () => import('@/components/Contentful/RichTextItemsCentered');
```
mcstover, line 93: Let's use `RichTextItemsCentered` for the Component Name and then also add `richTextItemsCentered` to the Content Group in Contentful.

src/pages/ContentfulPage.vue
```diff
@@ -142,7 +146,7 @@ const getComponentFromType = type => {
 		case 'homepageBottomCTA':
 			return HomepageBottomCTA;
 		case 'homepageHowItWorks':
-			return HomepageHowItWorks;
+			return MultiStepCentered;
```
mcstover, line 149: We'll need to preserve the previous switch here and add a new case switch for `RichTextItemsCentered` return `richTextItemsCentered`.

src/pages/ContentfulPage.vue
```diff
@@ -69,7 +70,7 @@ const CampaignThanks = () => import('@/components/CorporateCampaign/CampaignThan
 const HomepageBottomCTA = () => import('@/components/Homepage/HomepageBottomCTA');
 const HomepageCorporateSponsors = () => import('@/components/Homepage/HomepageCorporateSponsors');
 // const HomepageGeneralStats = () => import('@/components/Homepage/HomepageGeneralStats');
-const HomepageHowItWorks = () => import('@/components/Homepage/HomepageHowItWorks');
+// const HomepageHowItWorks = () => import('@/components/Homepage/HomepageHowItWorks');
```
mcstover, line 73: We need to keep this intact for now.

src/pages/ContentfulPage.vue
```diff
@@ -54,6 +54,7 @@ import { preFetchAll } from '@/util/apolloPreFetch';
 import { processPageContent } from '@/util/contentfulUtils';
 import logFormatter from '@/util/logFormatter';
 import * as siteThemes from '@/util/siteThemes';
+import contentfulStylesMixin from '@/plugins/contentful-ui-setting-styles-mixin';
```
mcstover, line 57: is this needed? Why?

src/pages/ContentfulPage.vue
```diff
@@ -235,6 +239,7 @@ export default {
 			title: this.title,
 		};
 	},
+	mixins: [contentfulStylesMixin],
```
mcstover, line 242: Same... maybe not needed...

BoulderBrains:
@mcstover all your comments have been addressed in my latest commit.

mcstover (approved):
So concise!

src/components/Contentful/RichTextItemsCentered.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+	<section-with-background-classic
+		class="how-it-works"
```
emuvente, line 3: This class can be dropped (I'm guessing this is leftover from copying)

emuvente (approved):
Well done! Great to have this component now

eddieferrer:
This is great, very concise. 

I think we can refactor this slightly to allow for any number of rich text items in columns by following these steps:
- change `lg:tw-col-span-3` to be based on the length of richContentfulContent array
- replace the `figure` element for a generic div - I know semantically in this case it works, but to make this more general I think a div would be fine. 

@BoulderBrains I was also looking at the content on contentful -- please make sure to remove the Generic Content Block that is the Content Items for this entry, since this will be confusing, and I believe the "heading" is rendered by a totally different entry in contentful

BoulderBrains:
@eddieferrer Thanks for the review.
- I've replaced `figure` with a `div`
- I also removed the headings from the `hp-classic-how-it-works` contentGroup in contentful. Thanks for mentioning this.

I don't think the suggested change of
_change lg:tw-col-span-3 to be based on the length of richContentfulContent array_
works though. Currently the length of richContentfulContent array is 4, which doesn't return the right column display when rendered on a large screen.
I tested that change locally with `:class="`tw-col-span-12 lg:tw-col-span-${richContentfulContent.length}`"`

Did I understand you suggestion correctly?

eddieferrer:
> @eddieferrer Thanks for the review.
> 
> * I've replaced `figure` with a `div`
> * I also removed the headings from the `hp-classic-how-it-works` contentGroup in contentful. Thanks for mentioning this.
> 
> I don't think the suggested change of
> _change lg:tw-col-span-3 to be based on the length of richContentfulContent array_
> works though. Currently the length of richContentfulContent array is 4, which doesn't return the right column display when rendered on a large screen.
> I tested that change locally with `:class="`tw-col-span-12 lg:tw-col-span-${richContentfulContent.length}`"`
> 
> Did I understand your suggestion correctly?

I was thinking it should be: 
:class="tw-col-span-12 lg:tw-col-span-${12/richContentfulContent.length}"

src/components/Contentful/RichTextItemsCentered.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+	<section-with-background-classic
+		:background-content="sectionBackground"
+		:vertical-padding="verticalPadding"
+	>
+		<template #content>
+			<kv-page-container>
+				<kv-grid class="tw-grid-cols-12 tw-text-center">
```
eddieferrer, line 8: We'll also need to add `:style="customGridStyles"` to this kv-grid. This will allow other uses of this content group where we can pass a setting of `align: left` and overwrite tw-text-center

BoulderBrains:
Thanks for the review team 👏 @eddieferrer, all those changes are now implemented.

mcstover (approved):

PR #3229, opened 2021-09-16
fix: minor webpack dev issues

eddieferrer:
While doing research and testing on: https://github.com/kiva/ui/pull/3223 I came across these 2 minor webpack improvements. 

Improving ExtractCSS which should not use hashes for dev - https://webpack.js.org/plugins/mini-css-extract-plugin/#advanced-configuration-example
Refactoring deprecated compiler hook - reference: https://webpack.js.org/api/compiler-hooks/


mcstover (approved):
I had read about removing the hash for dev mode recently as well. Should also be applicable for the js, but this is a fine improvement.

PR #3232, opened 2021-09-16
feat: Adds custom event fields to buttons in RTF's

eddieferrer:
* Adds a field on contentful Button types. 'webClickEventName'
* For the Rich Text Renderer, if this field is present, render the button with an onclick which passes this value up to dynamic rich text component
* In dynamic rich text component, emit a global event of the same name
* TLDR: This allows a button inside of a rich text field in contentful to open up the monthly good interactive selector.

GD-116

### DIFF
--- a/src/components/Contentful/DynamicRichText.vue
+++ b/src/components/Contentful/DynamicRichText.vue
@@ -20,9 +20,17 @@ export default {
 				},
 				props: [
 					/* define props as needed */
-				]
+				],
+				methods: {
+					buttonClick(customEventName, event) {
+						// Current behavior is to replace a button navigation if a custom event name is passed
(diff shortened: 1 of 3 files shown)


mcstover (approved):
Looks good. I suppose for independent usages of the button we could do something similar. Maybe a future update is to encapsulate the logic from richTextRenderer into a "ContentfuButton" wrapper component.

eddieferrer:
> Looks good. I suppose for independent usages of the button we could do something similar. Maybe a future update is to encapsulate the logic from richTextRenderer into a "ContentfuButton" wrapper component.

Yeah definitely. We currently don't have a use case for a ContentfulButton but I could see one in the future

PR #3235, opened 2021-09-20
Core 24 kc basic loan card

BoulderBrains:
[Core-24](https://kiva.atlassian.net/browse/CORE-24) (starting pieces)
This PR covers the subtask Core-81, core-88
& core-84

Good morning Matt, I thought I'd stage up this PR so we can get it into master and keep working on it. I've got everything in place for the amount to go portion of the loan card, but I'm going to wait until this frame is in place in master before including.

Was there anything else we needed to add to this?


mcstover (approved):
Good to get this started. FYI @emuvente I essentially started with your Recommended loan card and plan to start to abstract out the data fetch and some formatting into utils.

emuvente (approved):

emuvente:
Coveralls is in maintenance mode right now, which caused the coverage report upload to fail, which is why the ui test is failing and why the coveralls check hasn't completed. It's ok to ignore those checks in this case.

BoulderBrains:
@emuvente @mcstover I've actually not able to merge this without the coveralls test to past, as it's required. Is there a way around that?

PR #3241, opened 2021-09-20
core-98 adding loanUse to KivaClassicBasicLoanCard

BoulderBrains:
[Core-98](https://kiva.atlassian.net/browse/CORE-98)

Added loanUse to the KivaClassicBasicLoanCard.

*loanUse truncated to 52 characters, by passing loan-use-max-length into the loan-use component, which uses loan-use-filter.js

Screen Shot 2021-09-21 at 5 08 19 PM


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -59,6 +59,13 @@
 			class="tw-mb-1.5" :style="{width: '100%', height: '5.5rem'}"
 		/>
 
+		<div
+			v-if="!isLoading"
+			class="tw-mb-1.5 lg:tw-mb-1"
+		>
+			{{ loanUse }}
```
mcstover, line 66: I think we need to determine the character count restriction for this field, then truncate and add an ellipsis if we cut the content at that limit.

BoulderBrains: Will add that and have it up shortly.

mcstover (approved):

PR #3247, opened 2021-09-21
Core 99

BoulderBrains:
[Core-99](https://kiva.atlassian.net/browse/CORE-99)

New loan matcher component built using tailwinds and inserted into KivaClassicBasicLoanCard

Screen Shot 2021-09-21 at 2 17 21 PM

Screen Shot 2021-09-21 at 2 17 30 PM

### DIFF
--- a/src/components/LoanCards/LoanMatchingText.vue
+++ b/src/components/LoanCards/LoanMatchingText.vue
@@ -4,7 +4,7 @@
 			class="
 				tw-text-h4
 				tw-mr-1
-				tw-pl-0.5
+				tw-text-center
 				tw-h-3
 				tw-w-3
 				tw-rounded-full
--- a/src/components/LoanCards/LoanMatchingText.vue
+++ b/src/components/LoanCards/LoanMatchingText.vue
@@ -1,0 +1,72 @@
+<template>
+	<span class="tw-flex tw-items-center">
+		<h4 class="tw-inline-block ">
+			<span
+				class="
+				tw-mr-1
+				tw-p-1
+				tw-h-3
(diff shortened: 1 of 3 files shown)


src/components/LoanCards/LoanMatchingText.vue
```diff
@@ -0,0 +1,72 @@
...
+		<h4 class="tw-inline-block ">
+			<span
+				class="
+				tw-mr-1
+				tw-p-1
+				tw-h-3
+				tw-rounded-full
+				tw-bg-brand-100
+				tw-border-2
+				tw-border-brand
```
mcstover, line 12: I think you may need an explicit width on this too that matches the height to ensure it's always a circle. Both the width and the height seem like they should be a bit smaller based on the screenshot.

As for layout, you might consider wrapping the icon in it's own span to gain more control over it as well if need be.

BoulderBrains: I made some adjustments in my latest commit.

src/components/LoanCards/LoanMatchingText.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+	<span class="tw-flex tw-items-center">
+		<span
+			class="
+				tw-text-h4
+				tw-mr-1
+				tw-pl-0.5
```
ryan-ludwig, line 7: Rather then padding left to get the emoji in the right spot, would text-center get you there?

mcstover (approved):

emuvente (approved):
👍

PR #3252, opened 2021-09-22
core-5 fixing up cols on medium and large screens

BoulderBrains:
The column break was not working properly on dev. 

:class="`tw-col-span-12 lg:tw-col-span-${12/richContentfulContent.length}`"

This was because I was breaking the tailwinds class into pieces, which wasn't allowing the function to run before rendering, preventing tailwinds from working correctly, even though the right classes were rendered on the front end. 

The fix passes through full tailwind class names based on the calculation... which works. 
https://tailwindcss.com/docs/just-in-time-mode#arbitrary-value-support

* I spoke to Nathan and he did want to go with the 2 column view for medium view ports. 

**Small:**
Screen Shot 2021-09-22 at 12 26 04 PM

**Medium:** 
Screen Shot 2021-09-22 at 12 25 54 PM

**Large:**
Screen Shot 2021-09-22 at 12 25 44 PM

### DIFF
--- a/src/components/Contentful/RichTextItemsCentered.vue
+++ b/src/components/Contentful/RichTextItemsCentered.vue
@@ -12,7 +12,13 @@
 					<div
 						v-for="(item, index) in richContentfulContent"
 						:key="index"
-						:class="`tw-col-span-12 lg:tw-col-span-${12/richContentfulContent.length}`"
+						:class="{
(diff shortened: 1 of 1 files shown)


mcstover (approved):

ryan-ludwig (commented):
👍

PR #3255, opened 2021-09-22
feat: Allow horizontal alignment of grid items

ryan-ludwig:
This sets the max-width on the children of the grid instead of the grid.

Instead of `tw-mx-auto` on the grid element, it uses `tw-justify-items-...` to horizontally align the grid children and allow overrides from Contentul.

### DIFF
--- a/src/components/Contentful/DynamicHeroClassic.vue
+++ b/src/components/Contentful/DynamicHeroClassic.vue
@@ -6,12 +6,15 @@
 		<template #content>
 			<kv-page-container>
 				<kv-grid
-					class="tw-grid-cols-12 tw-gap-2 tw-items-center"
-					:class="{ 'tw-grid-cols-1 tw-mx-auto': singleColumn }"
+					class="tw-grid-cols-12 tw-gap-2 tw-items-center tw-justify-items-center"
 					:style="customGridStyles"
 				>
 					<div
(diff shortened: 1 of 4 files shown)


src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -6,12 +6,15 @@
...
+					class="tw-grid-cols-12 tw-gap-2 tw-items-center tw-justify-items-center"
 					:style="customGridStyles"
 				>
 					<div
-						class="tw-mx-auto tw-col-span-12 md:tw-col-span-6"
+						class="tw-mx-auto tw-col-span-12"
+						:class="{
+							'md:tw-col-span-6': !singleColumn
+						}"
+						:style="maxWidthStyles"
```
mcstover, line 17: It's kind of odd that the maxWidthStyles are not applying to the entire content area...I think there are designs in the pipeline that may need to restrict the width in that way.

mcstover (approved):

src/components/Contentful/DynamicHeroClassic.vue
ryan-ludwig: Good feedback, will work on that aspect

BoulderBrains (approved):

emuvente (approved):

PR #3256, opened 2021-09-22
fix: add tw-text-jumbo to the safelist since it's only used in conten…

ryan-ludwig:
…tful

### DIFF
--- a/tailwind.purge.safelist.txt
+++ b/tailwind.purge.safelist.txt
@@ -1,3 +1,4 @@
+tw-text-jumbo
 tw-pt-0
 md:tw-pt-0
 lg:tw-pt-0
--- a/tailwind.purge.safelist.txt
+++ b/tailwind.purge.safelist.txt
@@ -1,4 +1,4 @@
-tw-text-jumbo
+# Padding classes
 tw-pt-0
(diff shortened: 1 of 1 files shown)


ryan-ludwig:
This one had me scratching my head for a while, til I realized I couldn't find the class used in UI anywhere else.

mcstover:
@ryan-ludwig do you think we should go ahead and add the new themeable text classes

ryan-ludwig:
@mcstover Yeah, I think that makes sense to have all the themeable colors + all the type styles whitelisted. I'll update this PR.

mcstover (approved):

tailwind.purge.safelist.txt
```diff
@@ -38,3 +39,33 @@ tw-pb-8
 md:tw-pb-8
 lg:tw-pb-8
 xl:tw-pb-8
+
+# Themable color classes
+tw-bg-primary
```
ryan-ludwig, line 44: Hmm, now thinking we wouldn't ever set a `tw-bg-*` class within a contentful span.

mcstover (approved):

emuvente (approved):

PR #3258, opened 2021-09-23
Check for properties of each padding size before testing them as stri…

mcstover:
…ngs.

### DIFF
--- a/src/components/Contentful/SectionWithBackgroundClassic.vue
+++ b/src/components/Contentful/SectionWithBackgroundClassic.vue
@@ -150,10 +150,8 @@ export default {
 				if (key === 'sm') hasSmSizes = true;
 				if (key === 'lg') hasLgSizes = true;
 				// process sizes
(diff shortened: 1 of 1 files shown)


src/components/Contentful/SectionWithBackgroundClassic.vue
```diff
@@ -150,8 +150,10 @@ export default {
 				if (key === 'sm') hasSmSizes = true;
 				if (key === 'lg') hasLgSizes = true;
 				// process sizes
-				const top = sizeObj.top.toString() ? `${sizePrefix}tw-pt-${sizeObj.top}` : '';
-				const bottom = sizeObj.bottom.toString() ? `${sizePrefix}tw-pb-${sizeObj.bottom}` : '';
```
ryan-ludwig, line 154: I think this can be simplified with optional chaining to
```const bottom = sizeObj.bottom?.toString() ? `${sizePrefix}tw-pb-${sizeObj.bottom}` : '';```

ryan-ludwig (approved):

PR #3259, opened 2021-09-23
VUE-687 Move Dockerfile to jenkins library

emuvente:


mcstover (approved):
If this doesn't have to be removed for testing the new setup, we may want to keep it...or test the remove in the `test` env first...

emuvente:
> If this doesn't have to be removed for testing the new setup, we may want to keep it...or test the remove in the `test` env first...

We can hold off on merging this until we feel confident with the new location for it. I'm planning on making some significant changes to the Dockerfile, and keeping the changes in-sync between the two repos could get tricky, so I don't want to keep this file here for too long.

PR #3260, opened 2021-09-23
Design site: Ensure the top nav isn't covering the content when clicked

ryan-ludwig:
Since our nav is sticky at the top we need to scroll a little farther down the page.

### DIFF
--- a/src/components/WwwFrame/WwwPageDesign.vue
+++ b/src/components/WwwFrame/WwwPageDesign.vue
@@ -26,26 +26,40 @@
 						</div>
 
 						<!-- nav for mobile and desktop -->
-						<ul class="
+						<nav id="top-nav">
+							<ul class="
 								tw-flex tw-gap-2 tw-items-center md:tw-justify-end tw-py-2.5
 								tw-bg-secondary md:tw-bg-primary"
-						>
-							<li>
-								<kv-button variant="ghost" to="#creative-studio-logo-intro">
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/WwwPageDesign.vue
```diff
@@ -26,26 +26,40 @@
...
-							</li>
-						</ul>
+							>
+								<li>
+									<kv-button
+										variant="ghost"
+										href="#creative-studio-logo-intro"
+										@click.native.prevent="scrollPastNav('#creative-studio-logo-intro')"
+									>
+										Logosz
```
mcstover, line 40: `z`

src/components/WwwFrame/WwwPageDesign.vue
```diff
@@ -26,26 +26,40 @@
...
-									Colors
-								</kv-button>
-							</li>
-							<li>
-								<kv-button variant="ghost" to="#creative-studio-typography-intro">
-									Type
-								</kv-button>
-							</li>
-						</ul>
+							>
```
mcstover, line 33: Eventually, It would be useful to drive the nav via the dataObject in Ui Setting from contentful. Each object in the array would have the button text and the href which you could then cycle through to build the nav making it easy to manage.

mcstover (approved):

emuvente (approved):

PR #3261, opened 2021-09-23
VUE-702 Convert Cypress to run in real env without a mocked server

emuvente:
This changes the cypress testing to run against a real instead of mocked environment, and also removes the cypress test from the pull request checks. Cypress testing will now be run in Jenkins. 

The coverage drop is expected, because the Cypress tests aren't running in github anymore.

The commented code will be cleaned up in VUE-704 and VUE-707.

### DIFF
--- a/test/e2e/specs/AutolendingPage.spec.js
+++ b/test/e2e/specs/AutolendingPage.spec.js
@@ -1,7 +1,7 @@
 import { subDays } from 'date-fns';
 import wwwPageMock from '../fixtures/wwwPageMock';
 
-describe('Autolending Page Spec', () => {
+describe.skip('Autolending Page Spec', () => {
 	beforeEach(() => {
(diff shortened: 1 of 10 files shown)


test/e2e/specs/AutolendingPage.spec.js
```diff
@@ -1,7 +1,7 @@
 import { subDays } from 'date-fns';
 import wwwPageMock from '../fixtures/wwwPageMock';
 
-describe('Autolending Page Spec', () => {
+describe.skip('Autolending Page Spec', () => {
```
emuvente, line 4: Skipping this test suite for now because it requires authentication. This will be revisited and refactored after support for authentication in testing is added during VUE-721.

mcstover (approved):

BoulderBrains (approved):

PR #3262, opened 2021-09-23
Contentful component - Only apply a routerLink if we have a url

ryan-ludwig:


mcstover (approved):
I think this should be safe with existing usages. @eddieferrer

eddieferrer (approved):
This is fine. 👍

PR #3263, opened 2021-09-23
Core-102 adding the allSharesReserved state to KivaClassicBasicLoanCard

BoulderBrains:
Core-102 All shares reserved state of KivaClassicBasicLoanCard

When all shares are reserved we disable the button click, update the funding text to "Funding complete" and set the fundraising meter to 100%. 

Once we have the click interaction in place for the borrower image, that will also be disabled if allSharesReserved is true. 

Here it is:
Screen Shot 2021-09-23 at 12 55 42 PM

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -8,6 +8,11 @@
 			v-if="isLoading"
 			class="tw-mb-1 tw-rounded" :style="{width: '100%', height: '15.75rem'}"
 		/>
+
+		<!-- TODO: Once click interaction for this is in place, we need to disable the click
+		if allSharesReserved is true 
+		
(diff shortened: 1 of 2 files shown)


ryan-ludwig (approved):

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -8,6 +8,11 @@
 			v-if="isLoading"
 			class="tw-mb-1 tw-rounded" :style="{width: '100%', height: '15.75rem'}"
 		/>
+
+		<!-- TODO: Once click interaction for this is in place, we need to disable the click
+		if allSharesReserved is true 
+		
+		click="`${allSharesReserved ? 'correct redirect' : ''}`" -->
```
mcstover, line 15: Please fix these spacing issues

mcstover (approved):

PR #3265, opened 2021-09-23
Design Microsite: Add scrollspy to highlight what section we're in on the topnav

ryan-ludwig:


ryan-ludwig:
Will require the latest version of kv-components to be pulled in to see the `active` state of KvButton

mcstover (approved):

PR #3266, opened 2021-09-23
Core-107 click interactions and tracking events for KivaClassicBasicL…

BoulderBrains:
…oanCard

[Core-107](https://kiva.atlassian.net/browse/CORE-107)

- Click interaction added to borrower image by wrapping it in a router-link. 
      - There's a conditional in place that if allSharesReserved is true, we disable the < router-link > by converting it to a span tag.
- Tracking info added for borrower image click and "Read more" cta button click added based on spec in ticket.

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -136,7 +136,7 @@
 			:to="`/lend/${loanId}`"
 			v-kv-track-event="['Lending', 'click-Read-more', 'Read more', loanId]"
 		>
-			Read more
+			Learn more
 			<kv-material-icon
 				class="tw-align-middle"
 				:icon="mdiChevronRight"
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -9,7 +9,7 @@
 			class="tw-mb-1 tw-rounded" :style="{width: '100%', height: '15.75rem'}"
 		/>
 
-		<!-- If allSharesReserved, disable link by make it a span -->
+		<!-- If allSharesReserved, disable link by making it a span -->
 		<router-link
 			:is="allSharesReserved ? 'span' : 'router-link'"
 			:to="`${!allSharesReserved ? `/lend/${loanId}` : diable}`"
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -15,11 +15,10 @@
 			<!-- If allSharesReserved, disable link by making it a span -->
 			<router-link
 				:is="allSharesReserved ? 'span' : 'router-link'"
-				:to="`${!allSharesReserved ? `/lend/${loanId}` : diable}`"
-				v-kv-track-event="['KivaClassicBasicLoanCard', 'click-borrower-image', loanId]"
+				:to="`/lend/${loanId}`"
+				v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId]"
 			>
 				<borrower-image
-					v-if="!isLoading"
 					class="
 					tw-relative
 					tw-w-full
@@ -41,7 +40,7 @@
 				/>
 				<div v-if="countryName">
 					<summary-tag
-						class="tw-absolute tw-bottom-2 tw-left-1"
+						class="tw-absolute tw-bottom-2 tw-left-1 tw-text-primary"
 						:city="city"
 						:state="state"
 						:country-name="countryName"
@@ -135,7 +134,7 @@
 			class="tw-mb-2"
 			:state="`${allSharesReserved ? 'disabled' : ''}`"
 			:to="`/lend/${loanId}`"
-			v-kv-track-event="['KivaClassicBasicLoanCard', 'click-Read-more-CTA-button', loanId]"
+			v-kv-track-event="['Lending', 'click-Read-more', 'Read more', loanId]"
 		>
 			Read more
 			<kv-material-icon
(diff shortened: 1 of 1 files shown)


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -9,30 +9,32 @@
...
 		/>
 
-		<!-- TODO: Once click interaction for this is in place, we need to disable the click
-		if allSharesReserved is true
-		ie.
-		:click="`${allSharesReserved ? 'correct redirect' : ''}`" -->
-		<borrower-image
-			v-if="!isLoading"
-			class="
+		<!-- Is there a way to display the click completely if allSharesAreReserved? -->
```
mcstover: You can use a v-if on the router-link and it's containing borrower-image then have the v-else on an duplicated instance of the image when you don't want to show a click...I do question why we would turn off the link at all in this scenario. I don't think we should be removing this link...

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -9,30 +9,33 @@
...
-			:images="[
-				{ width: 336, viewSize: 1024 },
-				{ width: 336, viewSize: 768 },
-				{ width: 416, viewSize: 480 },
-				{ width: 374, viewSize: 414 },
-				{ width: 335, viewSize: 375 },
-				{ width: 280 },
-			]"
-		/>
+		<!-- If allSharesReserved, disable link by make it a span -->
```
mcstover, line 12: Why do we need to remove this link at all?

BoulderBrains: The requested allSharesReserved state from Nathan is "disabled button, (no click targets) - and funding complete". 

I mentioned that it felt odd to block users from learning more about a borrower. But he wants to keep this block in place until we get the recommended loan cards on funded borrower pages.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -3,52 +3,57 @@
...
-				:distribution-model="distributionModel"
-			/>
-			<summary-tag
-				v-if="countryName"
-				class="tw-absolute tw-bottom-2 tw-left-1"
-				:city="city"
-				:state="state"
-				:country-name="countryName"
+			<!-- If allSharesReserved, disable link by making it a span -->
+			<router-link
```
eddieferrer, line 16: Does this work? This is interesting syntax. I would expect this to be `<component :is="allSharesReserved ? 'span' : 'router-link'"`

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -3,52 +3,57 @@
...
+				:is="allSharesReserved ? 'span' : 'router-link'"
+				:to="`${!allSharesReserved ? `/lend/${loanId}` : diable}`"
```
eddieferrer, line 18: diable?

BoulderBrains: Thanks for calling out the redundancy, fixed that up.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
BoulderBrains: What I have in place does work, but I don't see a difference between what I have in there and what you suggested above. Am I misunderstanding you?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
BoulderBrains: Conversation discussed and resolved in Slack.

mcstover (approved):

BoulderBrains:
I'd like to get this merged into master branch and then work on the additional changes requested in slack and captured in CORE-110 & CORE-111

The conclusion of that discussion was the following work. 
- We will add the “Learn more” text link after the loan use statement,
- When all loan shares are reserved, 
     - we will replace the CTA button with the “another lender has selected this loan” message
     -  Hide the “Learn more” text link from #1
     - Remove the link from the image _(This change is present in this code review)_

src/components/LoanCards/KivaClassicBasicLoanCard.vue
BoulderBrains: This is rendering in storybook. 
**When router-link renders as a link (while loan is fundraising)**
Screen Shot 2021-09-24 at 1 54 31 PM

**When router-link renders as a span (while allSharesReserved)**
Screen Shot 2021-09-24 at 1 55 18 PM

mcstover: Is the `span` still clickable? Could you also empty the 'to' field in this case?

BoulderBrains: I'll be sure to verify this functionality in dev across the major browsers.

BoulderBrains: <router-link is used in the same way within [CheckoutItemImg.vue on lines 2 & 3](https://github.com/kiva/ui/blob/master/src/components/Checkout/CheckoutItemImg.vue#L3). That change was introduced about 8 months ago.

BoulderBrains: @mcstover I had that :to field return an empty string at an earlier point in my code review, but it was raised that it was a little redundant to have the allSharesReserved check on both fields, so I removed that. The Span is not clickable when returned and the mouse over event is that of a span. Is it prefered to clear out to :to field in the allSharesReserved state?

mcstover: SOunds good, thanks for checking! It's fine as is.

eddieferrer: > What I have in place does work, but I don't see a difference between what I have in there and what you suggested above. Am I misunderstanding you?
> 
> `<router-link :is="allSharesReserved ? 'span' : 'router-link'"`

@BoulderBrains I meant to literally make it: 
`<component
:is="allSharesReserved ? 'span' : 'router-link'"
:to="`/lend/${loanId}`"
v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId]"
>`

But looks like you tested it and it's working, so it's fine as is. :thumbs

PR #3269, opened 2021-09-24
core-110 added new optional kv-text-link to LoanUse.vue that directs …

BoulderBrains:
…to borrower profile if a loanId is passed into <loan-use>

I added a new optional kv-text-link to LoanUse.vue that directs to the borrower profile if a loanId is passed into the < loan-use > component. Also added the condition that if allSharesReserved is true, pass through a blank string for the loanId, which hides the link to the loan as stated in core-111

Looks like this: 
Screen Shot 2021-09-24 at 3 10 57 PM

### DIFF
--- a/src/components/BorrowerProfile/LoanUse.vue
+++ b/src/components/BorrowerProfile/LoanUse.vue
@@ -3,7 +3,6 @@
 		{{ loanUseFiltered }}
 		<kv-text-link
 			v-if="loanId"
-			class="tw-inline"
 			:to="`/lend/${loanId}`"
 			v-kv-track-event="['Lending', 'click-Read more', 'Learn more', loanId]"
 		>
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -1,10 +1,19 @@
...
+		>
+			Learn more
+		</kv-text-link>
 	</p>
 </template>
 
 <script>
+import KvTextLink from '~/@kiva/kv-components/vue/KvTextLink';
+
 const loanUseFilter = require('../../plugins/loan-use-filter');
```
ryan-ludwig, line 17: Unrelated to this PR, but can this `require` statement be removed in favor of a standard `import`?

ryan-ludwig (approved):

BoulderBrains:
@mcstover I've had dialogue with both Nathan and Ryan this morning and the "Learn more" link after the loanUse in the < h4 > style is what we want to go with here. ✔️

eddieferrer (approved):

emuvente (commented):
👍

PR #3273, opened 2021-09-27
feat: contentful button wrapper component

eddieferrer:
* Refactor button handling in DynamicHeroClassic
* Add new component
* Enable buttons to be rendered from contentful, inside of both RTF's and as items in content with support for global event handlers (like openMonthlyGoodSelector)

Sample page that uses buttons inside RTF's - /design
Sample using buttons as content items /lp/homepage-classic

VUE-713

### DIFF
--- a/src/components/Contentful/DynamicHeroClassic.vue
+++ b/src/components/Contentful/DynamicHeroClassic.vue
@@ -34,26 +34,15 @@
 							</kv-carousel>
 						</template>
 						<template v-if="isHeroImage">
-							<template v-if="buttonTo">
-								<router-link
-									:to="buttonTo"
-									v-kv-track-event="[
-										'Hero',
-										'click-hero-loancards',
-										heroMedia[0].description,
-									]"
-								>
-									<kv-contentful-img
-										v-if="heroMedia[0].url"
-										class="tw-block tw-mx-auto tw-mt-0 tw-mb-4 md:tw-mb-0"
-										:contentful-src="heroMedia[0].url"
-										:width="500"
-										fallback-format="jpg"
-										:alt="heroMedia[0].description"
-									/>
-								</router-link>
-							</template>
-							<template v-else>
+							<component
+								:is="buttonTo ? 'router-link' : 'span'"
+								:to="buttonTo"
+								v-kv-track-event="[
+									'Hero',
+									'click-hero-loancards',
+									heroMedia[0].description,
+								]"
+							>
 								<kv-contentful-img
 									v-if="heroMedia[0].url"
 									class="tw-block tw-mx-auto tw-mt-0 tw-mb-4 md:tw-mb-0"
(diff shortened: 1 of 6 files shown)


src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -34,26 +34,15 @@
...
-										class="tw-block tw-mx-auto tw-mt-0 tw-mb-4 md:tw-mb-0"
-										:contentful-src="heroMedia[0].url"
-										:width="500"
-										fallback-format="jpg"
-										:alt="heroMedia[0].description"
-									/>
-								</router-link>
-							</template>
-							<template v-else>
+							<component
```
eddieferrer, line 37: This syntax avoids the duplication of `kv-contentful-img` and conditionally wraps it

src/util/contentful/richTextRenderer.js
```diff
@@ -58,31 +58,11 @@ export function richTextRenderer(content) {
...
-			};
-			return `
-				<kv-button
-					variant="${contentfulEntryNode?.data?.target?.fields?.style ?? 'primary'}"
-					${analyticsDirective()}
-					${clickFunctionality()}
-				>${contentfulEntryNode?.data?.target?.fields?.label ?? ''}</kv-button>
-			`;
+			const buttonObject = JSON.stringify(contentfulEntryNode?.data?.target?.fields).replace(/"/g, '\'');
+			return `<button-wrapper class="tw-whitespace-normal" :content="${buttonObject}"/>`;
```
eddieferrer, line 65: Purge doesn't recognize this class, so I manually added it to the safelist. We want `tw-whitespace-pre-wrap` to support soft breaks from contentful but this breaks the buttons, so we override on the buttons with `tw-whitespace-normal`

tailwind.purge.safelist.txt
```diff
@@ -60,3 +60,4 @@ tw-text-subhead
 tw-text-base
 tw-text-small
 tw-text-link
+tw-whitespace-normal
```
mcstover, line 63: Do we also need `tw-whitespace-pre-wrap` here?

mcstover (approved):
Nice addition and cleanup enabled by it!

tailwind.purge.safelist.txt
eddieferrer: I think it does recognize it after all we don't need this

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -130,16 +107,9 @@
...
-							v-if="buttonText"
-							class="tw-w-full md:tw-w-auto"
-							:to="buttonTo"
-							:variant="buttonStyle"
-							@click.native="buttonClick"
-							v-kv-track-event="buttonAnalytics"
-						>
-							{{ buttonText }}
-						</kv-button>
+						<button-wrapper
```
eddieferrer, line 110: Sample use of new component

mcstover: So clean! But what about the responsive classes on the button. Is it behaving that way by default now? Full width on mobile and inline-block on medium and up?

BoulderBrains (approved):

tailwind.purge.safelist.txt
```diff
@@ -60,3 +60,7 @@ tw-text-subhead
 tw-text-base
 tw-text-small
 tw-text-link
+
+# Whitespace Classes
```
eddieferrer, line 64: In the end I was getting some flaky behavior so I decided to add these to the purge list

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -34,26 +34,15 @@
...
-										fallback-format="jpg"
-										:alt="heroMedia[0].description"
-									/>
-								</router-link>
-							</template>
-							<template v-else>
+							<component
+								:is="buttonTo ? 'router-link' : 'span'"
+								:to="buttonTo"
+								v-kv-track-event="[
```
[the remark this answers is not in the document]
eddieferrer: Good catch. I made a small change that allows for this syntax:
```
v-kv-track-event="buttonTo ? [
	'Hero',
	'click-hero-loancards',
	heroMedia[0].description,
] : null"
```

@BoulderBrains Just pinging you on this. I believe your code in: src/components/LoanCards/KivaClassicBasicLoanCard.vue That does: 
 
May still be firing the analytics event when clicking on a span?

ryan-ludwig (approved):

mcstover (approved):

PR #3280, opened 2021-09-28
core-108 setting faceZoom value on images, gets the correct sizes ret…

BoulderBrains:
…urned

[Core-108](https://kiva.atlassian.net/browse/CORE-108) 

**Before changes:** 
Screen Shot 2021-09-28 at 3 38 09 PM

**After changes:**
Screen Shot 2021-09-28 at 3 39 13 PM

I was able to get the correctly sized images returned only when I had an optional 'faceZoom' value set on the images that were being requested. @emuvente Do you have any insight as to why we wouldn't be getting the correctly sized images with the code that was previously in place? Additionally, now with the faceZoom value, a few of the returned images are zooming in on the borrower's torso. In this example, they are both wearing a lot of patterned fabric, perhaps those are being mis-identified as faces in these cases? Is there a way to correct that? 

Example: 
Screen Shot 2021-09-28 at 3 39 55 PM

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -32,11 +32,11 @@
 					:default-image="{ width: 336 }"
 					:hash="imageHash"
 					:images="[
-						{ width: 336, faceZoom: 0.7, viewSize: 1024 },
-						{ width: 336, faceZoom: 0.7, viewSize: 768 },
-						{ width: 416, faceZoom: 0.7, viewSize: 480 },
-						{ width: 374, faceZoom: 0.7, viewSize: 414 },
-						{ width: 335, faceZoom: 0.7, viewSize: 375 },
+						{ width: 336, faceZoom: 1, viewSize: 1024 },
+						{ width: 336, faceZoom: 1, viewSize: 768 },
+						{ width: 416, faceZoom: 1, viewSize: 480 },
+						{ width: 374, faceZoom: 1, viewSize: 414 },
+						{ width: 335, faceZoom: 1, viewSize: 375 },
 						{ width: 280 },
 					]"
 				/>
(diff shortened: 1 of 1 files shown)


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -25,17 +25,18 @@
...
 					:aspect-ratio="3 / 4"
 					:default-image="{ width: 336 }"
 					:hash="imageHash"
 					:images="[
-						{ width: 336, viewSize: 1024 },
-						{ width: 336, viewSize: 768 },
-						{ width: 416, viewSize: 480 },
-						{ width: 374, viewSize: 414 },
-						{ width: 335, viewSize: 375 },
+						{ width: 336, faceZoom: 0.7, viewSize: 1024 },
```
mcstover, line 35: `faceZoom` can just be `1` as it's converted in code to a whole number.

Otherwise this is a really good improvement.

emuvente:
I would recommend against using the faceZoom prop to solve this issue. faceZoom is specifically for zooming in on the primary face in the photo, for usage in avatar images.

To get the right cropping for the loan cards the loan image sizes just need to be added to the cloudinary approved sizes list in the admin settings manager.

emuvente:
So for the images defined in this loan card these strings will need to be added to the `config.cloudinary.allowed_sizes` array in the settings manager:
```
"w_280,h_210","w_335,h_252","w_336,h_252","w_374,h_281","w_416,h_312","w_560,h_420","w_670,h_503","w_672,h_504","w_748,h_561","w_832,h_624"
```
Once those are there and the cache for the images that have already been loaded is purged, the loan card images will be properly cropped.

BoulderBrains:
Thanks @emuvente. I've made the additions to admin dev, I'm just waiting on the cache to clear to verify before closing this PR.

mcstover:
Thanks @emuvente I thought about the config and checked that but when I found it didn't have your new 80x80 avatar sizes, I thought we were totally bypassing it now... 🤦

BoulderBrains:
FYI: changes have also been added to settingManager in production

PR #3284, opened 2021-09-29
removing the emoji, replacing it with a .png and spacing it correctly

BoulderBrains:
I removed the actual emoji and replaced with with a .png and then forced the sizing and positioning using inline styles. Here it is all fixed up.

Screen Shot 2021-09-28 at 6 31 07 PM


mcstover (approved):
It would good to give the icon a little more space still and try out the 1px border.

BoulderBrains:
With a 1px border and recentered slightly

Screen Shot 2021-09-28 at 7 20 10 PM

PR #3288, opened 2021-09-29
Add background color (powered by hex code selected in Contentful) to …

mcstover:
…WwwPage via ContentfulPage.

The Page Layout content model in Contentful now has a color picker coupled with validation! You can select only 3 hex codes at this time to pass that validation: #ffffff, #f5f5f5, #e0e0e0.

We originally were going to have a dropdown of acceptable tailwind classes but due to the need of a wrapping kv-tailwind class we couldn't do that just yet.

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -2,7 +2,7 @@
 	<component :is="pageFrame"
 		:header-theme="headerTheme"
 		:footer-theme="footerTheme"
-		:background-color="pageBackgroundColor"
+		:main-class="pageBackgroundColor"
 	>
 		<template v-if="!pageError">
 			<component
@@ -294,7 +294,7 @@ export default {
 				this.pageError = true;
 			} else {
 				this.title = (pageData?.page?.pageLayout?.pageTitle || pageData?.page?.pageTitle) ?? undefined;
-				this.pageBackgroundColor = pageData?.page?.pageLayout?.pageBackgroundColor ?? undefined;
+				this.pageBackgroundColor = pageData?.page?.pageLayout?.pageBackgroundColor ?? '';
 				this.headerTheme = siteThemes[pageData?.page?.pageLayout?.headerTheme] || {};
 				this.footerTheme = siteThemes[pageData?.page?.pageLayout?.footerTheme] || {};
 				this.pageFrame = getPageFrameFromType(pageData?.page?.pageType);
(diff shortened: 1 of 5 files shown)


BoulderBrains (approved):

emuvente (approved):

ryan-ludwig:
Totally get it about the `kv-tailwind` class issue. 

Since it's not super easy to find and replace in Contentful, what if the Contentful dropdown was still Tailwind classes, `tw-bg-primary, tw-bg-secondary, tw-bg-tertiary` etc. But then in this Vue file, the style tag would have those Tailwind classes built out so this component is covered. Something like
```
<style scoped>
.tw-bg-primary {
  background-color: rgb(var(--bg-primary));
}
.tw-bg-secondary {
  background-color: rgb(var(--bg-secondary));
}
.tw-bg-tertiary {
  background-color: rgb(var(--bg-tertiary));
}
</style>
```

mcstover:
@ryan-ludwig are these vars `--bg-primary` just already available or would we need to declare them here too?

eddieferrer:
I think the user experience in Contentful is a little confusing for this due to several factors:
- If a page layout has no background color, the hex color picker defaults to showing a black square. 
- The user isn't aware that there is a limitation to 3 colors until after they've selected a color that isn't one of the 3. 

Why not just change the field to a select field in contentful allowing you to select one of the 3 acceptable hex codes? Also add help text to describe the hex codes

src/pages/ContentfulPage.vue
```diff
@@ -292,6 +294,7 @@ export default {
 				this.pageError = true;
 			} else {
 				this.title = (pageData?.page?.pageLayout?.pageTitle || pageData?.page?.pageTitle) ?? undefined;
+				this.pageBackgroundColor = pageData?.page?.pageLayout?.pageBackgroundColor ?? undefined;
```
eddieferrer, line 297: The default value here - `undefined`, should probably match the component default in `data:` above - ''. I think either undefined or '' would work

test/unit/fixtures/CorporateCampaignContentfulPageRaw.json
```diff
@@ -62,6 +62,7 @@
 			},
 			"fields": {
 				"name": "Promo Campaign Test",
+				"pageBackgroundColor": "",
```
eddieferrer, line 65: This won't come back at all from contentful if the field is undefined in contentful.

ryan-ludwig:
@mcstover those CSS vars (like `--bg-primary` are available on all pages

mcstover:
Ok, I've reworked this to accept a Tailwind class string, then used variables to set the values. This allows use to use the existing `mainClass` prop to apply the bg. Should be a little more future proof and matches Storybook better.

Also cleaned up the test data.

ryan-ludwig (approved):

PR #3291, opened 2021-09-30
fix: Use html comment to disable max-length lint check for data uri

emuvente:
I had to try several different ways of setting the eslint comment, but this finally worked for me locally


src/components/Contentful/StoryCard.vue
```diff
@@ -126,6 +126,7 @@ export default {
 };
 </script>
 
+<!-- eslint-disable max-len -->
```
mcstover, line 129: Well this is interesting! 🤣

mcstover (approved):

PR #3295, opened 2021-10-01
 Core-48 Centered partner logos fed from contentful.

BoulderBrains:
[Core-48](https://kiva.atlassian.net/browse/CORE-48)

Partner logos section

**Mobile:** 
Screen Shot 2021-09-30 at 7 31 28 PM

**Tablet:** 
Screen Shot 2021-09-30 at 7 31 09 PM

**Desktop:**
Screen Shot 2021-09-30 at 7 31 47 PM

### DIFF
--- a/src/components/Contentful/MediaItemsCentered.vue
+++ b/src/components/Contentful/MediaItemsCentered.vue
@@ -1,0 +1,76 @@
+<template>
+	<section-with-background-classic
+		:background-content="sectionBackground"
+		:vertical-padding="verticalPadding"
+	>
+		<template #content>
+			<kv-page-container>
+				<kv-grid
+					:style="customGridStyles"
+					class="tw-grid-cols-12 lg:tw-grid-cols-10 lg:tw-gap-y-4
+						tw-justify-items-center tw-text-center tw-rounded tw-bg-white
+						tw-py-5 lg:tw-px-[177px]"
+				>
+					<span
+						v-for="(item, index) in contentfulMediaItems"
+						:key="index"
+						class="tw-flex tw-justify-items-center tw-items-center tw-mb-2"
+						:class="[itemClasses, {
+							'md:tw-col-span-12': index === 9
+						}]"
+					>
+						<kv-contentful-img
+							:contentful-src="item.file.url"
+							:alt="item.file.description"
+							:width="item.file.details.image.width"
+							:height="item.file.details.image.height"
+							:style="maxWidthStyles"
+						/>
+					</span>
+				</kv-grid>
+			</kv-page-container>
+		</template>
+	</section-with-background-classic>
+</template>
+
+<script>
+import SectionWithBackgroundClassic from '@/components/Contentful/SectionWithBackgroundClassic';
+import contentfulStylesMixin from '@/plugins/contentful-ui-setting-styles-mixin';
+import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
+import KvGrid from '~/@kiva/kv-components/vue/KvGrid';
+import KvContentfulImg from '~/@kiva/kv-components/vue/KvContentfulImg';
+
+export default {
+	components: {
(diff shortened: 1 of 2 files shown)


mcstover (approved):

emuvente (approved):

src/components/Contentful/MediaItemsCentered.vue
```diff
@@ -0,0 +1,76 @@
...
+						:key="index"
+						class="tw-flex tw-justify-items-center tw-items-center tw-mb-2"
+						:class="[itemClasses, {
+							'md:tw-col-span-12': index === 9
+						}]"
+					>
+						<kv-contentful-img
+							:contentful-src="item.file.url"
+							:alt="item.file.description"
+							:width="item.file.details.image.width"
```
ryan-ludwig, line 25: Not super urgent, but I'm concerned with this width and height coming from the Contentful item details, as those are the width and height of the originally uploaded image and it's encouraged to upload a very large/high fidelity item to Contentful (and then we properly size them in the code here).

Can we determine what the max width is of this container and use that for the width? Since I assume the aspect ratios are different that would mean removing the height.

src/components/Contentful/MediaItemsCentered.vue
```diff
@@ -0,0 +1,76 @@
...
+			});
+		},
+		contentfulMediaItems() {
+			const mediaItemsContentfulData = this.content?.media ?? [];
+			console.log('length', mediaItemsContentfulData.length);
+			return mediaItemsContentfulData;
+		},
+		itemClasses() {
+			return ['tw-col-span-6', {
+				'md:tw-col-span-4': this.contentfulMediaItems.length >= 6,
```
ryan-ludwig, line 70: If you like you could combine these two lines into one, e.g.,
`'md:tw-col-span-4 lg:tw-col-span-2': this.contentfulMediaItems.length >= 6,`

src/components/Contentful/MediaItemsCentered.vue
```diff
@@ -0,0 +1,76 @@
...
+		:background-content="sectionBackground"
+		:vertical-padding="verticalPadding"
+	>
+		<template #content>
+			<kv-page-container>
+				<kv-grid
+					:style="customGridStyles"
+					class="tw-grid-cols-12 lg:tw-grid-cols-10 lg:tw-gap-y-4
+						tw-justify-items-center tw-text-center tw-rounded tw-bg-white
+						tw-py-5 lg:tw-px-[177px]"
```
ryan-ludwig, line 12: These bracketed tailwind classes make me squirm, since they add that class to the global stylesheet loaded on all pages even though it's only used once, but I acknowledge there's not always a better way.

Let's take a second look at this next week and see if there's not a way we can solve this with grid or flexbox.

src/components/Contentful/MediaItemsCentered.vue
```diff
@@ -0,0 +1,76 @@
...
+				>
+					<span
+						v-for="(item, index) in contentfulMediaItems"
+						:key="index"
+						class="tw-flex tw-justify-items-center tw-items-center tw-mb-2"
+						:class="[itemClasses, {
+							'md:tw-col-span-12': index === 9
```
eddieferrer, line 19: Is this necessary?

mcstover: This is a temporary measure that relates to the number of sponsors we have for the current campaign. Def. part of the layout revisit that Ryan mentioned. We ultimately want this to support an arbitrary number of media items and handle the layout without specific intervention.

src/components/Contentful/MediaItemsCentered.vue
mcstover: These images are actually all smaller than the max of the container (most are 93px wide) so I think we're ok for now. Def and important improvement to make for the next round.

PR #3298, opened 2021-10-04
fix: sets index of loanCarousel to 0 when new category is clicked, al…

BoulderBrains:
…so re-initializes to ensure proper number of indexes is displayed below category arrow

- [Core-136](https://kiva.atlassian.net/browse/CORE-136)
     - When user chooses new category, loan cards are staying on previously scrolled to index.
     - FIXED by setting the index to 0 by calling goToIndex() in KvCarousel
- [Core-137](https://kiva.atlassian.net/browse/CORE-137)
     - On production, if one category has 4 carousel clicks of loans then 4 will show for all categories even if that category only has 2 "carousel clicks. 
     - Results in this error: 
     - Screen Shot 2021-10-04 at 2 13 32 PM
     - FIXED by calling the reInit() function

### DIFF
--- a/src/components/LoanCollections/KivaClassicLoanCarousel.vue
+++ b/src/components/LoanCollections/KivaClassicLoanCarousel.vue
@@ -15,6 +15,7 @@
 			:embla-options="{
 				loop: false,
 			}"
+			ref="categoryCarousel"
 			:multiple-slides-visible="true"
 			slides-to-scroll="visible"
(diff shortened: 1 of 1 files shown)


mcstover (approved):

ryan-ludwig (approved):

emuvente (approved):
:+1:

PR #3303, opened 2021-10-05
Typography: VUE-737 remove kv-tailwind class prefixes and foundation reset/typography

ryan-ludwig:
This is going into `typography_dev` branch


emuvente (commented):
👍

PR #3323, opened 2021-10-12
feat: Hide page elements based on query param

eddieferrer:
* Hides global promo, nav menu and secondary menu if query param kivaAppReferral=true exists

GD-139


eddieferrer:
Used a local resolved. My other idea was to use a cookie. Not sure if there is a benefit to using either approach. I tested this and LGTM, it also prevents any flash of the header and works even when not logged in after auth0 redirect

eddieferrer:
> I think the local resolver option is good as it "should" persist for page transitions within the Ui app. I'm not really interested in setting something as persistent as a cookie for this task.

@mcstover This will not persist for page transitions

mcstover:
👍

PR #3341, opened 2021-10-19
feat: Wrap KvTooltip in kiva classic theme provider and use the mint …

mcstover:
…theme for existing usages.


ryan-ludwig:
Love seeing KvThemeProvider come in handy!

PR #3353, opened 2021-10-21
feat: ResponsiveImageSet processing enhancement to handle all u…

eddieferrer:
…se cases

This PR solves the rest of the issues for `ResponsiveImageSet` rendered from contentful. This gets rid of the old paradigm of "retina" and "std". 

The `responsiveImageSetSourceSets` function can be used to pass a responsive Image Set from contentful into the `KvContentfulImg` component. Currently used in `DynamicHeroClassic` and enabled in `richTextRenderer`, but no RTF's use this yet

**Contentful Changes**
This adds a JSON object to a responsive image set, this json object looks like: 
```
{
    "sm": {
        "width": 537
    },
    "lg": {
        "width": 394
    },
    "med": {
        "width": 397
    }
}
```
**What is expected in Contentful for a responsive image set?** 
A single set of images at the highest resolution possible, and an optional JSON object above. 

For example, for a page implementing a responsive image set in the hero, where there is 1 image for small and medium resolutions and 1 image for large and above, the contentful entry would look like: 

Responsive Image Set
 - name
 - description
 - Images:
         *image1: The image to display at sm or med resolutions. This image should be the highest quality possible and can have huge dimensions. It can include "sm" in the filename
         *image2: The image to display at large and above. This image should be the highest quality possible and can have huge dimensions. It MUST include "lg" in the filename
- JSON Object: This json object includes the width that the image should be displayed at at the corresponding resolutions. If missing, it will default to the max width at that breakpoint, which works ok if the image is inserted in a col-12 container that takes up the full width. 

This will greatly clean this content up, since every image in the ResponsiveImageSet should be a unique image, not just the same image in a different resolution

TODO: 
- After this we can remove all "std" images from contentful ResponsiveImageSets. 
- We can also remove use cases where the same image has been added in contentful at different resolutions, one for large at 1200x900 and one for xl at 1600x1000 for example. In this case they should only be an "lg" image at the highest resolution


mcstover:
This is looking good. I like that the json object is optional. We should probably pose a question on defaults for discussion with Design.

PR #3381, opened 2021-10-29
Core 239

BoulderBrains:
[Core-239](https://kiva.atlassian.net/browse/CORE-239)

Updated the time remaining display on borrower profile. Added an additional check to see if the differenceInDays is larger than 28, if it is we pass through the unit we want returned, which is days. 

I've verified this between the old and new borrower profiles for a few loans and the display of time remaining is the same. 
A loan I used on my VM: https://dev-vm-01.kiva.org/lend-beta/2260639

**Old Borrower profile:** 
Screen Shot 2021-10-29 at 11 55 02 AM

**New Borrower profile:**
Screen Shot 2021-10-29 at 11 54 14 AM

**Borrower profile with less than 28 days remains unchanged**
Screen Shot 2021-10-29 at 11 55 41 AM

I've also added a test to cover a 40 days case.

### DIFF
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -107,14 +107,11 @@ function fundraisingTimeLeft(loan) {
 	}
 
 	const now = parseISO(new Date().toISOString());
-
 	const diffInDays = differenceInDays(plannedExpirationDate, now);
-	console.log('plannedEXP', plannedExpirationDate);
-	console.log('NOW', parseISO(new Date().toISOString()));
-	console.log('diffInDays', diffInDays);
 
-	// use differenceInDays to see if the plannedExpirationDate is greater than 28 day from now
-	if (diffInDays > -28) {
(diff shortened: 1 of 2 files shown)


mcstover (approved):

src/api/localResolvers/loan.js
```diff
@@ -105,7 +106,15 @@ function fundraisingTimeLeft(loan) {
 		return '';
 	}
 
+	const now = parseISO(new Date().toISOString());
+	const diffInDays = differenceInDays(plannedExpirationDate, now);
```
emuvente, line 110: `new Date()` can be passed directly into the function:
```
differenceInDays(plannedExpirationDate, new Date())
```

emuvente (approved):
Looks good! Minor style comment, but this works as-is so feel free to merge.

mcstover (approved):

PR #3390, opened 2021-11-04
feat: require valid names during login/registration AUTH-305

emuvente:


ryan-ludwig (approved):

ryan-ludwig:
Nice! Loved reading through those tests

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -2,42 +2,60 @@
...
+					v-model.trim="lastName"
+					:validation="$v.lastName"
+				>
+					Last name
+					<template #required>
+						Enter last name.
+					</template>
+				</kv-base-input>
+				<kv-base-input name="newAcctTerms"
+					type="checkbox"
```
ryan-ludwig, line 41: I think it's cool that `type` can be set to checkbox like this. I think a future improvement for `kv-base-input` would be to take type as a prop and then display either a `kv-text-input` or `kv-checkbox`, maybe a `kv-radio` too since they have the Kiva Classic styling (and the default form styling looks very non-Kiva Classic in a Tailwind world).

mcstover (approved):

PR #3392, opened 2021-11-04
Causes Product

eddieferrer:
This PR will be a combination of all the work necessary for a user to create a cause subscription: 

GD-162
GD-161

Test url example: 

/causes/signup?cause=climate&beta=true

There is no change to the waitlist functionality, so going to: /causes/signup is still the waitlist signup as it is currently

TODO:
- Update eligibility for causes form


src/components/Causes/CausesDropInPaymentWrapper.vue
```diff
@@ -0,0 +1,175 @@
+<template>
```
eddieferrer, line 1: The plan is to reuse this wrapper for both Update and Register -- similar to how the monthly good drop in wrapper works

src/pages/Causes/Signup.vue
```diff
@@ -1,154 +1,111 @@
...
+			this.hasAutoDeposits = pageQueryResult?.my?.autoDeposit ?? false;
+			const legacySubs = pageQueryResult?.my?.subscriptions?.values ?? [];
+			this.hasLegacySubscription = legacySubs.length > 0;
+			this.hasMadeLoan = pageQueryResult?.my?.lender?.loanCount > 0;
+		} catch (e) {
+			logReadQueryError(e, 'Causes Signup causesSignupEligibilityQuery');
+		}
+	},
+	computed: {
+		// Temporary query param feature flag
```
eddieferrer, line 96: This query param flag is just so we can test this and validate in lower environments

src/pages/Causes/Signup.vue
```diff
@@ -1,154 +1,111 @@
...
+		isBeta() {
+			return this.$route.query.beta === 'true';
+		},
+		isEligibleForCausesSignup() {
```
eddieferrer, line 100: This needs to be updated to reflect the use case of previous cause subscribers coming back to sign up again

src/pages/Causes/Signup.vue
```diff
@@ -1,154 +1,111 @@
 <template>
```
eddieferrer, line 1: This new wrapper component will direct to waitlist or signup form.

mcstover (approved):
Getting close!

src/pages/Causes/SignupForm.vue
```diff
@@ -0,0 +1,195 @@
...
+			// disable form inputs while routing to thanks
+			this.success = true;
+			this.$kvTrackEvent('Registration', 'successful-causes-reg', 'register-causes');
+			this.$router.push({
+				path: 'causes/thanks',
+			});
+		},
+	},
+	computed: {
+		causeCategoryId() {
```
eddieferrer, line 177: This feels super hacky. But we need a way to get the category id from the api for a specific category. Obviously if the name changes this would break. And it would be nice to use the name as the display name on the ui as well. 

I think a better approach is to maybe add a key to the category that will never change and we can use that to map key to category id. @mcstover thoughts on this?

mcstover: I'd agree that it would be better to have an ID passed around to reference a specific "cause" from your list. That may be a future piece of work though as I think you'd need to be passing the id from the initial Cause selection through login and to this page.

mcstover (approved):

PR #3397, opened 2021-11-05
refactor: remove deprecated contentfulUtils

eddieferrer:
* removed processContent and determineResponsive image size functions

VUE-814


mcstover (approved):
I think this is fine... I believe the default page is still reachable by an experiment force url but no one expressed intent to continue using it...

...and cheers for the code cleanup!

eddieferrer:
Yeah I think we're good here. LendByCategoryHomepage.vue is not being used. DefaultHomepage will still load with the covid hero.

PR #3405, opened 2021-11-09
feat: Core-236 disclaimer text parsed and rendered from contentful wi…

BoulderBrains:
…thin the footer

[Core-236](https://kiva.atlassian.net/browse/CORE-236)

Disclaimer text was added to the Global Promo Banner content type in contentful. If the banner is active or within the set date range & disclaimer text is present within a global promo banner, we now display it directly above the footer. 

If the enabled promo banner has an associated disclaimer, we display an indication in the promo banner using a superscript # 1. When that superscript is clicked we drop the user to the disclaimers located above the footer on the page.

**Global site maintenance banner:**
Screen Shot 2021-11-09 at 12 36 20 PM

**Appeal banner open**
Screen Shot 2021-11-10 at 9 56 40 AM

**Appeal banner closed**
Screen Shot 2021-11-10 at 9 56 29 AM

**Disclaimers now above footer**
Screen Shot 2021-11-10 at 9 58 27 AM


src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,135 @@
+<template>
+	<div class="row">
+		<ol class="text-left">
+			<li
+				id="disclaimers"
```
mcstover, line 5: id's should be unique so I would move this up to the `ol`. Then, if you want to generate an id for each item you can append the index to some string like so `:id="`disclaimer_${index}`"`

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,135 @@
...
+				}
+			}
+		}
+	},
+	computed: {
+		// Constructing the final form of the disclaimer text for display
+		fullyBuiltDisclaimerText() {
+			const builtDisclaimertext = [];
+			this.disclaimerContent.forEach(disclaimer => {
+				builtDisclaimertext.push(`Disclaimer: ${disclaimer}`);
```
mcstover, line 129: What's the markup look like with this? Guessing it's `Disclaimer: <p>You're disclaimer...`

I think we want a result that's more like `<p>Disclaimer: You're disclaimer text...`

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,134 @@
...
+
+			// exit if missing setting or fields
+			if (!uiGlobalPromoSetting || !uiGlobalPromoSetting.fields) {
+				return false;
+			}
+			// uiGlobalPromoSetting can contain an array of different banners with
+			// different start/end dates first determine if setting is enabled.
+
+			// DO I NEED THIS CHECK, SINCE IF THERE'S INACTIVE BANNERS WITHIN DATE RANGE
+			// WE STILL WANT TO DISPLAY THE DISCLAIMERS... RIGHT?
```
ryan-ludwig, line 50: :) Do you know the answer to this?

src/components/WwwFrame/PromotionalBanner/Banners/AppealBanner/AppealBannerCircular.vue
```diff
@@ -225,6 +249,11 @@ export default {
 
 	&__title {
 		margin-bottom: rem-calc(4);
+		display: inline;
```
ryan-ludwig, line 252: Wondering if this should be inline-block? I don't think the margin-bottom will do anything if it's inline

BoulderBrains: Thanks for calling this out Ryan. I set it to inline, which allowed the superscript #1 to jump up on the same line as the headline text. But, yeah this does make the headline loose it's small bottom margin. Do you think there would be a problem with just adding that small margin-bottom to margin-top on the element below it?

ryan-ludwig: Probably fine adding margin top on the element below, but I think inline-block will let you have both the #1 on the same line and also the margin-bottom.

BoulderBrains: I've tried making that update, but doesn't seem to be working. If I update the heading to `display: inline-block`, the 1 jumps to the next line. If I then alter the display settings on the <a>1</a>, to any within this available display options, I still get them displayed on two lines.

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,131 @@
...
+				if (activePromoBanner) {
+					// check for visibility based on current route and hiddenUrls field
+					const hiddenUrls = _get(activePromoBanner, 'fields.hiddenUrls', []);
+					if (hiddenUrls.includes(this.$route.path)) {
+						return false;
+					}
+
+					// check for visibility on promo session override
+					const showForPromo = _get(activePromoBanner, 'fields.showForPromo', false);
+					if (this.hasPromoSession && !showForPromo) {
```
mcstover, line 92: `this.hasPromoSession` isn't present in this component so we may want to add that check or pass down that state.

mcstover: As of this time, this is just never going to trigger, so it's not a blocker, but something we should get in place at some point soon.

mcstover (approved):

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,131 @@
...
+					}
+				}
+
+				// go through the inactive promoBanners, if within date range and disclaimer text exists
+				// push that disclaimer text to disclaimerContent
+				if (inactivePromoBanners.length > 0) {
+					inactivePromoBanners.forEach(item => {
+						const itemDisclaimer = item?.fields?.disclaimers ?? null;
+						if (itemDisclaimer) {
+							this.disclaimerContent.push(documentToHtmlString(itemDisclaimer));
```
emuvente, line 112: Does it matter if the list number that is used for a given disclaimer changes over time? With this forEach the numbers could change as different promos end. If they need to be consistent for the entire duration of the promo, then the list number might need to be defined in Contentful. I could see that getting really complicated though, so I hope that the number doesn't matter for the extra promos.

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -0,0 +1,169 @@
+s<template>
```
emuvente, line 1: An extra 's' slipped through here. Surprisingly, the Vue loader didn't complain about it 🤔

BoulderBrains: 👀 Damn, good catch, ty

src/components/WwwFrame/DisclaimersContentful.vue
mcstover: I think we're ok without numeric assignment at this time as only an "active" promo could be visible at the top with a jump link requiring any correlation. The "active" promo is extracted first and filtered out of the "inactive" set so it will also always be 1st in the list to match that `1` link at the top.

Let's keep an eye on this as it gets used though as we may have to get more sophisticated with tracking disclaimers along any presented content...

mcstover (approved):

src/util/settingsUtils.js
```diff
@@ -65,6 +65,23 @@ export function settingEnabled(data, enabledKey, startTimeKey, endTimeKey) {
 	return enabled && isWithinInterval(new Date(), { start: startTime, end: endTime });
```
eddieferrer, line 65: This could be refactored to use the new function below and avoid repeating code. So this return would read something like `return enabled && settingWithinDateRange(foo, bar, blah)`

eddieferrer (approved):

emuvente (approved):
Looks great!

mcstover (approved):

PR #3407, opened 2021-11-09
feat: adds optional button content to hero with carousel

eddieferrer:
GD-167

### DIFF
--- a/src/components/Contentful/HeroWithCarousel.vue
+++ b/src/components/Contentful/HeroWithCarousel.vue
@@ -31,6 +31,12 @@
 							</template>
 						</kv-carousel>
 					</div>
+					<div class="tw-col-span-12 tw-text-center">
+						<button-wrapper
+							class="tw-w-full md:tw-w-auto tw-mt-2"
+							:content="buttonContent"
+						/>
+					</div>
 				</kv-grid>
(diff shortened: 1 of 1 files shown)


eddieferrer:
Screen Shot 2021-11-09 at 1 54 47 PM
Screen Shot 2021-11-09 at 1 50 43 PM

src/components/Contentful/HeroWithCarousel.vue
```diff
@@ -31,6 +31,12 @@
 							</template>
 						</kv-carousel>
 					</div>
+					<div class="tw-col-span-12 tw-text-center">
```
mcstover, line 34: Should this be conditional so it doesn't introduce the extra margin or any other space from the empty containers?

mcstover: Or maybe the ButtonWrapper just won't render it's container due it's internal `v-if`...that would be ok too!

eddieferrer: I just tested this it actually renders as 0 height, so it won't add any space, and the button wrapper won't render if there is no button content, so no v-if required

mcstover:
👍

PR #3415, opened 2021-11-11
Core 198 core 223 promo credit display and disclaimer

mcstover:


src/components/Checkout/OrderTotals.vue
```diff
@@ -53,6 +53,12 @@
 				>
 					Apply
 				</button>
+				<div v-if="hasUPCCode" class="upc-disclaimer">
```
torgie-kiva, line 56: I'm guessing `hasUPCCode` is set somewhere already?

It's a very minor nitpick, but the C in UPC is already "Code", so this is kind of like `hasATMMachine` :P Not going to block the PR by any means, though!

torgie-kiva (approved):

BoulderBrains (approved):

src/components/Checkout/OrderTotals.vue
mcstover: yea kinda silly to repeat...

PR #3419, opened 2021-11-11
build: update token for release

emuvente:
Need to use a token that is allowed to push to protected branches to avoid errors during the release action. This technically allows the branch protection rules to be bypassed by anyone with write access to the repo. However, we don't hand out write access to many people, and actually bypassing branch protections takes determined effort (a new workflow is required to be made, configured with this new token, and it needs to run git push commands). Pull requests from outside contributors won't be able to run actions, so there is no concern there.

### DIFF
--- a/.github/workflows/release-workflow.yml
+++ b/.github/workflows/release-workflow.yml
@@ -19,5 +19,5 @@ jobs:
           node-version: '16.x'
       - name: Release
         env:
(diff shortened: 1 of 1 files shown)


mcstover (approved):

emuvente:
This access token has now been removed, so we don't have to be concerned about it being used to bypass branch protection anymore.

PR #3425, opened 2021-11-15
feat: update /settings/security/mfa to kiva classic

ryan-ludwig:


mcstover (approved):

emuvente (approved):
That all looks good me! I'll test this out when it's deployed as well.

PR #3432, opened 2021-11-18
Core 290 ml landing team application fixes

mcstover:


src/components/Checkout/TeamAttribution.vue
```diff
@@ -55,6 +55,15 @@ export default {
 			return _orderBy(this.teams, 'name');
 		}
 	},
+	watch: {
+		// watch for updated changes to the teamId prop
+		// typically this is derived from the team assigned to the loan reservation
+		teamId(newId, prevId) {
+			if (newId !== prevId) {
+				this.selectedId = newId;
```
torgie-kiva, line 63: Will this also capture cases where new loans are added to the basket (after, say, clicking "Join team"), or will they still need to manually select a team for each new loan added?

mcstover: From my testing in the VM all subsequent loans added to basket were given the campaign team if the lender was now a member of that team.

BoulderBrains (approved):

torgie-kiva (approved):

PR #3435, opened 2021-11-19
fix: sync browser clock before checkout page loads VUE-847

emuvente:


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -364,6 +364,14 @@ export default {
 			this.activeLoginDuration = parseInt(_get(data, 'general.activeLoginDuration.value'), 10) || 3600;
 		}
 	},
+	beforeRouteEnter(to, from, next) {
```
emuvente, line 367: This navigation guard is called before the router.onReady callback in client-entry.js, so this sync will happen before the vue app is mounted on the client after a server render.

mcstover (approved):

PR #3436, opened 2021-11-20
fix: update layout of global promo banner to ensure an inline display…

mcstover:
… of the disclaimer link

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/GenericPromoBanner.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/GenericPromoBanner.vue
@@ -1,27 +1,28 @@
 <template>
 	<div class="row align-center generic-banner">
-		<component
-			:is="currentWrapperComponent"
-			:to="trimmedLink"
-			:href="trimmedLink"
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/GenericPromoBanner.vue
```diff
@@ -1,27 +1,28 @@
...
+				:is="currentWrapperComponent"
+				:to="trimmedLink"
+				:href="trimmedLink"
+				:target="isExternalLink ? '_blank' : '_self'"
+				:class="{ 'banner-link' : trimmedLink, 'banner-wrapper' : !trimmedLink}"
+				v-kv-track-event="handleTracking"
+			>
+				{{ processedContent }}
+			</component>
+			<a
```
mcstover, line 15: By stripping out the `p` tags injected by Contentful we can use `processedContent` directly and set these items to be inline instead of one block alongside the other. Previously the `a` tag wrapped around the `p` from contentful created invalid markup which forced a block layout.

emuvente (approved):

BoulderBrains (approved):
Nice

PR #3438, opened 2021-11-22
fix: prevent content from breaking mobile layout

ryan-ludwig:
Following up on an email from Google's search console today, 

```
Content wider than screen
We recommend that you fix these issues when possible to enable the best experience and coverage in Google Search.
```

Screen Shot 2021-11-22 at 10 16 52 AM

This doesn't fix the root problem, something's odd with the content coming in and adding a bunch of `&nbsp;`
image

but it does prevent the layout from breaking in those cases.

https://www.kiva.org/lend-beta/2272570

### DIFF
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -287,8 +287,8 @@ export default {
 				this.loan.loanTermLenderRepaymentTerm = loan?.terms?.lenderRepaymentTerm ?? 0;
 				this.loan.lossLiabilityCurrencyExchange = loan?.terms?.lossLiabilityCurrencyExchange ?? '';
 				this.loan.repaymentInterval = loan?.repaymentInterval ?? '';
-				this.loan.disbursalDate = loan.disbursalDate ?? '';
-				this.loan.status = loan.status ?? '';
+				this.loan.disbursalDate = loan?.disbursalDate ?? '';
+				this.loan.status = loan?.status ?? '';
 
 				this.partner.arrearsRate = partner?.arrearsRate ?? 0;
 				this.partner.avgBorrowerCost = partner?.avgBorrowerCost ?? 0;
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/DetailsTabs.vue
```diff
@@ -287,8 +287,8 @@ export default {
 				this.loan.loanTermLenderRepaymentTerm = loan?.terms?.lenderRepaymentTerm ?? 0;
 				this.loan.lossLiabilityCurrencyExchange = loan?.terms?.lossLiabilityCurrencyExchange ?? '';
 				this.loan.repaymentInterval = loan?.repaymentInterval ?? '';
-				this.loan.disbursalDate = loan.disbursalDate ?? '';
-				this.loan.status = loan.status ?? '';
+				this.loan.disbursalDate = loan?.disbursalDate ?? '';
+				this.loan.status = loan?.status ?? '';
```
ryan-ludwig, line 291: These are unrelated, just fixing console errors when pulling up the page on my vm.

src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<section>
-		<div class="tw-prose">
+		<div class="tw-prose tw-overflow-auto">
```
mcstover, line 3: Seems ok. Also wonder if there is some "text-wrap" property that should be set here..

mcstover (approved):

src/components/BorrowerProfile/MoreAboutLoan.vue
mcstover: `overflow-wrap: break-word;` could also work

ryan-ludwig: Nice!

mcstover (approved):

PR #3439, opened 2021-11-22
feat: if a user is logged in and has an account, get their email for …

eddieferrer:
…the waitlist signup

GD-161

Last part for GD-161, if a logged in user end up on the waitlist page, either now or when we launch the product, instead of displaying the email input, just submit their email to iterable.

### DIFF
--- a/src/pages/Causes/Signup.vue
+++ b/src/pages/Causes/Signup.vue
@@ -94,18 +94,21 @@ export default {
 		}
 	},
 	created() {
-		try {
-			const pageQueryResult = this.apollo.readQuery({
-				query: pageQuery,
-			});
-			this.isMonthlyGoodSubscriber = pageQueryResult?.my?.autoDeposit?.isSubscriber ?? false;
-			this.hasAutoDeposits = pageQueryResult?.my?.autoDeposit ?? false;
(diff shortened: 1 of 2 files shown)


src/pages/Causes/Signup.vue
```diff
@@ -94,18 +94,21 @@ export default {
...
-			this.isMonthlyGoodSubscriber = pageQueryResult?.my?.autoDeposit?.isSubscriber ?? false;
-			this.hasAutoDeposits = pageQueryResult?.my?.autoDeposit ?? false;
-			const legacySubs = pageQueryResult?.my?.subscriptions?.values ?? [];
-			this.hasLegacySubscription = legacySubs.length > 0;
-			this.hasMadeLoan = pageQueryResult?.my?.lender?.loanCount > 0;
-			this.mySubscriptions = pageQueryResult?.mySubscriptions?.values ?? [];
-		} catch (e) {
-			logReadQueryError(e, 'Causes Signup causesSignupEligibilityQuery');
+		// TODO temporary beta query param
+		if (this.$route.query.beta === 'true') {
```
eddieferrer, line 98: Guard this read query if the cause beta is not on to avoid invariant error

mcstover (approved):

PR #3442, opened 2021-11-23
fix: ensure richText content shows for globalPromo banners and hide d…

mcstover:
…isclaimer link when not present


emuvente (approved):
👍

PR #3444, opened 2021-11-23
feat: adds dynamic date to causes signup form

eddieferrer:
* adds dynamic withdrawal date to causes signup, based on todays date
* fix eligibility for causes
* minor styling fix for when we have the users email

GD-166
GD-161

### DIFF
--- a/src/pages/Causes/Signup.vue
+++ b/src/pages/Causes/Signup.vue
@@ -33,7 +33,7 @@ const pageQuery = gql`query causesSignupEligibilityQuery {
 			}
 		}
 	}
-	mySubscriptions {
+	mySubscriptions(includeDisabled: true) {
 		values {
 			id
 			enabled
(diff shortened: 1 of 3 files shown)


src/pages/Causes/Signup.vue
```diff
@@ -133,12 +133,11 @@ export default {
 			const hasInactiveCauseSubscription = causesSubscriptions.find(
 				subscription => !subscription.enabled
 			);
-
 			return !hasActiveCauseSubscription
 				&& !this.isMonthlyGoodSubscriber
 				&& !this.hasAutoDeposits
 				&& !this.hasLegacySubscription
-				&& (!this.hasMadeLoan && !hasInactiveCauseSubscription);
+				&& (!this.hasMadeLoan || hasInactiveCauseSubscription);
```
eddieferrer, line 140: I believe this is correct. If has not made a loan or has inactive subscription they should be eligible

mcstover: makes sense

mcstover (approved):

PR #3452, opened 2021-11-29
fix: remove hard-coded image designations in favor of the ordered obj…

mcstover:
…ects from contentful


mcstover:
Currently broken in dev (b/c we set a different image in the right rail content group) and using mis-matched images in prod.

emuvente (approved):
👍  nice simplification!

BoulderBrains (approved):

PR #3455, opened 2021-11-29
fix: ensure instant donation thanks is excluded

mcstover:


src/util/urlUtils.js
```diff
@@ -6,7 +6,7 @@ export const globalBannerDenyList = [
 	'/confirm-instant-donation/*',
 	'/donate/support-kiva',
 	'/error',
-	'/instant-donation-thanks',
+	'/instant-donation-thanks/*',
```
mcstover, line 9: needed to use the new dynamic url signifier

PR #3458, opened 2021-11-30
Kvaccordion vue 875

roger-in-kiva:
KvAccordionItem was updated to have tw classes and no more custom classes.

Note: Negative rotation for icon is not working with tw classes, so, is working with positive rotation.
_tw--rotation-180_

### DIFF
--- a/src/components/Kv/KvAccordionItem.vue
+++ b/src/components/Kv/KvAccordionItem.vue
@@ -1,30 +1,26 @@
 <template>
 	<div
-		class="kv-accordion"
-		:class="{
-			'kv-accordion--open' : isOpen,
-		}"
+		class="tw-border-b tw-border-tertiary tw-relative"
 	>
 		<button
-			class="kv-accordion__header-button"
+			class="tw-w-full tw-flex tw-justify-between tw-items-center tw-py-[0.75rem] tw-px-0
+				tw-text-left disabled:tw-cursor-not-allowed disabled:tw-opacity-low"
 			@click.prevent="toggle"
 			:disabled="disabled"
 			:aria-controls="`kv-accordion-${id}`"
 			:aria-expanded="isOpen ? 'true' : 'false'"
 		>
-			<span class="kv-accordion__header-text">
+			<span class="tw-flex-1">
 				<slot name="header"></slot>
 			</span>
-			<kv-icon
-				class="kv-accordion__header-icon"
-				name="small-chevron"
-				:from-sprite="true"
-				aria-hidden="true"
+			<kv-material-icon
+				class="tw-h-3 tw-w-[25px] tw-transition tw-transform tw-duration-500 tw-ease"
+				:class="{ 'tw-rotate-180' : isOpen }"
+				:icon="mdiChevronDown"
 			/>
 		</button>
 		<kv-expandable>
 			<div
-				class="kv-accordion__pane"
 				:id="`kv-accordion-${id}`"
 				v-show="isOpen"
 				:aria-hidden="isOpen ? 'false' : 'true'"
@@ -49,12 +45,13 @@
 // pass a prop like 'tag' that sets the parent node of the button. <accordion tag="h3">...
 
 import KvExpandable from '@/components/Kv/KvExpandable';
-import KvIcon from '@/components/Kv/KvIcon';
+import { mdiChevronDown } from '@mdi/js';
+import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
 
 export default {
 	components: {
+		KvMaterialIcon,
 		KvExpandable,
-		KvIcon,
 	},
 	props: {
 		/**
@@ -82,7 +79,8 @@ export default {
 	},
 	data() {
 		return {
-			isOpen: this.open
+			isOpen: this.open,
+			mdiChevronDown
 		};
 	},
 	methods: {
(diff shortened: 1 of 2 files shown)


src/components/Kv/KvAccordionItem.vue
```diff
@@ -1,30 +1,26 @@
...
 	<div
-		class="kv-accordion"
-		:class="{
-			'kv-accordion--open' : isOpen,
-		}"
+		class="tw-border-b tw-border-tertiary tw-relative"
 	>
 		<button
-			class="kv-accordion__header-button"
+			class="tw-w-full tw-flex tw-justify-between tw-items-center tw-py-[0.75rem] tw-px-0
```
mcstover, line 6: Please use a style attribute for custom dimensions such as `tw-py-[0.75rem]` OR convert the size to use the closest measure in our design token/tailwind size matrix.

src/components/Kv/KvAccordionItem.vue
```diff
@@ -1,30 +1,26 @@
...
+			<span class="tw-flex-1">
 				<slot name="header"></slot>
 			</span>
-			<kv-icon
-				class="kv-accordion__header-icon"
-				name="small-chevron"
-				:from-sprite="true"
-				aria-hidden="true"
+			<kv-material-icon
+				class="tw-h-3 tw-w-[25px] tw-transition tw-transform tw-duration-500 tw-ease"
```
mcstover, line 17: Same as above for `tw-w-[25px]`. Should either become `tw-w-3` or move to `style="width: 25px;`

mcstover: I'd prefer we move towards our size/spacing scale rather than custom values.

src/components/Kv/KvAccordionItem.vue
emuvente: 0.75rem is 12px, which is 1.5 is our design token sizes. So `tw-py-1.5` should work.

src/components/Kv/KvAccordionItem.vue
emuvente: +1 to using `tw-w-3`

src/components/Kv/KvAccordionItem.vue
roger-in-kiva: Good, value replaced.

src/components/Kv/KvAccordionItem.vue
roger-in-kiva: Good, value replaced.

mcstover (approved):

emuvente (approved):
Nice!

src/components/Kv/KvAccordionItem.vue
```diff
@@ -1,30 +1,26 @@
 <template>
 	<div
-		class="kv-accordion"
-		:class="{
-			'kv-accordion--open' : isOpen,
-		}"
+		class="tw-border-b tw-border-tertiary tw-relative"
```
ryan-ludwig, line 3: To prevent a bottom border on the last entry the previous CSS used last-child. To bring that back with Tailwind, you can use `last:tw-border-b-0`

src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -1,14 +1,14 @@
 <template>
-	<div class="row">
-		<ol id="disclaimers" class="text-left">
+	<kv-page-container>
```
ryan-ludwig, line 2: I like that you removed `<div class="row">` here. However, I would suggest using wrapping it with `<kv-page-container>` in the context it's used in, which right now is TheFooter.vue. Rather than in the component itself.

This will make it so `<disclaimers-contentful>` can be used in other contexts and layouts, maybe taking up only half-of the screen width, etc. 

Just generally keeping layout handled by what is using the component, rather than putting layout in the component itself.

src/components/Kv/KvAccordionItem.vue
```diff
@@ -1,30 +1,26 @@
...
 				<slot name="header"></slot>
 			</span>
-			<kv-icon
-				class="kv-accordion__header-icon"
-				name="small-chevron"
-				:from-sprite="true"
-				aria-hidden="true"
+			<kv-material-icon
+				class="tw-h-3 tw-w-3 tw-transition tw-transform tw-duration-500 tw-ease"
+				:class="{ 'tw-rotate-180' : isOpen }"
```
ryan-ludwig, line 18: niiiice

src/components/Kv/KvAccordionItem.vue
roger-in-kiva: Good point.
Class is included now.

src/components/WwwFrame/DisclaimersContentful.vue
roger-in-kiva: Understood and agree!

src/components/Kv/KvAccordionItem.vue
```diff
@@ -1,30 +1,26 @@
...
-		class="kv-accordion"
-		:class="{
-			'kv-accordion--open' : isOpen,
-		}"
+		class="tw-border-b tw-border-tertiary tw-relative last:tw-border-b-0"
 	>
 		<button
-			class="kv-accordion__header-button"
+			class="tw-w-full tw-flex tw-justify-between tw-items-center tw-py-1.5 tw-px-0
+				tw-text-left disabled:tw-cursor-not-allowed disabled:tw-opacity-low"
```
ryan-ludwig, line 7: Sorry, just noticed one other thing. The old accordion had a hover color of blue. I think we need to add
`hover:tw-text-action-highlight`

ryan-ludwig (approved):

src/components/Kv/KvAccordionItem.vue
roger-in-kiva: Good.
Hover and focus options were added

PR #3459, opened 2021-11-30
fix: updated TOU, updated last changed date and highlighted new subst…

BoulderBrains:
…antive changes

[Core-307](https://kiva.atlassian.net/browse/CORE-307) (previously gd-187)

Updates to the TOU. 

I've implemented Carolyn's requested changes. I've removed the highlight from the previous updates, and applied highlighting to new substantive changes. Then I changed the last updated date at the top of the document. ✔️


emuvente (approved):
Looks good! We should probably wait to merge this until the 6th, so that it doesn't get deployed early as a part of a hotfix.

BoulderBrains:
@emuvente @mcstover I can wait to merge it, or I could change the _Last updated date_ to today (or tomorrow) and then we wouldn't have to worry about it. Do you see any harm in that or have a preference?

PR #3461, opened 2021-11-30
fix: fixes issue with double initializing of BT drop in

eddieferrer:
VUE-881


eddieferrer:
Wrapping the component in <keep-alive> will keep the components VUE uuid and Dropin.create will only get executed one time.

mcstover (approved):

PR #3462, opened 2021-12-01
fix: fixes the drop in error in a more robust way

eddieferrer:
Figured out what was happening with the braintree drop in error...

While the solution we hotfixed allows users to checkout, the component is still being mounted multiple times, and instead of the drop in error we are seeing a harmless `Cannot read properties of null (reading 'innerHTML')` -- https://sentry.io/organizations/kiva/issues/2826679294/?project=1201288&query=is%3Aunresolved

What is really happening is that the footer uses the `serverCacheKey` component option. The footer now includes the `disclaimers` component. This component contains dynamic queries so it cannot be cached in SSR. On page load, even if the disclaimer was hidden, the queries would execute, resulting in a `The client-side rendered virtual DOM tree is not matching the server-rendered content.` error. This would bail on the hydration and perform a full client-side render. Which would cause the BT Drop in to mount again and fire Dropin.create() multiple times.  The last of these dropIns would render the button disabled by erroring out. 

Our hotfix fixed the issue by erroring out with `Cannot read properties of null (reading 'innerHTML')` -- this is because in some instances of the remounting of BraintreeDropInInterface.vue, this is undefined. If we guarantee the document has rendered by wrapping our mount code like 
```
	mounted() {
		this.$nextTick(() => {
			// Prevents BT error in the case this component gets initialized multiple times
			const isElementEmpty = document?.getElementById('dropin-container')?.innerHTML === '';
			if (isElementEmpty) {
				this.getDropInAuthToken();
			}
		});
	},
```
We can then reproduce our original error.

The solution is to only cache the non dynamic part of the footer. I left the check for `isElementEmpty` with added optional chaining. This will prevent failures like this in the future and also not fire the entry to Sentry when document is undefined


src/components/AutoDeposit/AutoDepositDropInPaymentWrapper.vue
```diff
@@ -1,17 +1,15 @@
 <template>
 	<div class="row">
 		<div class="dropin-payment-holder small-12 columns">
-			<keep-alive>
```
eddieferrer, line 4: This does nothing

mcstover: ok oh well lolz

mcstover (approved):
This ensures we can preserve the static footer content remains cached!

emuvente (approved):

emuvente:
Great investigation!

PR #3484, opened 2021-12-08
feat: disclaimers footer component was upgraded to have the correct d…

roger-in-kiva:
DisclaimersContentful and TheFooter components were upgraded to have the correct design.


src/components/WwwFrame/TheFooter.vue
```diff
@@ -3,7 +3,9 @@
 				lg:tw-px-2 lg:tw-py-3 tw-bg-tertiary tw-text-primary
 				print:tw-p-0 print:tw-font-book print:tw-min-h-[auto]"
 	>
-		<disclaimers />
+		<kv-page-container class="tw-px-0">
```
ryan-ludwig, line 6: I'd prefer to find a solution using only DisclaimersContentfull.vue since KvPageContainer's padding changes through the breakpoints.

roger-in-kiva: One way is adding max width, margin and padding classes to Disclaimers component in order to adjust it to the footer's width.
Other way is using kv-page-container inside Disclaimers component.
Do you have other idea?

roger-in-kiva: Fixed @ryan-ludwig

emuvente (commented):
👍

emuvente (approved):

ryan-ludwig (approved):

PR #3490, opened 2021-12-09
VUE-830 frame for TopNav in Kiva Classic

ryan-ludwig:
All the dropdowns and search are kinda borked, but will hit those in the coming tickets.

Found it easiest to test the different states and screensizes using storybook.

Current status:
- [x] responsive layout
- [x] search dropdown
- [x] lend megadropdown
- [x] about menu
- [x] logged in custom profile image
- [ ] product and design review
- [ ] a11y testing


src/components/WwwFrame/SearchBar.vue
```diff
@@ -229,41 +239,11 @@ export default {
 <style lang="scss">
 @import 'settings';
 
-form.search-form {
```
ryan-ludwig, line 232: Will do more tailwind work here in the Search Bar ticket

roger-in-kiva (approved):
Nice, so clean!

ryan-ludwig:
Wondering if this should stay out of "typography_dev" for the time being. It would be bad to hold up that release in case something gets squirrelly with this header.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -618,356 +620,36 @@ export default {
...
-	position: relative;
-	display: flex;
-	align-items: stretch;
-	flex-wrap: nowrap;
-	height: $header-height;
-
-	@include breakpoint(large) {
-		height: $header-height-large;
-	}
+/* CSS grid areas to manage position changes across breakpoints without markup duplication */
```
emuvente, line 630: I really like this solution! It's probably worth noting though that grid-area is not supported by Internet Explorer at all. How much extra work do you think it would be to do this using flex/display/order properties?

src/components/WwwFrame/TheHeader.vue
```diff
@@ -502,35 +521,18 @@ export default {
...
 		preFetch: true,
 		result({ data }) {
-			this.isVisitor = !_get(data, 'my.userAccount.id');
-			this.isBorrower = _get(data, 'my.isBorrower');
-			this.loanId = _get(data, 'my.mostRecentBorrowedLoan.id');
-			this.trusteeId = _get(data, 'my.trustee.id');
-			this.basketCount = _get(data, 'shop.nonTrivialItemCount');
-			this.balance = Math.floor(_get(data, 'my.userAccount.balance'));
-			this.profilePic = _get(data, 'my.lender.image.url');
+			this.isVisitor = !data?.my?.userAccount?.id ?? false;
```
emuvente, line 529: I'm a bit confused by this line because it's not clear to me if the `!` will be applied to all of `data?.my?.userAccount?.id ?? false`, or just the `data?.my?.userAccount?.id` part. I think this could also be written as `this.isVisitor = !data?.my?.userAccount?.id;`, which would avoid that question altogether.

src/components/WwwFrame/TheHeader.vue
ryan-ludwig: That's a good question. I think there would have to be some duplication if we use flexbox, mobile search jumps to mind. That's not the end of the world though. I could time-box it to a half a day and see where we get?

Because of Tailwind's heavy use of CSS custom properties, and KvGrid using CSS grid, I assumed we had moved on from IE. We may want to pull up test.kiva.org in IE to see if it's even worth trying for the nav.

ryan-ludwig:
> Wondering if this should stay out of "typography_dev" for the time being. It would be bad to hold up that release in case something gets squirrelly with this header.

I think I'd like to just keep pushing to this branch. The lend dropdown mega menu is the next beast to take on.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -618,356 +614,43 @@ export default {
...
-			@include breakpoint(large) {
-				display: block;
-			}
-		}
-	}
+<style lang="postcss" scoped>
+.header__button {
+	@apply tw-flex tw-items-center tw-flex-shrink-0;
+	@apply tw-text-primary hover:tw-text-action-highlight hover:tw-no-underline focus:tw-no-underline;
+	@apply tw-h-8 md:tw-h-9 tw-whitespace-nowrap tw-flex-shrink-0;
```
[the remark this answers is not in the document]
ryan-ludwig: I went with 64 mobile, 72 tablet since those are values in our design system. And I couldn't rationalize why tablet would have a smaller nav than desktop.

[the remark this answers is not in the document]
mcstover: We're going with 64 on mobile and 72 for medium and desktop as those values fit our sizing scale.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -136,6 +136,7 @@ export default {
 				}`
 			}).subscribe({
 				next: ({ data }) => {
+					console.log(data);
```
mcstover, line 139: please remove at some point

src/components/WwwFrame/TheHeader.vue
```diff
@@ -571,10 +578,79 @@ export default {
...
+		onLendMenuPointerEnter(e) {
+			console.log('onLendMenuPointerEnter');
+			if (e.pointerType === 'touch') {
+				return;
+			}
+			this.isLendMenuDesired = true;
+			this.toggleLendMenu();
+		},
+		onLendMenuPointerLeave(e) {
+			console.log('onLendMenuPointerLeave');
```
mcstover, line 646: lots o consoles

mcstover (approved):

src/components/WwwFrame/LendMenu/LendListMenu.vue
```diff
@@ -166,126 +214,26 @@ export default {
...
-				this.$refs.regions.collapse();
-				this.$refs.regionCountries.forEach(region => region.collapse());
-			}
-			if (this.userId) {
-				this.$refs.myKiva.collapse();
-			}
-			if (this.hasSearches) {
-				this.$refs.searches.collapse();
-			}
+			// if (this.categories.length) {
```
ryan-ludwig, line 217: TODO: Look through this in the context of kv-accordion

src/components/WwwFrame/LendMenu/LendMegaMenu.vue
```diff
@@ -1,128 +1,217 @@
...
+								>
+									All loans
+								</router-link>
+							</li>
+						</ul>
+					</div>
+
+					<!-- my kiva -->
+					<!-- TODO: Where should this go? -->
+					<!-- My Kiva -->
```
ryan-ludwig, line 80: TODO: Figure this out with design

PR #3493, opened 2021-12-10
docs: workaround storybook console errors when stories are named Pages

ryan-ludwig:
Really not sure why this is happening but there's tons of console errors that look like
Screen Shot 2021-12-09 at 5 29 09 PM

Renaming from Pages to Page makes them all disappear 🤷


mcstover (approved):
Weird indeed, maybe it was trying to auto match to a directory somehow...

PR #3494, opened 2021-12-10
fix: ensure matching shows when applicable, clean up initial visibili…

mcstover:
…ty logic

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -378,10 +378,6 @@ export default {
 				this.lenderCountVisibility = true;
 				this.statScrollAnimation = true;
 			}
-
-			// if (this.lenderCountVisibility || this.matchingTextVisibility) {
-			// 	this.statScrollAnimation = true;
-			// }
 		},
 	},
 	methods: {
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -235,7 +235,7 @@
 							<span
 								class="tw-inline-block tw-align-middle"
 								key="loanMatchingText"
-								v-if="!statScrollAnimation"
+								v-if="!statScrollAnimation && !isMatchAtRisk"
 							>
 								<span
 									class="tw-text-h3 tw-inline-block tw-align-middle tw-px-1"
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -430,10 +431,10 @@ export default {
 			);
 		},
 		cycleStatsSlot() {
-			if (this.matchingText.length && !isMatchAtRisk) {
+			if (this.matchingText.length) {
 				const cycleSlotMachine = () => {
-					if (this.statScrollAnimation) {
-						this.statScrollAnimation = false;
+					if (!this.isMatchAtRisk) {
+						this.statScrollAnimation = !this.statScrollAnimation;
```
mcstover, line 437: This covers the previous if/else condition and we only enable this swapping if the loan is matching and not at risk.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -235,7 +235,7 @@
 							<span
 								class="tw-inline-block tw-align-middle"
 								key="loanMatchingText"
-								v-if="!statScrollAnimation"
+								v-if="!statScrollAnimation && !isMatchAtRisk"
```
mcstover, line 238: Hide this section if matching is at risk

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -376,9 +376,6 @@ export default {
 
 			if (this.status === 'fundraising' && this.numLenders > 0) {
 				this.lenderCountVisibility = true;
-			}
-
-			if (this.lenderCountVisibility && this.matching !== '') {
```
mcstover, line 381: This block wasn't needed b/c we can `statScrollAnimation` at the same time we determine `lenderCountVisibility` . Previously this check `(this.lenderCountVisibility && this.matching !== '')` was actually short-circuited b/c `this.matching` was always `undefined` thus we always activate `statScrollAnimation`.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -430,10 +427,10 @@ export default {
 			);
 		},
 		cycleStatsSlot() {
-			if (this.matchingText.length && !isMatchAtRisk) {
```
mcstover, line 433: This was broken b/c I forgot `this.` 🤦

emuvente (approved):

PR #3499, opened 2021-12-14
Credit banner vue 876

roger-in-kiva:
PromoCreditBanner component was created in order to have one component for BonusBanner and LendingRewardsBanner.

A story for SB was created to test it.


src/components/WwwFrame/PromotionalBanner/Banners/PromoCreditBanner.vue
```diff
@@ -0,0 +1,177 @@
+<template>
+	<div
+		v-if="lendingRewardOffered"
+		class="tw-bg-brand tw-text-white md:tw-py-1.5 md:tw-px-0"
+		v-kv-track-event="['TopNav','click-Promo','Lending Reward Banner']"
+	>
+		Make a Kiva loan <br class="tw-inline">and receive a $25 free credit to lend again.
+	</div>
+	<div v-else class="bonus-banner-holder tw-bg-brand tw-text-center tw-py-1.5 tw-px-5">
```
emuvente, line 9: This should only show if the bonus balance is greater than 0, `v-else-if="bonusBalance > 0"`. I think that will mean that you have to add a third element to this component for the `v-else` condition, `<div v-else></div>`, or use v-show instead.

roger-in-kiva: Thanks for that.

Original condition is this

               `currentActivePromo() {
			if (this.lendingRewardOffered) {
				return LendingRewardsBanner;
			}
			if (this.bonusBalance > 0) { // TODO: skip if on a checkout/basket page
				return BonusBanner;
			}
			return null;
		},`
		
So, I think is enough with `v-else-if="bonusBalance > 0".`

mcstover (approved):
I think this is good. Is there follow on work to use this in the UI app code to replace the existing banner and remove the swapping between small and large versions?

ryan-ludwig (approved):
Nice work!

ryan-ludwig:
> I think this is good. Is there follow on work to use this in the UI app code to replace the existing banner and remove the swapping between small and large versions?

Yes, I'll add this into the new Header work.

ryan-ludwig:
Since this replaces several components, can you delete those please and add to this PR.

emuvente (approved):
Great!

ryan-ludwig:
Thanks for removing the old banners @roger-in-kiva . I think this looks good. @mcstover  can you take a quick look please?

mcstover (approved):
👏

PR #3504, opened 2021-12-16
feat: add kiva classic style to causes settings components

eddieferrer:
GD-192
Screen Shot 2021-12-16 at 1 52 20 PM
Screen Shot 2021-12-16 at 1 52 11 PM

### DIFF
--- a/src/components/AutoDeposit/AutoDepositDropInPaymentWrapper.vue
+++ b/src/components/AutoDeposit/AutoDepositDropInPaymentWrapper.vue
@@ -10,16 +10,14 @@
 				:preselect-vaulted-payment-method="action === 'Registration'"
 				@transactions-enabled="enableConfirmButton = $event"
 			/>
-			<div id="dropin-button">
+			<div id="dropin-submit" class="tw-w-full">
 				<kv-button
 					class="tw-mb-2"
-					value="submit"
-					id="dropin-submit"
-					:state="!enableConfirmButton || submitting ? 'disabled' : ''"
+					:state="buttonState"
 					@click="submitDropInAutoDeposit"
 				>
 					<kv-icon name="lock" />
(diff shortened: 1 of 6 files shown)


src/components/AutoDeposit/AutoDepositDropInPaymentWrapper.vue
```diff
@@ -10,16 +10,14 @@
...
 				<kv-button
 					class="tw-mb-2"
-					value="submit"
-					id="dropin-submit"
-					:state="!enableConfirmButton || submitting ? 'disabled' : ''"
+					:state="buttonState"
 					@click="submitDropInAutoDeposit"
 				>
 					<kv-icon name="lock" />
-					Confirm <kv-loading-spinner v-if="submitting" />
```
eddieferrer, line 22: I think we can remove these in place spinner in favor of the button states

src/components/Forms/CausesUpdateForm.vue
```diff
@@ -95,7 +93,7 @@
 					<strong>Your contribution will:</strong>
 				</div>
 				<div class="column">
-					<kv-select v-model="form.categoryId" class="group-dropdown">
+					<kv-select v-model="form.categoryId" class="group-dropdown" id="causes-categories-select">
```
eddieferrer, line 96: KvSelect requires id prop

ryan-ludwig: The goal with the required id is to help remind that all inputs need `<label>` elements. I think you could use the "Your contribution will", or make a new `<label class="tw-sr-only" for="causes-category-select">Some descriptive label of the select</label>` nearby

ryan-ludwig (approved):
Small thing with the label, otherwise looks great! Thanks for doing this

PR #3506, opened 2021-12-20
Core 148 paying back repayment schedule

BoulderBrains:
Another step toward [core-148](https://kiva.atlassian.net/browse/CORE-148)

Repayment schedule

These changes cover the repayment schedule for Field Partner loans that are payingBack. 

If a repayment is received on time, that month's repayment is marked as paid. If the borrower is delinquent on their repayments we also mark that on the repayment schedule.

**Paying back loan in good standing:**
Screen Shot 2021-12-20 at 11 48 31 AM

**Paying back loan that's delinquent:** (page data was manipulated to get a delinquent row to show because we strip out repayment data in dev and our local dbs)
Screen Shot 2021-12-20 at 11 32 20 AM

### DIFF
--- a/src/components/BorrowerProfile/RepaymentSchedule.vue
+++ b/src/components/BorrowerProfile/RepaymentSchedule.vue
@@ -256,7 +256,8 @@ export default {
 					// push the monthly repayments to a new array to be used to check if repayments are on time
 					monthlyTotalRepayments.push(totalMonthlyPayment);
 
-					// iterate through the monthlyRepayments and add together each month's payments to get a total monthly payment amount
+					// iterate through the monthlyRepayments and add together each
+					// month's payments to get a total monthly payment amount
 					const totalMonthlyPaymentValue = monthlyTotalRepayments.reduce((runningTotal, monthlyAmount) => {
 						return runningTotal + parseFloat(monthlyAmount);
 					}, 0);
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -217,23 +253,46 @@ export default {
...
+						return runningTotal + parseFloat(monthlyAmount);
+					}, 0);
+
+					// set the repaid boolean
+					repaid = this.repaidAmount > totalMonthlyPaymentValue;
+
+					const now = new Date();
+					const parsedRepaymentDate = parseISO(repaymentDate);
+					// if a payment is not repaid and the repayment data is before now, then mark the loan delinquent
+					delinquent = !repaid && isBefore(parsedRepaymentDate, now);
```
mcstover, line 271: Seems like this should prevent the "Future" dates from being shown as delinquent but that doesn't appear to be the case in the attached screenshot which shows delinquency for several months into the future. Please see if you can get those future date to reflect the "expected" state instead of delinquent.

BoulderBrains: @mcstover After looking into this a little, I think it was a result of how if forced delinquent repayments to show. In order to do that I changed line 271

delinquent = !repaid && isBefore(parsedRepaymentDate, now);
I updated that to: 
delinquent = !repaid && !isBefore(parsedRepaymentDate, now);

Doing this caused all the rows to show as delinquent. With real data, I don't think that will be an issue.

mcstover (approved):
Good work!

PR #3507, opened 2021-12-20
feat(auth): set dev graphql query params when fake authentication cookie is present

emuvente:
Allows for faking authentication in development and testing. If this fake authentication is attempted in environments that do not allow it (like prod) there will be odd behavior, like the login sync not happening and baskets being lost, but no protected information will be exposed by the api.

### DIFF
--- a/config/dev-vm.js
+++ b/config/dev-vm.js
@@ -48,6 +48,7 @@ module.exports = merge(base, {
 				ouGKxT4mE4wQEKqpfsHSE96c9rHXQqZF: 'https://dev-vm-01.kiva.org/ui-login?force=true',
 			},
 			enable: true,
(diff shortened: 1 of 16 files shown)


mcstover (approved):

torgie-kiva (approved):
I got a little confused by `noteLoggedIn`, because it looked a lot like a typo of `notLoggedIn` :P But I figured it out.

JoeTravisKiva (approved):
Looks good to me. The only thing I might consider is either disabling this explicitly by default and only enabling it on certain environments, or just ensuring it won't run on prod.

mcstover (approved):

JoeTravisKiva (approved):
Nice! Those log changes in particular look slick!

PR #3509, opened 2021-12-21
feat: update causes subscription payment method

eddieferrer:
GD-194

See ticket for more details
Screen Shot 2021-12-21 at 12 40 02 PM
Screen Shot 2021-12-21 at 12 39 47 PM

### DIFF
--- a/src/components/Causes/CausesDropInPaymentWrapper.vue
+++ b/src/components/Causes/CausesDropInPaymentWrapper.vue
@@ -33,6 +33,8 @@ import * as Sentry from '@sentry/vue';
 import braintreeDropinError from '@/plugins/braintree-dropin-error-mixin';
 
 import createSubscription from '@/graphql/mutation/createSubscription.graphql';
+import updateSubscriptionPaymentMethod from '@/graphql/mutation/updateSubscriptionPaymentMethod.graphql';
+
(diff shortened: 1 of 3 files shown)


src/components/Causes/CausesDropInPaymentWrapper.vue
```diff
@@ -110,10 +112,48 @@ export default {
 				});
 		},
 		doBraintreeCausesSubscription(nonce, deviceData, paymentType) {
-			/** ! TODO add if (this.action === 'Update') logic here similar to
-			* src/components/MonthlyGood/MonthlyGoodDropInPaymentWrapper.vue
-			* for editing a cause subscription GD-155
-			*/
+			if (this.action === 'Update') {
+				console.log(nonce, deviceData, paymentType);
```
mcstover, line 116: bump

mcstover (approved):
Full feature parity!

src/components/Causes/CausesDropInPaymentWrapper.vue
eddieferrer: 🤦  Did we not have a lint warning for this at one time?

mcstover: I thought we had a warning but it def doesn't block anything...I don't see it either at this point in the builds

PR #3511, opened 2021-12-22
feat: causes update payment method kiva classic styling

eddieferrer:
GD-192
GD-194
Screen Shot 2021-12-22 at 10 10 41 AM
Screen Shot 2021-12-22 at 10 10 32 AM


ryan-ludwig (approved):
nice!

PR #3515, opened 2022-01-03
fix: add storycard spacing between storycard content and cta

eddieferrer:
CORE-359

### DIFF
--- a/src/components/Contentful/StoryCard.vue
+++ b/src/components/Contentful/StoryCard.vue
@@ -46,7 +46,7 @@
 				:html="cardTitle"
 			/>
 			<dynamic-rich-text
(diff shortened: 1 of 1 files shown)


ryan-ludwig (approved):

mcstover (approved):
Should we also add this to the instance above?

eddieferrer:
> Should we also add this to the instance above?

You mean to the cardTitle for space between the headline and content? That would be ok, but right now, there is some additional spacing created on kv-contentful-img so we would need to fix that first to not add twice the top padding

mcstover:
@eddieferrer def find to hold on any other changes if this helps out in the now

PR #3519, opened 2022-01-04
feat: add FB tracking event for Causes signup

eddieferrer:
GD-189

### DIFF
--- a/src/pages/Causes/SignupForm.vue
+++ b/src/pages/Causes/SignupForm.vue
@@ -167,7 +167,13 @@ export default {
 		completeCausesBraintree() {
 			// disable form inputs while routing to thanks
 			this.success = true;
+			// fire events
 			this.$kvTrackEvent('Registration', 'successful-causes-reg', 'register-causes');
+			// Track facebook event
+			if (typeof window !== 'undefined' && typeof fbq === 'function') {
+				window.fbq('track', 'CausesSignUp');
+			}
+			// route to thanks
 			this.$router.push({
 				path: '/causes/thanks',
 			});


eddieferrer:
@mcstover this is the last item needed for Causes web release

src/pages/Causes/SignupForm.vue
```diff
@@ -167,7 +167,13 @@ export default {
 		completeCausesBraintree() {
 			// disable form inputs while routing to thanks
 			this.success = true;
+			// fire events
 			this.$kvTrackEvent('Registration', 'successful-causes-reg', 'register-causes');
+			// Track facebook event
+			if (typeof window !== 'undefined' && typeof fbq === 'function') {
+				window.fbq('track', 'CompleteRegistration');
```
mcstover, line 174: I think `CompleteRegistration` should have a signifier that points to "Causes", maybe "CompleteCauseRegistration" so that we don't confuse this with an account Registration.

src/pages/Causes/SignupForm.vue
```diff
@@ -167,7 +167,13 @@ export default {
 		completeCausesBraintree() {
 			// disable form inputs while routing to thanks
 			this.success = true;
+			// fire events
 			this.$kvTrackEvent('Registration', 'successful-causes-reg', 'register-causes');
+			// Track facebook event
+			if (typeof window !== 'undefined' && typeof fbq === 'function') {
+				window.fbq('track', 'CausesSignUp');
```
eddieferrer, line 174: Verified with product, see ticket. Let's go with this so it matches the event for MG sign ups

mcstover (approved):

PR #3520, opened 2022-01-04
feat: core-357 adding anonmizationLevel check for repayment schedule,…

BoulderBrains:
… removing isPartnerLoan checks

[Core-357](https://kiva.atlassian.net/browse/CORE-357)

In this PR: 
- I've added an additional check in the detailsTabs.vue component to check the anonomization level of the loan before loading any of the repayment schedule component/data.
- I've removed all checks preventing a "Detailed repayment schedule" link from displaying on direct loans.
- I've added some conditional text under the table for direct loans in repayment to bring it in line with the one in production.
- I've verified that all data should be in place for directLoans in payingBack status to the best of my abilities without actually having access to what the expected repayments looks like on a direct loan

### DIFF
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -50,6 +50,7 @@
 						@show-definition="showDefinition"
 					/>
 					<repayment-schedule
+						v-if="loan.status !== 'fundraising' || loan.amonimizationLevel !== 'none'"
 						:loan-id="loanId"
 						:status="loan.status"
 					/>
@@ -260,6 +261,7 @@ export default {
 							lenderRepaymentTerm
 							repaymentInterval
 							disbursalDate
+							anonymizationLevel
 							terms {
 								currency
 								flexibleFundraisingEnabled
@@ -315,6 +317,7 @@ export default {
 				this.loan.disbursalDate = loan?.disbursalDate ?? '';
 				this.loan.status = loan?.status ?? '';
 				this.loan.name = loan?.name ?? '';
+				this.loan.anonymizationLevel = loan?.anonymizationLevel ?? 'none';
 
 				this.partner.arrearsRate = partner?.arrearsRate ?? 0;
 				this.partner.avgBorrowerCost = partner?.avgBorrowerCost ?? 0;
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -50,7 +50,7 @@
 						@show-definition="showDefinition"
 					/>
 					<repayment-schedule
-						v-if="loan.amonimizationLevel !== 'none'"
+						v-if="loan.anonymizationLevel !== 'full'"
 						:loan-id="loanId"
 						:status="loan.status"
 					/>
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -99,7 +99,6 @@
 <script>
 import gql from 'graphql-tag';
 import { formatContentGroupsFlat } from '@/util/contentfulUtils';
-import { createIntersectionObserver } from '@/util/observerUtils';
 // TODO: replace the loading placeholder with component from kv-components when available.
 import KvLoadingPlaceholder from '@/components/Kv/KvLoadingPlaceholder';
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -50,7 +50,7 @@
 						@show-definition="showDefinition"
 					/>
 					<repayment-schedule
-						v-if="loan.status !== 'fundraising' || loan.amonimizationLevel !== 'none'"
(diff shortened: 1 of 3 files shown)


src/components/BorrowerProfile/DetailsTabs.vue
```diff
@@ -50,6 +50,7 @@
 						@show-definition="showDefinition"
 					/>
 					<repayment-schedule
+						v-if="loan.status !== 'fundraising' || loan.amonimizationLevel !== 'none'"
```
mcstover, line 53: Need to fix spelling for `anonymizationLevel` and update the check to be `!== 'full'`. Most loans are set to `none` so this current check would exclude almost all loans.

In the near future we can get more sophisticated to check for `public` too.

src/components/BorrowerProfile/DetailsTabs.vue
```diff
@@ -50,6 +50,7 @@
 						@show-definition="showDefinition"
 					/>
 					<repayment-schedule
+						v-if="loan.anonymizationLevel !== 'full'"
```
mcstover, line 53: Based on the legacy code I think there is one other check that needs to happen at this level.
- Create computed property that returns true/false if the disbursalDate is past now
- Combine that condition with loan type so we can omit the repayment schedule for Direct loans if the disbursalDate hasn't occurred...

This could be another ticket.

mcstover (approved):

src/components/BorrowerProfile/DetailsTabs.vue
BoulderBrains: @mcstover
After looking through the code more last night and this morning, I don't think that check is actually correct. It's my understanding that a directLoan without a disbursal date has not been disbursed. The loan fundraises to completion then it transitions to funded. At that point, the loan takes a few days to process until it's disbursed, at that point the loan gets a disbursal date. 
If a directLoan doesn't have a disbursal date then we show the following portion of code, 
`<div
				v-if="!isPartnerLoan && this.status === 'fundraising'"
				class="tw-prose"
			>
				<p>
					This loan is for {{ loanAmountFormatted }}.
				</p>
				<p>
					
					Repayments on this loan will be made in monthly installments of {{ calculateMonthlyPayment }} USD over a period of {{ lenderRepaymentTerm }} months. After the funds are disbursed, the borrower(s) will have the standard 1 month before the first payment is due, and may have an additional grace period per the terms outlined in their loan agreement. The detailed repayment schedule will be published here at the time that the funds for this loan are disbursed.<br>
				</p>
				<p>
					
					Disbursement and repayments will be made via PayPal, a web-based payment system. Repayments made on delinquent loans will be applied toward the oldest payment due until the loan becomes current.
				</p>
			</div>`
So if I put this disbursal date check to hide the repayment schedule the code above would never be shown.

BoulderBrains, line 53: Prevent repayment schedule from loading if loan is anonymous.

src/components/BorrowerProfile/JumpLinks.vue
```diff
@@ -13,7 +13,7 @@
 			Borrower story
 		</a>
 		<a
-			class="tw-text-primary tw-pl-1 tw-pr-1 tw-font-medium"
+			class="tw-text-primary tw-pl-1 tw-pr-1 tw-font-medium tw-cursor-pointer"
 			v-kv-track-event="['Borrower profile', 'click-jump-link', 'Loan details']"
```
BoulderBrains, line 17: minor change to fix the cursor hover on the jump links.

src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -14,8 +11,7 @@
 			title="Loan repayment schedule"
 			@lightbox-closed="closeLightbox"
 		>
-			<!-- Field Partner loan fundraising -->
-			<div v-if="isPartnerLoan">
+			<div v-if="isPartnerLoan || !isPartnerLoan && loanDisbursed">
```
BoulderBrains, line 14: instead of checking if loan is fundraising, we check if the loan is disbursed.

src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -193,9 +196,7 @@ export default {
 				this.repaidAmount = data?.lend?.loan?.paidAmount || 0;
 				this.loanAmount = data?.lend?.loan?.loanAmount || 0;
 				this.lenderRepaymentTerm = data?.lend?.loan?.terms?.lenderRepaymentTerm || 0;
-				if (this.isPartnerLoan) {
-					this.firstRepaymentDate = this.repaymentSchedule[0].dueToKivaDate || '';
-				}
+				this.firstRepaymentDate = this.repaymentSchedule[0]?.dueToKivaDate || '';
```
BoulderBrains, line 199: This was throwing a console error, so I fixed it up.

src/components/BorrowerProfile/DetailsTabs.vue
```diff
@@ -395,10 +369,8 @@ export default {
 		},
 	},
 	mounted() {
-		this.createObserver();
-	},
-	beforeDestroy() {
-		this.destroyObserver();
+		this.loadData();
```
mcstover, line 372: We removed this as the new jump link was jumping to a missing or empty section.

mcstover (approved):

PR #3521, opened 2022-01-05
fix: kivaClassic styling tweaks for Causes cancellation and BT drop i…

eddieferrer:
…n wrappers

GD-204
Screen Shot 2022-01-05 at 9 49 15 AM
Screen Shot 2022-01-05 at 9 38 08 AM

* KivaClassic for Causes cancel subscription lightbox
* Some touchups to drop in payment wrappers - make button full width, style attribution text

### DIFF
--- a/src/components/Subscriptions/SubscriptionsCausesCancellationFlow.vue
+++ b/src/components/Subscriptions/SubscriptionsCausesCancellationFlow.vue
@@ -8,6 +8,18 @@
 		<div class="causes-cancel-lightbox__content">
 			<label>
 				<textarea
+					class="
+						tw-w-full
+						tw-p-2
+						tw-mb-4
+						tw-border tw-border-tertiary
+						tw-rounded-sm
+						tw-appearance-none
+						tw-text-base
+						tw-bg-primary
+						tw-ring-inset
+						focus:tw-outline-none focus:tw-ring-2 focus:tw-ring-action focus:tw-border-transparent
+					"
 					name="reasonTextArea"
 					rows="8"
 					:placeholder="textAreaPlaceholder"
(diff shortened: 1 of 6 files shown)


src/components/Subscriptions/SubscriptionsCausesCancellationFlow.vue
```diff
@@ -8,6 +8,18 @@
 		<div class="causes-cancel-lightbox__content">
 			<label>
 				<textarea
+					class="
```
ryan-ludwig, line 11: Nice, think we should make a KvTextArea to keep these consistent?

ryan-ludwig (approved):
Nice!

src/pages/Donate/DonateFormDropInPaymentWrapper.vue
```diff
@@ -231,28 +231,23 @@ export default {
...
-
-		.icon-lock {
-			height: rem-calc(20);
-			width: rem-calc(20);
-			fill: white;
-			top: rem-calc(3);
-			position: relative;
-			margin-right: rem-calc(8);
-		}
+	.icon-lock {
```
ryan-ludwig, line 234: Do we need these loading and icon styles anymore?

src/components/Subscriptions/SubscriptionsCausesCancellationFlow.vue
eddieferrer: Yeah that would be nice, it would probably be a really simple component

src/pages/Donate/DonateFormDropInPaymentWrapper.vue
eddieferrer: Good call, these are handled in the global drop in style, fixed the wrapper class to get these to display

PR #3524, opened 2022-01-05
kv-material-icon Safari fixes

ryan-ludwig:
Adds a width and height classes to all `<kv-material-icon>`s.  If a width and height class aren't specified Safari lets the SVG icon take all available space, where Chrome and Firefox use the SVG's intrinsic size. 

Brings in latest kv-components which has these same fixes.

Fixes stuff like this happening on Safari (mobile and desktop I believe)
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.27.0",
+	"version": "2.34.1",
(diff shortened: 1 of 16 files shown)


mcstover (approved):
Thank you sir

BoulderBrains (approved):
Nice

PR #3526, opened 2022-01-06
feat: causes homepage redirect experiment

eddieferrer:
GD-205

~~Is this the best approach for this?~~

~~Unfortunately, this loads the homepage, then the created hook will redirect to the landing page. I could not find a way to avoid this. Trying to handle the experiment in preFetch doesn't work because the experiment depends on the user being on mobile, which the server does not know about in preFetch.~~

When the homepage loads, if new user and is mobile, assign the experiment, fire the event and redirect to '/lp/causes/' if experiment version is shown

On repeated visits, we can read the session cookie and redirect in the preFetch to avoid having to load the homepage. This should work as expected since only  eligible users should have the experiment assignment.

If a user is assigned to the experiment and they sign up (becoming a known user) they will still have the experiment assignment and will get redirected on prefetch on a repeated visit, this is an acceptable behavior (see GD-205)

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -4,12 +4,20 @@
 
 <script>
 import gql from 'graphql-tag';
+import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
 import { preFetchAll } from '@/util/apolloPreFetch';
+import { parseExpCookie } from '@/util/experimentUtils';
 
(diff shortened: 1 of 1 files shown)


mcstover (approved):
For the experiment group, this setup could end up firing both the initial pageview call and the redirect pageview due to the isMobile check requiring this loading in the browser. Thus, we'll see see an odd pageview count for  `/` for experiment visitors that are redirected.

mcstover (approved):

PR #3528, opened 2022-01-10
feat: core-148 mobile view of repayment schedule

BoulderBrains:
[Core-148](https://kiva.atlassian.net/browse/CORE-148) These changes get the mobile view of the repayment schedule in place.

[Figma spec](https://www.figma.com/file/EI9I5sR9mATRxIV8hBWKwm/Repayment-Schedule---Borrower-Profile-v2-Nov.2021%5BNT%5D?node-id=1%3A8)

A few different versions of this to take a look at. 
**Loan payment incoming:**
Screen Shot 2022-01-10 at 4 24 36 PM

**Loan payment made**
Screen Shot 2022-01-10 at 4 25 11 PM

**Loan payment deliquent**
Screen Shot 2022-01-10 at 4 25 35 PM

### DIFF
--- a/src/components/BorrowerProfile/RepaymentSchedule.vue
+++ b/src/components/BorrowerProfile/RepaymentSchedule.vue
@@ -34,7 +34,7 @@
 						:key="index"
 						class="tw-mb-1"
 					>
-						<div class="tw-bg-secondary tw-rounded tw-text-center tw-mb-2 tw-pb-1.5">
+						<td class="tw-inline-block tw-w-full tw-bg-secondary tw-rounded tw-text-center tw-mb-2 tw-pb-1.5">
 							<p class="tw-text-h4 tw-py-1.5">
 								{{ repayment.formattedRepaymentDate }}
 							</p>
@@ -47,7 +47,8 @@
 							</p>
 							<!-- if payment is received -->
 							<p
-								class="tw-bg-primary tw-mx-auto tw-w-[11.5rem] tw-py-1 tw-rounded"
+								class="tw-bg-primary tw-mx-auto tw-py-1 tw-rounded"
+								style="width: 11.5rem;"
 								v-if="repayment.repaid && !repayment.delinquent"
 							>
 								<kv-material-icon
@@ -58,7 +59,8 @@
 							</p>
 							<!-- if payment is not received on time -->
 							<p
-								class="tw-bg-primary tw-mx-auto tw-w-[7.5rem] tw-py-1 tw-rounded"
+								class="tw-bg-primary tw-mx-auto tw-py-1 tw-rounded"
+								style="width: 7.5rem;"
 								v-if="!repayment.repaid && repayment.delinquent"
 							>
 								<kv-material-icon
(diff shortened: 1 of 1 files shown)


mcstover (approved):
Need to fix up that syntax issue but otherwise this looks really nice!

src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -26,7 +26,53 @@
...
-				<table class="tw-table-auto">
+
+				<!-- Table for small screens -->
+				<table class="md:tw-hidden tw-w-full">
+					<tr
+						v-for="(repayment, index) in parsedRepaymentSchedule"
+						:key="index"
+						class="tw-mb-1"
+					>
+						<div class="tw-bg-secondary tw-rounded tw-text-center tw-mb-2 tw-pb-1.5">
```
ryan-ludwig, line 37: should this be a `<td>`?

src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -26,7 +26,53 @@
...
+							<hr class="tw-mb-1.5 tw-mx-1.5">
+							<p class="tw-mb-1.5">
+								Expected: {{ repayment.formattedMonthlyPayment }}
+							</p>
+							<p v-if="!repayment.repaid && !repayment.delinquent">
+								Available {{ repayment.formattedRepaymentDate }}
+							</p>
+							<!-- if payment is received -->
+							<p
+								class="tw-bg-primary tw-mx-auto tw-w-[11.5rem] tw-py-1 tw-rounded"
```
ryan-ludwig, line 50: Ideally try to avoid these bracketed `tw-w-[555rem]` classes because they end up going into the stylesheet that is loaded on all pages even though they're probably just used on this page.  Over time the stylesheet will get bloated.

For these, adding a `style="width: 11.5rem;"` would work well.

The main reason I've seen to use the bracketed styles is when you have to change things responsively like
`tw-w-[11.5rem] lg:tw-w-[12rem]`. Since you can't do media queries with inline styles.

src/components/BorrowerProfile/RepaymentSchedule.vue
```diff
@@ -26,7 +26,53 @@
...
+							>
+								<kv-material-icon
+									:icon="mdiCheckboxMarkedCircle"
+									class="tw-w-3 tw-h-3 tw-text-brand-700 tw-align-middle"
+								/>
+								Repayment received
+							</p>
+							<!-- if payment is not received on time -->
+							<p
+								class="tw-bg-primary tw-mx-auto tw-w-[7.5rem] tw-py-1 tw-rounded"
```
ryan-ludwig, line 61: Same bracket comment here

src/components/BorrowerProfile/RepaymentSchedule.vue
BoulderBrains: Thanks for the explanation @ryan-ludwig

ryan-ludwig (approved):
There's an eslint thing squawking about line breaks, but otherwise looks great!

PR #3535, opened 2022-01-12
fix: make CC landing page status bar stick at the right spot, because of the new header height

ryan-ludwig:


mcstover (approved):
Good catch!

PR #3538, opened 2022-01-12
fix: core-151 setting width to full, so margin can have space

BoulderBrains:
In my code review yesterday, I was testing on a loan who's location and activity were so long it pushed my bookmark into the right place like this.
Screen Shot 2022-01-12 at 9 23 38 AM
 
Once things got up to dev, I realized that I forgot to set the 'tw-w-full' class on the parent component allowing it to take up the full width of the container and then giving space for my 'tw-ml-auto' on the bookmark component.to work.

**Error state:**
Screen Shot 2022-01-12 at 9 23 00 AM

**After 'tw-w-full' class is added:** FIXED
Screen Shot 2022-01-12 at 9 27 44 AM

### DIFF
--- a/src/components/BorrowerProfile/LoanBookmark.vue
+++ b/src/components/BorrowerProfile/LoanBookmark.vue
@@ -2,7 +2,7 @@
 	<div>
 		<button
 			v-if="!isBookmarked"
-			class="tw-text-action tw-inline-flex tw-p-1 tw-cursor-pointer tw-whitespace-nowrap"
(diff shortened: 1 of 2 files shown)


ryan-ludwig (approved):

ryan-ludwig:
This is totally unrelated, but should the font weight of the button be `font-medium`? I checked the Figma and it wasn't clear to me what weight was used, but it _looked_ like the others on the page which I think are all medium.

BoulderBrains:
Good eye @ryan-ludwig, thanks for that, I'll make that update before merging.

mcstover (approved):

PR #3539, opened 2022-01-12
Core 148 fix

BoulderBrains:
During QA of the repayment schedule Lauren happened to look at a "Funded" loan in development while being logged out, in that situation she was getting this result, as the check to show the repayment schedule was only looking at the loans "anonymizationLevel".
[Loan](https://www.dev.kiva.org/lend-beta/2293062) 
Screen Shot 2022-01-12 at 9 43 35 AM
So I was showing the repayment schedule but no data was coming through for it.  

I compared the older borrower profile, logged in and logged out states and noticed that for funded loans we were only showing the repayment schedule while being logged in. 

The check I've now put in place: 
```
displayRepaymentSchedule() {
	// check anonymization of loan, if it's not set to
	// full anonymization, continue
	if (this.loan.anonymizationLevel !== 'full') {
		// if loan is fundraising, always display the repayment schedule
		if (this.loan.status === 'fundraising') {
			return true;
		}
		// otherwise look at the isSupporter flag to determine if
		// the repayment schedule should be shown to current user,
		// if they are logged in
		return this.isSupporter;
	}
	// if loan is set to full anonymization, return false
	// because the repayment schedule should not be shown.
	return false;
}
```

Now we're only showing the repayment schedule: 
- if the loan.anonymizationLevel !== 'full'
& if the loan is fundrasing
& if the loan is not fundraising, if the user is a supporter of that loan

### DIFF
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -198,7 +198,7 @@ export default {
 			return `tab-panel-${this.name}-trustee`;
 		},
 		isSupporter() {
-			return this.loan.lentTo;
+			return this.loan?.lentTo;
 		},
 		displayRepaymentSchedule() {
 			// check anonymization of loan, if it's not set to
--- a/src/components/BorrowerProfile/DetailsTabs.vue
+++ b/src/components/BorrowerProfile/DetailsTabs.vue
@@ -207,12 +207,11 @@ export default {
 				// if loan is fundraising, always display the repayment schedule
 				if (this.loan.status === 'fundraising') {
 					return true;
-				} else {
-					// otherwise look at the isSupporter flag to determine if
-					// the repayment schedule should be shown to current user,
-					// if they are logged in
-					return this.isSupporter;
 				}
+				// otherwise look at the isSupporter flag to determine if
+				// the repayment schedule should be shown to current user,
+				// if they are logged in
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/DetailsTabs.vue
```diff
@@ -197,6 +197,26 @@ export default {
 		trusteeTabId() {
 			return `tab-panel-${this.name}-trustee`;
 		},
+		isSupporter() {
+			return this.loan.lentTo;
```
mcstover, line 201: prob want too use optional chaining on `loan` here.

mcstover: Same below

BoulderBrains: @mcstover Below are checks looking at the value and not assigning it. It was my understanding that optional chaining is only needed when making an assignment to a const. Is that wrong?

mcstover: Yep, that's good. Please add your "[nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)" operator so this returns `false` if it fails.

mcstover (approved):

ryan-ludwig (approved):

PR #3540, opened 2022-01-12
feat: add video options for dynamicHeroClassic

eddieferrer:
GD-199

Takes a uiSetting object like:

```
    "video": {
        "autoplay": true,
        "muted": true
    }
```

Default for component if setting is missing is our previous options of autoplay true and mute true. if Options are not default display the video controls (So user can mute video and/or click play)

TODO: 
Add to Setting details to: `admin-component-reference`
Add these changes to component in `cms-page-server`

Important note: _In some browsers (e.g. Chrome 70.0) autoplay doesn't work if no muted attribute is present._


mcstover (commented):
Very useful! Good note on the compatibility of autoplay which we'll want to state in our docs too!

PR #3541, opened 2022-01-12
test: fix selector since the input is already visible

ryan-ludwig:


test/e2e/specs/HeaderNavigation.spec.js
```diff
@@ -11,8 +11,8 @@ describe('Header Navigation', () => {
 
 		// Go to the home page
 		cy.visit('/');
-		// Open the search bar in the header
-		cy.contains('Open Search').click();
+		// Click the search bar in the header
+		cy.get('#top-nav-search').click();
 		// Type 'f' in the currently focused element (the search bar)
 		cy.focused().type('f');
```
emuvente, line 17: Because the search is always open on desktop now, I think we can get the search bar by the placeholder text, and that way we don't need to click the search bar or use `cy.focused()`:
```
// Type 'f' in the search bar
cy.findByPlaceholderText('Search all loans').type('f');
```

emuvente (approved):
👍

PR #3545, opened 2022-01-13
feat: update kiva credit checkout button to kiva classic

mcstover:


BoulderBrains (approved):

src/components/Checkout/KivaCreditPayment.vue
```diff
@@ -1,18 +1,19 @@
 <template>
 	<kv-button
-		type="submit"
-		class="smaller"
+		value="submit"
```
ryan-ludwig, line 3: Assuming this is inside a form we probably want it to remain `type="submit"` instead of value

mcstover: Hmmm, yea, maybe I need to add the `prevent` modifier back to the click b/c we don't actually submit the form unless JS isn't loaded...

PR #3557, opened 2022-01-18
fix: guard ad upsell readQuery

mcstover:


mcstover:
@emuvente Any thoughts on why this might throw an Invariant Error. Guarding this way to prevent 500.

mcstover:
This wasn't an issue in DEV so I'm thinking that missing most of our new service endpoints in QA is causing this...this exp checks `mySubscriptions`.

PR #3564, opened 2022-01-19
fix: update methods for basket-bar visibility to use route. ensure it…

mcstover:
…'s hidden on checkout

### DIFF
--- a/src/components/WwwFrame/TheBasketBar.vue
+++ b/src/components/WwwFrame/TheBasketBar.vue
@@ -72,7 +72,7 @@ export default {
 		},
 		hideBasketBar() {
 			// hide this banner on managed lending landing + checkout pages
-			const routeExclusions = ['/join-team', '/checkout'];
+			const routeExclusions = ['/join-team', '/checkout', '/lend-beta'];
 			const routePath = this.$route?.path;
 			const matchedRoutes = routeExclusions.filter(item => {
 				return routePath.indexOf(item) !== -1;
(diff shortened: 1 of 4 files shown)


src/components/WwwFrame/TheBasketBar.vue
```diff
@@ -1,8 +1,8 @@
 <template>
 	<!-- eslint-disable max-len -->
 	<div
-		v-if="count > 0"
-		class="tw-flex md:tw-hidden tw-fixed tw-z-sticky"
+		v-if="count > 0 && !hideBasketBar"
+		class="basket-bar tw-flex md:tw-hidden tw-fixed tw-z-sticky"
```
mcstover, line 5: Prob could have fixed things by just adding this class but worked up the no css approach instead. Then realized this class is used for locating this component from the monthly good selector that can show on homepage so had to add it anyway...

src/components/WwwFrame/TheBasketBar.vue
```diff
@@ -69,6 +69,15 @@ export default {
 	computed: {
 		basketLink() {
 			return this.corporate ? this.addHashToRoute('show-basket') : '/basket';
+		},
+		hideBasketBar() {
+			// hide this banner on managed lending landing + checkout pages
+			const routeExclusions = ['/join-team', '/checkout'];
```
ryan-ludwig, line 75: We have some route exclusions on `TheBasketBar.vue`'s parent, `WwwPage.vue` that determines if basket bar should be on the page. I think these route exclusions would make more sense to go there.

ryan-ludwig: Or we consolidate those route rules into this component instead.

emuvente (approved):
👍

ryan-ludwig (approved):

PR #3568, opened 2022-01-19
fix: add accessible text to social media icons

ryan-ludwig:
Found from the lighthouse report

### DIFF
--- a/src/components/WwwFrame/TheFooter.vue
+++ b/src/components/WwwFrame/TheFooter.vue
@@ -347,6 +347,7 @@
 				<div class="tw-my-2">
 					<a
 						href="//www.facebook.com/kiva/"
+						class="tw-inline-flex"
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheFooter.vue
```diff
@@ -347,6 +347,7 @@
 				<div class="tw-my-2">
 					<a
 						href="//www.facebook.com/kiva/"
+						class="tw-inline-flex"
```
ryan-ludwig, line 350: On hover, was seeing a blank space with an underline because of the new space added before the invisible span. Inline-flex takes care of that.

mcstover (approved):

emuvente (approved):
Nice find!

roger-in-kiva (approved):

PR #3570, opened 2022-01-19
refactor: core-370 removing scss from file and replacing with tailwin…

BoulderBrains:
…ds classes

[Core-370](https://kiva.atlassian.net/browse/CORE-370)

Replaced scss in DonateRepaymensToggle.vue with tailwinds classes & reworked DonateRepaymentToggle to display the label on the left of the checkbox. I've also added data-testid classes on all elements with ids within this files during my pass through.

**Donate repayment checkout option**

**SM screen:**
Screen Shot 2022-01-27 at 10 01 22 AM

**MED screen:**
Screen Shot 2022-01-27 at 10 01 30 AM

**LG screen:**
Screen Shot 2022-01-27 at 10 01 40 AM

To Test: 
- Go to the checkout page with a loan in your basket.
- Set the donation amount to 0
The donate repayments option with be shown.

### DIFF
--- a/src/components/Checkout/DonateRepaymentsToggle.vue
+++ b/src/components/Checkout/DonateRepaymentsToggle.vue
@@ -2,11 +2,9 @@
 	<div>
 		<div
 			v-if="showToggle"
-			class="tw-mt-2"
-			:class="{ 'tw-flex' : checkboxRight }"
+			class="tw-mt-2 tw-flex donate-repayments-toggle"
 		>
 			<button
-				v-if="checkboxRight"
 				class="tw-flex-1 tw-pt-0.5 md:tw-pt-0 tw-text-base
 					tw-font-medium tw-text-left md:tw-text-right tw-mr-2
 					tw-cursor-pointer"
@@ -16,17 +14,17 @@
 				Donate loan repayments instead?
 			</button>
 			<kv-checkbox
-				class="tw-relative tw-cursor-pointer md:tw-text-right kv-checkbox"
-				:class="{ 'tw-pt-0 ' : checkboxRight }"
+				class="tw-relative tw-cursor-pointer md:tw-text-right kv-checkbox tw-pt-0 donate-repayments-label"
 				id="donate-repayments"
+				data-testid="donate-repayments"
 				v-if="!myDonateRepayments"
 				v-model="donateRepayments"
 				@change="toggleDonateRepayments"
 			>
 				<span
 					id="donate-repayments-tooltip"
-					class="tw-text-small tw-font-medium"
-					:class="{ 'tw-sr-only' : checkboxRight }"
+					data-testid="donate-repayments-tooltip"
+					class="tw-text-small tw-font-medium tw-sr-only"
 				>
 					Donate loan repayments instead?
 				</span>
@@ -40,21 +38,8 @@
 				helping us cover operating costs and reach even more borrowers worldwide.
 				Repayments from selected loans will not be added back to your account as Kiva credit.
 			</kv-lightbox>
-
-			<kv-tooltip v-if="!checkboxRight" controller="donate-repayments-tooltip" theme="mint">
-				<template #title>
-					Thanks for your support!
-				</template>
-				When you check this box, repayments go back to Kiva in the form of donations,
-				helping us cover operating costs and reach even more borrowers worldwide.
-				<br>
-				Repayments from selected loans will not be added back to your account as Kiva credit.
-			</kv-tooltip>
 		</div>
-		<div
-			v-if="checkboxRight"
-			class="tw-flex tw-mt-2"
-		>
+		<div class="tw-flex tw-mt-2">
 			<kv-text-link
 				@click="isLightboxVisible = true;"
 				class="tw-text-base tw-text-left md:tw-text-right tw-w-full"
@@ -72,7 +57,6 @@ import _forEach from 'lodash/forEach';
 import numeral from 'numeral';
 import initializeCheckout from '@/graphql/query/checkout/initializeCheckout.graphql';
 import updateLoanReservationDonateRepayments from '@/graphql/mutation/updateLoanReservationDonateRepayments.graphql';
-import KvTooltip from '@/components/Kv/KvTooltip';
 import KvLightbox from '~/@kiva/kv-components/vue/KvLightbox';
 import KvTextLink from '~/@kiva/kv-components/vue/KvTextLink';
 import KvCheckbox from '~/@kiva/kv-components/vue/KvCheckbox';
@@ -82,7 +66,6 @@ export default {
 		KvCheckbox,
 		KvLightbox,
 		KvTextLink,
-		KvTooltip,
 	},
 	inject: ['apollo'],
 	data() {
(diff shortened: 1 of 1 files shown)


src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -140,35 +140,3 @@ export default {
...
-		line-height: 1;
-		cursor: pointer;
-
-		@include breakpoint(medium) {
-			text-align: right;
-			padding: 0.05rem 0 0 0.5rem;
-		}
-	}
-
-	.donate-repayments-icon {
```
BoulderBrains, line 160: This icon was no longer in the file, so I was able to just remove the scss

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-	<div v-if="showToggle" class="donate-repayments-toggle">
+	<div v-if="showToggle">
 		<kv-checkbox
-			class="donate-repayments-label"
+			class="tw-relative tw-cursor-pointer tw-leading-4 tw-pt-1 tw-pl-1 md:tw-text-right"
```
ryan-ludwig, line 4: Unless removing the leading class causes a problem, please remove it. Ideally line-heights are handled by our base typography rules.

ryan-ludwig (approved):
Nice!

BoulderBrains:
I'm working on the replacement of the hover tip with a lightbox currently.

BoulderBrains:
@ryan-ludwig 
I re-requested your review because I made some significant changes that I wanted to get your eyes on.

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -1,29 +1,69 @@
...
-			When you check this box, repayments go back to Kiva in the form of donations,
-			helping us cover operating costs and reach even more borrowers worldwide.
-			<br>
-			Repayments from selected loans will not be added back to your account as Kiva credit.
-		</kv-tooltip>
+			</label>
+			<kv-checkbox
+				class="tw-relative tw-cursor-pointer md:tw-text-right kv-checkbox"
+				:class="{ 'tw-pt-0 ' : checkboxRight }"
+				id="donateRepayments"
```
ryan-ludwig, line 22: This id is a clever workaround to hook into your label element above! There's one catch, which is that currently `kv-checkbox` generates a [random id for itself](https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-components/vue/KvCheckbox.vue#L10), so this id doesn't get applied to the checkbox.

I think we can still make all this work by changing your `<label>` element above to a `<button>`.  Then I'd add a ` aria-hidden="true"` to hide that button from screenreaders so they don't hear "Donate loan repayments instead?" twice.

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -1,29 +1,69 @@
...
+				:visible="isLightboxVisible"
+				title="Donate your loan repayments"
+				@lightbox-closed="closeLightbox"
+			>
+				When you check this box, repayments go back to Kiva in the form of donations,
+				helping us cover operating costs and reach even more borrowers worldwide.
+				Repayments from selected loans will not be added back to your account as Kiva credit.
+			</kv-lightbox>
+
+			<kv-tooltip v-if="!checkboxRight" controller="donate-repayments-tooltip" theme="mint">
```
ryan-ludwig, line 46: Since `checkboxRight` is always true, I don't think this will ever be used?

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -49,6 +93,8 @@ export default {
 			myDonateRepayments: false,
 			totals: {},
 			loans: [],
+			isLightboxVisible: false,
+			checkboxRight: true,
```
ryan-ludwig, line 97: What does this boolean help you with?

src/components/Checkout/DonateRepaymentsToggle.vue
BoulderBrains: I left this in there so we had the ability to flip the checkbox to it's previous state by just flipping a boolean. But perhaps cleaning it up is a better option?

src/components/Checkout/DonateRepaymentsToggle.vue
BoulderBrains: isLightboxVisible - Allows me to open and close the newly included lightbox.
checkboxRight - is swapping the display of the label and input.

BoulderBrains: this checkboxRight boolean allows me to toggle between a checkbox with a label on the right and a label on the right. Going to refactor this now and just always set it to right.

src/components/Checkout/DonateRepaymentsToggle.vue
BoulderBrains: Will be removed in my next commit.

src/components/Checkout/DonateRepaymentsToggle.vue
BoulderBrains: Thanks @ryan-ludwig

BoulderBrains:
@ryan-ludwig @mcstover 
Alright, I've made the requested changes and added some data-testid's to the elements with ids. This is now ready for final review. Thanks

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -1,29 +1,52 @@
...
+			>
+				When you check this box, repayments go back to Kiva in the form of donations,
+				helping us cover operating costs and reach even more borrowers worldwide.
+				Repayments from selected loans will not be added back to your account as Kiva credit.
+			</kv-lightbox>
+		</div>
+		<div class="tw-flex tw-mt-2">
+			<kv-text-link
+				@click="isLightboxVisible = true;"
+				class="tw-text-base tw-text-left md:tw-text-right tw-w-full"
```
ryan-ludwig, line 45: Not sure if the `tw-text-base` will always take effect, but I haven't tested it. I think kv-text-link wants to be `tw-text-h4` generally.

src/components/Checkout/DonateRepaymentsToggle.vue
```diff
@@ -1,29 +1,52 @@
...
+			v-if="showToggle"
+			class="tw-mt-2 tw-flex donate-repayments-toggle"
 		>
-			<span
-				id="donate-repayments-tooltip"
-				class="tw-text-small tw-text-link tw-font-medium"
+			<button
+				class="tw-flex-1 tw-pt-0.5 md:tw-pt-0 tw-text-base
+					tw-font-medium tw-text-left md:tw-text-right tw-mr-2
+					tw-cursor-pointer"
```
ryan-ludwig, line 10: Does this button also need a `v-if="!myDonateRepayments"` like the checkbox?

ryan-ludwig (approved):
Couple minor questions, but looks good! If the ultimate plan is give kv-checkbox an option for flipping the label position, it'd be a good idea to add a TODO in this file and mention the TODO in whatever ticket there is for kv-checkbox work.

mcstover (approved):

PR #3571, opened 2022-01-20
Remove remaining Foundation typography classes

ryan-ludwig:
VUE-849

### DIFF
--- a/src/assets/scss/global/typography.scss
+++ b/src/assets/scss/global/typography.scss
@@ -1,9 +1,5 @@
 @import "settings";
 
-h1.huge {
-	@include huge-headline();
-}
-
 .small-text {
 	@include small-text();
 }
--- a/src/assets/scss/global/typography.scss
+++ b/src/assets/scss/global/typography.scss
@@ -12,10 +12,6 @@
 	@include large-text();
 }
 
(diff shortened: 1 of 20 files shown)


src/assets/scss/global/typography.scss
```diff
@@ -1,13 +1,5 @@
...
 
-h1.huge {
-	@include huge-headline();
-}
-
-.small-text {
-	@include small-text();
-}
-
 .featured-text {
```
ryan-ludwig, line 3: These remaining 3 classes are still in use on Homepage[Contentful] components. I'm not sure which are still in play. Probably best to wait for CORE-51 to remove.

mcstover: Oh interesting...

mcstover (approved):
Would like to hold for merge until after we deploy the promo bar hotfix

src/assets/scss/global/typography.scss
eddieferrer: @ryan-ludwig can you make a note of this in CORE-51 so we don't forget to remove these classes when that work is done

eddieferrer (approved):

PR #3573, opened 2022-01-20
fix: bump spacing between borrower image and name in LoanStory component

mcstover:
This has been bugging me for a while and affects all viewports.

Prod Broken:
image

Fixed:
image


BoulderBrains (approved):

ryan-ludwig (approved):
Curious why there are negative margins on `<borrower-image`>

PR #3575, opened 2022-01-20
feat: clickable imageCard theme story cards

eddieferrer:
ACK-206

Allows for a layout similar to the one in ACK-206 ticket. Cards with link can be put into card rows. they will be the height of the background image. Analytics is also supported. 

We'll be reusing the StoryCard 'link' field for mobile needs as well (will support a deep link for mobile)

See sample cards at bottom of:  `/lp/eddie-test-page`

### DIFF
--- a/src/components/Contentful/StoryCard.vue
+++ b/src/components/Contentful/StoryCard.vue
@@ -8,14 +8,21 @@
 			:class="themeClass"
 			v-if="hasBackgroundImage"
 		>
-			<kv-contentful-img
-				class="kv-contentful-img tw-h-full tw-w-full tw-object-cover"
-				:width="520"
-				fallback-format="jpg"
-				:contentful-src="backgroundImage.url"
-				:alt="backgroundImage.description"
-			/>
-			<p class="story-card__imageCard-title tw-text-h4">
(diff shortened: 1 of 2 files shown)


src/components/Contentful/StoryCard.vue
```diff
@@ -8,14 +8,21 @@
...
+			>
+				<kv-contentful-img
+					class="kv-contentful-img tw-h-full tw-w-full tw-object-cover"
+					:width="520"
+					fallback-format="jpg"
+					:contentful-src="backgroundImage.url"
+					:alt="backgroundImage.description"
+				/>
+			</component>
+			<p class="story-card__imageCard-title tw-text-h4" v-if="!cardLink">
```
mcstover, line 25: Why would we hide this due to having a link on the card? Seems like we'd still like to show this interface element if it's present...not a blocker but likely a point of confusion for future usage.

mcstover (approved):

src/components/Contentful/StoryCard.vue
eddieferrer: I can double check with stakeholders.

PR #3578, opened 2022-01-21
update /styleguide page content to point to Kiva Classic resources

ryan-ludwig:
image


src/pages/StyleGuide.vue
```diff
@@ -1,167 +1,68 @@
...
-					<router-view />
-				</div>
-			</div>
-		</div>
-	</div>
+						</h3>
+					</li>
+				</ul>
+			</section>
+		</build-page-wrapper>
```
mcstover, line 45: Would like to see a small "alert" section to note that some portions of UI are still using embedded Foundation layout and style rules which we are incrementally replacing.

eddieferrer (approved):

BoulderBrains (approved):

emuvente (approved):
This looks great!

src/router/routes.js
```diff
@@ -334,48 +334,6 @@ export default [
 	{
 		path: '/styleguide',
 		component: () => import('@/pages/StyleGuide'),
-		children: [
```
mcstover, line 337: @ryan-ludwig Could would create a fallback route for `styleguide/*` to redirect back to just `/styleguide`?

mcstover (approved):

PR #3579, opened 2022-01-21
fix: allow website link to be processed, fire sentry in the client only

mcstover:
Explanation here: https://kiva.atlassian.net/browse/VUE-923?focusedCommentId=323887

### DIFF
--- a/src/components/BorrowerProfile/BorrowerBusinessDetails.vue
+++ b/src/components/BorrowerProfile/BorrowerBusinessDetails.vue
@@ -154,7 +154,7 @@ export default {
 			}
 		},
 		verifyUrl(value) {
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/BorrowerBusinessDetails.vue
```diff
@@ -154,7 +154,7 @@ export default {
 			}
 		},
 		verifyUrl(value) {
-			if (this.filteredSocialLinks.length) {
+			if (value) {
```
mcstover, line 157: `value` will be `null` or a string

BoulderBrains (approved):
nice

emuvente (approved):

PR #3584, opened 2022-01-26
fix: core-411 only fetching fundraising loan data if loan is fundraising

BoulderBrains:
[Core-411](https://kiva.atlassian.net/browse/CORE-411)

Before calling fundraising related data (fundraisingPercent, fundraisingTimeLeft & unreservedAmount) I'm now looking at the loan.status to verify it is fundraising.


mcstover (approved):
I think this will do it!

PR #3585, opened 2022-01-26
fix: exclude private tags from the tag filter list

ryan-ludwig:
This keeps them out of the filter lightbox on /cc pages


mcstover (approved):
Nice, let's give it go!

PR #3586, opened 2022-01-27
feat: restrict Causes users from seeing MG and AD signups

eddieferrer:
ACK-219
ACK-220

I tested all instances of this. 

We're now doing this in quite a few places. A future improvement would be better to centralize all this business logic somewhere. Maybe straight from the API? my.userAccount.isEligibleForMonthlyGood or maybe a local resolver of some sort?

### DIFF
--- a/src/components/MonthlyGood/AlreadySubscribedNotice.vue
+++ b/src/components/MonthlyGood/AlreadySubscribedNotice.vue
@@ -1,7 +1,7 @@
 <template functional>
 	<div v-if="!props.onetime">
 		<h1 class="tw-text-center tw-text-h2 tw-mb-2">
-			You're already signed up for Monthly Good
+			You're already signed up for a subscription
 		</h1>
 		<p class="tw-text-center">
 			Changes to this contribution can be made in your
--- a/src/components/Settings/SubscriptionsSettingsCards.vue
+++ b/src/components/Settings/SubscriptionsSettingsCards.vue
@@ -10,7 +10,7 @@
 
 		<!-- Monthly Good Settings -->
 		<subscriptions-monthly-good
-			v-if="!isOnetime"
+			v-if="!isOnetime && !hasModernSub"
 			@cancel-subscription="cancelSubscription"
 			@unsaved-changes="setUnsavedChanges"
 			ref="subscriptionsMonthlyGoodComponent"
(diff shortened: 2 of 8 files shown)


src/components/MonthlyGood/AlreadySubscribedNotice.vue
```diff
@@ -1,7 +1,7 @@
 <template functional>
 	<div v-if="!props.onetime">
 		<h1 class="tw-text-center tw-text-h2 tw-mb-2">
-			You're already signed up for Monthly Good
+			You're already signed up for a subscription
```
eddieferrer, line 4: I changed the wording here to be able to reuse this messaging for users who have MG or Causes

src/components/Settings/SubscriptionsSettingsCards.vue
```diff
@@ -154,13 +155,13 @@ export default {
...
 			this.isLegacySubscriber = legacySubs.length > 0;
-			this.mySubscriptions = data?.mySubscriptions?.values ?? [];
-			const causesSubscriptions = this.mySubscriptions.filter(
+
+			const modernSubscriptions = data?.mySubscriptions?.values ?? [];
+			this.hasModernSub = modernSubscriptions.length !== 0;
+			const causesSubscriptions = modernSubscriptions.filter(
 				subscription => subscription.category.subscriptionType === 'CAUSES'
 			);
-			this.hasActiveCauseSubscription = !!causesSubscriptions.find(
```
eddieferrer, line 161: cleaned up this logic a little bit.

src/pages/LandingPages/MGCovid19/MGCovid19.vue
```diff
@@ -50,8 +50,6 @@ export default {
 	},
 	data() {
 		return {
-			isExperimentActive: true,
-			isMonthlyGoodSubscriber: false,
```
eddieferrer, line 54: This data is not used in this template. I also did not modify `MGCovidHero` which does have a MG signup. But I don't think we are going to be using that landing page again.

mcstover: Oh man I sure hope we don't!

mcstover (approved):
Thank so you much!

PR #3597, opened 2022-01-28
refactor: remove unused theme props

ryan-ludwig:
The latest header and footer do not theme via a theme object like before, since they use the new themable tailwind colors. This PR just removes those objects which aren't doing anything on prod right now. For example http://www.kiva.org/build does not have a green header.

If we'd like to make theming the header/footer an option again, a way to do it would be to wrap whatever needs to be themed with a `kv-theme-provider` component and then pass either a predefined theme from kv-tokens/kivaColors to it, or a custom one.

roughly
```
import {
  defaultTheme, darkTheme, darkGreenTheme, mintTheme,
} from '@kiva/kv-tokens/configs/kivaColors.cjs';

data() {
   darkGreenTheme
}

<kv-theme-provider :theme="darkGreenTheme"><the-header /></kv-theme-provider>
```

### DIFF
--- a/src/components/WwwFrame/WwwPage.vue
+++ b/src/components/WwwFrame/WwwPage.vue
@@ -3,7 +3,6 @@
 		<the-banner-area v-show="!isKivaAppReferral" />
 		<the-header v-show="!isKivaAppReferral"
 			:hide-search-in-header="hideSearchInHeader"
-			:theme="headerTheme"
 		/>
 		<slot name="secondary" v-if="!isKivaAppReferral"></slot>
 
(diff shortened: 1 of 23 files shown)


ryan-ludwig:
This PR is meant to be a conversation starter too. 

Is there a better way to document KvThemeProvider to make it more understandable? https://608b4cf87f686c00213841b1-oexdydovsp.chromatic.com/?path=/docs/kvthemeprovider--default

Do we want to bring back contentful theming support? Also, maybe having different header themes isn't great for a consistent user experience and is a good thing to sunset. etc.

eddieferrer (approved):
I'm all for this. I agree that the inconsistent header and footer is not a good user experience. Once this is merged in -- We should remove these options from Contentful. I can help with that if needed. 

Inevitably there will be a one off header/footer color take over for some crazy special event campaign, but we can deal with that when we get there.

mcstover (approved):

PR #3616, opened 2022-02-09
feat: core-405 MG personalized value prop

BoulderBrains:
[Core-405](https://kiva.atlassian.net/browse/CORE-405)

Getting this piece into place, will need a little polish once all the page is put together.

Screen Shot 2022-02-09 at 4 50 09 PM

### DIFF
--- a/src/pages/MonthlyGood/PersonalizedMonthlyGoodLandingPage.vue
+++ b/src/pages/MonthlyGood/PersonalizedMonthlyGoodLandingPage.vue
@@ -64,24 +64,15 @@
 			/>
 		</section>
 
-		<section class="row md:tw-flex">
-			<div class="tw-block md:tw-hidden">
-				<kv-contentful-img
-					:contentful-src="valueImage"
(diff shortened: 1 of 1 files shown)


src/pages/MonthlyGood/PersonalizedMonthlyGoodLandingPage.vue
```diff
@@ -1,66 +1,96 @@
...
+			</kv-hero>
+
+			<personalized-mg-lightbox
+				:show-lightbox="showLightbox"
+				@hide-lightbox="showLightbox = false"
+			/>
+		</section>
+
+		<section class="row md:tw-flex">
+			<div class="tw-block md:tw-hidden">
```
mcstover, line 68: Please use flex-box to change the flex order so you can only load 1 image into the page. We do this in dynamicHeroClassic using `tw-order-first` which you may be able toggle by screen size.

mcstover (approved):

PR #3630, opened 2022-02-15
feat: add unbounce capture to additional routes

eddieferrer:
ACK-223

I wasn't  sure about the best place to put the unbounce trigger to make sure it loads every time for the selected routes. So added it to App.vue created lifecycle hook.I had some trouble getting it to work in client-entry on initial page ssr load. 

See ticket for list of routes this should be added to.

### DIFF
--- a/src/plugins/unbounce-event-mixin.js
+++ b/src/plugins/unbounce-event-mixin.js
@@ -9,7 +9,11 @@ export default {
 
 		// Set cookie to disabled unbounce for future visits
 		if (utmEmailMedium) {
-			this.cookieStore.set('kvutm_medium_email', true);
+			this.cookieStore.set(
+				'kvutm_medium_email',
+				true,
+				{ expires: new Date(new Date().setFullYear(new Date().getFullYear() + 2)) }
+			);
 		}
 
 		const utmEmailMediumCookie = this.cookieStore.get('kvutm_medium_email');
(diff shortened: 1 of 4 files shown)


emuvente (approved):
Looks good!

mcstover (approved):

src/plugins/unbounce-event-mixin.js
```diff
@@ -0,0 +1,43 @@
...
+
+export default {
+	inject: ['apollo', 'cookieStore'],
+	created() {
+		const unbounceRoute = this.$route?.meta?.unbounce ?? false;
+		const utmEmailMedium = this.$route?.query?.utm_medium === 'email';
+
+		// Set cookie to disabled unbounce for future visits
+		if (utmEmailMedium) {
+			this.cookieStore.set('kvutm_medium_email', true);
```
mcstover, line 12: I'm not sure if our cookieStore is working properly with data setting atm but ideally this would be set for at least a year.

mcstover: I believe it will be a session cookie by default

src/plugins/unbounce-event-mixin.js
```diff
@@ -0,0 +1,47 @@
...
+	created() {
+		const unbounceRoute = this.$route?.meta?.unbounce ?? false;
+		const utmEmailMedium = this.$route?.query?.utm_medium === 'email';
+
+		// Set cookie to disabled unbounce for future visits
+		if (utmEmailMedium) {
+			this.cookieStore.set(
+				'kvutm_medium_email',
+				true,
+				{ expires: new Date(new Date().setFullYear(new Date().getFullYear() + 2)) }
```
eddieferrer, line 15: 2 years.

mcstover (approved):

PR #3631, opened 2022-02-15
feat: modifies cause hp experiment to include non mobile users as well.

eddieferrer:
ACK-249

Since we no longer need to check this on the client, we can do the experiment assignment in prefetch and route accordingly. Also added a check for the experiment on Contentful page. This will allow the tracking event to fire for users who are in the shown group when they get rerouted to lp/causes. 

ACK team wants this hotfixed asap to try to increase the rate at which users are funnelled into the experiment to gather statistical significance

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -38,6 +38,7 @@ import { preFetchAll } from '@/util/apolloPreFetch';
 import { processPageContent } from '@/util/contentfulUtils';
 import logFormatter from '@/util/logFormatter';
 import contentfulEntries from '@/graphql/query/contentfulEntries.graphql';
+import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 
 // Page frames
 const WwwPage = () => import('@/components/WwwFrame/WwwPage');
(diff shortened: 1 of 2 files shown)


src/pages/ContentfulPage.vue
```diff
@@ -240,5 +241,20 @@ export default {
 			}
 		}
 	},
+	created() {
+		// Mobile Causes homepage experiment (GD-205)
+		const causesHomeExp = this.apollo.readFragment({
```
mcstover, line 246: Def see the need for this but should we gate this to the expected routes?

src/pages/Homepage/Homepage.vue
```diff
@@ -39,16 +39,26 @@ export default {
...
 			return client.query({
 				query: homePageQuery
-			}).then(() => {
+			}).then(pageQueryResult => {
+				if (!pageQueryResult?.data?.hasEverLoggedIn) {
+					return client.query({
+						query: experimentAssignmentQuery,
+						variables: { id: 'home_mobile_causes' },
+					}).then(expResult => {
+						console.log('expResult', expResult);
```
mcstover, line 50: log

mcstover (approved):

emuvente (approved):
👍

PR #3643, opened 2022-02-18
feat: kvanalytics plugin was updated to work as an isolated plugin

roger-in-kiva:
kvanalytics plugin was updated to work as an isolated plugin.

*The idea is to use this as a seeder for our plugin package.


emuvente (approved):
👍 looks good!

mcstover (approved):
Should be fine. @roger-in-kiva Please verify in dev that the general pageview and queued event tracking is operating properly

PR #3650, opened 2022-02-24
fix: ensure kiva credit checkout errors bubble up and show a toast me…

mcstover:
…ssage

### DIFF
--- a/src/plugins/checkout-utils-mixin.js
+++ b/src/plugins/checkout-utils-mixin.js
@@ -114,36 +114,42 @@ export default {
 		/* Handle Various errors from GraphQL
 		 * @param {Object} errorResponse contains errors node with array of errors
 		 */
-		showCheckoutError(errorResponse, ignoreAuth = false) {
+		showCheckoutError(errorResponse = [], ignoreAuth = false) {
 			checkInjections(this, injections);
 
 			// const errors = _get(errorResponse, 'errors');
 			let errorMessages = '';
 			// When validation or checkout fails and errors object is returned along with the data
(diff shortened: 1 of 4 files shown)


mcstover:
With the expansion of `showCheckoutError` any previous error signatures should still be supported. I was able to verify this by seeing toast messages for `validatePreCheckout` errors returned which use the legacy signature. I did also check and saw that the Braintree Checkout errors are still working properly as they use their own error message handler.

mcstover:
FYI @rubytidwell

src/plugins/checkout-utils-mixin.js
```diff
@@ -114,36 +114,42 @@ export default {
...
 
 				/* eslint-disable max-len */
 				// update error messages for new checkout context (Original messages reference a different user flow)
-				if (error === 'ERROR_OWN_LOAN') {
+				if (errorType === 'ERROR_OWN_LOAN') {
 					// Original Message: As a Kiva borrower, you cannot support your own fundraising loan on Kiva.  Please go back to the basket to remove your loan.
 					errorMessage = 'As a Kiva borrower, you cannot support your own fundraising loan on Kiva. Please remove your loan before completing checkout.';
 				}
-				if (error === 'ERROR_OVER_DAILY_LIMIT') {
+				if (errorType === 'ERROR_OVER_DAILY_LIMIT') {
```
mcstover, line 138: I was able to verify existing signature works with this error and that the new signature works with the UPC error.

emuvente (approved):
Nice! good find/fix

eddieferrer (approved):

PR #3656, opened 2022-03-01
fix: update donation interface test ids

mcstover:


src/components/Checkout/DonationItem.vue
```diff
@@ -104,7 +104,7 @@
 			>
 				<button
 					class="donation-amount"
-					data-testid="basket-donation-edit-button"
+					data-testid="basket-donation-edit-button-combined"
```
mcstover, line 107: So long as you are testing on a viewport >= 734px wide this one will be visible/active. Below that number and the two above would be visible/active.

davidskiva (approved):

PR #3659, opened 2022-03-03
fix: ensure link tracking is active on router-links

mcstover:
Found several links in the LendMenu's that weren't triggering their click events. This seems to work. Also, I realize the `.native` modifier is a Vue2 construct and won't transfer to Vue3.

### DIFF
--- a/src/components/WwwFrame/LendMenu/LendListMenu.vue
+++ b/src/components/WwwFrame/LendMenu/LendListMenu.vue
@@ -3,12 +3,11 @@
 		<router-link
 			to="/monthlygood"
 			class="tw-inline-flex tw-gap-0.5 tw-py-2 tw-mb-2 tw-border-b tw-border-tertiary tw-font-medium"
-			@click="trackMgLinkClick"
+			@click.native="trackMgLinkClick"
 		>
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/LendMenu/LendListMenu.vue
```diff
@@ -51,12 +50,20 @@
 							</li>
 						</template>
 						<li class="tw-border-t tw-border-tertiary">
-							<router-link to="/lend" class="lend-link">
+							<router-link
+								to="/lend"
+								class="lend-link"
+								v-kv-track-event.native="['TopNav','click-Lend-All_Loans']"
```
mcstover, line 56: Who knew 🤷

emuvente: That works?! Interesting 🤔

emuvente (approved):

src/components/WwwFrame/LendMenu/LendListMenu.vue
mcstover: indeed it is lolz

PR #3667, opened 2022-03-07
Ml landing loan display

mcstover:
The updated logic and settings around which loan display is visible and how we move between them is to resolve [MRKT-493](https://kiva.atlassian.net/browse/MRKT-493).

The Footer theme update just needed to be done as it wasn't resolved when we removed themes ;(

Before:
Screen Shot 2022-03-07 at 10 05 43 AM

After:
Screen Shot 2022-03-07 at 10 05 16 AM

### DIFF
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -42,6 +42,7 @@
 							:initial-filters="initialFilters"
 							:excluded-tags="excludedTags"
 							:initial-sort-by="initialSortBy"
+							:active-loan-display="activeLoanDisplay"
 							:show-loan-display-toggle="showLoanDisplayToggle"
 							:total-count="totalCount"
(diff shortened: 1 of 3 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -944,7 +965,9 @@ export default {
 			// Query to update basket state
 			this.updateBasketState();
 			// TEMPORARY: Obstruct ability to click the "Checkout" button on the loan card to prevent redirect
-			this.$refs.loandisplayref.loadingLoans = true;
+			if (this.$refs.loandisplayref) {
+				this.$refs.loandisplayref.loadingLoans = true;
```
mcstover, line 969: Oh man! accessing refs that don't exist cause "phantom" errors that don't bubble up. Guarding these before use allowed me to convert the loan displays to v-if's instead of v-show's.

emuvente (approved):
👍

eddieferrer (approved):

PR #3669, opened 2022-03-08
feat: improved open graph/twitter meta properties for sharing borrowe…

eddieferrer:
…r profiles

ACK-252

After doing some research, it seems some platforms (looking at you Slack) will use either `twitter:` or `og:` or a combination of both, depending on what order they are rendered on the page. 

This PR does a couple of things: 
* Add vmid to `twitter:image` in App.vue. This allows this property to be replaced by the `twitter:image` from SummaryCard
* Add `twitter` equivalent meta tags to SummaryCard and LoanUse

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -100,6 +100,7 @@ export default {
 				},
 				{
 					name: 'twitter:image',
+					vmid: 'twitter:image',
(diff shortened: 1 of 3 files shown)


emuvente (approved):
good additions!

mcstover (approved):
👍

PR #3670, opened 2022-03-08
feat(lend-by-category): add experiment for selecting hero loan VUE-917

emuvente:


mcstover (approved):
Nice extraction of these methods for reuse and a cleaner interface in the hero component!

maxhutch:
We have logic to handle the experiment strategy pattern on the backend -- if you pass the experiment header through to the hero channel method (i.e. [this query](https://github.com/kiva/ui/pull/3670/files#diff-bbd431cea2350bf5d0bb7a1561c336650733c062bff36bd2f1e5d77763c82138R88)), it will treat the two groups as expected.  That much being said, doing that split here isn't wrong; it just might be redundant.

src/util/categoryUtils.js
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Get the current Hero Channel id synchronously.
+ *
+ * @param  {ApolloClient} client     the client to read the data from
+ * @return {Object}                  { loanChannelIds, loansQuery }
+ */
+export function getHeroChannelCached(client) {
+	try {
+		// get hero channel id from chache if available
```
maxhutch, line 132: What are they keys in this cache, and what's its lifetime?  Are there docs for UI caching anywhere that I could read?

roger-in-kiva (approved):

mcstover:
> 

@maxhutch Are you referring to the distribution portion of our experiment setup? We typically run the distribution algo on the client which is then passed in the header `ex. x-experiments: EXP-ML-Service-Bandit-LendByCategory;a` and then used in the `EXP-` tracking event as well.

src/util/categoryUtils.js
maxhutch: Gotcha, thanks!

emuvente:
> We have logic to handle the experiment strategy pattern on the backend -- if you pass the experiment header through to the hero channel method (i.e. [this query](https://github.com/kiva/ui/pull/3670/files#diff-bbd431cea2350bf5d0bb7a1561c336650733c062bff36bd2f1e5d77763c82138R88)), it will treat the two groups as expected. That much being said, doing that split here isn't wrong; it just might be redundant.

@maxhutch Thanks for taking a look! The split I'm doing in this PR is on whether the experiment is even running or not; I'm not checking which experiment group the user is assigned to. I'm doing that so that we don't even request the `ml { heroChannel { id } }` query unless the experiment is actually running, which will let us quickly fall back to the pre-experiment behavior in case there's a problem with the experiment.

The experiment header with the user's assigned version will be added to the request automatically because of [this line](https://github.com/kiva/ui/pull/3670/files#diff-663064d0327435a24d722fef1ca7d01e28fd4fa6d8a24579689db8f0c5aca5efR8).

maxhutch:
> > 
> 
> @maxhutch Are you referring to the distribution portion of our experiment setup? We typically run the distribution algo on the client which is then passed in the header `ex. x-experiments: EXP-ML-Service-Bandit-LendByCategory;a` and then used in the `EXP-` tracking event as well.

No; I was expecting the distribution portion to happen here.  It's the switch that I was expecting to be in the backend.

emuvente:
> The kiva-ml-service supports dynamically changing the experiment and control behavior using a config server that reads state from kiva-config.

Yes, I like that pattern too! To be clear, this PR gives control entirely to the ML service for the control group vs experiment group behavior, taking advantage of everything you mentioned. This PR is not taking a stance on what the control behavior should be, it's just preserving the page experience that existed before the experiment. That was part of the requirements for this ticket, because it was not clear if this code would be released into production before you were back! We wanted to have a way to turn things off from the UI side of things in case of problems with querying the ML service.

maxhutch:
> it's just preserving the page experience that existed before the experiment.

Ahh, got it.  The heroChannel query is also currently returning the same thing as the ui settings query, but this makes sense from forwards/backwards compatibility perspective!  Sorry for the confusion!

PR #3676, opened 2022-03-10
PBJ-263: Adding a flag to allow migrated monthly good subscriptions r…

JoeTravisKiva:
…ecords to be modified using monolith API and existing UI until phase 2 of the migration starts using the Subscription API directly.

I'm totally open to other options/suggestions, but we need to be able to continue to modify migrated monthly good records in the existing UI for now.

### DIFF
--- a/src/components/Settings/SubscriptionsSettingsCards.vue
+++ b/src/components/Settings/SubscriptionsSettingsCards.vue
@@ -10,7 +10,7 @@
 
 		<!-- Monthly Good Settings -->
 		<subscriptions-monthly-good
-			v-if="!isOnetime && (!hasModernSub || hasActiveMonthlyGoodSubscription)"
+			v-if="!isOnetime && !hasActiveCauseSubscription"
 			@cancel-subscription="cancelSubscription"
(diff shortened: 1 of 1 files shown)


src/components/Settings/SubscriptionsSettingsCards.vue
```diff
@@ -10,7 +10,7 @@
 
 		<!-- Monthly Good Settings -->
 		<subscriptions-monthly-good
-			v-if="!isOnetime && !hasModernSub"
+			v-if="!isOnetime && (!hasModernSub || hasActiveMonthlyGoodSubscription)"
```
mcstover, line 13: Another approach would be to replace `!hasModernSub` with `! hasActiveCauseSubscription` and that would allow them through too without the extra logic added below.

mcstover (approved):
Left an alternative approach but am fine with either direction.

src/components/Settings/SubscriptionsSettingsCards.vue
JoeTravisKiva: Your suggestion is quite a bit cleaner so I'll update for that. Thanks!

PR #3679, opened 2022-03-11
fix: add analytics event on mg category selection

mcstover:


src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -567,8 +572,10 @@ export default {
 	}
 }
 
-#carousel_exp >>> section > div:nth-child(2) {
-	display: none;
+@include breakpoint(xxlarge) {
```
mcstover, line 575: These are only hidden when the page is at full width and all loan cards are visible.

roger-in-kiva (approved):

PR #3682, opened 2022-03-16
feat: remove kiva us from lbc page experimental layout

eddieferrer:
ACK-271

This will be hotfixed out. Hoping to release all of main. 

Removes `kiva-u-s` from this  experiment. I kept the image asset and category specs for `kiva-u-s`  incase it needs to be added back in in the future.


mcstover (commented):
👍

PR #3684, opened 2022-03-17
Add bundles core 461

roger-in-kiva:
Loan bundles were added for the lend by category page


src/components/LoanCollections/KivaClassicLoanCarouselExp.vue
```diff
@@ -35,15 +42,15 @@
 </template>
 
 <script>
-import KivaClassicBasicLoanCardExp from '@/components/LoanCards/KivaClassicBasicLoanCardExp';
```
mcstover, line 38: Need to keep this import as it's still used in the component

roger-in-kiva: Nice catch, thanks!

mcstover (approved):
Looks good!

PR #3686, opened 2022-03-18
feat: embeded unbounce landing pages

eddieferrer:
Allows you to go to '/ub/:unbouncePath` which will load up the landing page from unbounce. 

On first load, the iframe will not load, until a user accepts the OneTrust cookies, once they are accepted, a global event is fired which loads the iframe with the groups and the users optout preference sent as query params. 

On the unbounce side, if the query params are present, bypass the normal onetrust script and just load the appropriate scripts depending on the query param. 

Test page: 
`/ub/unbounce-test-page-embed`

Test direct unbounce link:
https://go.kiva.org/unbounce-test-page-embed/?groups=C0001%3A1%2CC0003%3A1%2CC0002%3A1%2CC0004%3A1&optout=true

Note: New landing pages should be created on the unbounce domain go.kiva.org @ unbounce so we can possibly deprecate the us.kiva.org domain.


src/pages/LandingPages/Unbounce/IFrameEmbed.vue
```diff
@@ -0,0 +1,123 @@
...
+
+			/** This event is fired either when this component mounts
+			* or from scripts.js when the cookie policy is changed */
+			window.addEventListener('oneTrustAccepted', this.handleOneTrustAccepted);
+
+			window.addEventListener('resize', _throttle(() => {
+				this.resizeIframe();
+			}, 200));
+
+			oneTrustGlobalEvent();
```
eddieferrer, line 103: This needs to be fired here and in script.js because we cannot guarantee that the component will have loaded to listen to the global event when it is fired in script.js

src/server-entry.js
```diff
@@ -90,7 +92,8 @@ export default context => {
 		// render content for template
 		context.renderedConfig = renderGlobals({ __KV_CONFIG__: config });
 		context.renderedNoscript = noscriptTemplate(config);
-		context.renderedExternals = `<script>(${serialize(headScript)})(window.__KV_CONFIG__);</script>`;
+		// eslint-disable-next-line max-len
+		context.renderedExternals = `<script>(${serialize(headScript)})(window.__KV_CONFIG__, ${serialize(oneTrustEvent)});</script>`;
```
eddieferrer, line 96: Inject oneTrustEvent code into head script

mcstover (approved):
Looks good and thanks for working out a way to transport these user settings. Global code over in the UB looks like a nice improvement too!

src/server-entry.js
mcstover: The addition of this oneTrust transport script is a nice lightweight amount of overhead to ensure we can access that state in the embed.

PR #3688, opened 2022-03-21
fix: update model-value on mg kv-select and guard loan arrays

mcstover:


roger-in-kiva (approved):
Nice!

src/pages/MonthlyGood/LandingFormExperiment.vue
```diff
@@ -20,7 +20,7 @@
 			<kv-select
 				class="select-category"
 				id="select-category"
-				:value="selectedGroup"
+				:model-value="selectedGroup"
 				@change="updateSelected"
```
emuvente, line 24: Could also replace `@change` with `@update:modelValue="updateSelected"` like in https://github.com/kiva/ui/pull/3651/commits/a3c92d42b1d16d857fef89bf962208ec2e1e3585#diff-3817250708c23e4a9dc53d5c8e90962c702520b5245b3c88bdae89899abf3d5bR28

emuvente (approved):
👍  looks great!

PR #3689, opened 2022-03-21
feat: dynamic text for bundles was added

roger-in-kiva:


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -290,6 +298,14 @@ export default {
 			}
 			return title;
 		},
+		bundleText() {
+			let text = '';
```
mcstover, line 302: can we please add a default here: `Support these three loans with just one click`

roger-in-kiva: Ready!

mcstover (approved):

PR #3694, opened 2022-03-22
Feature/mars 2

christian14b:
I modified the regex pattern in order to support the use of 'dot' in the router and parse function. It worked well in regexr.com/659in. I do not know if there is another approach to do this.

### DIFF
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -123,12 +123,12 @@ module.exports = function liveLoanRouter(cache) {
 	});
 
 	// Filter URL Router
-	router.use('/f/:id([a-zA-Z0-9%,_-]{0,})/url/:offset(\\d{0,})', async (req, res) => {
+	router.use('/f/:id([a-z.A-Z.0-9.%,_-]{0,})/url/:offset(\\d{0,})', async (req, res) => {
 		await redirectToUrl('filter', cache, req, res);
 	});
 
 	// Filter IMG Router
-	router.use('/f/:id([a-zA-Z0-9%,_-]{0,})/img/:offset(\\d{0,})', async (req, res) => {
+	router.use('/f/:id([a-z.A-Z.0-9.%,_-]{0,})/img/:offset(\\d{0,})', async (req, res) => {
 		await serveImg('filter', cache, req, res);
 	});
 
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -123,12 +123,12 @@ module.exports = function liveLoanRouter(cache) {
 	});
 
 	// Filter URL Router
-	router.use('/f/:id([a-z.A-Z.0-9.%,_-]{0,})/url/:offset(\\d{0,})', async (req, res) => {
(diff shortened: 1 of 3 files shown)


server/util/live-loan/live-loan-fetch.js
```diff
@@ -165,15 +165,24 @@ async function fetchThemes() {
...
+}
+
 // Parse the filter string and return an array of filter arrays, e.g. [['sector', 'arts'], ['country', 'ke']]
 // Returns an empty array if the filter string does not contain any valid matches.
 const getFilterArrays = filterString => {
 	// This regex matches strings like 'sector_arts,country_ke' and captures each
 	// param name and value (for example 'sector', 'arts', 'country', 'ke' would be captured).
 	// See tests, examples, and a more detailed explanation at regexr.com/659in
-	const searchParamRegex = /(?:([a-z]+)_([a-z0-9 \\/-]+),?)+?/g;
+	const searchParamRegex = /(?:([a-z]+)_([a-z.0-9. \\/-]+),?)+?/g;
```
emuvente, line 182: The `.` only needs to be included once, and it should be escaped: `/(?:([a-z]+)_([a-z0-9 \.\\/-]+),?)+?/g`. I've updated regexr.com/659in with this change, and a test.

server/live-loan-router.js
```diff
@@ -123,12 +123,12 @@ module.exports = function liveLoanRouter(cache) {
 	});
 
 	// Filter URL Router
-	router.use('/f/:id([a-zA-Z0-9%,_-]{0,})/url/:offset(\\d{0,})', async (req, res) => {
+	router.use('/f/:id([a-z.A-Z.0-9.%,_-]{0,})/url/:offset(\\d{0,})', async (req, res) => {
```
emuvente, line 126: On this line and on line 131, the `.` only needs to be included once, and it should be escaped: `[a-zA-Z0-9\.%,_-]{0,}`

server/util/live-loan/live-loan-fetch.js
```diff
@@ -294,6 +304,11 @@ async function parseFilterStringLegacy(filterString) {
 					if (theme) {
 						addArrayFilterValue(name, theme.name);
 					}
+				} else if (name === 'tag') {
+					const tag = findFilterOption(tags, name, value);
+					if (tag) {
+						addArrayFilterValue(name, tag.name);
```
emuvente, line 310: The legacy search expects an array of tag IDs, using the filter name 'loanTags', so this line should be changed to `addArrayFilterValue('loanTags', tag.id);`. This will mean the tests need to be updated as well, for example: `await testFilterParsing('tag_u.s. black-owned businesses', { loanTags: [43] });`

emuvente (commented):
This is looking great!

server/util/live-loan/live-loan-fetch.js
christian14b: Hey @emuvente, I'm getting an ESlint error with `\.`, Should I use  `\\.` instead?

mcstover (approved):
Very cool to see this new feature coming along!

emuvente (approved):
Good work!

PR #3695, opened 2022-03-23
chore: update follow-redirects

mcstover:


mcstover:
Replaces https://github.com/kiva/ui/pull/3691 due to conflicts.

PR #3702, opened 2022-03-24
chore: add some data-testids for borrower page

SqueakAttack:
These are the ids we talked about for CORE-412, should be all of them in conjunction with Davids PR


mcstover (approved):

davidskiva (approved):
These look good.

PR #3708, opened 2022-03-28
fix: thanks page layout and upsell for new subscription service users

eddieferrer:
Now  that the subscriptions service is available in testing environments, we can clean this up to solve layout issues for modern subs users. 

ACK-270


mcstover:
I think we need to hold on this for a bit. Having that query in the preFetch will 500 this page in Test + QA.

eddieferrer:
@mcstover I think we can merge this in now that the subscriptions service is in the lower environments

PR #3710, opened 2022-03-30
fix: use data-testid instead of data-test for element attributes

mcstover:


src/components/Checkout/OrderTotals.vue
```diff
@@ -90,7 +90,7 @@
 			</kv-tooltip>
 		</div>
 
-		<div class="order-total" data-test="total-due">
+		<div class="order-total" data-test="total-due" data-testid="total-due">
```
davidskiva, line 93: Question.  why did you leave in the old data-test instead of just changing it.

mcstover: Just didn't want to break anything in that case...

mcstover: For that file in particular there are several of those as that was the original approach I took when refactoring the Checkout sections to Kiva Classic. I saw the versions without `id` and just made duplicates.

mcstover: I'd be happy to make a sweep and remove these too though.

As for autolend settings which has a bunch of these, I'm going to create a separate task for that work.

davidskiva: Okay, I'm fine with leaving as is.  @SqueakAttack you okay having as is?

davidskiva (approved):

src/components/Checkout/OrderTotals.vue
SqueakAttack: that's fine. if it might break something to change, maybe best to leave as is

mcstover: Ok I'm removing them from the OrderTotals file. This won't affect anything in the Ui repo. Please look out for any issues around the totals in your robot tests. Everything removed has a `data-testid` now.

SqueakAttack (approved):
👍

PR #3714, opened 2022-03-31
fix: use node 16 in ui tests github actions

mcstover:


mcstover:
Builds started failing due to Node 15 usage: https://github.com/kiva/ui/runs/5779819135?check_suite_focus=true

PR #3719, opened 2022-04-04
feat: deprecate causes homepage experiment and sign up pages

eddieferrer:
Deprecates causes experiment
Deprecate Causes sign up pages. We do not want users signing up for causes. Causes settings/management pages are still there to allow users to cancel or modify their causes subscriptions

ACK-275


emuvente (approved):
👍 🧹  It continues to be great to see experiment code cleaned up. Thank you!

mcstover (approved):

PR #3722, opened 2022-04-06
chore: add more data-testids for borrower profile page

davidskiva:
These should be all the data-testids need.

### DIFF
--- a/src/components/BorrowerProfile/BorrowerBusinessDetails.vue
+++ b/src/components/BorrowerProfile/BorrowerBusinessDetails.vue
@@ -1,6 +1,8 @@
 <template>
-	<section class="tw-rounded tw-border-tertiary tw-bg-primary tw-p-2.5 md:tw-p-3.5 lg:tw-p-4">
-		<h2 class="tw-mt-0 tw-mb-1">
+	<section class="tw-rounded tw-border-tertiary tw-bg-primary tw-p-2.5 md:tw-p-3.5 lg:tw-p-4"
+		data-test-id="bp-direct-business-info"
+	>
+		<h2 class="tw-mt-0 tw-mb-1" data-testid="bp-direct-business-name">
 			About {{ borrowerBusinessName }}
 		</h2>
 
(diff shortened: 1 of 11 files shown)


src/components/BorrowerProfile/DescriptionListItem.vue
```diff
@@ -1,11 +1,15 @@
 <template>
 	<div class="tw-flex tw-mb-1.5">
 		<dt class="tw-flex-1 tw-mb-0">
-			<button @click="$emit('show-definition')" class="tw-underline tw-font-book">
+			<button @click="$emit('show-definition')" class="tw-underline tw-font-book"
+				data-testid="bp-details-lightbox-btn"
```
davidskiva, line 5: These would need to be used in conjunction with the parent in either LoanDetails.vue, FieldPartnerDetails.vue and TrusteeDetails.vue.

src/components/BorrowerProfile/BorrowerBusinessDetails.vue
```diff
@@ -36,6 +39,7 @@
 					tw-inline-flex tw-items-center tw-justify-center
 					tw-mt-2 tw-mr-2
 					tw-h-6 tw-w-6"
+			:data-testid="`bp-direct-business-social-links-${index}`"
```
mcstover, line 42: If you wanted some additional specificity you could use `link.type` in place of `index` here to show "facebook" or "twitter" etc...

src/components/BorrowerProfile/DescriptionListItem.vue
mcstover: Correct and likely some "term" matching if you want to click any specific item...

mcstover: Oh I see you have some specific wrappers below too!

mcstover (approved):

src/components/BorrowerProfile/BorrowerBusinessDetails.vue
davidskiva: I did like the idea of using link.type, but after careful consideration I think using index is best.  It is just easier to iterate through a loop to see which, if any, social link exist.  We can either get the href attribute for the element or alt attribute for the child svg to determine which site it is.

PR #3728, opened 2022-04-08
fix: added some top level data-testids for elements with same data-te…

davidskiva:
…stid

I think I got them all.
Some of the elements that have same data-testid will need to have a top level data-testid, that I added, prepended.

### DIFF
--- a/src/components/Checkout/CheckoutReceipt.vue
+++ b/src/components/Checkout/CheckoutReceipt.vue
@@ -29,7 +29,7 @@
 				</div>
 				<ul class="checkout-receipt__item-list">
 					<!-- Loans -->
-					<li
+					<li data-testid="receipt-item-loan"
 						class="section"
 						v-for="loan in loans"
 						:key="loan.id"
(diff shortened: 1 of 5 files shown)


src/components/Checkout/CheckoutReceipt.vue
```diff
@@ -198,65 +208,68 @@
...
+						<span class="total__amount tw-text-h3" data-testid="amount-charged-payment-amount">
 							${{ receipt.totals.depositTotals.depositTotal }}
 						</span>
 					</li>
 				</ul>
 			</section>
 			<section class="section section--print hide-for-print">
-				<button class="print tw-text-link tw-flex tw-items-center tw-gap-1" @click="printReceipt">
+				<button data-testid="print-receipt"
+					class="print tw-text-link tw-flex tw-items-center tw-gap-1" @click="printReceipt">
```
mcstover, line 272: The lint warning here is b/c the `>` should be on the next line. Additionally, there is a 3 item limit for a single line of attributes along with a line-length limit which I'm sure you've hit.

Once we move to multiple lines, we usually also move to one attribute per line to clean it up like I've shown below.

```
<button
    data-testid="print-receipt"
    class="print tw-text-link tw-flex tw-items-center tw-gap-1"
    @click="printReceipt"
>
```

mcstover (approved):
Please clean up that one warning.

src/components/Checkout/CheckoutReceipt.vue
davidskiva: Okay, I made the change to fix the link warning.

PR #3732, opened 2022-04-12
fix: loan bundles initialization was updated

roger-in-kiva:


mcstover (approved):
👍  Perfect. Let's verify assignment is happening without the force url in DEV.

PR #3734, opened 2022-04-13
feat: add meta info to lend-by-category pages

christian14b:
Update specific meta titles and descriptions depending on the category visited.

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -69,8 +69,8 @@ export default {
 			pageLayout: 'control',
 			addBundlesExp: false,
 			meta: {
-				title: '',
-				description: ''
+				title: undefined,
+				description: undefined
 			}
 		};
 	},
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -43,6 +43,19 @@ const pageQuery = gql`
 `;
 
 export default {
+	metaInfo() {
+		this.retrieveMetaInfo();
+		return {
+			title: this.meta.title,
+			meta: [
+				{
+					vmid: 'description',
+					name: 'description',
+					content: this.meta.description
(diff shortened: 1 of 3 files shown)


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -55,6 +68,10 @@ export default {
 			addToBasketExpActive: false,
 			pageLayout: 'control',
 			addBundlesExp: false,
+			meta: {
+				title: '',
+				description: ''
```
emuvente, line 73: Setting the title and description as `undefined` instead of empty strings will allow the default title/description to be used when one isn't set (like the world-refugee-day category which doesn't have a description) https://vue-meta.nuxtjs.org/guide/special.html#use-child-property-conditionally

christian14b: Good to know, thank you!

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -43,6 +43,19 @@ const pageQuery = gql`
 `;
 
 export default {
+	metaInfo() {
+		this.retrieveMetaInfo();
+		return {
+			title: this.meta.title,
+			meta: [
```
eddieferrer, line 50: We should also add the meta tags for `twitter:description` and `og:description` with the same value as `description`

emuvente: That's a good point! Should we do the same for `twitter:title` and `og:title`, setting them to the same value as `title`?

eddieferrer: Yep, lets add titles as well

christian14b: Good catch @eddieferrer ! I'll add those meta tags

emuvente (approved):
👍  Looks great!

eddieferrer (approved):

PR #3735, opened 2022-04-13
feat: set canonical urls

christian14b:
Setting canonical URLs following these conditions:

- Any page that uses pagination should include the page url param in the canonical url.
- Any page that is significantly changed by a URL param should include that URL param in the canonical url (e.g. lending filters).
- /lend-beta/* pages, which should use /lend/* as their canonical url instead


emuvente (approved):
👍  Looks good! This will help to clarify to search engines which URLs actually matter on our site

PR #3737, opened 2022-04-14
feat: optional choice test settings were built

roger-in-kiva:


mcstover (approved):
Perfect core setup!

@egoroza You can see here the stages of fetching an experiment from Kiva Admin, Using the setting to "assign" a version of the experiment followed by using the version to set a flag (`isOptionalChoiceExperiment`) that can be used in components for altering the interface. Also present is the experiment level tracking we can use for analysis.

PR #3740, opened 2022-04-18
feat: monthly good optional choice experiment landing page

egoroza:
updates monthly good page to add variant landing page experience for monthly good optional choice experience

desktop:
Screen Shot 2022-04-19 at 10 59 15 AM

mobile:
Screen Shot 2022-04-19 at 10 59 28 AM
Screen Shot 2022-04-19 at 10 59 37 AM


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.102.0",
+	"version": "2.108.1",
```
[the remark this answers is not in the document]
mcstover: It should be ok and will get updated with later releases

src/pages/MonthlyGood/HowItWorksOptionalChoiceExp.vue
```diff
@@ -0,0 +1,79 @@
...
+	}
+};
+
+</script>
+
+<style lang="scss" scoped>
+@import 'settings';
+
+.how-works-section-wrapper {
+	.mg-image {
```
mcstover, line 71: Was there a reason you couldn't use TW classes for these? Widths and measures should use one of our standard "space" definitions: https://608b4cf87f686c00213841b1-ebjvofhmyr.chromatic.com/?path=/story/base-styling-primitives--primitives

ie. `tw-w-6` or `tw-w-7` for the width and then use `tw-mt-0 tw-pb-4 tw-mx-auto` for your margin classes.

src/pages/MonthlyGood/HowItWorksOptionalChoiceExp.vue
```diff
@@ -0,0 +1,79 @@
+<template>
```
mcstover, line 1: @egoroza @vijayabhandari Please make a follow up ticket to move this to Contentful if we go forward with these after the experiment.

Alternatively, Emily and I could work on a much smaller component wrapper to just load this content from Contentful before launch too.

src/pages/MonthlyGood/HowItWorksOptionalChoiceExp.vue
mcstover: Thanks, please see if you can use the classes instead.

If you absolutely need an off scale size you can just put the width inline using a `style=""` attribute.

src/pages/MonthlyGood/HowItWorksOptionalChoiceExp.vue
egoroza: would be good to have this loaded from contentful regardless of whether or not we go through with this experiment imo. it's hardcoded for the control/current experience as well

mcstover: Agree! But we should probably set out a separate task to do that work.

mcstover (commented):
Nice work!

src/pages/MonthlyGood/HowItWorksOptionalChoiceExp.vue
egoroza: created https://kiva.atlassian.net/browse/CORE-543

roger-in-kiva (commented):
Looking pretty good!

PR #3744, opened 2022-04-19
Feature/mars 89

christian14b:
This feature aim is to run an experiment about the influence in the order of the categories listed in the Lend Menu

### DIFF
--- a/src/components/WwwFrame/LendMenu/TheLendMenu.vue
+++ b/src/components/WwwFrame/LendMenu/TheLendMenu.vue
@@ -67,8 +67,15 @@ export default {
 		const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
 		if (enabled) {
 			trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
+			getExperimentSettingAsync(this.apollo, lendMenuExpKey)
+				.then(({ version }) => {
+					this.weighedCategoriesExp = version === 'shown';
+					this.apollo.query({
+						query: experimentAssignmentQuery,
+						variables: { id: lendMenuExpKey }
+					});
+				});
 		}
-		this.weighedCategoriesExp = enabled;
 	},
 	inject: ['apollo', 'cookieStore'],
 	data() {
@@ -98,16 +105,7 @@ export default {
 	},
 	apollo: {
 		query: pageQuery,
-		preFetch(config, client) {
-			return getExperimentSettingAsync(client, lendMenuExpKey)
-				.then(({ enabled }) => {
-					this.weighedCategoriesExp = enabled;
-					return client.query({
-						query: experimentAssignmentQuery,
-						variables: { id: 'lend_menu_category_sort' }
-					});
-				});
-		},
+		preFetch: true,
 		result({ data }) {
 			this.userId = _get(data, 'my.userAccount.id');
 		},
--- a/src/components/WwwFrame/LendMenu/TheLendMenu.vue
+++ b/src/components/WwwFrame/LendMenu/TheLendMenu.vue
@@ -63,22 +63,6 @@ export default {
 		LendListMenu,
 		LendMegaMenu,
 	},
-	created() {
-		if (!this.$isServer) {
-			const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
-			if (enabled) {
-				trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
-				getExperimentSettingAsync(this.apollo, lendMenuExpKey)
-					.then(({ version }) => {
-						this.weighedCategoriesExp = version === 'b';
-						this.apollo.query({
-							query: experimentAssignmentQuery,
-							variables: { id: lendMenuExpKey }
-						});
-					});
-			}
-		}
-	},
 	inject: ['apollo', 'cookieStore'],
 	data() {
 		return {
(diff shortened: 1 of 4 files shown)


src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -55,6 +63,20 @@ export default {
 		LendListMenu,
 		LendMegaMenu,
 	},
+	created() {
+		const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
```
mcstover, line 67: I think this should all be within an `!this.$isServer` check here so it only runs in the browser. Otherwise it will try to run during server render but likely not complete.

The other usage of `getExperimentSettingAsync` in the Featured Hero is in the `prefetch` function which is specifically built to resolve on the server and sets up the cached data for the created method in that context.

Being that the lend menu is hidden on page load, I don't think we need to add this to prefetch but keeping it in created should ensure it's ready when someone hovers over or activates the menu...

christian14b: Thank Matt, I'll add that flag to ensure is running in the browser.

I saw the other usage of `getExperimentSeetingsAsync`, I tried to put it in the `prefetch` method but it was being ignored

emuvente: @mcstover using preFetch doesn't work in this case because the TheLendMenu component is asynchronously loaded in TheHeader, so it isn't found by the getDeepComponents function of the preFetch process.

mcstover: Right on. As mentioned above, I don't think we should use prefetch for this. I just also think we should prevent it from initiating a promise during the server render that may or may not complete.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -55,6 +63,22 @@ export default {
...
 		LendMegaMenu,
 	},
+	created() {
+		if (!this.$isServer) {
+			const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
+			if (enabled) {
+				trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
+				getExperimentSettingAsync(this.apollo, lendMenuExpKey)
+					.then(({ version }) => {
+						this.weighedCategoriesExp = version === 'shown';
```
emuvente, line 73: Because `b` was used as the variant key in the SettingsManager entry, `b` should be checked for here instead of `shown`

src/components/WwwFrame/LendMenu/TheLendMenu.vue
mcstover: `TheLendMenu: () => import('@/components/WwwFrame/LendMenu/TheLendMenu'),` would def have an ill affect on prefetch but not really b/c we intentionally loaded it this way.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
christian14b: You're right, sorry ! I solved this.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -55,6 +63,22 @@ export default {
 		LendListMenu,
 		LendMegaMenu,
 	},
+	created() {
+		if (!this.$isServer) {
+			const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
+			if (enabled) {
+				trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
+				getExperimentSettingAsync(this.apollo, lendMenuExpKey)
```
emuvente, line 71: `getExperimentSettingAsync` doesn't include the version assignment. I know that using this didn't work:
```js
const lendMenuCategorySortEXP = this.apollo.readFragment({
	id: 'Experiment:lend_menu_category_sort',
}) || {};
```
I think that didn't work because the experiment assignment query wasn't called. Instead of using `readFragment`, I would recommend adding this to `lendMenuData.graphql`:
```gql
experiment(id: "Experiment:lend_menu_category_sort") @client {
	id
	version
}
```
Which will make the experiment version available in the `onLoad` function, allowing the experiment flag to be set before the categories are set:
```js
this.apollo.watchQuery({ query: publicLendMenuQuery }).subscribe({
	next: ({ data }) => {
		const version = _get(data, 'experiment.version');
		const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
		if (enabled) {
			trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
			this.weighedCategoriesExp = version === 'b';
		}
		this.categories = _get(data, 'lend.loanChannels.values');
		this.isChannelsLoading = false;
	}
});
```

emuvente (approved):
:+1: Looks good!

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -149,6 +156,12 @@ export default {
 			});
 			this.apollo.watchQuery({ query: publicLendMenuQuery }).subscribe({
 				next: ({ data }) => {
+					const version = _get(data, 'experiment.version');
+					const { enabled } = getExperimentSettingCached(this.apollo, lendMenuExpKey);
+					if (enabled) {
+						trackExperimentVersion(this.apollo, this.$kvTrackEvent, 'TopNav', lendMenuExpKey);
```
emuvente, line 162: For MARS-90 the tracking event action needs to be `EXP-MARS-89-Apr2022`. By default the `trackExperimentVersion` function uses `key` as the action, so I think what we should do is modify the `trackExperimentVersion` function to have an optional 5th argument called `action`. When `trackExperimentVersion` calls `trackEvent` it can pass in `action ?? key` for the action argument, so that it still falls back to using `key` if `action` is not defined.

christian14b: Done!

mcstover:
Nice improvements happening here!

emuvente (approved):
👍  thanks for the trackExperimentVersion improvements!

PR #3746, opened 2022-04-20
feat: copy was updated in the mg setup page

roger-in-kiva:


mcstover (approved):

roger-in-kiva:
Note - this should be merged after the MG migration when $5 notes is live.

PR #3747, opened 2022-04-20
feat: monthly good optional choice exp false door lightbox

egoroza:
- implements false door experience/lightbox for monthly good optional choice experiment
- also updates `max-width` for mg optional choice landing page to be wider per design request

mobile:
Screen Shot 2022-04-22 at 12 01 35 AM

desktop:
Screen Shot 2022-04-22 at 12 01 44 AM

### DIFF
--- a/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
+++ b/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
@@ -199,10 +199,12 @@
 
 		<!-- Monthly Good False Door Exp Optional Choice -->
 		<kv-lightbox
-			title="Thanks for your interest!"
 			:visible="showActiveLightbox"
 			@lightbox-closed="closeLightbox"
 		>
+			<h2 class="tw-text-h2 tw-text-center">
+				Thanks for your interest!
+			</h2>
 			<div class="tw-text-center tw-flex tw-flex-col tw-gap-3 tw-my-8 lg:tw-w-8/12 md:tw-w-full">
 				<!-- eslint-disable-next-line max-len -->
 				<p>Choosing a loan in Monthly Good is not yet available. If you want to choose a loan each month, you may like auto-deposit.</p>
@@ -246,12 +248,12 @@ import WwwPage from '@/components/WwwFrame/WwwPage';
 
 import KvHero from '@/components/Kv/KvHero';
 import KvContentfulImg from '@/components/Kv/KvContentfulImg';
-import KvLightbox from '@/components/Kv/KvLightbox';
 import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
 import KvFrequentlyAskedQuestions from '@/components/Kv/KvFrequentlyAskedQuestions';
 import KivaClassicLoanCarouselExp from '@/components/LoanCollections/KivaClassicLoanCarouselExp';
 import AutomaticallySupportNotice from '@/components/MonthlyGood/AutomaticallySupportNotice';
 import loanGroupCategoriesMixin from '@/plugins/loan-group-categories';
+import KvLightbox from '~/@kiva/kv-components/vue/KvLightbox';
 import KvButton from '~/@kiva/kv-components/vue/KvButton';
 
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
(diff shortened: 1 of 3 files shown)


src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -195,6 +196,41 @@
...
+		<!-- Monthly Good False Door Exp Optional Choice -->
+		<kv-lightbox
+			title="Thanks for your interest!"
+			:visible="showActiveLightbox"
+			@lightbox-closed="closeLightbox"
+		>
+			<div class="tw-text-center tw-flex tw-flex-col tw-gap-3 tw-my-8 lg:tw-w-8/12 md:tw-w-full">
+				<!-- eslint-disable-next-line max-len -->
+				<p>Choosing a loan in Monthly Good is not yet available. If you want to choose a loan each month, you may like auto-deposit.</p>
+				<div class="tw-flex tw-flex-col tw-gap-2 lg:tw-w-10/12 md:tw-w-full tw-self-center">
```
mcstover, line 209: These buttons look really wide on desktop... maybe we need a max width of  300px or so added as a style attribute...

egoroza: oops - grabbed the wrong screencap but yeah it's still a little wide. can set that 👍

roger-in-kiva:
Content for mobile is aligned at the middle of the screen in figma, having a gap on the top and rounded corners.

roger-in-kiva:
Maybe decreasing the font size for the title, we could have a better view for ipad...

roger-in-kiva:
We need to update this copy, Susan is dummy data
Screen Shot 2022-04-20 at 13 36 52

egoroza:
> We need to update this copy, Susan is dummy data Screen Shot 2022-04-20 at 13 36 52

This supposed to be part of the image preview and is not actually functional according to product/design, i.e. just an image example and not implemented in the code as of now

egoroza:
> Maybe decreasing the font size for the title, we could have a better view for ipad...

I'm not sure if it makes sense to have a custom title for this lightbox since it would be good to have parity between the other `kv-lightbox` components across the site but happy to change that if others feel strongly about this

mcstover:
We should use the standard lightbox title and sizing for this. No need to alter type sizes per viewport as those changes are already built into the lightbox component.

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -210,11 +246,13 @@ import WwwPage from '@/components/WwwFrame/WwwPage';
 
 import KvHero from '@/components/Kv/KvHero';
 import KvContentfulImg from '@/components/Kv/KvContentfulImg';
+import KvLightbox from '@/components/Kv/KvLightbox';
```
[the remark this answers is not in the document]
mcstover: @egoroza please use the new version of our lightbox.

If you need to style the header differently than the default you can pass in a named slot with your special tailwind classes.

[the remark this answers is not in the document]
mcstover: There's an example of a custom header within our new lightbox here: https://github.com/kiva/ui/blob/main/src/components/BorrowerProfile/LendersAndTeams.vue

One notable change you'll see is that the mobile version will be more like an overlay card with rounded corners at the top.

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -195,6 +196,43 @@
 				:questions="frequentlyAskedQuestions"
 			/>
 		</div>
+
+		<!-- Monthly Good False Door Exp Optional Choice -->
+		<kv-lightbox
+			:visible="showActiveLightbox"
+			@lightbox-closed="closeLightbox"
+		>
+			<h2 class="tw-text-h2 tw-text-center">
```
egoroza, line 205: design wants this centered and not in line with the close icon, looks like putting this in `title` or `<template #header>` as a custom header keeps this in the same row as the close icon

mcstover (approved):

PR #3748, opened 2022-04-20
fix: type name should not be part of id argument MARS-89

emuvente:
The `experiment` resolver `id` argument is not meant to include the `Experiment` typename. This caused the version to always resolve to `null` because the experiment setting could not be found.


christian14b (approved):
Good catch! Thanks

PR #3750, opened 2022-04-20
perf: dynamically load loan card components VUE-486

emuvente:
Helps with performance by only loading the loan card component that is going to be displayed instead of loading every possible loan card component.


mcstover (approved):
Oh what a collection, almost time to prune some of them out!

eddieferrer (approved):

PR #3752, opened 2022-04-24
fix: prevent unreservedAmount from being less than zero

mcstover:
image

Based on this code, I think this will also resolve the "add to basket button" still being shown which is part of how extra reservations are being made.

### DIFF
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -151,7 +151,12 @@ function unreservedAmount(loan) {
 
 	// Calculate unfunded and unreserved amounts
 	const unfunded = loanAmount - fundedAmount;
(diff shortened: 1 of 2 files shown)


vijayabhandari:
Thanks @mcstover .  Looks like if it's negative number, the UI will show 0 now? But in the backend the reservation and charged amount will still be the same right? We are just changing the presentation of it?

roger-in-kiva (approved):

vijayabhandari (approved):

mcstover:
@vijayabhandari This will now show 0 remaining as the loan is fully reserved AND prevent the Lend Now button from still showing allowing additional reservations.

emuvente (approved):

egoroza (approved):

PR #3755, opened 2022-04-26
feat: dynamic lend amount button

egoroza:
A new lend button to be implemented across marketplace when a loan has <$25 left.

Screen Shot 2022-04-26 at 12 13 21 PM
Screen Shot 2022-04-26 at 12 13 39 PM

### DIFF
--- a/src/components/LoanCards/Buttons/LendAmountButton.vue
+++ b/src/components/LoanCards/Buttons/LendAmountButton.vue
@@ -1,0 +1,44 @@
+<template>
+	<lend-button
+		price="amountLeft"
+		:loan-id="loanId"
+		@add-to-basket="$emit('add-to-basket', $event)"
+	>
+		Lend ${{ amountLeft }} <span v-if="showNow">now</span>
+	</lend-button>
+</template>
+
+<script>
+import LendButton from './LendButton';
+
+export default {
+	components: {
+		LendButton,
+	},
+	props: {
+		loanId: {
+			type: Number,
+			default: null
+		},
+		loan: {
+			type: Object,
+			default: () => {}
+		},
+		showNow: {
+			type: Boolean,
+			default: false
+		}
+	},
+	computed: {
+		amountLeft() {
+			const { loanAmount, loanFundraisingInfo } = this.loan;
+			const { isExpiringSoon, fundedAmount, reservedAmount } = loanFundraisingInfo;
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/Buttons/LendAmountButton.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+	<lend-button
+		price="amountLeft"
+		:loan-id="loanId"
+		@add-to-basket="$emit('add-to-basket', $event)"
+	>
+		Lend ${{ amountLeft }} <span v-if="showNow">now</span>
```
mcstover, line 7: Please move the "space" before the `<span>` to be inside the span before " now".

src/components/LoanCards/Buttons/LendAmountButton.vue
```diff
@@ -0,0 +1,44 @@
...
+			type: Object,
+			default: () => {}
+		},
+		showNow: {
+			type: Boolean,
+			default: false
+		}
+	},
+	computed: {
+		amountLeft() {
```
mcstover, line 33: This may end up being better set as a prop but it's fine to keep here for now. I suspect the the parent component  will need to calculate this to understand if we want to show this button or the lend increment button with the drop-down.

egoroza: ah yes... OK, i'll add a little `TODO` comment here to swap this for a prop if necessary in one of the later tix

roger-in-kiva (approved):
Looks good!

src/components/LoanCards/Buttons/LendAmountButton.vue
mcstover: Thank yoU!

mcstover (approved):

src/components/LoanCards/Buttons/LendAmountButton.vue
egoroza: thinking about this now, wondering if it makes sense to turn this into a mixin if we have several cards calculating this

egoroza: or global js file

mcstover: We have a loanUtils file that could house a method to make this calculation.

Additionally, the LoanCardController already knows about this and passes it down to loan cards so you may already have it available in the loan cards you're working on.

PR #3757, opened 2022-04-26
fix: category control component overwriting meta title

christian14b:
This fix prevents `Loan Channel Category` meta title from being overwritten by  `Loan Channel Category Control `child component.


mcstover (approved):
I think this is fine even though "un-mapped" categories will get the default title and description from the parent page.

emuvente (approved):

PR #3759, opened 2022-04-27
feat: update LBC to accommodate less than 25 lend button + add tracking

egoroza:
this PR updates Lend by Category loan cards to use the `LendAmountButton` when there is <$25 less to be funded on a single loan. this PR also:
* updates tracking for `LendAmountButton`
* removes `amountLeft` computed prop since it now is coming from the parent component(s)
* updates `amountLeft` computed prop on `LoanCardController` such that a negative value is never returned, fixing progress meters that display a negative amount

Screen Shot 2022-04-27 at 12 31 42 PM
Screen Shot 2022-04-27 at 12 33 17 PM
Screen Shot 2022-04-27 at 12 35 10 PM
Screen Shot 2022-04-27 at 12 35 57 PM

### DIFF
--- a/src/components/LoanCards/Buttons/ActionButton.vue
+++ b/src/components/LoanCards/Buttons/ActionButton.vue
@@ -6,6 +6,8 @@
 		:loan="loan"
 		:minimal-checkout-button="minimalCheckoutButton"
 		@add-to-basket="handleAddToBasketEvent"
+		:show-now="showNow"
+		:amount-left="amountLeft"
 	/>
 </template>
 
@@ -71,6 +73,14 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		showNow: {
+			type: Boolean,
+			default: null
+		},
+		amountLeft: {
+			type: Number,
+			default: null,
+		}
 	},
 	data() {
 		return {};
--- a/src/components/LoanCards/Buttons/ActionButton.vue
+++ b/src/components/LoanCards/Buttons/ActionButton.vue
@@ -75,11 +75,11 @@ export default {
 		},
 		showNow: {
 			type: Boolean,
-			default: null
+			default: false
 		},
 		amountLeft: {
 			type: Number,
-			default: null,
+			default: 0,
 		}
 	},
 	data() {
--- a/src/components/LoanCards/Buttons/LendAmountButton.vue
+++ b/src/components/LoanCards/Buttons/LendAmountButton.vue
@@ -2,7 +2,7 @@
 	<lend-button
 		price="amountLeft"
 		:loan-id="loanId"
-		:is-partial-share="true"
+		v-kv-track-event="['Lending', 'Add to basket (Partial Share)', 'lend-button-click', loanId, amountLeft]"
 		@add-to-basket="$emit('add-to-basket', $event)"
 	>
 		Lend ${{ amountLeft }}<span v-if="showNow"> now</span>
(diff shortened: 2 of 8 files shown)


src/components/LoanCards/Buttons/ActionButton.vue
```diff
@@ -71,6 +73,14 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		showNow: {
+			type: Boolean,
+			default: null
```
mcstover, line 78: default value should `false` on these seeing as they are Bools

src/components/LoanCards/Buttons/ActionButton.vue
```diff
@@ -71,6 +73,14 @@ export default {
...
+		},
+		amountLeft: {
+			type: Number,
+			default: null,
```
mcstover, line 82: This should prob have a default of `0` to match the type.

roger-in-kiva (approved):

src/components/LoanCards/Buttons/LendButton.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<kv-button
-		v-kv-track-event="['Lending', 'Add to basket', 'lend-button-click', loanId, loanId]"
+		v-kv-track-event="['Lending', eventInfo, 'lend-button-click', loanId, loanId]"
```
mcstover, line 3: I don't think we should be changing this event. As mentioned in the slack thread we should have a new and separate even in the LendAmountButton component.

Changes to this event will cascade through an untold number of automated metrics dashboards that look at overall add to basket activity.

mcstover (approved):

src/components/LoanCards/Buttons/LendAmountButton.vue
```diff
@@ -2,6 +2,7 @@
 	<lend-button
 		price="amountLeft"
 		:loan-id="loanId"
+		:is-partial-share="true"
```
mcstover, line 5: I suggest adding a new event here like so which also captures the smaller remaining amount value in the last param.
`v-kv-track-event="['Lending', 'Add to basket (Partial Share)', 'lend-button-click', loanId, amountLeft]"`

egoroza: ooh gotcha, yeah wasn't sure how `v-kv-track-event` works exactly and thought it was only tied to `kv-button` components for some reason. makes sense! 👍

mcstover: We should test to verify.

Alternatively, you could break out a new method, `addToBasket` which would replace `$emit('add-to-basket', $event)` in the listener and contain both the new tracking event and the original emit statement.

```
@add-to-basket="addToBasket($event)"
```

```
addToBasket(event) {
    this.$kvTrackEvent('Lending', 'Add to basket (Partial Share)', 'lend-button-click', this.loanId, this.amountLeft);
    $emit('add-to-basket', event);
}
```

src/components/LoanCards/Buttons/LendAmountButton.vue
```diff
@@ -2,7 +2,7 @@
 	<lend-button
 		price="amountLeft"
```
mcstover, line 3: @egoroza this needs to be `:price="amountLeft"` to ensure the dynamic value of `amountLeft` is passed in.

mcstover: fixed with: https://github.com/kiva/ui/pull/3764

PR #3760, opened 2022-04-27
Feature/mars 95

christian14b:
This feature aims to create a new `Social Share` component (Facebook, Twitter, Email, and Copy Link) on the Borrower Profile page. Also, it sets an a/b test to show this component.


src/components/BorrowerProfile/BorrowerSocialShare.vue
```diff
@@ -0,0 +1,293 @@
...
+			}
+		}
+	},
+	mounted() {
+		this.handleFacebookResponse();
+	},
+};
+</script>
+
+<style lang="scss" scoped>
```
mcstover, line 167: Did most of this come from our existing Social Share component? A little bummed to have so much here that isn't in tailwinds but suppose we can roll with it for experimentation sake.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -153,6 +153,12 @@
 							{{ ctaButtonText }}
 						</kv-ui-button>
 					</span>
+					<borrower-social-share
+						v-show="socialShareExp"
```
mcstover, line 157: Can we use `v-if` instead of `show` here?

mcstover: Same with the other include below

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -245,6 +270,24 @@ export default {
 		if (expCookieSignifier === 'b') {
 			this.$kvTrackEvent('Borrower Profile', 'EXP-GROW-655-Aug2021', expCookieSignifier);
 		}
+		// EXP-MARS-95-May2022
+		this.apollo.watchQuery({ query: publicSocialShareQuery }).subscribe({
+			next: ({ data }) => {
+				const version = _get(data, 'experiment.version');
```
mcstover, line 276: Can we convert this to use optional chaining and remove the `_get` import?

mcstover:
@christian14b @emuvente I really like that this is not including any new 3rd party scripts!!!

src/components/BorrowerProfile/LendCta.vue
christian14b: Yeah! I forgot about the conditional display with `v-show`

src/pages/BorrowerProfile/BorrowerProfile.vue
christian14b: Right away

src/components/BorrowerProfile/BorrowerSocialShare.vue
christian14b: Yeah, I would love to work on that. I think we can do it if everything goes well with the experiment. What do you think @emuvente ?

src/components/BorrowerProfile/BorrowerSocialShare.vue
```diff
@@ -0,0 +1,293 @@
...
+@import 'settings';
+@import "foundation";
+
+$color-facebook: #3b5998;
+$color-twitter: #08a0e9;
+$color-email: #AB4147;
+$color-clipboard: #DEB13C;
+
+// layout of blocks
+.share {
```
emuvente, line 177: Our Foundation settings and Tailwind settings use different breakpoints, so I think this block of rules should be applied using Tailwind classes instead, as the rest of the page is using the Tailwind breakpoints.

src/components/BorrowerProfile/BorrowerSocialShare.vue
```diff
@@ -0,0 +1,293 @@
...
+			shareLink: `https://${this.$appConfig.host}${this.$route.path}`
+		};
+	},
+	computed: {
+		facebookShareUrl() {
+			return this.getFullUrl('https://www.facebook.com/dialog/share', {
+				app_id: this.$appConfig.fbApplicationId,
+				display: 'page',
+				href: `${this.shareLink}?utm_source=facebook.com&utm_medium=social&utm_campaign=social_share_checkout`,
+				redirect_uri: `${this.shareLink}?kiva_transaction_id=${this.$route.query.kiva_transaction_id}`,
```
emuvente, line 93: There isn't a transaction id on this page, so let's use the `href` value for `redirect_uri` as well.

src/components/BorrowerProfile/BorrowerSocialShare.vue
```diff
@@ -0,0 +1,293 @@
...
+	},
+	data() {
+		return {
+			copyStatus: {
+				class: '',
+				disabled: false,
+				text: ''
+			},
+			shareMessage: '',
+			shareLink: `https://${this.$appConfig.host}${this.$route.path}`
```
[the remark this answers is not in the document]
christian14b: You're right, thanks!

src/components/BorrowerProfile/BorrowerSocialShare.vue
christian14b: Forgot to ask about that, sorry!

src/components/BorrowerProfile/BorrowerSocialShare.vue
christian14b: Hey @emuvente I'm trying to translate the css to tw as `tw-my-[21px] tw-px-[15px] tw-max-w-[600px] md:tw-bg-primary md:tw-p-4 md:tw-rounded`, but it seems that arbitrary values don't work if I don't set `mode: 'jit'` in tailwind.config.js.

Should I modify the config file or change the breakpoint in css to be the same as tailwind's?

christian14b: Don't mind me. It works fine without it! Cache I guess

emuvente: The max-width can be set in an inline-style on the element `style="max-width:600px;"`, and for the margin and padding I would recommend rounding to the nearest built-in classes `tw-my-2.5 tw-px-2`.

mcstover: Yes, the dev cache will get you sometimes with these.

We typically don't like to use this dynamic value approach. Rather when we have to break out of our preset size classes, we would just add them to a `style=""` attribute instead.

I don't think we should be generating classes for 21 and 15 which are really odd values against our sizing scale which is based on multiple of 4px. The design should be adjusted to use our sizing scale so you can use our preset classes.

christian14b: I see... great to know that! I'll change it right away. Thank you both!

src/graphql/query/borrowerSocialShareData.graphql
```diff
@@ -0,0 +1,14 @@
+query borrowerSocialShareData {
+	general {
+		socialShareExperiment: uiExperimentSetting(
+			key: "borrower_social_share"
+		) {
+			key
+			value
+		}
+	}
+	experiment(id: "borrower_social_share") @client {
```
emuvente, line 10: This can't be queried for until after the uiExperimentSetting query completes, so this will need to be done in 2 separate queries.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -245,6 +269,24 @@ export default {
 		if (expCookieSignifier === 'b') {
 			this.$kvTrackEvent('Borrower Profile', 'EXP-GROW-655-Aug2021', expCookieSignifier);
 		}
+		// EXP-MARS-95-May2022
+		this.apollo.watchQuery({ query: publicSocialShareQuery }).subscribe({
+			next: ({ data }) => {
+				return getExperimentSettingAsync(this.apollo, socialShareExpKey)
```
emuvente, line 275: The publicSocialShareQuery already fetches the experiment setting, so this call is redundant. Instead, this should be replaced with a call to get the experiment version:
```js
import experimentAssignment from '@/src/graphql/query/experimentAssignment.graphql';
...
return this.apollo.query({
	query: experimentAssignment,
	variables: { id: socialShareExpKey },
}).then(versionResult => {
	const version = versionResult?.data?.experiment?.version;
	if (version) {
		trackExperimentVersion(
			this.apollo,
			this.$kvTrackEvent,
			'Borrower Profile',
			socialShareExpKey,
			'EXP-MARS-95-May2022'
		);
		this.showSocialShareBlock = version === 'b';
	}
});
```

christian14b: Thanks @emuvente, I'll keep this in mind

PR #3765, opened 2022-04-27
fix: use a 25 increment in the dropdown for loans with 25 or more rem…

mcstover:
…aining

Category Page:
image

Lend By Category page:
image

Borrower Profile:
image

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -484,6 +484,12 @@ export default {
 				this.cycleStatsSlot();
 			}
 		},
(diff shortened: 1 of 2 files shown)


roger-in-kiva (approved):

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -484,6 +484,12 @@ export default {
 				this.cycleStatsSlot();
 			}
 		},
+		unreservedAmount(newValue, previousValue) {
```
mcstover, line 487: This is just for backwards compatibility a little later down the road. Generally, the drop-down will be hidden for loans with less than $25 remaining.

PR #3767, opened 2022-04-27
Stranded loans checkout core 554

roger-in-kiva:
LoanPrice component was updated to hide the dropdown element if the price is less than 25.

### DIFF
--- a/src/components/Checkout/LoanPrice.vue
+++ b/src/components/Checkout/LoanPrice.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="loan-price-wrapper tw-flex">
-		<div class="loan-price-select tw-flex-grow">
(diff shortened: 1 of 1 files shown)


src/components/Checkout/LoanPrice.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 	<div class="loan-price-wrapper tw-flex">
-		<div class="loan-price-select tw-flex-grow">
+		<div v-if="price > 24" class="loan-price-select tw-flex-grow">
```
mcstover, line 3: That will work. Just need to add a `v-else` below showing the price...something like...
```
<p v-else class="tw-text-h3 tw-mb-0">
    {{ price }}
</p>

mcstover (approved):

PR #3769, opened 2022-04-28
feat: stranded loans lend button on borrower profile

egoroza:
Implement `LendAmountButton` on Borrower Profile when there is <$25 left to fund on a loan. 

Screen Shot 2022-04-28 at 11 35 07 AM

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -120,7 +120,7 @@
 									:show-now="true"
 									:amount-left="unreservedAmount"
 									@add-to-basket="addToBasket"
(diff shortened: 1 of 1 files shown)


mcstover (approved):
Cool let's check it out in dev

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -309,6 +323,14 @@ export default {
 	},
 	data() {
 		return {
+			loan: {
```
mcstover, line 326: @egoroza The new LendAmountButton no longer has a dependency on the full loan being passed as prop so it can be removed from that file. Could you please work on that clean up and then go through and remove this extra loan setup that was added to this component?

PR #3775, opened 2022-05-02
feat: donate seo page updates

edwinf-barreram:
- Added meta info for donate SEO pages

### DIFF
--- a/src/pages/Donate/DonateSupportUs.vue
+++ b/src/pages/Donate/DonateSupportUs.vue
@@ -45,8 +45,15 @@ const pageQuery = gql`query donateContent {
 
 export default {
 	metaInfo() {
-		return {
-			title: `${this.pageTitle}`,
+		return 	{
+			title: 'Donate to Kiva and create real change',
+			meta: [
+				{
+					vmid: 'description',
+					name: 'description',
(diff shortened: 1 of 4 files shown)


src/pages/Donate/DonateSupportUs.vue
```diff
@@ -45,8 +45,16 @@ const pageQuery = gql`query donateContent {
 
 export default {
 	metaInfo() {
-		return {
-			title: `${this.pageTitle}`,
+		return 	{
+			title: 'Donate to Kiva and create real change',
```
mcstover, line 49: Why not just set this in Contentful seeing as it's already wired up there?

edwinf-barreram: @mcstover In my understanding this is the way it has been done in several tickets, I'm just following some previous solutions.

mcstover: I think most of the other tickets do not have a corresponding Contentful page like these 2 do.

It would be more future proof to have this title set properly in Contentful but I'm ok in the short term to move it here.

@emuvente your call.

emuvente (approved):
Looks good!

christian14b (approved):

PR #3779, opened 2022-05-03
Experiment refactor

emuvente:
There are 3 major things happening in this PR:

1. Getting the current experiment version has been simplified to 1 request, hopefully making it much easier to use experiments in components. Example new usage:
```js
apollo: {
  preFetch: true,
  query: gql`query myCoolExperiment {
    experiment(id: "my_cool_experiment") @client {
      id
      version
    }
  }`,
  result({ data }) {
    this.myCoolExperimentVersion = data?.experiment?.version ?? 'a';
  }
}
```
2. Experiments are exclusively enabled by being added to the `ui.active_experiments` string in SettingsManager. That string is a comma-separated list of experiment ids, for example `unbounce_trigger,lend_menu_category_sort`. Previous enabled/startTime/endTime settings are ignored, hopefully simplifying the process of running and stopping experiments. 
3. Any experiments that are not in the active experiment list are removed from the `uiab` cookie for each user, avoiding potential issues with the cookie becoming too large from filling up with old experiments.


eddieferrer (approved):
Great approach to this. Also thanks for cleaning up lodash functions!

dyersituations:
@emuvente I think we can close this PR? Most of these changes made it into the work I did last year on Denali.

emuvente:
@dyersituations  yeah I think you're right!

PR #3781, opened 2022-05-04
feat: decimal place was removed from the lendamountbutton component

roger-in-kiva:
This component is getting a number from the lend-by-category page and a string from the borrower profile page, an array type was added to the prop to consider this.

### DIFF
--- a/src/components/LoanCards/Buttons/LendAmountButton.vue
+++ b/src/components/LoanCards/Buttons/LendAmountButton.vue
@@ -31,7 +31,7 @@ export default {
 	},
 	computed: {
 		amountValue() {
-			return parseInt(parseFloat(this.amountLeft), 10);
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/Buttons/LendAmountButton.vue
```diff
@@ -25,10 +25,15 @@ export default {
...
 		},
 		amountLeft: {
-			type: Number,
+			type: [Number, String],
 			default: 20
 		}
 	},
+	computed: {
+		amountValue() {
+			return parseFloat(this.amountLeft);
```
mcstover, line 34: I think you'll need to use parseInt here so the decimal places are removed.

roger-in-kiva: Fixed!

mcstover (approved):

PR #3786, opened 2022-05-06
feat: borrower profile urgency experiment

eddieferrer:
* Adds lend countdown to borrower profile when loan ends in 3 days or less

ACK-288

This is the revised design: 
Screen Shot 2022-05-10 at 2 14 04 PM
Screen Shot 2022-05-10 at 2 13 43 PM

This will have to be refactored to accommodate the new experiment setup


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -275,7 +295,31 @@ export default {
 				return `${loanUse}\n\n${this.description.substring(0, 120)}...`;
 			}
 			return 'For the borrower\'s privacy, this loan has been made anonymous.';
+		},
+		showUrgencyExp() {
+			return this.hasThreeDaysOrLessLeft && this.isUrgencyExpVersionShown;
 		}
 	},
+	created() {
```
eddieferrer, line 303: I added the experiment query here in created, since I had some issues with getting the experiment to work in prefetch in conjunction with calculating `hasThreeDaysOrLessLeft`

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -275,7 +294,31 @@ export default {
...
+		/*
+		 * Experiment Initializations
+		*/
+		this.apollo.query({ query: experimentAssignmentQuery, variables: { id: 'lend_urgency' } })
+			.then(result => {
+				const urgencyEXP = result?.data?.experiment;
+				// Only track event if loan is expiring soon
+				this.isUrgencyExpVersionShown = urgencyEXP.version === 'shown';
```
mcstover, line 310: Couldn't this go within the 3 days condition? Why change the experience if you're not tracking that session?

mcstover: not a deal breaker, just kinda odd that some folks would see this without reason.

mcstover (approved):

src/pages/BorrowerProfile/BorrowerProfile.vue
mcstover: I wonder if you could have chained promises to get `hasThreeDaysOrLessLeft` ready for a final segment that run the experiment assignment...

mcstover: Yep. too sticky for an exp!

src/pages/BorrowerProfile/BorrowerProfile.vue
eddieferrer: I dont think it makes a difference. The experience won't be changed unless has 3 days or less (based on the computed prop above)

mcstover: Ok, I see now you have the `showUrgencyExp` prop with that additional check so all clear. Thanks!

eddieferrer: 👍  I'm gonna have to refactor this anyway, since this same experiment key will also drive the homepage part of this.

PR #3789, opened 2022-05-10
feat: flss filter styles VUE-1004

georgebs:
**AC**
Stub out the FLSS filters and the Advanced Filters container following the design mocks.

Screen Shot 2022-05-10 at 8 32 40

**Steps to test:**
Go to [http://localhost:8888/lend/filter-alpha](http://localhost:8888/lend/filter-alpha)
Filter menu should match the figma design

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchFilter.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchFilter.vue
@@ -1,6 +1,9 @@
 <template>
-	<div class="tw-bg-white tw-border-primary-inverse tw-rounded tw-p-3">
-		<p>Reset all</p>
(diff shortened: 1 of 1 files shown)


mcstover (approved):

src/components/Lend/LoanSearch/LoanSearchFilter.vue
```diff
@@ -180,3 +212,9 @@ export default {
 	},
 };
 </script>
+
+<style lang="scss" scoped>
+	.filter-min-w {
```
mcstover, line 217: In future work we may move this up onto the template as a style attribute instead.

PR #3792, opened 2022-05-10
feat: Remove category experiment info

edwinf-barreram:
#Description
- Removed code related to lend_by_category_v2 experiment info

#Ticket

[MARS-103](https://kiva.atlassian.net/browse/MARS-103)


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -220,7 +219,7 @@ export default {
 			return _without(this.categoryIds, ...this.customCategoryIds);
 		},
 		leadHeaderFilterLink() {
-			return this.lendFilterExpVersion === 'b' ? '/lend/filter' : '/lend';
+			return '/lend';
```
mcstover, line 222: This should be the other way around. We should not be intentionally directing traffic into legacy `/lend` unless there are a subset of query params that can't be fulfilled by the `/lend/filter` page. Seeing as this is a clean link, it should be `/lend/filter`.

edwinf-barreram: @mcstover Will do that change, thank you very much for correcting me! 👍

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -308,9 +307,7 @@ export default {
 		filterUrl() {
 			// initial release sends us back to /lend
 			// return `/lend/${this.$route.params.category || ''}`;
-			return this.lendFilterExpVersion === 'b'
-				? this.getAlgoliaFilterUrl()
-				: `/lend/${this.$route.params.category || ''}`;
+			return `/lend/${this.$route.params.category || ''}`;
```
mcstover, line 310: I think this one should also use `this.getAlgoliaFilterUrl()` as the return value which will check `import loanChannelQueryMapMixin from '@/plugins/loan-channel-query-map';` to get a `lend/filter` compatible url or progressively fall back to the loan channel url from the api or legacy `/lend` if needed.

src/pages/Lend/LendByCategoryPage.vue
mcstover: It's ok, this would generally make sense but the lend_by_category_V2 exp is tightly coupled across the UI app and our legacy app and actually originates here: https://github.com/kiva/kiva/blob/development/httpdocs/classes/API/Experiment/LendFilter.php

We need to do a bit more analysis on what removing this UI code does as we don't want to move back to a world where the legacy page gets traffic by default.

We can move this convo to slack to get into the details.

FYI @emuvente

emuvente: To be clear, this is about lend_filter_v2 not lend_by_category_v2

edwinf-barreram:
@emuvente @mcstover Guys, I have reverted the changes for lend_filter_v2 experiment

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -130,6 +130,7 @@ export default {
 		KvLoadingOverlay,
 		LendHeader,
 		AddToBasketInterstitial,
+		lendFilterExpVersion: '',
```
mcstover, line 133: This needs to be returned to the `data` object down below.

edwinf-barreram: You are right, I forgot to revert that part! 👍

mcstover, line 133: Please remove this line...I'm surprised you're not getting an error locally seeing as this isn't a component...

edwinf-barreram: Yeah, that's weird because I didn't see any compiling errors and also the linting checks not detected the bug

mcstover (approved):

src/pages/Lend/LendByCategoryPage.vue
mcstover: Thanks for fixing this up!

emuvente (approved):
👍

christian14b (approved):
👍🏾

PR #3793, opened 2022-05-11
feat: add promo creds to checkout page

egoroza:
adds promo credits to the checkout page and removes the legacy redirect.

### DIFF
--- a/src/graphql/query/checkout/initializeCheckout.graphql
+++ b/src/graphql/query/checkout/initializeCheckout.graphql
@@ -49,6 +49,18 @@ query initializeCheckout($basketId: String) {
 			items {
 				totalCount
 				values {
+					creditsUsed {
+						amount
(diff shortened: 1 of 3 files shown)


mcstover:
This looks correct and should set the stage for follow on items. I think it would be best to get these updates in incrementally and then start patching in the other features. In particular, I'd like to see this in dev or even just the removed redirect conditional if you feel the rest of this isn't ready...

egoroza:
> This looks correct and should set the stage for follow on items. I think it would be best to get these updates in incrementally and then start patching in the other features. In particular, I'd like to see this in dev or even just the removed redirect conditional if you feel the rest of this isn't ready...

okay cool! yeah i wanted to double check to see if i missed some stuff but this has at least the redirect removed + the credit row + attribution so i'm fine with merging it

PR #3797, opened 2022-05-11
feat: added property to enable rounded corners on the list loan card

dyersituations:
https://kiva.atlassian.net/browse/VUE-1008

- Added new prop to enable rounded corners on loan card
- Added true to new filter alpha
- Added stories for loan card variations

Without rounded (default)

image

image

With rounded

image

image

### DIFF
--- a/src/components/LoanCards/ListLoanCard.vue
+++ b/src/components/LoanCards/ListLoanCard.vue
@@ -1,6 +1,9 @@
 <template>
-	<div class="row list-loan-card">
-		<div class="list-loan-card-desktop-column show-for-large large-4 xlarge-4 columns">
+	<div class="row list-loan-card" :class="{ 'tw-rounded': roundedCorners }">
+		<div
+			class="list-loan-card-desktop-column show-for-large large-4 xlarge-4 columns"
+			:class="{ 'tw-rounded-l': roundedCorners }"
+		>
 			<div class="list-loan-card-desktop-column-container row">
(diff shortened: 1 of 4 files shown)


src/components/LoanCards/ListLoanCard.vue
```diff
@@ -1,6 +1,9 @@
 <template>
-	<div class="row list-loan-card">
-		<div class="list-loan-card-desktop-column show-for-large large-4 xlarge-4 columns">
+	<div class="row list-loan-card" :class="{ 'tw-rounded': roundedCorners }">
+		<div
+			class="list-loan-card-desktop-column show-for-large large-4 xlarge-4 columns"
+			:class="{ 'tw-rounded-l': roundedCorners }"
```
mcstover, line 5: Another way to do this would be to apply `tw-overflow-hidden` alongside `tw-rounded` on the `row list-loan-card` div. But, if does the trick and looks fine across various widths let's roll with it!

mcstover (approved):

src/components/LoanCards/ListLoanCard.vue
dyersituations: Ah yup, that's slightly simpler, thanks. I was thinking of not adding any new CSS that could potentially change the existing card behavior, but if we're good adding overflow hidden here, works for me. Done!

mcstover (approved):

PR #3798, opened 2022-05-11
Personalized bundle core 586

roger-in-kiva:


mcstover:
There's a lot of pending changes coming into the Lend by Category page from @emuvente 's team so I wanted to loop him in to help evaluate. @roger-in-kiva It may be that we need to let these other pending exps into `main` then get your updates added afterwards.

mcstover (approved):

PR #3799, opened 2022-05-12
feat: update event tracking for checkout/thanks pages

edwinf-barreram:
# Description

- Updated update event tracking for checkout/thanks pages

# Ticket

[MARS-83](https://kiva.atlassian.net/browse/MARS-83)

### DIFF
--- a/src/components/Checkout/MonthlyGoodCTA.vue
+++ b/src/components/Checkout/MonthlyGoodCTA.vue
@@ -190,10 +190,10 @@ export default {
 	},
 	computed: {
 		selectedAmountTracking() {
-			return ['MonthlyGood', 'click-amount-option', `[actual amount selected - ${this.mgOptionSelected}]`];
+			return ['MonthlyGood', 'click-amount-option', numeral(this.mgOptionSelected).format('$0,0.00')];
 		},
 		selectedCategoryTracking() {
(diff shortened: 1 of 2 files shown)


src/components/Checkout/MonthlyGoodCTA.vue
```diff
@@ -186,6 +188,14 @@ export default {
 			]
 		};
 	},
+	computed: {
+		selectedAmountTracking() {
+			return ['MonthlyGood', 'click-amount-option', `[actual amount selected - ${this.mgOptionSelected}]`];
+		},
+		selectedCategoryTracking() {
+			return ['MonthlyGood', 'click-category-option', `[actual category select - ${this.selectedGroup}]`];
```
mcstover, line 196: Both of these would be easier to filter on and create reports from if the label was more concise. I think we can reduce these to `amount: ###` + `category: ###`.

mcstover, line 196: Both of these would be easier to filter on and create reports from if the label was more concise. I think we can reduce these to `amount: ###` + `category: ###`.

emuvente (approved):
👍  Looks good!

mcstover (approved):

christian14b (approved):

PR #3801, opened 2022-05-12
Feat: Lazy loading loan rows on /lend-by-category

christian14b:
**Description**
This feature aims to fetch loan data rows depending on the viewport position using intersection observer browser api.

**Ticket**
[MARS-118](https://kiva.atlassian.net/browse/MARS-118?atlOrigin=eyJpIjoiYzgzOTM5NzNkMjAwNDBhZmFiNzgyZWU4NTQ4ZjlkOWUiLCJwIjoiaiJ9)


mcstover:
Thanks for keeping this up to date with the other exps coming into the page...there will be more soon lolz.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -557,6 +520,101 @@ export default {
...
+							});
+							this.recommendedLoans = allRecLoans;
+							this.rowLazyLoadComplete = true;
+						});
+					} catch (e) {
+						logReadQueryError(e, 'LendByCategory recommendedLoansQuery');
+					}
+				} else {
+					try {
+						return this.apollo.watchQuery({
```
emuvente, line 594: This might make more sense as a regular apollo.query instead of apollo.watchQuery.

emuvente (commented):
This is looking really good! I added a couple comments, but I also observed some odd behavior that I recorded.

https://user-images.githubusercontent.com/4149025/168931521-3fb4ac70-f8ff-446a-9c9f-3969a97489db.mov

1. The recommended loans row disappeared at one point, but then came back with a different title, different description and different loans. It sees like the loading of later rows is overwriting that somehow.
2. Sometimes scrolling to the bottom of the page would not show any loans. This seemed to be caused by different things, like no loans being returned for some rows. I think this could be fixed by calling fetchLoanData again after a row finishes loading if the foot is still in the viewport at that point.
3. It looks like the tracking events will only fire once all the rows have loaded. I'm assuming that the right thing to do would be to fire a new tracking event each time a row loads, but I think we would have to check in with data science about that. @mcstover thoughts?

emuvente (approved):
This is ready to go! Thanks for putting this all together!

mcstover (approved):
👍

PR #3804, opened 2022-05-13
feat: kvlightbox added to filter menu VUE-1021

georgebs:
Added functionality to show the filter menu within a KvLightbox component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [2.144.0](https://github.com/kiva/ui/compare/v2.143.1...v2.144.0) (2022-05-13)
+
+
+### Features
+
+* dedicated new gender radio filter using KvRadio ([8d7537d](https://github.com/kiva/ui/commit/8d7537d339a1b99ca55fcacf0c9cc0e606cfa1ed))
+
 ### [2.143.1](https://github.com/kiva/ui/compare/v2.143.0...v2.143.1) (2022-05-13)
 
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "kvue",
-	"version": "2.143.1",
+	"version": "2.144.0",
 	"description": "The Kiva ui",
 	"author": "braincrave",
 	"private": true,


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -1,15 +1,16 @@
...
+	<div class="tw-flex" @keyup.esc="closeLightbox">
 		<div class="tw-flex tw-flex-col tw-mr-4">
 			<div class="md:tw-hidden tw-mb-3">
-				<kv-button variant="secondary">
+				<kv-button variant="secondary" @click="openLightbox">
 					Filter & Sort
 				</kv-button>
+
+				<kv-lightbox :visible="isLightboxVisible" title="" preventClose
+					noPaddingTop variant="lightbox" @lightbox-closed="closeLightbox"
```
mcstover, line 10: This doesn't existing on the new version of KvLightbox you're using. Please remove for now. There is some precedent for when we need to make hard-coded changes to the new lightbox styles. Here's an example which is similar to what you were doing I think, please try to keep a light touch ;)
https://github.com/kiva/ui/blob/main/src/components/Checkout/AutoDepositUpsell.vue#L282

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -1,15 +1,16 @@
...
-		</div>
+	<div class="tw-flex" @keyup.esc="closeLightbox">
 		<div class="tw-flex tw-flex-col tw-mr-4">
 			<div class="md:tw-hidden tw-mb-3">
-				<kv-button variant="secondary">
+				<kv-button variant="secondary" @click="openLightbox">
 					Filter & Sort
 				</kv-button>
+
+				<kv-lightbox :visible="isLightboxVisible" title="" preventClose
```
mcstover, line 9: We shouldn't need the `preventClose` prop here. There's no reason to prevent closing the filter menu

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -1,15 +1,16 @@
 <template>
-	<div class="tw-flex">
-		<div id="mobile-menu" class="tw-rounded md:tw-hidden
-				tw-z-dropdown tw-drop-shadow tw-absolute tw-top-4 md:tw-top-0"
-		>
-			<loan-search-filter />
-		</div>
+	<div class="tw-flex" @keyup.esc="closeLightbox">
```
mcstover, line 2: The `@keyup` portion of this shouldn't be needed as it's built in to the component

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -42,25 +43,38 @@ import LoanCardController from '@/components/LoanCards/LoanCardController';
...
 		};
 	},
+	methods: {
+		openLightbox() {
+			this.isLightboxVisible = true;
+		},
+		closeLightbox() {
+			this.isLightboxVisible = false;
+			this.lightboxTitle = '';
+			this.lightboxContent = null;
```
mcstover, line 70: Is this being used?

src/components/Lend/LoanSearch/LoanSearchInterface.vue
dyersituations: @mcstover I'm assuming you're referring to `noPaddingTop`. If not, VSCode is auto formatting that to use hyphens for me, so @georgebs maybe check your auto save actions.

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -13,7 +13,7 @@
 							</p>
 						</div>
 						<button class="tw-mb-2 tw-mt-3 tw-border-r tw-border-tertiary tw-px-1 md:tw-px-2">
-							<kv-material-icon :icon="mdiEarth" class="tw-text-secondary tw-w-3 tw-h-3" />
+							<kv-material-icon :icon="mdiEarth" class="tw-text-brand tw-w-3 tw-h-3" />
```
dyersituations, line 16: It looks like this change is making both icons green/selected, is that what we want?

dyersituations:
Two UX questions:
- Does this ticked include making the filters full width of the modal?
- Should clicking outside the modal close it? Currently you have to click the "X" button.
image

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: Correct `noPaddingTop` was a prop on our older version of KvLightbox, but we're using the version from kv-ui-components now which doesn't have that option. We have to use some hacky override styles if we "really" need to do something beyond the standard.

mcstover:
@georgebs @dyersituations For this PR, I'm more concerned about the functionality and keeping the code pretty clean for now. @georgebs you can create some follow up PR's for any polish and visual work you'd like to add in.

PR #3805, opened 2022-05-13
Vue 1006 basic loan search state

mcstover:
This may prove to be a temporary construct, but for now it provides a place to hold, way to transport and subscribe to changes of a generic search state object. It will need to be kept in sync with any/all filters we may use across the standard `lend > loans` and `flss` queries.


mcstover:
Also, I've got a follow up PR including this that will re-enable showing loan cards with a sample for collecting and subscribing to changes in this localResolver object.

dyersituations (approved):

PR #3806, opened 2022-05-13
Vue 1006 v2 stub for loan search state watcher

mcstover:
This is mostly a temporary implementation as it hard codes and initial query. It does however provide a short path to get loans showing on the page again. And marks items in the Filter Interface for removal and cleanup or for extraction into their specific filter components.

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchFilter.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchFilter.vue
@@ -85,24 +85,24 @@
 </template>
 
 <script>
-
 import {
-	fetchData,
-	filterGender,
-	filterSector,
+	// fetchData,
+	// filterGender,
+	// filterSector,
 	fetchSectors,
 	fetchCountryFacets,
-	filterCountry,
-	filterLoanTerm,
+	// filterCountry,
+	// filterLoanTerm,
 } from '@/util/flssUtils';
 import KvAccordionItem from '@/components/Kv/KvAccordionItem';
 import { mdiClose, mdiArrowRight } from '@mdi/js';
+import updateLoanSearchMutation from '@/graphql/mutation/updateLoanSearchState.graphql';
 import KvRadio from '~/@kiva/kv-components/vue/KvRadio';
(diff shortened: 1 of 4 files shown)


src/api/localResolvers/loanSearch.js
```diff
@@ -0,0 +1,61 @@
+import loanSearchStateQuery from '@/graphql/query/loanSearchState.graphql';
+// import gql from 'graphql-tag';
+
+// eslint-disable-next-line no-underscore-dangle
```
dyersituations, line 4: I've noticed this eslint exclusion a lot, have we considered adding an exception to the config?

src/components/Lend/LoanSearch/LoanSearchFilter.vue
```diff
@@ -85,24 +85,24 @@
...
 import KvAccordionItem from '@/components/Kv/KvAccordionItem';
 import { mdiClose, mdiArrowRight } from '@mdi/js';
+import updateLoanSearchMutation from '@/graphql/mutation/updateLoanSearchState.graphql';
 import KvRadio from '~/@kiva/kv-components/vue/KvRadio';
 import KvCheckbox from '~/@kiva/kv-components/vue/KvCheckbox';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
 
 export default {
-	inject: ['apollo'],
+	inject: ['apollo', 'cookieStore'],
```
dyersituations, line 105: Where is `cookieStore` used?

dyersituations (approved):

src/api/localResolvers/loanSearch.js
mcstover: Yea it does show up in a few places but always in the context of using `__typename`. So yea if we could make that an exception it would be good to clean out those eslint bypasses

src/components/Lend/LoanSearch/LoanSearchFilter.vue
mcstover: Checkout https://github.com/kiva/ui/blob/main/src/api/apollo.js to see how cookieStore is a dependency of some of our Apollo links.

PR #3807, opened 2022-05-13
feat: dedicated new gender radio filter using KvRadio

dyersituations:
https://kiva.atlassian.net/browse/VUE-1010

- Implemented dedicated radio group for gender
- Added TODOs for some decisions pending new search cache + behavior (new ticket: https://kiva.atlassian.net/browse/VUE-1028)
- Radios won't be clickable until kv-components fix is merged (see other PR coming soon)
- Added new stories + test sections for loan search

image

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
@@ -1,0 +1,76 @@
+<template>
+	<fieldset class="tw-flex tw-flex-col tw-gap-2 tw-my-2">
+		<kv-radio
+			v-for="option in genderOptions"
+			:key="option.key"
+			:value="option.key"
+			v-model="selectedGender"
+			@change="setGender"
+		>
+			{{ option.title }}
+		</kv-radio>
+	</fieldset>
+</template>
+
+<script>
+import KvRadio from '~/@kiva/kv-components/vue/KvRadio';
+
+export const BOTH_KEY = 'both';
+export const BOTH_TITLE = 'All genders';
+export const FEMALE_KEY = 'female';
+export const FEMALE_TITLE = 'Women';
+export const MALE_KEY = 'male';
+export const MALE_TITLE = 'Men';
+const GENDER_KEYS = [BOTH_KEY, FEMALE_KEY, MALE_KEY];
+
+export default {
+	components: {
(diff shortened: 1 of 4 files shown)


src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
```diff
@@ -0,0 +1,76 @@
...
+<script>
+import KvRadio from '~/@kiva/kv-components/vue/KvRadio';
+
+export const BOTH_KEY = 'both';
+export const BOTH_TITLE = 'All genders';
+export const FEMALE_KEY = 'female';
+export const FEMALE_TITLE = 'Women';
+export const MALE_KEY = 'male';
+export const MALE_TITLE = 'Men';
+const GENDER_KEYS = [BOTH_KEY, FEMALE_KEY, MALE_KEY];
```
mcstover, line 24: Clever use of constants, making them available for the test files too!

src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
```diff
@@ -0,0 +1,76 @@
...
+		gender: {
+			type: String,
+			// TODO: determine if the default should be "null" (old component behavior) or "both"
+			default: BOTH_KEY
+		}
+	},
+	data() {
+		return {
+			selectedGender: this.gender || BOTH_KEY,
+			genderOptions: [
```
mcstover, line 40: The Gender options will likely become a dynamic data set passed in at some point but in this limited case the data set is currently small enough that it's ok to have them static and hard-coded here.

mcstover (approved):
Nice clean up and de-coupling of this code

src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
dyersituations: Yeah I was wondering if this component should be a generic radio group, but it'll be easy to adjust later.

src/components/Lend/LoanSearch/LoanSearchGenderFilter.vue
dyersituations: I miss TypeScript in a situation like this, but using constants instead of magic strings works well enough.

PR #3812, opened 2022-05-16
feat: implement campaign verification form on checkout

egoroza:
adds the campaign verification form and verify remove promo credit lightbox to `CheckoutPage`

Screen Shot 2022-05-16 at 12 59 05 PM
Screen Shot 2022-05-17 at 11 17 35 AM

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -178,7 +178,6 @@
 					/>
 				</div>
 			</div>
-
 			<campaign-verification-form
 				:form-id="externalFormId"
 				:ma-id="String(managedAccountId)"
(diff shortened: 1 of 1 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -730,7 +771,18 @@ export default {
...
+		getPromoInformationFromBasket() {
+			getPromoFromBasket(this.derivedPromoFund?.id, this.apollo).then(({ data }) => {
+				this.promoData = data?.shop?.promoCampaign;
+			});
+		},
+		verificationComplete() {
+			this.verificationSubmitted = true;
+		},
+		handleVerificationOptOut() {
+			this.showVerification = false;
```
mcstover, line 784: This method will also need to set `this.showVerifyRemovePromoCredit = true` which likely means we'll need to add the `verify-remove-promo-credit` component here in a similar fashion as the CCLandingPage does. We can make another ticket for this if you want but both have to be in place before this code is released in prod.

Note: the VerifyRemovePromoCredit component does function on this page already but it's nested down in the OrderTotals component. I don't "think" there will be an issue with including it in both places as we do this on CCLandingPage > InContextCheckout too.

mcstover (approved):
I say let's give it a go. We'll def want to do some full scope testing in DEV for these new flows

roger-in-kiva (approved):

PR #3815, opened 2022-05-16
Vue 1029 move update search state method to new src utils loan search utils js file

dyersituations:
https://kiva.atlassian.net/browse/VUE-1029

- Created new util file
- Added associated tests file
- Fixed a couple vue errors from previous PRs
- Loans still appearing in search like before
- Added missing auth0 mock for local dev that was causing an error

image


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -29,7 +34,7 @@
 					:key="loan.id"
 					:loan="loan"
 					loan-card-type="ListLoanCard"
-					rounded-corners="true"
+					:rounded-corners="true"
```
mcstover, line 37: 👍  Saw this happening when used the other day too!

src/util/KvAuth0.js
```diff
@@ -357,6 +357,7 @@ export const MockKvAuth0 = {
 	getKivaId: () => undefined,
 	getLastLogin: () => 0,
 	getMfaEnrollToken: () => Promise.resolve({}),
+	fakeAuthAllowed: () => undefined,
```
mcstover, line 360: @emuvente I think this is ok, though I suppose it could return `false` as well...

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -6,7 +6,12 @@
 					Filter & Sort
 				</kv-button>
 
-				<kv-lightbox :visible="isLightboxVisible" variant="lightbox" @lightbox-closed="closeLightbox">
+				<kv-lightbox
+					:visible="isLightboxVisible"
+					variant="lightbox"
+					title=""
```
mcstover, line 12: Fine for now but not very "descriptive" for a screen reader. We should eventually add something like "Loan Filter Controls" or something of that nature...

mcstover (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
dyersituations: Ah I didn't realize this `title` was also used for `aria-label`. I found a workaround, passing `{{ null }}` to a slot template. Found here: https://github.com/vuejs/vue/issues/7064

src/util/KvAuth0.js
dyersituations: Where `fakeAuthAllowed` is used currently, anything falsy would work, but I just changed it to `false` to be more explicit.

mcstover (approved):

PR #3816, opened 2022-05-17
feat: implement monthly good digest lightbox

egoroza:
adds a lightbox when the `utm_campaign=liked_loan` query is present for users who click through the MG digest email

MOBILE

Screen Shot 2022-05-17 at 12 04 07 PM

DESKTOP 

Screen Shot 2022-05-17 at 12 05 11 PM


src/components/LoansByCategory/MGDigestLightbox.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+	<kv-lightbox
+		:visible="isLightboxVisible"
+		title=""
```
mcstover, line 4: Please add `Thank you for the feedback!` as the value here.

src/components/LoansByCategory/MGDigestLightbox.vue
```diff
@@ -0,0 +1,92 @@
...
+		@lightbox-closed="closeLightbox"
+	>
+		<h2 class="tw-text-h2 tw-text-center">
```
mcstover, line 7: Wrap this `h2` in a template to use it as the header `<template #header>` and gain control over how it's styled.

mcstover (approved):
The 2 comments around the lightbox header are about ensuring the aria label are properly set on the components. Please try to get this dialed in before release.

egoroza:
> The 2 comments around the lightbox header are about ensuring the aria label is properly set on the components. Please try to get this dialed in before release.

gotcha, ya i think i have to do a bit of hacky css since design doesn't want the title aligned with the `x` and they want it centered - a bit tricky with the title and `x` being in the same flexbox even when using the custom header - might be a good opportunity to add some kind of centered title functionality to `KvLightbox` in the future?

src/components/LoansByCategory/MGDigestLightbox.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+	<kv-lightbox
+		title="Thank you for the feedback!"
+		:visible="isLightboxVisible"
+		@lightbox-closed="closeLightbox"
+	>
+		<template #header>
+			<h2 class="tw-hidden">
```
mcstover, line 8: this could just be an empty span as well but it's fine for now

PR #3818, opened 2022-05-17
fix: remove legacy redirect for promo code state when adding a promo …

mcstover:
…code in new checkout

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -179,30 +179,6 @@
 				</div>
 			</div>
 
-			<kv-lightbox
-				:visible="redirectLightboxVisible"
-				title="This checkout is being tested right now, but doesn't support some functions yet."
-				@lightbox-closed="redirectLightboxClosed"
-				data-testid="checkout-legacy-redirect-lightbox"
-			>
-				<p class="tw-mb-4">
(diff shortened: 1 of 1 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -645,19 +645,12 @@ export default {
...
 			}).then(({ data }) => {
-				// when updating basket state, check for free credits and redirect if present
+				// when updating basket state, check for free credits and remove guest checkout option
 				const hasFreeCredits = _get(data, 'shop.basket.hasFreeCredits');
-				if (hasFreeCredits) {
-					if (refreshEvent === 'kiva-card-applied') {
-						this.$kvTrackEvent('basket', 'free credits applied', 'exit to legacy');
-						this.disableGuestCheckout();
-					}
-					this.redirectLightboxVisible = true;
```
dyersituations, line 655: Can `redirectLightboxVisible` and the associated lightbox be removed? I'm not seeing that flag being set to `true` anymore in this file.

mcstover: Yea good call. Was also just looking at that.

There is still one redirect case but we're handling it directly in the apollo prefetch

dyersituations (approved):

PR #3819, opened 2022-05-18
feat: wrong message was removed in MG thanks page

roger-in-kiva:


mcstover (approved):
Are we actually getting any traffic `fromCovidLanding` ???

roger-in-kiva:
> Are we actually getting any traffic `fromCovidLanding` ???

Not sure, but Vijaya suggested to keep this.

Are we able to remove this now? @mcstover

PR #3824, opened 2022-05-19
Test credit checkout

davidskiva:
This is and e23 test that validates a lender, with kiva credit, can purchase a loan and that their balance is dropped by correct amount.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ build/schema.graphql
 .config.env
 config.env
 jsconfig.json
+cypress.env.json
--- a/test/e2e/specs/ExistingUserCreditCheckout.spec.js
+++ b/test/e2e/specs/ExistingUserCreditCheckout.spec.js
@@ -2,14 +2,17 @@ describe('Existing User Checkout with Kiva Credit', () => {
 	it('Baskets a loan, proceed to checkout, log in user with credit and purchases using kiva credit', () => {
 		let beginBalance;
 		let newBalance;
+		// Cypress.env();
+		const lenderID = Cypress.env().userInfo.kivaCreditLender.userID;
+		const loginCookie = `${lenderID}:recent/active/mfa`;
 
-		// Mock log in
-		cy.setCookie('kvfa', '1003394:recent/active/mfa', 'domain=.kiva.org');
-		// go to portfolio and get current credit
+		// Mock log in as kivaquality7
+		cy.setCookie('kvfa', loginCookie, 'domain=.kiva.org');
+		// go to portfolio and get current credit for later calculations
 		cy.visit('/portfolio');
 		cy.get('.strong .kiva-green').invoke('text').then(fullText => {
-			const number = fullText.match(/[0-9]+/);
-			beginBalance = Number(number); // text to numeric
+			const number = fullText.match(/[0-9,.]+/);
+			beginBalance = parseFloat(number); // text to float
 		});
 		// Go to lend page
 		cy.visit('/lend?x');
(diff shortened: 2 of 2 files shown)


test/e2e/specs/ExistingUserCreditCheckout.spec.js
```diff
@@ -0,0 +1,55 @@
...
+		// Mock login our test lender
+		cy.setCookie('kvfa', loginCookie, 'domain=.kiva.org');
+		// go to portfolio and get current credit for later calculations
+		cy.visit('/portfolio');
+		cy.get('.strong .kiva-green').invoke('text').then(fullText => {
+			const number = fullText.match(/[0-9,.]+/);
+			beginBalance = parseFloat(number); // text to float
+		});
+		// Go to lend page
+		cy.visit('/lend?x');
```
mcstover, line 18: Can we use the loan carousel on the `/checkout` page to add a loan instead of adding a dependency on the legacy stack being present?

If not, some other alternatives would be to use a Ui app category page such as `/lend-by-category/agriculture`

.gitignore
```diff
@@ -20,3 +20,4 @@ build/schema.graphql
 .config.env
 config.env
 jsconfig.json
+cypress.env.json
```
mcstover, line 23: it does seem like we'll need to add this config somewhere so that other devs can pull it onto their machine and so that the pipeline will have it when attempting to run the test.

test/e2e/specs/ExistingUserCreditCheckout.spec.js
```diff
@@ -0,0 +1,55 @@
...
+		// Set sort to random
+		cy.get('#filter-loan-sort-dropdown', { timeout: 10000 }).as('sortFilter');
+		cy.wait(5000);
```
dyersituations, line 21: What is this particular wait needed for? Would it be possible to wait for an element to appear or request to return, depending on the situation?

mcstover: My guess is this is just to deal with slow loading pages in the lower envs...

.gitignore
davidskiva: @mcstover where do you suggest it should be added at?

test/e2e/specs/ExistingUserCreditCheckout.spec.js
davidskiva: Okay, I'll make a change to use on category page

.gitignore
mcstover: It will live over in `marketplace-web-ui-ci` where you're adding it. we'll prob just want to add some documentation somewhere so folks know where to get it...

test/e2e/specs/ExistingUserCreditCheckout.spec.js
mcstover: Oh yea, good call on test ids. None of the lend buttons currently have data-testid attributes :(

test/e2e/specs/ExistingUserCreditCheckout.spec.js
```diff
@@ -0,0 +1,51 @@
...
+		cy.findAllByTestId('total-amount').first().should('to.contain', '$28.75');
+		// get the order total and convert it to a float for caluations
+		cy.findAllByTestId('total-amount').first().invoke('text').then(fullText => {
+			const number = fullText.match(/[0-9,.]+/);
+			const usedBalance = parseFloat(number); // text to float
+			// subtract order total from beginning to get new balance
+			newBalance = beginBalance - usedBalance;
+		});
+		// return to portfolio and get new balance
+		cy.visit('/portfolio');
```
mcstover, line 43: This is taking us out of the UI application. We already know the Transaction went through b/c we made it to the thanks page so I would generally say this isn't even needed.

If you feel it's important to verify that a users balance was changed due to the transaction, could you instead move to the home page or refresh the thanks page and then check against the value in the header? I realize that is a rounded number but it should still be sufficient to know if the money was deducted.

mcstover (approved):

test/e2e/specs/ExistingUserCreditCheckout.spec.js
```diff
@@ -0,0 +1,51 @@
...
+		cy.setCookie('kvfa', loginCookie, 'domain=.kiva.org');
+		// go to portfolio and get current credit for later calculations
+		cy.visit('/portfolio');
+		cy.get('.strong .kiva-green').invoke('text').then(fullText => {
+			const number = fullText.match(/[0-9,.]+/);
+			beginBalance = parseFloat(number); // text to float
+		});
+		// Go to category page for women
+		cy.visit('/lend-by-category/women');
+		// // Wait for the lend or lend again button to exist, then click it
```
emuvente, line 19: Rather than using an explicit timeout for these commands, I would recommend intercepting the GraphQL requests before the `cy.setCookie` and `cy.visit` commands, and waiting for all those requests to finish each time:

```js
// Spy on GraphQL requests
cy.intercept('POST', '**/graphql*').as('graphqlRequest');
// Mock login our test lender
cy.setCookie('kvfa', loginCookie, 'domain=.kiva.org');
// go to portfolio and get current credit for later calculations
cy.visit('/portfolio');
cy.get('.strong .kiva-green').invoke('text').then(fullText => {
	const number = fullText.match(/[0-9,.]+/);
	beginBalance = parseFloat(number); // text to floats
});
// Go to category page for women
cy.visit('/lend-by-category/women');
// Wait for the lend or lend again button to exist, then click it
cy.get('@graphqlRequest.all');
cy.findAllByText('Lend now').first().click();
// wait for the basket add lightbox
cy.get('@graphqlRequest.all');
cy.get('.kv-lightbox').should('be.visible');
// Wait for the checkout button to exist in the lightbox, then click it
cy.get('.button-checkout').click();
// Wait for basket to load and check for complete order button
cy.get('@graphqlRequest.all');
cy.findByText('Basket');
cy.findByText('Complete order').click();
// wait for order confirmation page to load
cy.get('@graphqlRequest.all');
cy.findByText('Thank you!');
```

mcstover: I like this approach to understanding when you are clear to take the next action.

On another note, I realized there's another exit out of the UI app to portfolio. We should be able to get the balance from the Header and not leave the Ui app.

davidskiva: @mcstover the reason why I did not use the amount from the header is because it displays a rounded number.  We need to have the full amount including cents and the portfolio page has that.

davidskiva: @emuvente thanks for the recommendation, I have updated the test to use cy.intercept.

emuvente (approved):
👍 looks good!

mcstover (approved):

mcstover (approved):
Thanks @davidskiva Good stuff!

mcstover (approved):

PR #3828, opened 2022-05-20
feat: lightbox for personalization wizard

roger-in-kiva:
@egoroza feel free to make any upgrade as you want; should be great having green buttons in the results page.


egoroza (approved):
awesome!!

PR #3830, opened 2022-05-20
fix: fix query param check and lbc route on LBC MG digest lightbox

egoroza:


mcstover:
👍

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -723,7 +723,7 @@ export default {
 			this.setLeftArrowPosition();
 		}
 
-		if (this.$route.query.utm_campaign === 'liked_loan') {
+		if (this.$route.query.utm_campaign === 'liked_loan' || this.$route.query.utm_campaign.includes('liked_loan')) {
```
mcstover, line 726: Prob could just use the new `includes` check to cover all cases...

PR #3835, opened 2022-05-24
feat: redirect to new variant for paid ads lp

christian14b:
# Description
This feature sets an experiment for redirecting to a new version of the paid ads landing page

# Ticket
[MARS-124](https://kiva.atlassian.net/browse/MARS-124?atlOrigin=eyJpIjoiYWU0ODY1OWQ3MDRkNGNkOTg4YWRiZmI3NzEzYmQzN2QiLCJwIjoiaiJ9)

### DIFF
--- a/src/graphql/query/contentfulEntries.graphql
+++ b/src/graphql/query/contentfulEntries.graphql
@@ -1,7 +1,19 @@
 # The preview parameter should come from a route query param and
 # should not be passed in manually. It is handled by src/api/ContentfulPreviewLink.js
-query contentfulEntries($contentType: String!, $contentKey: String, $preview: Boolean) {
+query contentfulEntries(
+	$contentType: String!
+	$contentKey: String
+	$preview: Boolean
+) {
 	contentful {
-		entries(contentType: $contentType, contentKey: $contentKey, preview: $preview)
+		entries(
+			contentType: $contentType
+			contentKey: $contentKey
+			preview: $preview
+		)
+	}
+	experiment(id: "Experiment:manual_lending_lp") @client {
+		id
+		version
 	}
 }
--- a/src/graphql/query/contentfulEntries.graphql
+++ b/src/graphql/query/contentfulEntries.graphql
@@ -12,8 +12,4 @@ query contentfulEntries(
 			preview: $preview
 		)
 	}
-	experiment(id: "Experiment:manual_lending_lp") @client {
-		id
-		version
-	}
 }
--- a/src/graphql/query/wwwHeader.graphql
+++ b/src/graphql/query/wwwHeader.graphql
@@ -47,4 +47,12 @@ query wwwHeader($basketId: String) {
 			id
 		}
 	}
+	general {
+		manual_lending_lp_exp: uiExperimentSetting(
+			key: "manual_lending_lp"
+		) {
+			key
+			value
+		}
+	}
 }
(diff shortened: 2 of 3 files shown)


src/graphql/query/wwwHeader.graphql
```diff
@@ -47,4 +47,12 @@ query wwwHeader($basketId: String) {
 			id
 		}
 	}
+	general {
```
mcstover, line 50: Is this actually needed? With this it will be fetched on every page load...

src/graphql/query/contentfulEntries.graphql
```diff
@@ -1,7 +1,19 @@
...
+) {
 	contentful {
-		entries(contentType: $contentType, contentKey: $contentKey, preview: $preview)
+		entries(
+			contentType: $contentType
+			contentKey: $contentKey
+			preview: $preview
+		)
+	}
+	experiment(id: "Experiment:manual_lending_lp") @client {
```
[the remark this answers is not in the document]
christian14b: You are absolutely right. I'm going to be working in better way. Thanks!

src/pages/ContentfulPage.vue
```diff
@@ -215,6 +237,34 @@ export default {
 			};
 		},
 		preFetch(config, client, args) {
+			if (args?.route?.path === '/lp/home-ml') {
+				client.query({ query: pageQuery }).then(() => {
```
emuvente, line 241: This promise needs to be returned, otherwise it won't prevent the contentful entries fro being fetched down on line 268. It might be cleaner to write this using async/await:
```
preFetch: async function(config, client, args) {
	if (args?.route?.path === '/lp/home-ml') {
		const result = await expQuery...
		const version = result?.data?.experiment?.version;
		const { enabled } = getExperimentSettingCached(client, manualLendingLPExpKey);
		
		if (enabled) {
			...
		}
	} else {
		return client.query({
				query: contentfulEntries,
				...rest of existing prefetch

Comment (christian14b) [reply src/pages/ContentfulPage.vue:241]: Thanks @emuvente ! Just one question regarding `pageQuery` that should be executed before the experiment query, right? As:

```
...
await client.query({ query: pageQuery });
const result = await client.query({ query: experimentQuery, variables: { id: manualLendingLPExpKey } })
...
```

File: src/pages/ContentfulPage.vue [hunk 4]
```diff
@@ -214,34 +236,57 @@ export default {
...
-				return Promise.all([
-					pageFrame(),
-					...contentGroups.map(g => g.component()),
-				]);
-			}).then(resolvedImports => {
-				// Call preFetch for page frame and content group components
-				const components = resolvedImports.map(resolvedImport => resolvedImport.default);
-				return preFetchAll(components, client, args);
-			});
+			} else {
```
emuvente, line 261: Now that I'm looking at it, I don't think this needs to be wrapped in an `else`. The `return Promise.reject` statement above will handle the case where a redirect is needed, and everything else should fall through to making the contetnfulEntries query. Sorry for the back-and-forth on this!

christian14b: No problem! Didn't realize that either, thanks

emuvente (approved):
👍  Looks great!

mcstover:
Thanks for optimizing this a bit to prevent it from always running

PR #3837, opened 2022-05-25
fix: validation was added for utm campaign

roger-in-kiva:


src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -723,7 +723,8 @@ export default {
 			this.setLeftArrowPosition();
 		}
 
-		if (this.$route.query.utm_campaign === 'liked_loan' || this.$route.query.utm_campaign.includes('liked_loan')) {
+		if (this.$route.query.utm_campaign && (!!this.$route.query.utm_campaign === 'liked_loan'
```
mcstover, line 726: @roger-in-kiva I think this can be simplified to a single check using optional chaining. Something like:

`if (this.$route?.query?.utm_campaign?.includes('liked_loan')) { // code ...}`

roger-in-kiva: Oh yeah, and working fine!

mcstover (approved):
I think this will help. We'll still want to look out for this issue incase it's a matter of support for the `includes` method but hopefully this will clear it up.

PR #3838, opened 2022-05-25
Vue 1037 location filter numbers are not always accurate

dyersituations:
https://kiva.atlassian.net/browse/VUE-1037

- Location filters are now based on the results of an FLSS facets query (filled out data from lend API, since FLSS doesn't return all of the data)
- Location checkboxes now get disabled when no fundraising loans match (like previous filters)
- Location option numbers now change based on the other filters
- Renamed utils test file to match naming convention

image

image


mcstover (approved):
Excellent updates!

PR #3841, opened 2022-05-26
feat: upsells checkout a/b test

egoroza:
implements upsells a/b test on checkout page

### DIFF
--- a/src/graphql/query/checkout/initializeCheckout.graphql
+++ b/src/graphql/query/checkout/initializeCheckout.graphql
@@ -30,6 +30,10 @@ query initializeCheckout($basketId: String) {
 			value
 			description
 		}
+		upsellsEnabled: uiExperimentSetting(key: "upsells_checkout") {
+			key
+			value
+		}
 	}
 	shop (basketId: $basketId) {
 		id
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -362,18 +362,6 @@ export default {
 				});
 		},
 		result({ data }) {
-			const upsellsExperiment = this.apollo.readFragment({
-				id: 'Experiment:upsells_checkout',
-				fragment: experimentVersionFragment,
-			}) || {};
-			this.isUpsellsExperimentEnabled = upsellsExperiment.version === 'b';
-			if (upsellsExperiment.version) {
-				this.$kvTrackEvent(
-					'Basket',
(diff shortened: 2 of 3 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -323,6 +324,7 @@ export default {
 			verificationSubmitted: false,
 			showVerification: false,
 			showVerifyRemovePromoCredit: false,
+			isUpsellsExperimentEnabled: false,
```
mcstover, line 327: What will this flag control?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -354,11 +356,24 @@ export default {
...
 				.then(() => {
 					return Promise.all([
-						client.query({ query: initializeCheckout, fetchPolicy: 'network-only' })
+						client.query({ query: initializeCheckout, fetchPolicy: 'network-only' }),
+						client.query({ query: experimentQuery, variables: { id: 'upsells_checkout' } }),
 					]);
 				});
 		},
 		result({ data }) {
+			const upsellsExperiment = this.apollo.readFragment({
```
mcstover, line 365: Please move all of this new code from the `result` block to either `created` or `mounted`

src/pages/Checkout/CheckoutPage.vue
mcstover: I see the design now. Looks nice!

mcstover (approved):

egoroza:
noticed that the `Experiment:upsells_checkout` query returns a response, but returns `null` for the version 🤔  unsure why...

src/graphql/query/checkout/initializeCheckout.graphql
```diff
@@ -30,6 +30,10 @@ query initializeCheckout($basketId: String) {
 			value
 			description
 		}
+		upsellsEnabled: uiExperimentSetting(key: "upsells_checkout") {
```
mcstover, line 33: @egoroza Please move this Experiment setting to https://github.com/kiva/ui/blob/main/src/graphql/query/checkout/checkoutSettings.graphql and things will start to work.

The `initializeCheckout` is running a network only call and bypassing the cache. Experiment assignment depends on the setting existing in the cache.

mcstover: @egoroza Please remove this additional query block from this file.

PR #3843, opened 2022-05-26
fix: call trackExperimentVersion in mounted hook

christian14b:
Fix calling `trackExperimentVersion` in mounted hook instead of calling it in `prefetch` function that caused an internal error.


mcstover (approved):
Looks right to me

emuvente (approved):
👍  good fix!

PR #3844, opened 2022-05-26
Vue 1025 activate theme attribute filters

dyersituations:
https://kiva.atlassian.net/browse/VUE-1025

- Implemented new theme filter
- Theme filter apply to loan search
- Theme filter numbers influenced by results like location filter numbers
- Bug logged where filter numbers aren't always matching
- Moved facets query method from `loanUtils`
- Updated cache graphql return values for consistency (even though they don't affect the data returned)

image


mcstover (approved):
Nice additions and updates!

PR #3846, opened 2022-05-27
Vue 1026 loan search sort by

mcstover:


src/api/localResolvers/loanSearch.js
```diff
@@ -24,7 +24,7 @@ export default () => {
 				gender: '', // expects a string
 				countryIsoCode: [], // expects an array of strings
 				sectorId: [], // expects an array of ints
-				sortBy: '', // expects a string
+				sortBy: null, // expects a string
```
mcstover, line 27: With `enum` types in graphql you can pass null rather than an empty string if you intend for the default value to be used.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -142,15 +143,19 @@ export default {
 				if (!this.allFacets) this.allFacets = await fetchLoanFacets(this.apollo);
 
 				// Get filtered facet options from FLSS
+				// TODO: Prevent this from running on every query
 				const { isoCodes, themes } = await runFacetsQueries(this.apollo, data?.loanSearchState);
```
mcstover, line 147: @dyersituations This method is currently called with every loan query. A future optimization would be to structure the data so we could check for changes that require updates to the facet counts. ie. Sort and Paging updates wouldn't require an update.

dyersituations (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: No big rush on this one, we'll make a ticket for later

PR #3847, opened 2022-05-27
feat: add lending page components caching

edwinf-barreram:
# Description
- Added cache function to Lending page components

# Ticket
[MARS-131](https://kiva.atlassian.net/browse/MARS-131)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.156.2",
+	"version": "2.157.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "kvue",
-			"version": "2.156.2",
+			"version": "2.157.0",
 			"license": "UNLICENSED",
 			"dependencies": {
 				"@babel/eslint-parser": "^7.15.4",
--- a/src/components/Kv/KvLoadingOverlay.vue
+++ b/src/components/Kv/KvLoadingOverlay.vue
@@ -9,9 +9,11 @@
 <script>
 import _throttle from 'lodash/throttle';
 import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
+import getCacheKey from '@/util/getCacheKey';
 
 export default {
 	name: 'KvLoadingOverlay',
+	serverCacheKey: () => getCacheKey('KvLoadingOverlay'),
 	components: {
 		KvLoadingSpinner
 	},
--- a/src/pages/Lend/LendHeader.vue
+++ b/src/pages/Lend/LendHeader.vue
@@ -21,6 +21,7 @@
 
 <script>
 import ViewToggle from '@/components/LoansByCategory/ViewToggle';
+import getCacheKey from '@/util/getCacheKey';
 
 export default {
 	name: 'LendHeader',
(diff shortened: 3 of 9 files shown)


src/pages/Lend/LendHeader.vue
```diff
@@ -49,6 +50,19 @@ export default {
...
+			browseUrl,
+			filterUrl,
+			sidePinnedFilterPadding,
+			hardLeftAlign,
+			sideArrowsPadding
+		} = props;
+		const getValue = value => {
+			return value ? `-${value}` : '';
+		};
+		getCacheKey(`${filterUrl}${getValue(browseUrl)}${getValue(sidePinnedFilterPadding)}${getValue(hardLeftAlign)}${getValue(sideArrowsPadding)}`); // eslint-disable-line max-len
```
mcstover, line 64: It seems like the filterUrl + browserUrl shouldn't be used on this component. Rather they would go on the ViewToggle where they would actually differ? Or, does it need to be the parent (this component) that manages that...

I just wonder if either `filterUrl` or `browserUrl` will contain query params and result in a lot of instances of this being cached...

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.156.2",
+	"version": "2.157.0",
```
dyersituations, line 3: Which `npm` version are you running? This change can be reverted.

src/components/Kv/KvLoadingOverlay.vue
```diff
@@ -9,9 +9,11 @@
 <script>
 import _throttle from 'lodash/throttle';
 import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
+import getCacheKey from '@/util/getCacheKey';
 
 export default {
 	name: 'KvLoadingOverlay',
+	serverCacheKey: () => getCacheKey('KvLoadingOverlay'),
```
dyersituations, line 16: Consider using constants to share strings in these components where they are exact matches, such as `KvLoadingOverlay`, if the name/key match is important here. If the caching is only done based on the `serverCacheKey` and doesn't consider the `name` prop, ignore this comment :).

package-lock.json
edwinf-barreram: I am using version 8.5.0

mcstover: I think this is ok... it will get updated in the background again later...

src/pages/Lend/LendHeader.vue
edwinf-barreram: @mcstover this is a result is that is currently being generated there **_/lend/filter-/lend-by-category-true_**

package-lock.json
dyersituations: Ah this version will be auto-updated to match the package.json version.

src/pages/Lend/LendHeader.vue
mcstover: Ok suppose that's ok. The "/" characters are a little odd but no query params in site!

mcstover (approved):

src/components/Kv/KvLoadingOverlay.vue
edwinf-barreram: Name is being set here because this component does not contains props.

emuvente: @dyersituations The name/key match isn't important here. They are both used together to create the final cache key: `${component.name}::${component.serverCacheKey(props)}`

emuvente (commented):
Looking good! There are a just a couple changes needed

mcstover (approved):

PR #3850, opened 2022-05-31
feat: added loan sector filter VUE-1023

georgebs:
This is a draft yet

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchSectorFilter.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchSectorFilter.vue
@@ -33,7 +33,7 @@ export default {
 	data() {
 		return {
 			displayedSectors: this.sectors,
-			// results: []
+			filterResults: []
 		};
 	},
 	computed: {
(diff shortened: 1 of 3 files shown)


mcstover:
@georgebs This is looking great, you'll need to make an addition [here](https://github.com/kiva/ui/blob/main/src/util/loanSearchUtils.js#L234) to map the selected sectors into the query. Then you're sectors will start to be applied.

georgebs:
Working on the Unit tests now

src/components/Lend/LoanSearch/LoanSearchSectorFilter.vue
```diff
@@ -0,0 +1,59 @@
...
+		items() {
+			return this.displayedSectors.map(c => ({
+				value: c.id.toString(),
+				title: c.name,
+			}));
+		}
+	},
+	methods: {
+		updateSectors(sectors) {
+			this.filterResults = sectors.map(sector => parseInt(sector, 10));
```
mcstover, line 49: Seeing as this isn't used outside of this scope, it could probably just be a local const. Not a blocker to merge.

mcstover (approved):
Let's see how things are going with this in place!

PR #3853, opened 2022-05-31
Core 610 expanding loan bug

roger-in-kiva:


mcstover (approved):
This is a good fix to get things working in a suitable manner albeit with a little degradation to the closing animation state.

So long as things are working well in DEV we should be good to go with this fix as we consider a broader re-work of the page in the near future.

PR #3855, opened 2022-05-31
feat: update query string when filters change

dyersituations:
https://kiva.atlassian.net/browse/VUE-1045

- Gender filter now updates the query string params
- Updated router to allow an optional param to prevent auto-scroll to top when route changes
- Using the browser back/forward buttons also now apply the gender filter


mcstover (approved):
Rad!

PR #3859, opened 2022-06-01
fix: remove cache from loan card controller

edwinf-barreram:


dyersituations:
Just out of curiosity, what issue did this cause?

emuvente (approved):
Thanks! This is needed unfortunately, because the style blocks for the loan cards are not being loaded by the server renderer, because it skips loading the children entirely if the parent is cached.

emuvente:
@dyersituations It's causing a flash of unstyled content on /lend-by-category and /lend-by-category/eco-friendly. Specifically, the spacing and size of any of the loan cards is broken on page load, but then is fixed a moment later when the client hydration takes over and adds the style blocks for the loan cards to the page:

https://user-images.githubusercontent.com/4149025/171469457-d32b7d06-9483-4064-b0fd-ca959c701250.mov

dyersituations:
@emuvente Thanks for the info!

mcstover:
It was def worth a try!

PR #3861, opened 2022-06-02
fix: update seo borrower profile title and description

edwinf-barreram:
# Description
- Update seo borrower profile title and description

# Ticket
[MARS-24](https://kiva.atlassian.net/browse/MARS-24)

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -146,8 +146,16 @@ export default {
 	},
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('-beta', '');
+		let title = '';
+		let description = '';
+
+		if (this.$route.path.includes('lend') || this.$route.path.includes('-beta')) {
+			title = `Lend to ${this?.name} in ${this?.countryName}`;
+			description = `A loan of ${this.loanAmount ?? '0'} ${this.use}`;
+		}
+
 		return {
-			title: this.pageTitle,
+			title,
 			link: [
 				{
 					vmid: 'canonical',
@@ -163,6 +171,11 @@ export default {
 					vmid: 'og:image',
 					content: this.imageShareUrl
 				},
+				{
+					vmid: 'description',
+					name: 'description',
+					content: description,
+				}
 			].concat(this.$appConfig.enableFB ? [
 				{
 					vmid: 'facebook_label',
--- a/src/pages/BorrowerProfile/FundedBorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/FundedBorrowerProfile.vue
@@ -111,6 +111,14 @@ export default {
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('funded', 'lend');
 		return {
+			title: `${this.loan?.name} From ${this.loan?.geocode?.country?.name}'s has been funded`,
+			meta: [
+				{
+					vmid: 'description',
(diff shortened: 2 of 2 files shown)


src/pages/BorrowerProfile/FundedBorrowerProfile.vue
```diff
@@ -111,6 +111,14 @@ export default {
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('funded', 'lend');
 		return {
+			title: `${this.loan?.name} From ${this.loan?.geocode?.country?.name}'s has been funded`,
```
dyersituations, line 114: Do we want fallbacks if these `name` props or the `use` prop below are undefined?

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -146,8 +146,16 @@ export default {
 	},
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('-beta', '');
+		let title = '';
+		let description = '';
+
+		if (this.$route.path.includes('lend') || this.$route.path.includes('-beta')) {
+			title = `Lend to ${this?.name} in ${this?.countryName}`;
```
dyersituations, line 153: Isn't `this` always defined here? If so, you don't need the `this?`.

src/pages/BorrowerProfile/FundedBorrowerProfile.vue
```diff
@@ -111,6 +111,14 @@ export default {
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('funded', 'lend');
 		return {
+			title: `${this.loan.name ?? ''} From ${this.loan.geocode.country.name ?? ''}'s has been funded`,
```
dyersituations, line 114: `this.loan` is always defined for this component?

edwinf-barreram: @dyersituations It is defined as an empty object.

FundedBorrowerProfile vue — kiva 2022-06-06 16-20-49

edwinf-barreram: So, if I remove **_this_** it breaks because the undefined properties

FundedBorrowerProfile vue — kiva 2022-06-06 16-23-54

dyersituations: In that case, you probably want `this.loan.geocode?.country?.name`? Since by default `loan` is an empty object.

dyersituations (commented):
@edwinf-barreram @christian14b @emuvente @mcstover Just FYI, I'm watching changes through the `ui` repo and adding code-based comments and questions. Since I'm not familiar with the other team's backlogs, I'm holding off on actually approving the PRs :).

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -146,8 +146,16 @@ export default {
 	},
 	metaInfo() {
 		const canonicalUrl = `https://${this.$appConfig.host}${this.$route.path}`.replace('-beta', '');
+		let title = '';
+		let description = '';
+
+		if (this.$route.path.includes('lend') || this.$route.path.includes('-beta')) {
+			title = `Lend to ${this.name} in ${this.countryName}`;
```
eddieferrer, line 153: We need to consider the loans anonymization level. If it is full then we should hide the name and loan description like it is done in the computed properties.

emuvente (approved):
This looks good to me!

PR #3862, opened 2022-06-02
Personalized bundle core 586

roger-in-kiva:


config/dev-local.js
```diff
@@ -8,6 +8,9 @@ module.exports = merge(base, devVm, {
 		publicPath: '/',
 		photoPath: 'https://www-dev-kiva-org.freetls.fastly.net/img/',
 		graphqlUri: 'https://marketplace-api.dk1.kiva.org/graphql',
+		enableAnalytics: true,
+		enableSnowplow: true,
+		snowplowUri: 'events.fivetran.com/snowplow/v5qt54ocr2nm',
```
mcstover, line 13: @roger-in-kiva Please set `enableAnalytics` to `false` here. We don't want this on all of the time for all engineers.

PR #3866, opened 2022-06-03
Vue 1052 initialize the page with a sort by option when the query param is present on page load

dyersituations:
https://kiva.atlassian.net/browse/VUE-1052
https://kiva.atlassian.net/browse/VUE-1047

- The sortBy filter can now be initialized via the query param
- Changing the sortBy filter now gets applied to the query params
- Added filter validation against available facets before the search state is stored
- Filter validation ensures that bad values in the query params don't result in empty searches
- Bad values result in default filter values being used instead for the problem query param(s)
- Consolidated all interactions with the search state to `LoanSearchInterface` for easier maintenance
- Added `TODO` placeholders for clarity for where future query params can be added
- While not absolutely necessary due to the simple filter options, pulled gender options from the API
- Added `queryType` to `LoanSearchInterface` as a temp location for that value until we need it in the search state
- Updated `LoanSearchSortBy` to handle selection change via prop (needed for query param initialization)
- Storing uppercase facet values as a minor simplification and optimization for validating search state

image

### DIFF
--- a/.storybook/stories/LoanSearchGenderFilter.stories.js
+++ b/.storybook/stories/LoanSearchGenderFilter.stories.js
@@ -17,8 +17,6 @@ const story = (args = {}) => {
 
 export const Default = story();
 
-export const Both = story({ gender: 'both' });
-
 export const Female = story({ gender: 'female' });
 
 export const Male = story({ gender: 'male' });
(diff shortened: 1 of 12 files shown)


mcstover (approved):
Nice work, full speed ahead!

PR #3867, opened 2022-06-03
fix: set meta information only if it defined

edwinf-barreram:
# Description

- Fix: Show Meta info just when it is defined

# Ticket 
[MARS-146](https://kiva.atlassian.net/browse/MARS-146)

### DIFF
--- a/src/pages/ContentfulPage.vue
+++ b/src/pages/ContentfulPage.vue
@@ -205,23 +205,29 @@ export default {
 		};
 	},
 	metaInfo() {
-		return {
-			title: this.title,
-			meta: [
+		const metaObj = { title: this.title.replace('| Kiva', '') };
+
+		if (this.description) {
+			metaObj.meta = [
 				{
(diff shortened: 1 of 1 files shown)


src/pages/ContentfulPage.vue
```diff
@@ -205,23 +205,29 @@ export default {
 		};
 	},
 	metaInfo() {
-		return {
-			title: this.title,
-			meta: [
+		const metaObj = { title: this.title.replace('| Kiva', '') };
+
+		if (this.description) {
```
dyersituations, line 210: In what situations do we not have `description` and/or `canonicalUrl`?

emuvente: Most of the time there won't be a custom canonical URL, and we're still in the process of creating custom descriptions, so we have many pages at the moment that don't have one.

emuvente (approved):
:+1: Looks good!

PR #3868, opened 2022-06-04
fix: update track event details

christian14b:
This PR updates the track event information for the funded borrower profile experiment.


emuvente (approved):
:+1: thanks for finding this!

PR #3869, opened 2022-06-06
fix: check title undefined value

edwinf-barreram:
# Description

- Fix: check title undefined value

# Ticket 
[MARS-146](https://kiva.atlassian.net/browse/MARS-146)


emuvente (approved):
👍 Thanks for the quick fix!

PR #3872, opened 2022-06-06
feat: fire events when user cancels, shares or gets an error from facebook

christian14b:
This feature tracks the following events in facebook share button:

- Hit `Cancel` button
- Hit `Share` button
- Get an `error` from facebook.

using the `handleFacebookResponse` function.


emuvente (approved):
👍 Looks good

PR #3873, opened 2022-06-06
VUE-1051, VUE-1046, VUE-1044

dyersituations:
https://kiva.atlassian.net/browse/VUE-1051
https://kiva.atlassian.net/browse/VUE-1046

- Sector filter can now be initialized via the query param
- Handles both FLSS/legacy "1,2" and Algolia "Arts~Agriculture" formats
- Algolia format will be replaced with the new FLSS format after page load
- Changing the sector filter now populates the query param
- `KvCheckboxList` selection can now be changed via a prop
- Fixed issue where `updateSearchState` could be comparing the wrong objects

image

Update: https://kiva.atlassian.net/browse/VUE-1044

- Theme filter can now be initialized via the query param
- Handles both FLSS/legacy "1,2" and Algolia "Rural Exclusion~Conflict Zones" formats
- Algolia format will be replaced with the new FLSS format after page load
- Changing the theme filter now populates the query param
- Because the theme ID FLSS filter is not yet on prod, the filtering uses theme name, but the query param uses theme ID
- Added missing `LoanSearchSectorFilter` stories
- Theme names in state now stored in upper case to make comparisons consistent, FLSS appears to use fuzzy matching, so caps don't change results

image

Update: https://kiva.atlassian.net/browse/VUE-1080

- Sector filter now has number of loans fundraising listed in parens, matching other filters

image

### DIFF
--- a/.storybook/stories/KvCheckboxList.stories.js
+++ b/.storybook/stories/KvCheckboxList.stories.js
@@ -12,7 +12,7 @@ const story = (args) => {
 		template: `<kv-checkbox-list
 			:show-select-all="showSelectAll"
 			:items="items"
-			:initial-selected="initialSelected" />`,
+			:selected-values="selectedValues" />`,
 	})
 	template.args = args;
 	return template;
@@ -24,6 +24,6 @@ export const Default = story({ items });
 
 export const ShowSelectAll = story({ items, showSelectAll: true });
 
-export const InitialSelected = story({ items, initialSelected: items.slice(0, 2).map(i => i.value) });
+export const SelectedValues = story({ items, selectedValues: items.slice(0, 2).map(i => i.value) });
 
 export const Disabled = story({ items: items.map(i => ({ ...i, disabled: true })) });
--- a/.storybook/stories/LoanSearchSectorFilter.stories.js
+++ b/.storybook/stories/LoanSearchSectorFilter.stories.js
@@ -1,0 +1,25 @@
+import LoanSearchSectorFilter from '@/components/Lend/LoanSearch/LoanSearchSectorFilter';
+
+export default {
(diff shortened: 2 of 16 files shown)


dyersituations:
@mcstover @georgebs @emuvente @eddieferrer Just FYI, my next changeset for VUE-1044 has also been pushed to this PR. Normally I wouldn't combine PRs, but the changes were building on the previous changeset. The commits are separate, so the individual changesets are still clear. My next work will also directly build on this work.

dyersituations:
@mcstover @georgebs @emuvente @eddieferrer Pushed the work for VUE-1080 (sector filter counts).

eddieferrer (commented):
I checked this out locally and did some testing. It is looking pretty good so far. One thing I noticed is that on changing a checkbox, 2 identical queries are made to `fetchFacets` and because of the way `runFacetsQueries` is written these are sequential blocking queries. 

I think at the very least `runFacetsQueries` should run the queries in parallel. Ideally, the facet filters are combined into a single query

dyersituations:
> I checked this out locally and did some testing. It is looking pretty good so far. One thing I noticed is that on changing a checkbox, 2 identical queries are made to `fetchFacets` and because of the way `runFacetsQueries` is written these are sequential blocking queries.
> 
> I think at the very least `runFacetsQueries` should run the queries in parallel. Ideally, the facet filters are combined into a single query

Good catch, thanks @eddieferrer! The identical calls were to get the different available facets with loan numbers based on the other filters (so same query, different params). I don't know how I missed combining these earlier (there was initially just one), but my latest commit pushed to this PR combines the facet queries into one graphql query with a few filter params.

eddieferrer (approved):
👍

PR #3874, opened 2022-06-07
feat: require deposits on matched loans exp

egoroza:
sets up a/b test for requiring deposits on matched loans on borrower profile and checkout pages


roger-in-kiva (approved):
Looks good!

PR #3875, opened 2022-06-07
fix:  layout of the borrower profile loan progress section

christian14b:
This pr fix the broken layout of `Loan Progress` in `Borrower Profile` caused by the conditional section of the experiment.

### DIFF
--- a/src/components/BorrowerProfile/LoanProgress.vue
+++ b/src/components/BorrowerProfile/LoanProgress.vue
@@ -6,7 +6,7 @@
 			:value="progressPercent * 100"
 		/>
 		<figcaption class="tw-flex">
-			<div v-if="!fundedPage">
+			<template v-if="!fundedPage">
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LoanProgress.vue
```diff
@@ -29,7 +29,7 @@
 						{{ progressPercent | numeral('0%') }} funded
 					</p>
 				</div>
-			</div>
+			</template>
 			<div v-else>
```
eddieferrer, line 33: I think this also needs to be changed from div to template for funded loans.

christian14b: Yeah, I tested that out but it shows everything in the same row while design has two rows for `This loan is fully funded!` and the other section

eddieferrer (approved):

PR #3876, opened 2022-06-07
Fix: add subheader and copy tweaks in funded borrower profile variant

christian14b:
This pr fix showing subheader in `Recommended Borrower` section and make copy tweaks for each row in the `Funded Borrower Profile` variant.


emuvente (approved):
👍

PR #3881, opened 2022-06-08
fix: typo in tag name

emuvente:
This typo causes the 'view loan' button to not render and to not be interactive.
Screen Shot 2022-06-08 at 12 08 30 PM


christian14b (approved):
👍🏾  Good catch!

PR #3882, opened 2022-06-08
feat: add page description to contentful utils

christian14b:
This change adds `pageDescription` and `canonicalUrl` fields to `contentfulObject`.


edwinf-barreram (approved):

emuvente (commented):
Looking good! The same changes should be added to the `processPageContent` function as well, as that's used in a few places still.

emuvente (approved):
👍 Thanks for this fix!

mcstover:
💯

PR #3883, opened 2022-06-09
Vue 1032 add loading state overlay for loan section of the page

dyersituations:
https://kiva.atlassian.net/browse/VUE-1032
https://kiva.atlassian.net/browse/VUE-1072

- Created new `KvSectionModalLoader` component that displays a green `KvLoadingSpinner` centered horizontally slightly below the top of the relative parent, with a white slightly transparent background
- Replaced last CSS section with Tailwind JIT classname
- Fixed constrained filter width in mobile view
- Updated `LoanSearchInterface` to use `Promise.all` for graphql queries that aren't interdependent

Initial load:
image

Filter change:
image

Mobile filters now full width:
image

### DIFF
--- a/src/components/Kv/KvSectionModalLoader.vue
+++ b/src/components/Kv/KvSectionModalLoader.vue
@@ -3,7 +3,7 @@
 		v-if="loading"
 		:aria-label="LOADING_LABEL"
 		class="tw-absolute tw-w-full tw-h-full tw-bg-opacity-[65%] tw-z-1 tw-top-0 tw-left-0"
-		:class="{ [`tw-bg-${bgColor}`]: bgColor, 'tw-rounded': rounded }"
+		:class="[`tw-bg-${bgColor}`, {'tw-rounded': rounded }]"
 	>
 		<div class="tw-absolute tw-top-5 tw-left-1/2 tw--translate-x-1/2">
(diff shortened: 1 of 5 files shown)


src/components/Kv/KvSectionModalLoader.vue
```diff
@@ -0,0 +1,48 @@
...
+		},
+		bgColor: {
+			type: String,
+			default: 'white'
+		},
+		rounded: {
+			type: Boolean,
+			default: false
+		},
+		size: {
```
eddieferrer, line 37: Can we do prop validation here to limit this to the 3 acceptable values `small, medium, large` ?

eddieferrer:
Looks good, we should be able to use this to fully deprecate KvLoadingSpinner.vue in the future 👍

src/components/Kv/KvSectionModalLoader.vue
dyersituations: Good call. I almost added a `validator` but decided against it because it was a simple prop, but you're right that it's best practice. My secret hope is that long-term these validations can be replaced by TypeScript :P. Fixed!

emuvente (approved):
Awesome! Looking forward to seeing the old loading spinner fully deprecated soon hopefully!

PR #3884, opened 2022-06-09
fix: remove redirect routes from site map

edwinf-barreram:
# Description
- Remove redirects from the ui site map

# Ticket
[MARS-135](https://kiva.atlassian.net/browse/MARS-135)

### DIFF
--- a/server/sitemap/generators/static.js
+++ b/server/sitemap/generators/static.js
@@ -6,7 +6,7 @@ const { info } = require('../../util/log');
 function getRoutes(haystack, root = '') {
 	return haystack
 		// Remove excluded routes
-		.filter(route => !route.meta?.excludeFromStaticSitemap || route.hasOwnProperty('redirect'))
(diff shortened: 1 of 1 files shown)


server/sitemap/generators/static.js
```diff
@@ -6,7 +6,7 @@ const { info } = require('../../util/log');
 function getRoutes(haystack, root = '') {
 	return haystack
 		// Remove excluded routes
-		.filter(route => !route.meta?.excludeFromStaticSitemap)
+		.filter(route => !route.meta?.excludeFromStaticSitemap || route.hasOwnProperty('redirect'))
```
emuvente, line 9: This should be inverted, because this filter function includes routes when the return value is `true`:
```
route => !route.meta?.excludeFromStaticSitemap && !route.redirect
```

emuvente (approved):
👍

PR #3885, opened 2022-06-09
style: change heading for sector loans row

christian14b:
Change from `Support these loans for ${sectorName}”` to `“Support these ${sectorName} loans"` for better understanding 

Screen Shot 2022-06-08 at 22 23 11


emuvente (approved):
👍

PR #3887, opened 2022-06-10
Vue 1074 setup and fire the loans shown event on every loan query execution

dyersituations:
https://kiva.atlassian.net/browse/VUE-1074
https://kiva.atlassian.net/browse/VUE-1075
https://kiva.atlassian.net/browse/VUE-1076
https://kiva.atlassian.net/browse/VUE-1077
https://kiva.atlassian.net/browse/VUE-1078
https://kiva.atlassian.net/browse/VUE-1079

- Added loan results/hits analytics
- Added gender filter analytics
- Added sort by analytics
- Added location filter analytics
- Added sector filter analytics
- Added theme filter analytics
- `KvCheckboxList` now emits more change event data to allow analytics to be more detailed

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -147,7 +147,9 @@ export default {
 			totalCount: 0,
 			isLightboxVisible: false,
 			loanSearchState: {},
-			queryType: FLSS_QUERY_TYPE
+			queryType: FLSS_QUERY_TYPE,
+			// Holds comma-separated list of loan IDs from the query results
+			trackedHits: undefined,
 		};
 	},
 	async mounted() {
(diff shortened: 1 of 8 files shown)


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -204,10 +206,27 @@ export default {
...
+
+				// Add analytics event for loans query result
+				this.trackLoans();
 			}
 		});
 	},
 	methods: {
+		trackLoans() {
+			// Check for matching hits, for example when sorting a single page of results
+			const hits = this.loans.map(l => l.id).sort().join();
```
mcstover, line 218: I think we need to preserve the delivered order of the loan ids. Does this sort alter that?

dyersituations: The order delivered to analytics?

mcstover: Correct. It should represent the same order in which the loans are shown on the page

dyersituations: Do we want to send a new event if the order is different by IDs the same? Previously we checked the IDs to ensure there weren't duplicate lists, regardless of order.

dyersituations: Got it, thanks. I made the false assumption that we wanted to check result duplicates in general. Updated!

mcstover (approved):

mcstover (approved):

PR #3891, opened 2022-06-10
feat: add tag to UI_REVISION page, add tag as release param in Sentry

eddieferrer:


eddieferrer:
I thought this would be a more convenient way to find out which tag is currently live on prod

emuvente (approved):
Nice! That makes total sense to me

PR #3892, opened 2022-06-13
feat: change UI fetch implementations

christian14b:
Replace `isomorphic-fetch/node-fetch` with `make-fetch-happen` in server side and `whatwg-fetch` in client side to improve performance by using faster streams, keeping connections alive, and pooling requests.


emuvente (approved):
👍 Looks good!

edwinf-barreram (approved):
LGTM 👍

PR #3894, opened 2022-06-13
Core 609 bp complete loan exp

roger-in-kiva:


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -500,9 +510,17 @@ export default {
...
 		unreservedAmount(newValue, previousValue) {
 			// set initial selected value for sub 25 loan if shown
-			if (newValue !== previousValue && previousValue === '' && newValue < 25) {
+			if (this.completeLoan && this.isBetween25And100) {
+				this.selectedOption = Number(this.unreservedAmount).toFixed();
+			} else if (newValue !== previousValue && previousValue === '' && newValue < 25) {
 				this.selectedOption = parseInt(newValue, 10);
 			}
+		},
+		isCompleteLoanActive() {
```
mcstover, line 519: This should be renamed so it doesn't share the name with an existing property.

roger-in-kiva: That is the watcher of existing computed property.

mcstover: 🤦

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -438,6 +442,12 @@ export default {
 				loanId: this.loanId,
 			}).then(() => {
 				this.isAdding = false;
+				if (this.isCompleteLoanActive) {
+					// eslint-disable-next-line max-len
+					this.$kvTrackEvent('Borrower profile', 'Complete loan', 'click-amount-left-cta', this.loanId, this.selectedOption);
+				} else {
+					this.$kvTrackEvent('Lending', 'Add to basket', this.ctaButtonText);
```
mcstover, line 449: This event should fire for ALL add-to-basket actions regardless of loan share size.

roger-in-kiva: Nice, fixed!

mcstover (approved):

PR #3895, opened 2022-06-13
Vue 1033 implement loan pager component

dyersituations:
https://kiva.atlassian.net/browse/VUE-1033
https://kiva.atlassian.net/browse/VUE-1059

- Implemented new `KvPager` component
- The existing `KvPagination` used the Vue router directly, both accessing the query params and pushing routes, so I created a new component without that dependency
- The new pager component is entirely tailwind and has the new arrow icons, but it retains the existing behavior (I fixed one edge case as well from the existing component)
- Added `pageOffset` and `pageLimit` to `loanSearchState`
- There are some conversions between offset and page number in the areas that need page numbers (FLSS + pager), but the state holds the more generic offset and limit values
- Created `numberUtils`, since I kept needing a simple method to check if a value is a number

image

image

### DIFF
--- a/.storybook/stories/KvPager.stories.js
+++ b/.storybook/stories/KvPager.stories.js
@@ -1,0 +1,43 @@
+import KvPager from '@/components/Kv/KvPager';
+
+export default {
+	title: 'Kv/KvPager',
+	component: KvPager,
+};
+
+const story = (args) => {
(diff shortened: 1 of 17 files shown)


mcstover (approved):
We can follow up with updates to KvPagination at a later date.

PR #3896, opened 2022-06-14
feat: activate redirect experiment to split traffic between lend filt…

mcstover:
…er pages

### DIFF
--- a/src/graphql/query/lendFilterPage.graphql
+++ b/src/graphql/query/lendFilterPage.graphql
@@ -15,10 +15,4 @@ query lendFilterPage($basketId: String) {
 			}
 		}
 	}
-	general {
-		lendFilterFlssExp: uiExperimentSetting(key: "lend_filter_flss_mock") {
-			key
-			value
-		}
-	}
 }
--- a/src/pages/Lend/Filter/LendFilterPage.vue
+++ b/src/pages/Lend/Filter/LendFilterPage.vue
@@ -103,8 +103,11 @@
 import _get from 'lodash/get';
 import _map from 'lodash/map';
 import _forEach from 'lodash/forEach';
-
-import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
+import {
+	getExperimentSettingAsync,
+	getExperimentSettingCached,
+	trackExperimentVersion
+} from '@/util/experimentUtils';
 import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
 
 // Algolia Imports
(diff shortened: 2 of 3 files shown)


src/graphql/query/lendFilterPage.graphql
```diff
@@ -15,10 +15,4 @@ query lendFilterPage($basketId: String) {
 			}
 		}
 	}
-	general {
```
mcstover, line 18: EXP setting fetch has moved to new async function in the preFetch promise chain

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -134,6 +137,30 @@ import lendFilterPageQuery from '@/graphql/query/lendFilterPage.graphql';
...
+	const eligibleQueryParams = ['page', 'sortBy', 'gender', 'registration'];
+	const queryParamKeys = Object.keys(route?.query);
+	const allowedSorts = ['expiringSoon', 'loanAmountDesc', 'loanAmount', 'popularity'];
+	// eligible by default, no params is also eligible
+	let isEligible = true;
+
+	queryParamKeys.forEach(queryParam => {
+		// ensure any query params are eligible
+		if (!eligibleQueryParams.includes(queryParam)) {
+			isEligible = false;
```
dyersituations, line 153: Not a huge performance difference in this situation, but you could use a short circuited loop.

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -182,23 +209,18 @@ export default {
...
-
-		if (lendFilterFlssExp.version && lendFilterFlssExp.version !== 'unassigned') {
-			if (lendFilterFlssExp.version === 'b') {
-				this.hideNonFlssFilters = true;
-				this.initiallyExpandedFilters = true;
+		if (isFLSSEligible(this.$route)) {
+			const { enabled } = getExperimentSettingCached(this.apollo, lendFilterRedirectEXP);
+			if (enabled) {
+				// this method will get the version from the apollo cache
+				trackExperimentVersion(
```
dyersituations, line 216: Is there a reason we track this here instead of on the FLSS filter page?

mcstover: This event will fire for visitors in the control group of the experiment with `a` as the "version" whereas the event on our new page will fire with `b` as the version. This page will redirect before it's loaded in the browser if the are in the experiment group and the event fires in the client.

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -134,6 +137,30 @@ import lendFilterPageQuery from '@/graphql/query/lendFilterPage.graphql';
 
 import lendFilterExpMixin from '@/plugins/lend-filter-page-exp-mixin';
 
+const lendFilterRedirectEXP = 'lend_filter_flss_v1';
+
+function isFLSSEligible(route = {}) {
+	// check route for eligibility
+	const eligibleQueryParams = ['page', 'sortBy', 'gender', 'registration'];
```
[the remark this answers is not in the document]
dyersituations: Thanks for the info. We already are mapping the Algolia names to IDs for `sector` and `attribute`.

src/pages/Lend/Filter/LendFilterPage.vue
dyersituations: Ah, so we're tracking eligibility regardless.

dyersituations (approved):

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -134,6 +137,30 @@ import lendFilterPageQuery from '@/graphql/query/lendFilterPage.graphql';
 
 import lendFilterExpMixin from '@/plugins/lend-filter-page-exp-mixin';
 
+const lendFilterRedirectEXP = 'lend_filter_flss_v1';
```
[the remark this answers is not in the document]
mcstover: We'll walk through this setup together soon.

PR #3900, opened 2022-06-14
feat: add selected and deseleted notation on filter click events

mcstover:


src/components/Lend/LoanSearch/LoanSearchLocationFilter.vue
```diff
@@ -88,7 +88,19 @@ export default {
...
+			if (wasSelectAll) {
+				appliedState = values.length ? 'select-all' : 'deselect-all';
+			} else {
+				appliedState = values.includes(changed) ? 'selected' : 'deselected';
+			}
+
+			this.$kvTrackEvent?.(
+				'Lending',
+				'click-location-filter',
+				wasSelectAll ? `Region: ${region}` : `Country: ${changed}`,
```
mcstover, line 101: I've "enhanced" these values to notate if the click was on a region vs a country. There is also a slight difference in the `appliedState`.

dyersituations (approved):

PR #3905, opened 2022-06-16
feat: results per page component

dyersituations:
https://kiva.atlassian.net/browse/VUE-1053
https://kiva.atlassian.net/browse/VUE-1093
https://kiva.atlassian.net/browse/VUE-1094

- Created new `KvResultsPerPage` component modeled after the existing Algolia component
- Couldn't use the old component because it was a wrapped Algolia component
- Resets to the first page when results per page changed, like existing component
- Added behavior of scrolling back to the top of the page when results per page changed
- We can remove the scroll behavior, but I thought the UX made sense, since the pager also scrolls
- Fixed potential edge case in pager
- Added 0 loan state

image

image

### DIFF
--- a/.storybook/stories/KvResultsPerPage.stories.js
+++ b/.storybook/stories/KvResultsPerPage.stories.js
@@ -1,0 +1,22 @@
+import KvResultsPerPage from '@/components/Kv/KvResultsPerPage';
+
+export default {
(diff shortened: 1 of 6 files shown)


mcstover (approved):
Nice!

PR #3906, opened 2022-06-16
fix: do not show reserved loans for upsells on checkout

egoroza:
this PR updates the upsells module experiment to ensure we do not show reserved loans to users. if user A reserves the upsell module loan, other users will see a different loan instead. 

per: https://kivahq.slack.com/archives/C02TDSFQ551/p1655394514010749

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -276,6 +276,7 @@ import UpsellModule from '@/components/Checkout/UpsellModule';
 import updateLoanReservation from '@/graphql/mutation/updateLoanReservation.graphql';
 import * as Sentry from '@sentry/vue';
 import _forEach from 'lodash/forEach';
+import { isLoanFundraising } from '@/util/loanUtils';
 import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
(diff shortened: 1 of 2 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -792,7 +793,9 @@ export default {
 				query: upsellQuery,
 				fetchPolicy: 'network-only',
 			}).then(({ data }) => {
-				this.upsellLoan = data?.lend?.loans?.values[0];
+				const loans = data?.lend?.loans?.values;
+				// Temp solution so we don't show reserved loans on upsell
+				this.upsellLoan = loans.filter(loan => isLoanFundraising(loan))[0] || data?.lend?.loans?.values[0];
```
mcstover, line 798: I would think the OR condition here should actually be an empty object or something that can turn off the visual result. If the first check doesn't pass that means you'd be showing a loan with some funny state...

egoroza: ya good point - also added another conditional to the `v-if` statement to check for this for the upsell module

mcstover (approved):

emuvente (approved):
👍

roger-in-kiva (approved):
Nice job!

PR #3909, opened 2022-06-16
feat: preconnect to fastly

christian14b:
This pr creates a `preconnect` `rel` to prioritize `fastly` connections in the app.


edwinf-barreram (approved):
👍

PR #3911, opened 2022-06-16
fix: update layout of mobile vs desktop filter interface

mcstover:


src/pages/Lend/LoanSearchPage.vue
```diff
@@ -1,33 +1,31 @@
 <template>
 	<www-page id="lend-filter-alpha">
 		<article class="tw-bg-secondary tw-relative tw-pt-6">
-			<div>
```
mcstover, line 4: I really just removed this extraneous div. No other changes.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -15,15 +18,31 @@
 						{{ null }} <!-- Hide title text -->
 					</template>
 					<loan-search-filter
+						style="min-width: 285px;"
```
dyersituations, line 21: The two `min-width` entries here could be `tw-min-w-[285px]` or are we preferring the inline style in this case?

dyersituations (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: We prefer inline styles for items that don't have an automated tailwind unit.

PR #3912, opened 2022-06-16
Core 609 bp complete loan exp

roger-in-kiva:


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -517,7 +517,8 @@ export default {
 		},
 		isCompleteLoanActive() {
 			if (this.isCompleteLoanActive && this.completeLoanView) {
-				this.$kvTrackEvent('Borrower profile', 'Complete loan', 'view-amount-left-cta', null, null);
+				// eslint-disable-next-line
+				this.$kvTrackEvent('Borrower profile', 'Complete loan', 'view-amount-left-cta', this.loanId, this.selectedOption);
```
mcstover, line 521: The 5th parameter must be a whole Int. Does `selectedOption` fit that type?

mcstover (approved):

src/components/BorrowerProfile/LendCta.vue
mcstover: Approved if so...otherwise, you'll need to do something else as I think the value will be removed...

PR #3913, opened 2022-06-16
feat: zero loans reset analytics

dyersituations:
https://kiva.atlassian.net/browse/VUE-1095

- Added missing analytics to zero loans reset
- Added `h3` to match newer styling

image


mcstover (approved):
I don't see why we would need optional chaining on `kvTrackEvent`...there are tons of calls to that method across the code base as it's installed globally...

dyersituations:
> I don't see why we would need optional chaining on `kvTrackEvent`...there are tons of calls to that method across the code base as it's installed globally...

Unit testing + storybook. We could also mock it, if preferred.

PR #3916, opened 2022-06-16
feat: reset page offset when filtering, fixed analytics URL

dyersituations:
https://kiva.atlassian.net/browse/VUE-1092
https://kiva.atlassian.net/browse/VUE-1096

- Fixed issue where analytics URL wasn't updated when visible results matched previous search
- Fixed issue where changing a filter to reduce results would result in empty page if user was on page beyond new results
- Updated behavior to reset to first page when a new filter gets changed, like lend page

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -82,7 +82,7 @@
 					:limit="loanSearchState.pageLimit"
 					:total="totalCount"
 					:offset="loanSearchState.pageOffset"
-					@page-changed="handleUpdatedFilters"
+					@page-changed="handlePageChange"
 				/>
 				<kv-results-per-page
 					:selected="loanSearchState.pageLimit"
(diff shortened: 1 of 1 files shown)


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -290,11 +290,15 @@ export default {
 			updateSearchState(this.apollo, filters, this.allFacets, this.queryType, this.loanSearchState);
 		},
 		handleUpdatedFilters(filters) {
-			this.updateState({ ...this.loanSearchState, ...filters });
+			this.updateState({ ...this.loanSearchState, ...filters, pageOffset: 0 });
```
mcstover, line 293: Is this always 0 or should it be set based on current state?

dyersituations: The `handleUpdatedFilters` is called when a filter/sorting changes, so it resets to the first page. This aligns with the previous behavior and ensures the user never has a page beyond the newly filtered results. Alternatively we could ensure the page is only changed to the last page available when filters change and shrink the results, but I noticed that we used to reset, and that's a simpler approach.

mcstover: Ok cool, that makes sense!

mcstover (approved):

PR #3918, opened 2022-06-17
create filter chips VUE-1042

georgebs:
WIP

### DIFF
--- a/src/components/Kv/KvChipClassic.vue
+++ b/src/components/Kv/KvChipClassic.vue
@@ -1,0 +1,27 @@
+<template>
+	<button class="tw-rounded tw-bg-tertiary tw-p-1 tw-uppercase tw-text-small tw-shrink-0">
+		<slot></slot> X
+	</button>
+</template>
+
+<script>
+export default {
+	name: 'KvChipClassic',
+	props: {
+		// eslint-disable-next-line vue/require-default-prop
+		items: {
+			type: Array,
+			required: false,
+		},
+	},
+	methods: {
+		onChange(event) {
+			this.$emit('change', event);
+		},
+	}
+};
+</script>
+
+<style>
+
+</style>
--- a/src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
@@ -1,0 +1,111 @@
+<template>
+	<div class="tw-flex tw-flex-wrap tw-gap-1.5 tw-my-.5 tw-mb-1">
+		<div v-for="(item, i) in items" :key="i" class="tw-inline-block">
+			<kv-chip-classic :value="item.value">
+				{{ item.name }}
+			</kv-chip-classic>
+		</div>
+	</div>
+</template>
+
+<script>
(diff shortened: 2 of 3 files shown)


src/components/Kv/KvChipClassic.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+	<button class="tw-rounded tw-bg-tertiary tw-p-1 tw-uppercase tw-text-small tw-shrink-0">
+		<slot></slot> X
+	</button>
+</template>
+
+<script>
+export default {
+	name: 'KvChipClassic',
+	props: {
```
mcstover, line 10: If you're using slots to pass the visible text I think you can skip passing down any props.

src/components/Kv/KvChipClassic.vue
```diff
@@ -0,0 +1,27 @@
...
+		// eslint-disable-next-line vue/require-default-prop
+		items: {
+			type: Array,
+			required: false,
+		},
+	},
+	methods: {
+		onChange(event) {
```
mcstover, line 18: I would convert this to `onClick` for the method name, `click` in the emit below and then you'll want to add `@click="onClick"` to your `button` element above.

src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
```diff
@@ -0,0 +1,111 @@
+<template>
+	<div class="tw-flex tw-flex-wrap tw-gap-1.5 tw-my-.5 tw-mb-1">
+		<div v-for="(item, i) in items" :key="i" class="tw-inline-block">
+			<kv-chip-classic :value="item.value">
```
mcstover, line 4: You won't need to pass a prop down here but you'll need to listen for `@click` and call a new method (to be defined below) with your full "item" object. This new method "handleChipClick" will receive your "item" and can then emit the update event and include a tracking event to represent the interaction.

src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
```diff
@@ -0,0 +1,111 @@
...
+         * }
+         */
+		allFacets: {
+			type: Object,
+			default: () => {}
+		}
+	},
+	data() {
+		return {
+			items: []
```
mcstover, line 87: I think you should convert this to a "computed" property. That way `items` will be reactive. Within your new `items` computed property you can create a sequence of steps to aggregate and combine data for "active" filters and return an array of objects representing them. Each item will contain all the info you need to pass back up to the updated method and the display name for the chip.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -37,6 +37,9 @@
 					</template>
 				</kv-lightbox>
 			</div>
+			<div>
+				<loan-search-filter-chips :loan-search-state="loanSearchState" :all-facets="allFacets" />
```
mcstover, line 41: You'll need to add `@updated="handleUpdatedFilters"` to this instance and ensure the data associated with the `updated` event emitted from LoanSearchFilterChips.vue is formatted correctly and contains the correct data fields.

PR #3921, opened 2022-06-17
fix: exclude popularity sort as eligible for the experiment

mcstover:


src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -143,7 +143,7 @@ function isFLSSEligible(route = {}) {
 	// check route for eligibility
 	const eligibleQueryParams = ['page', 'sortBy', 'gender', 'registration'];
 	const queryParamKeys = Object.keys(route?.query);
-	const allowedSorts = ['expiringSoon', 'loanAmountDesc', 'loanAmount', 'popularity'];
+	const allowedSorts = ['expiringSoon', 'loanAmountDesc', 'loanAmount']; // 'popularity'
```
dyersituations, line 146: Do we also need to remove "Recommended" from the available alpha sorts in the UI?

dyersituations (approved):

src/pages/Lend/Filter/LendFilterPage.vue
mcstover: Recommended is the default FLSS sort. I never actually went through mapping popularity to recommended which was good b/c we determined that wouldn't be what we'll do. Popularity is distinct. We do have a switch to flip to "popularity" when the query type isn't FLSS but I don't see a way that would flip at this time.

mcstover: Also, "Recommended" is the Label mapped from "personalized" for FLSS.

dyersituations: Oh right, recommended maps to personalized, not popularity.

PR #3923, opened 2022-06-17
feat: increase number of allowed webpack chunks

christian14b:
This pr changes `webpack.client.base.conf.js` configuration file to ensure that modules are only loaded when they are actually used by the page.


edwinf-barreram (approved):
👍

PR #3925, opened 2022-06-20
fix: create specific route to handle theme_refugees/displaced path se…

mcstover:
…gment snafu

### DIFF
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -137,6 +137,12 @@ module.exports = function liveLoanRouter(cache) {
 			await redirectToUrl('filter', cache, req, res);
 		});
 	});
+	// Filter URL Router (refugees/displaced)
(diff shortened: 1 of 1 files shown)


server/live-loan-router.js
```diff
@@ -137,6 +137,12 @@ module.exports = function liveLoanRouter(cache) {
 			await redirectToUrl('filter', cache, req, res);
 		});
 	});
+	// Filter URL Router (refugees/displaced)
+	router.use('/f/theme_refugees/displaced/url/:offset(\\d{0,})', async (req, res) => {
+		await tracer.trace('live-loan.filter.redirectToUrl', { resource: req.path }, async () => {
+			await redirectToUrl('filter', cache, req, res);
```
emuvente, line 143: The handler will be expecting req.params.id to be defined, so we should set `req.params.id = 'theme_refugees/displaced';` before calling redirectToUrl.

Also FYI, we'll need to include ef9049e in the hotfix in order to use `tracer.trace`.

PR #3936, opened 2022-06-21
feat: always show certain themes

dyersituations:
https://kiva.atlassian.net/browse/VUE-1071

- 9 themes/attributes are now always shown, regardless of loans fundraising

image


mcstover (approved):
Cool thank you!

PR #3937, opened 2022-06-21
feat: add basket state + keep it up to date after add to basket actions

mcstover:


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -222,12 +225,41 @@ export default {
...
+				query: userIdQuery
+			}).then(() => {
+				return client.query({
+					query: itemsInBasketQuery
+				});
+			});
 		},
 	},
+	created() {
+		const basketId = this.cookieStore.get('kvbskt');
```
dyersituations, line 239: I know we use this everywhere, but `kvbskt` seems like a good constant to share.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -222,12 +225,41 @@ export default {
...
+
+		try {
+			const userIdData = this.apollo.readQuery({
+				query: userIdQuery
```
dyersituations, line 243: I haven't dug into this usage of `apollo` yet, I'm assuming this will pull from the cache filled by `preFetch`?

dyersituations (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: Some type of global, but also update access could be useful and worth exploring more. Def an improvement I'd like to see in the CMS page server. Ultimately, this data exists in it's most up to date form in the cookie does reside in the cookie and at this point the ability to access on both the server and client this way works pretty well.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: Correct. read the same query that pushed data to the cache to get that data outside of the preFetch method.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -222,12 +225,41 @@ export default {
...
+			this.apollo.watchQuery({
+				query: itemsInBasketQuery,
+				variables: { basketId },
+			}).subscribe({
+				next: ({ data }) => {
+					this.itemsInBasket = data?.shop?.basket?.items?.values.map(item => item.id) ?? [];
+				},
+			});
+		} catch (e) {
+			logReadQueryError(e, 'LoanSearchInterface itemsInBasketQuery');
```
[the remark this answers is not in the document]
mcstover: This is the approach we settled on to help us detect queries that cause invariant errors due to cache collisions!

src/components/Lend/LoanSearch/LoanSearchInterface.vue
dyersituations: That's true. I also just meant the string `kvbskt` itself in code in a global constant.

PR #3939, opened 2022-06-22
fix: redirect to lend/id when loan status is not fundraising

edwinf-barreram:
# Description
- Redirect to lend/id when loan status is not fundraising

# Ticket
[MARS-153](https://kiva.atlassian.net/browse/MARS-153)


eddieferrer (approved):

emuvente (approved):
👍 That should do it!

PR #3940, opened 2022-06-22
fix: fixed styles on thanks social share page

edwinf-barreram:


mcstover (approved):

emuvente (approved):
Thanks for the quick fix!

PR #3942, opened 2022-06-22
Funded borrower profile: Improvements

christian14b:
Add the following fixes and interaction in the funded borrower profile page experiment:

- Add a dollar sign into the button: “Lend $25 now”
- When a user clicks on a “Lend now” CTA, the button should go into a loading state while the loan is added to the basket. Then the CTA should change to say ‘Checkout now’. 

Screen Shot 2022-06-22 at 11 40 01


mcstover (approved):

emuvente (approved):
👍 Looks good!

PR #3943, opened 2022-06-22
fix: error handling for upsell loans

egoroza:
fixes an issue on upsell loans where if a user tries to add a reserved loan, it will pull a new non-reserved loan for them as the upsell loan in their cart

Screen Shot 2022-06-22 at 1 41 25 PM


mcstover (approved):
Looks good!

roger-in-kiva (approved):

PR #3945, opened 2022-06-22
Vue 1112 create loan channel mapping to flss

dyersituations:
https://kiva.atlassian.net/browse/VUE-1112

- Added new mappings for `flssLoanSearch`
- Checked all channels in mapping file, all channels in category header, and all high traffic channels via pageviews in Looker
- Only found one missing/incorrect URL in the mapping file, `ending-soon` had the wrong URL
- Cleaned up eslint disable rule usage


src/plugins/loan-channel-query-map.js
```diff
@@ -9,25 +11,29 @@ export default {
...
+					algoliaParams: 'gender=female',
+					flssLoanSearch: { gender: 'female' },
 				},
 				{
 					id: 8,
 					url: 'agriculture',
 					queryParams: 'status=fundRaising&riskRating=0,5&sector=1',
-					algoliaParams: 'sector=Agriculture'
+					algoliaParams: 'sector=Agriculture',
+					flssLoanSearch: { sector: [1] },
```
mcstover, line 22: I think we need a key of `sectorId` when we're using the `id`.

src/plugins/loan-channel-query-map.js
```diff
@@ -170,37 +186,43 @@ export default {
...
+					algoliaParams: 'sector=Education',
+					flssLoanSearch: { sector: [15] },
 				},
 				{
 					id: 54,
 					url: 'trending-now',
 					queryParams: 'status=fundRaising&riskRating=0,5',
-					algoliaParams: ''
+					algoliaParams: '',
+					flssLoanSearch: {},
```
mcstover, line 204: This will need to be updated to the `popularity` sort when it's available in FLSS.  It's not in the `queryParams` above b/c it's the default. We won't be able to include this category in the exp this go round.

src/plugins/loan-channel-query-map.js
```diff
@@ -170,37 +186,43 @@ export default {
...
+					algoliaParams: 'sector=Agriculture',
+					flssLoanSearch: { sector: [1] },
 				},
 				{
 					id: 56,
 					url: 'loans-with-research-backed-impact',
 					queryParams: 'status=fundRaising&riskRating=0,5',
-					algoliaParams: ''
+					algoliaParams: '',
+					flssLoanSearch: {},
```
mcstover, line 218: This one is based on a different sort param for impact that isn't yet compatible with FLSS so it can't be included either.

src/plugins/loan-channel-query-map.js
```diff
@@ -224,119 +247,108 @@ export default {
...
+					algoliaParams: 'attributes=Clean%20Energy',
+					flssLoanSearch: { theme: ['CLEAN ENERGY'] },
 				},
 				{
 					id: 66,
 					url: 'lenders-like-you-supported-these-loans',
 					queryParams: 'status=fundRaising&riskRating=0,5',
-					algoliaParams: ''
+					algoliaParams: '',
+					flssLoanSearch: {},
```
mcstover, line 258: This is a "reccommened" row that would come out of the ML service instead of FLSS, so it's not eligible either.

src/plugins/loan-channel-query-map.js
dyersituations: Oops, good catch. Updated!

src/plugins/loan-channel-query-map.js
dyersituations: Removed.

src/plugins/loan-channel-query-map.js
dyersituations: Looks like some of these are missing some params for Algolia. Removed.

src/plugins/loan-channel-query-map.js
dyersituations: Removed.

src/plugins/loan-channel-query-map.js
```diff
@@ -397,55 +407,54 @@ export default {
...
 					queryParams: 'status=fundRaising&riskRating=0,5&state=CA&city_state=San Jose,CA&distributionModel=direct',
 					algoliaParams: ''
 				},
 				{
 					id: 43,
 					url: 'super-power-a-woman-on-kiva',
 					queryParams: 'status=fundRaising&gender=female&riskRating=0,5&distributionModel=field_partner',
-					algoliaParams: 'gender=female'
+					algoliaParams: 'gender=female',
+					flssLoanSearch: { gender: 'female' },
```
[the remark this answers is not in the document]
dyersituations: Removed.

mcstover (approved):

PR #3948, opened 2022-06-23
fix: add missing word loan

edwinf-barreram:


emuvente (approved):
👍 Looks good! Thanks!

PR #3949, opened 2022-06-23
fix: move exp tracking to mounted hook and remove eligibility gating …

mcstover:
…around analytics on lend/filter

### DIFF
--- a/src/pages/Lend/Filter/LendFilterPage.vue
+++ b/src/pages/Lend/Filter/LendFilterPage.vue
@@ -208,20 +208,6 @@ export default {
 
 		// Get Lend Filter Exp version
 		this.getLendFilterExpVersion();
-
-		if (isFLSSEligible(this.$route)) {
-			const { enabled } = getExperimentSettingCached(this.apollo, lendFilterRedirectEXP);
-			if (enabled) {
-				// this method will get the version from the apollo cache
-				trackExperimentVersion(
-					this.apollo,
-					this.$kvTrackEvent,
-					'Lending',
-					lendFilterRedirectEXP,
-					'EXP-VUE-1061-June2022'
-				);
-			}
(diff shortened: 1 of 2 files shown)


src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -208,20 +208,6 @@ export default {
 
 		// Get Lend Filter Exp version
 		this.getLendFilterExpVersion();
-
-		if (isFLSSEligible(this.$route)) {
```
mcstover, line 212: This extra check around the tracking call is the primary issue as the Algolia components immediately push in the `popularity` sort param causing the url read to be in-eligible by the time this is caused.

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -70,7 +70,7 @@ export default {
 				});
 		}
 	},
-	created() {
+	mounted() {
```
dyersituations, line 73: Why was this change needed?

src/pages/Lend/Filter/LendFilterPage.vue
```diff
@@ -316,6 +302,20 @@ export default {
 
 		// update global lend filter experiment setting
 		this.updateLendFilterExp();
+
+		// Check cached for lend/filter-alpha experiment state and track if present
```
dyersituations, line 306: No need to check `isFLSSEligible`?

dyersituations (approved):

src/pages/Lend/LoanSearchPage.vue
mcstover: Analytics calls can only execute in the client, so it's a more explicit location to attempt the call. More preference than necessity.

PR #3956, opened 2022-06-28
Vue 1113 activate loan channel experiment in channel control component

dyersituations:
https://kiva.atlassian.net/browse/VUE-1113
https://kiva.atlassian.net/browse/VUE-1115

- Created GraphQL query that pulls FLSS data and needed other API data for a loan channel
- Added switch between new and old data in required places: `preFetch`, `created`, and `activateLoanChannelWatchQuery`
- `loanChannelUtils` handles new toggling and functionality covered with unit tests
- Created experiment settings in dev: 
```
{"name":"LoanChannelFlssQuery","enabled":true,"startTime":"2022-06-22T23:59","endTime":"2025-06-22T23:59","control":{"key":"a","name":"Control"},"distribution":{"a":0.5,"b":0.5},"population":0.05}
```

-----

Update: pushed https://kiva.atlassian.net/browse/VUE-1114

- Added experiment tracking when the FLSS channel experiment is enabled

-----

Updated: pushed https://kiva.atlassian.net/browse/VUE-1129

- Added `distributionModel` to the FLSS channel mappings


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -336,65 +338,64 @@ export default {
 		}
 	},
 	apollo: {
-		preFetch(config, client, args) {
+		preFetch(_config, client, args) {
```
eddieferrer, line 341: Whats the thinking behind renaming these function arguments to add "_" ? Renaming `loan` to `loans` in remaining loans above is ok, but renaming this preFetch functions makes this one different than all the other components. Personally when I see the underscore I immediately think lodash.... but maybe thats just me

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -336,65 +338,64 @@ export default {
...
-			});
+				await preFetchChannel(
+					client,
+					// Access map directly since SSR doesn't have mixins available
+					loanChannelQueryMapMixin.data().loanChannelQueryMap,
+					targetedLoanChannelURL,
+					// Build loanQueryVars since SSR doesn't have same context
+					{ ids: [targetedLoanChannelID], limit: loansPerPage, offset: fromUrlParams(pageQuery).offset }
+				);
+			}).catch(e => logReadQueryError(e, 'LoanChannelCategoryControl preFetch loanChannelPageQuery'));
```
eddieferrer, line 362: I don't think we can use `logReadQueryError` in preFetch, since the error will just get swallowed up server side -- but I could be wrong

dyersituations: The error gets logged to the server-side logs (verified locally). I can remove this if we don't actually access these logs. My assumption was that all of the logs were somewhere in a cloud logging instance.

src/pages/Lend/LoanChannelCategoryControl.vue
dyersituations: Adding an underscore prefix is a relatively common practice for designating unused parameters. For example, the Vetur VSCode extension has this as part of its linting rules by default, though I know we don't have that as part of our eslint configs. I can revert these underscores if that's not a practice we want to employ here. It's just something I've seen/used for a while, so I just do it by default when I come across an unused parameter.

eddieferrer:
This seems to generate a server error for a user who does not yet have a basket and gets assigned to the experiment: 

Error Stack (with some console messages thrown in):
`{"meta":{},"level":"info","message":"session fetch: 9ms"}
{"meta":{},"level":"info","message":"fragment fetch: 32ms"}
{"meta":{},"level":"info","message":"fragment fetch: 29ms"}
{"meta":{},"level":"info","message":"session fetch: 669ms"}
preFetchChannel { gender: 'female' } { ids: [ 5 ], limit: 12, offset: 0 }
experimenActive
fetchLoanChannel { gender: 'female' } { ids: [ 5 ], limit: 12, offset: 0 }
loanSearchState { gender: 'female' }
{"meta":{},"level":"log","message":"data pre-fetch: 6804ms"}
[vue-router] <router-link>'s event prop is deprecated and has been removed in Vue Router 4. Use the v-slot API to remove this warning: https://next.router.vuejs.org/guide/migration/#removal-of-event-and-tag-props-in-router-link.
loanSearchState { gender: 'female' }
[Vue warn]: Error in render: "TypeError: Cannot read property 'id' of null"

found in

---> <LoanChannelCategoryControl> at src/pages/Lend/LoanChannelCategoryControl.vue
       <WwwPage> at src/components/WwwFrame/WwwPage.vue
         <LoanChannelCategoryPage> at src/pages/Lend/LoanChannelCategoryPage.vue
           <App> at src/App.vue
             <Root>
/Users/eddief/kiva/ui/node_modules/vue/dist/vue.runtime.common.dev.js:1895
    throw err
    ^

TypeError: Cannot read property 'id' of null
    at eval (webpack:///./src/pages/Lend/LoanChannelCategoryControl.vue?./node_modules/vue-loader/lib/loaders/templateLoader.js??vue-loader-options!./node_modules/vue-loader/lib??vue-loader-options:301:39)
    at Proxy.renderList (/Users/eddief/kiva/ui/node_modules/vue/dist/vue.runtime.common.dev.js:2640:16)
    at Proxy.render (webpack:///./src/pages/Lend/LoanChannelCategoryControl.vue?./node_modules/vue-loader/lib/loaders/templateLoader.js??vue-loader-options!./node_modules/vue-loader/lib??vue-loader-options:299:29)
    at /Users/eddief/kiva/ui/node_modules/@vue/composition-api/dist/vue-composition-api.common.js:1908:35
    at activateCurrentInstance (/Users/eddief/kiva/ui/node_modules/@vue/composition-api/dist/vue-composition-api.common.js:1858:16)
    at Proxy.$options.render (/Users/eddief/kiva/ui/node_modules/@vue/composition-api/dist/vue-composition-api.common.js:1907:24)
    at VueComponent.Vue._render (/Users/eddief/kiva/ui/node_modules/vue/dist/vue.runtime.common.dev.js:3559:22)
    at resolve (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8444:27)
    at waitForServerPrefetch (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8316:3)
    at renderComponentInner (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8455:3)
    at renderComponent (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8412:5)
    at RenderContext.renderNode (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8323:5)
    at RenderContext.next (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:2611:23)
    at cachedWrite (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:2464:9)
    at renderStringNode$1 (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8530:5)
    at RenderContext.renderNode (/Users/eddief/kiva/ui/node_modules/vue-server-renderer/build.dev.js:8321:5)`

To reproduce set experiment to 100% b variant -- open https://dev-vm-01.kiva.org/lend-by-category/women in an incognito window

eddieferrer:
As an aside, you can add more `flssLoanSearch` queries to the `loan-channel-query-map.js` -- see: `ui/src/pages/Lend/LoanChannelCategoryExperiment.vue` 

For example: `eco-friendly` maps to FLSS: loanTags: [9, 8], single-parents maps to FLSS: loanTags: [17]

dyersituations:
> As an aside, you can add more `flssLoanSearch` queries to the `loan-channel-query-map.js` -- see: `ui/src/pages/Lend/LoanChannelCategoryExperiment.vue`
> 
> For example: `eco-friendly` maps to FLSS: loanTags: [9, 8], `single-parents` maps to FLSS: loanTags: [17]

Interesting, I didn't know about that list of channels, thanks. Luckily I'm not seeing a channel that isn't already in `loan-channel-query-map.js`. Unfortunately FLSS doesn't support `loanTags` filters, so I didn't add mappings to `loan-channel-query-map.js` for FLSS if `loanTags` were included.

eddieferrer:
> > As an aside, you can add more `flssLoanSearch` queries to the `loan-channel-query-map.js` -- see: `ui/src/pages/Lend/LoanChannelCategoryExperiment.vue`
> > For example: `eco-friendly` maps to FLSS: loanTags: [9, 8], `single-parents` maps to FLSS: loanTags: [17]
> 
> Interesting, I didn't know about that list of channels, thanks. Luckily I'm not seeing a channel that isn't already in `loan-channel-query-map.js`. Unfortunately FLSS doesn't support `loanTags` filters, so I didn't add mappings to `loan-channel-query-map.js` for FLSS if `loanTags` were included.

 Good call, forgot that FLSS isn't supporting loanTags yet 👍

dyersituations:
> This seems to generate a server error for a user who does not yet have a basket and gets assigned to the experiment:

I did test the use case of an `undefined` basket before, and I just checked again. Locally I'm not seeing any errors, but maybe my setup is slightly different, since I've been working from `localhost`:
- Launched the ui via `npm run dev` and updated `local.js` config to point at the dev GraphQL
- User isn't logged in since login doesn't work in this configuration
- The basket cookie getter `this.cookieStore.get('kvbskt')` returns `undefined` -> so I'm assuming it's a similar situation to your test case
- The experiment is `version: 'b'`
Is there anything else that needs to align for our test cases to be the same? It's hard to tell from the stacktrace. This is the first time I've worked with the basket, so I could be missing something important.

dyersituations:
> > This seems to generate a server error for a user who does not yet have a basket and gets assigned to the experiment:
> 
> @eddieferrer I did test the use case of an `undefined` basket before, and I just checked again. Locally I'm not seeing any errors, but maybe my setup is slightly different, since I've been working from `localhost`:
> 
> * Launched the ui via `npm run dev` and updated `local.js` config to point at the dev GraphQL
> * User isn't logged in since login doesn't work in this configuration
> * The basket cookie getter `this.cookieStore.get('kvbskt')` returns `undefined` -> so I'm assuming it's a similar situation to your test case
> * The experiment is `version: 'b'`
> 
> Is there anything else that needs to align for our test cases to be the same? It's hard to tell from the stacktrace. This is the first time I've worked with the basket, so I could be missing something important.

src/pages/Lend/LoanChannelCategoryControl.vue
dyersituations: @eddieferrer Good to keep this log then or should I remove?

src/pages/Lend/LoanChannelCategoryControl.vue
dyersituations: @eddieferrer Thoughts on using underscores or should I remove for code consistency?

dyersituations: Removed underscores for consistency.

src/pages/Lend/LoanChannelCategoryControl.vue
dyersituations: Checked all usages of `preFetch`, and we generally don't log errors, so I removed the `catch`. While having a `catch` would be good defensive programming, there are many issues if the connections fail for the UI, so adding a `catch` here doesn't save the app from erroring out in that edge case.

dyersituations:
> This seems to generate a server error for a user who does not yet have a basket and gets assigned to the experiment:

Did extensive testing in the VM setup, and it turns out that this error will occur if FLSS fails to return data. For some reason, it looks like both @eddieferrer and I encountered the same issue with FLSS locally. Once I recreated my VM and dry-run DB, this error longer occurred, but I did encounter the error before recreating the VM environment. I added in a check just in case this edge case occurs in prod, so that the app doesn't crash, but I doubt anyone will encounter this error in prod.

```
firstLoan() {
    // Handle an edge case where a backend error could lead to a null loan
    return this.loans[0] ? [this.loans[0]] : [];
}
```

eddieferrer (approved):
Good sleuthing on that FLSS bug 👍 

Based on the error we were getting I do wonder if FLSS vs legacy have slightly different returns from the API... Does FLSS return `loans: null` and legacy return `loans: []` when there are no loans? That would be my guess since the legacy API call on this page handles no loans without issue....

mcstover (approved):

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -339,62 +347,61 @@ export default {
...
-					query: loanChannelQuery,
-					variables: _merge(
-						{
-							ids: [targetedLoanChannelID],
-							limit: loansPerPage
-						},
-						fromUrlParams(pageQuery)
-					)
-				});
+				await preFetchChannel(
```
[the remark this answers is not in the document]
dyersituations: Return isn't needed because of the `await`, but for consistency with the other chained promises in this `preFetch`, I removed the `await` and returned `preFetchChannel`.

dyersituations:
> Does FLSS return `loans: null` and legacy return `loans: []` when there are no loans?

Good question. Just checked, and both return `[]`. My guess is that the error was more related to the FLSS server not being reachable for whatever reason with the VM setup at the time. Since I don't know how to repro the bad FLSS setup in the VM, testing might have to be done via changing to bad API URLs, or something along those lines. And luckily this issue shouldn't happen in prod, but we should probably do some testing around unreachable servers.

PR #3962, opened 2022-06-29
feat: update copy text for matched loans

egoroza:
update copy text for matched loans on borrower profile page as part of required deposits on matched loans experiment

Screen Shot 2022-06-29 at 3 00 17 PM


emuvente (approved):
👍

PR #3963, opened 2022-06-29
fix: update copy for borrower profile share card

emuvente:
This also adds the anonymization level to the loan use filter to reduce the number of places the level had to be checked.


christian14b (approved):
Nice refactoring!

PR #3965, opened 2022-06-30
fix: add inviter name to shareLink func

christian14b:
This pr changes `shareLink` function inside `SocialShareV2` component to add `inviterName`.


emuvente (approved):
Thanks!

PR #3967, opened 2022-07-01
feat: adding category mouseover effects

galen94:
Adding a "View all loans" button and hover effects when hovering over a category.

### DIFF
--- a/src/components/Categories/MainCategoryTile.vue
+++ b/src/components/Categories/MainCategoryTile.vue
@@ -1,23 +1,25 @@
 <template>
-	<div v-if="tileSize === 'large' || tileSize === 'medium'" class="tw-mb-6">
+	<div v-if="tileSize === 'large' || tileSize === 'medium'" class="tw-mb-6 tw-group">
 		<router-link
-			class="tw-no-underline
-			active:tw-no-underline visited:tw-no-underline
-			hover:tw-no-underline focus:tw-no-underline
-			tw-text-black
-			active:tw-text-black visited:tw-text-black
-			hover:tw-text-black focus:tw-text-black"
+			class="remove-link-decoration"
 			:to="`/lend-by-category/${cleanURL}`"
 			v-kv-track-event="['Lending', 'click-Category', categoryName]"
 		>
-			<kv-responsive-image class="category-image"
-				:images="chooseImage(tileSize)"
-				loading="lazy"
-				:alt="altText"
-			/>
+			<div class="tw-relative">
+				<kv-responsive-image class="category-image"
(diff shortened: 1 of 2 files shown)


src/pages/Categories/CategoriesBeta.vue
```diff
@@ -45,6 +45,20 @@
 					/>
 				</div>
 			</kv-grid>
+			<kv-grid class="tw-bg-secondary tw-pt-5 tw-mb-5 tw-rounded">
```
eddieferrer, line 48: The button should have 48px of padding top and bottom, there's too much space here, I think you need to remove the `tw-pt-5` from this kv-grid

src/pages/Categories/CategoriesBeta.vue
```diff
@@ -45,6 +45,20 @@
...
+				<div class="tw-flex tw-items-center tw-justify-center tw-pt-6 tw-pb-6">
+					<router-link
+						class="tw-text-black
+						active:tw-text-black visited:tw-text-black
+						hover:tw-text-black focus:tw-text-black"
+						to="/lend"
```
eddieferrer, line 54: please add analytics to this link we can use: `['Lending', 'click-view-all', 'View All']`

src/components/Categories/MainCategoryTile.vue
```diff
@@ -40,27 +42,29 @@
...
 		>
 			<div class="tw-flex">
 				<div class="tw-mr-2 tw-flex-none">
-					<kv-responsive-image class="category-image-small"
-						:images="chooseImage(tileSize)"
-						loading="lazy"
-						:alt="altText"
-					/>
+					<div class="tw-relative">
+						<kv-responsive-image class="category-image image-small"
```
eddieferrer, line 54: We try to stick to BEM class naming for CSS -- more info here: http://getbem.com/naming/

For this class -- it would be: `category-image category-image__small`and you can modify your css below to target the `category-image__small` class

src/pages/Categories/CategoriesBeta.vue
galen94: I've been playing around with inspecting the screen for sizing. It seems like tw-pt-6 is good for the padding size of 48px, but I made the button itself smaller so that everything fits Nathan's dimensions. 

Before I push again, there's a nice part of KvButton that allows you to choose the state. When I choose active it's always the secondary color like in Figma, but when I choose default it's white and on hover it's secondary. I like it with the default state because you can actually see the button changing when you hover. When it's always active it seems like it doesn't really work. 

Should I just keep it active to match Figma?

PR #3971, opened 2022-07-05
live loan drawing memory usage optimizations

emuvente:
`node-canvas` has a [known memory leak](https://github.com/Automattic/node-canvas/issues/1974) that occurs when `registerFont` is called multiple times. This PR changes our usage of `registerFont` to only happen on startup to avoid that memory leak. Additionally, this PR sets up a pool of canvases to be used and re-used for drawing loan cards to avoid the memory and cpu cost of creating a new canvas for every request.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.206.4",
+	"version": "2.214.1",
 	"lockfileVersion": 2,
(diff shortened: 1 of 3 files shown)


mcstover (approved):

dyersituations (approved):
Great find! The `canvas` is definitely uses a ton of resources in general.

PR #3976, opened 2022-07-06
fix: fixed copy share link params structure

edwinf-barreram:
# Description

- Add small change to fix copy share link url params structure


emuvente (approved):
Good catch!

PR #3985, opened 2022-07-07
fix: use optimized default make-fetch-happen agent instead of custom agent

emuvente:
Because we were providing a custom agent we weren't getting any of the optimized configurations (specifically keepAlive). We only needed the custom agent for allowing invalid VM certificates via `rejectUnauthorized`, and that can be controlled using [the `strictSSL` option](https://github.com/npm/make-fetch-happen#--optsca-optscert-optskey-optsstrictssl) instead.

### DIFF
--- a/server/util/fetch.js
+++ b/server/util/fetch.js
@@ -1,12 +1,10 @@
 const minipassFetch = require('make-fetch-happen');
-const https = require('https');
 
 module.exports = function fetch(url, options) {
+	const onVm = url.indexOf('vm') > -1;
 	return minipassFetch(url, {
+		// fix request blocked b/c of self-signed certificate on dev-vm.
+		strictSSL: !onVm,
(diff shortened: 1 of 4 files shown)


server/util/fetch.js
```diff
@@ -1,12 +1,10 @@
 const minipassFetch = require('make-fetch-happen');
-const https = require('https');
```
dyersituations, line 2: Can the `https` npm package be removed? I don't see it being used directly elsewhere.

dyersituations (approved):

mcstover (approved):

christian14b (approved):

server/util/fetch.js
emuvente: Yes it can, and should! I'll add a commit to remove it.

PR #3989, opened 2022-07-07
chore: update graphql and related dependencies

eddieferrer:


dyersituations (approved):

mcstover:
Looks like our use of `IntrospectionFragmentMatcher` instead of the other introspectionQuery constants kept use out of the breaking changes path!

PR #3990, opened 2022-07-07
Vue 1116 add show hide reset toggles to filter chips interface

dyersituations:
https://kiva.atlassian.net/browse/VUE-1116

- Added "Show all x filters" and "Hide filters | Reset filters" links
- Modeled general functionality after Algolia page
- Added new tracking as needed
- No unit tests for toggling behavior, since it relies on `clientHeight`, which isn't available in the jsdom, and there doesn't appear to be a good way to mock `clientHeight` for multiple elements at once
- Used a `style` block in order to access tailwind breakpoint and to avoid new tailwind class with `[Xpx]`
- Used lodash throttle, since we don't have a non-lodash implementation yet, and it didn't seem like it was worth the time at the moment to implement the slightly more complex utility custom

Wide screen default single line of chips:

image

Wide screen shown filters:

image

Small screen default 3 lines of chips:

image

Small screen shown filters:

image


src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
```diff
@@ -81,7 +122,7 @@ export default {
 				itemList = [...itemList, ...themeFacets];
 			}
 			// Gender
-			if (loanSearchState.gender !== null) {
+			if (loanSearchState.gender) {
```
mcstover, line 125: Does this result in showing a generic "All" chip for gender? Or is `.gender` just not present if it's not set?

src/components/Lend/LoanSearch/LoanSearchFilterChips.vue
```diff
@@ -105,7 +146,58 @@ export default {
...
+	watch: {
+		items() {
+			this.determineIsCollapsable();
+		}
+	}
 };
 </script>
+
+<style lang="postcss" scoped>
+	.container { max-height: 148px; }
```
mcstover, line 200: We may want to use `chip-container` here to be more specific... I don't think there are any other generic usages on this page of the `container` class but there may be elsewhere.

mcstover (approved):

PR #3991, opened 2022-07-07
feat: /lend/* pages server switch-up

christian14b:
This pr makes the following server changes for /lend/* pages :

/lend/id | ui | redirects to monolith for users that should see old page

/lend/category_name | ui | new url for /lend-by-category/category_name? just a redirect?
/lend-classic/id | monolith | renamed from /lend/id. canonicalizes to /lend/id

### DIFF
--- a/src/components/BorrowerProfile/FundedBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FundedBorrowerProfile.vue
@@ -1,0 +1,553 @@
+<template>
+	<div>
+		<div v-if="!enabledExperiment">
+			<div class="row borrower-profile-wrapper">
+				<div class="small-12 medium-4 columns">
+					<!-- Borrower photo -->
+					<loan-card-image
+						:loan-id="loan.id"
+						:name="loan.name"
+						:retina-image-url="loan.image.retina"
+						:standard-image-url="loan.image.default"
+						:is-visitor="true"
+						v-kv-track-event="['basket', 'basket-loan-profile', 'basket-loan-profile']"
+						:open-in-new-tab="true"
+						:use-default-styles="false"
+					/>
+				</div>
+				<div class="small-12 medium-8 columns">
+					<h2
+						class="tw-text-action tw-inline-block"
+						v-if="loan.status==='funded'"
+					>
+						100%
+					</h2>
+					<h2 class="tw-text-action tw-inline-block tw-capitalize">
+						{{ loan.status }}
+					</h2>
+					<!-- Total funded/loan amount -->
+					<div class="tw-mt-1">
+						Total loan: ${{ loan.loanFundraisingInfo.fundedAmount | numeral('0,0') }}
+					</div>
+					<!-- Borrower Name -->
+					<h1> {{ loan.name }} </h1>
+					<!-- Borrower location -->
+					<div class="tw-mt-2">
+						<div class="loan-location-flag">
+							<kv-flag :country="`${ loan.geocode.country.isoCode }`" />
+						</div>
+						<span>
+							{{ loan.geocode.city }}, {{ loan.geocode.state }}, {{ loan.geocode.country.name }}
+							/ {{ loan.sector.name }}
+						</span>
+
+						<div>
+							<!-- Link to see full borrower profile in old stack -->
+							<router-link
+								:to="`/lend/${loan.id}?minimal=false`"
+								v-kv-track-event="['Lending', 'full-borrower-profile-exit-link']"
+							>
+								See full borrower profile
+							</router-link>
+						</div>
+					</div>
+				</div>
+			</div>
+			<!-- Loans you might light section -->
+			<div class="row lyml-wrapper">
+				<div class="small-12 columns">
+					<hr class="tw-border-tertiary">
+					<h3 class="tw-text-center tw-mb-2 tw-mt-3" v-html="lymlHeading"></h3>
+					<l-y-m-l
+						class="tw-mb-3"
+						:basketed-loans="itemsInBasket"
+						:target-loan="loan"
+						@add-to-basket="handleAddToBasket"
+					/>
+				</div>
+			</div>
+			<hr class="tw-border-tertiary">
+			<div class="row">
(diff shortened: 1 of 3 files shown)


src/components/BorrowerProfile/FundedBorrowerProfile.vue
```diff
@@ -0,0 +1,553 @@
+<template>
```
dyersituations, line 1: Is there a good way to share this fairly large component. It looks like the only difference from the existing page is the use of `WwwPage`.

src/router/routes.js
```diff
@@ -242,6 +235,14 @@ module.exports = [
 			unbounce: true,
 		},
 	},
+	{
+		path: '/lend/:id(\\d+)',
+		redirect: '/lend-beta/:id'
```
mcstover, line 240: Won't this still have pages loading at the `/lend-beta/###` url? Why not define the BorrowerProfile page component here as well and just use `/lend/###`...

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -384,6 +389,14 @@ export default {
 		},
 	},
 	created() {
+		const expCookieSignifier = this.cookieStore.get('kvlendborrowerbeta');
+		if (expCookieSignifier !== 'b') {
+			const { query } = this.$route;
+			const queryString = Object.keys(query)
+				.map(key => `${key}=${query[key]}`)
+				.join('&');
+			this.$router.push(`/lend-classic/${this.$route.params.id}?query=${queryString}`);
```
mcstover, line 398: I think there are some other conditions we'll need to check here...and maybe be able to get rid of this exp cookie check altogether...
1. If a loan is `inPfp` we redirect to legacy (https://github.com/kiva/kiva/blob/development/sites/www_kiva/server/views/Xb/Loan/LoanView.php#L95)
2. If the lender is an RTP Translator we skip the experiment (https://github.com/kiva/kiva/blob/development/httpdocs/classes/API/Experiment/LoanDetailBeta.php#L22)

src/router/routes.js
```diff
@@ -131,13 +131,6 @@ module.exports = [
 			excludeFromStaticSitemap: true,
 		}
 	},
-	{
-		path: '/funded/:id',
```
mcstover, line 135: Excited to see this removed an incorporated into the new BP!!!

We'll eventually need to clean up the [redirect in legacy](https://github.com/kiva/kiva/blob/development/sites/www_kiva/server/views/Xb/Loan/LoanView.php#L151) but can probably just turn off the Experiment (`API_Experiment_FundedProfileRedirect`) if we haven't already...

src/pages/BorrowerProfile/BorrowerProfile.vue
christian14b: @mcstover Sounds good! Is there a way to get if lender has a RTP Volunteer role?

src/router/routes.js
christian14b: Yeah.. sounds right. @emuvente would this imply setting a `canonical url` for the `Borrower Profile` page?

src/pages/BorrowerProfile/BorrowerProfile.vue
eddieferrer: @eddieferrer As of my latest merged PR #1 is not true. UI will now render inPfp loans

eddieferrer:
@christian14b I noticed in `KivaClassicBasicLoanCard.vue` there is a check for `this.$route.path.includes('funded');` -- this will have to be refactored since you are removing the funded route

src/components/BorrowerProfile/FundedBorrowerProfile.vue
```diff
@@ -447,7 +408,7 @@ export default {
 					fetchPolicy: 'network-only',
 				}).then(data => {
 					// need to update this.itemsInBasket here.
-					this.itemsInBasket = _get(data, 'data.shop.basket.items.values');
+					this.basketedLoans = _get(data, 'data.shop.basket.items.values');
```
emuvente, line 411: I think that the itemsInBasket prop will be updated by the watchQuery for the borrower profile. Did you see that being updated or was this line required for it to update?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -408,9 +383,6 @@ export default {
 		isLessThan25() {
 			return this.unreservedAmount < 25 && this.unreservedAmount > 0;
 		},
-		inBorrowerProfilePage() {
```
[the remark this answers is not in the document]
emuvente: Actually, it looks like this component in main already has something like this, using the lendNowButton prop. So this PR will just need to set lendNowButton to `true` in the funded borrower profile.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -108,6 +108,14 @@
 		</article>
 		<what-is-kiva-modal v-if="kivaModuleExpEnabled && !shownModal" />
 		<!-- <aside>Similar loans</aside> -->
+		<article v-else>
```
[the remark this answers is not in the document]
emuvente: Yes, good point!

emuvente (approved):
Looks good to me 👍

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -377,24 +422,36 @@ export default {
...
 				})
 				.then(({ data }) => {
+					const expCookieSignifier = cookieStore.get('kvlendborrowerbeta');
+					if (expCookieSignifier === 'a' || expCookieSignifier === 'c') {
+						const { query } = route;
+						return Promise.reject({
+							path: `/lend-classic/${route.params.id}`,
+							query,
+						});
+					}
```
mcstover, line 445: Maybe in the near future we'll be able to get rid of this expCookie stuff and just provide a query param to allow bypass to the legacy bp...or a link in the page.

mcstover (approved):

PR #3994, opened 2022-07-08
fix: added missing loan card padding

dyersituations:
https://kiva.atlassian.net/browse/VUE-1102

Screen Shot 2022-07-07 at 5 01 00 PM

Screen Shot 2022-07-07 at 5 02 01 PM

### DIFF
--- a/src/components/LoanCards/ListLoanCard.vue
+++ b/src/components/LoanCards/ListLoanCard.vue
@@ -66,7 +66,7 @@
 						@track-loan-card-interaction="trackInteraction"
 					/>
 				</div>
-				<div class="large-7 columns show-for-large">
+				<div class="large-7 columns show-for-large tw-py-1">
 					<action-button
 						:loan-id="loan.id"
 						:loan="loan"


src/components/LoanCards/ListLoanCard.vue
```diff
@@ -66,7 +66,7 @@
 						@track-loan-card-interaction="trackInteraction"
 					/>
 				</div>
-				<div class="large-7 columns show-for-large">
+				<div class="large-7 columns show-for-large tw-py-1">
```
mcstover, line 69: Does this 'double" the padding at larger + smaller screen sizes?

mcstover: Maybe not if there is a margin collapse situation occurring. Just checking as the small and xl versions seem to already have the intended padding around the elements.

dyersituations: This element didn't have top/bottom padding before, so it works out. First screenshot is large, second is small. Small actually looks a bit better now, since the top of the green box was pretty close to the element above it. The extra whitespace below in the small example is there regardless of this new padding.

image

image

dyersituations: This padding is also applied to the lend buttons, since it's the same element wrapper, but again, the changes actually look a bit better in small view:

image

mcstover (approved):

PR #4000, opened 2022-07-08
feat: adding loan spotlight

galen94:
Adding a spotlight loan from the recommended by lenders channel with the women channel as a back up.


src/pages/Categories/CategoriesBeta.vue
```diff
@@ -10,7 +10,12 @@
 					future for themselves and their families.
 				</p>
 			</kv-grid>
-			<kv-grid class="md:tw-pt-6 lg:tw-pt-8">
+		</kv-page-container>
+		<div class="tw-pt-4 tw-mb-2">
```
eddieferrer, line 14: I think you can apply these classes directly on the loan-spotlight component and this wrapper div is unnecessary

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+
+import { toParagraphs } from '@/util/loanUtils';
+import gql from 'graphql-tag';
+import logReadQueryError from '@/util/logReadQueryError';
+import KvResponsiveImage from '@/components/Kv/KvResponsiveImage';
+import KvGrid from '~/@kiva/kv-components/vue/KvGrid';
+import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
+import KvButton from '~/@kiva/kv-components/vue/KvButton';
+
+const routePath = 'recommended-by-lenders';
```
eddieferrer, line 60: Let's make the default category string be a prop  that is passed to this component. This will allow the component to be used to spotlight other categories

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+
+function getTargetedChannel(targetedRoute, allChannels) {
+	const loanChannels = allChannels.lend?.loanChannels?.values;
+	const targetedLoanChannel = loanChannels.filter(
+		loanChannel => loanChannel.url.split('/').pop() === targetedRoute
+	);
+	// no channel that matches the name recommended-by-lenders
+	if (targetedLoanChannel.length === 0) {
+		// return id for channel women
+		return 5;
```
eddieferrer, line 125: We can't guarantee that women will be id = 5. You'll have to search for the id based on the url like you are doing for RBL

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+	},
+};
+
+</script>
+
+<style lang="postcss" scoped>
+
+.spotlight-loan-image >>> img {
+	@apply tw-object-cover tw-object-top tw-rounded;
+	@apply tw-w-full tw-h-full md:tw-w-[390px] md:tw-h-[320px] lg:tw-w-[520px] lg:tw-h-[390px];
```
eddieferrer, line 252: These widths and heights seem oddly specific, for this component to work in different contexts we might not need to be so specific. What happens if we do something like: `@apply tw-w-full tw-h-full md:tw-w-1/2 md:tw-max-h-[320px] lg:tw-max-h-[390px]` ? Setting width to be 50% and height to be a max-height instead?

galen94: Yeah they were pretty specific haha. The large size dimensions were what Nathan had, but the medium size dimensions were meant to avoid wrapping of the title "Today's loan spotlight".

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+					}
+				}
+			}
+		}
+	}
+`;
+
+function getTargetedChannel(targetedRoute, allChannels) {
+	const loanChannels = allChannels.lend?.loanChannels?.values;
+	const targetedLoanChannel = loanChannels.filter(
```
dyersituations, line 119: The optional chaining to get `loanChannels` implies that it could be undefined, so we probably want optional chaining here as well `loanChannels?.filter`.

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+	if (targetedLoanChannel.length !== 0 && targetedLoanChannel[0].loans.totalCount === 0) {
+		return 5;
+	}
+	// isolate recommended-by-lenders channel id
+	return targetedLoanChannel[0]?.id || null;
+}
+
+function filterByAnonymizationLevel(spotlightData) {
+	const firstFiveRecommendedLoans = spotlightData.lend?.loanChannelsById[0]?.loans?.values ?? [];
+	const nonAnonymousLoans = firstFiveRecommendedLoans.filter(
```
dyersituations, line 137: Same comment about probably wanting optional chaining here.

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+		},
+	},
+	created() {
+		let allChannelsData = {};
+		try {
+			allChannelsData = this.apollo.readQuery({
+				query: allChannelsQuery,
+			});
+		} catch (e) {
+			logReadQueryError(e);
```
dyersituations, line 205: We generally also pass in the location of the error as the second arg:
`logReadQueryError(e, 'LoanSpotlight created');`

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
...
+		let spotlightData = {};
+		try {
+			spotlightData = this.apollo.readQuery({
+				query: spotlightLoanQuery,
+				variables: {
+					ids: [this.targetedLoanChannelID],
+				},
+			});
+		} catch (e) {
+			logReadQueryError(e);
```
dyersituations, line 222: Similar comment about `logReadQueryError`

src/components/Categories/LoanSpotlight.vue
galen94: Just as a reference, this is how it looks with the original dimensions I had.

Screen Shot 2022-07-08 at 11 45 00 AM

dyersituations: I did some quick testing locally and the spotlight image was really tall in small view. Do we want a max height? Otherwise looks good!

image

src/components/Categories/LoanSpotlight.vue
eddieferrer: The nullish coalescing above means this  won't return undefined

dyersituations: Ah yup, I missed the `[]`.

src/components/Categories/LoanSpotlight.vue
eddieferrer: > @apply tw-w-full tw-h-full md:tw-w-[1/2] md:tw-max-h-[320px] lg:tw-max-h-[390px];

I actually like this solution with maybe adding a max width as well? Lets create 2 stories one for vertical and one with horizontal loan image and we can run this by Nathan

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,255 @@
+<template>
+	<div class="tw-bg-primary md:tw-bg-secondary">
+		<kv-page-container>
+			<kv-grid>
+				<h2 class="md:tw-hidden">
+					Today's loan spotlight
+				</h2>
+				<div class="md:tw-flex md:tw-pt-8 md:tw-pb-8">
+					<div class="md:tw-mr-3 lg:tw-mr-4 md:tw-flex-none">
```
eddieferrer, line 9: You could also try making this wrapper div 50% width or max width 50%

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,270 @@
...
+				}
+				return `${this.spotlightLoan.geocode.country.name ?? ''}`;
+			}
+			return '';
+		},
+		imageExists() {
+			return this.spotlightLoan.image?.default ?? '';
+		},
+		getSpotlightImage() {
+			if (this.imageExists === '') {
```
eddieferrer, line 205: For this method, and the  getSpotlightLoanLocation above, you could just do:
`if (this.spotlightLoan.image?.default)` instead of the varExists computed

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,270 @@
...
+				ids: [this.targetedLoanChannelID],
+			},
+		}).then(result => {
+			// filter out loans with anonymizationLevel of full, then take first in list
+			this.isLoading = false;
+			this.spotlightLoan = filterByAnonymizationLevel(result.data);
+		});
+	},
+	mounted() {
+		this.apollo.query({
```
eddieferrer, line 234: I think you can wrap this query in an if getSpotlightImage exists -- since created will have run on the server and you don't need to do this query if an image exists

src/pages/Categories/CategoriesBeta.vue
```diff
@@ -52,7 +64,7 @@
 					<kv-button
 						to="/lend"
 						variant="secondary"
-						state="active"
+						state=""
```
eddieferrer, line 67: remove empty prop

src/components/Categories/LoanSpotlight.vue
```diff
@@ -0,0 +1,270 @@
+<template>
+	<div>
+		<h2 class="md:tw-hidden tw-pb-2">
+			Today's loan spotlight
```
eddieferrer, line 4: I like leaving this text up while it loads, good choice! the loading state looks great

eddieferrer (commented):
Looking good. 
Overall, I would like this component to be more flexible so that it can be used in other contexts. Right now it really only works if it is the full page width. 

Heres an example story template I would like for this to work in. 

```
	template: `
		<div style="width: 100%; max-width: 1200px; display:flex">
			<div>
				<loan-spotlight
				:route-name='recommended-by-lenders'
				:fallback-route-name="women"
				/>
			</div>
			<div>
				<loan-spotlight
				:route-name='recommended-by-lenders'
				:fallback-route-name="women"
				/>
			</div>
		</div>
	`,
```

This component should render if used on 50% of the page like this, for example. 

I think you can get a lot of the way there by removing flex-none from the image

src/components/Categories/LoanSpotlight.vue
eddieferrer: I thought about this a little more --  Lets remove all the placeholder image code. Let's filter to only display loans that have an image. If a loan doesnt have an image it's not spotlight worth, lets not show it

galen94: I agree! I was thinking the same thing.

.storybook/stories/LoanSpotlight.stories.js
```diff
@@ -0,0 +1,180 @@
...
+							}
+						]
+	  				}
+				}
+			]
+		}
+	}
+};
+
+export default {
```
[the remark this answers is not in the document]
galen94: Isn't that for when a component gets its data from props? I thought in this instance we had to specify what data we wanted in the query above because we don't get anything from props.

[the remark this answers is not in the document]
galen94: Nevermind, I think you just mean the category names.

eddieferrer (approved):
Looks great! 👍

galen94:
@eddieferrer Not sure why this test is failing...

PR #4001, opened 2022-07-08
chore: update sentry packages

eddieferrer:


.github/dependabot.yml
```diff
@@ -12,4 +12,8 @@ updates:
     - ">= 2.7.a"
   - dependency-name: vue-svg-loader
     versions:
-    - ">= 0.15.0"    
+    - ">= 0.15.0"
+  # Version 2 of this package is for Vue3
+  - dependency-name: vue-countdown
```
eddieferrer, line 17: I installed this package, and version 2 is strictly for vue3 support, so pinning version

package.json
```diff
@@ -51,8 +51,8 @@
 		"@kiva/kv-components": "^3.0.6",
 		"@kiva/kv-tokens": "^2.1.0",
 		"@mdi/js": "^5.9.55",
-		"@sentry/tracing": "^6.13.3",
-		"@sentry/vue": "^6.13.3",
+		"@sentry/tracing": "^6.19.6",
```
eddieferrer, line 54: version 7 of Sentry drops IE11 support, not sure if we are ready for that yet

mcstover (approved):
I think we're close on IE11 but this is a good step forward for now

PR #4004, opened 2022-07-08
chore: update eslint-vue and fix lint errors

eddieferrer:
**There are no functional code changes in this PR**

This updated version of eslint is required for Vue 2.7.4 which has backported Vue 3 features, like the composition API, see: 
https://github.com/vuejs/vue/blob/main/CHANGELOG.md

_New rules that came up with this eslint upgrade:_ 

`vue/first-attribute-linebreak`
I auto fixed this and this is 95% of all the file changes in this PR, spacing only

These rules were all ignored where the violations came up: 

`vue/multi-word-component-names`
I ignored this rule in all the components that have a single word name

`vue/no-v-text-v-html-on-component`
This came up in GenericPromoBanner.vue I ignored as well, seems like an edge case. 

`vue/no-computed-properties-in-data` in LendIncrementButton.vue
This one is interesting, it leads me to believe that the LendIncrementButton will never initialize with the computed property, since the computed value is not available in data. 

`vue/no-useless-template-attributes` 
this one is also interesting. There are several places where we add attributes to the <template> tag like: 
```
<template
	#content class="tw-bg-primary"
>
```
According to my research, since template isn't actually rendered to the DOM this class would never be applied?

Or 
```
<template
	v-else
	v-kv-track-event="['TopNav','click-Promo','Lending Reward Banner']"
>
```
In which case this track event would never fire?


dyersituations (approved):

roger-in-kiva (approved):
Checked!

mcstover (approved):
All good on the formatting and we should probably try to fix up that interesting case of computed in data in a followup task.

PR #4006, opened 2022-07-11
Core 630 join team checkout

roger-in-kiva:


mcstover:
@michelleinez We have fallback mechanisms in our actual `/checkout` page to handle promo users that exit the single page experience and end up going through the standard checkout. We've already added support for handling checkout properly and the employee verification form as the components are shared and we just needed to include the logic to activate them.

This work also adds in the team join prompt.

I've added you to this PR b/c we'll need to do some deliberate testing on this code to ensure it's functioning properly for promo account AND that the standard checkout flow isn't obstructed. If you are both ready to roll on that testing we can merge  and start, otherwise I'd suggest waiting to merge until after lunch on Monday so we can ready the primary release without this included.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -212,6 +212,13 @@
 				@credit-removed="handleCreditRemoved"
 				@promo-opt-out-lightbox-closed="handleCancelPromoOptOut"
 			/>
+			<campaign-join-team-form
+				v-if="this.showTeamForm"
+				:campaign-name="campaignPartnerName"
+				:team-id="this.teamId"
+				:promo-id="this.promoFundId"
+				@team-process-complete="handleTeamJoinProcess"
```
mcstover, line 220: @roger-in-kiva Need to remove `this` on all of these fields.

mcstover: I see that's how it's setup in the CCLandingPage too. But `this` shouldn't be necessary...

mcstover (approved):

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Agreed!

PR #4007, opened 2022-07-11
fix: set loans per page value according promo card availability

edwinf-barreram:
# Description
- Set loans per page value according promo card availability

#Ticket
[MARS-170](https://kiva.atlassian.net/browse/MARS-170)

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -363,7 +363,7 @@ export default {
 			return client.query({
 				query: loanChannelPageQuery
 			}).then(({ data }) => {
-				const { route } = args
+				const { route } = args;
 				const { query, params, path } = route;
 
 				// Filter routes on route.param.category to get current path
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -176,7 +176,7 @@ import {
 } from '@/util/loanChannelUtils';
 import KvButton from '~/@kiva/kv-components/vue/KvButton';
 
-const loansPerPage = 12;
+let loansPerPage = 11;
 
 // A map of functions to transform url query parameters to/from graphql variables.
 // Each key in urlParamTransform is a url query parameter (e.g. the 'page' in ?page=2).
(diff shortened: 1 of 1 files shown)


mcstover:
@edwinf-barreram I think you'll actually need to run this switch in the prefetch around line 360 "before" the loan channel loan data is requested. This will likely require factoring the initializeMGPromo method out of the vue components moving it up under the imports...

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -254,7 +269,7 @@ export default {
 	data() {
 		return {
 			offset: 0,
-			limit: loansPerPage,
+			limit: 0,
```
mcstover, line 272: Probably want to keep this as is if possible...otherwise this looks good.

mcstover (approved):

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -559,28 +578,15 @@ export default {
 			// Update Lend Filter Exp CASH-545
 			this.getLendFilterExpVersion();
 		},
-		initializeMonthlyGoodPromo() {
+		async initializeMonthlyGoodPromo() {
```
dyersituations, line 581: It doesn't look like this method needs `async`

emuvente: Yes, this looks like it will work without `async`

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -357,6 +372,11 @@ export default {
 				// Extract query
 				const pageQuery = _get(args, 'route.query');
 
+				const currentRoute = _get(args, 'route.path').replace('/lend-by-category/', '');
+				const matchedRoutes = _filter(targetRoutes, route => route.route === currentRoute);
```
eddieferrer, line 376: Can we use native JS filtering instead of lodash here?

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -357,6 +372,11 @@ export default {
 				// Extract query
 				const pageQuery = _get(args, 'route.query');
 
+				const currentRoute = _get(args, 'route.path').replace('/lend-by-category/', '');
```
eddieferrer, line 375: No need for lodash here either

mcstover (approved):

emuvente (approved):
Looks great!

PR #4009, opened 2022-07-11
feat: pager now uses new KvMaterialIcon

dyersituations:
https://kiva.atlassian.net/browse/VUE-1151

image


mcstover (approved):
oooh so fresh!

PR #4023, opened 2022-07-14
fix: issue with montly good and loans per page

dyersituations:
https://kiva.atlassian.net/browse/VUE-1175

- The issue has to do with subtracting 1 from `loansPerPage` for the monthly good promo
- The specific issue seemed to be that the `loansPerPage` variable in memory kept getting subtracted by 1 with each render, leading to a mismatch in `loansPerPage`, both in the Apollo cache (leading to Apollo cache errors) and between `loansPerPage` and `this.limit` (leading to weird pager behavior)
- I took the opportunity to generally simplify the logic around handling the `offset`, removing several `lodash` imports

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -155,12 +155,9 @@
 
 <script>
 import _get from 'lodash/get';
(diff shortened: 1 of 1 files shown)


mcstover (approved):

eddieferrer:
Nice work on this @dyersituations  !!

PR #4024, opened 2022-07-15
fix: update paypal config after changes to amount, stup out the same …

mcstover:
…for google and apple

### DIFF
--- a/src/components/Payment/BraintreeDropInInterface.vue
+++ b/src/components/Payment/BraintreeDropInInterface.vue
@@ -83,6 +83,36 @@ export default {
 			updatingPaymentWrapper: false,
 		};
 	},
+	computed: {
+		formattedAmount() {
+			return numeral(this.amount).format('0.00');
+		},
+		applePaymentRequest() {
+			return {
+				total: {
+					label: 'Kiva',
+					amount: this.formattedAmount,
+				},
+				requiredBillingContactFields: ['postalAddress']
+			};
+		},
+		googleTransactionInfo() {
+			return {
+				totalPriceStatus: 'FINAL',
+				totalPrice: this.formattedAmount,
+				currencyCode: 'USD',
+			};
+		}
+	},
+	watch: {
+		amount(next, prev) {
+			if (next !== prev) {
+				this.btDropinInstance.updateConfiguration('paypal', 'amount', this.formattedAmount);
+				this.btDropinInstance.updateConfiguration('googlePay', 'transactionInfo', this.googleTransactionInfo);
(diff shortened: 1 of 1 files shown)


src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -83,6 +83,37 @@ export default {
...
+			};
+		}
+	},
+	watch: {
+		amount(next, prev) {
+			if (next !== prev) {
+				this.btDropinInstance.updateConfiguration('paypal', 'amount', this.formattedAmount);
+				// eslint-disable-next-line max-len
+				// this.btDropinInstance.updateConfiguration('googlePay', 'transactionInfo', this.googleTransactionInfo);
+				// this.btDropinInstance.updateConfiguration('applePay', 'paymentRequest', this.applePaymentRequest);
```
mcstover, line 113: Uncommenting these locally didn't result in any errors but I'm unsure of how we'll test them. @eddieferrer did you help with this originally? Do you know if we can verify them in DEV?

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -83,6 +83,37 @@ export default {
...
+				totalPriceStatus: 'FINAL',
+				totalPrice: this.formattedAmount,
+				currencyCode: 'USD',
+			};
+		}
+	},
+	watch: {
+		amount(next, prev) {
+			if (next !== prev) {
+				this.btDropinInstance.updateConfiguration('paypal', 'amount', this.formattedAmount);
```
mcstover, line 110: I verified that with this config update sequence in place the amount value that Paypal is tracking is indeed updated after a donation change resulting in a basket total amount change. Once logged into the Paypal popup you can see the amount notated under the credit option, close the popup adjust the basket then re-open Paypal and see the new amount reflected.

dyersituations (approved):

src/components/Payment/BraintreeDropInInterface.vue
mcstover: Ok, it looks like we can at least make sure both Google and Apple initialize correctly once we're in dev. So, I've uncommented these too.

emuvente (approved):
Awesome!

src/components/Payment/BraintreeDropInInterface.vue
eddieferrer: @mcstover instructions for testing the complete flow for Google pay and applePay can be found here: https://kiva.atlassian.net/wiki/x/3QA9Kg

eddieferrer (approved):

PR #4029, opened 2022-07-18
Ack 325 monthly good

galen94:
ACK-325, Adding Monthly Good module to categories beta page


src/pages/Categories/CategoriesBeta.vue
```diff
@@ -87,13 +87,17 @@
 				</div>
 			</kv-grid>
 		</kv-page-container>
+		<div>
```
galen94, line 90: Just realized I probably don't need this wrapper div

src/components/Categories/MonthlyGoodModule.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+	<div class="tw-relative">
+		<kv-responsive-image
+			class="image-sizing"
+			:images="monthlyGoodImages"
+			loading="lazy"
+			:alt="'Monthly Good'"
```
dyersituations, line 7: Do we need the extra single quotes?

galen94: Or actually it doesn't seem to work without them, nevermind. :)

src/components/Categories/MonthlyGoodModule.vue
```diff
@@ -0,0 +1,78 @@
...
+		};
+	},
+};
+
+</script>
+
+<style lang="postcss" scoped>
+
+.image-sizing >>> img {
+	@apply tw-w-full tw-max-h-[311px] md:tw-max-h-[428px] md:tw-h-[428px] tw-object-cover tw-object-top;
```
dyersituations, line 71: In general we want to avoid the Tailwind JIT classes, since they lead to a larger compiled Tailwind CSS file. If a sizing doesn't exist in our preset Tailwind classes, we want to use inline/normal CSS. These classes are applied in CSS here, so it would be quick to convert these to plain CSS right here.

src/components/Categories/MonthlyGoodModule.vue
dyersituations: Oh right, you can change to `alt="Monthly Good"`

src/components/Categories/MonthlyGoodModule.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+	<div class="tw-relative">
```
eddieferrer, line 2: Since this module is the full width of the screen, on very wide screens, the image will zoom in and look very bad, let's give this a max width so that on ultra wide screens it is just centered with a max width.

.storybook/stories/MonthlyGoodModule.stories.js
```diff
@@ -0,0 +1,13 @@
+import MonthlyGoodModule from '@/components/Categories/MonthlyGoodModule';
```
eddieferrer, line 1: I actually don't think a story is necessary for this module. It is pretty "dumb" and doesnt really have much functionality

src/components/Categories/MainCategoryTile.vue
```diff
@@ -160,15 +160,18 @@ export default {
 	@apply tw-rounded;
 }
 
+/* tw-transform-gpu is used to eliminate readjustments/wiggles after the zoom effect on the images.
+Rendering by the GPU here instead of the CPU ensures a smoother transition. */
+
 .category-image__large >>> img {
-	@apply group-hover:tw-scale-110 tw-transition-all tw-duration-500 tw-ease-in-out tw-transform;
+	@apply group-hover:tw-scale-110 tw-transition-all tw-duration-500 tw-ease-in-out tw-transform-gpu;
```
eddieferrer, line 167: I tested this locally, looks great, good find on this fix 👍

src/components/Categories/MonthlyGoodModule.vue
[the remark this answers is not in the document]
galen94: For these, I based the image heights off of Nathan's design dimensions, small 311px, md/lg 428px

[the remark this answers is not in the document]
eddieferrer: The assets are reasonably close, but not the exact size as Nathans specs, you can use your judgement here if you think the max-height heights look better than the asset heights.

eddieferrer (approved):

dyersituations (approved):

PR #4035, opened 2022-07-19
feat: adding FAQs to categories page

galen94:


src/pages/Categories/CategoriesBeta.vue
```diff
@@ -173,6 +213,14 @@ export default {
 			return category.retinaImage?.url ?? '';
 		}
 	},
+	computed: {
+		contentGroups() {
+			return this.pageData?.page?.pageLayout?.contentGroups ?? [];
+		},
+		faqContentGroup() {
+			return this.contentGroups[0] ?? {};
```
mcstover, line 221: Please `find` or `filter` for the faq content group by it's type, which should be `frequentlyAskedQuestions`.

dyersituations (commented):
Looks good other than the comment from @mcstover. Holding off on approving.

src/pages/Categories/CategoriesBeta.vue
mcstover: thanks @galen94 !

src/pages/Categories/CategoriesBeta.vue
```diff
@@ -156,15 +171,40 @@ export default {
...
 		};
 	},
 	apollo: {
-		query: allCategoriesQuery,
-		preFetch: true,
-		result(result) {
-			this.categories = result.data?.lend?.loanChannels?.values ?? [];
+		preFetch(config, client) {
+			return client.query({
+				query: allCategoriesQuery
```
eddieferrer, line 180: You can combine both of these two queries together by just combining the Graphql query object so that they happen asynchronously instead of waiting for the categories query, then executing the contentful query

src/pages/Categories/CategoriesBeta.vue
```diff
@@ -173,6 +185,16 @@ export default {
 			return category.retinaImage?.url ?? '';
 		}
 	},
+	computed: {
+		contentGroups() {
+			return this.pageData?.page?.pageLayout?.contentGroups ?? [];
+		},
+		faqContentGroup() {
+			return this.contentGroups?.find(({ type }) => {
+				return type ? type === 'frequentlyAskedQuestions' : false;
```
dyersituations, line 194: Could be simplified to `return type === 'frequentlyAskedQuestions'`

dyersituations (approved):

src/pages/Categories/CategoriesBeta.vue
eddieferrer: > Could be simplified to `return type === 'frequentlyAskedQuestions'`

I thought this too -- this exact code is actually on several pages, and I probably wrote it and it was copied and pasted :)

eddieferrer (approved):

src/pages/Categories/CategoriesBeta.vue
galen94: Yes, I was trying to stay consistent with the other FAQ code on the MonthlyGoodLandingPage.vue. Should I leave it? @eddieferrer

mcstover (approved):

PR #4039, opened 2022-07-20
fix: fetch more lenders to avoid small filtered list

christian14b:
## Description
As there isn't a fast way of filtering lenders collection in loan type to avoid fetching anonymous lenders.

The idea behind this pr is to fetch more lenders (10) in order to have more options and always show the required amount of lenders (3 lenders in desktop and 2 in mobile).

Problem: Fetching 3 lenders and filtering one because was an anonymous lender. This image should show 3 people.
image

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -164,7 +164,7 @@ const pageQuery = gql`
 					hash
 				}
 				plannedExpirationDate
-				lenders(limit: 3) {
+				lenders(limit: 10) {
 					values {
 						id
 						name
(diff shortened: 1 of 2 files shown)


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -164,7 +164,7 @@ const pageQuery = gql`
 					hash
 				}
 				plannedExpirationDate
-				lenders(limit: 3) {
+				lenders(limit: 10) {
```
mcstover, line 167: I wish this query wasn't part of the loan data and page load...but it'll do for now...

mcstover (approved):

PR #4040, opened 2022-07-20
Core 644 mg entrypoint experiment setup

roger-in-kiva:


mcstover:
@roger-in-kiva can you please see if re-using the mgSwapText experiment instead of adding a new one. Or, if you could at least replace that existing experiment. At the least, your new exp will need to be setup in the same way as that existing exp.

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -177,24 +178,27 @@ export default {
...
+			);
+		}
+
+		this.$nextTick(() => {
+			this.showMGUpsellLink = true;
 		});
 	}
 };
 </script>
+0
```
dyersituations, line 204: Remove trailing `0`

dyersituations (approved):

mcstover (approved):

src/components/WwwFrame/LendMenu/TheLendMenu.vue
roger-in-kiva: Removed

PR #4041, opened 2022-07-20
feat: create SavedSearchItem component + use for entry list

tiarascoleman:
Decided to make grid layout into cards (hopefully this does not result into an issue with Design)
Screen Shot 2022-07-20 at 4 02 08 PM
Screen Shot 2022-07-20 at 4 02 39 PM

 The second image is here to show that the cards are responsive

### DIFF
--- a/src/components/Settings/SaveSearchItem.vue
+++ b/src/components/Settings/SaveSearchItem.vue
@@ -1,0 +1,61 @@
+<template>
+	<article>
+		<div
+			class="tw-flex tw-flex-col"
+		>
+			<span
+				v-if="savedSearch.name"
+				class="tw-text-medium tw-m-1 tw-p-1 tw-font-medium"
+			>
+				{{ savedSearch.name }}
+			</span>
+			<span>
+				<kv-checkbox
+					v-model="showAlerts"
+					class="tw-text-small tw-m-1 tw-p-1"
+				>
+					Email me when new loans become available
+				</kv-checkbox>
+			</span>
+			<span>
+				<kv-button
+					variant="secondary"
+					class="tw-text-small tw-m-1 tw-p-1"
+				>
+					Delete Search
+				</kv-button>
+				<kv-button
+					variant="secondary"
+					class="tw-text-small tw-m-1 tw-p-1"
+				>
+					View Loans
+				</kv-button>
+			</span>
+		</div>
+	</article>
+</template>
+
+<script>
+import KvButton from '~/@kiva/kv-components/vue/KvButton';
+import KvCheckbox from '~/@kiva/kv-components/vue/KvCheckbox';
+
+export default {
+	name: 'SaveSearchItem',
+	data() {
+		return {
+			showAlerts: false,
+		};
+	},
+	inject: ['apollo', 'cookieStore'],
+	components: {
+		KvButton,
+		KvCheckbox,
+	},
+	props: {
+		savedSearch: {
+			type: Object,
+			default: () => {},
+		}
+	}
+};
+</script>
--- a/src/pages/Settings/SavedSearchBeta.vue
+++ b/src/pages/Settings/SavedSearchBeta.vue
@@ -1,33 +1,46 @@
 <template>
-	<www-page>
+	<www-page
+		:gray-background="true"
+	>
+		<template #secondary>
+			<the-my-kiva-secondary-menu />
+		</template>
 		<kv-page-container>
-			<kv-grid class="tw-pt-4 md:tw-pt-6 lg:tw-pt-8">
-				<h1 class="tw-mb-2">
-					Saved searches
-				</h1>
+			<kv-grid class="row-column">
+				<h2 class="tw-mb-2 tw-justify-start">
+					Saved Searches
+				</h2>
 			</kv-grid>
-			<kv-grid class="md:tw-pt-6 lg:tw-pt-8">
-				<div
-					v-for="(search, index) in savedSearches"
-					:key="index"
-				>
-					{{ search.name }}
-				</div>
-			</kv-grid>
+
+			<div>
+				<kv-grid class="tw-row-col tw-p-6 tw-m-1">
+					<save-search-item
+						v-for="(search, index) in savedSearches"
+						class="tw-grid tw-grid-col-1 tw-gap-4 tw-rounded tw-border-transparent
+						tw-max-w-lg tw-bg-primary"
+						:key="index"
+						:saved-search="search"
+					/>
+				</kv-grid>
+			</div>
 		</kv-page-container>
 	</www-page>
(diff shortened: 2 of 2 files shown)


src/components/Settings/SaveSearchItem.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+	<main class>
```
dyersituations, line 2: We probably don't want `main` here, since that semantic tag is meant for the main content of the document. For example, we currently use that tag in the page container components. Using `div` here would work, or a semantic tag like `article`. Also, the empty `class can be removed.

src/components/Settings/SaveSearchItem.vue
```diff
@@ -0,0 +1,69 @@
...
+
+export default {
+	name: 'SaveSearchItem',
+	data() {
+		return {
+			showAlerts: false,
+
+		};
+	},
+	inject: ['apollo', 'cookieStore'],
```
dyersituations, line 42: At the moment, this component doesn't need these.

src/components/Settings/SaveSearchItem.vue
```diff
@@ -0,0 +1,69 @@
...
+			type: Object,
+			default: () => {},
+		}
+	}
+};
+</script>
+
+<style scoped>
+	main {
+		max-width: 10000px;
```
dyersituations, line 58: This seems like a huge number, why is it needed?

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -6,27 +6,34 @@
...
-			<kv-grid class="md:tw-pt-6 lg:tw-pt-8">
-				<div
-					v-for="(search, index) in savedSearches"
-					:key="index"
-				>
-					{{ search.name }}
-				</div>
-			</kv-grid>
+
+			<main class>
```
dyersituations, line 10: The `main` here can probably be replaced with another element tag as well, and the empty `class` removed.

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -6,27 +6,34 @@
...
+				<kv-grid class>
```
dyersituations, line 11: Empty `class` can be removed.

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -68,3 +75,28 @@ export default {
 	}
 };
 </script>
+
+<style scoped>
```
dyersituations, line 79: Try to replace as much of this as possible with tailwind.

src/components/Settings/SaveSearchItem.vue
mcstover: We'll eventually need these for queries around the loan count...so I'm ok with keeping them as that work is up soon.

dyersituations:
If we keep the cards, we would probably want some padding/spacing here.
image

src/pages/Settings/SavedSearchBeta.vue
mcstover: Agree all of these styles should be converted to `tw-` classes.

Except for the 10k...please remove that everywhere you have it.

mcstover:
Nice to see the content rows stubbed out. If you can do the tailwind styles + div clean up work we can merge this initial pass.

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -68,3 +75,28 @@ export default {
...
+
+.container {
+	padding: 2px 16px;
+}
+
+.card {
+	display: grid;
+	grid-template-columns: repeat (auto-fit, minmax(300px, 1fr));
+	grid-gap: 1rem;
+	box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
```
mcstover, line 93: We generally don't use drop shadows in Kiva Classic. This should be a border instead. `tw-border`

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -68,3 +75,28 @@ export default {
...
+	transition: 0.3s;
+	border-radius: 5px;
```
mcstover, line 95: We have standard sizes for border radius, `tw-rounded-sm` should work on the row container.

mcstover:
Also, I do think that going with a list of cards like this is a fine and suitable substitute for the old table layout! Good choice. A little sprucing up and we'll have an easy to use and look at interface.

eddieferrer:
I think a better example for the layout of this page would be any of the settings pages. They also contain an h1 header and a series of "cards" down the page. Check out src/pages/Settings/SettingsPage.vue or src/pages/Settings/EmailSettings.vue

tiarascoleman (commented):
Screen Shot 2022-08-02 at 2 46 23 PM
This is the new page!

src/pages/Settings/SavedSearchBeta.vue
```diff
@@ -1,41 +1,54 @@
 <template>
-	<www-page>
+	<www-page
+		:gray-background="true"
+	>
+		<template #secondary>
+			<the-my-kiva-secondary-menu />
```
dyersituations, line 6: Just out of curiosity, why do we need to specify the secondary menu?

dyersituations:
Looks great! Some general UX observations in regards aligning with the loan cards at `/lend/filter-alpha`. This is assuming that we don't have a mock we're working directly from.
- Might be good to add a bit more padding inside the card
- Might be good to turn the "View loans" button into a green action button
- Might be good to add a gray border

Loan card for reference:
image

mcstover:
@dyersituations these a nice suggestions but I'd like to defer any "polish" work to a later stage as we're going to have a quick design review to help drive those decisions. If the stubbed out code is clean at this point we should be good to proceed.

mcstover (approved):

src/pages/Settings/SavedSearchBeta.vue
mcstover: The secondary menu will provide navigation within the "logged-in" user section of the site, similar to `/portfolio/lending-stats` or `/settings`.

mcstover: The legacy page has a very outdated version of this "secondary" nav as well.

dyersituations: Sounds good, thanks.

dyersituations (approved):

src/pages/Settings/SavedSearchBeta.vue
tiarascoleman: Oh the secondary menu is the secondary navigation which is VUE-1128!

PR #4043, opened 2022-07-21
fix: better tailwind jit compiling

eddieferrer:
I know this makes no sense, because we are getting the prefix property from KvConfig, which is getting it from the primitives. But I believe it could be a tailwind bug?

I tried adding this and I got much better results locally. 

Specifically for the use case of adding a class in `ui` repo that is not being used in the `kv-ui-elements` repo and having that class show  up in the compiled code without needing to add it to the purge safelist

I'm curious for others to verify

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -1,6 +1,7 @@
 const kvConfig = require('@kiva/kv-components/tailwind.config.cjs');
 
 module.exports = {
+	prefix: 'tw-',
 	presets: [kvConfig],
 	content: [
 		'./node_modules/@kiva/kv-components/**/*.vue',


dyersituations:
@eddieferrer Do you have some specific examples of classes we could test with locally in this branch?

eddieferrer:
@dyersituations you could try removing the classes in the safelist and testing with those.

dyersituations:
> @dyersituations you could try removing the classes in the safelist and testing with those.

I tested removing this from `main` and it worked: remove `tw-text-jumbo` from the safelist, remove the hard source cache folder, run dev, add `tw-text-jumbo` to a node, let it recompile -> change shows up.

Theoretically that shouldn't work from `main`, right? I've found JIT to work _sometimes_...

mcstover:
Well it depends on if tw-text-jumbo was used in any other markup in the repo...you can also test by finding something not used anywhere and then adding it while running dev mode...

dyersituations:
> Well it depends on if tw-text-jumbo was used in any other markup in the repo...you can also test by finding something not used anywhere and then adding it while running dev mode...

Just checked, and `tw-text-jumbo` wasn't being used anywhere, so it looks like while running from `main`, JIT worked for me adding new Tailwind classes. I've seen JIT work sometimes and not other times in general, like when I used to add `[]` JIT classes... It sometimes would pick up the new classes, sometimes I had to clear the cache and hope they appeared...

mcstover:
I've also seen intermittent "activation" of new classes with JIT and I suspect it's rooted in our threadLoader or buildCache as we seem to have all the tailwind specific build configs (jit mode, NODE_ENV = development) in place...

eddieferrer:
> > Well it depends on if tw-text-jumbo was used in any other markup in the repo...you can also test by finding something not used anywhere and then adding it while running dev mode...
> 
> Just checked, and `tw-text-jumbo` wasn't being used anywhere, so it looks like while running from `main`, JIT worked for me adding new Tailwind classes. I've seen JIT work sometimes and not other times in general, like when I used to add `[]` JIT classes... It sometimes would pick up the new classes, sometimes I had to clear the cache and hope they appeared...

I'm not sure thats a good test. tw-text-jumbo is used in the kv-ui-elements repo. Those classes have always worked for me. The ones I have a problem with are the classes that are not used in kv-ui-elements

dyersituations:
> I'm not sure thats a good test. tw-text-jumbo is used in the kv-ui-elements repo. Those classes have always worked for me. The ones I have a problem with are the classes that are not used in kv-ui-elements.

Gotcha, I forgot to check the elements repo. Later today I'll try with a class in neither repo.

dyersituations:
@eddieferrer I was going back through open PRs and noticed that I never did another quick test before Kiva Connect. Sorry for the delay! Interestingly I'm still just seeing inconsistent compilation of TL JIT. This morning TL classes not previously in the compiled TL CSS are working in both this branch and in `main`. Classes I used to test that weren't previously used or in the compiled code were float classes, like `tw-float-none`.

eddieferrer:
I'm just going to close this, TW JIT is still inconsistent with this, not sure this helps

dyersituations:
> I'm just going to close this, TW JIT is still inconsistent with this, not sure this helps

At one point I spend a couple hours poking at the build process, but it was a can of worms. I then heard we were long-term hoping to move everything to `cms-page-server`, so I stopped poking at it. Haha.

PR #4047, opened 2022-07-21
fix: link url for anonymous users in share button

eddieferrer:


dyersituations (approved):

mcstover (approved):
Generally interesting that `lender` is a required prop in this context. We prob should revisit that at somepoint.

PR #4048, opened 2022-07-21
feat: install hotjar tracking code

christian14b:
This pr installs the code for Hotjar as an effort to understand how users are reacting to site's elements.

**Hotjar** is a third-party tool that allows internal users to survey visitors to the site based on various targeting measures.


dyersituations:
General question: why do we need hotjar if we already have fullstory?

eddieferrer:
+1 to Caseys question. So much tracking.... 

@mcstover do we need to mirror this script load for the unbounce pages that are wrapped in the IFrameEmbed? That script is currently in unbounce. It might be better going forward to move that script to the UI repo and have it available externally and then just load it in unbounce. This would make it easier to keep in sync with changes to `src/head/script.js`

mcstover (changes requested):
This took needs a data audit, external configuration and potentially the addition of several data exclusion classes before being activated on the site.

mcstover (approved):
We've _now_ configured the HotJar account to "Suppress all on page content". Should this configuration change we will need to turn off the tool and alter our code to continue suppression of PII on the site.

config/audit.js
```diff
@@ -17,6 +17,8 @@ module.exports = merge(base, {
 		gaAlternateId: 'UA-11686022-6',
 		enableSnowplow: true,
 		snowplowUri: 'events.fivetran.com/snowplow/v5qt54ocr2nm',
+		enableHotjar: false,
+		hotjarId: '3071239',
```
mcstover, line 21: @christian14b @emuvente I think we should at minimum setup a 2nd site that can be used across all lower environments for testing. I'll check our account for that option.

mcstover (approved):

emuvente (approved):
👍

PR #4049, opened 2022-07-21
fix: remove the live-loans endpoint from our lighthouse run as it fai…

mcstover:
…ls, dropping all other results

### DIFF
--- a/lighthouserc-prod.js
+++ b/lighthouserc-prod.js
@@ -10,7 +10,7 @@ module.exports = {
 				'https://www.kiva.org/lend/filter',
 				'https://www.kiva.org/lend-by-category',
 				'https://www.kiva.org/lend-by-category/women',
-				'https://www.kiva.org/live-loan/f/sort_newest/url/1',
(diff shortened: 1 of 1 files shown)


lighthouserc-prod.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
 				'https://www.kiva.org/lend/filter',
 				'https://www.kiva.org/lend-by-category',
 				'https://www.kiva.org/lend-by-category/women',
-				'https://www.kiva.org/live-loan/f/sort_newest/url/1',
+				// 'https://www.kiva.org/live-loan/f/sort_newest/url/1',
```
mcstover, line 13: Unfortunately, giving the process more time didn't solve the access issues we're having trying to complete the redirect from a live loan url to an actual borrower profile page. I'm thinking now that Px is actually causing this but haven't been able to reproduce in isolation.

dyersituations: What is Px?

dyersituations (approved):

PR #4051, opened 2022-07-22
fix: kvdropdown menu showing up under borrower profile lend cta

eddieferrer:
I introduced this bug, on the non PFP pages, the about menu and user menu show up under the lend cta. This is because unlike the lend menu they use the kv-dropdown component which has a different z-index. 😭 

I'll hotfix this out

The rest are more minor spacing changes. 

The borrower profile is starting to feel a little bit hacky with all the state dependent styling and the 3 different states and all their permutations (share button on/off, logged in/out, social exp on/off). Maybe some refactoring and more thinking of these styles with style guide in mind.  Personally I prefer the approach with the share button of using slots from the borrower profile to slot the button in the summary card/lend cta. It at least keeps the component in one place vs having an instance nested in the lend cta and another in the summary card (like the jump links)....

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -29,6 +29,7 @@
 					},
 					'lg:tw-rounded-t',
 					'lg:tw-px-4',
+					'lg:tw-gap-1'
 				]"
 			>
 				<div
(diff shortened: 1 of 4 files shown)


mcstover (approved):
Good thoughts on the every growing complexity of this page. I agree that we should consider ways to simplify the logic and cleanup the switching between states at some point. There will be a couple of opportunities for that in the not to distant future.

PR #4055, opened 2022-07-22
fix: notes on exp version extraction

mcstover:


mcstover:
New commit hash to trigger a new tag. I used --amend on a previous commit that had already been merged which didn't create a new tag.

PR #4056, opened 2022-07-29
feat: add new cookies for Optimizely experiment

edwinf-barreram:
# Ticket

[MARS-194](https://kiva.atlassian.net/browse/MARS-194)


src/graphql/query/shared/hasEverLoggedIn.graphql
```diff
@@ -1,3 +1,13 @@
 query hasEverLoggedIn {
 	hasEverLoggedIn @client
+	my {
```
emuvente, line 3: The hasEverLoggedIn query is used in other components, so let's put this in a new query and add it as another call in the prefetch:
```
client.query({ query: hasEverLoggedInQuery }),
client.query({ query: optimizelyUserDataQuery }),
fetchAllExpSettings(config, client, {
	query: route?.query,
	path: route?.path
})
```

src/components/WwwFrame/WwwPage.vue
```diff
@@ -28,6 +28,9 @@ import TheFooter from './TheFooter';
 import TheBasketBar from './TheBasketBar';
 import TheBannerArea from './TheBannerArea';
 
+const hasLentBeforeCookie = 'kvu_lent_before';
+const hasDepositBeforeCookie = 'kvu_deposit_before';
```
emuvente, line 32: It's best to keep the cookie names short, like `kvu_lb` and `kvu_db`

src/graphql/query/shared/hasEverLoggedIn.graphql
```diff
@@ -1,3 +1,13 @@
 query hasEverLoggedIn {
 	hasEverLoggedIn @client
```
mcstover, line 2: Generally, I think we should keep this query file clean as it's used in several other context, and has a name indicating exactly what it does. Please use an inline query or create a different file to hold the new data queries.

src/components/WwwFrame/WwwPage.vue
edwinf-barreram: Got it, will change!

src/graphql/query/shared/hasEverLoggedIn.graphql
mcstover: oh, @emuvente already addressed above. Thanks!

src/components/WwwFrame/WwwPage.vue
```diff
@@ -72,9 +75,19 @@ export default {
...
 	created() {
+		const result = this.apollo.readQuery({
+			query: hasEverLoggedInQuery,
+		});
+
+		const hasLentBefore = result?.my?.loans?.totalCount > 0;
+		const hasDepositBefore = !result?.my?.userAccount?.isFirstTimeDepositor;
+
+		this.cookieStore.set(hasLentBeforeCookie, hasLentBefore);
+		this.cookieStore.set(hasDepositBeforeCookie, hasDepositBefore);
```
emuvente, line 89: These should probably be limited to our site and secure connections only:
```
this.cookieStore.set(hasLentBeforeCookie, hasLentBefore, { secure: true, sameSite: 'strict' });
this.cookieStore.set(hasDepositBeforeCookie, hasDepositBefore, { secure: true, sameSite: 'strict' });
```

src/graphql/query/shared/hasEverLoggedIn.graphql
```diff
@@ -1,3 +1,13 @@
...
+	my {
+		userAccount {
+			id
+			balance
+			isFirstTimeDepositor
+		}
+		loans {
+			totalCount
+		}
+	}
```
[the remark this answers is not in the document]
emuvente: Yeah, I'm concerned about that too. These were requested as general controls that can be used in optimizely, not for a specific experiment, but we could ask if we could limit this to a certain page initially. 

I think we could also try skipping the query if the cookies already exist, that way it only happens for the first request in the session.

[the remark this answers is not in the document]
mcstover: If we can focus this in or prevent subsequent queries I think that would be helpful. Otherwise, we'll just have to keep an eye out for any lengthening of the page load...

src/components/WwwFrame/WwwPage.vue
```diff
@@ -67,14 +82,27 @@ export default {
...
 				fetchAllExpSettings(config, client, {
 					query: route?.query,
 					path: route?.path
 				})
 			]);
-		},
+		}
 	},
 	created() {
+		if (this.cookieStore.get(hasLentBeforeCookie) === undefined || this.cookieStore.get(hasDepositBeforeCookie) === undefined) { // eslint-disable-line max-len
```
mcstover, line 94: This check would also need to be a conditional for including the new `client.query({ query: optimizelyUserDataQuery })` in the preFetch. [Here's an example of a conditional prefetch](https://github.com/kiva/ui/blob/main/src/pages/Thanks/ThanksPage.vue#L189)

edwinf-barreram: done!

mcstover (approved):

src/components/WwwFrame/WwwPage.vue
```diff
@@ -64,17 +79,32 @@ export default {
 		};
 	},
 	apollo: {
-		preFetch(config, client, { route }) {
+		preFetch(config, client, { cookieStore, route }) {
+			const getUserData = cookieStore.get(hasLentBeforeCookie) === undefined || cookieStore.get(hasDepositBeforeCookie) === undefined; // eslint-disable-line max-len
```
dyersituations, line 83: What happens when these cookies are `false`, then the user does these actions, will these cookies always be `false`?

edwinf-barreram: If cookie is false, it will be handle by the Optimizely experiment. Added an image for reference.

User Metrics: Audiences - Optimizely 2022-08-01 19-48-58

dyersituations: Right, but do we want the cookie value to change once the user takes an action, such as lend, where `hasLentBeforeCookie` would change from `false` to `true`?

src/components/WwwFrame/WwwPage.vue
```diff
@@ -64,17 +79,38 @@ export default {
...
-		},
+		}
 	},
 	created() {
+		let userData = {};
+		const currentHasLentBeforeValue = this.cookieStore.get(hasLentBeforeCookie);
+		const currentHasDepositBeforeValue = this.cookieStore.get(hasDepositBeforeCookie);
+
+		if (!currentHasLentBeforeValue || !currentHasLentBeforeValue) {
+			userData = this.apollo.readQuery({
```
dyersituations, line 101: Because the `preFetch` doesn't have the same check, I think this `readQuery` will encounter an error when this is called. The first question is: do we need to update the cookie value? If so, then we'll need to make sure the query also runs in the `preFetch` when needed here.

edwinf-barreram: Yeah, that is what I am working on right now!

src/components/WwwFrame/WwwPage.vue
```diff
@@ -28,6 +28,21 @@ import TheFooter from './TheFooter';
...
+	my {
+		userAccount {
+			id
+			isFirstTimeDepositor
+		}
+		loans {
+			totalCount
+		}
+	}
+}`;
```
emuvente, line 44: isFirstTimeDepositor is only true for people with only 1 deposit. Also, the `loans` endpoint is fetching 20 loans by default, which may slow things down. I would suggest checking the `transactions` endpoint to get the total number of deposits and also limit all results:
```graphql
{
  my {
    loans(limit:1) {
      totalCount
    }
    transactions(limit:1, filter:{category:deposit}) {
      totalCount
    }
  }
}
```

edwinf-barreram: Nice suggestion, thank you!

edwinf-barreram: Done!

src/components/WwwFrame/WwwPage.vue
```diff
@@ -70,11 +84,28 @@ export default {
 				fetchAllExpSettings(config, client, {
 					query: route?.query,
 					path: route?.path
-				})
+				}),
+				client.query({ query: optimizelyUserDataQuery })
```
mcstover, line 88: If there are no guards around this query, the data will be requested on every page load across the Ui application.

I think we need the check you previously had in place followed by wrapping the `readQuery` call in created in a try/catch statement. [Here's an example](https://github.com/kiva/ui/blob/main/src/pages/Lend/LoanChannelCategoryExperiment.vue#L875)

mcstover (approved):

mcstover (changes requested):
See my comment about checks before the prefetch query and using try/catch around the readQuery

src/pages/Thanks/ThanksPage.vue
```diff
@@ -274,6 +275,9 @@ export default {
 			.filter(item => item.basketItemType === 'loan_reservation')
 			.map(item => item.loan);
 
+		// MARS-194-User metrics A/B Optimizely experiment
+		handleUserMetricsCookies(this.cookieStore, this.loans, this.receipt);
```
mcstover, line 279: Could we pass only the loan count and the deposit total to this method instead? Seems like a whole lot of data to pass for 2 very small data points.

src/util/userMetricsCookie.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * This is a helper function to update cookies values for user metrics.
+ * it is going to be used for Optimizely A/B experiment.
+ * @param cookieStore
```
dyersituations, line 4: Recommend adding descriptions to these param comments.

src/util/userMetricsCookie.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param loans
+ * @param receipt
+ */
+export default function handleUserMetricsCookies(cookieStore, loans, receipt) {
+	const depositTotal = receipt?.totals?.depositTotals?.depositTotal;
+	const hasLentBefore = loans?.length > 0;
+	const hasDepositBefore = parseFloat(depositTotal) > 0;
+
+	if (hasLentBefore || hasDepositBefore) {
+		cookieStore.remove('kvu_lb');
```
dyersituations, line 14: Now that there is this dedicated file, recommend creating 2 new constants for these cookie names. Using a shared variable is generally preferred over magic strings.

For example, in this method at the top: `const lentBeforeCookieKey = 'kvu_lb';`

src/components/WwwFrame/TheHeader.vue
```diff
@@ -476,6 +478,20 @@ import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
 import SearchBar from './SearchBar';
 import PromoCreditBanner from './PromotionalBanner/Banners/PromoCreditBanner';
 
+const hasLentBeforeCookie = 'kvu_lb';
```
dyersituations, line 481: If you add these to the util file, import and use instead of these two constants.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -597,6 +624,24 @@ export default {
 			}
 		}
 	},
+	created() {
+		if (!this.isVisitor) {
+			try {
+				let userData = {};
```
dyersituations, line 630: You can just use a const here instead of having two lines.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -597,6 +624,24 @@ export default {
...
+				userData = this.apollo.readQuery({
```
dyersituations, line 631: This will throw an error if it wasn't called in the `preFetch`. Generally you don't want to create code that has _expected_ exceptions. Instead the `preFetch` calling should match the `created`. For example, this should never throw an exception unless there is some kind of expected network error.

mcstover: My suggestion here it to perform the same cookie check you do in the prefetch as a gate to using readQuery. If either are undefined, you would have prefetched the query.

src/pages/Thanks/ThanksPage.vue
mcstover: This is not a blocking suggestion, just a potential optimization to clarify what data we pass and how it's used.

src/components/WwwFrame/TheHeader.vue
dyersituations: The edge case was when a user has the cookies, then takes an action that would change the value saved to the cookie. Can we ignore that use case? If so, it simplifies this logic quite a bit.

mcstover: The actions that would change these cookies are confined to the new "thanks" contexts. This "global" query will only run when the user is logged in and does not have both cookies already set.

dyersituations: Sounds like we aren't worried about changing the cookie value, that simplifies things a little. CC: @edwinf-barreram

mcstover (approved):
I think the additional undefined cookie checks before call readQuery is important and @dyersituations 's suggestion about the constants is a good optimization to make here too. I'm approving to unblock that I put in with my prior comments which have been resolved.

edwinf-barreram:
@emuvente @mcstover Guys I have added some changes to use the Optimizely API features directly and not depend on the cookies value. The reference to it is [here](https://support.optimizely.com/hc/en-us/articles/4410289259277-Custom-Attributes-Capture-visitor-data-through-the-API-in-Optimizely)

mcstover (approved):
Thanks for all the updates

dyersituations (approved):

src/components/WwwFrame/TheHeader.vue
```diff
@@ -597,6 +625,31 @@ export default {
...
+					query: optimizelyUserDataQuery,
+				});
+
+				const hasLentBefore = userData?.my?.loans?.totalCount > 0;
+				const hasDepositBefore = userData?.my?.transactions?.totalCount > 0;
+
+				this.cookieStore.set(hasLentBeforeCookie, hasLentBefore);
+				this.cookieStore.set(hasDepositBeforeCookie, hasDepositBefore);
```
emuvente, line 647: `userHasLentBefore` and `userHasDepositBefore` should be called here too.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -572,18 +589,29 @@ export default {
...
-			this.basketCount = data?.shop?.nonTrivialItemCount ?? 0;
-			this.balance = Math.floor(data?.my?.userAccount?.balance ?? 0);
-			this.profilePic = data?.my?.lender?.image?.url ?? '';
-			this.profilePicId = data?.my?.lender?.image?.id ?? null;
-			this.basketState = data || {};
+		preFetch(config, client, { cookieStore }) {
+			return client.query({
+				query: headerQuery,
+			}).then(({ data }) => {
+				this.isVisitor = !data?.my?.userAccount?.id;
```
emuvente, line 596: `this` is not available in the prefetch function (it's run outside of the component context before the component is created). So we still need the `result` function and the `query: headerQuery` line. The rest of this `then` function will still work.

src/components/WwwFrame/TheHeader.vue
edwinf-barreram: done!

src/components/WwwFrame/TheHeader.vue
edwinf-barreram: done!

emuvente (approved):
👍 Looks great!

PR #4061, opened 2022-08-02
fix: only show expected FLSS sort options

dyersituations:
https://kiva.atlassian.net/browse/VUE-1180

- Instead of explicitly excluding, sort options are now explicitly included
- This will prevent this bug from happening the next time a sort option is released


mcstover (approved):
nice!

PR #4062, opened 2022-08-02
fix: update complete this loan price

egoroza:
updates complete this loan upper limit to $75 instead of $100

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -603,7 +603,7 @@ export default {
 		},
 		unreservedAmount(newValue, previousValue) {
 			// set initial selected value for sub 25 loan if shown
-			if (this.completeLoan && this.isBetween25And100) {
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -749,8 +749,8 @@ export default {
 		isLessThan25() {
 			return this.unreservedAmount < 25 && this.unreservedAmount > 0;
 		},
-		isBetween25And100() {
-			return this.unreservedAmount < 100 && this.unreservedAmount > 25;
+		isBetween25And75() {
+			return this.unreservedAmount < 75 && this.unreservedAmount > 25;
```
dyersituations, line 753: Do we want this to include `75` and `25`?

roger-in-kiva (approved):

src/components/BorrowerProfile/LendCta.vue
egoroza: nope exclusive!

PR #4065, opened 2022-08-04
"What is kiva" module in Borrower Profile (MARS-199)

christian14b:
This pr enables an experiment that shows a new modal within the Borrower Profile page for users that hasn't logged in before and may not have information about how kiva works.

image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "kvue",
-	"version": "2.243.0",
+	"version": "2.246.0",
 	"description": "The Kiva ui",
 	"author": "braincrave",
 	"private": true,
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -106,7 +106,7 @@
 				</content-container>
 			</div>
 		</article>
-		<what-is-kiva-modal v-if="kivaModuleExpEnabled && !shownModal" />
+		<what-is-kiva-modal v-if="true" />
 		<!-- <aside>Similar loans</aside> -->
 	</www-page>
 </template>
@@ -145,7 +145,7 @@ import loanUseFilter from '@/plugins/loan-use-filter';
 
 const socialElementsExpKey = 'social_elements';
 const whatIsKivaExpKey = 'what_is_kiva_module';
-
+const getPublicId = route => route?.query?.utm_content ?? route?.query?.name ?? '';
 const pageQuery = gql`
 	query borrowerProfileMeta($loanId: Int!, $publicId: String!, $getInviter: Boolean!) {
 		hasEverLoggedIn @client
@@ -261,6 +261,7 @@ export default {
 	metaInfo() {
 		const title = this.anonymizationLevel === 'full' ? undefined : this.pageTitle;
 		const description = this.anonymizationLevel === 'full' ? undefined : this.pageDescription;
+		const isSclePresent = this.$route.query?.utm_campaign?.includes('scle');
 
 		return {
 			title,
@@ -317,7 +318,20 @@ export default {
 					vmid: 'twitter:description',
 					content: this.shareDescription
 				},
-			])
+			]).concat(isSclePresent ? [
+				{
+					vmid: 'robots',
+					name: 'robots',
+					content: 'noindex',
+				},
+			] : []),
+			link: (isSclePresent ? [
+				{
+					vmid: 'canonical',
+					rel: 'canonical',
+					href: `https://${this.$appConfig.host}${this.$route.fullPath}`,
+				},
+			] : []),
 		};
 	},
 	data() {
@@ -359,13 +373,14 @@ export default {
 	apollo: {
 		query: pageQuery,
 		preFetch(config, client, { route }) {
+			const publicId = getPublicId(route);
 			return client
 				.query({
 					query: pageQuery,
 					variables: {
 						loanId: Number(route.params?.id ?? 0),
-						publicId: route.query?.utm_content ?? '',
-						getInviter: !!route.query?.utm_content
+						publicId,
+						getInviter: !!publicId
 					},
 				})
 				.then(({ data }) => {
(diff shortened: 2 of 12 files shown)


src/components/BorrowerProfile/WhatIsKivaModal.vue
```diff
@@ -0,0 +1,143 @@
...
+	},
+	data() {
+		return {
+			title: 'Welcome to Kiva! 💚',
+			description: 'Lend as little as $25 to make a big differences in someone\'s life.',
+			isLightboxVisible: true
+		};
+	},
+	methods: {
+		hideLightbox(type) {
```
mcstover, line 108: Pretty sure the param coming back will be an object containing `type`. So this would be better named as "payload" or "event" and then your code would access `type` on that object.

Also, need to merge your kv-ui-elements pr and update the version here.

dyersituations (approved):

eddieferrer:
How will this interact with the unbounce email capture modal? I believe that modal is still on the borrower profile

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -430,6 +441,17 @@ export default {
 				'EXP-MARS-158-Jul2022'
 			);
 		}
+
+		const kivaModuleExpData = getExperimentSettingCached(this.apollo, socialElementsExpKey);
```
eddieferrer, line 445: I think this should be `whatIsKivaExpKey` instead of `socialElementsExpKey`

eddieferrer:
I believe 2 of these icons can already be in src/assets/icons/inline: `lend-icon.svg` and `piggy-bank.svg` I think the repeat icon would be new

mcstover (approved):
The tracking update looks right. You should be able to easily verify in localhost (with dev env setting added) or full devvm

package.json
```diff
@@ -48,7 +48,7 @@
 		"@godaddy/terminus": "^4.11.0",
 		"@graphql-tools/load": "^7.7.0",
 		"@graphql-tools/url-loader": "^7.12.1",
-		"@kiva/kv-components": "^3.1.0",
+		"@kiva/kv-components": "^3.2.0",
```
mcstover, line 51: @christian14b I think you may need to run `npm i` so that this updates the package-lock file too.

mcstover: The release happened and this version is on npm https://www.npmjs.com/package/@kiva/kv-components

mcstover: I think/hope this will help github action pass...

eddieferrer:
I'm currently working on an experiment and I tried this same setup for tracking/reading value and fetching but `getExperimentSettingCached` is always returning undefined, had to go back to the previous experiment setup... 😕

mcstover:
@eddieferrer I'm wondering if you could more specific in your attempt to use this pattern for initializing an experiment? It's actually a bit of a hybrid seeing as it doesn't use the new `getExperimentSettingAsync` method at all.

In this PR, the setting is fetched in the pageQuery, then the exp is initialized in the 2nd stage of the prefetch. Then `getExperimentSettingCached` is called in the mounted hook. When were you calling the `getExperimentSettingCached` method? Created or Mounted? Did you setting exist to be read?

mcstover (approved):

eddieferrer:
> @eddieferrer I'm wondering if you could more specific in your attempt to use this pattern for initializing an experiment? It's actually a bit of a hybrid seeing as it doesn't use the new `getExperimentSettingAsync` method at all.
> 
> In this PR, the setting is fetched in the pageQuery, then the exp is initialized in the 2nd stage of the prefetch. Then `getExperimentSettingCached` is called in the mounted hook. When were you calling the `getExperimentSettingCached` method? Created or Mounted? Did you setting exist to be read?
> 
> I'm hoping to understand any constraints around these new methods as we're about to port them in the cms-page-server too!

I was using the same set up as this PR with my setting in Dev admin and running the UI via localhost, and also trying to force the setting via the setuiab url. I was calling getExperimentSettingCached in the mounted hook as well. My setting does exist  and I see it in my cookie value...

I will do some additional testing and put up a PR next week. The only difference I can see between my code and this PR is that i am working off localhost instead of the dev-vm

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -106,6 +106,7 @@
 				</content-container>
 			</div>
 		</article>
+		<what-is-kiva-modal v-if="kivaModuleExpEnabled && !hasEverLoggedIn" />
```
emuvente, line 109: With the way this is now this modal will be show for every visit to the borrower profile page until the visitor logs in. It should only be shown once. We should probably create a cookie after the modal has shown and only display the modal if the user does not have the cookie.

src/components/BorrowerProfile/WhatIsKivaModal.vue
```diff
@@ -0,0 +1,143 @@
...
+					</h2>
+					<div class="tw-mb-3">
+						<p
+							:data-testid="`bp-what-is-kiva-lightbox-description`"
+							class="md:tw-text-center tw-text-subhead"
+						>
+							{{ description }}
+						</p>
+					</div>
+					<h3 class="md:tw-hidden tw-text-action tw-mb-1">
```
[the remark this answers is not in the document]
emuvente: Ok, this is fine for now then.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -411,6 +424,7 @@ export default {
 			this.diffInDays = differenceInCalendarDays(parseISO(loan?.plannedExpirationDate), new Date());
 			this.hasThreeDaysOrLessLeft = this.diffInDays <= 3;
 			this.lender = result?.data?.my?.userAccount ?? {};
+			this.shownModal = !this.cookieStore.get('what-is-kiva-shown') && !result?.data?.hasEverLoggedIn;
```
emuvente, line 427: Shouldn't this be `this.cookieStore.get('what-is-kiva-shown') || result?.data?.hasEverLoggedIn`?

src/components/BorrowerProfile/WhatIsKivaModal.vue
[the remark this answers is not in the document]
mcstover: We haven't implemented any "normalization" to centering the header content in this version of the lightbox. I believe you can fix it though by adding padding to the left of your element(s) equivalent to the space taken up by the close X.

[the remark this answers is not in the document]
mcstover: That should at least allow you to move the content into the main content area. We took this approach with  `src/components/Checkout/AutoDepositUpsell.vue`

emuvente (approved):
Looks great! Thanks for the updates!

PR #4067, opened 2022-08-05
fix: add canonical url to Borrower Profile page

edwinf-barreram:
# Description
- Added canonical url to borrower profile page.

# Ticket
[MARS-206](https://kiva.atlassian.net/browse/MARS-206)

Lend to Alan in Bolivia | Kiva 🔊 2022-08-04 19-04-29

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -251,6 +251,7 @@ export default {
 	metaInfo() {
 		const title = this.anonymizationLevel === 'full' ? undefined : this.pageTitle;
 		const description = this.anonymizationLevel === 'full' ? undefined : this.pageDescription;
+		const isSclePresent = this.$route.query?.utm_campaign?.includes('scle');
 
 		return {
 			title,
@@ -307,7 +308,13 @@ export default {
 					vmid: 'twitter:description',
 					content: this.shareDescription
 				},
-			])
+			]).concat(isSclePresent ? [
+				{
+					vmid: 'canonical',
+					rel: 'canonical',
+					href: `https://${this.$appConfig.host}${this.$route.path}?${this.$route.query?.utm_campaign}`
+				},
+			] : [])
 		};
 	},
 	data() {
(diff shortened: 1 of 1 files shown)


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -307,7 +308,13 @@ export default {
 					vmid: 'twitter:description',
 					content: this.shareDescription
 				},
-			])
+			]).concat(isSclePresent ? [
+				{
+					vmid: 'canonical',
+					rel: 'canonical',
+					href: `https://${this.$appConfig.host}${this.$route.path}?${this.$route.query?.utm_campaign}`
```
mcstover, line 315: I think we need to include the entirety of the query param set here? @emuvente thoughts?

emuvente: It's best practice in canonical urls to only include parameters that change the content of the page, because a different page could be indexed for every different set of parameters listed in the canonical. If a bunch of pages are indexed that have the same content, the overall ranking of the base page will be lowered, because it's not unique content anymore.

That said, we could also use `noindex` to prevent indexing and then use the entire query string in the canonical url, which would future-proof this from any parameter naming changes and avoid anything besides the base url being indexed. I think that's probably the best way to go:
```js
]).concat(isSclePresent ? [
	{
		vmid: 'robots',
		name: 'robots',
		content: 'noindex',
	},
] : []),
link: (isSclePresent ? [
	{
		vmid: 'canonical',
		rel: 'canonical',
		href: `https://${this.$appConfig.host}${this.$route.fullPath}`,
	},
] : []),
```

mcstover: I think the approach you've outlined is the best way for now and I look forward to cleaning this out of the code.

edwinf-barreram:
@mcstover @emuvente @eddieferrer @bedo Approves are still missing here!

src/pages/BorrowerProfile/BorrowerProfile.vue
mcstover: @edwinf-barreram Please address the note above from @emuvente

edwinf-barreram: Done! @mcstover

mcstover (approved):
Thank you!

dyersituations (approved):

PR #4068, opened 2022-08-05
fix: fixed promise resolve

edwinf-barreram:
# Description
- Looks like  promise not being resolved correctly on the preFetch function

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -597,11 +597,8 @@ export default {
 				const hasLentBeforeValue = cookieStore.get(hasLentBeforeCookie);
 				const hasDepositBeforeValue = cookieStore.get(hasDepositBeforeCookie);
 
-				if (data?.my?.userAccount?.id && (hasLentBeforeValue === undefined || hasDepositBeforeValue === undefined)) { // eslint-disable-line max-len
-					return Promise.all([
-						client.query({ query: optimizelyUserDataQuery })
-					]);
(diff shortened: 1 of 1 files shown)


mcstover (approved):

src/components/WwwFrame/TheHeader.vue
```diff
@@ -597,11 +597,8 @@ export default {
 				const hasLentBeforeValue = cookieStore.get(hasLentBeforeCookie);
 				const hasDepositBeforeValue = cookieStore.get(hasDepositBeforeCookie);
 
-				if (data?.my?.userAccount?.id && (hasLentBeforeValue === undefined || hasDepositBeforeValue === undefined)) { // eslint-disable-line max-len
-					return Promise.all([
-						client.query({ query: optimizelyUserDataQuery })
-					]);
-				}
+				return data?.my?.userAccount?.id && (hasLentBeforeValue === undefined || hasDepositBeforeValue === undefined) // eslint-disable-line max-len
+					? client.query({ query: optimizelyUserDataQuery }) : Promise.resolve();
```
mcstover, line 601: When testing in dev we found that the user data from the `optimizelyUserDataQuery` query was not getting set into the Apollo state resulting in cookie values of `false` in all scenarios.

My suggestion was to perform the check inline and return the `optimizelyUserDataQuery` directly or resolve the promise if we aren't calling it.

emuvente (approved):
👍

dyersituations (approved):

PR #4075, opened 2022-08-09
feat: use the new flss version of lend/filter

mcstover:


src/router/routes.js
```diff
@@ -243,19 +243,17 @@ module.exports = [
 		},
 	},
 	{
+		name: 'lend-filter',
 		path: '/lend/filter',
-		component: () => import('@/pages/Lend/Filter/LendFilterPage'),
```
dyersituations, line 247: At some point do we want to delete the old components?

dyersituations (approved):

src/router/routes.js
mcstover: Yes we'll create some other tickets around the cleanup

src/router/routes.js
```diff
@@ -243,19 +243,17 @@ module.exports = [
...
+		component: () => import('@/pages/Lend/LoanSearchPage'),
 		meta: {
 			unbounce: true,
 		},
 	},
 	{
 		name: 'filter-alpha',
```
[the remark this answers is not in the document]
mcstover: maybe until the end of the month.

PR #4079, opened 2022-08-09
MARS-194: fix user logged attribute detection

edwinf-barreram:
# Description
- Added try catch block on set attributes values function

### DIFF
--- a/src/util/optimizelyUserMetrics.js
+++ b/src/util/optimizelyUserMetrics.js
@@ -1,17 +1,20 @@
 function setUserAttribute(key, value) {
-	if (typeof window === 'undefined') {
-		return;
+	try {
+		if (typeof window === 'undefined') {
+			return;
+		}
+
+		window.optimizely = window.optimizely || [];
+		window.optimizely.push({
+			type: 'user',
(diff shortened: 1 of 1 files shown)


src/util/optimizelyUserMetrics.js
```diff
@@ -1,17 +1,20 @@
...
-	});
+		window.optimizely = window.optimizely || [];
+		window.optimizely.push({
+			type: 'user',
+			attributes: {
+				[key]: value,
+			},
+		});
+	} catch (err) {
+		console.error(err);
```
dyersituations, line 15: We generally use `logFormatter` for errors, so `logFormatter(err, 'error');`

src/util/optimizelyUserMetrics.js
```diff
@@ -1,17 +1,20 @@
 function setUserAttribute(key, value) {
-	if (typeof window === 'undefined') {
-		return;
-	}
+	try {
```
dyersituations, line 2: Why is try catch needed?

edwinf-barreram:
Closing this, not necessary!

PR #4080, opened 2022-08-10
feat: add lbc eco category experimental layout

eddieferrer:
* Adds an experimental layout to the lbc page for eco categories
* Also removes old template for LoanChannelExperiment layout and the header images that were very briefly used in that previous experiment

ACK-357

Experiment to replace the layout of the `'eco-friendly-loans' and 'eco-friendly' lend-by-category pages with this new experimental layout. Instead of the current page, the page shows 5 loan card carousels with category titles and descriptions based on 5 new loan channels created for this purpose. 

Screenshot of the entire page:
screencapture-dev-vm-01-kiva-org-lend-by-category-eco-friendly-loans-2022-08-09-19_53_10


.storybook/stories/KivaClassicSingleCategoryCarousel.stories.js
```diff
@@ -0,0 +1,75 @@
+import KivaClassicSingleCategoryCarousel from '@/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue';
```
eddieferrer, line 1: Not a very good story since I am not mocking the loan card info

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -306,6 +306,10 @@ export default {
 		expLabel: {
 			type: String,
 			default: ''
+		},
+		lendNowButton: {
```
eddieferrer, line 310: Prop to display add directly to basket button instead of view loan. It doesn't feel great to pass this prop through KivaClassicSingleCategoryCarousel.vue > KivaClassicLoanCarousel.vue > /KivaClassicBasicLoanCard.vue but the previous approach (of checking the route for the funded borrower profile is also not very good.

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -0,0 +1,106 @@
+<template>
```
eddieferrer, line 1: Not sure if this deserves a new component, its very much like `KivaClassicMultiCategoryCarousel.vue` but only 1 category and the description and name of the category added. I tried using `KivaClassicMultiCategoryCarousel.vue` with a single category and having the description and name in the parent component, but the ideal UI is for the description and name to not show up until the carousel starts loading, which is why I went with a new component -- alternatively we could add some more props to the multi carousel to optionally display the description and name

mcstover: I'm glad you could at least reuse the carousel component. It has one major flaw right now which is that it needs a min height that matches the typical height of the loan card so that we don't have a giant layout shift when the content does finally load. Ideally we'd have a placeholder state for the name and description along with the loan cards.

In what scenario would a single row carousel not be loading? Lazy-loading down the page?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
mcstover: It think this an appropriate prop to add as it applies to "all" loan cards in the carousel and will be very useful! I can now add it as a config on the contenful version and we'll be able to use that in other places where we want folks to add to basket!

mcstover (approved):
I realize there is a lot of duplication to stand up this exp but for the short term that's ok. We can look to consolidate based on what we learn.

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -0,0 +1,106 @@
...
+					}
+				}`,
+				variables: {
+					loanChannelIds: this.loanChannelId,
+					loanLimit: this.loanQueryLimit
+				},
+			}).then(result => {
+				const loanChannelData = result?.data?.lend?.loanChannelsById ?? [];
+				// eslint-disable-next-line prefer-destructuring
+				this.selectedChannel = loanChannelData[0];
```
dyersituations, line 99: Might want to add optional chaining in case the fallback on line 97 ever gets hit: loanChannelData?.[0]

src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
```diff
@@ -0,0 +1,236 @@
...
+			&& to.path.indexOf('/lend/filter') === -1) {
+			// set cookie to signify redirect
+			this.cookieStore.set('redirectFromUi', true);
+		}
+		next();
+	}
+};
+</script>
+
+<style lang="scss" scoped>
```
dyersituations, line 220: I'm assuming this component is temporary, since there's scss usage, so I'll not comment on this file more.

dyersituations (approved):

src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
mcstover: @dyersituations we can probably cleanup this `heading-region` section to match the declarations from the LoanSearchPage header area. And do the same the LoanChannelCategoryControl... that way they would all match

mcstover: These kinds of things are good candidates for small cleanup tickets in Vue queue.

eddieferrer: This was just legacy CSS from LoanChannelPage, which shares the same heading region

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
eddieferrer: The page would load all the titles and the descriptions, which looks weird, and then there would be at least a 1sec delay before the loading state carousels showed up. This is due to the fetchLoans call resolving

PR #4081, opened 2022-08-10
Categories testing

galen94:
Cypress testing for CategoriesBeta page


src/pages/Categories/CategoriesBeta.vue
```diff
@@ -15,6 +15,7 @@
 			<kv-page-container>
 				<kv-grid>
 					<loan-spotlight
+						data-testid="all-categories-loan-spotlight"
```
eddieferrer, line 18: Please add the `data-testid` inside of the components, this will avoid duplicating them here and they will be available for future uses of the component

test/e2e/specs/CategoriesBetaPage.spec.js
```diff
@@ -0,0 +1,35 @@
...
+		cy.intercept('POST', '**/graphql*').as('graphqlRequest');
+
+		// Go to all categories beta page
+		cy.visit('/categories-beta');
+
+		// Wait for page buttons to exist
+		cy.get('@graphqlRequest.all');
+		cy.findAllByText('View all loans');
+		cy.findAllByText('Get started');
+		cy.wait(1000);
```
eddieferrer, line 13: Please remove this wait. The test is probably failing without it because of the query after the page loads to fetch the loan spotlight, but waiting an arbitrary amount of time makes this test very fickle and it will fail if that query takes longer than 1 sec. 

A better approach is to wait on the graphql query for the loan spotlight, see the example here: https://github.com/kiva/ui/blob/main/test/README.md about waiting for aliased, multiple queries.

eddieferrer (commented):
Nice job, just a couple of minor things. 👍

eddieferrer (approved):
looks great!

PR #4083, opened 2022-08-10
feat: update upsells checkout styles

egoroza:
- fix an issue on upsell module where flex components wrap with experiment copy
- adds a min-height to prevent moving checkout component from being pushed down when upsell module is rendered

Screen Shot 2022-08-10 at 3 04 48 PM
Screen Shot 2022-08-10 at 3 04 58 PM

Screen Shot 2022-08-11 at 10 36 36 AM
Screen Shot 2022-08-11 at 10 36 42 AM
Screen Shot 2022-08-11 at 10 39 53 AM
Screen Shot 2022-08-11 at 10 40 03 AM


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -920,6 +920,10 @@ export default {
 <style lang="scss">
 @import 'settings';
 
+.basket-list-container {
+	min-height: 600px;
```
mcstover, line 924: The min-height will need to be applied to your new upsell container rather than the item list. This list can vary well beyond 600 and would still have the issue of the loan upsell popping in and expanding it's height.

mcstover (approved):

dyersituations (approved):

PR #4089, opened 2022-08-11
Vue 1218 lend filter url process

mcstover:


src/plugins/loan-channel-query-map.js
```diff
@@ -53,7 +53,7 @@ export default {
 					url: 'livestock',
 					queryParams: 'status=fundRaising&riskRating=0,5&activity=73',
 					// algoliaParams: '', // not supported pass exclude link
-					fallbackUrl: '/lend?status=fundRaising&riskRating=0,5&activity=73&kexpn=lend_filter.lend_filter_versions&kexpv=c'
+					fallbackUrl: '/lend?status=fundRaising&riskRating=0,5&activity=73'
```
mcstover, line 56: Removing the old exp activation for fallback as you can now just hit lend without the override.

src/plugins/loan-channel-query-map.js
```diff
@@ -65,7 +65,7 @@ export default {
 				{
 					id: 3,
 					url: 'ending-soon',
-					queryParams: 'status=fundRaising&riskRating=0,5&expiringSoon=1',
+					queryParams: 'status=fundRaising&riskRating=0,5&sortBy=expiringSoon',
```
mcstover, line 68: This makes the query params compatible with lend/filter

src/plugins/loan-channel-query-map.js
```diff
@@ -238,7 +238,7 @@ export default {
 				{
 					id: 60,
 					url: 'loans-that-are-almost-funded',
-					queryParams: 'status=fundRaising&riskRating=0,5',
+					queryParams: 'status=fundRaising&riskRating=0,5&sortBy=amountLeft',
```
mcstover, line 241: amountLeft is now enabled on lend/filter flss so this makes it compatible.

src/util/loanSearch/queryParamUtils.js
```diff
@@ -10,8 +10,25 @@ const lendToFlssSort = new Map([
...
 
+/**
+ * Check for excluded query params
+ *
+ * @param {Object} query
+ * @returns {boolean}
+ */
+export function hasExcludedQueryParams(query) {
+	// Handle temporary query param exclusions
+	const excludedParams = ['loanTags', 'partner', 'activity', 'lenderTerm'];
```
mcstover, line 24: There are most definitely some others we should add. However, this first pass is against the categories shown in our lend drop down.

I'll test some more in dev against categories showing up in lend-by-category

src/util/loanSearch/queryParamUtils.js
```diff
@@ -23,12 +40,25 @@ const lendToFlssSort = new Map([
...
+
+				if (facet) {
+					prev.push(facet.id);
+				}
+			} else if (current !== '') {
+				return [...prev, parseInt(current, 10)];
+			}
+
+			return prev;
+		}, []);
```
mcstover, line 58: @dyersituations this is certainly uglier than your original process for comma separated values but enabled us to support lists of theme names.

src/util/loanSearch/queryParamUtils.js
```diff
@@ -97,7 +127,10 @@ export async function applyQueryParams(apollo, query, allFacets, queryType, page
 		countryIsoCode: getCountryIsoCodesFromQueryParam(query.country || query.countries, allFacets),
 		sectorId: getIdsFromQueryParam(query.sector, allFacets.sectorNames, allFacets.sectorFacets),
 		sortBy: queryType === FLSS_QUERY_TYPE ? lendToFlssSort.get(query.sortBy) : query.sortBy,
-		themeId: getIdsFromQueryParam(query.attribute || query.attributes, allFacets.themeNames, allFacets.themeFacets),
+		themeId: getIdsFromQueryParam(
+			query.attribute || query.attributes || query.theme,
```
mcstover, line 131: Check `query.theme` on incoming query params. This will be converted to attribute ids!

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -285,8 +291,8 @@ export default {
 					await this.fetchFacets(this.loanSearchState)
 				]);
 
-				// Store loan data in component
-				this.loans = loans;
+				// Store loan data in component, guarding against null loan objects
```
dyersituations, line 294: When does this happen?

src/plugins/loan-channel-query-map.js
dyersituations: This channel will need a `flssLoanSearch` if want it to be served by FLSS.

src/util/loanSearch/queryParamUtils.js
dyersituations: When are there lists of theme names? I thought it was comma separated IDs or names with ~

dyersituations (approved):

eddieferrer:
Great cleanup 🧹 

Could we also add the algolia url cleanup from `src/pages/Lend/LoanChannelCategoryControl.vue` to `LoanChannelCategoryRecommendedByLenders.vue` and `LoanChannelCategoryClimateExperiment.vue`

src/components/Lend/LoanSearch/LoanSearchInterface.vue
mcstover: This can happen in lower envs and the vm...I'm on the fence as to if we should keep this in.

When the personalized sort gets out of sync with the db loans may not exist causing a null value to be returned instead of a loan object. When this happens no loans are shown on the page. The page loads but in a broken state.

src/util/loanSearch/queryParamUtils.js
mcstover: There was at least one category in the query map that had 2 theme names so this was really to support that. A single name already worked with the small addition below!

src/plugins/loan-channel-query-map.js
mcstover: Good call. We'll do another pass soon on this file soon to solidify the FLSS queries on our category page and we can include this category then.

mcstover:
> Great cleanup 🧹 
> 
> 
> 
> Could we also add the algolia url cleanup from `src/pages/Lend/LoanChannelCategoryControl.vue` to `LoanChannelCategoryRecommendedByLenders.vue` and `LoanChannelCategoryClimateExperiment.vue`

Yessir, I'll add to those as well this afternoon

src/util/loanSearch/queryParamUtils.js
dyersituations: Gotcha. I wonder if that was an accident. No worries supporting a different use case though!

mcstover: Yea, a single theme name would go into the final process and get matched to the facet! Figured if it worked there, I could use for the comma separated process too...

I suppose in the future we could actually extract the query value prep and use a single reduce process for all formats.

mcstover:
@eddieferrer I've updated the 2 pages you've specified and also included the exp mixin call that wasn't present. I think that should be fine in all context for now. I'm also working on strategy to remove the lend_filter_v2 exp altogether soon.

eddieferrer (approved):

PR #4091, opened 2022-08-11
feat: show upsells module only if under 100

egoroza:
show upsell module only if it is less than $100

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -534,7 +534,7 @@ export default {
 		isUpsellUnder100() {
 			const amountLeft = this.upsellLoan?.loanAmount
 			- this.upsellLoan?.loanFundraisingInfo?.fundedAmount
-			- this.upsellLoan?.loanFundraisingInfo?.reservedAmount;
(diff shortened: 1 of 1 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -528,6 +529,12 @@ export default {
 		this.getUpsellModuleData();
 	},
 	computed: {
+		isUpsellUnder100() {
+			const amountLeft = this.upsellLoan?.loanAmount
+			- this.upsellLoan?.loanFundraisingInfo?.fundedAmount
+			- this.upsellLoan?.loanFundraisingInfo?.reservedAmount;
```
dyersituations, line 535: Fallback numbers (like 0) will be needed if these props are indeed ever undefined. There's optional chaining here, but I'm not sure if they are ever undefined in this situation. A number - undefined = NaN.

egoroza: yeah this seems to just be the standard across the ui repo for `amountLeft` but added a `0` fallback 👍

roger-in-kiva (approved):

PR #4096, opened 2022-08-11
feat: swap CampaignHero component for the newer Dynamic Hero Classic …

michelleinez:
…component

### DIFF
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -23,14 +23,14 @@
 
 			<!-- TODO: Alter CTA if Checkout is ready -->
 			<campaign-hero
-				v-show="pageData.page.contentGroups.mlCampaignHero"
+				v-if="hasCampaignHero"
 				:hero-area-content="heroAreaContent"
 				@add-to-basket="handleAddToBasket"
 				@jump-to-loans="jumpToLoans"
 			/>
 
 			<dynamic-hero-classic
-				v-show="pageData.page.contentGroups.dynamicHeroClassic"
+				v-if="hasDynamicHeroClassic"
 				:content="heroAreaContent"
 			/>
 
@@ -244,8 +244,6 @@ import syncDate from '@/util/syncDate';
 import trackTransactionEvent from '@/util/trackTransactionEvent';
 import checkoutUtils from '@/plugins/checkout-utils-mixin';
 import updateLoanReservationTeam from '@/graphql/mutation/updateLoanReservationTeam.graphql';
-import CampaignHero from '@/components/CorporateCampaign/CampaignHero';
-import DynamicHeroClassic from '@/components/Contentful/DynamicHeroClassic';
 import CampaignHowKivaWorks from '@/components/CorporateCampaign/CampaignHowKivaWorks';
 import CampaignJoinTeamForm from '@/components/CorporateCampaign/CampaignJoinTeamForm';
 import CampaignLoanGridDisplay from '@/components/CorporateCampaign/CampaignLoanGridDisplay';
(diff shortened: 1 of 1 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -23,11 +23,17 @@
 
 			<!-- TODO: Alter CTA if Checkout is ready -->
 			<campaign-hero
+				v-show="pageData.page.contentGroups.mlCampaignHero"
```
mcstover, line 26: Did you try to use `v-if` for these two visibility checks?

Also, it will be more safe to either check a direct value (maybe type) on `heroAreaContent` or create a computed property that returns a boolean so you can use a v-if/v-else on these two components.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -603,6 +611,8 @@ export default {
 		this.loadingPage = basketItems.some(item => item.__typename === 'LoanReservation'); // eslint-disable-line no-underscore-dangle, max-len
 	},
 	mounted() {
+		this.$root.$on('handleAddToBasket', this.handleAddToBasket);
+		this.$root.$on('jumpToLoans', this.jumpToLoans);
```
mcstover, line 615: Are these necessary?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
mcstover: The optional chaining within the js below is def the way to go with contentful based data. Ui doesn't have things setup to allow that in the template though.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: jumpToLoans is necessary to be an event handler because the event handler that the button in the hero section triggers is specified in contentful for DynamicHeroClassic.
Add to basket isn't really necessary for this PR though, I can remove it.  I imagine I might need that event for when I swap the loan cards out for the classic loan cards but I haven't really gotten there yet so it'll be cleaner to just add it in that PR when I get to it.

mcstover: Ahh yes, makes sense now. Shouldn't be a problem to keep them both here

mcstover: In other contexts, we manage these events at a component level and turn them off on unmount... I think this is ok  to just activate in this context for now b/c folks generally do not leave this page which might leave these events active.

dyersituations: @mcstover Where/how are these events triggered?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -239,6 +245,7 @@ import trackTransactionEvent from '@/util/trackTransactionEvent';
 import checkoutUtils from '@/plugins/checkout-utils-mixin';
 import updateLoanReservationTeam from '@/graphql/mutation/updateLoanReservationTeam.graphql';
 import CampaignHero from '@/components/CorporateCampaign/CampaignHero';
+import DynamicHeroClassic from '@/components/Contentful/DynamicHeroClassic';
```
mcstover, line 248: As a potential optimization, you might want to try and use dynamic imports for both hero components as well...

```
ex. const CampaignHero = () => import('@/components/CorporateCampaign/CampaignHero');
```

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
eddieferrer: @michelleinez please also add a beforeDestroy hook to remove these events when the component is destroyed. See `MonthlyGoodSelectorDesktop.vue` for a good example of this

eddieferrer: @dyersituations This comes from contentful. There is an option to add a web event name to a button in contentful, in this case you would add the `jumpToLoans` as the value

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: done

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: added the hook

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: done

PR #4098, opened 2022-08-12
fix: update loan channel query map with missing categories, adapt for…

mcstover:
… flss, add new excluded params

Went through ALL of our loan channels and added several that we didn't have mappings for. This is required as many of the previously unused channels are getting picked up and inserted by the bandit into Lend By Category.

Without these mappings they would just land on lend/filter without any query params to setup the loan search.

Within this pass, I've also matched up query params against those in the prod api and cleaned up some unnecessary params where they wouldn't be needed. _We'll def clean up some more unused query params over time! Especially the "fundraising" status!_

### DIFF
--- a/src/plugins/loan-channel-query-map.js
+++ b/src/plugins/loan-channel-query-map.js
@@ -10,7 +10,7 @@ export default {
 				{
 					id: 5,
 					url: 'women',
-					queryParams: 'status=fundRaising&gender=female&riskRating=0,5',
+					queryParams: 'status=fundRaising&gender=female',
 					algoliaParams: 'gender=female',
 					flssLoanSearch: {
 						gender: 'female'
@@ -19,7 +19,7 @@ export default {
 				{
 					id: 8,
 					url: 'agriculture',
-					queryParams: 'status=fundRaising&riskRating=0,5&sector=1',
+					queryParams: 'status=fundRaising&sector=1',
 					algoliaParams: 'sector=Agriculture',
 					flssLoanSearch: {
 						sectorId: [1]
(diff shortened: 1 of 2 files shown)


dyersituations (approved):

dyersituations (approved):

src/util/loanSearch/queryParamUtils.js
```diff
@@ -23,10 +23,18 @@ export function hasExcludedQueryParams(query) {
...
+		'defaultRate',
+		// We don't yet officially support this query param, but will soon
+		// Dropping it doesn't have any ill affect on the most popular categories
+		// If problems are reported we can uncomment this one to redirect to legacy
+		// 'distributionModel',
 		'isGroup',
 		'lenderTerm',
 		'loanTags',
 		'partner',
+		'riskRating',
```
mcstover, line 36: As seen above, I've removed all of the default riskRating query params leaving only those that modify the range, which will get redirected to legacy lend.

src/util/loanSearch/queryParamUtils.js
```diff
@@ -23,10 +23,18 @@ export function hasExcludedQueryParams(query) {
 	// Handle temporary query param exclusions
 	const excludedParams = [
 		'activity',
+		'city_state',
+		'defaultRate',
+		// We don't yet officially support this query param, but will soon
+		// Dropping it doesn't have any ill affect on the most popular categories
+		// If problems are reported we can uncomment this one to redirect to legacy
+		// 'distributionModel',
```
mcstover, line 31: As noted, I've not seen any big issues with this when landing on the flss version of lend/filter. Typically the combination of query params produces a proper loan search state. ie. A US loan search includes query params with the countryIsoCode so the result is correct.

src/plugins/loan-channel-query-map.js
```diff
@@ -10,7 +10,7 @@ export default {
 				{
 					id: 5,
 					url: 'women',
-					queryParams: 'status=fundRaising&gender=female&riskRating=0,5',
+					queryParams: 'status=fundRaising&gender=female',
```
mcstover, line 13: As noted below, I've removed all the default riskRating query params leaving only those that alter the range. Altered riskRatings will be redirected to legacy lend while the rest can be served by flss lend/filter

PR #4099, opened 2022-08-15
fix: typo in channel 90, add mixin for querymap on eco friendlly, cus…

mcstover:
…tom route url for recommended

### DIFF
--- a/src/plugins/loan-channel-query-map.js
+++ b/src/plugins/loan-channel-query-map.js
@@ -484,7 +484,7 @@ export default {
 				},
 				{
 					id: 90,
-					url: 'eco-friendly-loans',
+					url: 'ecofriendly-loans',
 					queryParams: 'status=fundRaising&gender=female&loanTags=9&distributionModel=both',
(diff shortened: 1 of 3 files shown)


dyersituations (approved):

src/plugins/loan-channel-query-map.js
```diff
@@ -484,7 +484,7 @@ export default {
 				},
 				{
 					id: 90,
-					url: 'eco-friendly-loans',
```
eddieferrer, line 487: Thanks, I'll add the other 3 eco channels. 

In our code, we are being specifically agnostic about the channel id's since they could hypothetically differ in prod vs dev. Doesn't this loan-channel-query-map now bind the id and channel url together to the values in prod?

mcstover: Loan channel ids should not differ in prod vs dev and they will not for more than 2 weeks. Please keep this file synced with prod.

mcstover: I'm not sure which code you're referring to here `In our code`... but we should try to clean it up to not diverge per environment.

PR #4104, opened 2022-08-16
feat: lbc experimental layout modifications

eddieferrer:
* Fix loading behavior on LBC control layout
* Add experiment routes to loan channel query map
* fix experiment creation on navigation for lbc template
* add view all loans card back to experiment layout
* modified page background color
* added better loading for category description and carousels

ACK-364

### DIFF
--- a/src/components/LoanCollections/KivaClassicLoanCarousel.vue
+++ b/src/components/LoanCollections/KivaClassicLoanCarousel.vue
@@ -3,7 +3,7 @@
 		<transition name="kvfade">
 			<div
 				v-if="isLoading"
-				class="spinner"
+				class="spinner tw-text-center"
 			>
 				<kv-loading-spinner />
 			</div>
@@ -22,27 +22,9 @@
 			@interact-carousel="onInteractCarousel"
 		>
 			<template v-for="(loanId, index) in augmentedLoanIds" #[`slide${index}`]>
-				<!-- Show View more Card -->
-				<router-link
-					v-if="loanId === 3"
-					:key="`view-more-card-${loanId}`"
-					class="tw-flex tw-items-center tw-h-full tw-w-full
-						hover:tw-bg-action-highlight hover:tw-text-primary-inverse tw-rounded"
-					:to="cleanUrl"
-					v-kv-track-event="[
-						'Homepage',
-						'click-carousel-view-all-category-loans',
-						`${viewAllLoansCategoryTitle}`]"
-				>
-					<div class="tw-w-full tw-text-center">
-						<h3>{{ viewAllLoansCategoryTitle }}</h3>
-					</div>
-				</router-link>
-
 				<!-- show loan card -->
 				<!-- TODO Re-implement card position analytics -->
 				<kiva-classic-basic-loan-card
-					v-else
 					:item-index="index"
 					:key="`loan-${loanId}`"
 					:loan-id="loanId"
@@ -50,6 +32,22 @@
 					:lend-now-button="lendNowButton"
 				/>
 			</template>
+			<!-- Show View more Card -->
+			<router-link
+				v-if="showViewMoreCard"
+				:key="`view-more-card`"
+				class="tw-flex tw-items-center tw-h-full tw-w-full
+						hover:tw-bg-action-highlight hover:tw-text-primary-inverse tw-rounded"
+				:to="cleanUrl"
+				v-kv-track-event="[
+					'Lending',
+					'click-carousel-view-all-category-loans',
+					`${viewAllLoansCategoryTitle}`]"
+			>
+				<div class="tw-w-full tw-text-center">
+					<h3>{{ viewAllLoansCategoryTitle }}</h3>
+				</div>
+			</router-link>
 			<div
 				v-if="showCheckBackMessage" class="tw-flex tw-items-center tw-h-full tw-w-full
 					tw-border-action-highlight tw-rounded"
@@ -132,7 +130,6 @@ export default {
 			const clonedLoanIds = [...this.loanIds];
 			// const promoCardId = 1;
 			// const loadMoreCardId = 2;
-			const viewMoreCardId = 3;
 			// TODO: splice if promoCard if active on row
 			// if (this.showPromoCard) {
 			// 	clonedLoanIds.splice(1, 0, promoCardId);
@@ -142,11 +139,6 @@ export default {
 			// 	clonedLoanIds.push(loadMoreCardId);
 			// 	return clonedLoanIds;
 			// }
-			// append viewMoreCard if active
-			if (this.showViewMoreCard) {
-				clonedLoanIds.push(viewMoreCardId);
-				return clonedLoanIds;
-			}
 			return clonedLoanIds;
 		},
 		cleanUrl() {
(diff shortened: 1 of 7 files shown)


src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -54,16 +73,17 @@ export default {
...
 		};
 	},
 	computed: {
 		loanQueryLimit() {
 			return this.loanDisplaySettings?.loanLimit ?? 1;
 		},
 		showViewMoreCard() {
-			return this.loanDisplaySettings?.showViewMoreCard ?? false;
+			if (this.loanDisplaySettings?.showViewMoreCard) {
+				return this.loanQueryLimit === this.selectedChannelLoanIds.length;
```
eddieferrer, line 84: Only show view more card if our loan length is equal to query limit, so we can assume there are more loans.

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -74,6 +94,14 @@ export default {
 		selectedChannelLoanIds() {
 			return this.selectedChannel?.loans?.values?.map(loan => loan.id) ?? [];
 		},
+		name() {
```
eddieferrer, line 97: We can use these to pass is the name/description to show them instantly if the parent component knows them while this component does it's loading

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -261,7 +263,7 @@ export default {
 			filters: { },
 			targetedLoanChannelURL: null,
 			targetedLoanChannelID: null,
-			loanChannel: () => {},
+			loanChannel: {},
```
eddieferrer, line 266: This should not be a function and returns a weird value before its set while loading

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -150,6 +150,8 @@
 				</div>
 			</div>
 		</div>
+
+		<kv-loading-overlay v-if="loading" />
```
eddieferrer, line 154: The loading states were not working properly on this page. This adds 1 loader, which displays inside the relative position wrapper div. Loader is visible, when this page loads without prefetch (linking from lbc experiment layout) or when changing pages. it shows for categories with the MG promo card and those without

src/components/LoanCollections/KivaClassicLoanCarousel.vue
```diff
@@ -3,7 +3,7 @@
 		<transition name="kvfade">
 			<div
 				v-if="isLoading"
-				class="spinner"
+				class="spinner tw-text-center"
```
eddieferrer, line 6: centers spinner

src/components/LoanCollections/KivaClassicLoanCarousel.vue
```diff
@@ -22,34 +22,32 @@
...
-					v-else
 					:item-index="index"
 					:key="`loan-${loanId}`"
 					:loan-id="loanId"
 					:exp-label="expLabel"
 					:lend-now-button="lendNowButton"
 				/>
 			</template>
+			<!-- Show View more Card -->
+			<router-link
```
eddieferrer, line 36: we can just move this view more card outside of the loop of loans, and clean up the code that assigns it an arbitrary loan id of 3.

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -1,19 +1,21 @@
...
-			:is-visible="showCarousel"
-			:loan-ids="selectedChannelLoanIds"
-			:selected-channel="selectedChannel"
-			:show-view-more-card="showViewMoreCard"
-			:lend-now-button="lendNowButton"
-			:show-check-back-message="showCheckBackMessage"
-		/>
+		<div>
+			<kiva-classic-loan-carousel
+				:is-visible="true"
```
eddieferrer, line 11: The carousel actually loads much nicer if we just set this to visible. Since it will go instantly into the loading state.

src/components/LoanCollections/KivaClassicLoanCarousel.vue
```diff
@@ -22,34 +22,32 @@
...
+				v-if="showViewMoreCard"
+				:key="`view-more-card`"
+				class="tw-flex tw-items-center tw-h-full tw-w-full
+						hover:tw-bg-action-highlight hover:tw-text-primary-inverse tw-rounded"
+				:to="cleanUrl"
+				v-kv-track-event="[
+					'Homepage',
```
mcstover, line 43: Maybe we should make this more generic or dynamic based on the page path???

src/components/LoanCollections/KivaClassicLoanCarousel.vue
mcstover: I don't remember why were were even doing this...

src/components/LoanCollections/KivaClassicLoanCarousel.vue
eddieferrer: Changed to 'Lending' which I think fits with the new event guidelines

src/plugins/loan-channel-query-map.js
```diff
@@ -397,6 +397,29 @@ export default {
...
+				},
+				{
+					id: 118,
+					url: 'recycle-and-re-use',
+					queryParams: 'queryString=used clothing&distributionModel=field_partner&sortBy=popularity',
+				},
+				{
+					id: 119,
+					url: 'other-eco-friendly-loans',
+					queryParams: 'tag=9,8&distributionModel=field_partner&sortBy=expiringSoon',
```
mcstover, line 420: shouldn't `tag` be `loanTags` ?

src/pages/Lend/LoanChannelCategoryControl.vue
mcstover: Wonder if this will affect reactivity when moving between categories...ie lbc to category to lbc and back to a different category. or even clicking into one of your exp categories...

mcstover (approved):

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -375,6 +377,7 @@ export default {
 		}
 	},
 	created() {
+		this.loading = true;
```
eddieferrer, line 380: When we land here from the experiment layout, the component won't prefetch, this loading will set the loading state to true. Then base data will miss the cache, so the loading state will stay on until the observer finished updating and loads the loans. loading is then clears in the `activateLoanChannelWatchQuery` next() function below. 

This addition adds the loader when coming from the experiment layout and the text "No loans" is displayed. It does not seem to affect normal operation of the page (coming from a link, with prefetch)

src/plugins/loan-channel-query-map.js
eddieferrer: I don't think so -- here is the link for solar: https://www.kiva.org/lend?tag=9&queryString=solar&sortBy=popularity

mcstover: ok so long as it's working properly in `/lend` it should be ok. Could you please also add this to the excludeParams

dyersituations (approved):

src/plugins/loan-channel-query-map.js
mcstover: Ok so both `tag` and `loanTags` works 🤷

PR #4113, opened 2022-08-17
feat: finish deprecating Causes product

eddieferrer:


mcstover:
@eddieferrer this generally seems ok. Is there any need for the Cancellation flow or can any remaining users cancel in a different way.

eddieferrer:
There is no way to cancel causes anymore. My understanding is that the users were migrated months ago -- so today, there should be 0 users with an active Causes subscriptions. I can verify that before merging this in.

eddieferrer:
@mcstover For reference this is the backend ticket: https://kiva.atlassian.net/browse/CORE-519

mcstover (approved):

PR #4115, opened 2022-08-17
feat: activate saved search at it's actual url

mcstover:


src/router/routes.js
```diff
@@ -501,13 +501,17 @@ module.exports = [
...
-		path: '/lend/saved-search-beta',
+		path: '/lend/saved-search',
 		component: () => import('@/pages/Settings/SavedSearchBeta'),
 		meta: {
 			authenticationRequired: true,
 			excludeFromStaticSitemap: true,
 		}
 	},
+	{
+		path: '/lend/saved-search-beta',
```
dyersituations, line 512: Do we need this fallback?

dyersituations (approved):

src/router/routes.js
mcstover: eh prob not I guess it was only @tiarascoleman who may have memorized that one...

mcstover: I'll clean it up here and also the Prod version of the puppet configs.

tiarascoleman (approved):
Yay!

PR #4116, opened 2022-08-17
fix: ensure external links open in new tabs

michelleinez:


src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -158,6 +159,10 @@ export default {
 			default: () => {},
 		},
 	},
+	async mounted() {
+		await this.$nextTick();
+		addBlankTargetToExternalLinks(this.$refs.heroBodyCopy);
```
mcstover, line 164: @michelleinez We'll need to create a prop that can be used to conditionally apply this modification to this page. We don't want this on "all" instances of DynamicHeroClassic, only the instances used in CCLandingPage.

src/util/contentful/richTextRenderer.js
```diff
@@ -151,3 +151,22 @@ export function richTextRenderer(content) {
...
 }
+
+/* eslint-disable import/prefer-default-export */
+/**
+ * Adds target="_blank" to links so they open in a new tab
+ *
+ * @param {String} String containing html of contentful entry
+ * @returns {void}
+ */
+export function addBlankTargetToExternalLinks(bodyCopy) {
```
mcstover, line 162: Nice job with this new reusable method!

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -158,10 +158,14 @@ export default {
...
 		},
+		pageSettingData: {
+			type: Object,
+			default: () => {},
+		},
 	},
 	async mounted() {
 		await this.$nextTick();
-		addBlankTargetToExternalLinks(this.$refs.heroBodyCopy);
+		addBlankTargetToExternalLinks(this.$refs.heroBodyCopy, this.$props.pageSettingData);
```
mcstover, line 168: you won't need to use `$props` as they available on this, like so, `this.pageSettingData`

mcstover (approved):

PR #4118, opened 2022-08-17
fix: added readQuery mock TheHeader.spec.js test

edwinf-barreram:


dyersituations (approved):

mcstover (approved):

test/unit/specs/components/WwwFrame/TheHeader.spec.js
```diff
@@ -21,6 +21,7 @@ describe('TheHeader', () => {
 					apollo: {
 						readFragment: () => {},
 						query: () => Promise.resolve({}),
+						readQuery: () => Promise.resolve({}),
```
eddieferrer, line 24: I dont think readQuery returns a promise, since it only reads from the cache I think this should be `readQuery: () => {},`

eddieferrer (changes requested):

test/unit/specs/components/WwwFrame/TheHeader.spec.js
edwinf-barreram: Done!

eddieferrer (approved):

emuvente (approved):

PR #4119, opened 2022-08-18
chore: reduce log level of missing thanks query and include transacti…

mcstover:
…onId in error message


emuvente (approved):
Good fix and log cleanup!

eddieferrer (approved):

dyersituations (approved):

PR #4121, opened 2022-08-18
fix: kivaClassic Loan card should not show checkout now button

eddieferrer:


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -147,49 +147,51 @@
...
-				'Lending',
-				'lend-button-loan-upsell',
-				expLabel
-			]"
-		>
-			{{ ctaButtonText }}
-		</kv-ui-button>
+		<!-- Template wrapper for different states of the "lend now" button -->
+		<template v-if="showLendNowButton">
+			<kv-ui-button
```
eddieferrer, line 152: This is the button that shows up now on the homepage when loan is in basket, I think we just need to wrap all these states

mcstover (approved):

PR #4123, opened 2022-08-19
feat: handleteamform trigger was moved after team verification

roger-in-kiva:


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -992,9 +992,6 @@ export default {
 						}
 					});
 				});
-				Promise.all(loans).then(() => {
```
mcstover, line 995: Did you verify that the cache will update the the loan data after the mutation?

The end result should be that the Team Selection box for each Loan in the basket shows the newly joined team as selected.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -879,6 +878,7 @@ export default {
 		},
 		verificationComplete() {
 			this.verificationSubmitted = true;
+			this.handleTeamForm();
```
mcstover, line 881: This is probably a fine location to handle the "after" the verification process. However, we also need to check trigger the check for Team Join on page load IF verification isn't needed.

I suggest you add an else statement around line 865 in the `getPromoInformatinoFromBasket` method

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Do you mean this?
Screen Shot 2022-08-19 at 18 39 19

mcstover: Correct. Just wanted to make sure the new team was selected after joining. You may be to remove your self from that team now to test this again...

roger-in-kiva: Nice, I tested a couple of times adding and removing myself from that team...

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: An else statement was added to trigger the **handleTeamForm** method

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Should be great to test it on dev with more than one joined team to check if selecting as default

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -863,6 +862,8 @@ export default {
 						&& this.loans.length
 					) {
 						this.showVerification = true;
+					} else {
```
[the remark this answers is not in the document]
roger-in-kiva: Right, they are in the method, are these correct? @mcstover 

```
handleTeamForm() {
			if (
				this.loans.length
				&& this.isActivelyLoggedIn
				&& this.teamId
				&& !this.teamJoinStatus
			) {
				// check for team join optionality
				this.showTeamForm = true;
			}
		}
```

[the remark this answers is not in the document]
mcstover: oh ok thanks!

mcstover (approved):

PR #4127, opened 2022-08-19
Feat: Optimizely custom metrics

christian14b:
This pr aims to track custom events:
- Deposits
- Loan share purchase
using [Optimizely api](https://docs.developers.optimizely.com/web/docs/event).


src/pages/Thanks/ThanksPage.vue
```diff
@@ -352,6 +358,17 @@ export default {
 				);
 			}
 		}
+
+		if (this.$appConfig.enableOptimizely) {
```
mcstover, line 362: Generally we're tracking these type of transaction events here: https://github.com/kiva/ui/blob/main/src/plugins/kv-analytics-plugin.js#L214

Thus, you should be able to skip any changes to the Thanks page, import your new event util (or just embed that one method) and use the transactionData we've already aggregated for these events.

emuvente: +1

mcstover (approved):

src/plugins/kv-analytics-plugin.js
```diff
@@ -331,6 +337,25 @@ export default {
 					event: 'refresh'
 				});
 			},
+			trackOPTransaction: transactionData => {
+				window.optimizely.push({
```
emuvente, line 341: I think these events should only fire when the deposit total/loan total is greater than 0, and the only tag that's needed would be the deposit amount/loan share purchase amount.

emuvente (approved):
👍 Looks good!

PR #4133, opened 2022-08-23
fix: enable load more button, add loading state, scaffold lsc in save…

mcstover:
…d search item query

### DIFF
--- a/src/graphql/fragments/loanSearchCriteriaFields.graphql
+++ b/src/graphql/fragments/loanSearchCriteriaFields.graphql
@@ -1,0 +1,52 @@
+fragment loanSearchCriteriaFields on LoanSearchCriteria {
+	filters {
+		activity
+		arrearsRate {
+			min
+			max
+		}
+		avgBorrowerCost {
+			min
+			max
+		}
+		cityState
+		country
+		currencyLossPossible
+		dafEligible
+		defaultRate {
+			min
+			max
+		}
+		distributionModel
+		excludeNonRated
+		expiringSoon
+		gender
+		hasResearchScore
+		isGroup
+		isMatched
+		lenderFavorite
+		lenderTerm {
+			min
+			max
+		}
+		loanIds
+		loanLimit
+		loanTags
+		matcherAccountId
+		partner
+		profitability {
+			min
+			max
+		}
+		riskRating {
+			min
+			max
+		}
+		sector
+		status
+		theme
+		trustee
+	}
+	queryString
+	sortBy
+}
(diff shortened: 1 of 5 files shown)


mcstover:
I'll add the loan count to the View loans button, here, in a subsequent commit.

src/pages/Settings/SavedSearch.vue
```diff
@@ -71,31 +93,80 @@ export default {
...
+										status
+										theme
+										trustee
+									}
+								}
+							}
 						}
 					}
-					
+
```
dyersituations, line 158: Remove extra newline

mcstover:
Ok, the rest has been added. Not sure why the "rename" is framed as a full file change :(

dyersituations (approved):

src/pages/Settings/SavedSearch.vue
```diff
@@ -35,32 +39,51 @@
...
-			savedSearches: [],
+			limit: 10,
+			loading: true,
 			offset: 0,
-			limit: 40 // TODO: Limit is 40 because of an error via (totalCount) API
+			savedSearches: [],
+			totalCount: 0,
 		};
 	},
+	apollo: {
```
tiarascoleman, line 80: Interesting! How are you able to use this as an option instead of original query initialization on Lines 75-88?

src/graphql/fragments/loanSearchCriteriaFields.graphql
```diff
@@ -0,0 +1,52 @@
+fragment loanSearchCriteriaFields on LoanSearchCriteria {
```
tiarascoleman, line 1: What's a fragment?

tiarascoleman (approved):
Great work! I only have a few questions :)

src/graphql/fragments/loanSearchCriteriaFields.graphql
mcstover: Fragments in graphql create a reusable sub-set of query fields, typically on graphql object type that is used in multiple places. The LoanSearchCriteria is a common object that can be found as a sub-field in various locations throughout the api so this fragment could be useful there. @michelleinez this may be helpful for you at some point when pulling the LSC off of a managed account.

src/pages/Settings/SavedSearch.vue
mcstover: This is our global apollo option that is used for the initial fetch of page or component data on the server. This data fetch occurs before the created hook.

PR #4138, opened 2022-08-25
fix: client side routing when routing to same route, different params

eddieferrer:
VUE-1245

I did not like the solution of using `beforeRouteUpdate` because then the onus would be on the developer to add that hook and prefetch the components all the way down if a route with a param could possibly be navigated client side with a change in params. 

This solution leaves navigation prefetch behavior as it was unless, the route is the same and the params are different

### DIFF
--- a/src/client-entry.js
+++ b/src/client-entry.js
@@ -135,12 +135,22 @@ router.onReady(() => {
 		// get newly activated components
 		const matched = router.getMatchedComponents(to);
 		const prevMatched = router.getMatchedComponents(from);
-		const activated = _dropWhile(matched, (c, i) => prevMatched[i] === c);
+		const areRoutesTheSame = JSON.stringify(to?.matched?.[0]?.path) === JSON.stringify(from?.matched?.[0]?.path);
+		const areParamsTheSame = JSON.stringify(to?.params) === JSON.stringify(from?.params);
+		let activated;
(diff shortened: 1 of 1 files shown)


src/client-entry.js
```diff
@@ -135,12 +135,22 @@ router.onReady(() => {
 		// get newly activated components
 		const matched = router.getMatchedComponents(to);
 		const prevMatched = router.getMatchedComponents(from);
-		const activated = _dropWhile(matched, (c, i) => prevMatched[i] === c);
+		const areRoutesTheSame = JSON.stringify(to?.matched?.[0]?.path) === JSON.stringify(from?.matched?.[0]?.path);
+		const areParamsTheSame = JSON.stringify(to?.params) === JSON.stringify(from?.params);
```
dyersituations, line 139: Do we need to check if params are not in the same order?

eddieferrer: vue router returns them in the order they are defined in the route. So for a fictional route of 
`/lend-by-category/:category/:test` they always come back as {category: "foo", test: "bar"}

dyersituations: Sounds good.

dyersituations (approved):

mcstover (approved):
While I don't believe this fully addresses re-fetching global components it _does_ focus to certain parts of the site. Probably lend-by-cat + settings sections at this time.

emuvente (approved):

PR #4143, opened 2022-08-26
MARS-178: Apollo 3 Migrate

edwinf-barreram:
# Description
- Update Apollo imports
- Update Apollo cache function

# Ticket
[MARS-178](https://kiva.atlassian.net/browse/MARS-178)


dyersituations:
@edwinf-barreram Is this PR ready to be reviewed? I'm often hesitant to review draft PRs until I get the go ahead that they are ready 😄.

edwinf-barreram:
I created it as draft because I am trying to fix an error related to my changes but I would like all of you guys (@mcstover, @emuvente, @dyersituations, @eddieferrer, @christian14b) give some feedback and suggestions on this because as I said in the zoom call I didn't work with Apollo before and feel like I'm doing things that are not very clear for me.

This is the error that I am working on.

Screen Shot 2022-08-26 at 17 22 05

dyersituations (changes requested):
I did some testing locally this morning, and I'm seeing some critical errors. On the home page there are warnings around the Apollo Cache. This warning typically means that queries on the same objects are overwriting each other. Sometimes that's desired, sometimes it's not, hence why Apollo warns you. In the current state it means that we might be getting the same query multiple times as different components are overwriting data on the same GraphQL objects. This doesn't occur in main, so investigation will need to be done around why this is happening (such as which queries) and determine if we either need to add a type policy in the Apollo Client or adjust other Apollo Client settings.

image

Then I tested another page that I know is in the `ui` repo and using GraphQL extensively, lend/filter. There are critical errors and the page doesn't load any data.

image

src/api/apollo.js
```diff
@@ -19,30 +14,26 @@ export default function createApolloClient({
...
 	fetch
 }) {
-	const cache = new InMemoryCache({
-		fragmentMatcher: new IntrospectionFragmentMatcher({
-			introspectionQueryResultData: {
-				__schema: { types }
-			}
-		}),
-		// Return a custom cache id for types that don't have an id field
-		dataIdFromObject: object => {
```
mcstover, line 33: @edwinf-barreram @dyersituations I suspect we'll need something like lines 26-41 that were removed here particularly this method which ensures there is an id on every object to associate in the cache. The simplified caching mechanism for the server may also be important – if it's still available.

edwinf-barreram: Yeah, will take a full review about this!

dyersituations: The simplified server caching mechanism is still available (see work in CPS), and this data ID method won't necessarily work 100% in the new Apollo Client. I had to tweak it a bit for the `lend` object, so I wonder if the new version has adjusted the caching mechanism a bit. But this would be part of the solution. @edwinf-barreram Take a look at my recent PRs to CPS for some Apollo-related work. You won't be able to copy and paste, but it might give some ideas.

edwinf-barreram: @dyersituations Thank you very much!

src/api/apollo.js
```diff
@@ -1,48 +1,150 @@
...
+
+const { resolvers } = loanResolverFactory();
+const { LoanPartner, LoanDirect } = resolvers;
+
+export const typePolicies = {
+	activeLoan: {
+		merge: true
+	},
+	LoanChannel: {
+		keyFields: ['id'],
```
emuvente, line 22: I'm not sure that we need to define the `keyFields` when it's only `id`, because that's the default.

src/api/localResolvers/activeLoan.js
```diff
@@ -53,7 +54,8 @@ export default () => {
 						}
 					};
 
-					context.cache.writeData({
+					context.cache.writeQuery({
+						query,
```
dyersituations, line 58: Why did the query need to be imported here and in `addToBasketInterstitial.js`?

src/plugins/apollo-plugin.js
```diff
@@ -41,7 +41,10 @@ export default Vue => {
 									...(isContentfulQuery(query) && isContentfulPreview && { preview: true })
 								}
 							});
-							result.call(this, { data });
+
+							if (data !== null) {
+								result.call(this, { data });
```
dyersituations, line 46: Do we not want to send an empty result? What happens if the query doesn't send back a result in this case?

src/api/localResolvers/experiment.js
```diff
@@ -11,12 +13,25 @@ export default ({ cookieStore }) => {
...
+					let experimentData = null;
+					try {
+						experimentData = cache.readQuery({
+							query: experimentSettingQuery,
+							variables: {
+								key: id || '',
+							}
+						});
+					} catch (e) {
+						logReadQueryError(e, 'AlgoliaPOC itemsInBasketQuery');
```
dyersituations, line 29: Update `logReadQueryError` to match this file/method.

src/api/apollo.js
```diff
@@ -1,48 +1,150 @@
...
+			},
+			// The contentful field doesn't have an ID
+			contentful: {
+				merge: true,
+			},
+			// The lend field doesn't have an ID
+			lend: {
+				merge: true,
+			},
+			// The my field doesn't have an ID
```
dyersituations, line 103: You could combine this identical comments (and some are missing on fields here) to one comment above line 84 (fields) to explain why we use merge below.

src/api/apollo.js
```diff
@@ -72,9 +173,7 @@ export default function createApolloClient({
 		assumeImmutableResults: true,
 	});
 
-	// set default local state
-	cache.writeData({ data: defaults });
-	client.onResetStore(() => cache.writeData({ data: defaults }));
+	client.resetStore();
```
dyersituations, line 176: What does this line do?

src/api/apollo.js
```diff
@@ -1,48 +1,150 @@
...
-			return defaultDataIdFromObject(object);
-		},
-		// Use a simpler underlying cache for server renders
-		resultCaching: typeof window !== 'undefined',
-		// Block modifying cache results outside of normal operations
-		// see https://github.com/apollographql/apollo-client/pull/4543
-		freezeResults: true,
+	// initialize local state resolvers
+	// TODO: convert resolvers into typePolicies
+	// const { resolvers } = initState({ appConfig, cookieStore, kvAuth0 });
```
dyersituations, line 135: Is more work needed here for resolvers or can this commented out code be removed (and the commented out `appConfig` above?

src/plugins/apollo-plugin.js
edwinf-barreram: I got and issue with data null at the moment I was working on the first approach, will take a new look about this.

src/api/apollo.js
```diff
@@ -1,48 +1,150 @@
...
+
+	const possibleTypes = {};
+	types.forEach(element => {
+		const typeList = [element.possibleTypes.map(type => type.name)];
+		// eslint-disable-next-line prefer-destructuring
+		possibleTypes[element.name] = typeList[0];
```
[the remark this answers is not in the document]
edwinf-barreram: Yeah, that is the only position being build in line above what means it always have 1 item.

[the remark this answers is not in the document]
dyersituations: In that case, couldn't it just be `possibleTypes[element.name] = element.possibleTypes.map(type => type.name);`? It doesn't look like `typeList` is used anywhere else.

src/plugins/apollo-plugin.js
dyersituations: Thanks for investigating. If possible, find a use case (or force a user case) that results in data null and see what happens with the query. For example, does it timeout and fail if this result never gets called?

src/api/localResolvers/activeLoan.js
emuvente: `writeData` did not require a query; it wrote the data directly to the cache object however it was provided. With Apollo 3 they have added some additional safety checks, including getting rid of the `writeData` function entirely. That leaves using `writeQuery`, which requires providing the query shape that the data is for which prevents writing data in a shape that is incompatible with the schema.

src/api/apollo.js
emuvente: This line shouldn't be needed

src/api/apollo.js
```diff
@@ -72,9 +153,7 @@ export default function createApolloClient({
 		assumeImmutableResults: true,
 	});
 
-	// set default local state
-	cache.writeData({ data: defaults });
-	client.onResetStore(() => cache.writeData({ data: defaults }));
```
emuvente, line 77: This default local state still needs to be set. Since we can't use `writeData` anymore, this will be a bit more complicated to do. I would recommend:

- changing `defaults` in each localResolver file to be a function that accepts the `cache` as an argument and calls `cache.writeQuery` with the data
- creating a function that calls `defaults(cache)` for every localResolver file
- calling that function here and in the client.onResetStore callback

emuvente (changes requested):
This is coming together great! There is still work needed with migrating the local resolvers and their defaults.

emuvente:
There are some conflicts to resolve, but I think this is otherwise finally ready to go!

emuvente (approved):

dyersituations (approved):

christian14b (approved):

PR #4147, opened 2022-08-30
saved search modal

egoroza:
modal for saved search experiment

Screen Shot 2022-08-30 at 2 05 35 PM
Screen Shot 2022-08-30 at 2 15 21 PM
Screen Shot 2022-08-30 at 2 22 25 PM

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -57,7 +57,11 @@
 		<div class="tw-col-span-2 tw-relative tw-grow">
 			<kv-section-modal-loader :loading="loading" bg-color="secondary" size="large" />
 			<div v-if="initialLoadComplete">
-				<loan-search-saved-search v-if="enableSavedSearch && showSavedSearch" />
+				<loan-search-saved-search
+					v-if="enableSavedSearch && showSavedSearch"
+					:loan-search-state="loanSearchState"
+					:theme-names="themeNames"
+				/>
 				<loan-search-filter-chips
 					:loan-search-state="loanSearchState"
 					:all-facets="allFacets"
@@ -155,7 +159,7 @@ export default {
 		enableSavedSearch: {
 			type: Boolean,
 			default: false,
-		}
+		},
 	},
 	data() {
 		return {
@@ -343,6 +347,9 @@ export default {
 				|| genderFilterApplied
 				|| sectorFilterApplied
 				|| themeFilterApplied;
+		},
+		themeNames() {
+			return this.allFacets?.themeNames;
 		}
 	},
 	methods: {
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -349,7 +349,7 @@ export default {
 				|| themeFilterApplied;
 		},
 		themeNames() {
-			return this.allFacets?.themeNames;
+			return this.allFacets?.themeNames ?? [];
 		}
 	},
 	methods: {
--- a/src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
@@ -50,7 +50,7 @@
 					/>
 					<kv-button
 						@click.native="saveSavedSearch"
-						:state="savedSearchName.length === 0 && 'disabled'"
+						:state="savedSearchName.length === 0 ? 'disabled' : null"
 					>
 						Create Saved Search
 					</kv-button>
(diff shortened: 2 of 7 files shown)


src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
```diff
@@ -1,52 +1,139 @@
...
+			);
+		},
+		saveSavedSearch() {
+			this.isLightboxVisible = false;
+			this.$kvTrackEvent(
+				'Lending',
+				'click-create-saved-search-new-modal',
+				'Create saved search'
+			);
+			createSavedSearch(this.apollo, this.reformattedSearchState, this.savedSearchName);
```
egoroza, line 136: `CORE-693` to handle error & success cases

src/graphql/mutation/createSavedSearch.graphql
```diff
@@ -0,0 +1,15 @@
+mutation createSavedSearch($name: String!, $filters: LoanSearchFiltersInput) {
+  my {
+    createSavedSearch (savedSearch :{
+      filters: $filters,
+      name: $name,
+      isAlert: false,
+      queryString: $name,
+      sortBy: popularity
+    })
```
mcstover, line 9: Can you please add and save the full URL into the `url` param?

mcstover: We use the `url` field to link back into lend/filter (or legacy lend) from the `/lend/saved-search` page.

mcstover: Nevermind. That's not a possible param!

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -343,6 +347,9 @@ export default {
 				|| genderFilterApplied
 				|| sectorFilterApplied
 				|| themeFilterApplied;
+		},
+		themeNames() {
+			return this.allFacets?.themeNames;
```
dyersituations, line 352: Looks like there probably needs to be a `?? []` here so that selecting based on prop in the child component doesn't potentially encounter an exception. Edge case, but just in case.

src/graphql/mutation/createSavedSearch.graphql
```diff
@@ -0,0 +1,15 @@
+mutation createSavedSearch($name: String!, $filters: LoanSearchFiltersInput) {
+  my {
+    createSavedSearch (savedSearch :{
+      filters: $filters,
+      name: $name,
+      isAlert: false,
+      queryString: $name,
```
mcstover, line 7: The queryString param should contain the current query params from the page, not including any `utm_` params.

It looks like the url that we use in the query presentation will be auto generated from the filters in the LSC.

src/util/loanSearch/searchStateUtils.js
```diff
@@ -73,3 +74,20 @@ export async function updateSearchState(apollo, loanQueryFilters, allFacets, que
...
 }
+
+/**
+ * Creates a saved search with a name and set of filters
+ * @param {Object} apollo The Apollo client instance
+ * @param {Object} loanQueryFilters The filters for the saved search
+ * @param {string} savedSearchName The name of the saved search
+ * @returns {Promise<Array>} Promise for the results of the mutation
+ */
+export async function createSavedSearch(apollo, loanQueryFilters, savedSearchName) {
```
mcstover, line 85: Please create a test for this method in `test/unit/specs/util/loanSearch/searchStateUtils.spec.js`

dyersituations, line 85: I'd recommend throwing this in a new util file `savedSearchUtils` that is dedicated to saved search within loan filtering. This file is currently focused on the search state in particular, not saved search.

dyersituations: If you do create a new file, please update the `README` in this folder with a brief description of the new util.

dyersituations: Or create a new test file, based on the other comment I added. Same process though.

src/graphql/mutation/createSavedSearch.graphql
egoroza: gotcha, just to confirm, i'm fixing it so that the `queryString` String value looks as such (pulled from the url query params) -

`{"gender":"female","country":"PH"}`

not sure if there is a way to verify this, the graphql schema just indicates it should be a string

mcstover: I think it's the url format for this one. You could test which form works best by passing the value to a lend > loams query with only the queryString filter applied.

That said, it may turn out to be something even more specifically formatted for php. We can check the monologue code in a little while

mcstover (approved):

PR #4151, opened 2022-09-01
fix: combine loan channel pages to avoid issues with max limit 100 in…

eddieferrer:
… prod

These are the only use cases of this `.graphql` query file. 

The only other uses of these type of query is in the LoanSpotlight component (which will never fail since it has a backup category) and the CategoryAdminControls (which is commented out, and not in use atm)


dyersituations (approved):

mcstover (approved):
1 + 2 and 1 + 2 and...

PR #4164, opened 2022-09-07
Core 696 loggedout savesearch

roger-in-kiva:


src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
```diff
@@ -97,6 +101,10 @@ export default {
 			'view-new-filter-saved-search',
 			''
 		);
+
+		if (this.$route?.query?.saved_search ?? false) {
+			this.openModal();
```
dyersituations, line 106: Which use case needs this modal opening when the page loads?

src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
```diff
@@ -112,16 +120,26 @@ export default {
 				sector: this.loanSearchState?.sectorId,
 				theme: this.loanSearchState?.themeId.map(themeId => this.themeNames[themeId])
 			};
-		}
+		},
+		loginUrl() {
+			if (this.$route.path === '/') {
+				return '/ui-login';
+			}
```
mcstover, line 127: We prod don't need lines 125-127 as this component would never load on the homepage.

src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
mcstover: Looks like the doneUrl setup in the loginUrl below is the case that would trigger this.

@roger-in-kiva I think you should add the `userId` to this if statement.

mcstover (commented):
Makes sense, just a small tweak.

src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
roger-in-kiva: We need to open the modal when coming back from login page triggered by openModal method.

The openModal method is considering the userId, if null is redirecting to login page. So if we trigger this as logged out via url query, the user will be redirected to login page automatically. Do you like to avoid this? @mcstover

src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
roger-in-kiva: Fixed, thanks!

dyersituations (approved):

mcstover (approved):

src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
mcstover: That works!

PR #4165, opened 2022-09-07
fix: set root level id on loanCategoriesByLoanChannelIds query

mcstover:


dyersituations (approved):

mcstover:
@emuvente @eddieferrer @dyersituations 
Sorry about that hiccup. This wasn't in prod yet with my first PR so full lint check didn't occur. Rather it just warned against a missing type. Also, I was a bit confused at how you can specify the root id but also have to use the `... on` types where the id also lives with this new api...

roger-in-kiva (approved):

eddieferrer (approved):

PR #4169, opened 2022-09-09
feat: stylized carousel for Story Cards MARS-241

emuvente:
Depends on kiva/kv-ui-elements#203

Large screens:
Screen Shot 2022-09-09 at 9 54 32 AM

Small screens:
Screen Shot 2022-09-09 at 9 55 11 AM

### DIFF
--- a/src/components/Contentful/StoryCard.vue
+++ b/src/components/Contentful/StoryCard.vue
@@ -49,17 +49,21 @@
 			v-else
 		>
 			<dynamic-rich-text
-				class="tw-text-action"
+				class="story-card__title tw-text-action"
 				:class="`tw-text-${alignment}`"
+				v-if="cardTitle"
 				:html="cardTitle"
 			/>
 			<dynamic-rich-text
 				class="story-card__content tw-h-full tw-pb-4 tw-pt-3"
 				:class="`story-card__content--${alignment} tw-text-${alignment}`"
+				v-if="cardContent"
 				:html="cardContent"
 			/>
 			<dynamic-rich-text
+				class="story-card__footer"
 				:class="`tw-text-${alignment}`"
+				v-if="cardFooter"
 				:html="cardFooter"
 			/>
 		</div>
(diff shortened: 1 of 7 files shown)


emuvente:
I based this component on Story Cards, but the number of custom styles I ended up applying makes me think there might be a better way to go about this. Any suggestions for alternatives, or is this ok for now?

src/components/Contentful/StoryCard.vue
```diff
@@ -49,17 +49,21 @@
 			v-else
 		>
 			<dynamic-rich-text
-				class="tw-text-action"
+				class="story-card__title tw-text-action"
 				:class="`tw-text-${alignment}`"
+				v-if="cardTitle"
```
mcstover, line 54: I'm a bit concerned that making these conditional will affect the layout of existing Story Cards, specifically the distribution of the flex column...

Could you please load up `/lp/own-the-change` locally to compare against https://www.dev.kiva.org/lp/own-the-change to verify.

mcstover: `/lp/eddie-test-page` is another good comparison point

mcstover (approved):
Pending layout check on the StoryCards

dyersituations (approved):

mcstover:
@emuvente I agree that StoryCardCarousel.vue is pretty specialized and overrides some significant portions of the StoryCard behavior but think we should roll with it for now. I can think of some fun variations on this we can discuss with design if we decide to keep and use this elsewhere... ie. what if the background color matched the Story Card theme once you switch cards!!!

src/components/Contentful/StoryCard.vue
mcstover: alternatively you can tests any of these pages in dev against cms page server by replacing `/lp` with `/cpslp`.

emuvente: I don't see any visual differences between dev and this PR running on localhost for both of those pages on both large and small screens.

eddieferrer (approved):

PR #4170, opened 2022-09-09
Core 688 saved search

roger-in-kiva:
kv-lightbox is triggering twice the closeModal method, a condition was added to fire tracking event only once as temporal solution for the experiment


src/components/Lend/LoanSearch/LoanSearchSavedSearch.vue
```diff
@@ -141,11 +143,14 @@ export default {
 		},
 		closeModal() {
 			this.isLightboxVisible = false;
-			this.$kvTrackEvent(
-				'Lending',
-				'click-new-saved-search-modal-dismiss',
-				''
-			);
+			if (!this.closedModal) {
```
dyersituations, line 146: Why was it triggered twice? The root cause might influence a different solution.

roger-in-kiva: I tried to find it looking and testing through kv-lightbox component but not successfully. Do you like to try? @dyersituations

dyersituations: I'm not currently in this repo, but I could help test early next week, just let me know.

roger-in-kiva: I think we can approve this as a temporal solution to the experiment and open a new ticket to check why is firing twice

mcstover: I think the duplication comes from KvLightbox being unmounted when the search is saved and the LoanSearchSavedSearch component is removed from the dom. saveSavedSearch toggles off the visibility causing the component to emit the 1st close event. Then events are emitted up with showSavedSearchSuccessMessage which will take the interface out of the dome and trigger unmount and the 2nd close event.

mcstover (approved):

PR #4171, opened 2022-09-09
MARS-239: new home page loan cards

edwinf-barreram:
# Description
- Added new Loan Card vue component 
- Added new Story book template

# Ticket
[MARS-239](https://kiva.atlassian.net/browse/MARS-239)

Screen Shot 2022-09-09 at 15 31 13

### DIFF
--- a/src/components/NewHomePage/LoanCards.vue
+++ b/src/components/NewHomePage/LoanCards.vue
@@ -7,7 +7,7 @@
 			class="link"
 			:is="allSharesReserved ? 'span' : 'router-link'"
 			:to="`/lend-beta/${loanId}`"
-			v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId]"
+			v-kv-track-event="['Lending', 'click-see more', 'card', loanId]"
 		>
 			<!-- Borrower image w/ location <summary-tag> -->
 			<kv-loading-placeholder
--- a/src/components/NewHomePage/LoanCards.vue
+++ b/src/components/NewHomePage/LoanCards.vue
@@ -298,7 +298,6 @@ export default {
 		},
 		processQueryResult(result) {
 			if (result.error) {
-				console.error(result.error);
 				this.$showTipMsg('There was a problem loading your loan recommendations', 'error');
 				try {
 					Sentry.withScope(scope => {
--- a/src/components/NewHomePage/LoanCards.vue
+++ b/src/components/NewHomePage/LoanCards.vue
@@ -340,7 +340,6 @@ export default {
 }
 
 .card-container {
-	@apply tw-m-[50px];
 	@apply tw-min-w-[260px] tw-max-w-[260px] tw-h-[100%] tw-p-[10px];
 	@apply tw-inline-block tw-rounded-[8px] tw-transition-all tw-duration-300 tw-ease-in-out;
 }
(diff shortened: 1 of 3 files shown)


src/components/NewHomePage/LoanCards.vue
```diff
@@ -0,0 +1,407 @@
...
+					apollo: this.apollo,
+					loanId: this.loanId,
+					loanCardQuery,
+					callback: result => this.processQueryResult(result),
+				});
+			}
+		},
+		processQueryResult(result) {
+			if (result.error) {
+				console.error(result.error);
```
dyersituations, line 301: Remove log?

src/components/NewHomePage/LoanCards.vue
```diff
@@ -0,0 +1,407 @@
...
+			if (this.queryObserver) {
+				this.queryObserver.setVariables({
+					loanId,
+				});
+			}
+		},
+	},
+};
+</script>
+<style lang="scss" scoped>
```
dyersituations, line 336: Is this a temporary experiment or new permanent element? If it's permanent, consider changing to tailwind + postcss. I see this uses the old foundation import as well.

mcstover: We should not be importing `settings` or `foundation` any longer.

Please refactor this to use `tw-` classes for the colors you need, which you can apply within the `postcss` setup if you'd like. You can keep some of this raw css but most of it should be converted to `tw-` classes except where tailwindcss doesn't support it. _I think everything below is supported._

mcstover: Checkout [TheHeader](https://github.com/kiva/ui/blob/main/src/components/WwwFrame/TheHeader.vue) and https://github.com/kiva/ui/blob/main/src/components/BorrowerProfile/WhatIsKivaModal.vue#L136 for samples of the new style usage paradigm.

edwinf-barreram:
@mcstover @dyersituations I have updated the styles to use tailwindcss.

Screen Shot 2022-09-09 at 15 31 13

src/components/NewHomePage/LoanCards.vue
```diff
@@ -0,0 +1,359 @@
...
+</script>
+
+<style lang="postcss" scoped>
+
+.link {
+	@apply tw-no-underline tw-text-black hover:tw-no-underline;
+}
+
+.card-container {
+	@apply tw-m-[50px];
```
dyersituations, line 343: We've been using regular styles instead of the tailwind JIT classes with square brackets. Those classes add extra classes to the compiled, where the sizings might only be needed in this file.

edwinf-barreram: Sorry, I need to delete that. It was just for testing purpose.

src/util/loanUtils.js
```diff
@@ -136,3 +136,24 @@ export function watchLoanData({
 	// Return the observer to allow modification of variables
 	return queryObserver;
 }
+
+export function watchLoanCardData({
```
mcstover, line 140: Was this really needed? Did reusing the existing method fail in some way?

edwinf-barreram: The other requires cookies data to add the loans in the basket, for this card that's not necessary.

mcstover (approved):
Nice work!

dyersituations (approved):

emuvente (approved):
👍 Looks great!

PR #4172, opened 2022-09-10
fix: turn off algolia analytics

mcstover:


src/main.js
```diff
@@ -82,10 +82,6 @@ export default function createApp({
 				if (eventAsString.indexOf("Cannot set property 'PX1065' of undefined") !== -1) {
 					return false;
 				}
-				// Skip sending failed Algolia analtyics related errors
-				if (eventAsString.indexOf('Before calling any methods on the analytics') !== -1) {
-					return false;
-				}
```
mcstover, line 88: As we wind down this implementation, we want to know if any of these are occurring so we can remove them.

mcstover:
FYI @emuvente @dyersituations Here's that tiny pr ;)

dyersituations (approved):

PR #4173, opened 2022-09-13
Feat/dynamic hero classic two cols mars 242

christian14b:
This pr allow Dynamic Hero Classic to support different number of columns depending on settings.

Final when columns are set to 2
image

Without grid elements:
image


.storybook/stories/CorporateCampaign.stories.js
```diff
@@ -9,6 +9,7 @@ import CorporateCampaignLanding from '@/pages/LandingPages/CorporateCampaign/CCL
 import CampaignHero from '@/components/CorporateCampaign/CampaignHero';
 import CampaignStatus from '@/components/CorporateCampaign/CampaignStatus';
 import cookieStoreStoryMixin from '../mixins/cookie-store-story-mixin';
+import DynamicHeroClassic from '@/components/Contentful/DynamicHeroClassic';
```
mcstover, line 12: Can you please create a new Story for DynamicHeroClassic so we don't conflate it with Corporate Campaigns.

mcstover: Otherwise this is looking good!

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -105,13 +106,39 @@
...
-						<button-wrapper
-							class="tw-w-full md:tw-w-auto"
-							:content="buttonContent"
-						/>
+						<div
+							v-if="gridHeroBody.length"
+							:class="{'tw-grid md:tw-grid-cols-2 md:tw-grid-rows-2 tw-gap-1 md:tw-gap-3'
+								: gridHeroBody.length }"
+						>
+							<template v-if="gridHeroBody.length">
```
mcstover, line 120: The `v-if` on this template shouldn't be needed as the same check is on the div above.

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -191,6 +231,17 @@ export default {
 			const text = this.genericContentBlock?.bodyCopy ?? '';
 			return text ? richTextRenderer(text) : '';
 		},
+		gridHeroBody() {
+			return ['first-grid-item', 'second-grid-item', 'third-grid-item', 'fourth-grid-item'].map(el => {
+				const block = this.getGridBlock(el);
+				const text = block?.bodyCopy ?? '';
+				if (!text) return false;
```
dyersituations, line 238: Does this `false` cause an error in the template? I see the template expects `block.text`, which should fail if the block is `false` like in this return. Should there instead be a `.filter` before this map?

mcstover:
@michelleinez You might be happy to see we're added a 2nd button to the DynamicHeroClassic!

src/components/Contentful/DynamicRichText.vue
```diff
@@ -23,12 +27,32 @@ export default {
...
+					classes += `md:tw-grid-cols-${this.bodyColumns.md} `;
+				}
+				if (this.bodyColumns?.lg) {
+					classes += `lg:tw-grid-cols-${this.bodyColumns.lg} `;
+				}
+				if (this.bodyColumns?.xl) {
+					classes += `xl:tw-grid-cols-${this.bodyColumns.xl} `;
+				}
+			} else {
+				classes += 'tw-prose';
```
mcstover, line 44: tw-prose sets spacing and type styles on various elements so it should probably be applied for all cases.

emuvente: tw-prose adds a a top and bottom margin to every element except for the first one. That lack of margin on the first element causes the grid items to be mis-aligned:
Screen Shot 2022-09-15 at 9 50 37 AM
Screen Shot 2022-09-15 at 9 51 55 AM

So to keep tw-prose those margins will have to be removed so that the spacing is handled by the grid instead when needed.

mcstover: oh yea, well that's a weird one... maybe we can address in tw-prose at somepoint. I was just thinking about when someone tries to use other elements or embeds with these blocks applied. Let's go with what you have for now.

How will the blocks be determined/configured in Contentful?

mcstover: Also, please review this code running locally against https://www.dev.kiva.org/lp/dynamic-hero-components + https://www.dev.kiva.org/lp/how-kiva-works to ensure the content layout of their dynamic hero usages are not altered.

mcstover: Ok so it's "activated" by adding configuration. That sounds good!

Thanks for verifying against the other pages!!!

mcstover (approved):

src/plugins/contentful-ui-setting-styles-mixin.js
```diff
@@ -56,5 +56,9 @@ export default {
 		verticalPadding() {
 			return this.uiSetting?.dataObject?.verticalPadding ?? {};
 		},
+
+		bodyColumns() {
+			return this.uiSetting?.dataObject?.bodyColumns;
```
mcstover, line 61: 👍

dyersituations (approved):

src/components/Contentful/DynamicRichText.vue
```diff
@@ -9,12 +9,16 @@ export default {
 		html: {
 			type: String,
 			default: ''
+		},
+		bodyColumns: {
+			type: Object,
+			default: () => {}
```
emuvente, line 15: I think the default might need to be `null`. Since `{}` evaluates as `true`, the grid classes will always be applied instead of the prose class.

src/components/Contentful/DynamicHeroClassic.vue
```diff
@@ -108,10 +109,27 @@
 						<div v-if="heroBody" class="tw-prose tw-mb-2 md:tw-mb-3" ref="heroBodyCopy">
 							<dynamic-rich-text :html="heroBody" />
 						</div>
-						<button-wrapper
-							class="tw-w-full md:tw-w-auto"
-							:content="buttonContent"
-						/>
+						<div v-if="gridHeroBody" class="tw-mb-2 md:tw-mb-3">
+							<dynamic-rich-text :body-columns="bodyColumns" :html="gridHeroBody" />
```
emuvente, line 113: I think the body-columns prop should be applied to the existing heroBody dynamic rich text, rather than having a separate "gridHeroBody".

PR #4175, opened 2022-09-13
fix: update lighthouse checkout step to gain visibility to ancestor hash

mcstover:
Force github checkout action to look deeper into the commit history.

ref. https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/troubleshooting.md#running-lhci-in-github-actions-i-constantly-get-ancestor-hash-not-determinable

### DIFF
--- a/.github/workflows/lighthouse-prod.yml
+++ b/.github/workflows/lighthouse-prod.yml
@@ -19,6 +19,10 @@ jobs:
       PSI_API_KEY: ${{ secrets.LHCI_PSI_API_KEY }}
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 40
+      - name: Fetch base_ref HEAD to use it as Ancestor hash in LHCI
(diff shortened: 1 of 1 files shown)


.github/workflows/lighthouse-prod.yml
```diff
@@ -19,6 +19,10 @@ jobs:
       PSI_API_KEY: ${{ secrets.LHCI_PSI_API_KEY }}
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 40
```
mcstover, line 23: I went with a larger number here, rather than 20 as mentioned, b/c sometimes we have more than 20 commits in a day...

mcstover:
Also created https://kiva.atlassian.net/browse/VUE-1297 to get the collector updated.

dyersituations (approved):

emuvente (approved):
Nice. Thanks for addressing this!

PR #4176, opened 2022-09-14
fix: comment out broken ref line

mcstover:


.github/workflows/lighthouse-prod.yml
```diff
@@ -20,9 +20,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 40
-      - name: Fetch base_ref HEAD to use it as Ancestor hash in LHCI
-        run: git fetch --depth=1 origin +refs/heads/${{github.base_ref}}:refs/remotes/origin/${{github.base_ref}}
+          fetch-depth: 30
+      # - name: Fetch base_ref HEAD to use it as Ancestor hash in LHCI
+      #   run: git fetch --depth=1 origin +refs/heads/${{github.base_ref}}:refs/remotes/origin/${{github.base_ref}}
```
mcstover, line 25: This was failing...I think the fix in the docs assumed the action was running as part of a PR which appears to [only be available in that context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) (search base_ref)

Will try and get the run going tonight and follow up with more later.

dyersituations (approved):

PR #4182, opened 2022-09-14
add a k8saudit config file

jmallatt-kiva:


config/k8saudit.js
```diff
@@ -0,0 +1,66 @@
+const { merge } = require('webpack-merge');
+var base = require('./index.js')
```
mcstover, line 2: change `index` to `audit` here to import the existing `audit.js` config.

config/k8saudit.js
```diff
@@ -0,0 +1,66 @@
...
+
+module.exports = merge(base, {
+	app: {
+		host: 'www.audit.kiva.org',
+		publicPath: 'https://www-audit-kiva-org.freetls.fastly.net/ui/',
```
mcstover, line 7: Remove all other fields under `app`. Keep `publicPath` and remove `/ui/` from the end of it. Everything else can go as it's already defined in audit.js.

config/k8saudit.js
```diff
@@ -0,0 +1,66 @@
...
+		paypal : {
+			url: 'www.sandbox.paypal.com',
+			environment: 'sandbox'
+		},
+	},
+	server: {
+		graphqlUri: 'https://marketplace-api-audit.dk1.kiva.org/graphql',
+		sessionUri: 'https://www.audit.kiva.org/start-ui-session',
+		memcachedEnabled: true,
+		memcachedServers: 'audit-memcached-01:11211',
```
mcstover, line 64: Remove `graphqlUri` + `sessionUri` here as they are already defined.

If `audit-memcached-01:11211` still exist and is reachable from within our cluster  (or can be made reachable with a security group) you could use this value. Otherwise we'll need to setup a memcached node and sg in the cluster.

Alternatively. If you change the Deployed number of nodes to just 1 in the deploy config. you could set `memcachedEnabled` to false and use an in app/memory cache mechanism. _this might work with more than 1 node but would likely bork logged in session._

jmallatt-kiva: memcached cluster is set up and reachable.   I don't think we should need a second, ui specific cluster in the case of the audit env

PR #4191, opened 2022-09-16
MARS-244: Loan card and Carousel

edwinf-barreram:
# Description

- Added New Loan Cards to Categories section
- Added New Loan Cards to Carousel

# Ticket
[MARS-244](https://kiva.atlassian.net/browse/MARS-244)

Screen Shot 2022-09-20 at 19 12 17
Screen Shot 2022-09-20 at 19 12 30
Screen Shot 2022-09-20 at 19 17 33


src/components/LoanCollections/HomeExp/KivaLoanCardCarousel.vue
```diff
@@ -0,0 +1,171 @@
+<template>
```
mcstover, line 1: @edwinf-barreram Did you explore making this loan card type a prop on the existing loan card carousel we use on homepage? That would be my preferred approach rather than having an entirely different component with a lot of duplicated code. It would also allow us to re-use the setup coming from contentful and just add that config to the dataObject there...

What I've described above will be the process for when we port this to cms-page-server

edwinf-barreram: @mcstover I tried it initially but due to the quantity of customizations that I was having to add there just to show the different versions and due cards have different sizes and the new carousel version must show 4 cards instead of 3, it was making it difficult to take control of everything. It was the reason why I decided to create a new component in a different folder just for the experiment.

Also... thinking on future when we have to delete either of them it don't be a pain in the ass. 😅

mcstover: Ok thanks for considering this. I'll likely still setup a switch to choose a loan card in the new app.

src/util/categoryUtils.js
```diff
@@ -135,3 +135,36 @@ export function getHeroChannelCached(client) {
...
 	}
 }
+
+/**
+ * Get the current selected category name
+ * @param {Number} categoryId
+ * @param {String} name
+ * @return {String} category name
+*/
+export function getCategoryName(categoryId, name) {
```
dyersituations, line 145: While this file doesn't have utils yet, this would be a good method to add a simple unit test for.

src/components/LoanCollections/HomeExp/KivaLoanCardCategory.vue
```diff
@@ -0,0 +1,137 @@
...
+			// const loadMoreCardId = 2;
+			// TODO: splice if promoCard if active on row
+			// if (this.showPromoCard) {
+			// 	clonedLoanIds.splice(1, 0, promoCardId);
+			// }
+			// TODO: append loadMoreCard if active
+			// if (this.showLoadMoreCard) {
+			// 	clonedLoanIds.push(loadMoreCardId);
+			// 	return clonedLoanIds;
+			// }
```
dyersituations, line 83: Will this commented out code be added? If not, this computed can be removed, and the `loanIds` used directly.

src/components/LoanCollections/HomeExp/KivaLoanCardCarousel.vue
```diff
@@ -0,0 +1,171 @@
...
+			return this.augmentedLoanIds.length === 0;
+		},
+		augmentedLoanIds() {
+			const clonedLoanIds = [...this.loanIds];
+			// const promoCardId = 1;
+			// const loadMoreCardId = 2;
+			// TODO: splice if promoCard if active on row
+			// if (this.showPromoCard) {
+			// 	clonedLoanIds.splice(1, 0, promoCardId);
+			// }
```
dyersituations, line 93: Looks like the same commented out block as the other file.

src/components/LoanCards/NewHomePageLoanCard.vue
```diff
@@ -355,4 +356,12 @@ export default {
 .borrower-name {
 	@apply tw-mt-[10px];
 }
+
+.loading-placeholder {
+	@apply tw-min-w-[260px] tw-max-w-[260px];
```
dyersituations, line 361: Use normal CSS instead of JIT classes with brackets.

src/components/LoanCards/NewHomePageLoanCard.vue
```diff
@@ -355,4 +356,12 @@ export default {
...
+}
+
+.loading-paragraph {
+	@apply tw-min-w-[260px] tw-max-w-[260px];
```
dyersituations, line 365: Use normal CSS instead of JIT classes with brackets.

src/components/LoanCards/NewHomePageLoanCard.vue
```diff
@@ -73,23 +73,24 @@
 			<kv-loading-paragraph
 				v-if="isLoading"
 				class="tw-mb-1.5 tw-flex-grow"
+				style="min-height: 1rem;"
 			/>
 			<div v-if="!isLoading" class="tw-text-left">
-				<p class="tw-m-0 tw-m-h[3rem] tw-overflow-hidden tw-text-ellipsis tw-line-clamp-2 tw-text-md">
+				<p class="tw-m-0 tw-h-6 tw-m-h[3rem] tw-overflow-hidden tw-text-ellipsis tw-line-clamp-2 tw-text-md">
```
dyersituations, line 79: Replace JIT class with brackets with inline styles.

edwinf-barreram: @dyersituations Could you add examples about this?

dyersituations: Actually what is `tw-m-h[3rem]`? I'm not seeing it in the TW docs, so I think it might be incorrect.

I just meant that we can to change TW JIT class, such as `tw-m-[5px]` to the inline style version, or `style="margin: 5px"`. The reason being is that we want to general reduce load time of assets, and each JIT class means an extra class in the compiled TW code that our app loads on EVERY page. So we are working with the idea that removing JIT classes in our somewhat large app could lead to faster load times.

dyersituations: JIT stands for "just in time", but just means they are new classes compiled into the TW CSS the app loads.

emuvente: FYI 3rem is a 6 in our tailwind sizing system (1 = 8px; 16px = 1rem). `tw-m-h` isn't a possible class, but if that was meant to be `tw-m-[3rem]` then it should be replaced with `tw-m-6`.

src/components/Homepage/HomeExp/KivaMultiCategoryGrid.vue
```diff
@@ -1,35 +1,40 @@
 <template>
-	<div class="md:tw-flex md:tw-items-start">
-		<div class="tw-mr-4">
-			<p class="tw-text-h3 tw-font-medium tw-mb-2 tw-text-secondary">
-				I want to support
-			</p>
-			<loan-category-selector-home-exp
+	<div class="row">
+		<div class="small-12 large-3 medium-3">
```
emuvente, line 3: The `row`, `small-12`, `large-3`, and `medium-3` classes are all from Foundation, not Tailwind, which has different breakpoints.

You can use the KvGrid component and `col-span` classes to do the same thing in Tailwind:
```vue
<kv-grid class="tw-grid-cols-12">
    <div class="tw-col-span-12 md:tw-col-span-3">
        ...
    </div>
    <div class="tw-col-span-12 md:tw-col-span-9">
        ...
    </div>
</kv-grid>
```

src/components/Contentful/HomePage/NewHomeLoansByCategoryGrid.vue
```diff
@@ -1,35 +1,21 @@
...
+			:new-home-exp="true"
+		/>
+	</div>
 </template>
 
 <script>
 import contentfulStylesMixin from '@/plugins/contentful-ui-setting-styles-mixin';
 import KivaMultiCategoryGrid from '@/components/Homepage/HomeExp/KivaMultiCategoryGrid';
-import SectionWithBackgroundClassic from '@/components/Contentful/SectionWithBackgroundClassic';
-import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
```
emuvente, line 25: SectionWithBackgroundClassic and KvPageContainer are both useful components to use here, as they let the spacing, theme, and background all be controlled from Contentful while keeping the content aligned with the rest of the page. We should try to keep both of these if possible. Were they causing styling issues?

emuvente (commented):
Looking good, just a couple comments!

src/components/Contentful/HomePage/NewHomeLoansByCategoryGrid.vue
edwinf-barreram: I have reverted this change!

src/components/LoanCollections/HomeExp/KivaLoanCardCategory.vue
```diff
@@ -0,0 +1,137 @@
+<template>
```
emuvente, line 1: This component is supposed to have a carousel for the loan cards on small screens. This currently hides all the loans by default, only showing loans for medium-and-up screen sizes.

src/components/Contentful/HomePage/NewHomeLoansCardCarousel.vue
```diff
@@ -0,0 +1,143 @@
+<template>
+	<section-with-background-classic
+		:background-content="background"
+		:theme-name="themeName"
+		:vertical-padding="verticalPadding"
+	>
+		<template #content>
+			<kv-page-container>
+				<div>
```
eddieferrer, line 9: Is this wrapper div necessary?

src/components/LoanCollections/HomeExp/KivaLoanCardCategory.vue
eddieferrer: > Will this commented out code be added? If not, this computed can be removed, and the `loanIds` used directly.

Agreed. I removed this in another carousel I built.

src/components/Homepage/HomeExp/KivaMultiCategoryGrid.vue
```diff
@@ -1,35 +1,48 @@
 <template>
-	<div class="md:tw-flex md:tw-items-start">
-		<div class="tw-mr-4">
-			<p class="tw-text-h3 tw-font-medium tw-mb-2 tw-text-secondary">
-				I want to support
-			</p>
-			<loan-category-selector-home-exp
+	<kv-grid class="tw-grid-cols-12">
+		<div class="tw-col-span-12 md:tw-col-span-3">
+			<div class="md:tw-items-start">
```
emuvente, line 4: Since this doesn't have `tw-flex` I don't think the `md:tw-items-start` class is doing anything. This wrapper div can probably be removed.

emuvente (approved):
This is looking good! This could be merged once the remaining comments are addressed.

mcstover (approved):

PR #4194, opened 2022-09-19
fix: add other file extensions so @vueuse modules can be found

mcstover:
We found that the vueuse components had cjs/mjs extensions so jest couldn't find them... maybe we need to add this elsewhere too???


SqueakAttack (approved):
👍 Thanks! I tried this locally and I see that the problem test and smattering of other tests run with it.

emuvente (approved):

PR #4196, opened 2022-09-19
feat: eco challenge thanks page

eddieferrer:
* Adds routing to eco thanks page and reading achievement information from service

On checkout, if user is part of the eco_challenge experiment, fetch the achievements milestones (which is passed loan ids in basket) if these loans would progress the users challenge progress, add a query param, the goes through post purchase route and directs them to the thank you page. 

Added some computed properties to the thank you page to parse achievement response. This thank you page is based on the mars exp, so it only shows 1 loan in the hero area. Ideally we would want this loan to be one of the loans in the checkout that contributed to the challenge progress, this proved to be a little difficult, so for the test, I am generalizing the message displayed next to the loan, and then just searching for any of a list of eco terms in the loan uses of the transaction to try to find the eco loan. I expect in some cases, with checking out multiple loans, this could miss and a non eco loan will be shown up top in the header area.... 

Screen Shot 2022-09-19 at 3 38 06 PM
Screen Shot 2022-09-19 at 2 57 17 PM

### DIFF
--- a/src/graphql/query/achievementMilestones.graphql
+++ b/src/graphql/query/achievementMilestones.graphql
@@ -1,0 +1,12 @@
+query achievementMilestonesForCheckout($loanIds: [String!]!) {
+	achievementMilestonesForCheckout(loanIds: $loanIds){
+		checkoutMilestoneProgresses {
+			achievement
+			milestoneName
+			postCheckoutProgress
+			progress
+			status
+			target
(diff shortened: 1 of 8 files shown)


mcstover (approved):
Looks good!

emuvente (approved):
👍 great!

PR #4209, opened 2022-09-21
Feat/green text in dynamic hero class

christian14b:
Add green text to DynamicHeroClassic

image


mcstover (approved):
We'll circle up with design during our brand refresh to set some standards around this type of style augmentation but this should be good for now...

edwinf-barreram (approved):
LGTM 🚀

PR #4212, opened 2022-09-22
feat: add query param to force dynamic upsells + tracking

egoroza:
* adds a query parameter `forceDynamicUpsell` to force experience to use dynamic upsell
* adds some tracking

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -589,7 +589,11 @@ export default {
 		// show toast for specified scenario
 		this.handleToast();
 		this.getPromoInformationFromBasket();
-		this.getUpsellModuleData();
+		if (this.$route.query.forceDynamicUpsell === 'true') {
+			this.getDynamicUpsellModuleData();
+		} else {
+			this.getUpsellModuleData();
+		}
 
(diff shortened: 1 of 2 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1003,7 +1019,11 @@ export default {
 							);
 							// eslint-disable-next-line max-len
 							this.$showTipMsg('Looks like that loan was reserved by someone else! Try this one instead.', 'info');
-							this.getUpsellModuleData();
+							if (this.$route.query.forceDynamicUpsell === 'true' && this.enableDynamicUpsells) {
```
mcstover, line 1022: What's the use case for this query param? Is it temporary?

egoroza: yes really just for testing with product/design! since the dynamic upsell loan condition will only be met if there are no loans with <$50 left.

mcstover (approved):

src/pages/Checkout/CheckoutPage.vue
mcstover: Ok, that makes sense. Just wanted to better understand if this hard condition to have the query param would end up in production and be a dependency there too.

PR #4216, opened 2022-09-22
Update lend cta lbc hover card

mcstover:
Screen Shot 2022-09-22 at 3 33 32 PM

Screen Shot 2022-09-22 at 3 33 17 PM

These Loan Cards now know when they are in the basket.


roger-in-kiva (approved):
thanks! 👊🏻

PR #4222, opened 2022-09-26
feat: a/b test for quick filters

egoroza:
setup a/b test for quick filters on category pages


roger-in-kiva:
Looks good!

You need to exclude the test from some pages:
`add a/b test on all category pages (Except WOmen, Eco Friendly, Kiva US) that shows filters at top`

egoroza:
> Looks good!
> 
> You need to exclude the test for some categories: `add a/b test on all category pages (Except WOmen, Eco Friendly, Kiva US) that shows filters at top`

forgot about that thank you! will update

PR #4223, opened 2022-09-26
MARS-249: New home page experiment setup

edwinf-barreram:
# Description
- Added new home page experiment query
- Added new home page experiment tracking
- Load home page based on experiment version

# Ticket 

[MARS-249](https://kiva.atlassian.net/browse/MARS-249)

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -100,10 +100,15 @@ export default {
 			}).then(result => {
 				// Call preFetch for the active homepage
 				const expVersion = result[0]?.data?.experiment?.version;
+
+				if (expVersion === 'c') {
+					return Promise.reject({	path: '/cps/home' });
+				}
+
 				if (expVersion === 'a' || expVersion === 'b') {
 					return preFetchAll([result[1]?.default], client, args);
 				}
-				return Promise.reject({	path: '/cps/home' });
+				Promise.resolve();
 			});
 		},
 	},
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -6,6 +6,7 @@
 /* eslint-disable vue/multi-word-component-names */
 import gql from 'graphql-tag';
 import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
+import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import { preFetchAll } from '@/util/apolloPreFetch';
 
 const ContentfulPage = () => import('@/pages/ContentfulPage');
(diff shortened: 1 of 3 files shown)


src/pages/Homepage/Homepage.vue
```diff
@@ -85,10 +92,31 @@ export default {
...
+						homePageComponent = await ContentfulPage();
+						break;
+					case 'b':
+						// Should prefetch the experimental home page
+						homePageComponent = await ContentfulPage();
+						break;
+					default:
+						// Should redirect to /cps/home
+						homePageComponent = await ContentfulPage();
+						break;
```
mcstover, line 116: Can we please add a version `c` check for the redirect to `/cps/home` and set the standard ui homepage as the default/fallback condition.

mcstover (approved):

src/router/routes.js
```diff
@@ -1,10 +1,19 @@
...
+import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
+
+const routes = [
 	{
 		path: '/',
 		name: 'homepage',
 		component: () => import('@/pages/Homepage/Homepage'),
 		meta: {
-			contentfulPage: () => 'home',
+			contentfulPage: (route, apollo) => {
```
eddieferrer, line 9: This is a real clever way to do the experiment at a route level. I think we could reuse this in the future for some other route level experiments

src/pages/Homepage/Homepage.vue
```diff
@@ -85,13 +93,33 @@ export default {
...
+				return Promise.all([
+					client.query({ query: experimentAssignmentQuery, variables: { id: 'new_home_layout' } }),
+					ContentfulPage(),
+				]);
+			}).then(result => {
 				// Call preFetch for the active homepage
-				const component = resolvedImport.default;
-				return preFetchAll([component], client, args);
+				const expVersion = result[0]?.data?.experiment?.version;
+				if (expVersion === 'a' || expVersion === 'b') {
```
eddieferrer, line 103: I think this is all the versions? you could simplify to just `if (expVersion)`

eddieferrer (approved):

src/router/routes.js
edwinf-barreram: It was a @emuvente Idea, so He is the mastermind behind this!

src/pages/Homepage/Homepage.vue
mcstover: We actually need to preserve a version `c` which represents a percentage of traffic we'll send to cms-page-server.

Still a concerned that it's set as the default with no controls to bypass. While we could use the exp distribution settings it would still be better to have an explicit `if (expVersion === 'c')` block to activate that redirect.

edwinf-barreram: Ok, I'll add the validation!

mcstover (approved):

emuvente (approved):
Looks great!

PR #4224, opened 2022-09-26
feat: use flss for lbc page for eco game categories

eddieferrer:
Use FLSS for the lend-by-category/"category" pages and the eco friendly game page with sub categories for the categories that are in the eco game. The "filter" link will still default to the legacy page, since the new one does not have support for keywords.

Since the subcategory carousels on the eco page are not set to prefetch, and we only have `getCachedChannel` `preFetchChannel` and `watch` in `loanChannelUtils.js ` I added one more function which is just a regular fetch. Not super elegant since it's basically the same as getCachedChannel....

### DIFF
--- a/src/util/loanChannelUtils.js
+++ b/src/util/loanChannelUtils.js
@@ -121,6 +121,33 @@ export function getCachedChannel(apollo, queryMap, channelUrl, loanQueryVars) {
 }
 
 /**
+ * Gets the loan channel data from the API
+ *
+ * @param {Object} apollo The Apollo client instance
+ * @param {Array} queryMap The map mixin from loan-channel-query-map.js
+ * @param {string} channelUrl The URL of the loan channel
+ * @param {Object} loanQueryVars The loan channel query variables
+ * @returns {Object} The loan channel data, transformed if FLSS
+ */
+export async function getLoanChannel(apollo, queryMap, channelUrl, loanQueryVars) {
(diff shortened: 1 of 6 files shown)


mcstover:
@eddieferrer Just noting that we're in the process of converting LBC to use the category service to resolve loan rows. Thus we should not be making any additional changes to inject FLSS in that context as that service will handle it. The code for this is already in place and will be activated as exp in a couple weeks.

src/util/loanChannelUtils.js
```diff
@@ -120,6 +120,33 @@ export function getCachedChannel(apollo, queryMap, channelUrl, loanQueryVars) {
...
+ * @param {string} channelUrl The URL of the loan channel
+ * @param {Object} loanQueryVars The loan channel query variables
+ * @returns {Object} The loan channel data, transformed if FLSS
+ */
+export async function getLoanChannel(apollo, queryMap, channelUrl, loanQueryVars) {
+	const queryMapFLSS = getFLSSQueryMap(queryMap, channelUrl);
+
+	if (queryMapFLSS) {
+		const experimentActive = checkCachedChannelExperiment(apollo);
+		console.log('experiment active', experimentActive);
```
mcstover, line 137: can remove

mcstover (approved):
makes sense now and thanks for the tests!

PR #4227, opened 2022-09-28
Core 751 quick filters queries

roger-in-kiva:
Following the filter page logic, we are returning the options for every dropdown needed for the quick filter experiment in the category page as `quickFiltersOptions` object like

quickFiltersOptions.gender
quickFiltersOptions.location
quickFiltersOptions.sorting

Experiment is working for some categories only, all of them are covered on this way


mcstover (approved):
This is going to be a pretty "expensive" data fetch but it's likely ok for a short term experiment.

PR #4233, opened 2022-09-29
fix: temporarily remove /cc/kiva-universal from lighthouse as it is f…

mcstover:
…ailing on most mobile loads

Lighthouse tends to use a mobile device for its runs, this path has caused failures for the past few days so we need to remove it for now until we can optimize the page to load properly on throttled devices.


mcstover:
This needs to go out with the hotfix for our homepage exp.

emuvente (approved):

PR #4234, opened 2022-09-29
feat: add donate/supportkiva duplicate page

eddieferrer:
* Creates duplicate of donate/supportus at donate/supportkiva
* Clean up donation form analytics success event

ACK-416

This new page will be used to test new content for the donation page, traffic will be routed/experiment will be driven with optimizely. 

Also renamed the contentful key of `donate/supportus` it is currently web-donate-support-us which doesnt really match the convention of the keys in contentful being the url, I will take care of syncing this change in contentful with the release of this to prod.

### DIFF
--- a/src/pages/Donate/DonateForm.vue
+++ b/src/pages/Donate/DonateForm.vue
@@ -99,15 +99,6 @@ export default {
 			type: String,
 			default: '',
 		},
-		formSubmitAnalytics: {
-			type: Object,
-			default: () => {
-				return {
-					category: 'Donate Form',
-					action: 'click-donate-support-us-form',
-				};
-			},
-		},
 		id: { // used when you have multiple instances of this form on one page.
 			type: String,
 			default: 'instance1',
@@ -197,11 +188,11 @@ export default {
 					});
 				} else {
 					this.$kvTrackEvent(
-						this.formSubmitAnalytics.category,
-						this.formSubmitAnalytics.action,
-						this.buttonText,
+						'donation',
+						'add-to-basket',
+						'donation-one-time',
+						null,
 						// pass donation amount as whole number
-						numeral(this.selectedAmount).value() * 100,
 						numeral(this.selectedAmount).value() * 100
 					);
 					this.$router.push({
--- a/src/pages/Donate/DonateSupportUs.vue
+++ b/src/pages/Donate/DonateSupportUs.vue
@@ -9,7 +9,6 @@
 					:button-text="buttonText"
 					:data="donationValues"
 					:form-disclaimer="formDisclaimer"
-					:form-submit-analytics="formSubmitAnalytics"
 					:show-disclaimer="false"
 					:activate-monthly-option="true"
 				/>
@@ -37,9 +36,9 @@ import DonateForm from '@/pages/Donate/DonateForm';
 import DonateSupportUsRightRail from '@/pages/Donate/DonateSupportUsRightRail';
 import { documentToHtmlString } from '~/@contentful/rich-text-html-renderer';
 
-const pageQuery = gql`query donateContent {
+const pageQuery = gql`query donateContent($contentKey: String) {
 	contentful {
-		entries (contentType: "page", contentKey: "web-donate-support-us")
+		entries (contentType: "page", contentKey: $contentKey)
 	}
 }`;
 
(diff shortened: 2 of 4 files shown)


src/pages/Donate/DonateForm.vue
```diff
@@ -197,11 +188,11 @@ export default {
 					});
 				} else {
 					this.$kvTrackEvent(
```
eddieferrer, line 190: The add a one time donation to basket should be a single analytics event, no matter where this form is displayed or what the route it. Cleaned this up and removed `formSubmitAnalytics` prop

src/pages/Donate/DonateSupportUs.vue
```diff
@@ -71,16 +70,22 @@ export default {
...
-			},
 		};
 	},
 	inject: ['apollo', 'cookieStore'],
 	apollo: {
 		preFetch: true,
 		query: pageQuery,
+		preFetchVariables({ route }) {
+			return {
+				contentKey: route?.meta?.contentfulPage(route)?.trim(),
```
eddieferrer, line 81: get page key from route definition

src/router/routes.js
```diff
@@ -121,6 +121,15 @@ module.exports = [
...
 		meta: {
+			contentfulPage: () => 'donate/supportus',
+			unbounce: true,
+		},
+	},
+	{
+		path: '/donate/supportkiva',
+		component: () => import('@/pages/Donate/DonateSupportUs'),
+		meta: {
+			contentfulPage: () => 'donate/supportkiva',
```
eddieferrer, line 132: Maybe a bit verbose, but it's clear what the contentful key is. I could've also gone with just setting the key to the path....

src/router/routes.js
```diff
@@ -121,6 +121,15 @@ module.exports = [
 		path: '/donate/supportus',
 		component: () => import('@/pages/Donate/DonateSupportUs'),
 		meta: {
+			contentfulPage: () => 'donate/supportus',
+			unbounce: true,
+		},
+	},
+	{
+		path: '/donate/supportkiva',
```
mcstover, line 129: I don't think we should use this path unless we are going to deprecate or combine the Large donor page that resides at `/donate/supportkiva`

src/pages/Donate/DonateSupportUs.vue
```diff
@@ -9,7 +9,6 @@
 					:button-text="buttonText"
 					:data="donationValues"
 					:form-disclaimer="formDisclaimer"
-					:form-submit-analytics="formSubmitAnalytics"
```
mcstover, line 12: We use this on `src/pages/Donate/DonateFromMacroHero.vue` to differentiate the source of these donation entries...

mcstover:
@eddieferrer are you planning to create some type of `donationForm` content group type in Contentful or just continue to use the content ad-hoc.

The previous usage of the `web-...` paradigm was deliberate, due to it not being a full Contentful Page but I'm totally fine moving away from it

src/router/routes.js
eddieferrer: I brought up the fact that `donate/support-kiva` exists with the large donor page. The team was ok with this. There is no plan to deprecate the large donor page

src/pages/Donate/DonateSupportUs.vue
eddieferrer: For the new analytics events the urls should not be present. It is very easy on looker to get the URL of where an event was fired. With the modified analytics event here, you can easily pull the data on looker to get all one-time donations from the donation form. If you wanted to to differentiate the donations that came from the DonateFromMacroHero you could filter that subset with events on URL `/donate/support-kiva`

eddieferrer:
> @eddieferrer are you planning to create some type of `donationForm` content group type in Contentful or just continue to use the content ad-hoc? If so, please plan to add the render component for this new content-group type to cms-page-server as well.
> 
> The previous usage of the `web-...` paradigm was deliberate, due to it not being a full Contentful Page but I'm totally fine moving away from it

No we are not planning to create a new content group. This page will only differ from /donate/supportus in the content of the elements already on contentful. 

I would prefer to move away from the `web-` paradigm. This is the only page that uses it, out of the couple that are rendered this way. This makes them easier to find and In the future we can move to these being full contentful pages without having to rename the key again.

src/router/routes.js
mcstover: Please make sure the business team (managed lending) is looped into this conversation, both b/c of the potential confusion of the url as well as the change in how they will track their donations going forward.

mcstover (approved):

PR #4239, opened 2022-09-30
feat: modify contentful content for donate route

eddieferrer:
I realized that this page was tightly coupled to the content keys on contentful it would not allow the creation of a duplicate page with differing content. 

This decouples a lot of the content and changes the main processing function for the page.


mcstover (approved):
Cool cool! I spoke with design about working on some standard "layouts" that will eventually allow us to generate various column based layouts with Contentful.

PR #4242, opened 2022-10-03
fix: update cps temporary homepage route

mcstover:
CPS routes have been updated, so we need to update the redirect route. Just want to verify this in Dev before I prep the hotfix updates.

I'm organizing additional PRs to update lighthoust-prod.js config and routes in puppet as well.

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -99,7 +99,7 @@ export default {
 				const expVersion = result?.data?.experiment?.version;
 
 				if (expVersion === 'c') {
-					return Promise.reject({	path: '/cps/home' });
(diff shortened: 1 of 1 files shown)


src/pages/Homepage/Homepage.vue
```diff
@@ -99,7 +99,7 @@ export default {
 				const expVersion = result?.data?.experiment?.version;
 
 				if (expVersion === 'c') {
-					return Promise.reject({	path: '/cps/home' });
+					return Promise.reject({	path: '/pgtmp/home' });
```
dyersituations, line 102: What is the `pgtmp` folder? Is `cps` no longer the parent path for CPS?

dyersituations (approved):

PR #4243, opened 2022-10-04
fix: hasOptOut should be false when kvgdpr cookie does not exist

emuvente:
Optimizely isn't loading at all for new users. Tracked down the issue to the `hasOptOut` definition in server-entry, which was returning `true` when the kvgpdr cookie didn't exist. I don't think script.js and IFrameEmbed have this issue because they aren't using optional chaining, but I changed the comparisons in those files as well just to be safe.

```js
const a = {};
a?.b?.c?.indexOf('string') !== -1; // true
a?.b?.c?.indexOf('string') > -1; // false
```

I also found that because the renderedExternals string is cached, all users were getting the same script tags, regardless of having the opt-out cookie or not. I addressed that by making separate strings for opted in and opted out users.

### DIFF
--- a/src/head/script.js
+++ b/src/head/script.js
@@ -3,7 +3,7 @@ export default (config, globalOneTrustEvent) => {
 	const cookies = typeof document !== 'undefined' ? document.cookie.split(';') : [];
 	let optout = false;
 	for (let i = 0; i < cookies.length; i++) { // eslint-disable-line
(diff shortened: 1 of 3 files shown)


dyersituations (approved):

mcstover:
nope, not equal to -1 in that case 🤦

PR #4245, opened 2022-10-04
fix: eco challenge thanks redirect

eddieferrer:
The `myFTDQueryUtil` and the `achievementsQuery` query were racing against each other. So it would sometimes redirect to the eco-challenge thanks page, but usually it myFTD would finish and it would redirect to the regular thanks. This makes sure that we have the achievements data before the redirect 

I'm not sure why this was the case, at first glance I thought the `await` would be enough, but I see they were getting batched by Apollo as 1 query...


mcstover (approved):
Looks good

dyersituations (approved):

PR #4252, opened 2022-10-06
feat: remove querying for category info again for single category car…

eddieferrer:
…ousel

### DIFF
--- a/src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
+++ b/src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="loanChannelId">
+	<div>
 		<div>
 			<h2 class="tw-mb-2 tw-inline">
 				{{ name }}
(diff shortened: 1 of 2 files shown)


src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div v-if="loanChannelId">
+	<div>
```
dyersituations, line 2: Is extra div wrapper still needed?

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -48,6 +48,13 @@ export default {
 		KvPill,
 	},
 	props: {
+		/** prefetched selected channel
+		 * if this data is passed in, it will be used instead of fetching the channel data
+		 * */
+		prefetchedSelectedChannel: {
+			type: Object,
+			default: () => {}
```
dyersituations, line 56: If the default was null, the check below could be simpler.

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
dyersituations: Oh haha, right, Vue 2... 🤦

mcstover (approved):

PR #4255, opened 2022-10-06
MARS-246: Send user attributes to hotjar API

edwinf-barreram:
# Description
- Send user data to hotjar API

# Ticket
[MARS-246](https://kiva.atlassian.net/browse/MARS-246)

### DIFF
--- a/src/components/CorporateCampaign/CampaignThanks.vue
+++ b/src/components/CorporateCampaign/CampaignThanks.vue
@@ -63,6 +63,7 @@ import SocialShareV2 from '@/components/Checkout/SocialShareV2';
 import thanksPageQuery from '@/graphql/query/thanksPage.graphql';
 import { joinArray } from '@/util/joinArray';
 import { userHasLentBefore, userHasDepositBefore } from '@/util/optimizelyUserMetrics';
+import setHotJarUserAttributes from '@/util/hotJarUserAttributes';
 import CampaignPartnerThanks from './CampaignPartnerThanks';
 
 const hasLentBeforeCookie = 'kvu_lb';
@@ -159,6 +160,14 @@ export default {
 
 				userHasLentBefore(hasLentBefore);
 				userHasDepositBefore(hasDepositBefore);
+
+				// MARS-246 Hotjar user attributes
+				setHotJarUserAttributes({
+					userId: this.userId,
+					hasEverLoggedIn: true,
+					hasLentBefore,
+					hasDepositBefore,
+				});
 
 				this.showReceipt = true;
 				await this.$nextTick();
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -467,6 +467,7 @@
 <script>
 import logReadQueryError from '@/util/logReadQueryError';
 import { userHasLentBefore, userHasDepositBefore } from '@/util/optimizelyUserMetrics';
+import setHotJarUserAttributes from '@/util/hotJarUserAttributes';
 import headerQuery from '@/graphql/query/wwwHeader.graphql';
 import gql from 'graphql-tag';
 import KivaLogo from '@/assets/inline-svgs/logos/kiva-logo.svg';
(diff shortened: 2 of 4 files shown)


src/components/CorporateCampaign/CampaignThanks.vue
```diff
@@ -160,6 +161,14 @@ export default {
 				userHasLentBefore(hasLentBefore);
 				userHasDepositBefore(hasDepositBefore);
 
+				// MARS-246 Hotjar user attributes
+				setHotJarUserAttributes({
+					userId: this.userId,
+					hasEverLoggedIn: true,
```
dyersituations, line 167: Can this page only be accessed by logged in user?

src/pages/Thanks/ThanksPage.vue
```diff
@@ -302,6 +303,14 @@ export default {
 		userHasLentBefore(hasLentBefore);
 		userHasDepositBefore(hasDepositBefore);
 
+		// MARS-246 Hotjar user attributes
+		setHotJarUserAttributes({
+			userId: this.userId,
+			hasEverLoggedIn: true,
```
dyersituations, line 309: Can this page only be accessed by logged in user?

src/components/WwwFrame/TheHeader.vue
```diff
@@ -644,6 +648,14 @@ export default {
 
 				userHasLentBefore(hasLentBefore);
 				userHasDepositBefore(hasDepositBefore);
+
+				// MARS-246 Hotjar user attributes
+				setHotJarUserAttributes({
+					userId: this.userId,
```
dyersituations, line 654: Do we need to check if `userId` is still `0` here?

src/components/CorporateCampaign/CampaignThanks.vue
mcstover: The page is open but this component would only show up after user was logged in and successfully executed a checkout...

src/pages/Thanks/ThanksPage.vue
mcstover: `this.userId` isn't defined on the thanks page. You can conditionally check for it though...
`userId: !data?.my?.userAccount?.id ?? null`

src/components/CorporateCampaign/CampaignThanks.vue
```diff
@@ -160,6 +161,14 @@ export default {
 				userHasLentBefore(hasLentBefore);
 				userHasDepositBefore(hasDepositBefore);
 
+				// MARS-246 Hotjar user attributes
+				setHotJarUserAttributes({
+					userId: this.userId,
```
mcstover, line 166: `this.userId` isn't defined on this component, you should however be able to check for it on the `lender` data property that is updated...something like `this.lender?.my?.userAccount?.id` ... need to check the actual page data structure at this point in time.

You can checkout on `/cc/kiva-universal` without a promo code to test this behavior.

mcstover (approved):

src/pages/Thanks/ThanksPage.vue
```diff
@@ -302,6 +303,14 @@ export default {
 		userHasLentBefore(hasLentBefore);
 		userHasDepositBefore(hasDepositBefore);
 
+		// MARS-246 Hotjar user attributes
+		setHotJarUserAttributes({
+			userId: data?.my?.userAccountId?.id,
+			hasEverLoggedIn: true,
```
emuvente, line 309: hasEverLoggedIn isn't necessarily true here. The user may have used guest checkout

mcstover: CampaignThanks is currently only shown after a successful checkout in that context which can only occur after being logged in. Guest Checkout isn't available here so it's prob ok for now to remain set to `true`

emuvente: Agreed that CampaignThanks can just be set to `true`. However, for this line in ThanksPage I think it should be changed to `hasEverLoggedIn: isLoggedIn`.

emuvente (approved):
👍 Looks good!

PR #4256, opened 2022-10-06
Vue 1308 add tags filter section

dyersituations:
https://kiva.atlassian.net/browse/VUE-1308

- Added tag filter to loan search
- Only public tags are displayed
- Displayed labels have `#` removed
- Ensured that saved search works with tags
- Loan channel query map updated with new tag filter
- Query string updated and parsed for both Algolia and new versions of the query param
- Added missing unit tests for existing method
- Silenced some Vue router errors that can be ignored

image


src/util/loanSearch/queryParamUtils.js
```diff
@@ -202,6 +204,14 @@ export function updateQueryParams(loanSearchState, router, queryType) {
 
 	// Vue throws duplicate navigation exception when identical paths are pushed to the router
 	if (!doParamsMatch) {
-		router.push({ ...router.currentRoute, query: newParams, params: { noScroll: true, noAnalytics: true } });
+		router.push({ ...router.currentRoute, query: newParams, params: { noScroll: true, noAnalytics: true } })
+			.catch(e => {
+				const { isNavigationFailure, NavigationFailureType } = VueRouter;
```
mcstover, line 209: These weren't available from the passed `router` object?

mcstover (approved):
Nice!

eddieferrer:
This is great, I think this will help the upcoming game experiment since the game version of the lend-by-category/eco-loans can now pull from flss 👍

src/plugins/loan-channel-query-map.js
```diff
@@ -715,6 +739,11 @@ export default {
 					id: 151,
 					url: 'b-i-p-o-c-business',
 					queryParams: 'status=fundRaising&country=US&loanTags=51&distributionModel=direct',
+					flssLoanSearch: {
+						countryIsoCode: ['US'],
+						sectorId: [51],
```
eddieferrer, line 744: Is this intentional? The query param has `loanTags=51` but the FLSS has `sectorId=51`

src/util/loanSearch/queryParamUtils.js
dyersituations: These aren't on the router instance but on the `VueRouter` module itself. These two items aren't exported individually from the Vue Router module. This pattern is pulled from the official Vue Router docs.

PR #4258, opened 2022-10-07
feat: use loanCardFields fragment for kiva classic basic loan card

eddieferrer:
Added fields to commonly used fragment (used in both flss and legacy lend by category queries) then kiva classic loan card can just load these fields if they are present. 

This makes it so the eco-challenge page loads all the cards almost instantly since we have already queried for them when getting the categories on `src/pages/Lend/LoanChannelCategoryClimateExperiment.vue`

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -231,10 +231,13 @@ import LoanProgressGroup from '@/components/LoanCards/LoanProgressGroup';
 import LoanMatchingText from '@/components/LoanCards/LoanMatchingText';
 import SummaryTag from '@/components/BorrowerProfile/SummaryTag';
 import { setLendAmount } from '@/util/basketUtils';
+import loanCardFieldsFragment from '@/graphql/fragments/loanCardFields.graphql';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
 import KvUiButton from '~/@kiva/kv-components/vue/KvButton';
 
-const loanQuery = gql`query kcBasicLoanCard($basketId: String, $loanId: Int!) {
+const loanQuery = gql`
+	${loanCardFieldsFragment}
+	query kcBasicLoanCard($basketId: String, $loanId: Int!) {
 	shop (basketId: $basketId) {
 		id
 		basket {
@@ -250,55 +253,12 @@ const loanQuery = gql`query kcBasicLoanCard($basketId: String, $loanId: Int!) {
 	lend {
 		loan(id: $loanId) {
 			id
-			distributionModel
-			geocode {
-				city
-				state
-				country {
-					name
-					isoCode
-				}
-			}
-			image {
-				id
-				hash
-			}
-			name
-			sector {
-				id
-				name
-			}
-			whySpecial
-
-			# for isLentTo
-			userProperties {
-				lentTo
-			}
-
-			# for loan-use-mixin
-			use
-			status
-			loanAmount
-			borrowerCount
-
-			# for percent-raised-mixin
-			loanFundraisingInfo {
-				fundedAmount
-				reservedAmount
-			}
-
-			# for time-left-mixin
-			plannedExpirationDate
+			...loanCardFields
 
 			# for loan-progress component
 			unreservedAmount @client
 			fundraisingPercent @client
 			fundraisingTimeLeft @client
-
-			# for matching-text component
-			isMatchable
-			matchingText
-			matchRatio
 		}
 	}
 }`;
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -511,6 +471,22 @@ export default {
 	beforeDestroy() {
 		this.destroyViewportObserver();
 	},
+	created() {
+		try {
```
eddieferrer, line 475: This will throw an invariant error when the fragment is in the cache but missing fields. Lets just swallow the error and do nothing. KivaClassicBasicLoanCard just does what it used to do before this PR in that case

src/graphql/fragments/loanCardFields.graphql
```diff
@@ -1,5 +1,6 @@
 fragment loanCardFields on LoanBasic {
 	id
+	distributionModel
```
eddieferrer, line 3: These additions here are the ones that were used in the loan card but were not present in this fragment

src/components/LoanCards/KivaClassicBasicLoanCard.vue
dyersituations: Since the fragment now has the extra fields, in what case will the loan in the cache have missing fields?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -511,6 +471,22 @@ export default {
...
+			// Attempt to read the loan card fragment from the cache
+			// If cache is missing fragment fields, this will throw and invariant error
+			const loanCardFieldFragment = this.apollo.readFragment({
+				id: `LoanPartner:${this.loanId}`,
```
mcstover, line 479: Should this be LoanBasic to cover both LoanDirect and LoanPartner?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
eddieferrer: I found at least one case, on the homepage (i guess now old homepage) the multi category carousel uses a query which just gets the loan ID. `src/components/LoanCollections/KivaClassicMultiCategoryCarousel.vue:121`

mcstover: That's correct b/c we don't want all the loan data prefetched.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
mcstover: From what I see in the Apollo Cache this will be either `LoanPartner` or `LoanDirect`...so maybe we need to somehow switch on the distribution type...

emuvente: I think it works to use `LoanBasic` there, as all the fields being requested exist on that type. The Apollo client understands the relationship between `LoanBasic`, `LoanPartner`, and `LoanDirect`, because we define the [possible types](https://github.com/kiva/ui/blob/5834fb39246bf87a97582e608fbbd008269afa2b/src/api/apollo.js#L29).

mcstover: That would be great!

mcstover: Using `LoanBasic` Does appear to work in the VM!

mcstover: hmm that's curious and the change I also made... I was testing by using the carousel switcher on the home page and seeing that clicking a previously loaded category would no longer hit the server so I thought it was getting it from the cache...

mcstover: Which context were you using?

eddieferrer: I tried both the homepage multi carousel and the lbc eco page -- attempting to read all 3 different fragments LoanBasic, LoanDirect and LoanPartner, but the only one that I can get to return anything is LoanPartner

eddieferrer: > I think it works to use `LoanBasic` there, as all the fields being requested exist on that type. The Apollo client understands the relationship between `LoanBasic`, `LoanPartner`, and `LoanDirect`, because we define the [possible types](https://github.com/kiva/ui/blob/5834fb39246bf87a97582e608fbbd008269afa2b/src/api/apollo.js#L29).

@emuvente thoughts on this? I could not get LoanBasic to work. And if I look at my cache I only see LoanPartner or LoanDirect entries

eddieferrer:
Added a check for either Partner or Direct fragment, confirmed working 👍

mcstover (approved):
Glad you found a functioning way to do this. Can you please make a ticket to port this optionality to cms-page-server.

emuvente (approved):
👍

PR #4251, opened 2022-10-06
fix: pass query origin context on all flss queries

mcstover:
Passes the `origin` variable into all `fundraisingLoans` queries.

Screen Shot 2022-10-05 at 5 56 07 PM

Screen Shot 2022-10-05 at 5 55 59 PM

PR #4261, opened 2022-10-10
refactor: climate lbc page to work with flss query

eddieferrer:
I used to do a query with multiple channel ids for the main category 'eco-friendly'. This worked because eco-friendly uses a tagId which was not supported with FLSS

With the added support for tagId now I need to do separate queries here to get all the categories. The problem stems from 
the function `transformFLSSData` returning `data?.lend?.loanChannelsById?.[0]` which only returns the first item in that array, so it does not support multiple categories.

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
+++ b/src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
@@ -92,6 +92,7 @@ import {
 	mdiClose
 } from '@mdi/js';
 import IconCalendar from '@/assets/icons/inline/eco-challenge/calendar.svg';
+import gql from 'graphql-tag';
 import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
 import KvTextLink from '~/@kiva/kv-components/vue/KvTextLink';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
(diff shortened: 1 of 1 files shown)


mcstover:
We need to get https://github.com/kiva/ui/pull/4251/files merged so this can be adapted to pass the `origin` parameter.

src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
```diff
@@ -191,14 +204,24 @@ export default {
...
-				// Build loanQueryVars since SSR doesn't have same context
-				{ ids: [...secondaryEcoLoanChannelIds, targetedLoanChannelID], limit, offset }
-			);
+
+			const promisesArrayOfSecondaryChannels = secondaryEcoLoanChannelIds.map(channelId => {
+				return preFetchChannel(
+					client,
+					loanChannelQueryMapMixin.data().loanChannelQueryMap,
+					loanChannelQueryMapMixin.data().loanChannelQueryMap.find(channel => channel.id === channelId)?.url,
+					{ ids: [channelId], limit, offset }
```
mcstover, line 213: I think if you merge in main, you'll have some variables already present or imported to pass the `origin` param.

eddieferrer: > I think if you merge in main, you'll have some variables already present or imported to pass the `origin` param.

Ok will do!

mcstover (approved):

PR #4264, opened 2022-10-11
feat: update category pages to show selected quick filters

egoroza:
adds `selectedQuickFilters` to query map that gets passed into cat query so that the filters get added to the `filterObject` and is updated through the watch query. when the experiment is off, `selectedQuickFilters` will always remain an empty object and should not affect the current query structure


src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -182,6 +186,7 @@ export default {
 		toggleRegions() {
 			this.showRegions = !this.showRegions;
 			this.selectedRegion = null;
+			this.updateLocation(this.selectedCountries);
```
dyersituations, line 189: We typically use the pattern of emitting values rather than passing a callback to a component. See the other loan filter components that emit the changed values.

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -556,7 +578,7 @@ export default {
 			this.quickFiltersOptions.gender = [
 				{
 					title: 'All genders',
-					key: '',
+					key: 'all',
```
dyersituations, line 581: When did this key change? It used to mean that empty meant all values.

src/util/loanChannelUtils.js
```diff
@@ -170,7 +173,7 @@ export function watchChannelQuery(apollo, queryMap, channelUrl, loanQueryVars, n
 		experimentActive = checkCachedChannelExperiment(apollo);
 
 		if (experimentActive) {
-			observer = watchLoanChannel(apollo, queryMapFLSS, loanQueryVars);
+			observer = watchLoanChannel(apollo, { ...queryMapFLSS, ...selectedQuickFilters }, loanQueryVars);
```
dyersituations, line 176: `{ ...queryMapFLSS, ...selectedQuickFilters }` can be stored in a const at the top of this method instead of needing to create this object twice in this method.

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -381,12 +385,19 @@ export default {
...
 
 		// Prevent pop-in by loading data from the Apollo cache manually here instead of just using the subscription
-		const baseData = getCachedChannel(
-			this.apollo,
-			this.loanChannelQueryMap,
-			this.targetedLoanChannelURL,
-			this.loanQueryVars
-		);
+		const baseData = this.enableQuickFilters
+			? getFilteredLoanChannel(
```
dyersituations, line 389: Is there a situation where the user would be able to get this filtered channel from the cache? In this current setup, it will always be a slightly slowly, since `fetchLoanChannel` returns a promise. On the same note, does this need an `await` before it?

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
egoroza: yeah we only want to update when the `See {{ totalLoans }} loans` button is clicked which uses this function

src/pages/Lend/LoanChannelCategoryControl.vue
egoroza: oops forgot to change this back from testing

PR #4266, opened 2022-10-11
feat: get thankspage data using window object

christian14b:
As an effort to better customize tests in Optimizely, Product team needs dynamic information about the user

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -2,6 +2,7 @@
 	<div
 		id="app"
 		class="tw-h-full tw-bg-primary"
+		:data-hydrated="dataHydrated"
 	>
 		<router-view />
 		<vue-progress-bar />
@@ -17,6 +18,11 @@ import unbounceEventMixin from '@/plugins/unbounce-event-mixin';
 
 export default {
 	name: 'App',
+	data() {
+		return {
+			dataHydrated: false
+		};
+	},
 	components: {
 		TheTipMessage,
 	},
@@ -207,6 +213,9 @@ export default {
 			]))
 		};
 	},
+	mounted() {
+		this.dataHydrated = true;
+	}
 };
 </script>
 


src/server-entry.js
```diff
@@ -182,6 +184,30 @@ export default context => {
 					logFormatter(`data pre-fetch: ${Date.now() - s}ms`);
 					sp = new Date();
 				}
+				const transactionId = router.currentRoute.query?.kiva_transaction_id
+					? numeral(router.currentRoute.query?.kiva_transaction_id).value()
+					: null;
+
+				const data = transactionId ? apolloClient.readQuery({
```
dyersituations, line 191: Is this new code running on each page load? And where does this query initially get ran? I see it's being pulled from the cache here in the server entry.

mcstover:
I think this all needs to go into the ThanksPage and come out of our global server scope. Additionally, the setting of this data needs to be gated behind 3rd party opt out.

mcstover (changes requested):
I think this all needs to go into the ThanksPage and come out of our global server scope. Additionally, the setting of this data needs to be gated behind 3rd party opt out.

src/server-entry.js
emuvente: Yes, this will run on each page load, but it will only try to read data from the cache. If the query has been executed this will return a result, but be undefined otherwise.

emuvente:
> I think this all needs to go into the ThanksPage and come out of our global server scope

@mcstover We've limited this to using the thank page query for now because that will unlock some of the functionality that's being requested, but the goal is to have similar data available on all pages, which is why I suggested putting it in server entry.

src/server-entry.js
```diff
@@ -182,6 +184,30 @@ export default context => {
 					logFormatter(`data pre-fetch: ${Date.now() - s}ms`);
 					sp = new Date();
 				}
+				const transactionId = router.currentRoute.query?.kiva_transaction_id
```
emuvente, line 187: I think this code to get the data should be in a function in `optimizelyUserMetrics.js`. Then server-entry can call it:
```js
import { buildUserDataGlobal } from '@/util/optimizelyUserMetrics';

...

__OPTIMIZELY_DATA__: buildUserDataGlobal(router, apolloClient)
```

src/server-entry.js
emuvente: @christian14b Calls to `readQuery` need to be wrapped in a try/catch because they will throw exceptions if the data isn't present.

emuvente:
> This seems like an opportunity to use the standard data_layer instead of a specialized selection just for Optimizely.

Agreed! Is there a standard for that? At the very least I think the global should be called `dataLayer` instead of `__OPTIMIZELY_DATA__`.

mcstover:
We currently setup the `dataLayer` [here](https://github.com/kiva/ui/blob/main/src/head/script.js#L13) so I think that would need to move up into server entry...so this could push to it...

mcstover (approved):

dyersituations (approved):

emuvente (approved):

emuvente (approved):

PR #4268, opened 2022-10-11
fix: enable some traffic to landing pages in lighthouse for ui and cp…

mcstover:
…s comparison


dyersituations (approved):
Looking forward to seeing what lighthouse thinks of prebuilt pages... I know it's not with these changes, but once we get pre-built working, it'll be cool to see.

emuvente (approved):

PR #4269, opened 2022-10-12
Cit 14 2

michelleinez:


src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -141,6 +133,20 @@ export default {
 		};
 	},
 	computed: {
+		loanIds() {
+			console.log(_map(this.loans, 'id'));
+			return _map(this.loans, 'id');
+		},
```
mcstover, line 139: Please refactor to use optional chaining. We're slowing removing any use of lodash as we refactor or update code so please anytime you see something like this, switch to using plain js.

src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -71,26 +59,26 @@
...
 </template>
 
 <script>
 import basicLoanQuery from '@/graphql/query/basicLoanData.graphql';
-import KvCarousel from '@/components/Kv/KvCarousel';
-import KvCarouselSlide from '@/components/Kv/KvCarouselSlide';
 import KvLoadingSpinner from '@/components/Kv/KvLoadingSpinner';
-import LoanCardController from '@/components/LoanCards/LoanCardController';
+
+import _map from 'lodash/map';
```
mcstover, line 71: Please remove this import and use plain js.

src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -203,12 +209,16 @@ export default {
...
+			this.$kvTrackEvent('carousel', 'click-carousel-horizontal-scroll', interaction);
+		},
 		addToBasket(payload) {
 			this.loanAdded = true;
 			this.$emit('add-to-basket', payload);
 		},
 		showLoanDetails(payload) {
-			const selectedLoan = this.loans.find(loan => loan.id === payload.loanId);
+			const selectedLoan = this.loans.find(loan => String(loan.id) === String(payload.loanId));
 			this.$emit('show-loan-details', selectedLoan);
```
mcstover, line 222: Does this method need to `preventDefault` or anything like that to ensure the clicks coming up from the child components do not navigate?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1418,7 +1415,7 @@ export default {
 			}
 		},
 		jumpToLoans() {
-			this.$refs.campaignLoanSection.scrollIntoView({ behavior: 'smooth' });
+			this.$refs.mlLoanDisplay[0].$refs.campaignLoanSection.scrollIntoView({ behavior: 'smooth' });
```
dyersituations, line 1418: Is there guaranteed to have 1 item here in `mlLoanDisplay`? If not, optional chaining would be good.

src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -3,8 +3,9 @@
 		{{ loanUseFiltered }}
 		<kv-text-link
 			v-if="loanId && showLearnMore"
-			:to="`/lend/${loanId}`"
+			:to="showLoanDetails ? '' : `/lend/${loanId}`"
 			v-kv-track-event="['Lending', 'click-Read more', 'Learn more', loanId]"
+			@click="showLoanDetails({loanId})"
```
dyersituations, line 8: Will clicking this link error if `showLoanDetails` is not defined?

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -12,11 +12,12 @@
 		<div
 			v-if="!isLoading"
 			class="tw-relative"
+			@click="showLoanDetails({loanId})"
 		>
 			<!-- If allSharesReserved, disable link by making it a span -->
 			<router-link
 				:is="allSharesReserved ? 'span' : 'router-link'"
-				:to="`/lend/${loanId}`"
+				:to="showLoanDetails ? '' : `/lend/${loanId}`"
```
dyersituations, line 20: Will this link be non-clickable if `showLoanDetails` is undefined?

src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -3,8 +3,9 @@
 		{{ loanUseFiltered }}
 		<kv-text-link
 			v-if="loanId && showLearnMore"
-			:to="`/lend/${loanId}`"
+			:to="showLoanDetails ? '' : `/lend/${loanId}`"
```
eddieferrer, line 6: I believe you can pass undefined to exclude the prop altogether, so this line could be: `:to="showLoanDetails ? undefined : `/lend/${loanId}`"` and then if showLoanDetails this prop will not be passed at all

mcstover: The click on this text link ultimately needs to emit an event we can capture... @michelleinez is it already bubbling up something you can use to show the In Context Borrower Profile? Or, should it actually call `showLoanDetails` ?

mcstover: That is good advice though @eddieferrer for the right situation

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -140,7 +142,7 @@
 				variant="secondary"
 				v-if="isInBasket"
 				v-kv-track-event="['Lending', 'click-Read more', 'checkout-now-button-click', loanId, loanId]"
-				to="/basket"
+				:to="customCheckoutRoute.length > 0 ? customCheckoutRoute : '/basket'"
```
eddieferrer, line 145: You can just check for falsy on the empty string here, you don't need to check for the length.

src/components/BorrowerProfile/LoanUse.vue
michelleinez: No, there's a default function that sets itself when it's not defined.  That default function is empty and does nothing.

michelleinez: I'm actually changing it now also -- so that it's not a function anymore but a boolean that causes the component to emit an event and pass the loan id up to the parent.  It's cleaner than this mess.

src/components/BorrowerProfile/LoanUse.vue
michelleinez: I'm changing this to emit an event now.  The to field will have a boolean instead which defaults to false so that the default behavior doesn't change when the prop is undef.

src/components/BorrowerProfile/LoanUse.vue
```diff
@@ -3,8 +3,9 @@
 		{{ loanUseFiltered }}
 		<kv-text-link
 			v-if="loanId && showLearnMore"
-			:to="`/lend/${loanId}`"
+			:to="customLoanDetails? '' : `/lend/${loanId}`"
```
mcstover, line 6: Seems like we should use `null` instead of `''` (empty String) if `customLoanDetails`. Did you try this advice from Eddie? Also spacing..

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -12,11 +12,12 @@
 		<div
 			v-if="!isLoading"
 			class="tw-relative"
+			@click="showLoanDetails"
 		>
 			<!-- If allSharesReserved, disable link by making it a span -->
 			<router-link
 				:is="allSharesReserved ? 'span' : 'router-link'"
-				:to="`/lend/${loanId}`"
+				:to="customLoanDetails? '' : `/lend/${loanId}`"
```
mcstover, line 20: Same comment about using `null` and spacing.

mcstover (commented):
Nice!

PR #4270, opened 2022-10-12
Vue 1252 distribution model filter

dyersituations:
https://kiva.atlassian.net/browse/VUE-1252

- Added distribution model filter to loan search
- Abstracted gender filter to generic radio group filter -> both filters now have a "All loans" default instead of the previous "All genders"
- New event is `click-distributionModel-filter` -> there wasn't an event for this filter in the legacy app
- Gender facets (and new distribution model facets) are now being pulled from FLSS -> non-binary hidden because it's not included in the order array

image

After initial PR feedback:

image


src/components/Lend/LoanSearch/LoanSearchFilter.vue
```diff
@@ -93,6 +99,20 @@
 					</h2>
 				</button>
 			</kv-accordion-item>
+			<kv-accordion-item id="acc-distribution-model" :open="false">
```
mcstover, line 102: Please move this above the Advanced Filters section outside of that section. My apologies for the bad spec.

This will need to be gated by the experiment Setting there as well.

dyersituations: Should it remain in an accordion like in the screenshot? It's currently in an accordion just below the advanced filters accordion. If it's outside an accordion, does it need a header like in legacy? The gender filter doesn't have a header, but I could see these options being more confusing without a header.

dyersituations:
> @dyersituations There will need to be optionality for the "All" option added to this setup so that `All genders` can be used for Gender and other custom phrases can be used for other radio group filter types where appropriate.

I wondered if this was needed. The notes in the tickets for the other filters all just mentioned "all filters". What should the all be for distribution model?

mcstover:
> > @dyersituations There will need to be optionality for the "All" option added to this setup so that `All genders` can be used for Gender and other custom phrases can be used for other radio group filter types where appropriate.
> 
> I wondered if this was needed. The notes in the tickets for the other filters all just mentioned "all filters". What should the all be for distribution model?

`All loans` will be fine for the first run of the other types, unless we have some other verbiage in place on legacy lend. We'll need to review.

src/components/Lend/LoanSearch/LoanSearchFilter.vue
mcstover: This should be an independent section without an accordian. It should have a header that matches legacy lend.

dyersituations: Header and sub header? We have taken some liberties in transferring filters to the new UI, for example legacy sector does have a header, where in the new UI there isn't. I think we need to get the specifics into the tickets to reduce churn on these PRs.

image

mcstover: Good point. Let's leave off the sub head for now and spend some time refining the other tickets before getting started on them.

dyersituations: Done. Reused capitalized header styling for consistency with filter weight of importance.

image

dyersituations:
> > > @dyersituations There will need to be optionality for the "All" option added to this setup so that `All genders` can be used for Gender and other custom phrases can be used for other radio group filter types where appropriate.
> > 
> > 
> > I wondered if this was needed. The notes in the tickets for the other filters all just mentioned "all filters". What should the all be for distribution model?
> 
> `All loans` will be fine for the first run of the other types, unless we have some other verbiage in place on legacy lend. We'll need to review.

@mcstover Done. The all loans title is now configurable via prop:

image

mcstover (approved):
Very nice!

PR #4274, opened 2022-10-13
TTI performance enhancements

emuvente:
A mix of changes to help with performance I found while trying to minimize time-to-interactive. There is much more that could be done for that, but these are some small wins along the way.

### DIFF
--- a/build/webpack.client.base.conf.js
+++ b/build/webpack.client.base.conf.js
@@ -17,7 +17,6 @@ module.exports = merge(baseWebpackConfig, {
 			chunks: 'all',
 			maxAsyncRequests: Infinity, // default is 6
 			cacheGroups: {
-				default: false,
 				vendors: false
 			}
 		}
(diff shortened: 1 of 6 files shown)


mcstover (approved):
Nice clean up and optimization!

build/webpack.client.base.conf.js
```diff
@@ -17,7 +17,6 @@ module.exports = merge(baseWebpackConfig, {
 			chunks: 'all',
 			maxAsyncRequests: Infinity, // default is 6
 			cacheGroups: {
-				default: false,
```
emuvente, line 20: This is causing common modules to be duplicated across chunks instead of being added to a shared chunk, greatly increasing the initial payload sent to users.

dyersituations (approved):

eddieferrer (approved):
Awesome. 🧹 👍

PR #4279, opened 2022-10-14
fix: images aren't showing up on cc page for production build

michelleinez:


mcstover (approved):

mcstover:
Just noting, that when `null` is passed to the router link `to` property rendering of the link and all it's content is skipped!

PR #4282, opened 2022-10-17
Core 775 helpme choose triggers

roger-in-kiva:


src/components/LoansByCategory/HelpmeChoose/HelpmeChooseTrigger.vue
```diff
@@ -0,0 +1,62 @@
...
+		return {
+			progressIcon,
+			kivaIcon,
+			appleIcon
+		};
+	},
+	computed: {
+		title() {
+			let str = '';
+			if (this.variant === 'amountLeft') {
```
dyersituations, line 36: Doesn't need to change, but these two methods with stacked if statements would typically be solved with switch statements instead.

dyersituations (approved):

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseTrigger.vue
roger-in-kiva: Nice suggestion, even we can update this removing str definition and just return the value for every case, but good to go for the experiment. Thanks @dyersituations !

mcstover (approved):

PR #4289, opened 2022-10-18
Core 811 fix select all

roger-in-kiva:


src/components/LoansByCategory/QuickFilters/CheckboxList.vue
```diff
@@ -78,7 +78,11 @@ export default {
 					if (exists) this.selected.splice(index, 1);
 				} else if (!exists) this.selected.push(item.value);
 			});
-			this.updateSelected(this.selected, undefined, true);
+			if (isAll && typeof item === 'object') {
```
mcstover, line 81: `item` wouldn't be defined here at all right?

roger-in-kiva: Logic updated, thanks!

src/components/LoansByCategory/QuickFilters/CheckboxList.vue
```diff
@@ -75,13 +75,16 @@ export default {
 				const index = this.selected.indexOf(item.value);
 				const exists = index !== -1;
 				if (isAll) {
-					if (exists) this.selected.splice(index, 1);
-				} else if (!exists) this.selected.push(item.value);
+					if (exists) {
+						this.updateSelected(this.selected, item.value);
```
dyersituations, line 79: It looks like the component is emitting an event for each item that was selected when select all is toggled. There should only be one event emitted for a user action.

roger-in-kiva: Yeah, this is b/c the updater on parent was planned for one event. Selecting all is the less probable action here, main idea is to select a country you are interested.
We need to launch this asap, but if you believe a refactor is needed for this experiment I can fix it.

dyersituations: Gotcha. If it works alright in functional testing, please add a `TODO` here to refactor into one emitted event. In general you really only want one distinct event per user action, unless there's a chain of actions. Here we have one user interaction with one result (reset filter).

roger-in-kiva: Good! Agree, I added a new ticket for this https://kiva.atlassian.net/browse/CORE-820

dyersituations: Thanks!

dyersituations (approved):

PR #4290, opened 2022-10-18
fix: avoid skipping userData query

christian14b:
Remove condition to get `hasEverLoggedIn` value for `hojtar` user attributes


emuvente (changes requested):
The optimizelyUserDataQuery is expensive to run for every request, which is which it was gated behind the if statement. Please leave that in place and make a separate query just to get the hasEverLoggedIn value, which executes very quickly.

mcstover (approved):

emuvente (approved):
Thanks! Looks good!

PR #4292, opened 2022-10-19
feat: body scroll was blocked when location dropdown is visible

roger-in-kiva:


src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -206,6 +206,11 @@ export default {
 		toggleRegions() {
 			this.showRegions = !this.showRegions;
 			this.selectedRegion = null;
+			if (this.showRegions) {
+				document.documentElement.style.overflow = 'hidden';
```
dyersituations, line 210: Usually you don't want to change the document directly in Vue. Does this dropdown behave different than other dropdowns we have?

roger-in-kiva: Yeah, behavior is diff for mobile, need to block the scroll of the body and allow it into the modal.

dyersituations: I guess I'm confused as to what this is solving exactly. Is there a reason this dropdown in particular needs a scrolling fix? Do we do this for other dropdowns? I didn't see this code anywhere else.

roger-in-kiva: Yeah context is here, 0:26 sec aprox

https://kivahq.slack.com/files/U016NHS0FM5/F0479EMGQM7/screenshare_-_2022-10-17_2_45_37_pm.webm?origin_team=TAY80LL2D&origin_channel=D03FN1M1KDY

mcstover: In the past, we've created a full screen lightbox for filtering on mobile which could be how we address this in a future, full implementation as well.

roger-in-kiva: Yeah, there is an overlay in design similar to having a lightbox, but is out of the scope for the experiment.
If this becomes a feature we'll be doing all needed refactor for a full implementation.

dyersituations: Thanks for the context.

dyersituations (approved):

PR #4295, opened 2022-10-20
MARS-274: Improve loan card performance in new home page experiment

edwinf-barreram:
# Description
- Added changes to improve loan card performance in Category and Carousel sections

# Ticket
[MARS-274](https://kiva.atlassian.net/browse/MARS-274)

### DIFF
--- a/src/components/Contentful/HomePage/NewHomeLoansCardCarousel.vue
+++ b/src/components/Contentful/HomePage/NewHomeLoansCardCarousel.vue
@@ -8,8 +8,8 @@
 			<kv-page-container>
 				<div>
 					<kiva-loan-card-carousel
-						:selected-channel="selectedChannel"
-						:loans="selectedChannelLoanIds"
+						:selected-channel="currentSelectedChannel"
+						:loans="loanList"
 					/>
 				</div>
 			</kv-page-container>
@@ -18,8 +18,6 @@
 </template>
 
 <script>
-import gql from 'graphql-tag';
-import loanCardFieldsFragment from '@/graphql/fragments/loanCardFields.graphql';
 import KivaLoanCardCarousel from '@/components/LoanCollections/HomeExp/KivaLoanCardCarousel';
 import contentfulStylesMixin from '@/plugins/contentful-ui-setting-styles-mixin';
 import SectionWithBackgroundClassic from '@/components/Contentful/SectionWithBackgroundClassic';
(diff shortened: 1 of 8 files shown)


src/components/Contentful/HomePage/NewHomeLoansCardCarousel.vue
```diff
@@ -9,7 +9,7 @@
 				<div>
 					<kiva-loan-card-carousel
 						:selected-channel="selectedChannel"
-						:loan-ids="selectedChannelLoanIds"
+						:loans="selectedChannelLoanIds"
```
dyersituations, line 12: Rename `selectedChannelLoanIds`, since it no longer just has IDs.

src/components/Contentful/HomePage/NewHomeLoansCardCarousel.vue
```diff
@@ -61,82 +68,12 @@ export default {
...
-				this.loanChannelData = loanChannels;
-				// Activate the first channel available
-				const initialChannel = this.combinedLoanChannelData[0];
-				this.selectedChannel = initialChannel;
-				// Make the carousel visible
-				this.showCarousel = true;
-			});
+	watch: {
+		loans() {
+			if (this.loans.length > 0 && this.selectedChannel.id) {
```
dyersituations, line 74: Why are the loans being passed in, but only the length checked? Can we just pass in the `selectedChannel` to this component?

src/pages/ContentfulPage.vue
```diff
@@ -348,6 +383,28 @@ export default {
 				this.pageBackgroundColor = pageLayout.pageBackgroundColor ?? '';
 				this.pageFrame = getPageFrameFromType(page.pageType);
 				this.contentGroups = getContentGroups(pageData);
+
+				if (page.key === 'hp/crowdfund-for-good') {
```
dyersituations, line 387: I'm not sure if moving this channel-specific logic to the contentful page is the most appropriate choice. The contentful page runs a bunch of different page types, so having this type of specific logic here is a little unusual. Have we discussed putting this type of page specific logic in this wrapper page component? If not, it might make more sense to extract this logic into a util and call the util in the two files that need this channel/loan data.

mcstover: I agree, we should not be including any component specific or loan related configuration or data fetchs in this file. We Also should not be calling any utils for any targted content groups.

This file needs to remain totally generic with the exeception of the redirect controls we have in place already.

mcstover (changes requested):

edwinf-barreram:
Closing because this improvement was addressed on PR [4302](https://github.com/kiva/ui/pull/4302)

PR #4296, opened 2022-10-20
feat: dynamic copy was added for every category and subcategory in he…

roger-in-kiva:
…lpme choose exp


src/components/LoansByCategory/HelpmeChoose/BorrowerSelector.vue
```diff
@@ -0,0 +1,4 @@
+<template>
+	<div>
```
dyersituations, line 2: Are these placeholders for future tickets?

dyersituations (approved):

PR #4297, opened 2022-10-20
feat: gender added to facets for location options

roger-in-kiva:


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -489,6 +489,8 @@ export default {
 		updateQuickFilters(filter) {
 			if (filter.gender !== undefined) {
```
dyersituations, line 490: The default of `gender` is `null`, so we might want to just do a `if (filter.gender)` check here. Or does this also receive `null` when filters are cleared?

roger-in-kiva: Yeah, receive `null` and becomes a false.

dyersituations (approved):

PR #4298, opened 2022-10-20
Vue 1307 add keyword search field

dyersituations:
https://kiva.atlassian.net/browse/VUE-1307

- Implemented keyword search filter/input
- Brought over placeholder from legacy, "Search borrower story"
- I know we're moving away from lodash, but using `debounce` from lodash is quite convenient... And we want debounce on a text input that changes filters...
- Used the same query param as legacy, `queryString`

image


mcstover (approved):
Very cool! Using the single import of lodash is the way to go in this case as well.

PR #4299, opened 2022-10-20
fix: make sure game users see the FLSS loan results in carousel

eddieferrer:
This will align the loan set in the carousels with what the achievement service is using.

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -160,6 +160,22 @@ export default {
 							});
 						}).then(() => {
 							return fetchExperimentSettings('eco_challenge', client);
+						}),
+						client.query(
+							{
+								query: experimentAssignmentQuery,
+								variables: { id: 'loan_channel_flss_query_v1' }
+							}
+						).then(() => {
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -160,6 +160,22 @@ export default {
...
+						).then(() => {
+							return client.mutate({
+								mutation: updateExperimentVersion,
+								variables: {
+									id: 'loan_channel_flss_query_v1',
+									version: 'b'
+								}
+							});
+						}).then(() => {
+							return fetchExperimentSettings('loan_channel_flss_query_v1', client);
```
mcstover, line 178: Isn't this just doing the same thing you've done above?

eddieferrer: This is the only sequence I found that works. You can't mutate an unassigned experiment.

So the sequence is assign it, mutate it, then run the assign query again so that we can read this query from the cache where we need to which returns the mutated version.

dyersituations (approved):

mcstover (approved):

PR #4301, opened 2022-10-21
fix: don't add background as a question in faqs

eddieferrer:
Fixes issue where background colors is added as a question, see Kyle's conversation in Slack. Seems to **not** be an issue in: https://github.com/kiva/cms-page-server/blob/main/components/contentful/FrequentlyAskedQuestions.vue


mcstover (approved):
Thank you much!

dyersituations (approved):

PR #4307, opened 2022-10-24
feat: display achievement service badge on lending stats page

eddieferrer:
Screen Shot 2022-10-24 at 9 58 25 AM

I just added this new query to the created hook, I suppose the page could be refactored to add this in prefetch, but it will need a sequence of requests since the achievements api request depends on the userId. 

Users without badges should not be shown anything, for now.

### DIFF
--- a/src/graphql/query/userAchievementsProgress.graphql
+++ b/src/graphql/query/userAchievementsProgress.graphql
@@ -1,0 +1,11 @@
+
+query userAchievementProgress(
+	$userId: String!
+) {
+	userAchievementProgress(userId: $userId) {
+		achievementProgress {
+			achievementId
+			status
+		}
+	}
+}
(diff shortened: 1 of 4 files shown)


dyersituations (approved):

mcstover (approved):

src/graphql/query/userAchievementsProgress.graphql
```diff
@@ -0,0 +1,11 @@
+
+query userAchievementProgress(
+	$userId: String!
+) {
+	userAchievementProgress(userId: $userId) {
```
emuvente, line 5: This endpoint should be refactored on the backend to read the user ID from the JWT, like other endpoints do. Then the userId argument isn't required and this can be called in a regular prefetch like other endpoints.

PR #4313, opened 2022-10-25
Core 779 helpme choose grid

roger-in-kiva:
Static data for testing.


mcstover (changes requested):
@roger-in-kiva Please use MDI icons instead of adding new images to the repo...the design team should know we use the free version of this library and adjust their designs to use icons included there
`mdiRefresh` + `mdiChevronLeft` should do the trick

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseTrigger.vue
```diff
@@ -33,27 +33,41 @@ export default {
...
-			if (this.variant === 'amountLeft') {
-				str = 'Almost funded';
-			}
-			if (this.variant === 'personalized') {
-				str = 'Recommended for you';
-			}
-			if (this.variant === 'researchScore') {
-				str = 'Research backed impact';
+			switch (this.variant) {
+				case 'amountLeft':
```
dyersituations, line 37: Doesn't need to change for this repo, but we'll want to convert these magic strings to a type when we move this code eventually to the `cms-page-server`.

roger-in-kiva:
> @roger-in-kiva Please use MDI icons instead of adding new images to the repo...the design team should know we use the free version of this library and adjust their designs to use icons included there `mdiRefresh` + `mdiChevronLeft` should do the trick

Ill keep this in mind, thank you @mcstover

mcstover (approved):

PR #4314, opened 2022-10-25
fix: change boolean flags to strings

christian14b:
Optimizely ‘Web’ product only supports `string` matching for custom attributes.


dyersituations (approved):

mcstover (approved):
interesting...

PR #4316, opened 2022-10-25
fix: remove cookie based exp redirect to lend-classic

mcstover:
This cookie will no longer be set. Suppose we can leave this code in for a while longer if we want though...

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -372,7 +372,7 @@
 											</li>
 											<li>
 												<router-link
-													:to="`/lend/${loanId}#loanComments`"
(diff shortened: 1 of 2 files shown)


dyersituations (approved):

emuvente (commented):
We could keep this as a convenience to the users that prefer to view the old version of the page. I didn't see this experiment removed in your other pr

mcstover:
Part of https://github.com/kiva/kiva/pull/11010/files was to remove the `httpdocs/classes/API/Experiment/LoanDetailBeta.php` exp which previously was setting this cookie.

mcstover:
We're going to keep this in place. The other part about My conversations has been moved to a different pr.

PR #4317, opened 2022-10-25
Vue 1255 partner id search

dyersituations:
https://kiva.atlassian.net/browse/VUE-1255

- Implemented Partner Info (ID) search/select
- The select box UI is mostly similar to legacy (see screenshots below)
- Extended new `KvSelectBox`
- Added new analytics events - including event to previously created keyword search (event was missed in changeset)
- Updated some mapping entries to use new `popularityScore` sort where appropriate

image

image

image


mcstover (approved):
Nice work and tests!

With the query map updates some categories could get activated to use flss. Pretty sure this flss facet is in prod already though...

dyersituations:
> Nice work and tests!
> 
> With the query map updates some categories could get activated to use flss. Pretty sure this flss facet is in prod already though...

Is there more that needs to be done to activate? I tested the updated map entries locally, and all pulled from FLSS when I changed pages.

mcstover:
@dyersituations I don't think there's any other work. Just wanted to make sure those new query maps would work in prod (ie partnerId is available there) if they were used.

dyersituations:
> @dyersituations I don't think there's any other work. Just wanted to make sure those new query maps would work in prod (ie partnerId is available there) if they were used.

@mcstover Sounds good, and yup, `partnerId` filter is live in prod.

PR #4321, opened 2022-10-26
fix: promo code only comes up for non-dynamically rendered pages

michelleinez:


mcstover (approved):
Makes sense.

Was the kv-loading-overlay you're removing a new instance or was it the primary overlay for when the page is initializing?

PR #4322, opened 2022-10-26
feat: update placement of user attributes

christian14b:
Polls based on user attributes are not triggering, but when sending a user attribute manually in the console, the poll appears immediately. We believe this is happening because user attributes were sent on the `created` hook when the DOM is not yet loaded.

If the behavior remains the same on `dev` the change will be reverted.

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -653,12 +653,14 @@ export default {
 				logReadQueryError(e, 'User Data For Optimizely Metrics');
 			}
 		}
+	},
+	mounted() {
 		// MARS-246 Hotjar user attributes
 		setHotJarUserAttributes({
 			userId: this.userId,
 			hasEverLoggedIn: this.hasEverLoggedIn,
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -653,12 +653,14 @@ export default {
 				logReadQueryError(e, 'User Data For Optimizely Metrics');
 			}
 		}
+	},
+	mounted() {
```
dyersituations, line 657: Just out of curiosity, why was this move needed?

dyersituations (approved):

src/components/WwwFrame/TheHeader.vue
emuvente: We're having to take some stabs in the dark to try to resolve issues with getting the hotjar surveys to display. With this we're testing if ensuring the attributes are only set in the browser and that they're called on every page load will make a difference.

emuvente (approved):

PR #4323, opened 2022-10-26
Core 784 helpme choose logic

roger-in-kiva:


src/components/LoansByCategory/HelpmeChoose/HelpmeChooseRecommendations.vue
```diff
@@ -61,16 +43,31 @@ export default {
...
+	},
 	data() {
 		return {
 			mdiRefresh,
-			selectedLoan: 0
+			selectedLoan: 0,
 		};
 	},
+	computed: {
+		loanData() {
```
dyersituations, line 59: Looks like this isn't being used currently.

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseRecommendations.vue
```diff
@@ -1,37 +1,19 @@
...
-
-			<!-- eslint-disable max-len -->
+		<div class="tw-flex tw-flex-col tw-gap-1 lg:tw-gap-1.5 lg:tw-w-1/3">
 			<helpme-choose-borrower-selector
-				image-url="https://res.cloudinary.com/kiva/w_960,h_600,c_fill,g_faces/a_ignore,fl_progressive,q_auto:best,f_auto/remote/ea3be8ae591d38de82c7a81b3657b22e.jpg"
-				name="BV Goudiry Group"
-				tag="Agriculture"
+				v-for="(loan, index) in loans"
+				:key="loan.id"
+				:image-url="loan.image.default"
```
dyersituations, line 7: Are we guaranteed that `image` and `activity` are defined for a loan? If not, we'd want computed values instead with optional chaining.

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseRecommendations.vue
roger-in-kiva: Yeah, this data will be injected to a loan card in other ticket.

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseRecommendations.vue
roger-in-kiva: Are we having a placeholder for default images?

dyersituations: Not sure about placeholder. I've personally never seen a loan without an image. @mcstover Is this a case that we'd need to cover (no loan image), or can we assume there will be an image?

roger-in-kiva: Some components are assuming there will be an image

```
<loan-card-image
							:loan-id="loan.id"
							:name="loan.name"
							:retina-image-url="loan.image.retina"
							:standard-image-url="loan.image.default"
							:is-visitor="isVisitor"
							:is-favorite="isFavorite"
							:class="{ 'tw-rounded-tl': roundedCorners }"

							@track-loan-card-interaction="trackInteraction"
							@favorite-toggled="toggleFavorite"
						/>
```

mcstover: You could v-if a placeholder in place of the image like we do in KivaClassicBasicLoanCard

roger-in-kiva: @dyersituations some methods to have optional chaining were added

@mcstover I like the idea of having a placeholder like that but the question is if a loan can have no image?

mcstover: Loans always have a image hash assigned to them so long ad you ask for it in your query it can be assumed to resolve to an image.

src/components/LoansByCategory/HelpmeChoose/HelpmeChooseRecommendations.vue
```diff
@@ -61,16 +43,44 @@ export default {
...
 			mdiRefresh,
-			selectedLoan: 0
+			selectedLoan: 0,
 		};
 	},
+	computed: {
+		loanData() {
+			return this.loans[this.selectedLoan];
+		},
+		activityName() {
```
dyersituations, line 62: Looks like these two computed can be removed, since you're using the methods.

mcstover: agree

roger-in-kiva: Sure, thanks!

dyersituations (approved):

mcstover (approved):

PR #4325, opened 2022-10-27
feat: add donate modules experiment

eddieferrer:
Adds 2 variations to the donation item in checkout. 

I decided to do 3 big `<template>` blocks one for each variant, I think it keeps the code a little more organized, and there are some subtle dom differences between each variant that were annoying to deal with otherwise. 

Also did some minor cleanup to this file:
* removed use of lodash forEach
* Combined contentful query into page query for prefetch

Variants:
Variant a - control, no change (driven by contentful)
Screen Shot 2022-10-27 at 10 47 31 AM

Variant b 
Screen Shot 2022-10-27 at 10 42 28 AM

Variant c
Screen Shot 2022-10-27 at 10 42 14 AM


src/components/Checkout/DonationItem.vue
```diff
@@ -1,170 +1,498 @@
...
+				<span class="tw-text-secondary tw-text-small tw-pr-1">Donate to Kiva</span>
+				<hr class="tw-flex-grow">
+			</div>
+			<h3>Help our team find more borrowers</h3>
+			<div class="tw-flex tw-flex-col md:tw-flex-row tw-pb-5 tw-mt-2">
+				<!-- donation image -->
+				<div class="tw-hidden md:tw-block tw-flex-none md:tw-mr-3 lg:tw-mr-4.5">
+					<img
+						class="donation-img tw-w-12 lg:tw-w-13 tw-h-12 lg:tw-h-13 tw-rounded"
+						src="@/assets/images/team_donate.png"
```
dyersituations, line 356: Did we want to do an image require here as well?

src/components/Checkout/DonationItem.vue
```diff
@@ -278,24 +603,33 @@ export default {
...
-						// Get contentful dynamic content
-						client.query({ query: donationItemQuery })
-					]).then(resolve).catch(reject);
-				}).catch(reject);
+			return client.query({
+				query: donationDataQuery
+			}).then(() => {
+				return client.query({ query: experimentQuery, variables: { id: 'basket_donate_modules' } });
+			}).catch(errorResponse => {
+				console.error(errorResponse);
```
dyersituations, line 616: Should this be `logFormatter(errorResponse, 'error');`?

dyersituations (approved):

src/components/Checkout/DonationItem.vue
eddieferrer: > Did we want to do an image require here as well?

Is there any benefit to using the image require?

dyersituations: It might help with deferring loading, but it was a difference I noticed between the sections.

mcstover (approved):

PR #4326, opened 2022-10-27
chore: removed inactive experiments

eddieferrer:
This removes referenced to experiments:
`uiexp.EXP-VUE-937-recommended-row-algo`
`uiexp.featured_loans`
`uiexp.funded_lyml_sort`

All 3 are disabled in production


src/graphql/query/fundedBorrowerProfile.graphql
```diff
@@ -7,10 +7,6 @@ query fundedBorrowerProfile(
 	$imgRetinaSize: String = "w960h720"
 ) {
 	general {
-		customSort: uiExperimentSetting(key: "funded_lyml_sort") {
```
eddieferrer, line 10: not referenced anywhere else

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -186,10 +186,6 @@ const pageQuery = gql`
 				key
 				value
 			}
-			customSort: uiExperimentSetting(key: "funded_lyml_sort") {
```
eddieferrer, line 189: not referenced anywhere else

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -572,21 +572,6 @@ export default {
 				);
 			}
 		},
-		initializeRecommendedRowAlgoExp() {
```
eddieferrer, line 575: this function is not referenced anywhere else

src/pages/Lend/admin/FeaturedAdminControls.vue
```diff
@@ -1,192 +0,0 @@
-<template>
```
eddieferrer, line 1: I don't know what this file is, but it is not referenced anywhere and hasn't been touched since 2018. It is the only place that'@/graphql/query/featuredLoansControl.graphql' and '@/graphql/mutation/setFeaturedLoans.graphql' are referenced, so I deleted those files too

emuvente (approved):

src/graphql/query/fundedBorrowerProfile.graphql
mcstover: I think I removed usage of this last week with the Funded BP cleanup...

src/pages/Lend/LendByCategoryPage.vue
mcstover: We actually had a ticket for this but I was unclear about  it's usage... https://kiva.atlassian.net/browse/VUE-972

Seems like any code that used this must have already been removed???

src/pages/Lend/admin/FeaturedAdminControls.vue
mcstover: This was all part of an embedded experiment/setting managment tool. I'd left the files in place b/c it was a really nice interface for the purpose, but we can revisit later if we need to set something like that up again.

mcstover (approved):

PR #4327, opened 2022-10-27
fix: use anchor links to avoid app routing for monolith lend/ routes

emuvente:


mcstover (approved):
Well it came to that ;)

PR #4330, opened 2022-10-27
fix: use category url to activate view more link and card on lend by …

mcstover:
…category

### DIFF
--- a/src/graphql/query/lendByCategory/categoryServiceLoanChannels.graphql
+++ b/src/graphql/query/lendByCategory/categoryServiceLoanChannels.graphql
@@ -16,6 +16,7 @@ query loanChannelData(
 			loanChannelId
 			name
 			description
+			url
 			savedSearch(limit: $numberOfLoans, excludeLoanIds: $excludeIds, origin: $origin) {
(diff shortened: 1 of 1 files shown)


src/graphql/query/lendByCategory/categoryServiceLoanChannels.graphql
```diff
@@ -16,6 +16,7 @@ query loanChannelData(
 			loanChannelId
 			name
 			description
+			url
```
dyersituations, line 19: Where does this get used?

dyersituations (approved):

src/graphql/query/lendByCategory/categoryServiceLoanChannels.graphql
mcstover: https://github.com/kiva/ui/blob/main/src/pages/Lend/LendByCategoryPage.vue#L250

The category object is based on this query which is then passed into the category row component. If the url is present the Row's header becomes a link and an additional "view more ##category name## loans" card is added at the end of the carousel.

PR #4332, opened 2022-10-27
fix: issues with new range min max slider (FF and colors)

dyersituations:
- Handles now work like expected in FF (both can be interacted with)
- Slider green now within handles in FF (it wasn't before)
- Slider colors now use new Kiva branding (and global CSS vars - missed that in my last PR)

image


mcstover (approved):
yay for brand color vars and pointer events!

PR #4343, opened 2022-11-02
fix: remove unused dependencies, fix unit test

eddieferrer:
Removed some dependencies which I believe are not being used. Ran tests, build, and local dev server to verify and everything ran fine.


package.json
```diff
@@ -125,7 +125,6 @@
 		"vue-server-renderer": "^2.6.14",
 		"vue-smooth-reflow": "^0.1.12",
 		"vue2-touch-events": "^2.3.2",
-		"vuedraggable": "^2.23.0",
```
eddieferrer, line 128: We're not using this dependency

package.json
```diff
@@ -154,10 +153,8 @@
 		"@testing-library/vue": "^5.8.2",
 		"app-manifest-loader": "^2.4.1",
 		"autoprefixer": "^10.4.2",
-		"babel-core": "^7.0.0-bridge.0",
```
eddieferrer, line 157: I believe with this jest update, and the removal of inject-loader, we can remove the babel-core dependency bridge.

package.json
```diff
@@ -179,7 +176,6 @@
 		"hard-source-webpack-plugin-fixed-hashbug": "^0.13.2",
 		"html-minifier": "^4.0.0",
 		"husky": "^5.2.0",
-		"inject-loader": "^4.0.1",
```
eddieferrer, line 182: I don't believe we are using this.

package.json
```diff
@@ -154,10 +153,8 @@
 		"@testing-library/vue": "^5.8.2",
 		"app-manifest-loader": "^2.4.1",
 		"autoprefixer": "^10.4.2",
-		"babel-core": "^7.0.0-bridge.0",
-		"babel-jest": "^23.4.2",
+		"babel-jest": "^27.5.1",
 		"babel-loader": "^8.1.0",
-		"babel-preset-vue": "^2.0.2",
```
eddieferrer, line 160: I don't think we are using this either

test/unit/specs/api/localResolvers/loan.spec.js
```diff
@@ -237,11 +237,11 @@ describe('loan.js', () => {
 				.toBe(resolvers.LoanPartner.fundraisingTimeLeftMilliseconds);
 		});
 
-		it('Returns the time remaining before expiration in words', () => {
+		it('Returns the time remaining before expiration in milliseconds', () => {
 			const now = new Date();
 			const tenMinutes = add(now, { minutes: 10 });
 			const twoHours = add(now, { hours: 2 });
-			const fiveDays = add(now, { days: 5 });
```
eddieferrer, line 244: This test actually fails currently because of DST (Nov 6) -- switching it to add hours instead of days seems to get around this.

dyersituations: Nice, thanks for the fix, I just ran into this one as well!

dyersituations (approved):

mcstover (approved):
Thanks, looks good

PR #4348, opened 2022-11-03
feat: added EEPR to Homepage Footer within Ui Server [VUE-1353]

tiarascoleman:
Screen Shot 2022-11-03 at 12 11 03 PM

### DIFF
--- a/src/components/WwwFrame/TheFooter.vue
+++ b/src/components/WwwFrame/TheFooter.vue
@@ -147,7 +147,7 @@
 							<li>
 								<a
 									:href="employeeEngagementUrl"
-									v-kv-track-event="['Footer', 'click-Work with us-Employee engagement']"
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

src/components/WwwFrame/TheFooter.vue
```diff
@@ -144,6 +144,15 @@
 									Partner with us
 								</router-link>
 							</li>
+							<li>
+								<a
+									:href="employeeEngagementUrl"
+									v-kv-track-event="['Footer', 'click-Work with us-Employee engagement']"
```
mcstover, line 150: This is under the Get to know us section so the 2nd param here should be:
`click-Get to know us-Employee engagement`.

Please update in both locations.

tiarascoleman: Revised and thanks for the find! :)

mcstover (approved):

PR #4349, opened 2022-11-03
Abstract loan filter

dyersituations:
Given that we were adding 3 new filters this sprint to the FLSS-driven filter page, it was a good opportunity to modularize our filters, to ensure the modularization is solid across different filters. Adding filters has generally been easy with the previous pattern, but you had to remember where to add all of the filter-specific changes, especially for tests, which meant pulling up an old PR as example. Having to rely on an old PR is not a good practice and means the code isn't clear enough (and yes, I wrote the original code I'm referring to here :P). This pattern encapsulates all configurations and logic into files for each specific filter, with tests to go along with them. Adding a new filter is as easy as creating a new file w/ tests, using an existing filter file as an example, no need to reference old PRs! This effort added ~1 day to this sprint, but in the end it seemed worth it based on the greater ease of extending and maintaining the filters. Also, it surfaced some bugs, one of which was resolved in this PR, look for the screenshot below. The new filter modules still need some unit test coverage, but the same existing unit test coverage is ported to these new modules. Using these new modules extracted a bunch of code out of existing components, some of which weren't previously tested.

- `LoanSearchFilter` now adds filters to the UI in order of properties of `filterConfig.config` and doesn't need to be changed in the future for new filters, unless a new filter type is added, or props of existing types are changed
- Files/modules created for each filter w/ associated tests
- Query parsing extracted to new util to eliminate circular import that occurred in new module structure

_Bug fixed: the wrong theme was being saved when saving a search_
image


mcstover (approved):
I like the programatic filter creation via config!

PR #4351, opened 2022-11-03
Add range slider functionality

dyersituations:
The following changes were needed for the new filters w/ range sliders. General behavior matches legacy.

- Handles now have tooltips that show values that track with the handles as they move
- A unit can be displayed in the tooltip
- A range slider can be a percentage, which assumes the number needs to be multiplied by 100 when displayed
- The number of decimals shown is based on the step provided
- Adjusted the z index on a couple other components to ensure everything layers correctly

Tooltips:
image

Unit:
image

Percentage w/ unit:
image


mcstover (approved):
cool!

PR #4354, opened 2022-11-04
feat: new partner filters

dyersituations:
https://kiva.atlassian.net/browse/VUE-1326
https://kiva.atlassian.net/browse/VUE-1372
https://kiva.atlassian.net/browse/VUE-1327

- Added parter risk rating, partner default rate, and partner profitability range slider filters
- Range slider added as option to `LoanSearchFilter` for rendering range slider filters
- Added `activities` filter (not displayed) so that the loan channels with `activity` work - there was also a mapping missing

image

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.371.1",
+	"version": "2.375.0",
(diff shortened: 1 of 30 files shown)


mcstover (approved):
Woohoo!

PR #4360, opened 2022-11-07
chore: deprecate more contentful components

eddieferrer:
typography.scss had already been removed, and I couldn't find any references to it. 

This PR also removes the use of the 3 font classes in typography.scss, most of the uses were in the deprecated contentful components, this removes the last couple of uses

Also removes - CategoryGrid.vue, not used anywhere

All content remove/migrated in contentful


mcstover (approved):
Thanks!

dyersituations (approved):

PR #4362, opened 2022-11-07
Core 839 checkout button error

roger-in-kiva:


dyersituations (approved):
Minor request: please add some simple details as to what exactly is being fixed in the PR description when requesting a cross-team review. I can guess that this is the checkout failing while the data is loading, but the context would help with the review. No need to update this PR, but it would definitely help with future PRs, thanks!

emuvente (approved):

PR #4370, opened 2022-11-09
feat: add how and press-center routes

christian14b:
Add routes to `how-kiva-works` and `press` contentful pages.


dyersituations (approved):

mcstover (approved):
This is fine for now and important to solidify the new pages at the existing routes. We will however, likely move these to be served by cms-page-server in the near future.

emuvente (approved):

PR #4372, opened 2022-11-11
fix: use 'both' instead of null when saving a search

mcstover:
- Adds custom option for "All loans" state with `both` as the value instead of `null`.

This appears to be working in the VM, prob need to checkout the tests for this filter...

### DIFF
--- a/src/util/loanSearch/filters/distributionModels.js
+++ b/src/util/loanSearch/filters/distributionModels.js
@@ -4,6 +4,7 @@ import { getEnumNameFromQueryParam } from '@/util/loanSearch/queryParseUtils';
 // Distribution model enum keys
 export const FIELDPARTNER_KEY = 'FIELDPARTNER';
 export const DIRECT_KEY = 'DIRECT';
+export const BOTH = 'BOTH';
 
 /**
  * Maps the distribution model enum names to display names
@@ -11,6 +12,7 @@ export const DIRECT_KEY = 'DIRECT';
 export const distributionModelDisplayMap = {
 	[FIELDPARTNER_KEY]: 'Partner',
 	[DIRECT_KEY]: 'Direct',
+	[BOTH]: 'All loans',
 };
 
 /**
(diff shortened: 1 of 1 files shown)


src/util/loanSearch/filters/distributionModels.js
```diff
@@ -63,7 +66,7 @@ export default {
 	},
 	getRemovedFacet: () => ({ distributionModel: null }),
 	getSavedSearch: loanSearchState => ({
-		distributionModel: distributionModelEnumMap[loanSearchState?.distributionModel?.toUpperCase()]
+		distributionModel: distributionModelEnumMap[loanSearchState?.distributionModel?.toUpperCase() ?? 'BOTH']
```
dyersituations, line 69: Is this long needed with the value map change?

src/util/loanSearch/filters/distributionModels.js
```diff
@@ -4,21 +4,24 @@ import { getEnumNameFromQueryParam } from '@/util/loanSearch/queryParseUtils';
 // Distribution model enum keys
 export const FIELDPARTNER_KEY = 'FIELDPARTNER';
 export const DIRECT_KEY = 'DIRECT';
+export const BOTH = 'BOTH';
```
dyersituations, line 7: Interesting that this is needed, since I did test saving with this filter as null. Was there an error?

src/util/loanSearch/filters/distributionModels.js
mcstover: hmm prob not... This may have been my first change as I wanted to make sure a default was always applied.

src/util/loanSearch/filters/distributionModels.js
mcstover: In my testing saved search urls were coming back with field partner if null was used. Direct worked but "All" was getting converted to field partner, likely due to the backend handling. There's some monolith code updates to go with this, which checks for `both` as the value before forcing a default of field partner if null is present...

dyersituations: Ah gotcha. I had assumed that was the auto-application of partner that the data team mentioned in slack recently, but makes sense that this was actually from the `null`.

mcstover:
This was ultimately setting a saved search to pass `both` as the value when "All loans" was selected but is currently unnecessary.

PR #4375, opened 2022-11-11
fix: remove ineffective exp force versions, add cps version of suppor…

mcstover:
…t-refugees


mcstover:
FYI @emuvente @eddieferrer @dyersituations Removing the force urls as they didn't really work and ended up redirecting to the cps versions for some reason. Also, we couldn't focus in on them in the new explores which have no mechasim for query params.

emuvente (approved):

PR #4378, opened 2022-11-14
feat: conditional share card image for new year campaign

christian14b:
Due to Year in Review Share we want to conditionally set a generic image in share cards. The output is to hand over a generic share URL for fb, twitter and linkedin to marketing.

### DIFF
--- a/src/pages/Homepage/Homepage.vue
+++ b/src/pages/Homepage/Homepage.vue
@@ -8,6 +8,7 @@ import gql from 'graphql-tag';
 import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
 import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import { preFetchAll } from '@/util/apolloPreFetch';
+import numeral from 'numeral';
 
 const ContentfulPage = () => import('@/pages/ContentfulPage');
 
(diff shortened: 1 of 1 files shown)


emuvente (commented):
We don't have final copy yet for the share card language, but I would like to see it included in this PR so that all we have to do later is replace the copy. It's important for this one because the copy will include dynamic data from URL parameters. Something like:
```js
computed: {
    nyShareCopy() {
        const loans = numeral(this.$route?.query?.ny-l);
        const loanString = `${loans.format('0,0')} ${loans.value() === 1 ? 'loan' : 'loans'}`;
        const borrowers = numeral(this.$route?.query?.ny-b);
        const borrowerString = `${borrowers.format('0,0')} ${borrowers.value() === 1 ? 'borrower' : 'borrowers'}`;
        const countries = numeral(this.$route?.query?.ny-c);
        const countryString = `${countries.format('0,0')} ${countries.value() === 1 ? 'country' : 'countries'}`;
        return `In 2022 I lent to ${loanString}, and helped ${borrowerString} in ${countryString} succeed.`;
    }
}
```

emuvente (approved):
👍 Looks good!

emuvente (commented):
Actually, I just remembered an issue we ran into with previous share cards. We need to be sure that we're including the sharing parameters in the canonical url when they are present. Like we did here https://github.com/kiva/ui/blob/main/src/pages/BorrowerProfile/BorrowerProfile.vue#L375-L381

emuvente (approved):
👍 Thanks for the update. It looks great!

PR #4379, opened 2022-11-15
feat: use lend/filter for saved searches, augment url where needed

mcstover:
This PR updates the saved search page `/lend/saved-search` to link to `/lend/filter` instead of legacy lend.
- updates the url path
- removes unnecessary loanLimit param when unset (ie, set to -1)
- removed duplicated separators after query param changes

### DIFF
--- a/src/components/Settings/SaveSearchItem.vue
+++ b/src/components/Settings/SaveSearchItem.vue
@@ -21,7 +21,7 @@
 				class="tw-border-tertiary tw-mb-2"
 			>
 			<kv-button
-				:href="savedSearch.url"
+				:href="savedSearchUrl"
 				class="tw-mr-2 tw-mb-2 md:tw-mb-0"
 			>
 				{{ viewLoansText }}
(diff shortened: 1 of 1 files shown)


mcstover:
Need to do a bit more validation and testing, potentially accomodating other filter formatting

src/components/Settings/SaveSearchItem.vue
```diff
@@ -65,6 +65,20 @@ export default {
...
+			// augment url for lend/filter compatibility
+			let ssUrl = this.savedSearch?.url ?? '';
+			// update url with lend/filter path
+			ssUrl = ssUrl.replace('lend', 'lend/filter');
+			// remove loanLimit of -1 which is equivalent to not being set
+			if (ssUrl.indexOf('loanLimit=-1') > -1) {
+				ssUrl = ssUrl.replace(/loanLimit=-1/, '');
+			}
+			// clear up any duplicate separators
+			ssUrl = ssUrl.replace(/\?&/, '?').replace(/&&/, '&');
```
dyersituations, line 78: Just out of curiosity, what would this look like with duplicate separators?

dyersituations (approved):

src/components/Settings/SaveSearchItem.vue
mcstover: `loanLimit=-1` could be in different locations. Sometimes just after the `?`, other times sandwhiched between ampersands so you couldend up with separators alongside each other once the query param and it's value are removed...

mcstover:
@dyersituations if any other query params come to mind that need a query param map to be compatible with lend/filter please let me know and we can fold it in here before merging.

dyersituations:
> @dyersituations if any other query params come to mind that need a query param map to be compatible with lend/filter please let me know and we can fold it in here before merging.

The only one at the moment I think is the distribution model ticket, but it's not related 100% to saved search.

PR #4383, opened 2022-11-15
'Stepper' on Focused Share Ask on Thanks page

christian14b:
Winning variant tested on Optimizely. This work contains:

- Remove confetti
- Remove confirmation email banner
- Add Stepper on Thanks Page

### DIFF
--- a/src/components/Thanks/ShareStepper.vue
+++ b/src/components/Thanks/ShareStepper.vue
@@ -2,35 +2,14 @@
 	<div class="tw-my-2 tw-px-2">
 		<div style="width: 136px" class="tw-mx-auto tw-px-1">
 			<div class="tw-flex tw-items-center">
-				<div class="tw-flex tw-items-center tw-text-brand tw-relative">
-					<div
-						class="tw-bg-brand tw-h-3.5 tw-rounded-full tw-w-3.5"
-					>
-						<kv-icon
-							class="tw-fill-current tw-inline-flex tw-max-h-full tw-ml-1 tw-mt-1 tw-text-white tw-w-1.5"
-							name="checkmark"
-						/>
-					</div>
-					<p
-						class="tw-text-small tw-absolute tw-font-medium tw-mt-7 tw-text-center tw-uppercase tw--ml-0.5"
-					>
-						Lend
-					</p>
-				</div>
-				<div
-					class="tw-flex-auto tw-border-t-2 tw-border-brand"
-				></div>
-				<div class="tw-flex tw-items-center tw-text-brand tw-relative">
-					<div
-						class="tw-border-2 tw-border-brand tw-h-3.5 tw-rounded-full tw-w-3.5"
-					>
-					</div>
-					<p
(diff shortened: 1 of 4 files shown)


src/components/Thanks/ShareStepperStep.vue
```diff
@@ -0,0 +1,52 @@
...
+
+export default {
+	name: 'ShareStepperStep',
+	components: { KvIcon },
+	props: {
+		step: {
+			type: Number,
+			default: 0
+		},
+		lastStep: {
```
dyersituations, line 38: Doesn't need to change necessarily, but I think this would be clearer as `isLastStep`, the comparison only needs to happen once in `ShareStepper`

src/components/Thanks/ShareStepperStep.vue
```diff
@@ -0,0 +1,52 @@
...
+			>
+				<kv-icon
+					v-if="step !== lastStep"
+					class="tw-fill-current tw-inline-flex tw-max-h-full tw-ml-1 tw-mt-1 tw-text-white tw-w-1.5"
+					name="checkmark"
+				/>
+			</div>
+			<p
+				:class="`tw-text-small tw-absolute tw-font-medium tw-mt-0.5 tw-text-center
+				tw-uppercase tw--ml-${0.5 * (step + 1)}`"
```
dyersituations, line 19: Are we guaranteed to have these dynamically created tailwind classes available? If not, this could be done with inline style instead.

src/components/Thanks/ThanksPageShare.vue
```diff
@@ -1,24 +1,6 @@
...
-						<div>Success, we've emailed your receipt to you.</div>
-					</div>
-					<div v-else>
-						<!-- eslint-disable-next-line max-len -->
-						<div>Success, your receipt has been sent to <strong class="fs-mask">{{ lender.email }}</strong></div>
-					</div>
-				</div>
-			</div>
-		</div>
+		<ShareStepper :lender-name="this.lender.firstName" />
```
mcstover, line 3: If this is replacing the `Success, your receipt has been sent to` string then the kiva-tests repo will also need to be updated as the presence of this string is used in those automated tests to signify we've reached the thanks page.

mcstover: @christian14b These test updates need to be scoped into this feature work to be merged at the same time.

mcstover (approved):

src/components/Thanks/ShareStepper.vue
```diff
@@ -0,0 +1,48 @@
...
+						:key="step.key"
+						:step="step.key"
+						:is-last-step="step.key === steps.length - 1"
+						:text="step.text"
+					/>
+				</template>
+			</div>
+		</div>
+		<p class="tw-text-center tw-mt-4 tw-text-h3">
+			{{ lenderName }}, complete your support by sharing this loan.
```
emuvente, line 16: The name should be masked from fullstory with `<span class="fs-mask">{{ lenderName }}</span>`

emuvente (approved):

PR #4386, opened 2022-11-17
Remove Quantcast, Algolia analytics, and duplicate Facebook pixel MARS-166

emuvente:
A similar PR will be created in the monolith repo.

### DIFF
--- a/src/head/script.js
+++ b/src/head/script.js
@@ -97,20 +97,6 @@ export default (config, globalOneTrustEvent) => {
 		/* eslint-enable */
 	};
 
-	// Facebook Pixel Code
-	const insertFB = () => {
-		/* eslint-disable */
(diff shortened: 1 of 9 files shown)


src/head/script.js
```diff
@@ -260,16 +228,9 @@ export default (config, globalOneTrustEvent) => {
...
-			if (config.algoliaConfig.enableAA && !optout) {
-				insertAlgoliaAnalytics();
-			}
 			if (config.enableHotjar && !optout) {
 				insertHotjar();
 			}
 		}
-
-		if (config.enableFB && !optout) {
-			insertFB();
```
mcstover, line 272: Are we going to be good with the specific fb events while relying on gtm to load the lib? I think they are guarded to check so prob so..

emuvente: Yeah, all the facebook events are guarded with the `fbLoaded` check or with a check to `window.fbq`

mcstover (approved):

PR #4387, opened 2022-11-17
fix: hide isMatchable filter

dyersituations:
https://kiva.atlassian.net/browse/VUE-1417

- `isMatchable` isn't being used anywhere else (not in the query map), so we can just remove it for now from the config


mcstover (approved):
Thank you. Hopefully a short lived removal!

PR #4388, opened 2022-11-17
fix: select box needs touch event

dyersituations:
https://kiva.atlassian.net/browse/VUE-1418

- Mobile select box couldn't be closed


mcstover (approved):
nice find and fix!

PR #4389, opened 2022-11-18
Core 854 finding page exp setup

roger-in-kiva:
setup for new loan finding page experiment added to lbc page
new page folder and file added to repo
new `lending-home` route added to router file


src/router/routes.js
```diff
@@ -290,6 +290,14 @@ module.exports = [
 			excludeFromStaticSitemap: true,
 		}
 	},
+	{
+		path: '/lending-home',
```
dyersituations, line 294: Looks like we need a PR to `puppet` for this new route.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -708,8 +731,15 @@ export default {
 					client.query({ query: experimentQuery, variables: { id: 'mfi_recommendations' } }),
 					// experiment: VUE- Category Service driven FLSS channels
 					client.query({ query: experimentQuery, variables: { id: 'flss_category_service' } }),
+					// experiment: CORE-854 Loan Finding Page
+					client.query({ query: experimentQuery, variables: { id: 'loan_finding_page' } }),
 				]);
 			})
+				.then(() => {
+					return client.query({
+						query: mlOrderedLoanChannels
```
dyersituations, line 740: Can this duplicate `mlOrderedLoanChannels` be removed?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -694,6 +696,27 @@ export default {
 				}
 			});
 		},
+		initializeLoanFindingPageExperiment() {
+			const layoutEXP = this.apollo.readFragment({
```
dyersituations, line 700: Recommend renaming `layoutEXP` to something more applicable to this exp.

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -1,5 +1,6 @@
 <template>
 	<www-page
+		v-show="showContent"
```
dyersituations, line 3: Can we instead do this redirect and exp check in the `preFetch`? See `LendFilterPage.vue` for an example of this:

```
preFetch(config, client, { cookieStore, route }) {
			// prefetch page data + experiment settings
			return client.query({
				query: lendFilterPageQuery,
				variables: {
					basketId: cookieStore.get('kvbskt')
				},
			}).then(() => {
				// check route for eligibility
				if (isFLSSEligible(route)) {
					return getExperimentSettingAsync(client, lendFilterRedirectEXP)
						.then(() => {
							// Assign the exp
							return client.query({
								query: experimentAssignmentQuery,
								variables: { id: lendFilterRedirectEXP }
							});
						}).then(({ data }) => {
							// Rediect if we are in the experiment group
							if (data?.experiment?.version === 'b') {
								return Promise.reject({
									path: '/lend/filter-alpha',
									query: route.query,
								});
							}
						});
				}
				// Otherwise just resolve
				Promise.resolve();
			});
		}
```

dyersituations (changes requested):
See comments.

src/pages/Lend/LendByCategoryPage.vue
mcstover: +1 dupe can be removed

src/router/routes.js
mcstover: It will be good to create this route in puppet too. But if we go the client only approach to the redirect it would still work b/c of router push navigating in page instead of refreshing...

src/pages/Lend/LendByCategoryPage.vue
mcstover: Roger and I have discussed trying to do a prefetch and while I do think it's possible, it's also ok as a first pass to do all of the redirecting "and tracking" together in the single init method below...It would be good to do a little more work and get this promise chain in place as shown above though.

That said @roger-in-kiva I do not think we should hide ALL of the content. We can show a loader over all of the in page content while your exp is bootstrapped but not hide everything.

src/router/routes.js
dyersituations: @mcstover Would you rather do the client-side redirect instead of the server-side that I mentioned in another comment?

src/pages/Lend/LendByCategoryPage.vue
mcstover: Extending the promise chain and redirecting on the server/page load is my preference...

I think we can use `my.userAccount.id` field a conditional on the experiment assignment and then if the assignment is present decide about the redirect...

```
	preFetch(config, client) {
			return client.query({
				query: lendByCategoryQuery
			}).then(({ data })=> {
				// Get the array of channel objects from settings
				return Promise.all([
					// experiment: GROW-330 Machine Learning Category row
					client.query({ query: experimentQuery, variables: { id: 'EXP-ML-Service-Bandit-LendByCategory' } }),
					// experiment: CORE-698 MFI Recommendations
					client.query({ query: experimentQuery, variables: { id: 'mfi_recommendations' } }),
					// experiment: VUE- Category Service driven FLSS channels
					client.query({ query: experimentQuery, variables: { id: 'flss_category_service' } }),
					// experiment: CORE-854 Loan Finding Page
					data?.my?.userAccount?.id ? client.query({ query: experimentQuery, variables: { id: 'loan_finding_page' } }) : Promise.resolve(),
				]);
			}).then(expAssignments => {
               // check for loan_finding_page exp Assignment
               const targetExp = expAssignments?.this?.that?.exp?.version...
               if (targetExp === 'b') {
				  return Promise.reject({
					path: '/lending-home',
				  });
			   }
               Promise.resolve();
            })
				.then(() => {
					return client.query({
						query: mlOrderedLoanChannels
```

dyersituations: In this case, will the puppet entry be required?

roger-in-kiva: This sounds good, but what about tracking events. Should they passed to the new page when created?

```
this.$kvTrackEvent(
       'Lending',
       'EXP-CORE-854-Dec2022',
	loanFindingEXP.version
);
```

dyersituations: There are probably a couple options, and you'll want to test which is working like expected. If traffic only hits the new page from this redirect, then tracking on the new page should work fine. But if we only want to track if the user gets redirect (as opposed to opening the new page directly), then you'll want to do something similar to what's done in the `mounted` of `LendFilterPage.vue`.

mcstover: Tracking events on exps are generic in the sens that that they always fire if the exp is assigned and the version matches the context.

As Casey mentioned, you'd have a block in mounted to get the exp assignment and then fire the tracking for the assignment.

For LBC, where the exp is originally assigned, you'd use `getExperimentSettingCached` + `trackExperimentVersion` but on the new test page you'd need to use `getExperimentSettingAsync` + `trackExperimentVersion` to ensure you've fetched the setting and assignment.

roger-in-kiva: updated! WDYT @dyersituations @mcstover ?

mcstover: @roger-in-kiva I think this is looking great. Please remove this v-show line though so we don't hide/flash the entire page of content. You shouldn't need the new `showContent` mechanism at all now that you've wired up the prefetch!

src/router/routes.js
mcstover: We will need to update .htaccess in the kiva/kiva repo to handle this route in the dev-vm and create a new ui route in puppet. I can help with the puppet portion as needed.

src/pages/Lend/LendByCategoryPage.vue
dyersituations: Looks good, and agreed regarding removing `showContent`

roger-in-kiva: removed, thanks guys!

dyersituations (approved):
Approved pending proxy changes in other repos.

roger-in-kiva:
New ticket added for proxy changes
https://kiva.atlassian.net/browse/CORE-901

PR #4402, opened 2022-11-23
fix: ui doesn't load loans when default loan display mode is set to "…

michelleinez:
…grid"

### DIFF
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -6,7 +6,7 @@
 			<!-- TODO: Add promo code entry input, if no promo query params exist and  no promo is applied -->
 			<campaign-status
 				v-if="!hideStatusBar"
-				class="corporate-campaign-landing__status tw-sticky tw-top-8 md:tw-top-9 tw-z-2"
+				class="corporate-campaign-landing__status tw-sticky tw-z-2"
 				:active-credit-type="activeCreditType"
 				:is-matching="isMatchingCampaign"
 				:loading-promotion="loadingPromotion"
@@ -743,6 +743,9 @@ export default {
 		}
 	},
 	computed: {
+		loanDisplayComponent() {
+			return this.$refs?.mlLoanDisplay?.$refs || this.$refs?.mlLoanDisplay?.[0]?.$refs;
+		},
 		campaignLoanWrapperProps() {
 			return {
 				filters: this.filters,
@@ -1122,10 +1125,6 @@ export default {
 			// TDOO: Consider extending loading state for basket updates
 			// Query to update basket state
 			this.updateBasketState();
-			// TEMPORARY: Obstruct ability to click the "Checkout" button on the loan card to prevent redirect
-			if (this.$refs.loandisplayref) {
-				this.$refs.loandisplayref.loadingLoans = true;
-			}
 		},
 		updateBasketState() {
 			// Ensure basket state is loading
@@ -1220,11 +1219,6 @@ export default {
 
 					// Update user Auth state
 					this.setAuthStatus(this.kvAuth0?.user ?? {});
-
-					// TEMPORARY: turn off loading loans
-					if (this.$refs.loandisplayref) {
-						this.$refs.loandisplayref.loadingLoans = false;
-					}
 					if (this.$refs.inContextCheckoutRef) {
 						this.$refs.inContextCheckoutRef.updatingTotals = false;
 					}
@@ -1322,10 +1316,6 @@ export default {
 			this.$nextTick(() => {
 				// re-enable visibility of loans, activates loan fetch within loan display
 				this.showLoans = true;
-				// up loan query for grid view
-				if (!this.showLoanRows && this.$refs.loandisplayref && this.$route.query.page) {
-					this.$refs.loandisplayref.updateFromParams(this.$route.query);
-				}
 			});
 		},
 
@@ -1411,8 +1401,7 @@ export default {
 			}
 		},
 		jumpToLoans() {
-			const loanDisplayComponent = this.$refs?.mlLoanDisplay?.$refs || this.$refs?.mlLoanDisplay?.[0]?.$refs;
-			loanDisplayComponent.campaignLoanSection.scrollIntoView({ behavior: 'smooth' });
+			this.loanDisplayComponent.campaignLoanSection.scrollIntoView({ behavior: 'smooth' });
 		},
 		adjustRouteHash(hash) {
 			const route = { ...this.$route };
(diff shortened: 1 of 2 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1322,10 +1316,6 @@ export default {
 			this.$nextTick(() => {
 				// re-enable visibility of loans, activates loan fetch within loan display
 				this.showLoans = true;
-				// up loan query for grid view
-				if (!this.showLoanRows && this.$refs.loandisplayref && this.$route.query.page) {
-					this.$refs.loandisplayref.updateFromParams(this.$route.query);
```
mcstover, line 1327: This might break paging in grid view...

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1122,10 +1125,6 @@ export default {
 			// TDOO: Consider extending loading state for basket updates
 			// Query to update basket state
 			this.updateBasketState();
-			// TEMPORARY: Obstruct ability to click the "Checkout" button on the loan card to prevent redirect
-			if (this.$refs.loandisplayref) {
-				this.$refs.loandisplayref.loadingLoans = true;
-			}
```
mcstover, line 1128: This block should be initializing a loading overlay after/during add to basket. This is probably good to preserve...

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1445,18 +1434,7 @@ export default {
 			this.loanDetailsVisible = true;
 		}
 	},
-	beforeRouteEnter(to, from, next) {
-		next(vm => {
-			if (vm.$refs.loandisplayref) {
-				vm.$refs.loandisplayref?.updateFromParams(to.query);
```
mcstover, line 1451: This was also meant to power paging in grid view. specifically the ability to hit page=2 and load those loans...

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: In what way?  I tried switching pages in grid view and to me it appears to work just fine...

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: yeah this appears to work just fine...  did you notice an issue with it?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: The button in the new KivaClassicBasicLoanCard changes the "Loan $25 now" button text to "Loading..."  Here take a look at this loom video to see the UI behavior now.  https://www.loom.com/share/5d217b088b04404da27ec588c9876f16

michelleinez: Is the delay still too noticeable with the button text changing the way that it does?  I can't remember if the old loan cards did this as well or if you're just talking about the very short like <1 second delay between when the button changes to "Checkout now" and when the lightbox appears.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
mcstover: If you refresh the page, the grid view doesn't load and the `x` page of loans aren't loaded...

mcstover: I do see that paging generally works though in grid view!

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
mcstover: Yep, the new Kiva Classic card def prevents any accidental clicking out to checkout.

The "percieved" delay of the page was an issue that got some significant number of complaints from users through SP which is why we added in the different loading states to indicate when some background process was in progress.

You could probably chat with the SP stakeholders and walk through the interface in prod to get an idea of how it's behaving and if they are ok with the current behavior...

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
mcstover: reload the page and it doesn't work for me

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: I'm having trouble reproducing this issue...  Here's a video of me testing and it appearing to work.  Maybe you could let me know if there's a discrepancy between what I did and what you did?
https://www.loom.com/share/eeb16160ab5b430c8617a74cb1fc1108

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -6,7 +6,7 @@
 			<!-- TODO: Add promo code entry input, if no promo query params exist and  no promo is applied -->
 			<campaign-status
 				v-if="!hideStatusBar"
-				class="corporate-campaign-landing__status tw-sticky tw-top-8 md:tw-top-9 tw-z-2"
+				class="corporate-campaign-landing__status tw-sticky tw-z-2"
```
mcstover, line 9: Does this make the status bar cover the header now? Or is it naturally falling below the header properly?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -743,6 +743,9 @@ export default {
 		}
 	},
 	computed: {
+		loanDisplayComponent() {
+			return this.$refs?.mlLoanDisplay?.$refs || this.$refs?.mlLoanDisplay?.[0]?.$refs;
+		},
```
mcstover, line 748: Note that computed properties looking at refs will only evaluate on the initial page setup. values nested in refs are not reactive. Should be fine for initially setting the grid view vs carousel view. Just need to make sure toggling works fine.

mcstover (approved):

dyersituations (approved):
Tested locally, and there don't appear to be any new issues. The page has some existing issues though, but I verified they exist in dev.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: image
It was actually looking like this when the user wasn't logged in (see image)
And when the user was logged in, the banner looked fine.

PR #4407, opened 2022-11-29
Fix apollo resolvers

dyersituations:
https://kiva.atlassian.net/browse/VUE-1451

- Turns out that Apollo local resolvers weren't being called
- While [Apollo does recommend type policies + local only fields](https://www.apollographql.com/docs/react/local-state/local-resolvers/), local resolvers still work as before
- The pages that encountered Apollo errors (borrower profile, lend/filter, etc.) seem to work fine now
- Before being able to push to git, the missing `id` fields had to be added to a number of local resolver queries/mutations

**Notes**
- I assumed there was a new requirement for these `id` fields with Apollo 3, and I didn't test these use cases
- The MARS-178 branch needs changes from `main` before it can be merged
- I created a branch off the feature branch to isolate these new changes for review

### DIFF
--- a/src/api/apollo.js
+++ b/src/api/apollo.js
@@ -1,6 +1,4 @@
-/* eslint-disable no-underscore-dangle */
 import { ApolloLink, ApolloClient, InMemoryCache } from '@apollo/client/core';
-
 import Auth0LinkCreator from './Auth0Link';
 import BasketLinkCreator from './BasketLink';
 import ContentfulPreviewLink from './ContentfulPreviewLink';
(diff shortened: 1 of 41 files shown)


emuvente (approved):
Thanks @dyersituations!! I think this will resolve the issue I was seeing in https://kiva.atlassian.net/browse/MARS-323.

> I assumed there was a new requirement for these `id` fields with Apollo 3, and I didn't test these use cases

What's changed in Apollo 3 is that we have to define how to combine field data in the cache for types that don't have an `id` field. Previously they were merged automatically, but now we have to opt-in to that behavior.

For the server types that we have no control over, I added a special `Mergable` parent type of all the server types without IDs, and defined the `merge: true` option for that type. For the client types that we have control over I added a constant `id` field to every type that didn't have one before.

Once a type has an `id` field, our linting rules require that the id is always requested in queries, to ensure any queries watching the same object will be updated correctly.

dyersituations:
@emuvente Makes sense, thanks for the info! I'll go ahead and merge this PR so that the feature branch has these changes.

PR #4409, opened 2022-11-30
feat: sort by dropdown updgraded in quick filters component

roger-in-kiva:
css rules overridden in quick filter component for sort-by dropdown to follow new design

Screen Shot 2022-11-29 at 19 58 33

### DIFF
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -91,8 +91,9 @@
 					Sort By
 				</label>
 				<kv-select
+					id="sortBy"
 					:disabled="!filtersLoaded"
-					v-model="sortBy" id="sortBy"
+					v-model="sortBy"
 					style="min-width: 180px;"
 					@click.native="trackDropdownClick('sort')"
 				>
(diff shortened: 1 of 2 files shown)


src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -249,3 +257,19 @@ export default {
 	}
 };
 </script>
+
+<style lang="postcss">
```
dyersituations, line 261: Instead of unscoped CSS, you can use `scoped` on the `style` tag and `>>>` below, for example:

`#customizedSortBySelector >>> select {`

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -114,26 +115,31 @@
...
-				:disabled="!filtersLoaded"
-				v-model="sortBy"
-				style="min-width: 180px;"
-				@click.native="trackDropdownClick('sort')"
-			>
-				<option
-					v-for="sortType in filterOptions.sorting"
-					:key="sortType.key"
-					:value="sortType.key"
+			<div id="customizedSortBySelector" class="tw-text-action tw-flex tw-items-center tw-gap-1 tw-text-small">
```
dyersituations, line 118: Do we want the entire component to have `tw-text-action` or just the sort type name? Currently the "Loan sort" text is green like it can be clicked, but it can't actually be clicked. Or do we want this entire `div` clickable?

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
roger-in-kiva: Nice suggestion! I'll update it.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
roger-in-kiva: This is from design; is intended to have the entire component with `tw-text-action` but the behavior of a select element.

dyersituations (approved):

PR #4410, opened 2022-11-30
test: optimizely only recognize in value in cents

christian14b:
Validate loan total value before sending optimizely revenue tracking event

### DIFF
--- a/src/util/checkoutUtils.js
+++ b/src/util/checkoutUtils.js
@@ -85,6 +85,10 @@ export function formatTransactionData(
 		depositTotal: totals.creditAmountNeeded || 0,
 		paymentType,
 		isFTD: false,
+		optimizely: {
+			loanTotal,
+			donationTotal
+		}
 	};
 
 	return transactionData;
--- a/src/util/checkoutUtils.js
+++ b/src/util/checkoutUtils.js
@@ -85,10 +85,6 @@ export function formatTransactionData(
 		depositTotal: totals.creditAmountNeeded || 0,
 		paymentType,
 		isFTD: false,
-		optimizely: {
-			loanTotal,
-			donationTotal
-		}
 	};
 
 	return transactionData;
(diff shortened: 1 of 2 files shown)


mcstover (approved):

src/util/checkoutUtils.js
```diff
@@ -85,6 +85,10 @@ export function formatTransactionData(
 		depositTotal: totals.creditAmountNeeded || 0,
 		paymentType,
 		isFTD: false,
+		optimizely: {
+			loanTotal,
+			donationTotal
```
mcstover, line 90: Does this value match what you want to send? I think it can come back sometimes looking like `4.5` or just `25`... What type of values did you see?

roger-in-kiva:
Looks like they are not in cents.
What happens if having a kiva card then the deposit will be '0.00'? `if('0.00') == true` so you'll be tracking this deposit.
I think Melissa is reporting a 0.00 tracking event as an issue

Screen Shot 2022-11-30 at 11 53 17

christian14b:
Thanks @roger-in-kiva , I completely misunderstood the request!

roger-in-kiva:
I think should be considered here too: https://github.com/kiva/ui/blob/9caee359fc1c7cb7922afc83fb9c7051a0622997/src/plugins/kv-analytics-plugin.js#L295

roger-in-kiva:
@christian14b validations look good.
From optimizely documentation here: https://support.optimizely.com/hc/en-us/articles/4410289428109 , we can see that we need to add a new `revenue` tag in both cases like this

```
"tags"      : {
                     "revenue": valueInCents
                    }

```
with

```
var value = ‘youDefineThisValue’;
// Revenue must be added in cents
var valueInCents = value * 100;
```

ThevalueInCents variable represents the purchase amount, in cents (purchase amount*100). It is crucial that this value is passed to Optimizely in cents. Otherwise, your revenue totals will be incorrect.

Sorry for confusion.

roger-in-kiva (approved):
Looks good!

PR #4414, opened 2022-11-30
[loan finding MVP] feat: load filters for loan finding page quick filters

egoroza:


roger-in-kiva:
@egoroza  are categories sorted in alphabetical order?

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -19,42 +19,36 @@
...
-					{
-						title: 'Amount low to high',
-						key: 'amountLowToHigh'
-					},
-					{
-						title: 'Ending soon',
-						key: 'expiringSoon'
-					}
-				]
+				gender: [{
```
dyersituations, line 40: Will this be replaced by facets later?

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -67,6 +61,62 @@ export default {
...
+			this.quickFiltersOptions.categories = [
+				...[{ title: 'All categories', key: 0 }],
+				...categories.lend?.loanChannels?.values
+			];
+
+			this.quickFiltersOptions.location = facets.regions;
+
+			this.quickFiltersOptions.sorting = [
+				{
+					title: 'Recommended',
```
dyersituations, line 83: I see we already have this sort array in `LoanChannelCategoryControl`. If these 2 components are supposed to always have the same sort options, it would be worth for maintenance putting that array in a util and sharing between these two components.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -67,6 +61,62 @@ export default {
...
+				{
+					title: 'All genders',
+					key: '',
+				},
+				{
+					title: 'Women',
+					key: 'female',
+				},
+				{
+					title: 'Men',
```
dyersituations, line 113: Do we want the nonbinary option here as well? If so, the facets could be loaded from the API like in lend/filter.

src/util/flssUtils.js
```diff
@@ -64,6 +65,26 @@ export async function fetchFacets(
 	}
 }
 
+/**
+ * Fetches list of categories from FLSS
+ *
+ * @param {Object} apollo The Apollo client instant
+ * @returns {Promise<Array<Object>>} Promise for category name/id
+ */
+export async function fetchCategories(
```
dyersituations, line 74: This file is unit tested, so we'll need tests for this new method, following a similar pattern to the other methods in this file.

egoroza:
> @egoroza  are categories sorted in alphabetical order?

Nope - can add that if we want as a requirement!

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: Nah this gets hard coded for now

roger-in-kiva:
> > @egoroza  are categories sorted in alphabetical order?
> 
> Nope - can add that if we want as a requirement!

Yeah, is a requirement here https://kiva.atlassian.net/browse/CORE-859

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: We will have some tech debt work to refactor quick filters so it's more seamless to share - also this should pull from the API eventually

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: Same comment as regards to above - we will want to pull from an API after experiment phase

dyersituations: In that case, we probably want to add nonbinary here for consistency with other filters. And I recommend adding a TODO here for returning to pull the data from the API.

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: I recommend adding a TODO here then for returning to pull this data from the API.

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: Yeah we currently support nonbinary in filtering on prod: https://www.kiva.org/lend/filter. One benefit of using the API is that it will return supported genders, so that should be the target longterm. But at the moment, it is safe to add nonbinary here as well manually.

egoroza: I'll add it manually for now and add a TODO note for integrating the API when we have this as a feature 👍

dyersituations (approved):

roger-in-kiva (approved):
Looks good!

PR #4415, opened 2022-11-30
feat: improve loan meta preview

eddieferrer:
* makes MARS-143 share language the default for meta preview description and title
* cleanup funded borrower profile variables

ACK-484

### DIFF
--- a/src/components/BorrowerProfile/FundedBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FundedBorrowerProfile.vue
@@ -116,7 +116,6 @@ import personalizedLoansQuery from '@/graphql/query/lendByCategory/personalizedL
 import mlLoansYouMightLikeData from '@/graphql/query/loansYouMightLike/mlLoansYouMightLikeData.graphql';
 import LoanCardController from '@/components/LoanCards/LoanCardController';
 import { FLSS_ORIGIN_BP_FUNDED } from '@/util/flssUtils';
-import loanUseFilter from '@/plugins/loan-use-filter';
 import KvGrid from '~/@kiva/kv-components/vue/KvGrid';
 import KvPageContainer from '~/@kiva/kv-components/vue/KvPageContainer';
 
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/FundedBorrowerProfile.vue
```diff
@@ -190,34 +189,19 @@ export default {
...
-			anonymizationLevel: 'none',
-			shareCardLanguageVersion: '',
-			inviterIsGuestOrAnonymous: false,
-			basketedLoans: []
+			shareCardLanguageVersion: ''
 		};
 	},
 	computed: {
 		shareTitle() {
-			if (this.shareCardLanguageVersion === 'b') {
```
dyersituations, line 202: Can the `shareCardLanguageVersion` be removed from this component now? I see the exp itself is being used in the thanks page, but these appear to be the only usages in this component.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -596,25 +596,20 @@ export default {
 				this.loanUseMaxLength, this.anonymizationLevel);
 		},
 		shareTitle() {
-			if (this.shareCardLanguageVersion === 'b') {
```
dyersituations, line 599: Same comment about exp.

mcstover (approved):

src/components/BorrowerProfile/FundedBorrowerProfile.vue
eddieferrer: `this.shareCardLanguage` is still being used in this component to track the experiment. It can probably be removed? We might be able to remove it from this page and also from the thanks page, but check with the MARS team first

dyersituations: Ah gotcha, it's an exp from a different team, thanks for the info.

PR #4416, opened 2022-11-30
Vue 1457

dyersituations:
https://kiva.atlassian.net/browse/VUE-1457

image

### DIFF
--- a/src/components/BorrowerProfile/LoanBookmark.vue
+++ b/src/components/BorrowerProfile/LoanBookmark.vue
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<template v-if="isLoading">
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/LoanBookmark.vue
```diff
@@ -49,7 +57,6 @@ export default {
 				}
 			}
 		`,
-		preFetch: true,
 		preFetchVariables({ route }) {
```
mcstover, line 60: The `preFetchVariables` block should be clear to remove here too...

dyersituations: Ah yup, good catch. Removed and confirmed to be unneeded.

mcstover (approved):

PR #4417, opened 2022-11-30
Mars 317 kiva context borrower profile

christian14b:
This experiment has two versions for loans that belongs to the following categories (these are ranked too):

1. Eco-friendly
2. Education
3. Arts
4. Agriculture
5. Women
6. Kiva US

**A** version adds `loan impact` section and chunk text in `story`  and `more about this loan` sections. Also, adds a new `vetted by` section.

image
image

**B** version displays the expected repayment date 
image


mcstover (changes requested):
@christian14b We are in the process of trying to reduce the initial page load for the Borrower Profile so should not be adding any additional prefetched data. As noted in other discussions we are actively auditing and removing unnecessary data from the prefetch over the next couple sprints.

Please move any new data needed for your experiments to it's own query that runs in the mounted hook - or created if wrapped with a `!this.$isServer` so it only runs in the browser. You will need to create placeholders where the content will eventually show up to prevent significant content jumps for at least the Loan Story additions.

christian14b:
@mcstover Let me know if that change fix the problem or if there's something else to change regarding prefetched data

src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -96,13 +131,30 @@ export default {
...
+						country {
+							isoCode
+						}
+					}
+					tags
+					gender
+					sector {
+						id
+						name
+					}
```
mcstover, line 144: This is a prefetched query. Can these items be placed in a new query activated by the mounted hook instead?

src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -96,13 +131,30 @@ export default {
...
 							id
 							bylineName
 							showName
 						}
+						partner {
+							id
+							countries {
+								name
+							}
+						}
```
mcstover, line 157: Same here...

src/components/BorrowerProfile/LoanStory.vue
mcstover: Looks like you've already duplicated them in mounted so maybe these can just be removed...

christian14b: Thought I removed them, thanks!

mcstover (approved):
Thank you @christian14b This exp has some really neat informational addtions to the page!

dyersituations (approved):

emuvente (approved):
Looks good!

PR #4427, opened 2022-12-02
fix: errors resolved for apollo cache on BP

dyersituations:
Earlier today some changes were merged that introduced a few invariant violation errors that occur when Apollo tries to read something from cache that doesn't exist. Those errors were blocking/confusing my borrower profile optimization work, so I spun out these fixes, since we are also currently seeing these errors on dev.

@christian14b Two items to look into:

1. I didn't want to add the experiment query to the `preFetch`, but I noticed that the experiment flag was used quite a bit in the template. If we can move this to clientside, please do in a follow-up PR.
2. I didn't spend time verifying that delaying this data fetching indeed works okay from a UX perspective (like if text pop-in is bad), since I'm not familiar with this new experiment. Please test this when you have a chance, because this PR enables the data to be loaded clientside, and before the changes in this PR, the data was not loading clientside (it was erroring out). An example is the change I made for `this.sector?.name?.toLowerCase()`. That was erroring out because sector hadn't loaded yet, but I didn't verify if this computed value causes bad pop-in when the data fetch is moved clientside. 
 
If you have any questions, please let me know.

Thanks!

### DIFF
--- a/src/components/BorrowerProfile/LoanStory.vue
+++ b/src/components/BorrowerProfile/LoanStory.vue
@@ -167,7 +167,7 @@ export default {
 			this.previousLoanId = loan?.previousLoanId ?? 0;
 		},
 	},
-	mounted() {
+	async mounted() {
 		const query = gql`query loanStoryContextExp($loanId: Int!) {
 			lend {
 				loan(id: $loanId) {
@@ -196,7 +196,7 @@ export default {
 		}`;
 
 		try {
-			const data = this.apollo.query({
+			const { data } = await this.apollo.query({
 				query,
 				variables: {
 					loanId: this.loanId,
--- a/src/components/BorrowerProfile/LoanStory.vue
+++ b/src/components/BorrowerProfile/LoanStory.vue
@@ -196,7 +196,7 @@ export default {
 		}`;
 
 		try {
-			const data = this.apollo.readQuery({
+			const data = this.apollo.query({
 				query,
 				variables: {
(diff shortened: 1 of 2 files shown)


src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -196,7 +196,7 @@ export default {
 		}`;
 
 		try {
-			const data = this.apollo.readQuery({
+			const data = this.apollo.query({
```
emuvente, line 199: `query` is asynchronous. It returns a promise for the data, not the actual data.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -604,7 +608,7 @@ export default {
 		}`;
 
 		try {
-			const data = this.apollo.readQuery({
+			const data = this.apollo.query({
```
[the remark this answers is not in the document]
dyersituations: Done.

src/components/BorrowerProfile/LoanStory.vue
dyersituations: Fixed, and data return format confirmed.

dyersituations:
> Thanks @dyersituations ! I tested this out in my machine and it's working well. Regarding the pop-in text I can add a placeholder on another pr so it won't look weird.

@christian14b Sounds good, thanks for testing!

emuvente (approved):

PR #4429, opened 2022-12-02
feat: show loans on MVP Quick Filters

egoroza:
Screen Shot 2022-12-02 at 12 53 52 PM

### DIFF
--- a/src/components/LoanFinding/QuickFiltersSection.vue
+++ b/src/components/LoanFinding/QuickFiltersSection.vue
@@ -14,29 +14,52 @@
 			@reset-filters="resetFilters"
 			@handle-overlay="handleQuickFiltersOverlay"
 		/>
+		<div
+			class="tw-flex tw-flex-wrap"
+		>
+			<kiva-classic-basic-loan-card
+				v-for="(loan, index) in loans"
+				:key="`loan-${loan.id}`"
+				:item-index="index"
+				:loan-id="loan.id"
+				:exp-label="expLabel"
+				:show-action-button="true"
+				style="margin-right: 16px; max-width: 300px;"
+			/>
+		</div>
 	</div>
 </template>
 
 <script>
 import QuickFilters from '@/components/LoansByCategory/QuickFilters/QuickFilters';
-import { runFacetsQueries, fetchLoanFacets } from '@/util/loanSearch/dataUtils';
+import { runFacetsQueries, fetchLoanFacets, runLoansQuery } from '@/util/loanSearch/dataUtils';
 import { fetchCategories, FLSS_ORIGIN_CATEGORY } from '@/util/flssUtils';
 import { transformIsoCodes } from '@/util/loanSearch/filters/regions';
+import KivaClassicBasicLoanCard from '@/components/LoanCards/KivaClassicBasicLoanCard';
 
 export default {
 	name: 'QuickFiltersSection',
 	components: {
-		QuickFilters
+		QuickFilters,
+		KivaClassicBasicLoanCard
 	},
 	inject: ['apollo'],
 	data() {
 		return {
-			filtersLoaded: false,
+			totalCount: 0,
 			targetedLoanChannelURL: '',
+			filtersLoaded: false,
 			flssLoanSearch: {},
-			loanSearchState: {},
+			loanSearchState: {
+				pageOffset: 0,
+				pageLimit: 3
+			},
 			loans: [],
 			quickFiltersOptions: {
+				category: [{
+					key: 0,
+					title: 'All categories',
+				}],
 				gender: [{
 					key: '',
 					title: 'All genders'
@@ -46,16 +69,42 @@ export default {
 		};
 	},
 	async mounted() {
-		this.allFacets = await fetchLoanFacets(this.apollo);
+		this.allFacets = await fetchLoanFacets(this.apollo, true);
 		await this.fetchFilterData(this.flssLoanSearch);
-	},
-	computed: {
-		totalCount() {
-			return 0;
-		},
+		const { loans, totalCount } = await runLoansQuery(
+			this.apollo,
+			{ ...this.flssLoanSearch, ...this.loanSearchState },
+			FLSS_ORIGIN_CATEGORY
+		);
+		this.loans = loans;
+		this.totalCount = totalCount;
 	},
 	methods: {
-		updateQuickFilters() {
+		async updateQuickFilters(filter) {
+			if (filter.gender !== undefined) {
+				this.flssLoanSearch.gender = filter.gender;
+			} else if (filter.sortBy) {
+				this.flssLoanSearch.sortBy = filter.sortBy;
+			} else if (filter.country) {
+				this.flssLoanSearch.countryIsoCode = filter.country;
+			} else {
+				// We want to reset the flss paramaters for categories
+				delete this.flssLoanSearch.sectorId;
+				delete this.flssLoanSearch.tagId;
+				delete this.flssLoanSearch.activityId;
+				delete this.flssLoanSearch.themeId;
+				this.flssLoanSearch = {
+					...this.flssLoanSearch,
+					...filter
+				};
+			}
+			const { loans, totalCount } = await runLoansQuery(
+				this.apollo,
+				{ ...this.flssLoanSearch, ...this.loanSearchState },
+				FLSS_ORIGIN_CATEGORY
+			);
+			this.loans = loans;
+			this.totalCount = totalCount;
 		},
 		resetFilters() {
 		},
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -139,6 +139,7 @@
 
 <script>
 import { mdiFilterVariant } from '@mdi/js';
+import loanChannelQueryMapMixin from '@/plugins/loan-channel-query-map';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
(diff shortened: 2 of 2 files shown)


src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -46,16 +69,42 @@ export default {
...
+			this.apollo,
+			{ ...this.flssLoanSearch, ...this.loanSearchState },
+			FLSS_ORIGIN_CATEGORY
+		);
+		this.loans = loans;
+		this.totalCount = totalCount;
 	},
 	methods: {
-		updateQuickFilters() {
+		async updateQuickFilters(filter) {
```
egoroza, line 83: this will be cleaned up in future tickets, i got ahead of myself and got a head start 😃

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -182,7 +183,22 @@ export default {
 			showBadge: false
 		};
 	},
+	mixins: [
+		loanChannelQueryMapMixin
+	],
 	watch: {
+		selectedCategory(category) {
```
egoroza, line 190: (same with this)

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -14,29 +14,52 @@
 			@reset-filters="resetFilters"
 			@handle-overlay="handleQuickFiltersOverlay"
 		/>
+		<div
+			class="tw-flex tw-flex-wrap"
+		>
+			<kiva-classic-basic-loan-card
```
egoroza, line 20: this UI will also be cleaned up to match the figma... again, accidentally got ahead of myself 😂

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
mcstover: Generally, I'm ok with in-progress work that doesn't affect any live prod code. However, this file is actively used in prod so we'll need to wait on merging until the code in this is cleaned up and verfiied in the context of the existing/running quick filters exp.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -54,16 +73,42 @@ export default {
 		};
 	},
 	async mounted() {
-		this.allFacets = await fetchLoanFacets(this.apollo);
+		this.allFacets = await fetchLoanFacets(this.apollo, true);
```
dyersituations, line 76: What is the `true` here? I'm only seeing one param for `fetchLoanFacets`.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -54,16 +73,42 @@ export default {
...
 		await this.fetchFilterData(this.flssLoanSearch);
-	},
-	computed: {
-		totalCount() {
-			return 0;
-		},
+		const { loans, totalCount } = await runLoansQuery(
+			this.apollo,
+			{ ...this.flssLoanSearch, ...this.loanSearchState },
+			FLSS_ORIGIN_CATEGORY
```
dyersituations, line 81: Is this the correct query context? It would be good to ask the data team what context we should use for this new page, if we haven't already.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -54,16 +73,42 @@ export default {
...
+				delete this.flssLoanSearch.sectorId;
+				delete this.flssLoanSearch.tagId;
+				delete this.flssLoanSearch.activityId;
+				delete this.flssLoanSearch.themeId;
+				this.flssLoanSearch = {
+					...this.flssLoanSearch,
+					...filter
+				};
+			}
+			const { loans, totalCount } = await runLoansQuery(
```
dyersituations, line 105: This `runLoansQuery` could be extracted into a method in this component, since this call and the results being stored, are the same in both cases in this file. It will help with maintenance, such as if the context changes.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -54,16 +73,42 @@ export default {
...
+				this.flssLoanSearch.sortBy = filter.sortBy;
+			} else if (filter.country) {
+				this.flssLoanSearch.countryIsoCode = filter.country;
+			} else {
+				// We want to reset the flss paramaters for categories
+				delete this.flssLoanSearch.sectorId;
+				delete this.flssLoanSearch.tagId;
+				delete this.flssLoanSearch.activityId;
+				delete this.flssLoanSearch.themeId;
+				this.flssLoanSearch = {
```
dyersituations, line 100: What props do we need from `flssLoanSearch`? If there's only one prop, it would be simpler to just pull our that prop instead of delete. And we wouldn't need to add more deletes as more quick filters are added.

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -14,27 +14,46 @@
...
 		/>
+		<div
+			class="tw-flex tw-flex-wrap"
+		>
+			<kiva-classic-basic-loan-card
+				v-for="(loan, index) in loans"
+				:key="`loan-${loan.id}`"
+				:item-index="index"
+				:loan-id="loan.id"
+				:exp-label="expLabel"
```
dyersituations, line 25: `expLabel` doesn't exist on this component, so there's an error in the console for `/lending-home`

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
dyersituations: I tested this category quick filter locally, and it doesn't currently change the displayed loans, so I would exclude this file from this PR (or complete the filter functionality).

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -14,27 +14,46 @@
...
 		return {
-			filtersLoaded: false,
+			totalCount: 0,
 			targetedLoanChannelURL: '',
+			filtersLoaded: false,
 			flssLoanSearch: {},
-			loanSearchState: {},
+			loanSearchState: {
+				pageOffset: 0,
+				pageLimit: 3
```
dyersituations, line 55: Is there a future ticket for a pager? I see we only show 3 loans currently.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: sounds good i can remove this!

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: oops, removing this. was testing something out

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: a bit complicated - basically, `flssLoanSearch` determines what flss properties to apply to the flss query. This is based on the quick filter selections the user makes. We have added a new dropdown for filtering by category - however, categories can have different kinds of props associated with them for flss. Some are `themeId`, some use `sectorId`, etc. so this just resets all the props that aren't gender/sort by/location (these have specific props and hence can just be replaced when a user updates their quick filter selections) but it's more ambiguous for categories

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: yes!

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: Got it, thanks for the info. This works for now then, but I suggest creating a follow-up ticket to re-arch this a little. For example, this component could pull from the query map instead, and be the "owner" of the filter state. That would allow this particular section of code to simply pull from the map to "reset" the state.

dyersituations: Though I guess its parent could also be seen as the owner. All the to say, there's probably a way to setup this filtering path (category) for easier maintenance, since we have the query map as a point of truth, which would be simpler than having to update that filter removal list as we add filters to this quick filter feature.

roger-in-kiva:
For this case, you can use the `loan-channel-query-map` plugin I think. https://github.com/kiva/ui/blob/main/src/plugins/loan-channel-query-map.js

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: We have a meeting with the data team to check this and any other requirements from them

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -182,9 +183,11 @@ export default {
 			showBadge: false
 		};
 	},
+	mixins: [
+		loanChannelQueryMapMixin
```
dyersituations, line 187: Will these changes be used soon? If not, this file can be reverted.

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: I'll approve this PR, but I recommend adding a follow up ticket to extract some identical code here.

roger-in-kiva (approved):

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: I actually think we will want to potentially use a different query depending on results from loan card work - but yes I will note this as a follow up for restructuring!

PR #4430, opened 2022-12-02
Fix user context exp

christian14b:
Track experiment to only non-depositor or non-lender users

### DIFF
--- a/src/components/BorrowerProfile/LoanStory.vue
+++ b/src/components/BorrowerProfile/LoanStory.vue
@@ -253,7 +253,7 @@ export default {
 					id: '5',
 					headline: 'Supports climate action',
 					// eslint-disable-next-line max-len
-					body: 'Loans enable borrowers to use clean energy, partake in recycling, and adopt sustainable practices',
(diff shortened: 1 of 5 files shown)


src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -206,7 +211,7 @@ export default {
 			this.isoCode = loan?.geocode?.country?.isoCode ?? '';
 			this.tags = loan?.tags ?? [];
 			this.gender = loan?.gender ?? '';
-			this.sector = loan?.sector ?? '';
+			this.sector = loan?.sector ?? { name: '' };
```
dyersituations, line 214: Is this change needed with the optional chaining I added?

dyersituations (approved):

src/components/BorrowerProfile/LoanStory.vue
mcstover: I think it would overwrite the initial value which contains this stub object in `data` above if it came back undefined.

mcstover (approved):

PR #4433, opened 2022-12-05
feat: migrating campaign loan filters onto Kiva Classic

tiarascoleman:
Still TODOs (fix icon svg bug, gender and sortby configuration, and toggling btw the view layout modes)

Screen Shot 2022-12-05 at 2 22 17 PM


src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,181 +19,197 @@
...
-					@pill-toggled="(val) => { $emit('set-loan-display', val === 'rows') }"
-				/>
+					@view-toggled="(val) => { $emit('set-loan-display', val === 'rows') }"
+				>
+					<h4 class="tw-text-h4 tw-font-medium tw-text-action">
+						Row View
+					</h4>
+					<img
+						class="tw-h-5 tw-mr-1 tw-inline-block"
+						src="@/assets/icons/inline/list-green.svg"
```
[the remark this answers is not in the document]
mcstover: @tiarascoleman Here's an example of a Kiva Classic icon setup. https://github.com/kiva/ui/blob/main/src/components/BorrowerProfile/LoanBookmark.vue#L7
To make this Kiva Classic, you'll need to import the global kv-component, import the mdi icons declaring them in data and then update usage in the template.

Once you have that working  you can delete the 2 new image files from this pr as well.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
 </template>
 
 <script>
 import _isEqual from 'lodash/isEqual';
 import _sortBy from 'lodash/sortBy';
 import gql from 'graphql-tag';
+import { mdiGridLarge, mdiLandRowsHorizontal } from '@mdi/js';
 import KvButton from '@/components/Kv/KvButton';
 import KvChip from '@/components/Kv/KvChip';
 import KvLightbox from '@/components/Kv/KvLightbox';
```
mcstover, line 231: This import should be converted to use kv-components along with the signature in the template.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
 	</div>
 </template>
 
 <script>
 import _isEqual from 'lodash/isEqual';
 import _sortBy from 'lodash/sortBy';
 import gql from 'graphql-tag';
+import { mdiGridLarge, mdiLandRowsHorizontal } from '@mdi/js';
 import KvButton from '@/components/Kv/KvButton';
 import KvChip from '@/components/Kv/KvChip';
```
mcstover, line 230: This import should at minimum be converted to use KvPill which is a tailwind based version of a chip like interface. You'd need to update your event handling on these with this change.

mcstover: Current implementation isn't Kiva Classic and this could be a new ticket

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
 import KvLightbox from '@/components/Kv/KvLightbox';
 import AttributeFilter from '@/components/CorporateCampaign/LoanSearch/AttributeFilter';
 import GenderFilter from '@/components/CorporateCampaign/LoanSearch/GenderFilter';
 import KvAccordionItem from '@/components/Kv/KvAccordionItem';
```
mcstover, line 234: There's a global kv-component for this now too

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
+			</kv-lightbox>
+		</div>
 	</div>
 </template>
 
 <script>
 import _isEqual from 'lodash/isEqual';
 import _sortBy from 'lodash/sortBy';
 import gql from 'graphql-tag';
+import { mdiGridLarge, mdiLandRowsHorizontal } from '@mdi/js';
```
mcstover, line 228: 👍

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -305,6 +328,8 @@ export default {
 			selectedSort: null,
 			isChipsCollapsable: true,
 			isChipsCollapsed: true,
+			mdiGridLarge,
+			mdiLandRowsHorizontal
```
mcstover, line 332: 👍

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
+		</div>
 	</div>
 </template>
 
 <script>
 import _isEqual from 'lodash/isEqual';
 import _sortBy from 'lodash/sortBy';
 import gql from 'graphql-tag';
+import { mdiGridLarge, mdiLandRowsHorizontal } from '@mdi/js';
 import KvButton from '@/components/Kv/KvButton';
```
mcstover, line 229: You'll need to updated this to the kv-components import of KvButton instead of this ui version. Then you can change to the "secondary" variant

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -19,196 +19,220 @@
...
+						:icon="mdiLandRowsHorizontal"
+					/>
+					<span class="divider"></span>
+					<h4 class="tw-text-h4 tw-font-medium tw-text-action">
+						Grid View
+					</h4>
+					<kv-icon
+						class="tw-h-5 tw-mr-1 tw-inline-block"
+						name="grid-green"
+						src="@/assets/icons/inline/grid-green.svg"
```
dyersituations, line 44: Looks like the `src` can be removed from the `kv-icon` usages in this file, and the associated `svg` files removed from this PR.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -4,215 +4,254 @@
...
-					ref="chipsContainer"
+					class="tw-flex tw-cursor-pointer tw-items-center"
+					id="view-toggle"
 				>
-					<div
-						ref="chipsInnerContainer"
+					<span
+						class="tw-flex tw-items-center tw-flex-wrap"
+						:selected="activeLoanDisplay"
+						@click=" (val) => { $emit('set-loan-display', val === true) }"
```
mcstover, line 29: There is no longer a `val` to evaluate as that was emitted from KvPillToggle specifically.

You should be able to just use `@click="$emit('set-loan-display', true)"` here and `@click="$emit('set-loan-display', false)"` on the Grid span below

tiarascoleman (commented):
Need help on these aspects: 
- (Lines 61-82) The KvChipClassic appears however it does not collapse- how can I make it close?
- (Lines 92-133) I managed to put both containers on a grid however I still can't figure out how to make the grids flow with its side by side alignment- thoughts?
- (Lines 135-206) All these filters uses Checklist from the Autolending page and yesterday I attempted to change the checklist to the Kiva Classic KvCheckbox List yet all of the countries disappeared- how can I can the checkboxes to Kiva Classic design while maintaining the autofilled info? (Thank you! Here's also what it looks like so far!) 
Screen Shot 2022-12-09 at 10 16 38 AM
Screen Shot 2022-12-09 at 10 16 51 AM
Screen Shot 2022-12-09 at 10 17 11 AM

mcstover:
@tiarascoleman @michelleinez I think you all might want to use a temporary feature branch so you can make smaller PRs into the feature branch before moving them into `main`. This PR is getting quite large, so using a feature branch approach will allow you to break up the work a little more while working toward the "full" Kiva Classic version of this interface.

Please, create a new branch off of the latest version of `main`, something like `managed-lending-fitler-refresh` and then change the base on this pr to be this new branch. Then let's finish up what's started here before starting on any other big changes like trying to replace the locations listing etc...

tiarascoleman:
> @tiarascoleman @michelleinez I think you all might want to use a temporary feature branch so you can make smaller PRs into the feature branch before moving them into `main`. This PR is getting quite large, so using a feature branch approach will allow you to break up the work a little more while working toward the "full" Kiva Classic version of this interface.
> 
> Please, create a new branch off of the latest version of `main`, something like `managed-lending-fitler-refresh` and then change the base on this pr to be this new branch. Then let's finish up what's started here before starting on any other big changes like trying to replace the locations listing etc...

Got it and it has been moved! Still curious on you thoughts for the other two questions!

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -89,28 +90,47 @@
...
-					:initial-gender="initialGender"
-					:selected-gender="selectedGender"
-					@updated-filters="handleUpdatedFilters"
-				/>
-
-				<kv-accordion-item
-					class="loan-filters__lightbox-accordian"
-					id="sort-order-accordian"
+				<span
+					class="tw-flex tw-items-center"
```
mcstover, line 94: I think you'll need to add `tw-grid-cols-2` here so that the 2 contained divs below are spaced evenly within this container.

mcstover: You'll prob want to remove `tw-items-center` as that is likely causing the top header of each section to no line up

dyersituations:
@tiarascoleman What do you mean regarding the chips? That they can't be removed once filters are applied? If so, check out which events are being emitted and caught in parent components.

For the radio button filter layout, here's a general (incomplete) approach to try. I experimented locally with this branch, and this gets it pretty close, but some further work is needed. Hopefully this gets you moving toward the desired final solution. With these changes, the width of the gender filters isn't right, but that can be adjusted with other tweaks, like changing `flex-basis`.

image

image

mcstover (approved):
I think this PR is fine to merge into the feature branch as currently set.

PR #4440, opened 2022-12-07
Core 875 qf empty state

roger-in-kiva:
- ui added when having zero loans in quick filters section
- `KivaClassicBasicLoanCard` updated with computed properties avoiding errors when having undefined data for the loan and showing loading animations correctly


src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -62,7 +71,12 @@ export default {
 				pageOffset: 0,
 				pageLimit: 6
 			},
-			loans: [],
+			// Default loans for loading animations
+			loans: [
+				{ id: 1 }, { id: 2 }, { id: 3 },
+				{ id: 4 }, { id: 5 }, { id: 6 }
```
mcstover, line 77: I think you can just use `0` for all of these id values and that's what we do for the placeholders [in the new homepage loan grid](https://github.com/kiva/ui/blob/main/src/components/Homepage/HomeExp/KivaMultiCategoryGrid.vue#L94).

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -488,8 +513,8 @@ export default {
 				}
 			}
 
-			this.isLoading = false;
 			this.loan = result.data?.lend?.loan || null;
+			if (this.loan) this.isLoading = false;
```
dyersituations, line 517: Is the component stuck in loading if `loan` is `null`? Is this a typical situation a user could encounter?

dyersituations (approved):

src/components/LoanFinding/QuickFiltersSection.vue
roger-in-kiva: there is an error for the key using `0` for all of these id values
`[Vue warn]: Duplicate keys detected: 'loan-0'. This may cause an update error.`

dyersituations: Change line 30 of this file to use index instead of loan ID.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
roger-in-kiva: this is to avoid a weird behavior when switching between default dummy data, for loading animations, and real data to display the loans.
only situation I think about this is to have a wrong data from api and stuck the component in loading state, but maybe is better than show a wrong loan.

src/components/LoanFinding/QuickFiltersSection.vue
mcstover: I think doing as @dyersituations has suggested will actually solve your issue below. Pretty sure 0 doesn't resolve any loan data but 1-6 "could" resolve loan data...

roger-in-kiva: oh yeah, nice suggestion! thanks @dyersituations  and @mcstover

PR #4444, opened 2022-12-08
Vue 1463 - Move some borrower profile prefetch

dyersituations:
https://kiva.atlassian.net/browse/VUE-1463

- I couldn't find usages of `state`, `description`, `isExpiringSoon`, or `reservedAmount`
- `my.userAccount` is used to populate the URLs/UTMs in `ShareButton`, so those were moved to `mounted`

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -175,7 +175,8 @@ const shareButtonExpKey = 'share_button_bp';
 const userContextExpKey = 'new_users_context';
 
 const getPublicId = route => route?.query?.utm_content ?? route?.query?.name ?? '';
-const pageQuery = gql`
+
+const preFetchQuery = gql`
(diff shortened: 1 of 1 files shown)


eddieferrer (approved):
👍 good cleanup! 🧹

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -212,7 +213,6 @@ const pageQuery = gql`
 				}
 				geocode {
 					city
-					state
```
mcstover, line 215: Is this not needed for comparison in @eddieferrer 's new background image work?

mcstover (approved):
Other than the comment about the state, which may be handled elsewhere? Looks good!

src/pages/BorrowerProfile/BorrowerProfile.vue
dyersituations: `HeroBackground` has it's own query.

PR #4448, opened 2022-12-08
feat: data fetching method added for recommeded loans in lending page

roger-in-kiva:
`getRecommendedLoans` method added to get loans sorted by personalized for recommended loans row.

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -5,7 +5,10 @@
 				Welcome back, <span class="tw-text-action fs-mask">{{ firstName }}</span>
 			</h3>
 
-			<lending-category-section class="tw-mt-2" />
+			<lending-category-section
+				:loans="recommendedLoans"
+				class="tw-mt-2"
+			/>
 
 			<quick-filters-section class="tw-mt-2" />
 		</div>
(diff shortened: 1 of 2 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -48,5 +54,18 @@ export default {
 			return this.userInfo?.firstName ?? '';
 		}
 	},
+	methods: {
+		async getRecommendedLoans() {
+			const { loans } = await runLoansQuery(
+				this.apollo,
+				{ sortBy: 'personalized', pageLimit: 6 },
+				FLSS_ORIGIN_NOT_SPECIFIED
```
dyersituations, line 62: Should this be a more specific origin?

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -48,5 +54,18 @@ export default {
...
+			);
+			this.recommendedLoans = loans;
+		}
+	},
+	created() {
```
dyersituations, line 67: Do we want this on created or mounted?

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: yeah, waiting for DS team answer, I can change it in a next ticket.

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: To get the loans data we don't need access to the DOM so its okay to have this on created I think.

mcstover: If it's in created it will kick off on server render, if that is the intent, please very by loading the page locally without js to see if this call is completing. I think it "should" complete b/c you have the async/await combo. If it doesn't complete you'll prob need to tweak things.

mcstover: Or just move it to mounted skip the extra initial server load.

dyersituations: Yeah because it's not being awaited, I'm not sure when it officially completes, like if the server render waits (probably not). Mounted is probably the more efficient route, if we're okay with an initial loading state.

roger-in-kiva: ok cool! lets move it to mounted. thanks guys!

mcstover (approved):

PR #4449, opened 2022-12-08
feat: add unbounce script for donation pop up

eddieferrer:
Screen Shot 2022-12-08 at 3 24 23 PM
Screen Shot 2022-12-08 at 3 24 11 PM

- Sends event trigger to GTM, which loads unbounce script
- Pop up and additional display conditions controlled through unbounce. 

Have not added GTM tag/trigger to production, added in dev.

### DIFF
--- a/src/plugins/unbounce-event-mixin.js
+++ b/src/plugins/unbounce-event-mixin.js
@@ -3,9 +3,14 @@ import hasEverLoggedInQuery from '@/graphql/query/shared/hasEverLoggedIn.graphql
 
 export default {
 	inject: ['apollo', 'cookieStore'],
-	created() {
-		const unbounceRoute = this.$route?.meta?.unbounce ?? false;
+	mounted() {
+		const unbounceEmailCaptureRoute = this.$route?.meta?.unbounceEmailCapture ?? false;
+		const unbouncePopUpRoute = this.$route?.meta?.unbouncePopUp ?? false;
+
 		const utmEmailMedium = this.$route?.query?.utm_medium === 'email';
+
+		// activateUnbounceEvent is a GTM trigger that fires the unbounce email capture script.
+		// activateunbouncePopUp is a GTM trigger that fires the unbounce pop up script.
 
 		// Set cookie to disabled unbounce for future visits
 		if (utmEmailMedium) {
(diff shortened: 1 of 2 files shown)


dyersituations (approved):

src/plugins/unbounce-event-mixin.js
```diff
@@ -7,6 +7,9 @@ export default {
 		const unbounceRoute = this.$route?.meta?.unbounce ?? false;
```
mcstover, line 7: Can we convert this to mounted So it doesn't try to run it on the server?

eddieferrer: Yes I think thats fine 👍

src/plugins/unbounce-event-mixin.js
```diff
@@ -17,30 +20,53 @@ export default {
...
+				query: hasEverLoggedInQuery
+			});
+			// Setup unbounce event trigger which is restricted to non-logged-in and unknown users
+			hasEverLoggedIn = data?.hasEverLoggedIn ?? true;
+			// Check for hasEverLoggedIn
+			if (!hasEverLoggedIn) {
+				// push data object if eligible + assigned
+				if (typeof window !== 'undefined') {
+					window.dataLayer.push({ event: 'activateUnbounceEvent' });
+				}
```
mcstover, line 38: I think we ened to remove this block and let the code below handle it when `unbounceRoute` is active.

Also, how are we going to gate this so it's not firing all over the site and in checkout and on settings pages? I think `activateUnbounceDonation` should also be gated by `unbounceRoute`

mcstover: All of the restrictions we put on the original event seem like they should apply here too. If someone is coming to the site from one of our emails they likely have a goal already and it may already be donating so we prob shouldn't get in the way.

eddieferrer: Will remove block, I did not mean to have it twice. 

I think the idea was to gate it on the unbounce side. But I will add the meta property, for redundancy

eddieferrer: @mcstover I circled back with the team, and the desired behavior varies from the behavior of the email capture modal. 

To be clear, this modal will have the unbounce rule of "When a visitor tries to exit the page." (Im not sure what that means technically, I assume its a window unload event)

* The modal should only load for hasLoggedInUsers
* The modal should fire on a couple of pages, specifically the thank you page (which currently does not have the `unbounce: true` meta property)

mcstover: Ok can we at least target a small set of specific pages in the Unbounce setup? This shouldn't be firing anywhere within portfolio or settings or teams or any of the Ui borrower onboarding process or on checkout and def NOT on the `/cc/` routes for managed lending...

mcstover (approved):

src/router/routes.js
```diff
@@ -83,6 +84,7 @@ module.exports = [
 		component: () => import('@/pages/LandingPages/CorporateCampaign/CCLandingPage'),
 		meta: {
 			excludeFromStaticSitemap: true,
+			unbounceDonation: true,
```
mcstover, line 87: I think this needs to be false. We don't provide donation controls to corporate users in the standard way for their incontext checkout. Also, @michelleinez you and the team should discuss any other repercussions of asking for donations from corporate partner audiences as it will disrupt the flow and usage of their promo funds.

src/router/routes.js
```diff
@@ -12,7 +12,8 @@ module.exports = [
 
 				return exp?.version === 'b' ? 'hp/crowdfund-for-good' : 'home';
 			},
-			unbounce: true,
+			unbounceEmailCapture: true,
+			unbounceDonation: true,
```
mcstover, line 16: Is there something that will cancel out one of these? Otherwise someone with no unbounce cookie hitting the homepage would get 2 popups...

eddieferrer: This will be the responsibility of product with the Unbounce rules.

src/router/routes.js
```diff
@@ -83,6 +84,7 @@ module.exports = [
 		component: () => import('@/pages/LandingPages/CorporateCampaign/CCLandingPage'),
 		meta: {
 			excludeFromStaticSitemap: true,
+			unbouncePopUp: true,
```
mcstover, line 87: This needs to be removed or set to false.

mcstover: There are a variety of other custom operations in the /cc landing page space that could be affected by an additional popup. For example, this page shows in-context checkout in a popup and requires folks to leave to login. This is when we would typically show an unbounce pop up which would interrupt both login and checkout for these users.

mcstover (approved):

PR #4450, opened 2022-12-09
Vue 1407 saved search doesnt refresh in lend menu

dyersituations:
https://kiva.atlassian.net/browse/VUE-1407

- The lend menu private user data was only being queried on mount - changed to query when the menu opened, so that we don't have to remember to refetch a query whenever changing data displayed in the lend menu (there are several locations currently where we'd need to, for saved loans and searches)
- Removed lend menu `onOpen` since it seems to no longer exist
- `onLoad` is only called when the lend menu is opened, and it's a relative lightweight gql query

### DIFF
--- a/src/components/WwwFrame/LendMenu/TheLendMenu.vue
+++ b/src/components/WwwFrame/LendMenu/TheLendMenu.vue
@@ -35,7 +35,7 @@ import _groupBy from 'lodash/groupBy';
 import _map from 'lodash/map';
 import _sortBy from 'lodash/sortBy';
 import gql from 'graphql-tag';
-
+import logReadQueryError from '@/util/logReadQueryError';
 import { indexIn } from '@/util/comparators';
 import publicLendMenuQuery from '@/graphql/query/lendMenuData.graphql';
 import privateLendMenuQuery from '@/graphql/query/lendMenuPrivateData.graphql';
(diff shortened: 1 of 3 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -669,6 +669,11 @@ export default {
 	},
 	methods: {
 		toggleLendMenu(immediate = false) {
+			if (!this.isLendMenuVisible) {
+				// If menu was previously hidden, run menu onLoad
+				this.$refs?.lendMenu?.onLoad?.();
+			}
```
mcstover, line 675: Moving this seems risky as it previously depended on updated state from that whole `immediate` block...

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -121,14 +121,11 @@ export default {
 		},
 	},
 	methods: {
-		onOpen() {
-			this.$refs?.mega?.onOpen?.();
-		},
```
mcstover, line 126: Interesting that this disappeared...

dyersituations: Definitely, but it no longer exists in the mega menu component...

src/components/WwwFrame/TheHeader.vue
mcstover: Ok I think all of those conditions below are to handle various states that dictate IF the menu should call the load function. Bypassing the checks looks like it could affect behavior on touch devices and change the auto close behavior as well...

dyersituations: I did test all browser sizes, but I will find an alternate solution.

PR #4452, opened 2022-12-09
Header /lend links

dyersituations:
https://kiva.atlassian.net/browse/VUE-1437
https://kiva.atlassian.net/browse/VUE-1443

- Search navigates either to new or old filter page based on query params
- Lend menu country list navigates to new filter page
- Used full navigation to ensure that `/lend/filter` state updates like expected - simpler solution than special handling if the navigation happens from the filter page

### DIFF
--- a/src/components/WwwFrame/SearchBar.vue
+++ b/src/components/WwwFrame/SearchBar.vue
@@ -239,7 +239,14 @@ export default {
 					filterUrl = '/lend';
 				}
 
-				window.location.href = `${filterUrl}?${new URLSearchParams(query).toString()}`;
+				// When already on the filter page, prevent identical paths from being pushed
+				if (this.$router.currentRoute.fullPath !== `${filterUrl}?${new URLSearchParams(query).toString()}`) {
+					this.$router.push({ path: filterUrl, query });
+				}
+
+				// Exit searching state, in case user is already on filter page
+				this.searching = false;
+				this.$refs.input.blur();
 			}
 		},
 		formatResult({ label, matches }) {
--- a/src/components/WwwFrame/SearchBar.vue
+++ b/src/components/WwwFrame/SearchBar.vue
@@ -1,8 +1,6 @@
 <template>
 	<form
 		class="search-form tw-relative"
-		action="/lend"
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/SearchBar.vue
```diff
@@ -233,7 +232,14 @@ export default {
...
 				}
-				this.$router.push({ path: '/lend', query });
+
+				// Fallback to legacy filter if there's an unsupported query param
+				let filterUrl = '/lend/filter';
+				if (hasExcludedQueryParams(query)) {
+					filterUrl = '/lend';
+				}
+
+				window.location.href = `${filterUrl}?${new URLSearchParams(query).toString()}`;
```
mcstover, line 242: Can we please use the `$router` here so we don't force a page refresh?

dyersituations: Sure, it'll just require some more logic here. I originally used the router, but there was extra searchbar behavior that had to be taken into account. The old app does a full page load, for example.

mcstover: Ok, I'm also wondering how many cases hit the `suggestion.url` case and if we need to augment that url before the window location.

The in page nav would only be an option for Ui so if it's too much muck and stuff we can skip for now. CPS and monolith would still have to do full page refreshes.

dyersituations: There are currently no suggestions with `url`. Will there ever be? Can we remove it? I don't know the history of that prop.

dyersituations: Pushed in page navigation for searchbar. Kept the full page load for country list, since we'd have to have prop drill the close method from the lend menu, and I didn't think that was worth the extra complication.

mcstover: I'm guessing it's a special case for certain items but also don't know offhand.

dyersituations: Okay, I'll leave as is for now, since I don't have an example of what a suggestion `url` looks like.

PR #4453, opened 2022-12-09
updates and fixes for MARS-317

emuvente:
Fixes and updates after product review:

- Expand experiment target matching ability with depositor and lender checks instead of checking that in vue components
- Change how the stats slot is cycled through to make it easier to add and remove items or change the conditions for when items display
- Use `expectedPayments` for repayment date when available
- Estimate first repayment date for loans without an expected payment date yet (non-disbursed loans)
- Drop 'First' language when there is only 1 payment (`repaymentInterval === 'at_end'`)
- Track what repayment date is displayed to the user
- Use line-clamp for truncating loan story and 'more about this loan' sections instead of character count
- Track when users click the 'read more' buttons
- A couple typo fixes

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -54,8 +54,6 @@
 						@togglelightbox="toggleLightbox"
 						:num-lenders="numLenders"
 						:user-context-exp-variant="userContextExpVariant"
-						:has-lent-before="hasLentBefore"
-						:has-deposit-before="hasDepositBefore"
 					>
 						<template #sharebutton v-if="inPfp || shareButtonExpEnabled">
 							<!-- Share button for PFP loans -->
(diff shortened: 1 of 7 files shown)


mcstover (approved):
Nice improvements and cleanup.

@dyersituations There may be some clean up needed to merge this into your upcoming PR

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -595,38 +588,33 @@ export default {
...
-				loan(id: $loanId) {
-					id
-					...on LoanPartner {
-						partnerName
-						partner {
-							id
-							countries {
-								name
+		// Async data fetch for MARS-317
+		const { data } = await this.apollo.query({
```
dyersituations, line 592: Do we want to wrap this in a try catch?

dyersituations (approved):

src/pages/BorrowerProfile/BorrowerProfile.vue
emuvente: No, `query` rejects the promise it returns when there is an error instead of throwing. I think `readQuery` might be the only ApolloClient function that throws an error (at least the only one we use regularly).

dyersituations: Ah I see we've set the `errorPolicy` to `all` in `apollo.js`.

PR #4456, opened 2022-12-12
feat: category dropdown logic for MVP quick filters section

egoroza:


src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -187,7 +187,22 @@ export default {
...
 	watch: {
-		selectedCategory() {
+		selectedCategory(categoryId) {
+			const catId = Number(categoryId);
+			const queryMap = loanChannelQueryMapMixin.data().loanChannelQueryMap;
+			const categoryFilter = catId === 0 ? {} : queryMap
+				.find(channel => channel.id === catId)?.flssLoanSearch;
+			// These categories use location/gender/sort by for FLSS and need
+			// to have dropdowns preset
+			if (catId === 5) { // women
```
dyersituations, line 197: Changing category "selects" other filters in the page, but it doesn't reset others, for example, selecting location, then "women", doesn't reset Location. Is that expected?

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -187,7 +187,22 @@ export default {
...
+				this.selectedGender = 'female';
+			} else if (catId === 28) { // kiva-u-s
```
dyersituations, line 199: Are there other special cases coming? If so, there probably needs to be a programmatic approach to setting the other filters, rather than special cases. Otherwise if there are just these 3 for the time being, fine to leave as is.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -187,7 +187,22 @@ export default {
...
+				this.setCountry('US');
+			} else if (catId === 3) { // ending-soon
+				this.sortBy = 'expiringSoon';
+			} else {
+				this.$emit('update-filters', categoryFilter);
+			}
```
dyersituations, line 205: Looks like this method needs a `$kvTrackEvent` call.

egoroza: We have a separate ticket for tracking!

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: Should just be these 3 for now, unless we add new categories to Kiva with similar presets!

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: I'm double checking this with design~

egoroza: I actually ended up changing this to reset!

roger-in-kiva (approved):
👍

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -113,6 +114,7 @@ export default {
 				this.flssLoanSearch.gender = filter.gender;
 			} else if (filter.sortBy) {
 				this.flssLoanSearch.sortBy = filter.sortBy;
+				this.loanSearchState.sortBy = filter.sortBy;
```
dyersituations, line 117: Why does this case need to update `loanSearchState` but not the others?

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -179,15 +179,50 @@ export default {
...
 	],
 	watch: {
-		selectedCategory() {
+		selectedCategory(categoryId) {
+			const catId = Number(categoryId);
+			const queryMap = loanChannelQueryMapMixin.data().loanChannelQueryMap;
+			const categoryFilter = catId === 0 ? {} : queryMap
+				.find(channel => channel.id === catId)?.flssLoanSearch;
+
+			if (this.presetFilterActive.women) {
```
dyersituations, line 201: Couldn't all filters just be reset instead when changing category?

egoroza: design/product wants just the preset ones to be reset for the "special" categories!

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: the query uses `loanSearchState` for the initial query call on first mount

PR #4463, opened 2022-12-13
Mars 313 mask hotjar

christian14b:
For being able to turn on the full session recording functionality in Hot Jar we have to suppress PII.

[Hotjar data supression documentation](https://help.hotjar.com/hc/en-us/articles/115012439167#specific_elements)

### DIFF
--- a/src/components/Checkout/BasketVerification.vue
+++ b/src/components/Checkout/BasketVerification.vue
@@ -34,7 +34,8 @@
 			@lightbox-closed="close"
 		>
 			<p class="tw-mb-4">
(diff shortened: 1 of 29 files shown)


src/components/Checkout/BasketVerification.vue
```diff
@@ -34,7 +34,7 @@
 			@lightbox-closed="close"
 		>
 			<p class="tw-mb-4">
-				We sent a validation link <span v-if="email" class="email fs-exclude">to {{ email }}</span>.
+				We sent a validation link <span v-if="email" class="email data-hj-suppress">to {{ email }}</span>.
```
dyersituations, line 37: Do we not need the FullStory exclude anymore?

PR #4467, opened 2022-12-14
chore: update passport lib and logout method signature

mcstover:
Relates to updates required for [the dependabot pr](https://github.com/kiva/ui/pull/4163).

So, far I've only encounted the need for providing a callback in the logout function but want to do more verification before merging.

Notes:
https://github.com/jaredhanson/passport/compare/v0.4.1...v0.6.0
https://www.passportjs.org/tutorials/auth0/logout/
https://github.com/auth0/passport-auth0/releases


server/auth-router.js
```diff
@@ -101,9 +101,12 @@ module.exports = function authRouter(config = {}) {
...
 		const returnUrl = encodeURIComponent(`https://${config.host}`);
 		const logoutUrl = `https://${config.auth0.domain}/v2/logout?returnTo=${returnUrl}`;
-		req.logout(); // removes req.user
-		noteLoggedOut(res);
-		res.redirect(logoutUrl);
+		req.logout({}, err => {
+			console.error(err);
+			// removes req.user
+			noteLoggedOut(res);
+			res.redirect(logoutUrl);
```
dyersituations, line 108: So this redirect is only needed if there's an error?

server/auth-router.js
```diff
@@ -101,9 +101,12 @@ module.exports = function authRouter(config = {}) {
 		info(`LoginUI: execute logout, session id:${req.sessionID}, cookie:${getSyncCookie(req)}, user id:${req.user && req.user.id}`); // eslint-disable-line max-len
 		const returnUrl = encodeURIComponent(`https://${config.host}`);
 		const logoutUrl = `https://${config.auth0.domain}/v2/logout?returnTo=${returnUrl}`;
-		req.logout(); // removes req.user
-		noteLoggedOut(res);
-		res.redirect(logoutUrl);
+		req.logout({}, err => {
```
eddieferrer, line 104: Looking at the example you linked in this PR, I think this should be:
`req.logout((err) => {` ?

mcstover: `err` is the only param passed into this callback.

The first param is an options object, but yea it may be ok to omit the empty `{}`.

server/auth-router.js
mcstover: `err` is the only param passed into this callback.

dyersituations: So it provides an error if there is one, but the same callback is used for success?

mcstover: Yea, providing a callback is now required. Without it there were errors

dyersituations: In that case, do we want an `if (err)` around the `console.error`?

mcstover: Yes I think that's a good callout.

mcstover:
Also, just to note, this PR is currently up to bring attention to a potentially "risky" update, just b/c it involves login + auth. I'd like for others to run this code in a full VM to help validate this before we actually merge it.

mcstover:
Also, I specifically didn't generate the package-lock b/c it would just introduce conflicts later as other items are moving into the repo over the next week or so. We can hold on finalizing this until after the new year.

server/auth-router.js
```diff
@@ -101,9 +101,14 @@ module.exports = function authRouter(config = {}) {
...
+			if (err) {
+				error('LoginUI: logout callback error:', err);
+			}
+			// removes req.user
```
emuvente, line 108: This comment seems to imply that calling `noteLoggedOut(res)` will remove `req.user`, but it's actually the call to req.logout that does that.

mcstover: good point will move up to correlate to the proper call

mcstover:
Verified login and logout in the VM

emuvente (approved):
👍 LGTM

michelleinez (approved):

dyersituations:
@mcstover Any concerns of getting the conflicts on this PR resolved and the PR merged? I can finish that out, if you think the changes here are safe/ready still.

mcstover (commented):
The primary change was that logout now provides a callback. So long as that's still the case in the latest version we should be ok to move forward.

dyersituations:
@mcstover due to conflicts, I'll move these changes to: https://github.com/kiva/ui/pull/5086

PR #4468, opened 2022-12-14
feat: add loan ids to various social share events

eddieferrer:
* Also removed `this` from inside templates
* Standardized `handleFacebookResponse` method. I believe one of these was firing an error event when it should not have

ACK-495

### DIFF
--- a/src/components/BorrowerProfile/ShareButton.vue
+++ b/src/components/BorrowerProfile/ShareButton.vue
@@ -8,6 +8,7 @@
 		:utm-content="utmContent"
 		:linked-in-title="linkedInTitle"
 		:open-lightbox="forceLightbox"
+		:loan-id="loan.id"
 	>
 		<template #modal-content>
 			<!-- eslint-disable-next-line max-len -->
--- a/src/components/Checkout/SocialShareV2.vue
+++ b/src/components/Checkout/SocialShareV2.vue
@@ -13,7 +13,7 @@
 				data-testid="share-facebook-button"
 				class="social__btn social__btn--facebook"
 				:href="facebookShareUrl"
-				v-kv-track-event="['thanks', 'Social-Share-Lightbox', 'click-Facebook-share']"
+				v-kv-track-event="['thanks', 'Social-Share-Lightbox', 'click-Facebook-share', selectedLoanId]"
 			>
 				<kv-icon name="facebook-round" title="Facebook" class="social__icon" />
 				<span>Share</span>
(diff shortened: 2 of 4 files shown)


src/components/BorrowerProfile/ShareButton.vue
```diff
@@ -8,6 +8,7 @@
 		:utm-content="utmContent"
 		:linked-in-title="linkedInTitle"
 		:open-lightbox="forceLightbox"
+		:loan-id="loan.id"
```
dyersituations, line 11: Do we need to add this to the `ThanksPageEcoChallenge` as well? Not sure if that page is still active.

dyersituations (approved):

src/components/BorrowerProfile/ShareButton.vue
eddieferrer: Good call, but no. That challenge is no longer active, and the share on that page shares a climate fact, not a loan.

src/components/Checkout/SocialShareV2.vue
```diff
@@ -47,11 +47,11 @@
 				class="social__btn social__btn--link tw-text-link tw-border-tertiary tw-border"
 				:class="copyStatus.class"
 				:disabled="copyStatus.disabled"
-				v-kv-track-event="['thanks', 'Social-Share-Lightbox', 'click-Copy-link-share']"
+				v-kv-track-event="['thanks', 'Social-Share-Lightbox', 'click-Copy-link-share', selectedLoan.id]"
```
mcstover, line 50: I think we need to break out a computed property for use in the template that returns `selectedLoanId` in a safe way in case it's not present. This would fix the potential undefined error down in the `handleFacebookResponse` as well.

src/components/Thanks/ThanksPageShare.vue
```diff
@@ -368,25 +372,15 @@ export default {
...
-							'error-Social-Share-Lightbox'
-						);
 					}
-				} else {
 					this.$kvTrackEvent(
 						'thanks',
 						'click-Facebook-share',
-						'error-Social-Share-Lightbox'
+						'error-Social-Share-Lightbox',
+						this.loan.id
```
mcstover, line 381: Please guard this against a missing id.

src/components/Checkout/SocialShareV2.vue
eddieferrer: I don't believe this will throw an error. the computed for `selectedLoan` will return an empty object if theres no loan for some reason. So accessing `.id` on this empty object will just send undefined to our track event function in analytics, which the function can handle

eddieferrer:
Added computed props to guard against weirdness

mcstover (approved):
Thanks!

PR #4470, opened 2022-12-14
Core 908 matched loans lending page

roger-in-kiva:
- matched lending section added to lending-home page
- `LendingCategorySection` component updated

* there is a ticket to update data fetching method: flss is not working fine in production when using `isMatchable: {eq: true}` as a filter and using `isMatched: true` in old stack is getting zero results in dev


src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -226,6 +226,14 @@ export default {
 				this.$emit('handle-overlay', false);
 			}
 		},
+		closeRegions() {
+			if (this.showRegions) {
+				this.showRegions = false;
+				this.selectedRegion = null;
+				document.documentElement.style.overflow = 'auto';
+				this.$emit('handle-overlay', false);
```
dyersituations, line 234: Can't we call `toggleRegions` instead?

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -67,12 +84,21 @@ export default {
 			);
 			this.recommendedLoans = loans;
 		},
-		trackRecommended({ success }) {
-			if (success) this.$kvTrackEvent('loan-card', 'add-to-basket', 'recommended-lending-home');
+		async getMatchedLoans() {
+			const { loans } = await runLoansQuery(
+				this.apollo,
+				{ filterObject: { isMatchable: { eq: true } }, sortBy: 'personalized', pageLimit: 9 },
```
dyersituations, line 90: Default is `personalized`, so that `sortBy` shouldn't be needed.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -6,12 +6,24 @@
...
+				:per-step="2"
 				class="tw-mt-2"
-				@add-to-basket="trackRecommended"
+				@add-to-basket="trackCategory($event, 'recommended')"
 			/>
 
 			<quick-filters-section class="tw-mt-6" />
+
+			<!-- Second category row: Matched loans section -->
+			<lending-category-section
```
dyersituations, line 20: Based on the slack convo, it sounds like this section should be hidden until matched worked as expected in FLSS?

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
roger-in-kiva: It was calling `toggleRegions` but there is a bug described here; toggling when dont needed: https://kivahq.slack.com/files/U0278NCFQ01/F04FC8U2TKN/screen_recording_2022-12-14_at_1.51.51_pm.mov

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: Right, but this will be updated in other ticket switching to old stack to work in production until matched worked as expected in FLSS

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: same above

src/pages/LoanFinding/LoanFinding.vue
dyersituations: Sounds good, but we still might want to hide until the final solution is complete, unless this page isn't public anywhere.

roger-in-kiva: oh, yeah, this page isn't public anywhere, is managed by an experiment turned off in production. this is for product QAing on dev

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
dyersituations: That link isn't working for me, but my guess based on the original code was that it displayed when outside of the location selector was clicked? Generally this type of logic should be consolidated into one path, for example if the emitted event ever changes, we'd have to remember to update all locations. I'll resolve comment for now since `toggleRegion` is used a bunch in this file, and I don't want to derail this work, but I recommend in the near future thinking of a way to combine these methods. One possible solution would be provide an override flag to the toggle method.

dyersituations (approved):

PR #4476, opened 2022-12-15
Core 932 borrower profile bug

roger-in-kiva:
loanId added as computed property to borrower profile page to fix the bug: 
`when you reach a fully funded borrower profile and click on any recommended borrower profiles (through their loan card) the information is incorrect.`

here to reproduce it https://www.kiva.org/lend/2456251


mcstover (approved):
From what I'm seeing in Prod, you can get into a state where "parts" of the spa loaded BPs don't fully update.

emuvente (approved):
Nice find!

dyersituations (approved):

PR #4479, opened 2022-12-16
feat: logic for mvp quick filters

egoroza:
this PR patches up the logic for quick filters on MVP loan finding page. this PR implements:
* logic for pagination (including showing a view more CTA on pg 2)
* resetting filters
* handling overlay for location selector
Screen Shot 2022-12-16 at 12 48 15 PM
Screen Shot 2022-12-16 at 12 48 31 PM
Screen Shot 2022-12-16 at 12 49 47 PM


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -1,7 +1,8 @@
 <template>
 	<div
 		class="tw-flex tw-flex-col"
-		style="min-width: 230px; max-width: 374px; height: 100%;"
+		style="min-width: 230px; height: 100%;"
```
mcstover, line 4: Can this line be removed now that it's replicated below?

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -102,14 +119,33 @@ export default {
 		const { loans, totalCount } = await runLoansQuery(
 			this.apollo,
 			{ ...this.flssLoanSearch, ...this.loanSearchState },
-			FLSS_ORIGIN_CATEGORY
+			FLSS_ORIGIN_LENDING_HOME
```
mcstover, line 122: Is this component also used on Category pages still? Seems like this should be conditional based on the context.

roger-in-kiva:
@egoroza  Can you include this close method in your work and verify is working fine with the overlay?
https://github.com/kiva/ui/blob/e89d5d18b696b2d3e626668db1d53d586f7fda13/src/components/LoansByCategory/QuickFilters/LocationSelector.vue#L2

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: Good catch thanks, I totally forgot it's still on cat pages for some reason I thought we removed it😅😂

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -102,14 +119,33 @@ export default {
...
 		);
 		this.loans = loans;
 		this.totalCount = totalCount;
 		this.backupLoans = this.loans.slice(3);
 	},
+	computed: {
+		filterPageUrl() {
+			const location = this.flssLoanSearch.countryIsoCode?.toString();
+			const paramStr = JSON.parse(JSON.stringify({
```
dyersituations, line 131: The JSON parse/stringify shouldn't be needed here. `URLSearchParams` can take an object directly.

egoroza: Sometimes some of the values are undefined, and using toString on URL Search Params seems to keep them there (i.e. `gender=undefined`) - the parsing removes the undefined values beforehand

dyersituations: Gotcha. In that case, I would add a comment above this line, since I could see this parsing being removed accidentally without context in the future.

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: oh wait - this is actually specific to the loan finding page and not re-used. `QuickFiltersSection` is the parent component on loan finding for `QuickFilters` while it's `LoanChannelCategoryControl` on Category pages. so should be ok!

src/components/LoanCards/KivaClassicBasicLoanCard.vue
dyersituations: Looks like this duplicate CSS can still be removed.

dyersituations:
I'm not sure if this is related to this PR or a previous changeset, but when the location selector is open, you can't scroll down to see the rest of the modal. I'd recommend logging a ticket, if we don't have one, or completing with this PR.
image

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -102,14 +119,33 @@ export default {
...
 			{ ...this.flssLoanSearch, ...this.loanSearchState },
-			FLSS_ORIGIN_CATEGORY
+			FLSS_ORIGIN_LENDING_HOME
 		);
 		this.loans = loans;
 		this.totalCount = totalCount;
 		this.backupLoans = this.loans.slice(3);
 	},
+	computed: {
+		filterPageUrl() {
```
dyersituations, line 129: When first changing the gender dropdown, this filter URL doesn't get updated. And if you select all genders, you get in state where you have a URL like `?gender=`

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -142,13 +178,17 @@ export default {
 				this.loans = this.backupLoans;
 			}
 		},
-		resetFilters() {
+		async resetFilters() {
```
dyersituations, line 181: When you click reset, the see more button gets the follow URL: `/lend/filter?location=`

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tw-relative tw-flex tw-flex-col tw-w-full">
+	<div class="tw-relative tw-flex tw-flex-col tw-w-full" v-click-outside="toggleRegions">
```
dyersituations, line 2: In order for `v-click-outside` to work, the component needs the `clickOutside` mixin. See the component `MonthlyGoodSelectorDesktop` for an example usage.

egoroza:
> I'm not sure if this is related to this PR or a previous changeset, but when the location selector is open, you can't scroll down to see the rest of the modal. I'd recommend logging a ticket, if we don't have one, or completing with this PR. image

yeah this is expected and how it worked on the category page... I'll probably add a ticket as a follow up

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
egoroza: thank you!! I was looking into this and that makes sense now

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: okay this should be fixed now! also moved it to `methods` that way it gets called each time a new CTA is rendered. I think the issue was it wasn't re-rendering since it bases the values off of `flssLoanSearch` which is an object which technically doesn't get updated? And also put in a fallback case for `?gender=` to replace empty strings with `undefined`

src/components/LoanFinding/QuickFiltersSection.vue
egoroza: This was caused by the `filterPageUrl` issue above, should be ok now^

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: It's likely related to Vue not detecting due to the property started out undefined (I didn't check if that's what is happening here, but Vue can't detect that situation). But yes, a method works fine.

dyersituations (approved):

PR #4480, opened 2022-12-19
Mars 318 update lbc page title tags

christian14b:
Update title tags in lend by category page


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -100,7 +100,7 @@ export default {
 				{
 					vmid: 'og:title',
 					property: 'og:title',
-					content: title ?? this.meta.title
+					content: title ?? `${this.meta.title} | Invest & Support`
```
emuvente, line 103: " | Invest & Support" should be added to `this.meta.title`, not just the og:title and twitter:title.

src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -372,130 +372,261 @@ export default {
 		},
 		getMetaInfo() {
 			switch (this.targetedLoanChannel) {
+				case '1-billion-in-change':
```
emuvente, line 375: I want us to use this ticket as an opportunity to set the meta title based on the category name from the API, so that we don't have to hard-code it in this file any more. The meta title just needs to be `${category.name} | Invest & Support`.

The category name is already fetched from the API in the `LoanChannelCategoryClimateExperiment` and `LoanChannelCategoryControl` components using functions from the `loanChannelUtils` file, so I would recommend defining the meta info in those components instead of this one. I would also recommend putting the function(s) that get/set the meta info in the `loanChannelUtils` file so that it can be shared by both category layout components.

emuvente (commented):
There's a conflict due to the 310 fix being merged, but once that's resolved this looks good to go!

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -202,8 +203,78 @@ function getPageOffset(query, limit) {
...
 export default {
 	name: 'LoanChannelCategoryControl',
+	metaInfo() {
+		const channelDescription = getMetaDescription(this.$route.params.category);
+		let image = '';
+		let title = null;
+		if (this.$route.query.category_share_version
+			&& ['women', 'education', 'agriculture'].includes(this.$route.params.category)) {
+			image = imageRequire(`./${this.$route.params.category}_share_card.png`);
+			title = `Can you help ${this.$route.query.lender} `;
```
emuvente, line 217: This should be
```
title = this.$route.query.lender ? `Can you help ${this.$route.query.lender} ` : 'Can you help ';
```

christian14b: Good catch! Thanks!

emuvente (approved):
Looks great!

PR #4481, opened 2022-12-19
Core 917 partner spotlight lending page

roger-in-kiva:
partner spotlight section added to lending page

there is a ticket to update tracking code for data fetching

### DIFF
--- a/src/components/LoanFinding/PartnerSpotlightSection.vue
+++ b/src/components/LoanFinding/PartnerSpotlightSection.vue
@@ -1,0 +1,66 @@
+<template>
+	<div>
+		<m-f-i-hero />
+
+		<mfi-loans-wrapper
+			v-if="selectedChannelLoanIds.length > 0"
+			:selected-channel-loan-ids="selectedChannelLoanIds"
+			:selected-channel="selectedChannel"
+			class="tw-my-4"
+		/>
+	</div>
+</template>
+
+<script>
+import MFIHero from '@/components/LoansByCategory/MFIRecommendations/MFIHero';
+import MfiLoansWrapper from '@/components/LoansByCategory/MFIRecommendations/MfiLoansWrapper';
+import mfiRecommendationsLoans from '@/graphql/query/lendByCategory/mfiRecommendationsLoans.graphql';
+
+export default {
+	name: 'PartnerSpotlightSection',
+	components: {
+		MFIHero,
+		MfiLoansWrapper
+	},
+	inject: ['apollo', 'cookieStore'],
+	data() {
+		return {
+			mfiRecommendationsLoans: [],
+			selectedChannel: {},
+		};
+	},
+	computed: {
+		selectedChannelLoanIds() {
+			return this.mfiRecommendationsLoans?.map(element => element.id) ?? [];
+		},
+	},
+	methods: {
+		fetchMFILoans() {
+			// Load mfi recommendations loans data
+			return this.apollo.query({
+				query: mfiRecommendationsLoans,
+			}).then(({ data }) => {
+				this.mfiRecommendationsLoans = data?.fundraisingLoans?.values ?? [];
+				const numberLoans = this.mfiRecommendationsLoans.length;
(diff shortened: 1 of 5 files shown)


src/components/LoanFinding/PartnerSpotlightSection.vue
```diff
@@ -0,0 +1,66 @@
...
+	methods: {
+		fetchMFILoans() {
+			// Load mfi recommendations loans data
+			return this.apollo.query({
+				query: mfiRecommendationsLoans,
+			}).then(({ data }) => {
+				this.mfiRecommendationsLoans = data?.fundraisingLoans?.values ?? [];
+				const numberLoans = this.mfiRecommendationsLoans.length;
+				if (numberLoans > 0) {
+					this.$kvTrackEvent(
```
dyersituations, line 46: Is this one of the tracking events that will change? I'm guessing we want different info for this event.

src/components/LoansByCategory/MFIRecommendations/MfiLoansWrapper.vue
```diff
@@ -18,7 +18,7 @@
 			<a
 				class="tw-text-action tw-text-base"
 				href="https://www.kiva.org/lend?partner=59&status=fundRaising&sortBy=newest"
-				v-kv-track-event="['Lending', 'click-MFI-feature', 'view all']"
+				v-kv-track-event="['event-tracking', 'click', 'mfi-view-all']"
```
dyersituations, line 21: Is `event-tracking` what we want here? It seems too general. Maybe that's also part of the tracking ticket.

src/components/LoanFinding/PartnerSpotlightSection.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+	<div>
+		<m-f-i-hero />
+
+		<mfi-loans-wrapper
```
dyersituations, line 5: If we don't already have a ticket for it, I recommend adding a ticket to change the `MfiLoansWrapper` to navigate to `/lend/filter` instead of legacy. We've been updating the app to navigate to the new lend page recently, so we don't have to introduce new links to the legacy filter page where we can. It looks like it's a one-line fix in that component, so you could do it with this PR or add a ticket.

src/components/LoansByCategory/MFIRecommendations/MfiLoansWrapper.vue
roger-in-kiva: we can fix it now, any suggestion from category list?

src/components/LoanFinding/PartnerSpotlightSection.vue
roger-in-kiva: Nice suggestion, Ill do it whit this PR, lemme update it

src/components/LoanFinding/PartnerSpotlightSection.vue
roger-in-kiva: There is one ticket to consider this. ty

src/components/LoansByCategory/MFIRecommendations/MfiLoansWrapper.vue
roger-in-kiva: lending is no more an options for categories, you can check available options here https://docs.google.com/spreadsheets/d/1pbxxYDfeOkrGhFi8HWl5K-vCPXaqO3bXS1Io3E_SDKc/edit#gid=1801894844

dyersituations: Oh I see we are needing to use a new style here for event strings. If we don't have it already, please log a new ticket to work with PM/Data on the expected strings here. I'll resolve this comment for now.

dyersituations: Yeah we'll need to expand the categories with the data team.

dyersituations (approved):

PR #4484, opened 2022-12-20
feat: tracking for mvp quick filters

egoroza:
adds tracking for quick filters on MVP lending page. also fixes a bug with the location selector.

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -313,6 +313,16 @@ export default {
 		useFullWidth: {
 			type: Boolean,
 			default: false
+		},
+		atcTrackingProps: {
+			type: Object,
+			default: () => {
+				return {
+					category: '',
+					action: '',
+					label: '',
+				};
+			}
 		}
 	},
 	inject: ['apollo', 'cookieStore'],
@@ -542,6 +552,13 @@ export default {
 
 				this.$showTipMsg(msg, 'error');
 			});
+			if (this.showActionButton) {
+				this.$kvTrackEvent(
+					this.atcTrackingProps.category,
+					this.atcTrackingProps.action,
+					this.atcTrackingProps.label
+				);
+			}
 		},
 	},
 	mounted() {
--- a/src/components/LoanCards/KivaClassicBasicLoanCard.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCard.vue
@@ -552,7 +552,7 @@ export default {
 
 				this.$showTipMsg(msg, 'error');
 			});
-			if (this.showActionButton) {
+			if (this.atcTrackingProps.category) {
 				this.$kvTrackEvent(
 					this.atcTrackingProps.category,
 					this.atcTrackingProps.action,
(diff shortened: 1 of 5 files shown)


src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -542,6 +552,13 @@ export default {
 
 				this.$showTipMsg(msg, 'error');
 			});
+			if (this.showActionButton) {
```
dyersituations, line 555: This check might want to be on the tracking props instead, just in case a future feature uses the action button and someone forgets the tracking info. Empty events would be confusing for data analytics.

Even just `if (this.atcTrackingProps.category)` might help with maintenance.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -168,6 +169,10 @@ export default {
 			type: Boolean,
 			default: false
 		},
+		trackingCategory: {
+			type: String,
+			default: 'search'
```
dyersituations, line 174: I'd recommend removing the default and instead making this a required prop. It looks like this changeset already sets this prop, so it might make sense to just make it required (also prevents the wrong category maybe being logged in the future if this component is reused).

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -181,6 +181,10 @@ export default {
 			type: Boolean,
 			default: false
 		},
+		trackingCategory: {
+			type: String,
+			default: 'search'
```
dyersituations, line 186: Same comment on potentially making this required.

dyersituations (approved):

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -36,6 +37,11 @@
 				:loan-id="loan.id"
 				:show-action-button="true"
 				:use-full-width="true"
+				:atc-tracking-props="{
```
mcstover, line 40: @dyersituations  @egoroza I think this would have been better handled by using the add-to-basket event and loan id that is already emitted from this component. We should look into refactoring this specific need to use that emitted event to trigger your tracking call instead of embedding it in the Ui version of KivaClassicLoanCard which now has a new discrepancy across applications.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -313,6 +313,16 @@ export default {
 		useFullWidth: {
 			type: Boolean,
 			default: false
+		},
+		atcTrackingProps: {
```
mcstover, line 317: We should look to remove this and use the existing `add-to-basket` event that is already emitted

src/components/LoanFinding/QuickFiltersSection.vue
dyersituations: @mcstover Good call, I wasn't aware of that event (I'll need to dig into this existing component more).

@egoroza It would be good to add a new ticket to tackle during the fixathon to refactor this usage to prevent tech debt.

mcstover: thanks!

PR #4490, opened 2022-12-22
Mars 324 thanks page for funded loans

christian14b:
There is currently an edge case in the Thanks page where a user sees a share ask appeal to share with “0 more people” when a loan that they’ve checked out with is fully funded. 

Changes are :
- New design
- Share links to the homepage with tailored tags.

## Current
image

## Change
image


mcstover (approved):
I think this is fine for now but could use some tuning around using computed values a little later on as we solidify the share feature. We'll also want to think about some alternatives or cache bypasses for all of the dynamic meta in the homepage in the new year.

dyersituations (approved):

emuvente (approved):

PR #4491, opened 2022-12-22
fix: shorten year review share title for twitter requirements

emuvente:
Twitter stops showing the title in the preview when it gets longer than 81 characters. Their docs say that it shouldn't be longer than 70 characters, but they apparently have some wiggle room.


mcstover (approved):
ok then ;)

dyersituations (approved):

christian14b (approved):

PR #4495, opened 2023-01-03
fix: paragraph spacing in expandable questions

christian14b:
There was not space between paragraphs in expandable questions. This bug was introduced while working on show one answer at a time. 

## Before
image

## After
image


mcstover (approved):
This is a good quick fix. We now have a kv-components version of these components for use as well which could be swapped in. The fix should already be in place in the global components...

dyersituations (approved):

eddieferrer (approved):

PR #4500, opened 2023-01-04
fix: qa fixes for quick filters on mvp lending page

egoroza:
This PR addresses a few things that came out of QA'ing the Quick Filters section of the MVP Lending Page:
* Update the progress bar to use the fundraising bar (so UI works with the grey background)
* Update some tracking (added tracking for selecting a category and refactor existing tracking for loan cards ATB)
* Ensure that "special loans" that use `partnerId` for their filter object work correctly by factoring `partnerId`s into the filter object, and added an LGBTQ channel map entry so that it works with the FLSS query
* Fix other "special loan" edge cases (women/Kiva US/ending soon)
* Some UI fixes requested by design

### DIFF
--- a/src/components/LoanCards/LoanProgressGroup.vue
+++ b/src/components/LoanCards/LoanProgressGroup.vue
@@ -3,7 +3,13 @@
 		<h4 class="tw-mb-0.5">
 			{{ fundingText }}
 		</h4>
+		<fundraising-status-meter
+			v-if="useNewProgress"
+			:is-funded="progressPercent === 100"
+			:percent-raised="progressPercent"
+		/>
 		<kv-progress-bar
+			v-else
 			class="tw-mb-1.5 lg:tw-mb-1"
 			aria-label="Percent the loan has funded"
 			:value="progressPercent * 100"
@@ -13,12 +19,14 @@
 
 <script>
 import numeral from 'numeral';
+import FundraisingStatusMeter from '@/components/LoanCards/FundraisingStatus/FundraisingStatusMeter';
 import KvProgressBar from '~/@kiva/kv-components/vue/KvProgressBar';
 
 export default {
 	name: 'LoanProgressGroup',
 	components: {
 		KvProgressBar,
+		FundraisingStatusMeter
 	},
 	props: {
 		moneyLeft: {
@@ -34,6 +42,10 @@ export default {
 			default: '',
 		},
 		allSharesReserved: {
+			type: Boolean,
+			default: false,
+		},
+		useNewProgress: {
 			type: Boolean,
 			default: false,
 		}
--- a/src/components/LoanCards/LoanProgressGroup.vue
+++ b/src/components/LoanCards/LoanProgressGroup.vue
@@ -3,13 +3,7 @@
 		<h4 class="tw-mb-0.5">
 			{{ fundingText }}
 		</h4>
-		<fundraising-status-meter
-			v-if="useNewProgress"
-			:is-funded="progressPercent === 100"
-			:percent-raised="progressPercent"
-		/>
 		<kv-progress-bar
-			v-else
 			class="tw-mb-1.5 lg:tw-mb-1"
 			aria-label="Percent the loan has funded"
 			:value="progressPercent * 100"
@@ -19,14 +13,12 @@
 
 <script>
 import numeral from 'numeral';
-import FundraisingStatusMeter from '@/components/LoanCards/FundraisingStatus/FundraisingStatusMeter';
 import KvProgressBar from '~/@kiva/kv-components/vue/KvProgressBar';
 
 export default {
 	name: 'LoanProgressGroup',
 	components: {
 		KvProgressBar,
-		FundraisingStatusMeter
 	},
 	props: {
 		moneyLeft: {
@@ -45,10 +37,6 @@ export default {
 			type: Boolean,
 			default: false,
 		},
-		useNewProgress: {
-			type: Boolean,
-			default: false,
-		}
 	},
 	computed: {
 		fundingText() {
--- a/src/components/LoanFinding/QuickFiltersSection.vue
+++ b/src/components/LoanFinding/QuickFiltersSection.vue
@@ -157,6 +157,9 @@ export default {
 			const params = new URLSearchParams(paramStr);
 			return `/lend/filter?${params.toString()}`;
 		},
+		// TODO: Rearchitect this at some point.
+		// This won't work for categories that have
+		// multiple criteria applied to their FLSSLoanSearch criteria.
 		async updateQuickFilters(filter) {
 			this.loanSearchState.pageOffset = 0;
 			if (filter.gender !== undefined) {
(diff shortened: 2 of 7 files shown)


roger-in-kiva:
looking good @egoroza!
are you considering to show recommended as default option for sort by dropdown in category page?
there is not a match with almost funded option for initial loans there.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -202,7 +202,7 @@ export default {
 			const queryMap = loanChannelQueryMapMixin.data().loanChannelQueryMap;
 			const categoryFilter = catId === 0 ? {} : queryMap
 				.find(channel => channel.id === catId)?.flssLoanSearch;
-
+			console.log(catId, categoryFilter);
```
dyersituations, line 205: Remove console.log

dyersituations:
Bug: there appears to be some issues with different categories, where the correct FLSS filters don't appear to be passed.

For example:
1. Load page
2. Change category dropdown to "Covid-19"
3. Look at network and notice that the filter object is empty and the loan count didn't change

There are other categories that don't change the results when selected.

src/components/LoanCards/LoanProgressGroup.vue
```diff
@@ -3,7 +3,13 @@
 		<h4 class="tw-mb-0.5">
 			{{ fundingText }}
 		</h4>
+		<fundraising-status-meter
```
dyersituations, line 6: What is the difference between `fundraising-status-meter` and `kv-progress-bar`? The loan cards look the same between this branch and dev on `lending-home`. If the components are basically the same, could the meter just replace the bar here? That would be ideal, since we could remove the new prop on the shared loan card component.

egoroza: They are slightly different, `FundraisingStatusMeter` has a darker grey background color while `KvProgressBar` has a lighter one (it matches the background color of the page). The Loan Card we are using is technically not supposed to be used anywhere else besides the homepage which has a background color that matches so we are doing some wonky stuff 😅 But when we revisit Loan Cards (after MVP) I think there is some room for tech debt/deletion/clean up

egoroza:
> looking good @egoroza! are you considering to show recommended as default option for sort by dropdown in category page? there is not a match with almost funded option for initial loans.

@roger-in-kiva Yes it should still be recommended on the category page! since `LoanChannelCategoryControl` sets the default to `Recommended` - i.e. https://www.dev.kiva.org/lend-by-category/agriculture

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: 😛

roger-in-kiva:
> > looking good @egoroza! are you considering to show recommended as default option for sort by dropdown in category page? there is not a match with almost funded option for initial loans.
> 
> @roger-in-kiva Yes it should still be recommended on the category page! since `LoanChannelCategoryControl` sets the default to `Recommended` - i.e. https://www.dev.kiva.org/lend-by-category/agriculture

Great! Is there a way to dont have an empty initial state?

src/components/LoanCards/LoanProgressGroup.vue
egoroza: (We'll also want to add this prop to the other sections of the MVP page so that it uses the correct progress bar - rfy + matched lending currently have the wrong one)

egoroza:
> Bug: there appears to be some issues with different categories, where the correct FLSS filters don't appear to be passed.
> 
> For example:
> 
> 1. Load page
> 2. Change category dropdown to "Covid-19"
> 3. Look at network and notice that the filter object was empty for `flssLoanData` and the loan count didn't change
> 
> There are other categories that don't change the results when selected.

Think I found the issue, but I think it should be just for `covid-19` and `mission-driven-orgs` categories. These categories also have a `sortBy` prop in their `flssLoanSearch` criteria which I think is messing this up maybe... will add an account for this similar to the Ending Soon category. 

Feels like there is a more efficient way to do this... the number of wonky categories we are including have gone up 🥲  we will probably want to refactor this some time soon...

dyersituations:
> Think I found the issue, but I think it should be just for `covid-19` and `mission-driven-orgs` categories. These categories also have a `sortBy` prop in their `flssLoanSearch` criteria which I think is messing this up maybe... will add an account for this similar to the Ending Soon category.
> 
> Feels like there is a more efficient way to do this... the number of wonky categories we are including have gone up 🥲 we will probably want to refactor this some time soon...

Thanks for the info! There's probably a way to handle the filtering without special casing, and it might be a good follow-up ticket to create for if/when the experiment is successful.

egoroza:
> > > looking good @egoroza! are you considering to show recommended as default option for sort by dropdown in category page? there is not a match with almost funded option for initial loans.
> > 
> > 
> > @roger-in-kiva Yes it should still be recommended on the category page! since `LoanChannelCategoryControl` sets the default to `Recommended` - i.e. https://www.dev.kiva.org/lend-by-category/agriculture
> 
> Great! Is there a way to dont have an empty initial state? Can we take advantage having two dropdowns for every case?

MVP shouldn't have an empty initial state but I see what you mean on the category page... Not entirely sure since it seems the loans get rendered on the client side (unless I'm misinterpreting what you're asking!)

src/components/LoanCards/LoanProgressGroup.vue
egoroza: I added this to the other components as well so this should be updated across the MVP page!

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -157,6 +157,9 @@ export default {
 			const params = new URLSearchParams(paramStr);
 			return `/lend/filter?${params.toString()}`;
 		},
+		// TODO: Rearchitect this at some point.
```
dyersituations, line 160: I recommend adding a ticket to the backlog with some context on what is needed for this refactor.

egoroza:
> > > > > looking good @egoroza! are you considering to show recommended as default option for sort by dropdown in category page? there is not a match with almost funded option for initial loans.
> > > > 
> > > > 
> > > > @roger-in-kiva Yes it should still be recommended on the category page! since `LoanChannelCategoryControl` sets the default to `Recommended` - i.e. https://www.dev.kiva.org/lend-by-category/agriculture
> > > 
> > > 
> > > Great! Is there a way to dont have an empty initial state? Can we take advantage having two dropdowns for every case?
> > 
> > 
> > MVP shouldn't have an empty initial state but I see what you mean on the category page... Not entirely sure since it seems the loans get rendered on the client side (unless I'm misinterpreting what you're asking!)
> 
> MVP is having an initial state for sort by dropdown that is almost funded option, but for category page is looking empty, these images could be helpful Screen Shot 2023-01-05 at 11 13 21
> 
> Screen Shot 2023-01-05 at 11 12 19

Ooh yah, my b I think I misunderstood. So we are loading loans CSR on Quick Filters / MVP page, I think it is more performant, the category pages load them on the server side first with some funky magic I haven't *fully* comprehended yet. But I think all sections of the page on MVP are like this, I think it would make more sense to potentially load RFY in as server side if we *wanted* to have some SSR loans since it's the first on the page, but it might makes sense to do that after the experiment imo!

dyersituations (approved):

PR #4508, opened 2023-01-06
Core 549 cleaning up one time subscription variations in mg

dyersituations:
https://kiva.atlassian.net/browse/CORE-549

- Redirected /covid19response to /monthlygood (and deleted code for page at /covid19response)
- Removed /monthlygood/setup `onetime` query param
- Removed /monthlygood/thanks `onetime` query param
- Note that I left some code for `onetime` in components that interact with the BE, in case that param was required by the BE endpoints - the `onetime` in those cases are either hard-coded false, or use the fallback `false` for the component prop
- Fixed issue with redirect on /monthlygood/thanks - previously it would encounter a server error and crash the site locally when trying to redirect


src/pages/MonthlyGood/MonthlyGoodThanksPage.vue
```diff
@@ -81,34 +71,39 @@ export default {
...
+					// Return user to /monthlygood
+					return Promise.reject({ path: '/monthlygood' });
+				}
+
+				this.autoDepositAmount = numeral(_get(data, 'my.autoDeposit.amount', 0)).format('0.00');
+				this.donation = numeral(_get(data, 'my.autoDeposit.donateAmount', 0)).format('0.00');
+				this.mgAmount = this.autoDepositAmount - this.donation;
+				this.dayOfMonth = _get(data, 'my.autoDeposit.dayOfMonth');
+				this.category = _get(data, 'my.monthlyGoodCategory');
+				this.autoDepositId = _get(data, 'my.autoDeposit.id');
```
mcstover, line 97: Generally, `this` isn't available in the then block of the preFetch function. You'll need to either keep the `result` block below or read the query in created and then set your data vars... ref. ContentfulPage, AutolendSettingsPage + CheckoutPage vue file.

dyersituations: Oh right, good call. I was testing the data being returned but not being set.

eddieferrer:
Great clean up 🧹 !

src/pages/MonthlyGood/MonthlyGoodThanksPage.vue
dyersituations: I went back and did some testing, the`this.` settings worked (what I saw), but `this` here is the `preFetch` function itself! That's not useful... Haha. Anyways, I backtracked to the original error and realized that the error could be resolved with just adding a `.catch` to the router push, since the push still goes through if you quiet the unimportant error it was throwing.

eddieferrer (approved):

roger-in-kiva (approved):
looking good!

mcstover (approved):

PR #4509, opened 2023-01-06
fix: fix issue with logo alignment in cc pages

michelleinez:


src/components/CorporateCampaign/CampaignLogoGroup.vue
```diff
@@ -31,6 +31,16 @@ export default {
 		corporateLogoUrl: {
 			type: String,
 			default: ''
+		},
+		logoClasses: {
+			type: String,
+			default: '',
+			required: false
```
dyersituations, line 38: Doesn't necessarily need to be changed but `required: false` shouldn't be needed on these props

dyersituations (approved):

src/components/CorporateCampaign/CampaignLogoGroup.vue
```diff
@@ -1,16 +1,16 @@
...
 		<span
 			v-if="corporateLogoUrl"
-			class="tw-text-secondary tw-inline-block tw-mx-1 tw-text-h2"
+			class="tw-text-secondary tw-inline-block tw-mx-2 tw-text-h2"
 			aria-hidden="true"
 		>+</span>
 		<kv-contentful-img
 			v-if="corporateLogoUrl"
-			class="campaign-logo-group__corporate"
+			:class="`campaign-logo-group__corporate ${logoClasses}`"
```
mcstover, line 11: @michelleinez Were you hoping to pass in tailwind classes for use here? If so we need to make sure those classes will be available by adding them to our tailwind.safelist file. In other words we can't just pass in arbitrary classes at run time and assume they will work. They "might" work if they've been used somewhere else in the code that was present during compile, but they may be missing b/c of the tailwind purge process and not apply at all.

Let's discuss this in slack a bit more to see if there is a small set of height classes we can agree as applicable...

src/components/CorporateCampaign/CampaignLogoGroup.vue
michelleinez: I was getting an error when I didn't have this in place stating that the props wasn't defined, even with the default key present...  :/

src/components/CorporateCampaign/CampaignLogoGroup.vue
michelleinez: They would be the tw-mb-1 or tw-mt-1 classes.  There is a separate height property as well which I added that manages the height.  These classes would be used to just offset the logo upwards or downwards to get the text to line up for both logos.

mcstover: Cool, those should be added along with whatever set we determine for the height modifiers. They will both be nice levers to have to adjust the the logo!

dyersituations: Tested locally, and the classes do get passed through and applied like expected. But like @mcstover mentioned, the classes need to be in the safelist so they don't get stripped from the compiled tailwind css.

src/components/CorporateCampaign/CampaignLogoGroup.vue
dyersituations: I tried removing all of the `required` props in this changeset locally, and the app built and ran without error. Which error did you encounter?

PR #4511, opened 2023-01-06
fix: redirect to lend-classic for unhandled loan statuses, allow rais…

mcstover:
…ed status

Previous broken "raised" status:
image

Previous broken "paying_back" status, this now redirects to `/lend-classic`
image

Fixed "raised" status:
image


eddieferrer (approved):
Looks good!

dyersituations (approved):

PR #4512, opened 2023-01-06
fix: run experiment only for new users

christian14b:
Experiment should only be shown and load for new users so there's a new computed that validates if the user is a non-lender, non-depositer nor has not logged in before.

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -537,18 +537,16 @@ export default {
 		}
 
 		// Async data fetch for MARS-317 and share button
-		if (this.enabledExperiment) {
-			const { data } = await this.apollo.query({
-				query: mountedQuery,
-				variables: {
-					loanId: this.loanId,
-				},
-			});
-			const loan = data?.lend?.loan;
-			this.partnerName = loan?.partnerName ?? '';
-			this.partnerCountry = loan?.partner?.countries[0]?.name ?? '';
-			this.lender = data?.my?.userAccount ?? {};
-		}
+		const { data } = await this.apollo.query({
+			query: mountedQuery,
+			variables: {
+				loanId: this.loanId,
+			},
+		});
+		const loan = data?.lend?.loan;
+		this.partnerName = loan?.partnerName ?? '';
+		this.partnerCountry = loan?.partner?.countries[0]?.name ?? '';
+		this.lender = data?.my?.userAccount ?? {};
 	},
 	computed: {
 		loanId() {
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -182,6 +182,7 @@ const preFetchQuery = gql`
 		$imgDefaultSize: String = "w480h360",
 		$imgRetinaSize: String = "w960h720"
 	) {
+		hasEverLoggedIn @client
 		general {
 			lendUrgency: uiExperimentSetting(key: "lend_urgency") {
 				key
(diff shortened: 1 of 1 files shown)


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -531,21 +537,28 @@ export default {
...
+		}
 	},
 	computed: {
 		loanId() {
 			return Number(this.$route.params.id || 0);
 		},
+		enabledExperiment() {
+			const hasLentBefore = this.cookieStore.get(hasLentBeforeCookie);
+			const hasDepositBefore = this.cookieStore.get(hasDepositBeforeCookie);
+			return hasLentBefore !== 'true' && hasDepositBefore !== 'true' && !this.hasEverLoggedIn;
```
dyersituations, line 560: So we're only running this for the first login of a user? At what point will `hasEverLoggedIn` change from returning `true` to returning `false`? For example, will it be `true` as soon as a user is created and logged in, or is it `true` on the second login of a new user only?

mcstover: Generally, so long as cookies haven't been recently cleared, hasEverLoggedIn will return true if it can find a current login session, remnants of a login session or the kvu cookie. ref. https://github.com/kiva/ui/blob/main/src/api/localResolvers/my.js#L14

It's not "absolute" but it's fine for the purposes of experimentation. We "could" do some additional filtering in the data analysis stage to exclude existing users who may have gotten shown the exp.

dyersituations: In that case, will `this.hasEverLoggedIn` only be `false` for non-logged-in users who have never logged in via their current browser? Assuming that cookie is set to `true` as soon as a user logs in.

mcstover: That's right. Hope that's your goal @christian14b !!!

christian14b: Thank you both for comments! Yeah @mcstover, that's the expected behavior!

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -531,21 +537,28 @@ export default {
...
-			query: mountedQuery,
-			variables: {
-				loanId: this.loanId,
-			},
-		});
-		const loan = data?.lend?.loan;
-		this.partnerName = loan?.partnerName ?? '';
-		this.partnerCountry = loan?.partner?.countries[0]?.name ?? '';
-		this.lender = data?.my?.userAccount ?? {};
+		if (this.enabledExperiment) {
```
mcstover, line 540: I don't think we should gate this entire query...We need the `lender` data for the share button.

dyersituations (approved):

mcstover (approved):

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -546,6 +552,11 @@ export default {
 		loanId() {
 			return Number(this.$route.params.id || 0);
 		},
+		enabledExperiment() {
+			const hasLentBefore = this.cookieStore.get(hasLentBeforeCookie);
+			const hasDepositBefore = this.cookieStore.get(hasDepositBeforeCookie);
+			return hasLentBefore !== 'true' && hasDepositBefore !== 'true' && !this.hasEverLoggedIn;
```
eddieferrer, line 558: Are these extra cookie checks necessary? Won't `hasLentBefore` and `hasDepositBefore` always be false if `!this.hasEverLoggedIn;`

emuvente:
Couldn't this be done with user targets in the experiment config?

emuvente:
Experiment this is for is finished.

PR #4518, opened 2023-01-09
fix: updated google play icon badge to 2022 version (ui repo) [VUE-1370]

tiarascoleman:
Before: 
Screen Shot 2023-01-09 at 10 28 01 AM

After:
Screen Shot 2023-01-09 at 10 28 33 AM


mcstover:
@tiarascoleman Could you please check the package-json file out from main so no changes are introduced here?

PR #4521, opened 2023-01-10
feat: journal updates on borrower profile

eddieferrer:
ACK-470

Adds updates to borrower profile: 

Loading state: 
Screen Shot 2023-01-09 at 12 16 36 PM

Loaded: 
Screen Shot 2023-01-10 at 12 53 16 PM

Updates load 3 at a time, this section is using the same pattern of loading as the other sections of the borrower profile to ensure that we don't slow down the main query in the borrower profile. 

Additional functionality coming in future PR's. (social sharing, linking)
Related backend change: https://github.com/kiva/kiva/pull/11265 to reverse the order of the API return.

### DIFF
--- a/src/components/BorrowerProfile/JournalUpdates.vue
+++ b/src/components/BorrowerProfile/JournalUpdates.vue
@@ -1,0 +1,188 @@
+<template>
+	<section>
+		<!-- Display Content -->
+		<h2 class="tw-text-h2 tw-mb-3" :data-testid="`bp-updates-header`">
+			<!-- eslint-disable-next-line max-len -->
+			{{ sectionTitle }} <kv-loading-placeholder v-if="firstLoading" class="tw-inline-block tw-align-middle tw-mb-0.5" :style="{width: '27%', height: '28px' }" />
+		</h2>
+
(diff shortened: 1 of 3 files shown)


mcstover (approved):
Pretty Cool!

src/components/BorrowerProfile/JournalUpdates.vue
```diff
@@ -0,0 +1,188 @@
...
+				See older updates
+			</kv-button>
+		</div>
+	</section>
+</template>
+
+<script>
+import { gql } from '@apollo/client';
+import { createIntersectionObserver } from '@/util/observerUtils';
+// TODO: replace the loading placeholder with component from kv-components when available.
```
dyersituations, line 58: The loader is in kv-components.

dyersituations (approved):

src/components/BorrowerProfile/JournalUpdates.vue
eddieferrer: Nice, that was just some copy pasta :) I can open a PR to replace all these

PR #4522, opened 2023-01-10
refactor: deprecate KvLoadingPlaceholder in favor of kv-elements version

eddieferrer:


dyersituations:
@eddieferrer Do we know for sure that the shared component behaves exactly the same in all cases?

eddieferrer:
> @eddieferrer Do we know for sure that the shared component behaves exactly the same in all cases?

The borrower profile for sure looks the same. That includes a lot of uses of the placeholder. I didn't check other pages. Looking at the code differences and uses, I believe it should behave the same.

dyersituations (approved):

mcstover (approved):
Should be ok but I'd double check on some of the existing instances that pass static and dynamic `styles` in.

eddieferrer:
Did some more QA, pretty confident in this, going to merge it in.

PR #4523, opened 2023-01-11
fix: loanId prop type

christian14b:


eddieferrer (commented):
Thank you for fixing this, I had also noticed this. 

I don't think the default should be 0, I believe it should be `null`. Since the loanId is used directly as the last parameter of the analytics events, null will omit the value in looker, where as 0 will send the value to looker and you'll have a bunch of events with loanId '0' which could be misleading

mcstover (approved):

eddieferrer (approved):

dyersituations (approved):

PR #4528, opened 2023-01-12
feat: add read more functionality to journal entries in bp

eddieferrer:
ACK-470

### DIFF
--- a/src/components/BorrowerProfile/UpdateDetails.vue
+++ b/src/components/BorrowerProfile/UpdateDetails.vue
@@ -8,7 +8,18 @@
 		>
 		<div class="tw-prose tw-pb-2">
 			<p v-html="subject"></p>
-			<p v-html="body"></p>
+			<p v-if="truncateBody && !readMoreClicked">
+				<span v-html="shortBody"></span>
+				<button
+					class="tw-text-link tw-mt-1"
+					@click="readMoreClicked = true"
+					v-show="truncateBody && !readMoreClicked"
+					v-kv-track-event="['borrower-profile', 'click', 'more-about-loan', 'read-more']"
+				>
+					read more
+				</button>
+			</p>
+			<p v-else v-html="htmlSafeBody"></p>
 		</div>
 
 		<div>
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/UpdateDetails.vue
```diff
@@ -57,9 +69,24 @@ export default {
...
+		},
+		truncateBody() {
+			const bodyLength = this.htmlSafeBody?.length ?? 0;
+			return bodyLength > 210;
+		},
+		shortBody() {
+			return `${this.htmlSafeBody?.substring(0, 210)}... `;
+		},
+		htmlSafeBody() {
+			return DOMPurify.sanitize(this.body, { ALLOWED_TAGS: ['b', 'i', 'em', 'strong', 'a', 'p', 'br'] });
```
mcstover, line 89: Would there be H_n_ header elements in here?

src/components/BorrowerProfile/UpdateDetails.vue
```diff
@@ -57,9 +69,24 @@ export default {
...
 			return format(parseISO(this.date), 'MMMM dd, yyyy');
+		},
+		truncateBody() {
+			const bodyLength = this.htmlSafeBody?.length ?? 0;
+			return bodyLength > 210;
+		},
+		shortBody() {
+			return `${this.htmlSafeBody?.substring(0, 210)}... `;
+		},
+		htmlSafeBody() {
```
eddieferrer, line 88: Some journals entries have html, so better sanitize them here.

dyersituations (approved):
Approved pending question on header elements from @mcstover

src/components/BorrowerProfile/UpdateDetails.vue
eddieferrer: Hmm yeah, there could possibly be. I'm not sure where this html is coming from..... I'll add h tags as allowed here as well.

mcstover (approved):

PR #4529, opened 2023-01-12
refactor: clean up eco_challenge experiment and layout

eddieferrer:


src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
```diff
@@ -30,9 +22,7 @@
 </template>
 
 <script>
-import IconClimateChallenge from '@/assets/icons/inline/eco-challenge/globe-leaf.svg';
 import KivaClassicLoanCarousel from '@/components/LoanCollections/KivaClassicLoanCarousel';
-import KvPill from '@/components/Kv/KvPill';
```
dyersituations, line 35: It looks like we don't use this component elsewhere, safe to delete the component entirely?

src/pages/Lend/LoanChannelCategoryClimateExperiment.vue
```diff
@@ -63,7 +63,6 @@
 				<kiva-classic-single-category-carousel
```
dyersituations, line 63: Are we saving this component?

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -313,12 +313,6 @@
 					</div>
 				</kv-grid>
 			</transition>
-
-			<eco-challenge-lightbox
```
dyersituations, line 317: Are we saving this component?

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -341,15 +335,9 @@ import {
 	isBetween25And500
 } from '@/util/loanUtils';
 import { createIntersectionObserver } from '@/util/observerUtils';
-import {
-	getExperimentSettingCached,
-	trackExperimentVersion
-} from '@/util/experimentUtils';
-import { achievementsQuery, hasMadeAchievementsProgression } from '@/util/ecoChallengeUtils';
```
dyersituations, line 348: Same with these utils?

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
mcstover: I'd like to keep these around for now as I think we'll prob revisit when we fill out our new component lib.

mcstover (approved):
I know @eddieferrer wanted to keep around some of the utils and templates as they may be needed later. Lots of nice cleanup and simplification otherwise!

src/components/LoanCollections/KivaClassicSingleCategoryCarousel.vue
eddieferrer: Yeah these are actually quite nice looking.

eddieferrer:
@dyersituations It is unclear to me wether we will have another challenge in the future, so I left the components in here, since there is a lot of work put in those, and it would certainly be easier to repurpose them for a future challenge. I will add comments to them to make this clear

dyersituations (approved):

PR #4531, opened 2023-01-13
fix: updated order totals test ID

dyersituations:
- The total text moved on the page, so the test ID had to be removed (robot changes coming later in that repo)


mcstover (approved):
Didn't see any conflicts with the kiva-tests

PR #4533, opened 2023-01-17
fix: mobile ui updates for quick filters

egoroza:
Screen Shot 2023-01-17 at 5 49 45 PM

### DIFF
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -136,7 +136,6 @@
 </template>
 
 <script>
-import _throttle from 'lodash/throttle';
 import { mdiFilterVariant } from '@mdi/js';
 import loanChannelQueryMapMixin from '@/plugins/loan-channel-query-map';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -101,8 +101,13 @@
 		</div>
 		<div class="tw-flex tw-justify-between tw-items-start tw-mt-2" v-if="withCategories">
 			<div id="customizedSortBySelector" class="tw-text-action tw-flex tw-items-center tw-gap-1">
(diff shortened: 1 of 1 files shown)


src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -258,6 +268,9 @@ export default {
 		}
 	},
 	methods: {
+		determineIsMobile() {
```
dyersituations, line 271: We do have this `isMobile` pattern elsewhere in the code, so no need to change for this PR, but for future mobile-only UI (CSS/text), consider using `@media`. That would avoid an extra event listener.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -181,11 +184,18 @@ export default {
...
 				endingSoon: false,
 			},
+			isMobile: false,
 		};
 	},
 	mixins: [
 		loanChannelQueryMapMixin
 	],
+	mounted() {
+		window.addEventListener('resize', _throttle(() => {
```
dyersituations, line 194: The `addEventListener` requires an associated `removeEventListener` in `beforeDestroy`. In order for that to work, the method (in this case the return `_throttle`) needs to be a unique instance of the function. So I'm not 100% if our other usages of `_throttle` for checking resize will actually remove the event listener. Please ensure the event listener is removed, or consider using the CSS-only approach mentioned in the other comment.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: Yeah I thought about it but I think we needed this for conditionally rendering some elements (different text displayed on mobile), unless there is a pure css way to do this 🤔

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: It seems to be working when switching views but yeah I'll double check and copy from the other files I took this from

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
roger-in-kiva: Yeah, you can add a class `md:tw-hidden` to every element you want to show on mobile only, this will be hidden for medium sizes or larger

dyersituations (approved):

PR #4534, opened 2023-01-17
feat: use legacy for matching in lending-home

dyersituations:
https://kiva.atlassian.net/browse/CORE-936

- Created temp lend query to use for the matching row


roger-in-kiva (approved):
👍

PR #4536, opened 2023-01-18
fix: journal updates in bp should use correctly sized image

eddieferrer:
ACK-508

### DIFF
--- a/src/components/BorrowerProfile/JournalUpdates.vue
+++ b/src/components/BorrowerProfile/JournalUpdates.vue
@@ -74,7 +74,7 @@ const updatesQuery = gql`query updatesQuery($loanId: Int!, $limit: Int, $offset:
 					date
 					image {
 						id
-						url
+						url(customSize: "h200w700")
 					}
 				}
 			}


dyersituations (approved):

src/components/BorrowerProfile/JournalUpdates.vue
```diff
@@ -74,7 +74,7 @@ const updatesQuery = gql`query updatesQuery($loanId: Int!, $limit: Int, $offset:
 					date
 					image {
 						id
-						url
+						url(customSize: "h200w700")
```
mcstover, line 77: This seems like a really specialized dimension ration...does the upload interface instruct how to optimize for this? Also, if this image were vertical instead of horizontal how would it look? Smooshed or cropped? Cropped to the top or the center?

eddieferrer: No, I pulled this dimension from the figma it is the ideal dimension. In reality, the updates that have images (not the majority of them) have image sizes all over the place. For images that do not have these dimensions, our API is smart enough to return a smaller image. Usually, for small portrait images, the returned image will have a height of 200 and whatever its natural width is.

mcstover (approved):

src/components/BorrowerProfile/JournalUpdates.vue
eddieferrer: This is an example of an update with a large square image, asking for "h200w700" the api will return the 200x200 image, and it looks like: 

Screen Shot 2023-01-18 at 11 04 02 AM

PR #4539, opened 2023-01-18
feat: modify sharing events

eddieferrer:
* Add new action 'share' and new category 'post-checkout'
* modify sharing events to use new category and action
* add utmCampaign to sharing events

ACK-510

Action and new category added in snowplow repo.

### DIFF
--- a/src/components/BorrowerProfile/UpdateDetails.vue
+++ b/src/components/BorrowerProfile/UpdateDetails.vue
@@ -30,7 +30,7 @@
 						:share-message="shareMessage"
 						:share-url="$route.path + '#updates'"
 						variant="link"
-						utm-campaign="bp_update_share"
+						utm-campaign="social_share_update"
 					>
 						Share this update
 					</kv-social-share-button>
--- a/src/components/Checkout/SocialShareV2.vue
+++ b/src/components/Checkout/SocialShareV2.vue
@@ -13,7 +13,12 @@
 				data-testid="share-facebook-button"
 				class="social__btn social__btn--facebook"
 				:href="facebookShareUrl"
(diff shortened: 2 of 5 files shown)


src/components/BorrowerProfile/UpdateDetails.vue
```diff
@@ -30,7 +30,7 @@
 						:share-message="shareMessage"
 						:share-url="$route.path + '#updates'"
 						variant="link"
-						utm-campaign="bp_update_share"
```
eddieferrer, line 33: All our other utm-campaigns are "social_share_<something>" so just modifying this one

src/components/Checkout/SocialShareV2.vue
```diff
@@ -178,12 +198,7 @@ export default {
 					if (code !== '4201') {
 						this.$showTipMsg(`There was a problem sharing to Facebook: ${message}`, 'warning');
 					}
-					this.$kvTrackEvent(
```
eddieferrer, line 181: I'm pretty confident nobody has ever looked at this event. Simplifying it and making it the same across the files, but I'm not what useful data we can know from sending an event on this failure. If anything I think this should be a Sentry log event.

src/components/Kv/KvSocialShareButton.vue
```diff
@@ -5,9 +5,10 @@
 			data-testid="bp-share-cta-button"
 			@click="isLightboxVisible = true;"
 			v-kv-track-event="[
-				'Lending',
-				'click-share-cta',
-				'Share'
+				trackingCategory,
```
eddieferrer, line 8: This is the event for clicking the button which shows the lightbox

dyersituations (approved):

PR #4540, opened 2023-01-19
Core 939 add loan labels to mvp

dyersituations:
https://kiva.atlassian.net/browse/CORE-939

- The lending-home page uses 3 (the grid component is on the category page) different loan card components
- All 4 cards now have an option prop (default `false`) for adding loan tag pills to the top of the card
- Consolidated the loan tag logic (other than `amountLeft`) into the tag component itself
- Once we finalize the loan card in the next body of work, we can clean up the loan card tag usage
- Updated matching text to reddish color (see screenshots)

3 cards on lending-home (1 has two screenshots, since it's used twice):

image

image

image

image

### DIFF
--- a/src/components/LoanCards/LoanMatchingText.vue
+++ b/src/components/LoanCards/LoanMatchingText.vue
@@ -19,7 +19,7 @@
 				:style="{ padding: '4px 0 2px 0', height: '1.2rem'}"
 			>
 		</span>
-		<h4 class="tw-inline-block" style="color: #A24536;">
+		<h4 class="tw-inline-block">
 			{{ constructedMatchingText }}
 		</h4>
 	</span>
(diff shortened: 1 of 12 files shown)


src/components/LoanCards/LoanMatchingText.vue
```diff
@@ -19,7 +19,7 @@
 				:style="{ padding: '4px 0 2px 0', height: '1.2rem'}"
 			>
 		</span>
-		<h4 class="tw-inline-block">
+		<h4 class="tw-inline-block" style="color: #A24536;">
```
mcstover, line 22: As it stands this will show up on the homepage, Managed Lending pages and any other locations that use KivaClassicBasicLoanCard. I don't think this change was meant to go site wide was it...

dyersituations: I checked, and they were OK with the propagation. But I'll ask again just in case.

mcstover (approved):

src/components/LoanCards/LoanMatchingText.vue
dyersituations: Team decided to revert the color change for now.

PR #4541, opened 2023-01-19
Core 953 recommended loan cards change size after scrolling

dyersituations:
https://kiva.atlassian.net/browse/CORE-958
https://kiva.atlassian.net/browse/CORE-967
https://kiva.atlassian.net/browse/CORE-953
https://kiva.atlassian.net/browse/CORE-965

- Scrolling issues resolved
- Medium browser window now also uses "large" location selector

### DIFF
--- a/src/components/LoanCollections/KivaClassicLoanCarouselExp.vue
+++ b/src/components/LoanCollections/KivaClassicLoanCarouselExp.vue
@@ -11,7 +11,7 @@
 
 		<kv-carousel
 			v-if="loanIds.length > 0 && isVisible"
-			class="tw-w-full tw-overflow-visible md:tw-overflow-hidden"
+			class="tw-w-full tw-overflow-hidden"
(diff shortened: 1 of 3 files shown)


roger-in-kiva (approved):
I tested this and working fine!
Should be useful to add tailwind config file and get the breakpoints from there instead of having hardcoded ones like this `document.documentElement.clientWidth < 734`?
Or having a mixin to get tailwind breakpoints from config file?

dyersituations:
> ...mixin to get tailwind breakpoints from config

Yeah I was hoping to not have to hardcode, but I hadn't previously seen us use a pattern of not hardcoding. You inspired me to look again, and I found one usage (which makes a lot of sense now that I see it), so I updated to use the primitives directly:

```
document.documentElement.clientWidth < kvTokensPrimitives.breakpoints.md
```

PR #4543, opened 2023-01-19
fix: resolved issues with default sorts

dyersituations:
https://kiva.atlassian.net/browse/CORE-970
https://kiva.atlassian.net/browse/CORE-960

- Fixed issues with default quick filter sort

### DIFF
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -146,7 +146,7 @@ export default {
 		},
 		filterOptions: {
 			type: Object,
-			default: () => {}
+			default: () => ({})
 		},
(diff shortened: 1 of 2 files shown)


src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -328,7 +332,6 @@ export default {
 </script>
 
 <style lang="postcss" scoped>
```
roger-in-kiva, line 334: is postcss still required? affecting some performance if added?

dyersituations: Postcss should only affect the compilation, so I don't think we have to worry about removing.

roger-in-kiva (approved):

PR #4547, opened 2023-01-23
feat: personalized sharing modal

eddieferrer:
* Make Bp sharing modal different and editable for borrowers vs lenders

ACK-462

Screen Shot 2023-01-23 at 8 54 31 AM
Screen Shot 2023-01-23 at 8 54 16 AM

Reference links for linkedin api change: https://github.com/bradvin/social-share-urls#-linkedin

### DIFF
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -20,7 +20,7 @@ function missingLoanField({
 	// Use optional chaining to see if nested loan properties exist
 	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? null, loan);
 	if (value === null) {
-		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'error', { loan });
+		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'info', { loan });
 		return true;
 	}
 }
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -20,7 +20,7 @@ function missingLoanField({
 	// Use optional chaining to see if nested loan properties exist
 	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? null, loan);
 	if (value === null) {
-		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'info', { loan });
+		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'error', { loan });
 		return true;
 	}
 }
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -18,8 +18,8 @@ function missingLoanField({
 }) {
 	const propChain = field.split('.');
 	// Use optional chaining to see if nested loan properties exist
-	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? null, loan);
-	if (value === null) {
+	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? undefined, loan);
+	if (value === undefined) {
 		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'error', { loan });
 		return true;
 	}
--- a/src/api/localResolvers/loan.js
+++ b/src/api/localResolvers/loan.js
@@ -18,7 +18,7 @@ function missingLoanField({
 }) {
 	const propChain = field.split('.');
 	// Use optional chaining to see if nested loan properties exist
-	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? undefined, loan);
+	const value = propChain.reduce((val, chainLink) => val?.[chainLink], loan);
 	if (value === undefined) {
 		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'error', { loan });
 		return true;
--- a/src/components/BorrowerProfile/ShareButton.vue
+++ b/src/components/BorrowerProfile/ShareButton.vue
@@ -1,30 +1,51 @@
 <template>
 	<kv-social-share-button
 		:modal-title="modalTitle"
-		:share-message="shareMessage"
+		:share-message="modifiedShareMessage"
 		:share-url="shareLink"
 		variant="caution"
 		:utm-campaign="utmCampaign"
 		:utm-content="utmContent"
-		:linked-in-title="linkedInTitle"
 		:open-lightbox="forceLightbox"
 		:loan-id="loan.id"
 	>
 		<template #modal-content>
-			<!-- eslint-disable-next-line max-len -->
-			<p>You can make change happen faster for {{ name }} by getting the word out. Each lender that supports {{ name }} brings them one step closer to being live for all to see on Kiva.org. Share their loan now.</p>
+			<div class="tw-relative">
+				<textarea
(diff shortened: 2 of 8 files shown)


src/components/Checkout/SocialShareV2.vue
```diff
@@ -168,11 +168,7 @@ export default {
 			});
 		},
 		linkedInShareUrl() {
-			return getFullUrl('https://www.linkedin.com/shareArticle', {
-				mini: 'true',
-				source: `https://${this.$appConfig.host}`,
-				summary: this.shareMessage.substring(0, 256),
-				title: `A loan for ${this.selectedLoan.name}`,
+			return getFullUrl('https://www.linkedin.com/sharing/share-offsite/', {
```
eddieferrer, line 171: The linked in sharing API has changed. It no longer supports any of these additional query params, only the url param of the page to be shared. The title and description are now pulled from the meta `og` tags.

src/components/Kv/KvSocialShareButton.vue
```diff
@@ -21,55 +21,46 @@
...
-					rel="noopener"
 					v-kv-track-event="[trackingCategory, 'share', 'facebook', utmCampaign, loanId]"
-					@click="$showTipMsg('Thanks for sharing to Facebook!')"
+					@click="showSharePopUp(facebookShareUrl, 'Thanks for sharing to Facebook!')"
 				>
 					<kv-material-icon
 						class="social-button__icon social-button__icon--facebook"
 						:icon="mdiFacebook"
 					/>
-					<span class="tw-font-medium">Share on Facebook</span>
```
eddieferrer, line 40: Removed "Share on" and modified css to allow or the 4 buttons to fit on one line, see screenshot

src/components/Kv/KvSocialShareButton.vue
```diff
@@ -227,20 +212,16 @@ export default {
...
+			return getFullUrl('https://www.linkedin.com/sharing/share-offsite/', {
+				url: `${this.shareLink}utm_source=linkedin.com&utm_medium=social${this.utmCampaignQueryParam}${this.utmContentQueryParam}${this.hashParams}` // eslint-disable-line max-len
 			});
 		},
 		twitterShareUrl() {
 			return getFullUrl('https://twitter.com/intent/tweet', {
 				text: this.shareMessage,
 				// eslint-disable-next-line max-len
 				url: `${this.shareLink}utm_source=t.co&utm_medium=social${this.utmCampaignQueryParam}${this.utmContentQueryParam}${this.hashParams}`,
+				hashtags: 'microloan,kivalove',
```
eddieferrer, line 224: These will show up as hashtags in twitter share

src/components/Kv/KvSocialShareButton.vue
```diff
@@ -284,6 +265,27 @@ export default {
 					};
 				}, 500);
 			}
+		},
+		/** displays the share pop up window for whatever service we are sharing on.  */
+		showSharePopUp(destinationUrl, thanksText) {
```
eddieferrer, line 270: This code is mostly from the twitter docs. It just opens a new pop up window with the share interface

src/components/Thanks/ThanksPageShare.vue
```diff
@@ -324,25 +324,7 @@ export default {
...
-					title += 'expand access to education around the world?';
-				}
-				title += 'support smallholder farmers around the world?';
-			}
-			return getFullUrl('https://www.linkedin.com/shareArticle', {
-				mini: 'true',
-				source: `https://${this.$appConfig.host}`,
-				summary: this.shareMessage.substring(0, 256),
-				title,
+			return getFullUrl('https://www.linkedin.com/sharing/share-offsite/', {
```
eddieferrer, line 327: This API has changed, see my previous comment

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -227,8 +227,10 @@ const preFetchQuery = gql`
 				loanAmount
 				status
 				use
+				fundraisingPercent @client
```
eddieferrer, line 230: We need these to calculate the percent for the share message. I don't think this will impact performance on this query

src/components/BorrowerProfile/ShareButton.vue
```diff
@@ -39,12 +60,55 @@ export default {
...
+			return `$${this.loan?.loanAmount - this.loan?.loanFundraisingInfo?.fundedAmount}`;
+		},
+		isBorrower() {
+			// If the loan id's match or the share query param is present, the user is the borrower
+			return this.loan?.id === this.borrowedLoanId || this.$route.query.share === 'true';
 		},
 		modalTitle() {
-			return `Help ${this.name} spread the word.`;
+			if (this.isBorrower) {
+				switch (true) {
```
dyersituations, line 96: I'm not sure what I think of the `switch (true)` syntax over if/else, but it's technically fine.

dyersituations (approved):

src/pages/BorrowerProfile/BorrowerProfile.vue
mcstover: I think this may re-introduce some excessive logging that we recently cleaned up. This local query has data requirements that oddly enough, aren't available on all loans...

Can you please move it out of the prefetch?

mcstover: @eddieferrer What I'm suggesting is to move just the client query out into it's own process and to check for it's data dependencies before calling it please.

src/components/Checkout/SocialShareV2.vue
emuvente: +1 Found out the same thing recently.

src/api/localResolvers/loan.js
```diff
@@ -20,7 +20,7 @@ function missingLoanField({
 	// Use optional chaining to see if nested loan properties exist
 	const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? null, loan);
 	if (value === null) {
-		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'error', { loan });
+		logFormatter(`Loan resolver: Missing required field '${field}' for ${type}.${resolver}`, 'info', { loan });
```
mcstover, line 23: @emuvente This is my suggestion for now. The code safely handles this state and we know it can happen depending on the loan state.

mcstover (approved):

src/api/localResolvers/loan.js
mcstover: It may be that we can remove the log or possibly gate it to a lower env...

emuvente: We could change the check above to be for `undefined` instead of `null`. The intent of this log is to tell developers that they need to request those fields in their data fetching. But as you point out, sometimes these fields are missing, in which case the graphql api returns `null` which causes this log message to appear. If we check for `undefined` instead (and use `undefined` as the default value after the nullish operator), then this log should only occur when the fields aren't requested at all:
```js
const value = propChain.reduce((val, chainLink) => val?.[chainLink] ?? undefined, loan);
if (value === undefined) {
    ...
```

BTW when the values are `null` the `fundraisingPercent` resolver returns 0.

mcstover: I'm fine with that adjustment as well. I suppose we'd still want to return 0 in the failure case as that keeps this functioning safely.

eddieferrer: I went with this approach and changed the log to trigger on undefined instead

eddieferrer:
I added a unit test to the local resolver for the null values.

PR #4551, opened 2023-01-23
fix: use revenue tag for custom optimizely events MARS-330

emuvente:
Custom events should use the `revenue` or `value` tags for metrics  https://docs.developers.optimizely.com/web/docs/event#parameters.

Based on fix from kiva/ui#4410


mcstover (approved):

roger-in-kiva (approved):
👍

christian14b (approved):

PR #4561, opened 2023-01-26
feat: hide help me choose feat when loans are fewer than 9 in categor…

roger-in-kiva:
Help me choose feature hidden in category page when the number of loans is fewer than 9

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -63,7 +63,7 @@
 						:show-tags="enableLoanTags"
 					/>
 					<helpme-choose-wrapper
-						v-if="enableHelpmeChoose"
+						v-if="showHelpMeChooseFeat"
 						:remaining-loans="helpmeChooseRemainingLoans"
 						:items-in-basket="itemsInBasket"
 						:is-visitor="isVisitor"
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -421,6 +421,10 @@ export default {
 		},
 		hasOnePageOfLoans() {
 			return this.totalCount <= this.limit;
+		},
+		showHelpMeChooseFeat() {
+			if (this.enableHelpmeChoose && this.allLoans.length > 8) return true;
+			return false;
```
mcstover, line 427: Couldn't this just be `return this.enableHelpmeChoose && this.allLoans.length > 8;` without the if block and second return?

roger-in-kiva: oooh yes 🤯
fixed

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -421,6 +421,9 @@ export default {
 		},
 		hasOnePageOfLoans() {
 			return this.totalCount <= this.limit;
+		},
+		showHelpMeChooseFeat() {
+			return this.enableHelpmeChoose && this.allLoans.length > 8;
```
dyersituations, line 426: Should we be using `this.allLoans.length` or the `totalCount` computed prop?

roger-in-kiva: Looks like `totalCount` is for all loans in the channel. `this.allLoans.length` is counting current loans in the page.

Logging this, I have a different number for livestoc for ex.
94 with `totalCount`
12 with `this.allLoans.length`

dyersituations: Interesting that `totalCount` doesn't appear to be respecting filters. Thanks for investigating!

dyersituations (approved):

PR #4562, opened 2023-01-27
feat: add cache control headers to passport redirects

christian14b:
Add `Cache-Control` header `no-store, max-age=0, no-transform` to passportjs redirects.

`/ui-login`
image

`/process-ssr-auth`
image

### DIFF
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -20,6 +20,7 @@ module.exports = function authRouter(config = {}) {
 
 	// Helper function to start slient authentication process
 	function attemptSilentAuth(req, res, next) {
+		res.set('Cache-Control', 'no-store, max-age=0, no-transform');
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

emuvente (approved):
That looks like that should do it!

emuvente (changes requested):
Actually, could you also please add those headers to the attemptSilentAuth function and the /error and /ui-logout routes?

mcstover:
Just noting I was able to see these new Cache-Control headers applied to `/process-ssr-auth` in the vm!

mcstover:
I think this one is ready with headers added to all auth routes. @emuvente Are these the updates you requested?

emuvente (approved):
Thanks, looks great!

mcstover (approved):

PR #4564, opened 2023-01-27
Cleanup category page flss exp vue 1235

christian14b:
Remove FLSS experiment scaffolding


dyersituations (changes requested):
App changes look good, but please update the unit tests, both removing the inappropriate tests and updating any that are still needed for the remaining methods in the util file.

dyersituations (approved):

mcstover (approved):

PR #4566, opened 2023-01-30
feat: pagination reset added in category page

roger-in-kiva:
Pagination reset in category page when filters are updated

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -116,7 +116,6 @@
 					:total="totalCount"
 					:limit="limit"
 					:offset="offset"
-					:scroll-to-top="true"
 					@page-changed="pageChange"
 				/>
 				<div v-if="totalCount > 0" class="loan-count tw-text-tertiary">
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -112,7 +112,6 @@
 					/>
 				</div>
 				<kv-pagination
-					ref="categoryPagination"
 					v-if="totalCount > 0"
 					:total="totalCount"
 					:limit="limit"
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -786,7 +786,7 @@ export default {
 			this.isLoadingHC = false;
 		},
 		resetPagination() {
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -781,6 +784,9 @@ export default {
 			const loans = loansData?.lend?.loanChannelsById[0]?.loans?.values ?? [];
 			this.helpmeChooseLoans = loans;
 			this.isLoadingHC = false;
+		},
+		resetPagination() {
+			this.$refs.categoryPagination.clickPage(0);
```
dyersituations, line 789: We probably don't want to call `clickPage`, since it will trigger an analytics click event, and the user didn't click the pagination in this case. Can we change the offset in this component instead? Looks like this component has a `pageChange` method.

roger-in-kiva: Nice catch! fixed

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -112,10 +112,12 @@
 					/>
 				</div>
 				<kv-pagination
+					ref="categoryPagination"
 					v-if="totalCount > 0"
 					:total="totalCount"
 					:limit="limit"
 					:offset="offset"
+					:scroll-to-top="true"
```
dyersituations, line 120: This scroll to top doesn't seem to be working when I test locally.

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -112,10 +112,12 @@
 					/>
 				</div>
 				<kv-pagination
+					ref="categoryPagination"
```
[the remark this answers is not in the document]
roger-in-kiva: Done

src/pages/Lend/LoanChannelCategoryControl.vue
roger-in-kiva: It works to me when clicking a page number or prev/next buttons

dyersituations: Oh I was testing the lending-home, oops. Actually though I don't think we need this prop set, since the new method calls the page change event, which changes the URL, causing the page to jump to the top. So I think we can remove this line.

roger-in-kiva: makes sense, removed!

dyersituations (approved):

PR #4567, opened 2023-01-30
feat: add component for embedding a form assembly form based on id

eddieferrer:
* optional query param props
* handles dynamic height resizing
* added story

ACK-507

### DIFF
--- a/src/components/Forms/FormAssemblyEmbed.vue
+++ b/src/components/Forms/FormAssemblyEmbed.vue
@@ -1,0 +1,72 @@
+<template>
+	<div class="tw-w-full">
+		<div
+			v-if="isLoading"
+			class="tw-w-full tw-flex tw-content-center tw-justify-center"
+			style="min-height: 200px;"
+		>
+			<kv-loading-spinner class="tw-self-center" />
+		</div>
+		<iframe
+			v-show="!isLoading"
+			class="tw-mx-auto tw-w-full"
+			:src="iFrameSrc"
+			:height="iFrameHeight"
+			frameborder="0"
+		></iframe>
+	</div>
+</template>
+
+<script>
+import {
+	resizeHelper
+} from '@/lib/formassembly/iframe-resize-helper';
+import KvLoadingSpinner from '~/@kiva/kv-components/vue/KvLoadingSpinner';
+
+export default {
+	name: 'FormAssemblyEmbed',
+	components: {
+		KvLoadingSpinner
+	},
+	props: {
+		// The form assembly id for the form to embed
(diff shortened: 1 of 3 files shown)


eddieferrer:
We're going to be adding the form assembly forms in the future to a new borrower dashboard page. 

@mcstover I based this off of your component from cms-page-server. That component uses the events `fa_frame_loaded` and `fa_frame_data` but I could not find any documentation for those, so I decided to just import the resize helper js that form assembly has in its documentation for inserting a form in iframes.

src/lib/formassembly/iframe-resize-helper.js
```diff
@@ -0,0 +1,201 @@
...
+// Source: https://tfaforms.com/js/iframe_resize_helper.js
+/*
+ * Helper script for iframe-published FormAssembly.com forms.
+ * Auto-resize iframe to fit content.
+ */
+function resizeHelper() {
+    /*
+     * A backwards compatible implementation of postMessage
+     * by Josh Fraser (joshfraser.com)
+     * released aunder the Apache 2.0 license.
```
dyersituations, line 11: It seems like we're good to use code under Apache 2.0, but I thought it would be good to confirm, have we used code under this license before? I'm not seeing references in the `ui` repo.

src/components/Forms/FormAssemblyEmbed.vue
```diff
@@ -0,0 +1,72 @@
...
+		}
+	},
+	async mounted() {
+		if (!this.$isServer && typeof window !== 'undefined') {
+			// resizes the iframe vertically based on the events fired from FA form
+			resizeHelper();
+
+			window.addEventListener('message', message => {
+				// if we receive a message we can assume the form is loaded
+				if (message.data) {
```
mcstover, line 65: There are a variety of "messages" that an iframe may send so this may trigger multiple times and potentially before your iframe is actually loaded and ready...

The `fa_frame_loaded` and `fa_frame_data` are setup and sent in custom code on the form in FormAssembly and allow us to take action specifically at those stages. May or may not be useful for your use case.

src/lib/formassembly/iframe-resize-helper.js
eddieferrer: Good call out 🤔

src/components/Forms/FormAssemblyEmbed.vue
eddieferrer: I found the code that triggers those events

eddieferrer:
Going to do this in CMSPS instead

PR #4568, opened 2023-01-30
feat: updated loan card use statement

dyersituations:
https://kiva.atlassian.net/browse/CORE-974

- Created a new `LoanUse` component for the cards w/ unit tests to verify use cases

image

### DIFF
--- a/src/components/LoanCards/LoanUse.vue
+++ b/src/components/LoanCards/LoanUse.vue
@@ -1,0 +1,64 @@
+<template>
+	<p class="tw-line-clamp-4">
+		{{ loanUse }}
+	</p>
+</template>
+
+<script>
+import numeral from 'numeral';
+
+export default {
+	name: 'LoanUse',
+	props: {
+		anonymizationLevel: {
+			type: String,
+			default: 'none',
+		},
+		use: {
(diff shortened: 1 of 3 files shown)


src/components/LoanCards/LoanUse.vue
```diff
@@ -0,0 +1,64 @@
...
+			default: '',
+		},
+		loanUseMaxLength: {
+			type: Number,
+			default: 100,
+		},
+	},
+	computed: {
+		helpLanguage() {
+			if (this.status === 'fundraising' || this.status === 'inactive' || this.status === 'reviewed') {
```
roger-in-kiva, line 44: Could this be a switch?

roger-in-kiva: Oh no, is a or condition...

roger-in-kiva (approved):

src/components/LoanCards/LoanUse.vue
dyersituations: It could theoretically be a switch, but when there's only two cases, I usually just keep it an if.

mcstover (approved):

PR #4575, opened 2023-02-01
feat: update progress text when amount left is low

dyersituations:
https://kiva.atlassian.net/browse/CORE-1031

- Progress bar text is now orange when amount left is under $100
- Exclamation mark added, if time left doesn't already add it
- Unit tests added to confirm behavior

image

### DIFF
--- a/src/components/LoanCards/LoanProgressGroup.vue
+++ b/src/components/LoanCards/LoanProgressGroup.vue
@@ -1,6 +1,6 @@
 <template>
 	<figure>
-		<h4 class="tw-mb-0.5">
+		<h4 class="tw-mb-0.5" :class="{ 'progress-group-amount-low': amountLow }">
 			{{ fundingText }}
 		</h4>
 		<kv-progress-bar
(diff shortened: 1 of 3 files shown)


src/components/LoanCards/LoanProgressGroup.vue
```diff
@@ -33,15 +33,32 @@ export default {
...
+			const exclamationMark = this.amountLow && !formattedTimeLeft.includes('!') ? '!' : '';
+			return `${formattedMoneyLeft} to go${formattedTimeLeft}${exclamationMark}`;
 		}
 	}
 };
 </script>
+
+<style scoped>
+.progress-group-amount-low {
+	color: #FF5C00;
```
mcstover, line 62: What's the acccessibilty score for this color over a white bg?

roger-in-kiva (approved):

src/components/LoanCards/LoanProgressGroup.vue
dyersituations: Interestingly, I just ran Lighthouse locally, and it called out colors in the footer, not this new orange. Is there another test you'd recommend?

mcstover: Thanks for checking that! I love orange and white together but it buzzes on the page a little oddly with text...the footer stuff was known.

dyersituations: Agreed about the lightness of the color buzzing. I guess we'll see how the experiment performs.

PR #4577, opened 2023-02-01
fix: tracking was missing on new page

dyersituations:
- Turns out that we are missing "b" group tracking for the `loan_finding_page` experiment
- The existing tracking occurs in the LBC page, in `mounted`, so it never gets hit when the redirect happens


roger-in-kiva (approved):
Thanks!

mcstover (approved):

PR #4585, opened 2023-02-03
fix: put recommended at the bottom of the sort options

dyersituations:
https://kiva.atlassian.net/browse/CORE-1002

- "Recommended" sort option is now at the bottom of the list in /lend/filter

image

### DIFF
--- a/src/util/loanSearch/filters/sortOptions.js
+++ b/src/util/loanSearch/filters/sortOptions.js
@@ -66,6 +66,7 @@ export function formatSortOptions(standardSorts, flssSorts, extendFlssFilters) {
 			sortSrc: STANDARD_QUERY_TYPE,
 		};
 	}) ?? [];
+
 	const labeledFlssSorts = flssSorts?.reduce((prev, current) => {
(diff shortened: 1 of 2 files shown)


mcstover:
We have generally moved towards alphabetizing options like this. Was that discussed? This is a pretty odd biz requirement to encode here...especially seeing as it's the default option. Maybe we should go back to a dropdown which can be coupled with a descriptive statement about each sort option.

dyersituations:
> We have generally moved towards alphabetizing options like this. Was that discussed? This is a pretty odd biz requirement to encode here...especially seeing as it's the default option. Maybe we should go back to a dropdown which can be coupled with a descriptive statement about each sort option.

It was discussed as the quickest option. I can always go back and ask for more time to expand the options to allow help text.

mcstover (approved):

roger-in-kiva (approved):

dyersituations:
As a follow-up, sounds like we'll move forward with this temporary solution, with hopes to discuss further improvements to the filters on lend/filter.

PR #4587, opened 2023-02-03
fix: add strict cache-control headers to private pages

mcstover:
Set's strict cache-control headers on authenticated, user specific pages.

### DIFF
--- a/server/vue-middleware.js
+++ b/server/vue-middleware.js
@@ -73,7 +73,12 @@ module.exports = function createMiddleware({
 
 		// set html response headers
 		res.setHeader('Content-Type', 'text/html');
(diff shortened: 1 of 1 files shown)


christian14b (approved):

dyersituations (approved):

emuvente (approved):
:+1: Looks good! I could also see this being based on whether the current route has an authentication guard, so that we don't have to remember to update the regex pattern.

mcstover:
@emuvente Went back and looked at maybe using our auth notations in the route meta but passed on that for now due to the way that file requires extra processing to clearup the dynamic imports.

PR #4591, opened 2023-02-06
Core 1035 a b test lend dropdown split

dyersituations:
https://kiva.atlassian.net/browse/CORE-1035

- Set up experiment assignment and tracking for new lend menu button experiment
- Fixed issue where errors were being sent to the console for the new loan card tags while loan data was loading


mcstover (approved):
Prob want to check on how this is working for the cached version of pages as that will be coming on soonish.

PR #4593, opened 2023-02-07
feat: query params updated when refreshing category page

roger-in-kiva:
Page query parameter removed when refreshing category page

I was not able to replicate the issue mentioned here https://kiva.atlassian.net/browse/CORE-1038 looks like already fixed

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -564,8 +564,8 @@ export default {
 		// Extract query
 		this.pageQuery = _get(this.$route, 'query');
 
-		// Ensure page offset gets set before loading cached data
-		this.updateFromParams(this.pageQuery);
+		// Reset pagination when loading the page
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -564,8 +564,8 @@ export default {
 		// Extract query
 		this.pageQuery = _get(this.$route, 'query');
 
-		// Ensure page offset gets set before loading cached data
-		this.updateFromParams(this.pageQuery);
+		// Reset pagination when loading the page
+		this.resetPagination();
```
dyersituations, line 568: Since this method triggers pushing to the router, I don't think this is the complete solution. Based on a conversation in Slack, we might want to pause on this changeset, if we are later moving all filters to the query string.

dyersituations:
@roger-in-kiva Looks like we can close this PR and address the filters via query string in the future. I moved ticket to done.

PR #4594, opened 2023-02-07
Old category redirects MARS-335

emuvente:
1. Redirects in routes.js now use status 301 (permanent) instead of 302 (temporary) so that google stops indexing the old routes
2. Added redirects for old category pages


dyersituations (approved):

mcstover (approved):

mcstover:
@emuvente Saw that test failure, maybe we could omit redirect routes for the call

emuvente:
@mcstover test failure was caused by a duplicate route listing that I fixed by removing the dupe

PR #4596, opened 2023-02-07
feat: loan callouts on loan cards

egoroza:
implements loan callouts on new loan cards with the following priority tag order (only show a max of 2):
p1: category (limited only to eco-friendly, refugees + IDP, + single parents)
p2: activity 
p3: sector (if it's not the same as activity)
p4: random public tag
p5: random theme

Screen Shot 2023-02-08 at 5 51 30 PM
Screen Shot 2023-02-07 at 5 30 58 PM
Screen Shot 2023-02-07 at 5 31 13 PM

we also ignore p1 on category pages to avoid redundancy.

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -245,7 +245,7 @@ export default {
 		},
 		categoryPageName: {
 			type: String,
-			default: null,
+			default: '',
 		}
 	},
 	inject: ['apollo', 'cookieStore'],
@@ -281,8 +281,8 @@ export default {
 	computed: {
 		loanCallouts() {
 			const callouts = [];
-			const activityName = this.loan?.activity.name;
-			const sectorName = this.loan?.sector.name;
+			const activityName = this.loan?.activity.name ?? '';
+			const sectorName = this.loan?.sector.name ?? '';
 			const tags = this.loan?.tags.filter(tag => tag.charAt(0) === '#')
 				.map(tag => tag.substring(1)) ?? [];
 			const themes = this.loan?.themes ?? [];
@@ -305,21 +305,28 @@ export default {
 			}
 
 			// P2 Activity
-			if (this.categoryPageName !== activityName) {
+			if (this.categoryPageName.toUpperCase() !== activityName.toUpperCase()) {
 				callouts.push(activityName);
 			}
 
 			// P3 Sector
 			if (sectorName
-			&& (activityName !== sectorName)
-			&& (sectorName !== this.categoryPageName)
+			&& (activityName.toUpperCase() !== sectorName.toUpperCase())
+			&& (sectorName.toUpperCase() !== this.categoryPageName.toUpperCase())
 			&& callouts.length < 2) {
 				callouts.push(sectorName);
 			}
 
 			// P4 Tag
-			if (tags && callouts.length < 2) {
-				callouts.push(tags[0]);
+			if (!!tags.length && callouts.length < 2) {
+				const position = Math.floor(Math.random() * tags.length);
+				callouts.push(tags[position]);
+			}
+
+			// P5 Tag
+			if (!!themes.length && callouts.length < 2) {
+				const position = Math.floor(Math.random() * themes.length);
+				callouts.push(themes[position]);
 			}
 
 			return callouts;
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -325,6 +325,10 @@ export default {
 			type: Boolean,
 			default: false
 		},
+		categoryPageName: {
+			type: String,
+			default: null,
+		}
 	},
 	inject: ['apollo', 'cookieStore'],
 	mixins: [percentRaisedMixin, timeLeftMixin, loanChannelQueryMapMixin],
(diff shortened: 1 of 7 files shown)


src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -239,6 +243,10 @@ export default {
 			type: Number,
 			default: 3
 		},
+		categoryPageName: {
```
mcstover, line 246: I think you'll need to add this prop to LoanCardController as well and then pass it on through to the `component` therein.

dyersituations: This component is being used outside the controller, and the callouts aren't being added to the old cards, so I don't think the `LoanCardController` needs to be changed.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -270,6 +279,51 @@ export default {
...
+					callouts.push('Eco-friendly');
+				} else if (categories.refugeesIdps) {
+					callouts.push('Refugees and IDPs');
+				} else if (categories.singleParents) {
+					callouts.push('Single Parent');
+				}
+			}
+
+			// P2 Activity
+			if (this.categoryPageName !== activityName) {
```
dyersituations, line 308: All of these string compares should probably have `.toUpperCase` to ensure casing differences don't cause issues.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -270,6 +279,51 @@ export default {
...
+			// P3 Sector
+			if (sectorName
+			&& (activityName !== sectorName)
+			&& (sectorName !== this.categoryPageName)
+			&& callouts.length < 2) {
+				callouts.push(sectorName);
+			}
+
+			// P4 Tag
+			if (tags && callouts.length < 2) {
```
dyersituations, line 321: `!!tags.length` needs to be checked here, otherwise loans without tags will have empty pills:
image

src/components/LoanCards/LoanTags/LoanCallouts.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+	<div class="tw-line-clamp-1">
```
dyersituations, line 2: It looks like clamping is just removing the second pill if there's wrap. My guess is that we instead want to truncate the pills themselves or something along those lines.
image

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -270,6 +279,51 @@ export default {
...
+				callouts.push(tags[0]);
```
dyersituations, line 322: Before pushing tags, we also need to check against the category, otherwise a tag that's actually the category will be pushed here.
image

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -270,6 +279,51 @@ export default {
 		};
 	},
 	computed: {
+		loanCallouts() {
```
dyersituations, line 282: Please create a follow-up ticket that turns this into a util and creates unit tests against it. We'll need to unit test all of the cases we expect, such the ones in the comments mentioned on this PR.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
dyersituations: The ticket also calls out pulling the tag by random, instead of the first.

egoroza: ooh did not even know that categories and tags can be the same, so much intersection!! :')

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
roger-in-kiva: This was turned into a util.
Here is the ticket for unit testing https://kiva.atlassian.net/browse/CORE-1080

src/components/LoanCards/LoanTags/LoanCallouts.vue
```diff
@@ -0,0 +1,45 @@
...
+				tw-whitespace-nowrap
+				tw-w-full
+				tw-h-4
+				"
+		>
+			<template v-for="tag in callouts">
+				<span
+					:key="tag"
+					class="
+                callout
```
dyersituations, line 17: Looks like this class can be removed?

src/components/LoanCards/LoanTags/LoanCallouts.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+	<div class="tw-overflow-hidden" style="width: inherit;">
```
dyersituations, line 2: Long names truncate in the pill, which is good, but If there are two long pill names, the second will no longer be a pill. Here's an extreme version. This could be a follow-up ticket.

image

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -62,6 +62,7 @@
 							:show-action-button="true"
 							:show-tags="enableLoanTags"
 							:in-grid="true"
+							:category-page-name="loanChannelName"
```
dyersituations, line 65: This needs to be passed in `HelpmeChooseWrapper` as well, otherwise you get a tag that matches the categories, like in Agriculture

src/components/LoanCards/LoanTags/LoanCallouts.vue
dyersituations: Callouts are now 50% width max w/ ellipsis

dyersituations (approved):

PR #4597, opened 2023-02-08
feat: new loan card exp setup added to lending home page

roger-in-kiva:
New loan card experiment setup added to lending home page

### DIFF
--- a/src/components/LoanFinding/PartnerSpotlightSection.vue
+++ b/src/components/LoanFinding/PartnerSpotlightSection.vue
@@ -7,7 +7,7 @@
 		<lending-category-section
 			title="Support Borrowers from Fundación Pro Mujer"
 			:loans="mfiRecommendationsLoans"
-			enable-loan-card-exp="enableLoanCardExp"
+			:enable-loan-card-exp="enableLoanCardExp"
 		/>
 
 		<div class="tw-flex tw-justify-center tw-mt-4">
--- a/src/components/LoanFinding/PartnerSpotlightSection.vue
+++ b/src/components/LoanFinding/PartnerSpotlightSection.vue
@@ -7,6 +7,7 @@
 		<lending-category-section
 			title="Support Borrowers from Fundación Pro Mujer"
 			:loans="mfiRecommendationsLoans"
+			enable-loan-card-exp="enableLoanCardExp"
 		/>
 
 		<div class="tw-flex tw-justify-center tw-mt-4">
@@ -33,6 +34,12 @@ export default {
 		LendingCategorySection
 	},
 	inject: ['apollo', 'cookieStore'],
+	props: {
+		enableLoanCardExp: {
+			type: Boolean,
+			default: false
+		}
+	},
 	data() {
 		return {
 			mfiRecommendationsLoans: [],
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -75,6 +75,7 @@ import KvToast from '~/@kiva/kv-components/vue/KvToast';
 import KvLightbox from '~/@kiva/kv-components/vue/KvLightbox';
 
 const EXP_KEY = 'loan_finding_page';
+const LOAN_CARD_EXP_KEY = 'lh_new_loan_card';
 
 export default {
 	name: 'LoanFinding',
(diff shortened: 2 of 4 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -109,7 +110,10 @@ export default {
 			return client.query({
 				query: userInfoQuery,
 			}).then(() => {
-				return client.query({ query: experimentQuery, variables: { id: EXP_KEY } });
+				return client.query(
+					{ query: experimentQuery, variables: { id: EXP_KEY } },
+					{ query: experimentQuery, variables: { id: LOAN_CARD_EXP_KEY } }
```
dyersituations, line 115: I don't think that this is needed with the new `active_experiments` process.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -185,6 +189,17 @@ export default {
 				'EXP-CORE-854-Dec2022'
 			);
 		}
+
+		const loanCardExpData = getExperimentSettingCached(this.apollo, LOAN_CARD_EXP_KEY);
```
dyersituations, line 193: This wasn't called out in the ticket, but we also need to use this experiment to swap cards. If you want this to be a separate ticket, please create, but we don't currently have a separate ticket for using this new experiment.

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: Both of them?

mcstover: I also wondered if it was needed, it isn't for assignment, but may be to ensure the readQuery can get the setting...

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: I can include on this.
Adding an enable param and passing to children to swap cards for testing group, right?

src/pages/LoanFinding/LoanFinding.vue
dyersituations: Actually I just did this in my locally header-exp code, this assignment is ran in the ancestor page component, and I didn't need to run it in the header component.

src/pages/LoanFinding/LoanFinding.vue
dyersituations: Yup, similar to the swapping we do in LBC based on the other exp flag. It will mean pulling more data in `LoanFinding.vue`, since the controller-based loan card needs all of the data, not just ID. If this effort ends up taking more time, we could instead focus on finishing the new cards, but one of these efforts (experiment or new loan card) needs to be completed this week. We used to use the controller-based cards in lending-home, so at the very least we have code in git history to check out for where we pulled the extra loan data before we introduced the new loan cards recently.

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: Yeah, Im getting the experiment data and cookie without this, agree removing both?

dyersituations: Yup, just verify in dev (removing cookie + getting assignment) once it gets merged, but the behavior should work the same locally.

dyersituations: We currently assign all active experiments in `experimentPreFetch.js`.

roger-in-kiva: ok, removed

src/pages/LoanFinding/LoanFinding.vue
roger-in-kiva: Section were not using card controller, is an easy fix now.

dyersituations: Oh right, we were using the old classic card, awesome!

src/components/LoanFinding/PartnerSpotlightSection.vue
```diff
@@ -7,6 +7,7 @@
 		<lending-category-section
 			title="Support Borrowers from Fundación Pro Mujer"
 			:loans="mfiRecommendationsLoans"
+			enable-loan-card-exp="enableLoanCardExp"
```
dyersituations, line 10: This needs to be `:enable-loan-card-exp`. There's currently an error in the console for this.

roger-in-kiva: fixed, ty

dyersituations (approved):

PR #4601, opened 2023-02-09
feat: updated loan use for direct loans

dyersituations:
https://kiva.atlassian.net/browse/CORE-1069

- Changed loan use statement string building for direct loans, since they have a more free-form loan data entry

image


roger-in-kiva (approved):
good

PR #4603, opened 2023-02-10
feat: new loand card styles updated

roger-in-kiva:
Style and height updated in new loan card to follow prototype aligments

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -1,8 +1,9 @@
 <template>
 	<div
 		:id="`${loanId}-loan-card`"
-		class="tw-flex tw-flex-col"
-		:class="{ 'loan-card-in-grid tw-px-2 tw-mb-4': inGrid, 'loan-card-active-hover': !allSharesReserved }"
+		class="tw-flex tw-flex-col tw-p-1 tw-bg-white tw-rounded"
+		:class="{ 'loan-card-in-grid tw-px-1': inGrid, 'loan-card-active-hover': !allSharesReserved }"
+		style="box-shadow: 0 4px 12px rgba(0, 0, 0, 0.08);"
 		:style="{ ...(!inGrid && { minWidth: '230px', maxWidth: cardWidth }) }"
 	>
 		<!-- Borrower image -->
@@ -50,7 +51,7 @@
 						:country-name="countryName"
 					>
 						<kv-material-icon
-							class="tw-h-2.5 tw-w-2.5 tw-mr-0.5"
+							class="tw-h-2 tw-w-2 tw-mr-0.5"
 							:icon="mdiMapMarker"
 						/>
 						{{ formattedLocation }}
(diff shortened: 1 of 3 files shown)


dyersituations:
Make sure the bottom padding on both the cards and MG card match the top padding on the loan cards. I tried to circle in the screenshot:

image

dyersituations:
Location pills are taller and grayer than the mocks:
image

dyersituations:
Loan callout background is darker than mocks:
image

dyersituations:
Mocks have black instead of green progress text and no upper case:
image

dyersituations:
It's really subtle, but the loan card images should have the black background removed. I don't know if it was used for a loading state or fallback, but the `tw-bg-black` creates a tiny black border on the image.
image

dyersituations:
In narrow (but not quite mobile) browser widths, 3 cards in a row are too many, as it cramps the contents. The previous grid only had 2 cards in a row, which is what we need with the new cards too.
image

dyersituations:
In mobile, the cards should either be centered (like they are with the old cards) or full width, instead of left aligned and not full width like the new cards are currently.

image

mcstover:
> It's really subtle, but the loan card images should have the black background removed. I don't know if it was used for a loading state or fallback, but the `tw-bg-black` creates a tiny black border on the image.
> 
> image
> 
> 

Is this a global class? We may need to keep it in place to cover image presentations that end up letterboxed.

dyersituations:
Having only 2 card rows in medium browsers also affects the old loan cards.
image

dyersituations:
Old loan cards MG card is too tall:
image

dyersituations:
Old loan card narrow view also no longer centered:
image

dyersituations:
Old loan cards lose gap as the window gets more narrow:
image

dyersituations:
> Is this a global class? We may need to keep it in place to cover image presentations that end up letterboxed.

@mcstover Not sure, I'm just functionally testing at the moment, but if it's used for letterbox, no worries.

dyersituations:
Narrow screens with old loan cards also gets a white background strip:
image

dyersituations:
The height of the cards jump as the carousel pages are changed. They start smaller, then jump to the height of the carousel.

dyersituations:
Side shadows of the card in carousel get cut off in smaller screens:

image

dyersituations:
The initial loading state of the new card is missing the border/shadow, and the border/shadow pops in after a second or two.
image

dyersituations:
> Is this a global class? We may need to keep it in place to cover image presentations that end up letterboxed.

Apparently this is a known browser issue, anti-aliasing with border-radius. We could re-structure the element to fix this, but I don't think it's worth the effort at the moment. We can wait until it gets potentially reported in the future.

dyersituations:
@roger-in-kiva @egoroza @mcstover This PR is ready for another review. The new loan cards work well in the category pages, and behavior of the old loan cards in the category pages is solid again. There will be some work to be done to resolve the remaining unresolved issues in this PR (see screenshots still visible), but I'll create follow-up tickets targeting lending-home. Thanks!

PR #4604, opened 2023-02-11
fix: add private and no-cache headers for auth routes, add more priva…

mcstover:
…te routes

To better guard against request collapsing and ensure a cache pass.
https://developer.fastly.com/learning/concepts/request-collapsing/#hit-for-pass
- Request collapsing could cause concurrent requests to our servers as requests are queued as they wait on origin
    - This can extend the page load time as similar/matching requests are held until the first request's cacheability is determined. If it's determined non-cacheable it will go to origin and the other requests will be let through. However, if there are still a bunch of similar/matching requests they may end up in the queue again.
- To solve for this, we should signal requests that should not be cached with the private flag OR use custom VCL to force pass
    - https://developer.fastly.com/learning/concepts/request-collapsing/#uncacheable-and-private-responses
        - Private flag disables request collapsing

### DIFF
--- a/server/vue-middleware.js
+++ b/server/vue-middleware.js
@@ -74,7 +74,8 @@ module.exports = function createMiddleware({
 		// set html response headers
 		res.setHeader('Content-Type', 'text/html');
 		// Set strict cache-control headers for protected pages
-		if (req?.url?.match(/(settings|portfolio|lend\/saved-search)/g)?.length) {
+		// eslint-disable-next-line max-len
+		if (req?.url?.match(/(checkout|settings|portfolio|lend\/saved-search|monthlygood\/thanks|process-browser-auth|register|start-verification|confirm-instant-donation|instant-donation-thanks)/g)?.length) {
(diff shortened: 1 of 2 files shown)


emuvente (approved):

christian14b (approved):

server/vue-middleware.js
```diff
@@ -74,7 +74,8 @@ module.exports = function createMiddleware({
 		// set html response headers
 		res.setHeader('Content-Type', 'text/html');
 		// Set strict cache-control headers for protected pages
-		if (req?.url?.match(/(settings|portfolio|lend\/saved-search)/g)?.length) {
+		// eslint-disable-next-line max-len
+		if (req?.url?.match(/(checkout|settings|portfolio|lend\/saved-search|monthlygood\/thanks|process-browser-auth|register|start-verification|confirm-instant-donation|instant-donation-thanks)/g)?.length) {
```
mcstover, line 78: To be clear, I do not like this. Prob move it to a simple list util soon or find a better way to import/clean our route declarations...

PR #4606, opened 2023-02-13
fix: ensure login links always go to login MARS-327

emuvente:
Lenders often return to the site without any cookies (new sessions, new devices, etc.) which includes the `kvu` cookie. That cookie is used to determine whether to send users to login or to sign-up by default, and without it the users are sent to sign-up. This is very confusing behavior when clicking a login button, and it has led to several lenders unintentionally creating new accounts. The PR forces all the existing login buttons to always go to login (which is the expected ux). The previous default behavior is now accessed with the `autoPage=true` url parameter to support the checkout "continue" buttons.

### DIFF
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -77,16 +77,17 @@ module.exports = function authRouter(config = {}) {
 			options.prompt = 'login';
 		}
 		// Go to register instead of login if the user has not logged in before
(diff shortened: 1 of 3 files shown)


mcstover (approved):
Sweet!

christian14b (approved):

PR #4609, opened 2023-02-14
fix: progress was changing on mount, callouts now just loaded on mount

dyersituations:
https://kiva.atlassian.net/browse/CORE-1078
https://kiva.atlassian.net/browse/CORE-1091

- Loan callouts are now rendered only client-side (onmount) - before they were rendered serverside, then the random choosing of callouts caused them to change on mount - ideally we'd pass the state from the SSR, but it'd have to be via apollo, and this feature doesn't seem worth that extra effort - we might later pull card data clientside, if the new loan cards win
- Moved the loan callouts util call into the component itself for encapsulation of logic
- Moved a few local resolver apollo props to the prefetched channel data, so that the component can sender serverside with the data needed for the progress bar (and it doesn't start at 0 when the page first loads)

### DIFF
--- a/src/graphql/fragments/loanCardFields.graphql
+++ b/src/graphql/fragments/loanCardFields.graphql
@@ -59,4 +59,9 @@ fragment loanCardFields on LoanBasic {
 	...on LoanDirect {
 		trusteeName
 	}
+
+	# for loan-progress component
+	unreservedAmount @client
+	fundraisingPercent @client
+	fundraisingTimeLeft @client
 }
--- a/src/graphql/fragments/loanCardFields.graphql
+++ b/src/graphql/fragments/loanCardFields.graphql
@@ -59,9 +59,4 @@ fragment loanCardFields on LoanBasic {
 	...on LoanDirect {
 		trusteeName
 	}
-
-	# for loan-progress component
-	unreservedAmount @client
-	fundraisingPercent @client
-	fundraisingTimeLeft @client
 }
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -89,27 +89,28 @@
 					/>
 				</router-link>
 
-				<!-- Loan callouts -->
-				<loan-callouts
-					v-if="!isLoading"
-					:loan="loan"
-					:category-page-name="categoryPageName"
-					class="tw-mt-1.5"
+				<!-- Loan call outs -->
+				<kv-loading-placeholder
+					v-if="typeof loanCallouts === 'undefined'"
+					class="tw-mt-1.5 tw-mb-2"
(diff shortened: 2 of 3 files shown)


src/graphql/fragments/loanCardFields.graphql
```diff
@@ -59,4 +59,9 @@ fragment loanCardFields on LoanBasic {
 	...on LoanDirect {
 		trusteeName
 	}
+
+	# for loan-progress component
+	unreservedAmount @client
+	fundraisingPercent @client
+	fundraisingTimeLeft @client
```
mcstover, line 66: This is a shared fragment and will affect all loan cards that use it so let's please not add this here at this time.

Also, we should have in place an animation on the progress bars so that these fields could always be dynamic and updated in the client.

dyersituations: Yeah... I wasn't sure if we cared about adding these to the shared. But I'll move them.

And for the loading, it's more about what data is prefetched and what data isn't. The prefetched data didn't have these three props. I find an alternative path.

mcstover: oh well, that's actually a good point. If the "actual" graphql fields are already included in the loanFragment then the data has been fetched and these would just add the local field processing... I don't know how much additional latency these local fields might add...

dyersituations: Thoughts on keeping on the shared then? I assumed it wasn't a huge impact. Otherwise I'll need to tweak the channel query or add a "client-side loading" flag to the new loan card. I don't mind doing those alternatives, if needed.

dyersituations: I have an alternate solution now that checks those local resolver values for `undefined`. Happy to go either path, but I'll push up the alternative after walking the dog...

mcstover: It may be that it turns out to be ok to make this change but I think we should separate it and look at how the other queries that implement this fragment might handle the initial presence of these fields.

I'm guessing having local fields in a fragment worked fine during dev but that may be one aspect we want to further verify as well...

dyersituations: Yeah the local fields in fragment worked fine, but I've just pushed up the alternate solution, pushing the data pull back to clientside, but adding some extra checks + loaders. Turns out the loaders for progress weren't working before, so these were good changes to get in regardless.

image

mcstover (approved):
Just curious is we removed the animation on this new progress setup? Previously, they were set to 0 and the updated value would animate from 0 to the proper state...

dyersituations:
> Just curious is we removed the animation on this new progress setup? Previously, they were set to 0 and the updated value would animate from 0 to the proper state...

@mcstover There is still animation, but unfortunately we now have the orange color when the amount is low, so starting it out at 0 w/ orange, then animating in the real amount looks weird.

roger-in-kiva (approved):

PR #4610, opened 2023-02-14
fix: downgrade @apollo/client to v3.3.21 to address memory leak

emuvente:
Hoping this will address the memory leak we've been seeing, based on https://github.com/apollographql/apollo-client/issues/9712#issuecomment-1197225283


mcstover (approved):
🤞

PR #4617, opened 2023-02-14
feat: add user attribute for when a user has a us loan in basket

eddieferrer:
ACK-540

I thought the `BasketItemList` would be a cleaner place to put this as opposed to the `CheckoutPage`. This fires a user attribute to signal that a user has a US loan in basket. I added this attribute on optimizely already, the idea is that we would like to exclude users with this attribute from the thanks page experiment.


emuvente (approved):
Makes sense. :+1:

mcstover (approved):

dyersituations (approved):

PR #4624, opened 2023-02-17
fix: card width fixed for carousels in lh page

roger-in-kiva:
new loan card width fixed to avoid shadow cut off in LH page

### DIFF
--- a/src/components/LoanFinding/LendingCategorySection.vue
+++ b/src/components/LoanFinding/LendingCategorySection.vue
@@ -9,7 +9,7 @@
 			</p>
 			<kv-carousel
 				class="tw-w-full tw-overflow-hidden tw-my-3"
-				:class="{ 'tw-pl-1' : enableLoanCardExp }"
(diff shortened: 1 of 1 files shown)


src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -9,6 +9,7 @@
 			</p>
 			<kv-carousel
 				class="tw-w-full tw-overflow-hidden tw-my-3"
+				:class="{ 'tw-pl-1' : enableLoanCardExp }"
```
dyersituations, line 12: Looks like this should be `tw-px-1`, since the pager is just barely off center with just left padding.

dyersituations: Looks like we also need a `tw-pt-1` so the top shadow doesn't get cut off.

roger-in-kiva: Added, ty!

dyersituations (approved):

PR #4626, opened 2023-02-17
fix: updated Apollo updateQuery usage for downgrade

dyersituations:
- The `updateQuery` method is simply `readQuery` followed by `writeQuery`
- Realized that the `writeQuery` doesn't actually return the updated state, and the return wasn't being used, so removed from `loanSearch`

https://github.com/apollographql/apollo-client/pull/8382/files


emuvente (approved):
Awesome!

mcstover (approved):

mcstover (approved):
I think is good merge

PR #4630, opened 2023-02-21
feat: new quick filters UI and functionality

egoroza:
New (2/27)
Screen Shot 2023-02-27 at 1 24 34 PM
Screen Shot 2023-02-27 at 1 25 30 PM
Screen Shot 2023-02-27 at 1 25 38 PM

OLD:
Screen Shot 2023-02-21 at 3 51 28 PM
Screen Shot 2023-02-21 at 3 52 14 PM


src/components/LoansByCategory/QuickFilters/FilterPills.vue
```diff
@@ -0,0 +1,94 @@
...
+					this.selectedOptions = this.selectedOptions.filter(val => val !== option);
+				}
+			}
+			this.$emit('update-values', { values: [...this.selectedOptions], selected, option });
+		}
+	}
+};
+</script>
+
+<style lang="postcss" scoped>
```
mcstover, line 70: Could we please convert as much of these styles as possible to use tailwind variables.

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -309,7 +356,29 @@ export default {
 };
 </script>
 
-<style scoped>
+<style lang="postcss" scoped>
+.filter-pill {
```
mcstover, line 360: Same here, we need to use our styleguide vars for this.

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -375,7 +375,7 @@ export default {
 			flssLoanSearch: {},
 			quickFiltersOptions: {
 				gender: [{
-					key: '',
+					key: 'all',
```
dyersituations, line 378: Just out of curiosity, why the change to `all` instead of empty string?

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -345,6 +369,29 @@ export default {
 </script>
 
 <style lang="postcss" scoped>
+	.filter-pill {
```
dyersituations, line 372: Some of the CSS here can be changed to tailwind as well, if not in the template (since we don't have all of the versions like hover in our compiled TW - we don't currently rely on JIT tailwind), here with `@apply`.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -179,13 +184,16 @@ export default {
...
+		KvMaterialIcon,
+		FilterPills
 	},
 	data() {
 		return {
 			mdiFilterVariant,
+			mdiSort,
 			selectedCategory: 0,
 			selectedGender: '',
+			selectedGenders: ['all'], // TODO: consolidate to just this when UI changes apply to QF on lending home, too
```
dyersituations, line 196: Not sure what this comment means. Please add a ticket and reference here with details on what this `TODO` entails.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -298,10 +310,21 @@ export default {
 				location
 			);
 		},
+		// We will need to factor in categories at some point
```
dyersituations, line 313: Not sure what this comment means. Consider adding a ticket and referencing here, if future changes are needed.

dyersituations:
In desktop view, I'm seeing a white box/border under + around the gender pills. This box doesn't appear when using mobile mode of chrome dev tools.
image

dyersituations:
Clicking either the location or sort by "activates" the textbox. The active/focus border can be removed.
image

dyersituations:
Clicking outside the textbox on the sort by filter doesn't open the dropdown. The clickable area is vaguely shown by this red area:

image

dyersituations:
The down arrows on the pills are slightly different sizes and misaligned

image

dyersituations:
The box shadow on the gender filter pills is cut off. It's subtle, but there's a difference between the gender filter shadows and the other filters
image

dyersituations:
Left padding missing in mobile view (notice difference between left and right padding).
image

dyersituations:
There's a small height different in filter pill heights

image

dyersituations:
LH filters now show an empty gender filter, and the filters no longer align.

image

dyersituations:
All genders in LH doesn't work.

image

src/pages/Lend/LoanChannelCategoryControl.vue
egoroza: Looks like if we have an empty string it won't apply it as an `id` property for html elements. Since the filter pills are technically checkbox inputs (`<input>`) when we pass in `:id=option.key` and `option.key === ''` it doesn't apply an ID at all and seems to ignore empty string, so any interactions with "All genders" doesn't get considered as part of the checkbox list state

egoroza: Obviously could have worked around this but I thought it made sense to just change the `key` across the board instead of interpolating a random char/phrase or something similar

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: ya will do, just meant mostly that i don't know if it makes sense to have 2 gender states. The dropdown on lending home for gender should just use the array as well, there will just only ever be 1 element in the array at a time

src/pages/Lend/LoanChannelCategoryControl.vue
dyersituations: Sounds good, thanks for the info, using `all` is fine.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
dyersituations: In that case, maybe update the comment to mention that the work on LH might mean removing the other gender state, I'm assuming `selectedGender`?

egoroza:
> In desktop view, I'm seeing a white box/border under + around the gender pills. This box doesn't appear when using mobile mode of chrome dev tools. image

How can one reproduce? Browser? I don't see this issue on my Chrome desktop view so need to investigate

dyersituations:
> How can one reproduce? Browser? I don't see this issue on my Chrome desktop view so need to investigate

Interesting. I see this in Chrome on my MacBook on both my MacBook screen and external monitor. Make sure that the devtool mobile emulator isn't enabled, since it doesn't occur in the emulator.

src/components/LoansByCategory/QuickFilters/FilterPills.vue
mcstover: Thanks for reducing the level of non-tailwind styles!

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
dyersituations: Looks like this comment could still use some clarity, maybe mention that `selectedGender` could be removed, if that's the change needed here.

dyersituations:
At a width slightly larger than small, the entire page gets horizontal scrolling. At a smaller view, the second row of filters scroll, but not at the width in this screenshot. Maybe the filters need horizontal scrolling at the medium screen breakpoint?

image

dyersituations:
In Chrome without mobile emulation, I'm still seeing a weird box around the gender filters. That box is also cutting off the gender filter drop shadow and making the sort filter pill taller.

image

dyersituations:
For consistency with the other filter pills, the sort filter should have the pointer icon for the entire pill. The pointer icon currently only shows up for the sort icon on hover.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -40,7 +30,16 @@
 					</option>
 				</kv-select>
 			</div>
-			<div v-if="!removeGenderDropdown" class="tw-flex tw-flex-col tw-grow">
+			<div v-if="!removeGenderDropdown && !withCategories" class="tw-overflow-scroll tw-pb-1">
```
dyersituations, line 33: This should probably be `tw-overflow-x-auto`. The box I'm seeing in the screenshot in another comment is the scrollbars always being visible.

dyersituations:
Please create a follow-up ticket to address the cut-off drop-shadows. It's subtle, so I'll approve as is, but we should address later when working on the filters:

image

image

dyersituations (approved):

PR #4632, opened 2023-02-22
[CIT-67, CIT-101, CIT-103] feat: adding and implementing a progress bar component on cc page

tiarascoleman:
UI mirrors design specs- currently I still have to check if the _progress bar_ functionality actually works (like if I click on an actual loan and add it to my basket, if the progress bar will self-edit itself). This is important especially for the two other modes (there are three: when an employee is just starting ("Let's Get Started"), when an employee is filling their basket ("Keep Going"), and when they have used all their credits ("You Did It"). 

The images are attached below- additionally the "Keep Going" mechanism isn't working; I'm attempting to make creditAmountLeft similar to moneyLeft in LoanProgress.vue however I can't figure out to work out unreservedAmount (in KivaBasicLoanCard.vue) to be referred to via promoAmount  (in CCLandingPage.vue)- any thoughts?

Current:
Screen Shot 2023-02-21 at 7 23 10 PM
Screen Shot 2023-02-21 at 7 26 11 PM
Screen Shot 2023-02-21 at 8 46 01 PM

Design Specs: https://www.figma.com/proto/kbM1wO1fXPq9U4G2XBuVBx/Campaign-page-checkout?page-id=71%3A1266&node-id=87%3A1283&viewport=-843%2C248%2C0.2&scaling=min-zoom&starting-point-node-id=71%3A1267


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -36,6 +36,8 @@
 				/>
 				<campaign-loan-wrapper
 					ref="mlLoanDisplay"
+					:promo-amount="promoAmount"
```
michelleinez, line 39: These props should get passed in via the campaignLoanWrapperProps object instead of directly as a prop itself.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+import KvProgressBar from '@/components/Kv/KvProgressBar';
+
+export default {
+	name: 'CampaignProgressBar',
+	components: {
+		KvProgressBar
+	},
+	props: {
+		promoAmount: {
+			type: String,
```
dyersituations, line 63: Should `promoAmount` be a string or number. I notice in the template that there are `toString()` calls that wouldn't be needed if it's a string. It looks like this should be a number due to the usage in `partialPromoAmount`.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+			default: '$0.00'
+		},
+		creditAmountLeft: {
+			type: String,
```
dyersituations, line 67: Based on the usage in `partialPromoAmount`, it looks like this should be a number.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+			default: '$0.00'
+		},
+		promoName: {
+			type: String,
+			default: null
+		},
+	},
+	computed: {
+		partialPromoAmount() {
+			return (this.creditAmountLeft / this.promoAmount) * (100);
```
dyersituations, line 77: This will need to convert props to numbers, if the props need to be strings.

src/components/Contentful/CampaignLoanWrapper.vue
```diff
@@ -81,7 +88,15 @@ export default {
 		componentProps: {
 			type: Object,
 			default: () => {}
-		}
+		},
+		promoAmount: {
+			type: String,
```
dyersituations, line 93: Based on other comments on this PR, this probably needs to be a number.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+
+export default {
+	name: 'CampaignProgressBar',
+	components: {
+		KvProgressBar
+	},
+	props: {
+		promoAmount: {
+			type: String,
+			default: '$0.00'
```
dyersituations, line 64: This default will result in two `$` in the template, if it's ever used.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+			</h3>
+			<kv-progress-bar
+				class="tw-mb-2"
+				aria-label="Percentage of credit amount"
+				:max="promoAmount.toString()"
+				:value="(0).toString()"
+			/>
+		</template>
+
+		<template v-if="creditAmountLeft >= '1.00' && creditAmountLeft !== promoAmount">
```
dyersituations, line 20: This compare will need to change, if the prop is changed to a number

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+			<h4 class="tw-mb-1 tw-px-3">
+				Keep Going!
+			</h4>
+			<h3 class="tw-mb-2 tw-px-3">
+				You have $ {{ partialPromoAmount | numeral }}
```
dyersituations, line 25: Looks like an extra space after `$`

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
...
+				Let's Get Started
+			</h4>
+			<h3 class="tw-mb-2 tw-px-4">
+				You have ${{ promoAmount | numeral }} in credit from {{ promoName }}
+			</h3>
+			<kv-progress-bar
+				class="tw-mb-2"
+				aria-label="Percentage of credit amount"
+				:max="promoAmount.toString()"
+				:value="(0).toString()"
```
dyersituations, line 16: Are these static bars? It looks this bar will always just be gray with this value.

image

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -0,0 +1,132 @@
+<template>
+	<div
+		class="tw-rounded tw-bg-brand-50 tw-w-full tw-py-1 tw-px-1 tw-mb-2 tw-mt-2"
+	>
+		<template v-if="promoAmount">
+			<h4 class="tw-mt-2 tw-mb-1 tw-px-4">
+				Let's Get Started
+			</h4>
+			<h3 class="tw-mb-2 tw-px-4">
+				You have ${{ promoAmount | numeral }} in credit from {{ promoName }}
```
dyersituations, line 10: Looks like `promoName` isn't making it to this component currently, potentially related to the comment by @michelleinez on `CCLandingPage`

dyersituations:
Are the two sections supposed to align on the left? I'm seeing just a little extra padding on the first one:

image

PR #4637, opened 2023-02-23
feat: dynamic data for spotlight section added with rotation based in…

roger-in-kiva:
- spotlight data added to a json file
- LoanFinding page updated to get data from json file
- cookie based method added to rotate spotlight variations


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -230,6 +221,11 @@ export default {
 		},
 		closeLightbox() {
 			this.showLightbox = false;
+		},
+		verifySpotlightIndex() {
+			const spotlightCookie = this.cookieStore.get('lh_spotlight') || null;
+			if (spotlightCookie) this.spotlightIndex = spotlightData.length - 1 <= Number(spotlightCookie) ? 0 : Number(spotlightCookie) + 1; // eslint-disable-line max-len
```
dyersituations, line 227: Storing `Number(spotlightCookie)` in a const with the body of the `if` statement would make this line a little more readable and wouldn't need the linting disable.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -200,15 +191,15 @@ export default {
...
-			});
+		async fetchSpotlightLoans() {
+			const flssFilterCriteria = this.activeSpotlightData?.flssLoanSearch ?? {};
+			const { loans } = await runLoansQuery(
+				this.apollo,
+				{ ...flssFilterCriteria, pageLimit: 6 },
+				FLSS_ORIGIN_LENDING_HOME
+			);
+
+			this.spotlightLoans = loans ?? [];
```
dyersituations, line 202: The fallback here should probably be `{}`

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -248,7 +244,8 @@ export default {
 	mounted() {
 		this.getRecommendedLoans();
 		this.getSecondCategoryData();
-		this.fetchMFILoans();
+		this.verifySpotlightIndex();
+		this.fetchSpotlightLoans();
```
dyersituations, line 248: Long-term we'll probably want to only run this fetch when needed (when the spotlight is visible), but we have another ticket for that lazy loading, and this query isn't adding a new query compared to before.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -147,6 +135,9 @@ export default {
 			if (this.matchedLoansTotal > 0) return 'Stretch your funds further with the help of our partners and Kivans just like you'; // eslint-disable-line max-len
 			return 'Loans that are ending soon or almost funded';
 		},
+		activeSpotlightData() {
+			return spotlightData[this.spotlightIndex] ?? [];
```
dyersituations, line 139: The fallback here should probably be {}

dyersituations (approved):

PR #4639, opened 2023-02-24
fix: resolve errors with read/write query for autolending

dyersituations:
https://kiva.atlassian.net/browse/CORE-1148

- Resolved issues in autolending local resolver with the `readQuery` and `writeQuery` combo that was needed to replace the `updateQuery` after Apollo downgrade
- Used lodash `mergeWith` in order to create a deep copy + overwrite arrays (specifically country list in `LoanSearchFilters`
- I would randomly get a weird "init" error when testing my autolending changes, and I eventually tracked it down to the 3rd party `FocusLock` component in the `KvLightbox` - I replaced the usages with the lightbox from our component library

### DIFF
--- a/src/api/localResolvers/autolending.js
+++ b/src/api/localResolvers/autolending.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-underscore-dangle */
 import _get from 'lodash/get';
+import _mergeWith from 'lodash/mergeWith';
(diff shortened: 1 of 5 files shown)


mcstover (approved):
Looks good

emuvente (commented):
Nice!

emuvente (approved):

PR #4650, opened 2023-02-28
feat: progress bar copy updated when loan is already funded

roger-in-kiva:
- copy text updated in progress bar component when loan is already funded

### DIFF
--- a/src/components/LoanCards/LoanProgressGroup.vue
+++ b/src/components/LoanCards/LoanProgressGroup.vue
@@ -51,7 +51,7 @@ export default {
 			const formattedTimeLeft = `${this.timeLeft !== '' ? `. ${this.timeLeft}` : ''}`;
 			// Some time left strings already include an exclamation mark
 			const exclamationMark = this.amountLow && !formattedTimeLeft.includes('!') ? '!' : '';
-			return `${formattedMoneyLeft} to go${formattedTimeLeft}${exclamationMark}`;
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/LoanProgressGroup.vue
```diff
@@ -51,7 +51,7 @@ export default {
 			const formattedTimeLeft = `${this.timeLeft !== '' ? `. ${this.timeLeft}` : ''}`;
 			// Some time left strings already include an exclamation mark
 			const exclamationMark = this.amountLow && !formattedTimeLeft.includes('!') ? '!' : '';
-			return `${formattedMoneyLeft} to go${formattedTimeLeft}${exclamationMark}`;
+			if (!this.numeralLeft.value()) return 'Funded!'; return `${formattedMoneyLeft} to go${formattedTimeLeft}${exclamationMark}`; // eslint-disable-line max-len
```
dyersituations, line 54: The second `return` should just be moved to a new line for readability.

mcstover (commented):
Please, break these conditions into an if block with brackets, followed by the default return statement. It's hard to know there are actually 2 different states here now...

src/components/LoanCards/LoanProgressGroup.vue
roger-in-kiva: Moved, ty!

dyersituations (approved):

PR #4652, opened 2023-03-01
feat: loading state for quick filter genders

egoroza:
This PR adds a loading state to the gender filter pills using placeholders when all of our filters are on one line. This ensures that the location selector and select sort dropdowns don't shift over dramatically once the data loads which can be confusing and jarring for the ux.

Screen Shot 2023-03-01 at 11 45 56 AM
Screen Shot 2023-03-01 at 11 46 09 AM

### DIFF
--- a/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
+++ b/src/components/LoansByCategory/QuickFilters/QuickFilters.vue
@@ -3,9 +3,8 @@
 		<div
 			class="tw-flex"
 			:class="{
-				'tw-px-1 lg:tw-pr-0 tw-justify-start md:tw-gap-1 tw-flex-wrap lg:tw-flex-nowrap'
-					: !withCategories,
-				'tw-gap-2 tw-flex-col lg:tw-flex-row tw-w-full' : withCategories
+				'tw-px-1 lg:tw-pr-0 tw-justify-start md:tw-gap-1 tw-flex-wrap lg:tw-flex-nowrap': !withCategories,
+				'tw-gap-2 tw-flex-col lg:tw-flex-row tw-w-full': withCategories
 			}"
 		>
 			<div v-if="withCategories" class="tw-flex tw-flex-col tw-grow">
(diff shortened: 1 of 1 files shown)


src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -403,6 +412,20 @@ export default {
...
+		@apply tw-rounded !important;
+	}
+
+	.placeholder {
+		display: none;
+	}
+
+	@media screen and (min-width: 600px) {
+		.placeholder {
+			display: flex;
```
dyersituations, line 425: These two displays could be converted to TW.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -30,13 +30,20 @@
...
+			<div v-if="!removeGenderDropdown && !withCategories" class="tw-flex tw-gap-1 tw-overflow-x-auto tw-pb-1">
 				<filter-pills
 					:filters-loaded="filtersLoaded"
 					:options="filterOptions.gender"
 					:selected-values="selectedGenders"
 					@update-values="updateGenders($event)"
 				/>
+				<div v-if="!filtersLoaded" class="tw-flex tw-gap-1 placeholder">
+					<kv-loading-placeholder
+						style="height: 2.75rem; width: 100px;"
```
dyersituations, line 42: Looks like the loading placeholders are 1px shorter, and the `height` can be removed here to fix that.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
egoroza: I think the TW breakpoint is slightly different (734px) ?

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
dyersituations: Done.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
dyersituations: Good call, leaving as is for now.

PR #4655, opened 2023-03-01
feat: [CIT-102] changing header of cc landing page to match desgin specs

tiarascoleman:
Before:
 
Screen Shot 2023-03-01 at 2 50 29 PM

After:
Screen Shot 2023-03-01 at 2 47 27 PM
Screen Shot 2023-03-01 at 2 47 50 PM

### DIFF
--- a/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
+++ b/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
@@ -2,22 +2,11 @@
 	<div class="loan-filters">
 		<div class="loan-filters__top-row">
 			<span class="tw-mb-2 md:tw-mb-0">
-				<kv-button
-					class="loan-filters__toggle tw-pr-2"
-					variant="secondary"
-					@click.native.prevent="showFilters()"
-				>
-					<span class="loan-filters__toggle-text">
-						Filter loans
-					</span>
-					<kv-material-icon
-						aria-hidden="true"
-						class="loan-filters__toggle-icon"
-						name="chevron-down"
-						:icon="mdiChevronDown"
-					/>
-				</kv-button>
-				<span class="tw-font-medium tw-whitespace-nowrap">{{ totalCount }} loans</span>
+				<div class="tw-inline-flex tw-items-center">
+					<h2 class="tw-text-center">
+						{{ promoName }} recommends these people
+					</h2>
+				</div>
 			</span>
 
 			<div v-if="showLoanDisplayToggle" class="loan-filters__loan-display">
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "kvue",
-	"version": "2.508.0",
```
michelleinez, line 3: Updating the vue version can introduce issues so we probably shouldn't do that unless it's part of a bigger effort where we have a test plan in place and everyone is made aware of the change.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+### [2.508.1](https://github.com/kiva/ui/compare/v2.508.0...v2.508.1) (2023-03-01)
+
+
+### Bug Fixes
+
+* filters should be able to handle undefined facets object ([0fec7f8](https://github.com/kiva/ui/commit/0fec7f8acc1a4abb49593734b03864e89ef8c32e))
```
dyersituations, line 6: You might consider merging `main` into `CIT-checkout-flow-improvements` branch directly first, so the PR just focuses on new changes.

package.json
mcstover: This one can sometimes happen but I'm more concerned that there is some much other work here.

My suggestion is to change this PR to be into `main` instead of a feature branch. This is a small segment of work that should be ok to work against main on.

dyersituations: This is the project build version. It can still be reverted though.

CHANGELOG.md
mcstover: this would also work. I think the CIT checkout flow branch has enough changes for now though and this doesn't relate to checkout so it's good to keep separate.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -648,6 +651,14 @@ export default {
 	}
 }
 
+.divider {
```
dyersituations, line 654: It looks like this class is no longer used.

package.json
dyersituations: @mcstover Yeah looks like the target branch needs to get `main` merged in first, I added a separate comment.

CHANGELOG.md
dyersituations: Yeah I'm not sure which branch is important here, like if it's a feature party branch.

michelleinez: The reason CIT-checkout-flow-improvements isn't merged is because CIT has a visa campaign where we can't make any major feature changes to prod until the emails are sent out and the employees have made their loans.

michelleinez: I don't have any intention to merge those changes in at least until after the March 8th campaign date.  @mcstover @dyersituations @tiarascoleman

mcstover: Makes sense @michelleinez thank you!

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
mcstover: Agree, this could be removed in some later cleanup

michelleinez (approved):

dyersituations (approved):

PR #4660, opened 2023-03-02
feat: server cache data for first two recommended loans

dyersituations:
https://kiva.atlassian.net/browse/CORE-1131

- Turns out the LCP score was being mostly affected by the large loan images being pulled in on mounted
- Prefetching the top 2 recommended loans ensures the images are visible on initial page load
- There are many other Lighthouse changes that we can make, but this is a good first step


roger-in-kiva (approved):
Looks good!
Is there a big diff prefetching 12 than 2? To reduce recommended calls.

dyersituations:
> Looks good! Is there a big diff prefetching 12 than 2? To reduce recommended calls.

Good question, and I was wondering the same. Our loan card details are fairly large objects, so my thought was just pulling 2 would be the quickest. Also, it prevents the 10 other cards from rendering. Our cards will render if they have data, otherwise they wait with an observer, so pulling 2 reduces the initial render time of the page.

mcstover (approved):
Prob need to handle filter params later but good for now!

PR #4661, opened 2023-03-02
fix: remaining tracking

dyersituations:
- Turns out the other 2 potential spots a user lands also needed the tweaked exp key


roger-in-kiva (approved):
👍

PR #4669, opened 2023-03-07
feat: allow guest users who have logged in to participate in LH exp

dyersituations:
https://kiva.atlassian.net/browse/CORE-1121

- Checks for `hasEverLoggedIn` instead of just logged in - the `kvu` cookie is create during log in process
- Removed experiment setting + assignment cookies that aren't needed in `preFetch`, since those get prefetched in parent wwwpage component

### DIFF
--- a/src/graphql/query/lendByCategory/lendByCategory.graphql
+++ b/src/graphql/query/lendByCategory/lendByCategory.graphql
@@ -1,17 +1,6 @@
 query lendByCategory($basketId: String) {
+	hasEverLoggedIn @client
 	general {
-		mlServiceBandit: uiExperimentSetting(key: "EXP-ML-Service-Bandit-LendByCategory") {
-			key
-			value
-		}
-		mfiRecommendations: uiExperimentSetting(key: "mfi_recommendations") {
-			key
-			value
-		}
-		lbcCategoryServiceRows: uiExperimentSetting(key: "flss_category_service") {
-			key
-			value
-		}
 		loanFindingPage: uiExperimentSetting(key: "loan_finding_page") {
 			key
 			value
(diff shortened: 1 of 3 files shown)


src/graphql/query/lendByCategory/lendByCategory.graphql
```diff
@@ -1,17 +1,6 @@
 query lendByCategory($basketId: String) {
+	hasEverLoggedIn @client
 	general {
-		mlServiceBandit: uiExperimentSetting(key: "EXP-ML-Service-Bandit-LendByCategory") {
-			key
-			value
-		}
-		mfiRecommendations: uiExperimentSetting(key: "mfi_recommendations") {
```
mcstover, line 7: didn't see this below, is it all clear to remove?

mcstover (approved):

src/graphql/query/lendByCategory/lendByCategory.graphql
roger-in-kiva: thanks!

roger-in-kiva (approved):

src/graphql/query/lendByCategory/lendByCategory.graphql
dyersituations: @mcstover Are you referring to these `uiExperimentSetting` settings in general or specifically `mfi_recommendations`? Because we now prefetch all of the active settings, we no longer need to call `uiExperimentSetting` or assignment, unless a component needs the results as part of its own prefetch. And `mfi_recommendations` specifically is an experiment we just removed, but this was a left over piece.

mcstover: just making sure it was cruft and ok to be removed

dyersituations:
Closed and reopened to try to get coveralls to run.

PR #4672, opened 2023-03-07
Core 1121 enable lending home for logged out users who we know have logged in before existing lenders regardless of log in status

dyersituations:
https://kiva.atlassian.net/browse/CORE-1121

- Checks for `hasEverLoggedIn` instead of just logged in - the `kvu` cookie is create during log in process
- Removed experiment setting + assignment cookies that aren't needed in `preFetch`, since those get prefetched in parent wwwpage component

**Creating a new PR to hopefully get coveralls to return status to the PR**

### DIFF
--- a/src/graphql/query/lendByCategory/lendByCategory.graphql
+++ b/src/graphql/query/lendByCategory/lendByCategory.graphql
@@ -1,17 +1,6 @@
 query lendByCategory($basketId: String) {
+	hasEverLoggedIn @client
 	general {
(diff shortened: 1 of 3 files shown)


mcstover (approved):
Also, I appreciate that descriptive branch name 🤣

dyersituations:
> Also, I appreciate that descriptive branch name 🤣

Haha you can thank Jira for that one...

PR #4676, opened 2023-03-08
Core 1195 filter pills test

roger-in-kiva:
- new quick filters ui showed under a test

### DIFF
--- a/src/components/LoansByCategory/QuickFilters/LocationSelector.vue
+++ b/src/components/LoansByCategory/QuickFilters/LocationSelector.vue
@@ -227,10 +227,6 @@ export default {
 			type: String,
 			required: true,
 		},
-		withCategories: {
-			type: Boolean,
-			required: true,
-		},
 		enableFilterPills: {
 			type: Boolean,
 			default: false
--- a/src/components/LoansByCategory/QuickFilters/LocationSelector.vue
+++ b/src/components/LoansByCategory/QuickFilters/LocationSelector.vue
@@ -5,12 +5,12 @@
 		v-click-outside="closeRegions"
 	>
 		<label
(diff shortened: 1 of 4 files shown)


dyersituations:
Seeing a scrollbar now around the gender pills.

image

src/components/LoansByCategory/QuickFilters/LocationSelector.vue
```diff
@@ -231,7 +230,11 @@ export default {
 		withCategories: {
```
dyersituations, line 230: Looks like this prop can be removed now.

src/components/LoansByCategory/QuickFilters/QuickFilters.vue
```diff
@@ -1,10 +1,24 @@
 <template>
 	<div class="tw-flex tw-flex-col tw-mb-2 tw-w-full">
+		<div v-if="!withCategories && !enableFilterPills" class="tw-flex tw-items-center tw-mb-2">
```
dyersituations, line 3: Why were these two blocks of new elements needed? I thought the old version of the filters already were shown on LH.

roger-in-kiva: These were removed with ui update, so, if we need to keep design like here https://www.kiva.org/lend-by-category/agriculture for testing, we need these new(old) blocks, right?
Like here https://github.com/kiva/ui/blob/155c55d83dde1f5180c9964ec9130635a4d41fa0/src/components/LoansByCategory/QuickFilters/QuickFilters.vue#L2

dyersituations: If this was how it was before, then makes sense, thanks.

dyersituations (approved):

PR #4678, opened 2023-03-09
fix: api error from querying trusteeType

christian14b:
The `trusteeType` was removed from db and graphql api so is failing to resolve.


src/components/LoanCards/HoverLoanCard/InfoPanels/TrusteeInfoPanel.vue
```diff
@@ -24,10 +24,10 @@
 						{{ trusteeName }}
 					</span>
 				</div>
-				<div v-if="this.trusteeType">
+				<div v-if="this.organizationType">
 					<dt>Trustee type:</dt>
```
mcstover, line 28: Please change the label too.

mcstover (approved):
Thanks!

emuvente (approved):

PR #4679, opened 2023-03-09
feat: add test id based on loan type to bp

eddieferrer:
ACK-552

Will add the `data-testid` to the main div in the borrower profile. Allowing us to use this property for exclusion of loans in optimizely tests.

looks like:
`<div class="www-page" id="borrower-profile" data-testid="partner-loan">`


mcstover (approved):
Seems fine as and addition. Maybe we should have a generic `borrower-profile` id somewhere in the page too...

dyersituations (approved):

emuvente (approved):
Thanks!

PR #4680, opened 2023-03-09
Add to basket cookie error vue 1583

christian14b:
Approach when trying to add a new loan to an expired basket:

1. Save loan information in a cookie
2. Refresh the page. 
3. Retrieve loan information.
4. Retry adding loan to the basket.

https://user-images.githubusercontent.com/56947778/224153354-46cfb5f4-7099-4f59-968e-3736755fa71c.mov


src/util/basketUtils.js
```diff
@@ -42,6 +44,7 @@ export function setLendAmount({ amount, apollo, loanId }) {
 			optimisticResponse: {
 				__typename: 'Mutation',
 				shop: {
+					id: basketId,
```
mcstover, line 47: I think this should just be `0` as I don't believe it's the same as the basketId we use from the cookie.

src/util/basketUtils.js
```diff
@@ -77,3 +80,9 @@ export function setLendAmount({ amount, apollo, loanId }) {
 export function setDonationAmount() {
 	// TODO
 }
+
+export function handleInvalidBasket({ loan, cookieStore }) {
+	cookieStore.remove('kvbskt', { path: '/', secure: true });
+	cookieStore.set('failedLoan', JSON.stringify(loan));
```
mcstover, line 86: Could we please namespace this cookie a bit... maybe `kvatbid` or something like that...

mcstover (approved):

dyersituations:
Do we want a tooltip message to indicate something is happening to refresh the page? It seemed like there was a bit of a delay where the user could just keep clicking around. What happens if the user clicks add to basket again while the page is reloading, for example? A tooltip about "hey we encountered an error, refreshing the page to try again" or whatever might let the users know to wait.

dyersituations (approved):

src/util/basketUtils.js
```diff
@@ -20,7 +20,7 @@ function logSetLendAmountError(loanId, err) {
 }
 
 export function setLendAmount({
-	amount, apollo, loanId, basketId = ''
+	amount, apollo, loanId, basketId = 0
```
mcstover, line 23: The mutation below is expecting a string.

src/components/LoanCards/Buttons/LendButton.vue
```diff
@@ -71,6 +71,8 @@ export default {
 							);
 							Sentry.captureMessage(`Add to Basket: ${error.message}`);
 							if (error?.extensions?.code === 'shop.invalidBasketId') {
+								// eslint-disable-next-line max-len
+								this.$showTipMsg('There was a problem adding the loan to your basket, refreshing the page to try again.', 'error');
```
mcstover, line 75: With this added, I think we'll need to use an `else` block below to prevent 2 messages from showing.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -452,6 +452,8 @@ export default {
 				Sentry.captureException(e);
 				// Handle errors from adding to basket
 				if (e?.[0]?.extensions?.code === 'shop.invalidBasketId') {
+					// eslint-disable-next-line max-len
+					this.$showTipMsg('There was a problem adding the loan to your basket, refreshing the page to try again.', 'error');
```
mcstover, line 456: With this added, I think we'll need to use an `else` block below to prevent 2 messages from showing.

src/components/LoanCards/Buttons/LendButton.vue
christian14b: `return handleInvalidBasket` is preventing second tip message to show

mcstover: Ahh ok, Thanks!

src/util/basketUtils.js
```diff
@@ -19,7 +19,9 @@ function logSetLendAmountError(loanId, err) {
 	}
 }
 
-export function setLendAmount({ amount, apollo, loanId }) {
+export function setLendAmount({
+	amount, apollo, loanId, basketId = '0'
```
mcstover, line 23: I don't think you need to pass in this new param at all. You can just pass the int `0` with `id: 0` in the optimisticResponse.

Sorry about my previous comment which was confusing this new param getting passed in with the mutation declaration for $basketId, which is fine as is.

mcstover: Seems like this would show an error when setting the optimisticResponse, but it may not b/c it just needs a placeholder until the real response is returned...

christian14b: Thank you! Setting `id: '0',` within `shop` optimisticResponse fix the error

mcstover (approved):

mcstover (approved):

PR #4681, opened 2023-03-09
fix: comment out lp route handling for cms-page-server verification

mcstover:
Commenting out `/lp` route handling. This ensures links from Ui handled landing pages or the footer aren't loaded from UI.

https://github.com/kiva/puppet/pull/345

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -445,6 +445,8 @@ module.exports = [
 		path: '/lp/support-refugees',
 		redirect: '/refugees'
 	},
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

christian14b (approved):

emuvente (approved):

mcstover:
I also checked the repo to ensure there were no router-links using these paths and there weren't. Probably won't be as clean when we do the homepage route.

PR #4684, opened 2023-03-10
fix: add new error code when basket.shop is null and move retry atb t…

christian14b:
…o mounted hook

When a basket cookie is expired and you refresh the page `basket.shop` is set to `null` and sending a different error code.

In `lending-home` page basket state on `TheHeader.vue` was not being updated within created hook. Moving on to the mounted hook fixes the problem.

### DIFF
--- a/src/components/LoanCards/Buttons/LendButton.vue
+++ b/src/components/LoanCards/Buttons/LendButton.vue
@@ -14,7 +14,7 @@ import numeral from 'numeral';
 import * as Sentry from '@sentry/vue';
 import updateLoanReservation from '@/graphql/mutation/updateLoanReservation.graphql';
 import loanCardBasketed from '@/graphql/query/loanCardBasketed.graphql';
-import { handleInvalidBasket } from '@/util/basketUtils';
+import { handleInvalidBasket, hasBasketExpired } from '@/util/basketUtils';
 import KvButton from '~/@kiva/kv-components/vue/KvButton';
 
 export default {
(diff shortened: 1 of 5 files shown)


src/components/LoanCards/Buttons/LendButton.vue
```diff
@@ -70,7 +70,7 @@ export default {
 								`Failed: ${error.message.substring(0, 40)}...`
 							);
 							Sentry.captureMessage(`Add to Basket: ${error.message}`);
-							if (error?.extensions?.code === 'shop.invalidBasketId') {
+							if (['shop.invalidBasketId', 'shop.basketRequired'].includes(error?.extensions?.code)) {
```
dyersituations, line 73: This might be a good line of code to extract to a util, since we have the same 2 error strings in 3 places now.

dyersituations (approved):

src/plugins/retry-after-expired-basket-mixin.js
```diff
@@ -6,7 +6,7 @@ const injections = ['apollo', 'cookieStore'];
 
 export default {
 	mixins: [addToBasketInsterstitial],
-	created() {
+	mounted() {
```
mcstover, line 9: Something I just realized about this method is that it assumes that the add to basket interstitial should be shown.

For this particular mitgation, let's just remove the invocation of the interstitial. In it's place we can show a success tip message that says: "Your loan has been added to the basket."

Maybe later we can augment that with the Loan name if it's available. ie. "Your loan to ##name## has been added to the basket."

src/components/LoanCards/Buttons/LendButton.vue
christian14b: After a few new attempts, shop was set to `null` because of hot reload (I think this is an edge case only seen in local development) but do you think I should set the new basket anyway? If not I can just revert these files

mcstover: Oh, that's a good catch. I think a util to encasulate the is repeated check logic is enough for now. THanks!

mcstover (approved):
Thank you!

PR #4685, opened 2023-03-10
Core 1200 add more new card

roger-in-kiva:
- `More` text added to `loanUse` component as requirement for new loan cards

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -6,7 +6,7 @@
 		:style="{ minWidth: '230px', maxWidth: cardWidth }"
 	>
 		<div class="tw-grow">
-			<div :class="{ 'loan-card-active-hover': !allSharesReserved }">
+			<div>
 				<!-- Borrower image -->
 				<kv-loading-placeholder
 					v-if="isLoading"
@@ -231,10 +231,6 @@ export default {
 			type: String,
 			default: '',
 		},
-		showMoreLink: {
-			type: Boolean,
-			default: false
-		}
 	},
 	inject: ['apollo', 'cookieStore'],
 	mixins: [percentRaisedMixin, timeLeftMixin],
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -511,7 +516,7 @@ export default {
 }
 
 .loan-card-active-hover:hover .loan-card-use {
-	@apply tw-underline;
+	@apply tw-no-underline;
```
dyersituations, line 519: I think `loan-card-active-hover` can be removed entirely.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -511,7 +516,7 @@ export default {
 }
```
dyersituations, line 516: The line right above here should be `@apply tw-text-primary tw-no-underline;`, otherwise the loan use anchor gets underlined on focus.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -229,6 +230,10 @@ export default {
 		categoryPageName: {
 			type: String,
 			default: '',
+		},
+		showMoreLink: {
```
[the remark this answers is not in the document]
mcstover: This was previously used to toggle the "View loan" version of the button that goes to BP. It would be nice to have that present so this loan card could be used in contentful landing pages if it becomes our new default...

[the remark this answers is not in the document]
dyersituations: We'll be adding the "view loan" CTA next sprint, since we're experimenting with the new loan card on the home page. Looks like we'll have to port this card to CPS...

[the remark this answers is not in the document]
roger-in-kiva: Removed, thanks both!

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
roger-in-kiva: Removed

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
roger-in-kiva: Fixed

dyersituations (approved):

PR #4687, opened 2023-03-10
fix: ensure e2e test uses ui header

dyersituations:
https://kiva.atlassian.net/browse/CORE-1142

- The one failure was due to there being a slight delay in focus listener in CPS when compared with ui for the search bar -> the test now just looks to a ui page for the time being, and disables a header exp


mcstover (approved):
We could also use something like `/ui-site-map` or `/build`...

mcstover (approved):

PR #4689, opened 2023-03-13
feat: replace router-link for hp and comment out hp route

christian14b:
Use `a` tag for homepage redirects since cps will handle it.


dyersituations (approved):

mcstover (approved):
This looks good. We may also need to look at any instances of router.push using `/`.

emuvente (approved):
The missing route handler should take care of anything that was missed.

PR #4693, opened 2023-03-15
refactor: clean up category share experiment

eddieferrer:
MARS-360


src/components/Thanks/ThanksPageShare.vue
```diff
@@ -216,10 +201,6 @@ export default {
 			type: Object,
 			default: () => ({}),
 		},
-		simpleSocialShareVersion: {
```
eddieferrer, line 219: Unrelated to this experiment, but this isn't used anywhere

src/components/Thanks/ThanksPageShare.vue
```diff
@@ -2,7 +2,6 @@
 	<div>
 		<ShareStepper
 			:lender-name="lender.firstName"
-			:show-lender-name="showLenderName"
```
dyersituations, line 5: Do we want to remove this prop from `ShareStepper`?

emuvente (approved):
Looks great! Thanks for cleaning this up!

src/components/Thanks/ThanksPageShare.vue
eddieferrer: No, I want to use the shareStepper for a guest version, where I will use this prop as `false`

christian14b (approved):
Cool! Thank you for doing this

PR #4694, opened 2023-03-15
Core 1189 experiment assignment create utility for determining experiment assignment w user

dyersituations:
https://kiva.atlassian.net/browse/CORE-1189

- Implemented a new experimentation utility, `assignVersionForLoginId`, for getting consistent assignments per experiment name + login ID combination -> utility not yet used (next changeset)
- New assignment method uses MIT-licensed `Alea` algorithm to replace `Math.random` -> `Alea` gets seeded in order for the assignments to be consistent
- Simplified the experiment settings needs to match the experimentation pattern that was ported to CPS -> no longer needing targets, enabled flag, and start/end dates
- Experiment settings will now only need: name, distribution, and population -> enabling controlled by active_experiments
- Added brute force unit tests, running generator 100 times per test to try to ensure stability -> individual tests still run in 6-12ms with the 100 iterations
- Ported over the unit tests from the existing JavaScript/PHP Alea libraries and extended with more cases (also running the ported tests with 100 iterations each)
- The assignment method unit tests run both with mocked and unmocked Alea library -> mocked to ensure assignment still falls in existing expectations (ported and extended existing assignment tests), and unmocked to ensure that assignment still behaves the same with Alea replacing Math.random

Note that these changes will be ported to CPS after they are stable in `ui` -> probably next week.


src/util/experiment/experimentUtils.js
```diff
@@ -180,6 +181,56 @@ export function assignVersion({
...
+		} else {
+			cutoff += weight * population;
+		}
+		// Exit the loop and return the current version
+		if (marker <= cutoff) return key;
+	}
+
+	// If no version was selected, mark them as 'unassigned' so that they will be re-assigned
+	// if/when the population percent changes.
+	return 'unassigned';
```
emuvente, line 231: I think the default could be `null` or `undefined` instead, because it's easier to compare against and this function automatically handles changes to the population number. The 'unassigned' string isn't needed anymore to force the experiment assignment function to try again.

emuvente (approved):
This is great! Those tests are excellent and that runManyTimes utility is a really useful addition.

roger-in-kiva (approved):
Good!

mcstover (approved):
Rad!

mcstover (approved):

PR #4697, opened 2023-03-15
Core 1104 five dollars notes test setup

roger-in-kiva:
Mixin added to get five dollars notes test assignment in next pages:

- Category page `/lend-by-category/*`
- Lend by Category page `/lend-by-category`
- Lending Home page `/lending-home`
- New filters page `/lend/filter`
- Borrower profile page `/lend/*`
- Checkout page `/checkout`

Mixin is returning `enableFiveDollarsNotes` to show a test planned in other tickets


roger-in-kiva:
> Looks like we'll probably need to pass the prop around a bit more, like to the loan cards, but that can happen in the follow up work.

Agree. It is a different level for each case, better in follow up work that is granulated.

mcstover (approved):

PR #4702, opened 2023-03-16
feat: show the focused share ask for guest checkout for us loans

eddieferrer:
For guests who are checking out with only US loans, show them the focused share ask. Instead of the portfolio link show a "Create account" link that just goes to the normal guest upsell state. 

ACK-553

Screenshots:
Screen Shot 2023-03-16 at 1 15 12 PM
Then clicking on "Create your account" goes to:
Screen Shot 2023-03-16 at 1 15 30 PM

### DIFF
--- a/src/components/Checkout/GuestUpsell.vue
+++ b/src/components/Checkout/GuestUpsell.vue
@@ -9,7 +9,7 @@
 		<form id="guestUpsellForm" action="." @submit.prevent.stop="submit">
 			<kv-base-input
 				name="firstName"
-				class="data-hj-suppress"
+				class="data-hj-suppress tw-mb-2"
 				type="text"
 				v-model.trim="firstName"
 				:validation="$v.firstName"
@@ -21,7 +21,7 @@
 			</kv-base-input>
 			<kv-base-input
 				name="lastName"
-				class="data-hj-suppress"
+				class="data-hj-suppress tw-mb-2"
 				type="text"
 				v-model.trim="lastName"
 				:validation="$v.lastName"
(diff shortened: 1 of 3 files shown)


src/components/Checkout/GuestUpsell.vue
```diff
@@ -9,7 +9,7 @@
 		<form id="guestUpsellForm" action="." @submit.prevent.stop="submit">
 			<kv-base-input
 				name="firstName"
-				class="data-hj-suppress"
+				class="data-hj-suppress tw-mb-2"
```
eddieferrer, line 12: The guest create account form was a little cramped, this looks better, see screenshot

src/components/Thanks/ThanksPageShare.vue
```diff
@@ -263,16 +276,21 @@ export default {
...
 				utm_content: this.utmContent,
 			};
-			if (!this.calculatePeopleQtyToGoal()) {
+
+			if (!this.calculatePeopleQtyToGoal() && !this.isGuest) {
 				if (this.lender.public && this.lender.publicName) {
 					args.lender = this.lender.publicName;
 				}
 				args.funded_share = 1;
 				return getFullUrl(`${base}/invitedby/${this.lender.inviterName}`, args);
```
mcstover, line 285: @emuvente Should we have a check here for `this.lender.inviterName` before attempting to use the invitedby url? Same scenario is below...

mcstover (approved):
I realize my comment about inviterName is current state, so maybe some follow up there later...

dyersituations (approved):

PR #4704, opened 2023-03-17
Core-1198-basketitemtests

SqueakAttack:
These tests do not test everything I think needs to be tested in the basket item components, but they are a start. I have separated more tasks into other related tickets to try and make sure we eventually cover everything.

A few questions since I am not familiar with ideal practices with Jest tests-
-will these automatically run once they are in dev as part of the marketplace tests or do I need to set something up?
-should each of these tests have so many assertions or should I split those up into more tests?

a note- I recall Matt had suggested that I not check if the objects I got were defined if that is all I'm checking (since just getting them means they are there), but unfortunately, the ui repo won't let me push the code up with those unused variables if I just get the objects but then don't use them in another statement.  Sorry for all the extra lines!

### DIFF
--- a/test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
+++ b/test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
@@ -1,0 +1,116 @@
+import BasketItem from '@/components/Checkout/BasketItem';
+import kvAnalytics from '@/plugins/kv-analytics-plugin';
+import VueRouter from 'vue-router';
+import numeralFilter from '@/plugins/numeral-filter';
+import CookieStore from '@/util/cookieStore';
+import { render, within } from '@testing-library/vue';
+import loanReservation from '../../../fixtures/MatchedPromoLoanReservation.json';
+import basketLoanTeams from '../../../fixtures/BasketLoanTeams.json';
+
+// Test that the basket item renders the components it contains:
+// Checkout Item Img/Borrower Link, Remove Basket Item Button, Loan Reservation, Team Attribution, Loan Promo Credits,
+// Loan Price, Remove Basket Item, Matching Text
+
+const emptyComponent = {
+	template: '<div></div>',
+};
+
+describe('BasketItem loan', () => {
+	beforeEach(() => {
+	});
+	afterEach(jest.clearAllMocks);
+
+	it('should contain these components and text', async () => {
+		loanReservation.expiryTime = '2050-09-19T19:02:10Z';
+		const {
+			getByText, getByTestId, getAllByTestId, getAllByRole, getByRole,
+			getByDisplayValue, queryByText, queryByDisplayValue
+		} = render(
+			BasketItem,
+			{
+				provide: {
+					apollo: {
+						readFragment: () => {},
+						query: () => Promise.resolve({}),
+						readQuery: () => {},
+					},
+					cookieStore: new CookieStore(),
+				},
+				routes: new VueRouter(),
+				props: {
+					disableMatching: false,
+					disableRedirects: false,
+					loan: loanReservation,
+					teams: basketLoanTeams,
+				},
+				stubs: {
+					LoanReservation: { ...emptyComponent }
+				}
+			},
+			vue => {
+				vue.use(kvAnalytics);
+				vue.filter('numeral', numeralFilter);
+			}
+		);
+
+		// not a component, but making sure this is showing
+		const LoanTitle = getByText('Twubakane Cb Group in Rwanda');
+		const LoanName = getByTestId('basket-loan-name');
+		expect(LoanTitle).toBe(LoanName);
+
+		// Can't use text 'reservation expires in' to select because it is inside a component rendered
+		// by LoanReservation, which has not actually been rendered here.
+		const LoanReservation = getByTestId('basket-loan-reservation-text');
+		const LoanResEx = LoanReservation.getAttribute('expiry-time');
+		// correct expiry time made it to component
+		expect(LoanResEx).toBe('2050-09-19T19:02:10Z');
+
+		const CheckoutItemImg = getByTestId('basket-loan-image');
+		const BorrowerLink = getByRole('link');
+		expect(BorrowerLink).toBe(CheckoutItemImg);
+		expect(BorrowerLink.href).toBe('http://localhost/#/lend/2444883');
+
+		const LoanPrice = document.getElementById('loan-price');
+		const LoanPriceSelected = LoanPrice.options[LoanPrice.selectedIndex].getAttribute('value');
+		const LoanPriceSelection = getByTestId('basket-loan-price-selector');
+		const LoanPriceDisplay = getByText('$25');
+		const LoanPriceDisplayNext = getByText('$50');
+		// 25 should be selected by default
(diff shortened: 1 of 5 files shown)


test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
```diff
@@ -0,0 +1,116 @@
...
+// Loan Price, Remove Basket Item, Matching Text
+
+const emptyComponent = {
+	template: '<div></div>',
+};
+
+describe('BasketItem loan', () => {
+	beforeEach(() => {
+	});
+	afterEach(jest.clearAllMocks);
```
dyersituations, line 21: This `afterEach` shouldn't be needed in each of these files, as there aren't mocks to clear.

test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
```diff
@@ -0,0 +1,116 @@
...
+		const TeamSelector = getAllByTestId('basket-loan-team-selector');
+		expect(TeamSelector).toBeDefined();
+
+		const RemoveBasketItem = getAllByTestId('removeBasketItem');
+		const BasketLoanInfo = getByTestId('basket-loan-info');
+		const RemoveButton = getAllByRole('button', { name: 'Close' });
+		expect(within(RemoveButton[0]).getByTitle('Remove from cart'));
+		expect(within(RemoveButton[1]).getByTitle('Remove from cart'));
+		expect(RemoveButton[0].classList[1]).toBe('md:tw-hidden'); // mobile version
+		expect(RemoveButton[1].classList[6]).toBe('md:tw-flex'); // tablet and up sizes
```
dyersituations, line 104: Using `classList[6]` might be tough to maintain. Do we need to check if there's a mobile button in this situation?

test/unit/specs/components/Checkout/CheckoutItemTeamDropDown.spec.js
```diff
@@ -0,0 +1,76 @@
+import { render } from '@testing-library/vue';
+import TeamAttribution from '../../../../../src/components/Checkout/TeamAttribution';
```
dyersituations, line 2: This can be `@/components...`

test/unit/specs/components/Checkout/CheckoutItemTeamDropDown.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+	it('should be hidden entirely if there are no teams given', async () => {
+		const {
+			getAllByTestId, getByRole
+		} = render(
+			TeamAttribution,
+			{
+				provide: {
+					apollo: {
+						readQuery: () => {
```
dyersituations, line 61: For these tests with `readQuery` empty, does the rendered component still show all of the elements/data that we need to check? Most of our components are influenced by the data we load.

test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
```diff
@@ -0,0 +1,116 @@
...
+		const LoanPriceDisplayValue = getByDisplayValue('$25');
+		const LoanPriceDisplayValueNext = queryByDisplayValue('$50');
+		expect(LoanPriceDisplayValueNext).toBe(null);
+		expect(LoanPriceSelected).toBe('25');
+		expect(LoanPrice.value).toBe('25');
+		expect(LoanPrice.value).not.toBe('50');
+		expect(within(LoanPrice).getByRole('option', { name: '$50' }));
+		expect(within(LoanPrice).getByRole('option', { name: '$975' }));
+		expect(LoanPriceSelection).toBeDefined();
+		expect(LoanPriceDisplay).toBeDefined();
```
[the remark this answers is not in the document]
mcstover: @SqueakAttack This is what I was trying to describe in our previous convos, that the simplified method should work if you are just checking the "existence" of your target value. I do see a similar usage [here](https://github.com/kiva/ui/blob/main/test/unit/specs/components/Kv/KvPagination.spec.js#L30) so it should be possible in this context too.

test/unit/specs/components/Checkout/CheckoutBasketLoanReservation.spec.js
```diff
@@ -0,0 +1,56 @@
...
+		const CloseLightboxHide = CloseLightbox.getAttribute('aria-hidden');
+		expect(KvLightboxSectionHidden).toBe(null);
+		expect(CloseLightboxHidden).toBe(null);
+		expect(CloseLightboxHide).toBe('true');
+		expect(KvLightboxSection).toBeDefined();
+		expect(ReservationWhyCloseButton).toBeDefined();
+
+		const LoanMessage = getByText('What does it mean that my loan is not reserved?');
+		const WhyLightbox = getByTestId('basket-loan-why-lightbox');
+		const Reservation = getByText('Reservation expires in 4m 59s');
```
eddieferrer, line 49: I don't believe this test will be reliable in all environments. Maybe we can wildcard or not include the seconds part in the selector to account for a possible latency between when Time is defined above and this line of code actually gets run

test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
SqueakAttack: that's fair. Is there a better way to grab this that I missed?

dyersituations: I haven't tested this locally, but something along these lines should work.

```
expect(RemoveButton[1].classList.contains('md:tw-flex')).toBe(true)
```

test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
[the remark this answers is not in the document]
SqueakAttack: done

test/unit/specs/components/Checkout/CheckoutBasketItem.spec.js
SqueakAttack: perfect, thanks! hadn't considered chaining them in that way.

test/unit/specs/components/Checkout/CheckoutItemTeamDropDown.spec.js
SqueakAttack: they all currently pass as they are written. should I be checking additional things?

test/unit/specs/components/Checkout/CheckoutBasketLoanReservation.spec.js
SqueakAttack: That is a good point. I had meant to come back to that and forgot. Thanks!

test/unit/specs/components/Checkout/CheckoutItemTeamDropDown.spec.js
dyersituations: Depends on how the components change when data is applied.

dyersituations: Never mind, these components don't use `readQuery` directly, so the props should be enough for rendering the component.

dyersituations (approved):

PR #4707, opened 2023-03-20
fix: row view on new campaign pages is showing fewer loans than grid …

michelleinez:
…view

### DIFF
--- a/src/components/Contentful/CampaignLoanWrapper.vue
+++ b/src/components/Contentful/CampaignLoanWrapper.vue
@@ -19,7 +19,7 @@
 
 				<campaign-loan-row
 					name="Loan Row"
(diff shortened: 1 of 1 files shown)


src/components/Contentful/CampaignLoanWrapper.vue
```diff
@@ -19,7 +19,7 @@
 
 				<campaign-loan-row
 					name="Loan Row"
-					v-if="componentProps.showLoanRows"
+					v-show="componentProps.showLoanRows"
```
dyersituations, line 22: Just out of curiosity, how does this change help? Theoretically the component should be visible in the same circumstances before and after the change.

dyersituations (approved):

PR #4708, opened 2023-03-21
feat: price array updated for five dollars notes in some pages

roger-in-kiva:
Price array updated for five dollars notes in next pages:

- Category page
- Lend by category page
- lending home page
- filter page
- borrower profile page

### DIFF
--- a/src/components/LoanCards/Buttons/LendCtaExp.vue
+++ b/src/components/LoanCards/Buttons/LendCtaExp.vue
@@ -108,6 +108,7 @@
 <script>
 import {
 	buildPriceArray,
+	build5DollarsPriceArray,
 	isLessThan25,
 	isBetween25And50,
 	isBetween25And500
@@ -132,6 +133,10 @@ export default {
 			default: true
 		},
 		isAdding: {
+			type: Boolean,
+			default: false
+		},
+		enableFiveDollarsNotes: {
 			type: Boolean,
 			default: false
 		}
@@ -199,8 +204,8 @@ export default {
 			// We don't want to open up $5 loan shares for loans with more than $25 at this time
 			// IF we wanted to show this interface on loans with less than 25 remaining they would see the selector
 			const minAmount = parseFloat(this.unreservedAmount < 25 ? this.minNoteSize : 25); // 25_hard_coded
-			// limit at 20 price options
-			const priceArray = buildPriceArray(parseFloat(this.unreservedAmount), minAmount).slice(0, 20);
+			// limit price options
+			const priceArray = this.enableFiveDollarsNotes ? build5DollarsPriceArray(parseFloat(this.unreservedAmount)).slice(0, 28) : buildPriceArray(parseFloat(this.unreservedAmount), minAmount).slice(0, 20); // eslint-disable-line max-len
 			// eslint-disable-next-line
 			if (this.isCompleteLoanActive && !priceArray.includes(Number(this.unreservedAmount).toFixed())) {
 				priceArray.push(Number(this.unreservedAmount).toFixed());
--- a/src/components/LoanCards/Buttons/LendIncrementButton.vue
+++ b/src/components/LoanCards/Buttons/LendIncrementButton.vue
@@ -31,6 +31,7 @@
 import LendButton from '@/components/LoanCards/Buttons/LendButton';
 import {
 	buildPriceArray,
+	build5DollarsPriceArray,
 	isLessThan25,
 	isBetween25And500
 } from '@/util/loanUtils';
@@ -60,6 +61,10 @@ export default {
 			type: Object,
 			default: () => {}
 		},
+		enableFiveDollarsNotes: {
+			type: Boolean,
+			default: false
+		}
 	},
 	computed: {
 		amountLeft() {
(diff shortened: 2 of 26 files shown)


src/util/loanUtils.js
```diff
@@ -75,6 +75,22 @@ export function buildPriceArray(amountLeft, minAmount) {
 	return priceArray;
 }
 
+export function build5DollarsPriceArray(amountLeft) {
+	const limit5 = amountLeft < 50 ? amountLeft : 50;
+	const N_5 = limit5 / 5;
```
dyersituations, line 80: It might be helpful for these variables to be more descriptive. I'm assuming this is number of $5 options? Something like `numberOf5` would be more clear when glancing through the code.

src/util/loanUtils.js
```diff
@@ -75,6 +75,22 @@ export function buildPriceArray(amountLeft, minAmount) {
 	return priceArray;
 }
 
+export function build5DollarsPriceArray(amountLeft) {
```
dyersituations, line 78: This function should get a few unit tests. It is an easily testable method, and this experiment has a lot of app implications, so it would be good to get some stability in before release.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -199,8 +204,8 @@ export default {
 			// We don't want to open up $5 loan shares for loans with more than $25 at this time
 			// IF we wanted to show this interface on loans with less than 25 remaining they would see the selector
 			const minAmount = parseFloat(this.unreservedAmount < 25 ? this.minNoteSize : 25); // 25_hard_coded
-			// limit at 20 price options
-			const priceArray = buildPriceArray(parseFloat(this.unreservedAmount), minAmount).slice(0, 20);
+			// limit price options
+			const priceArray = this.enableFiveDollarsNotes ? build5DollarsPriceArray(parseFloat(this.unreservedAmount)).slice(0, 28) : buildPriceArray(parseFloat(this.unreservedAmount), minAmount).slice(0, 20); // eslint-disable-line max-len
```
dyersituations, line 208: `parseFloat` can be moved out and stored in a different const to simplify this line of code.

Also, where does 28 come from, and I see that's different from the 20 options we had previously.

src/components/LoanCards/Buttons/LendIncrementButton.vue
```diff
@@ -81,7 +86,7 @@ export default {
 			// IF we wanted to show this interface on loans with less than 25 remaining they would see the selector
 			const minAmount = parseFloat(this.amountLeft < 25 ? this.loan.minNoteSize : 25); // 25_hard_coded
 			// cap at 20 prices
-			const priceArray = buildPriceArray(parseFloat(this.amountLeft), minAmount).slice(0, 20);
+			const priceArray = this.enableFiveDollarsNotes ? build5DollarsPriceArray(parseFloat(this.amountLeft)).slice(0, 28) : buildPriceArray(parseFloat(this.amountLeft), minAmount).slice(0, 20); // eslint-disable-line max-len
```
dyersituations, line 89: Since we have this line of code now in 3 components, it should probably be extracted to the `loanUtils` with the other new method and get a few unit tests.

dyersituations:
Sometimes there are a ton of options, sometimes fewer, like this loan. Based on seeing the larger list on some loans, I would expect this dropdown to go to up higher. Loan ID 2536027 on dev.

image

src/components/LoanCards/Buttons/LendCtaExp.vue
roger-in-kiva: Fixed with the function mentioned in the comment below.
Builders are returning all options from 5/25 to the amount left, so, if amount left is greater than 500 we are slicing the array to don't show more options than 500.
For $5 notes we are adding 8 options more than the current array.

src/components/LoanCards/Buttons/LendIncrementButton.vue
roger-in-kiva: Function added.

test/unit/specs/util/loanUtils.spec.js
```diff
@@ -154,4 +154,40 @@ describe('loanUtils.js', () => {
 			expect(result).toEqual(['Agriculture']);
 		});
 	});
+	describe('DropdownPrices', () => {
```
dyersituations, line 157: The describe usually just mentions the tested method, so `getDropdownPriceArray` here.

test/unit/specs/util/loanUtils.spec.js
```diff
@@ -154,4 +154,40 @@ describe('loanUtils.js', () => {
...
+
+		it('should return amount left or closer $25 increment as last option if amount left is less than 500', () => {
+			const unreservedAmountLess500 = 325;
+			const result = getDropdownPriceArray(unreservedAmountLess500, minAmount, enableFiveDollarsNotes);
+			const lastOption = result.length - 1;
+			expect(result[lastOption]).toEqual('325');
+		});
+
+		it('should return 5 as first option if $5 notes test enabled', () => {
+			const FiveDollarsNotes = true;
```
dyersituations, line 181: We usually use naming like `fiveDollarsNotes` or `FIVE_DOLLAR_NOTES` for constants. Or in this case, `true` can just be passed into `getDropdownPriceArray` directly without this variable.

src/util/loanUtils.js
```diff
@@ -75,6 +75,27 @@ export function buildPriceArray(amountLeft, minAmount) {
...
+	if (amountLeft > limit5Notes) {
+		for (let i = 3; i <= numberOf25; i += 1) {
+			priceArray.push(numeral(25 * i).format('0,0'));
+		}
+	}
+	return priceArray;
+}
+
+export function getDropdownPriceArray(unreservedAmount, minAmount, enableFiveDollarsNotes, inPfp = false) {
+	const parsedAmountLeft = parseFloat(unreservedAmount);
```
dyersituations, line 95: You also want a test for passing in a string for `unreservedAmount`, since that value is a string in the DB.

roger-in-kiva: Added!

dyersituations (approved):

PR #4712, opened 2023-03-22
chore: clean up MARS share_card_language experiment

eddieferrer:
* removed scle param and props

ACK-579

The 'b' version was actually already active in the meta description/title. This just removes the associated props and scle parameters. We want to run a very similar experiment but instead of sharing from the thanks, sharing from the borrower/funded borrower profile. That will be in a followup pr, and I will reuse the same basic idea and scle query parameter


dyersituations (approved):

emuvente (approved):
Thanks!

PR #4713, opened 2023-03-22
fix: temp fix for new id field on my endpoint

dyersituations:
- Temp fix while we wait for `id` field to be released for `my` endpoint


emuvente (approved):
Thanks!

PR #4714, opened 2023-03-23
Core 1171 five dollars notes checkout

roger-in-kiva:
$5 notes dropdown options added to checkout page and unit test for a new function

### DIFF
--- a/src/util/loanUtils.js
+++ b/src/util/loanUtils.js
@@ -96,12 +96,12 @@ export function getDropdownPriceArray(unreservedAmount, minAmount, enableFiveDol
 	return (enableFiveDollarsNotes && !inPfp) ? build5DollarsPriceArray(parsedAmountLeft).slice(0, 28) : buildPriceArray(parsedAmountLeft, minAmount).slice(0, 20); // eslint-disable-line max-len
 }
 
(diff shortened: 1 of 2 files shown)


src/util/loanUtils.js
```diff
@@ -96,6 +96,18 @@ export function getDropdownPriceArray(unreservedAmount, minAmount, enableFiveDol
 	return (enableFiveDollarsNotes && !inPfp) ? build5DollarsPriceArray(parsedAmountLeft).slice(0, 28) : buildPriceArray(parsedAmountLeft, minAmount).slice(0, 20); // eslint-disable-line max-len
 }
 
+export function getDropdownPriceArrayCheckout(unreservedAmount, minAmount, enableFiveDollarsNotes) {
```
dyersituations, line 99: Should the first prop instead be named `remainingAmount`?

dyersituations:
It looks like there's a special case on the checkout page that hides the dropdown for amounts under $25. That needs to be changed for the $5 notes experiment.

image

roger-in-kiva:
> It looks like there's a special case on the checkout page that hides the dropdown for amounts under $25. That needs to be changed for the $5 notes experiment.
> 
> image

yeah, makes sense. we can do this in CORE-1107 I think

PR #4719, opened 2023-03-24
feat: experimentation refactor with login ID assignment

dyersituations:
https://kiva.atlassian.net/browse/CORE-1214
https://kiva.atlassian.net/browse/CORE-1215

**Notes**
- I wasn't able to do a final functional testing run after writing unit tests + fixing small edge cases identified with tests, because the dev DB was empty today - I'll do more functional testing once the DB is back
- I won't merge this PR until after the release next week - the goal is to have the refactor on dev + tested by everyone on Denali for the 2 weeks between releases
- Changes will be ported to CPS in a following ticket
- I will be doing an audit of the experiment settings next in dev + prod and updating `active_experiments` according

**Details**
- Initial URL now passed to Apollo init to be able to pull `setuiab` from it
- Experiment assignment mutation removed, since it was used for an old experiment that is no longer running
- Removed mixin related to the removed mutation and old disabled experiment
- Cookie assignments are only stored if assignment forced via `setuiab`
- Cookie assignment logic implemented to respect existing experiment assignments
- Cookie assignment "forces" the assignment - this helps with future assignments via email URL using `setuiab` + keeps assignments from existing experiments
- Removed pulling active experiments, experiment settings, and assigning experiments from initial page load - that logic all exists in the local resolver now
- Went through entire application and updated experiment usage that used to rely on the global assignments - many of those experiments are disabled, but it was less work to get them compliant with this new pattern than to remove
- Removed the unneeded experiment clearing from the Algolia filter page - faster than removing that old page
- Unit tests hopefully cover all user paths and edge cases


eddieferrer:
Looks good, eager to test!

Can we add the ability and also a test case for forcing several experiments simultaneously with setuiab? This has proven to be very useful in testing interconnected experiments with a url like: `setuiab=asd.x&setuiab=xyz.x`

eddieferrer:
This is tangential, but it seems to me that editing the `active_experiments` setting in admin is annoying and error prone since the whole setting is in that itty bitty admin text box

dyersituations:
> This is tangential, but it seems to me that editing the `active_experiments` setting in admin is annoying and error prone since the whole setting is in that itty bitty admin text box

Haha yes. Editing experiments in general is a nightmare in Kiva Admin. More long-term work to come there.

eddieferrer:
> > Can we add the ability and also a test case for forcing several experiments simultaneously with setuiab? This has proven to be very useful in testing interconnected experiments with a url like: `?setuiab=asd.x&setuiab=xyz.x`
> 
> I didn't realize we used that use case (multiple forced experiments at once). That functionality was not ported from the old preFetch pattern, but I'll add a follow-up ticket to implement.

Thank you. Not used often, but when you need it you really need it :)

mcstover (approved):

src/api/apollo.js
```diff
@@ -14,7 +14,8 @@ export default function createApolloClient({
 	kvAuth0,
 	types,
 	uri,
-	fetch
+	fetch,
+	url = '',
```
emuvente, line 18: We might get more utility out of this if it was the current `route` object (or maybe even the entire `router`) instead of just the url string.

emuvente (commented):
This is great work!

src/api/apollo.js
dyersituations: I had assumed that the server wouldn't have access to a router object, but looking again, I see that we do use the router on the server entry as well. It looks like we can move the creation of the router around slightly and get the route object easily passed to the apollo `initState`. That will make it easier to grab the `setuiab` param, as well as other options in the future. I'll add this change to the ticket I'll be doing this week to extend the `setuiab` functionality. Thanks!

PR #4722, opened 2023-03-27
Core 1107 five dollars test less 25

roger-in-kiva:
- $5 notes added when amount left in a loan is less than $25


dyersituations:
It looks like an existing bug, but we might as well address this now. It looks like loans with less than $25 left don't show the "complete loan" state with stars. An example on dev: 2525049

image

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -164,14 +163,6 @@ export default {
 		},
 	},
 	watch: {
-		unreservedAmount(newValue, previousValue) {
```
dyersituations, line 167: Why was this change needed in the experiment component but not previous CTA component?

src/components/LoanCards/GridLoanCard.vue
```diff
@@ -47,7 +47,7 @@
 					:is-selected-by-another="isSelectedByAnother"
 					:is-amount-lend-button="lessThan25"
```
dyersituations, line 48: Do we need the same check here?

```
:is-amount-lend-button="lessThan25 && !enableFiveDollarsNotes"
```

dyersituations:
Amount flashes $25 before showing correct amount in dropdown, noticed in both BP and checkout.

1. Open http://localhost:8888/lend/2528356
2. Add $15
3. Refresh page -> notice $25 shown briefly before "continue to checkout"
4. Open checkout page -> notice $25 shown briefly before $15 selected

dyersituations:
"Lend" shown while adding to basket -> two buttons showing instead of one. I can't get a screenshot because of how fast it happens, but open any BP and click lend -> you'll see two buttons briefly.

dyersituations:
Checkout not showing all options for loan about to expire. This example is ID 2532682. It should show both 30 and 35 in the checkout in this situation.

image

roger-in-kiva:
> Checkout not showing all options for loan about to expire. This example is ID 2532682. It should show both 30 and 35 in the checkout in this situation.
> 
> image

We had a discussion about this here: https://kivahq.slack.com/archives/C03TW5EAFJ5/p1679516500995939

roger-in-kiva:
> It looks like an existing bug, but we might as well address this now. It looks like loans with less than $25 left don't show the "complete loan" state with stars. An example on dev: 2525049
> 
> image

Fixed, thanks!

src/components/LoanCards/Buttons/LendCtaExp.vue
roger-in-kiva: This watcher was there to assign correct amount for `selectedOption`. In this case was replaced by a computed prop having a better performance.

src/components/LoanCards/GridLoanCard.vue
roger-in-kiva: Added

roger-in-kiva:
> Amount flashes $25 before showing correct amount in dropdown, noticed in both BP and checkout.
> 
> 1. Open http://localhost:8888/lend/2528356
> 2. Add $15
> 3. Refresh page -> notice $25 shown briefly before "continue to checkout"
> 4. Open checkout page -> notice $25 shown briefly before $15 selected

Fixed, thanks!

roger-in-kiva:
> "Lend" shown while adding to basket -> two buttons showing instead of one. I can't get a screenshot because of how fast it happens, but open any BP and click lend -> you'll see two buttons briefly.

Fixed in BP and new loan card.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -129,7 +129,7 @@
 											:amount-left="unreservedAmount"
 											@add-to-basket="addToBasket"
 											:complete-loan="isCompleteLoanActive"
-											v-if="isLendAmountButton"
+											v-if="isLendAmountButton && !enableFiveDollarsNotes"
```
dyersituations, line 132: Is this button never needed for $5 notes?

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -778,6 +778,7 @@ export default {
 			return 'tw-transform tw-translate-y-7 md:tw--translate-y-7 lg:tw--translate-y-7';
 		},
 		isLessThan25() {
+			if (this.enableFiveDollarsNotes) return false;
```
dyersituations, line 781: I'd recommend adding a comment above this line as to why this should return false for $5 notes.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -272,6 +263,7 @@ export default {
 			return false;
 		},
 		isLessThan25() {
+			if (this.enableFiveDollarsNotes) return false;
```
dyersituations, line 266: Same comment about adding a comment here for clarity.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -294,6 +286,12 @@ export default {
 		showLendAgain() {
 			return this.isLentTo && !this.isLessThan25;
 		},
+		selectedOption() {
+			if (isBetween25And50(this.unreservedAmount) || isLessThan25(this.unreservedAmount)) {
+				return Number(this.unreservedAmount).toFixed();
+			}
+			return '25';
```
dyersituations, line 293: Should this be a number or should the other return be a string? Is there not an issue returning either a number or string from this computed?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -528,6 +526,22 @@ export default {
 			return hasCredits && isMatchedLoan;
 		});
 		this.matchedText = matchedLoansWithCredit[0]?.loan?.matchingText ?? '';
+
+		this.apollo.query({ query: experimentAssignmentQuery, variables: { id: FIVE_DOLLARS_NOTES_EXP } });
```
dyersituations, line 530: This assignment query should be needed here, since it's called in the preFetch.

src/components/BorrowerProfile/LendCta.vue
roger-in-kiva: yeah, the needed one is above.

src/components/LoanCards/Buttons/LendCtaExp.vue
roger-in-kiva: Previous value in the watcher was string; there is not an issue returning both.

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Is not working if removed

dyersituations: What isn't working? I just tried commenting out this line, and a version was successfully returned by the `readFragment` on the next line. The `preFetch` is called before this `create` method, so the assignment should be in the apollo cache.

roger-in-kiva: It was not working to me when removing; makes sense what you mentioned, lemme try again...

dyersituations (approved):

PR #4726, opened 2023-03-28
feat: localized meta properties experiment

eddieferrer:
ACK-469


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -179,7 +185,7 @@ const preFetchQuery = gql`
 				key
 				value
 			}
-			userContext: uiExperimentSetting(key: "new_users_context") {
```
eddieferrer, line 182: MARS confirmed this experiment setting can be removed here. This is the only reference in the code.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -510,8 +524,28 @@ export default {
 		this.partnerName = loan?.partnerName ?? '';
 		this.partnerCountry = loan?.partner?.countries[0]?.name ?? '';
 		this.lender = data?.my?.userAccount ?? {};
+
+		// ACK-469 Experiment Tracking for generating the scle in utm_campaign
+		const shareLanguageBpExp = getExperimentSettingCached(this.apollo, 'share_language_bp');
```
dyersituations, line 529: The exp key could be put in a const like `LEND_URGENCY_EXP` in this file.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -510,8 +524,28 @@ export default {
...
+		// Tracking for EXP-ACK-469-Apr2023
+		if (shareLanguageBpExp?.enabled) {
```
dyersituations, line 531: We don't need to check `enabled` anymore and won't need to add that to experiment settings moving forward. The `trackExperimentVersion` already checks that the experiment version is defined before logging:

```
if (exp.version && exp.version !== 'unassigned') {
```

dyersituations: Oh yup, the experiment assignment that we add to the component prefetch will get the experiment settings either from cache or async, if needed. And the assignment checks the `active_experiments` before assigning (replacing the enabled check).

dyersituations (approved):

emuvente (approved):

PR #4727, opened 2023-03-28
fix: add exp assignment for matching_highlight

eddieferrer:
Forgot to add the exp assignment query for this experiment. Since this component doesnt prefetch, this is all done in the initialize function
ACK-538


dyersituations (approved):
Thanks for covering this new requirement for experiments (assignment no longer happening in global preFetch)!

PR #4728, opened 2023-03-28
feat: query forced saved in cookie assignment

dyersituations:
https://kiva.atlassian.net/browse/CORE-1238

- Whether the assignment was query forced is now saved to the cookie value
- This is a temporary need while we have experiments active from before this refactor, since moving forward we won't need to store assignments in the cookie unless query forced, but this also isn't a bad addition to the cookie value
- The forced assignment lives regardless of distribution and population changes and will only be disregarded if the experiment is removed from the active_experiments

### DIFF
--- a/src/api/localResolvers/experiment.js
+++ b/src/api/localResolvers/experiment.js
@@ -86,6 +86,7 @@ export default ({ cookieStore, url = '' }) => {
 								: assignVersionForLoginId(experimentSetting, getLoginId(cookieStore)),
 							hash,
 							population,
+							queryForced: currentAssignment.queryForced,
 						};
 
 						// Update the "uiab" cookie if the assignment was forced via the "setuiab" query string param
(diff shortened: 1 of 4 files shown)


mcstover (approved):
Seems like rolling out these changes may "reset" all in progress experiments. Are we ready for that?

dyersituations:
> Seems like rolling out these changes may "reset" all in progress experiments. Are we ready for that?

@mcstover Existing assignments in the cookie will return `false` for `queryForced`, and the hash check is still in place. So if the assignment in the cookie wasn't query forced and the hash hasn't changed, the assignment should remain the same. Is this situation I missed?

There's a test that should cover this situation: `should return current assignment when assignment is forced`

emuvente (approved):
👍

roger-in-kiva (approved):
Looks good!

PR #4731, opened 2023-03-29
Cit 110: Fixing issues based on Casey's comments (migrating, removing, changing and extracting methods)

tiarascoleman:
Had to close the original branch/draft PR since it was outdated!


src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -55,7 +55,8 @@
 		/>
 
 		<div
-			:class="`in-context-login ${isCorporateCampaignPage ? 'tw-text-right' : ''}`"
+			class="in-context-login"
+			:class="{ 'tw-text-right' : !isCorporateCampaignPage }"
```
dyersituations, line 59: This component needs to import the new `isCorporateCampaignPage` method from the `checkoutUtils` file. That same method could be imported to the other components that also have the `isCorporateCampaignPage` method, centralizing that logic.

src/util/routeUtils.js
```diff
@@ -0,0 +1,13 @@
+/**
```
mcstover, line 1: Please move this to `/ui/src/util/urlUtils.js` where we already have similar route/url checks.

tiarascoleman: Sure thing!

mcstover: There is also a test file, `test/unit/specs/util/urlUtils.spec.js` where you could add a simple test for your new method.

src/util/routeUtils.js
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * @param {Object}
+ * @returns {String}
+ */
+
+export default function isCCPage(
+	route = {}
+) {
+	// Fetch route path
+	return route.path.substring(0, 4) === '/cc/';
```
dyersituations, line 12: Since `route` has a `{}` default, this should be:

```
route.path?.substring
```

src/components/Checkout/OrderTotals.vue
```diff
@@ -391,11 +392,11 @@ export default {
 				price: numeral(this.totals.donationTotal)
 			};
 		},
-		isCorporateCampaignPage() {
-			return this.$route.path.substring(0, 4) === '/cc/';
-		}
 	},
 	methods: {
+		ccPageCheck(route) {
+			return isCCPage(route);
```
dyersituations, line 398: Usage could be simplified by instead using `this.$route` in this method instead. And it could then be converted into a computed method.

```
ccPageCheck() {
			return isCCPage(this.$route);
}
```

src/components/CorporateCampaign/CampaignLoanRow.vue
dyersituations, line 1: It looks like this component is still referenced in `CampaignLoanWrapper.vue`

tiarascoleman: Hi Casey! I can't quite see this on my end- are you referring to referencing CampaignLoanRow or the former isCorporateCampaign method?

mcstover: I think he's just saying we should not delete this file...CampaignLoanRow is the carousel version of the loan display... not sure how it ended up deleted in this PR.

tiarascoleman: Oh! Agreed and it being deleted was an accident; I thought commit 05a33de restored the file; is it still not there?

dyersituations: Looks like the file is still deleted in this PR.

michelleinez (approved):

src/components/Checkout/OrderTotals.vue
mcstover: I agree, the existing computed prop can just be altered to the following so names across the file don't need to be changed.

```
isCorporateCampaignPage() {
    return isCCPage(this.$route);
}
```

test/unit/specs/util/urlUtils.spec.js
```diff
@@ -39,4 +39,19 @@ describe('urlUtils.js', () => {
 				.toBe('https://www.facebook.com/dialog/share');
 		});
 	});
+
+	describe('isCCPage', () => {
```
mcstover, line 43: Thank you!

src/components/CorporateCampaign/CampaignLoanRow.vue
mcstover: you should be able to `git checkout origin/main src/components/CorporateCampaign/CampaignLoanRow.vue` to bring the file back into your branch, afterwhich, you'll likely need to `git add` the file and commit it.

PR #4732, opened 2023-03-29
feat: prop added to new loan card to hide more cta in use component

roger-in-kiva:
`KivaClassicBasicLoanCardExp` component updated based in this requirements:

Let’s add the hover state back and remove the “more” cta for the next version of the test. 
Test group
category page, filter page, homepage include new loan card without the “more” (original design)

Control:
original /old loan cards in each of those loan card

Lending Home
by the time this ticket is done, we’ll have data on LH, so we can roll out the test and the loan card there can a/b more vs. no more 

Prop added to control lending home test in the future

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -6,7 +6,7 @@
 		:style="{ minWidth: '230px', maxWidth: cardWidth }"
 	>
 		<div class="tw-grow">
-			<div>
+			<div :class="{ 'loan-card-active-hover': !allSharesReserved }">
 				<!-- Borrower image -->
 				<kv-loading-placeholder
 					v-if="isLoading"
@@ -83,7 +83,7 @@
 								:borrower-count="loanBorrowerCount"
 								:name="borrowerName"
 								:distribution-model="distributionModel"
-								:show-more="true"
+								:show-more="enableMoreCta"
 							/>
 						</router-link>
 					</div>
@@ -230,7 +230,11 @@ export default {
 		enableFiveDollarsNotes: {
 			type: Boolean,
 			default: false
-		}
+		},
+		enableMoreCta: {
+			type: Boolean,
+			default: false
+		},
 	},
 	inject: ['apollo', 'cookieStore'],
 	mixins: [percentRaisedMixin, timeLeftMixin],
@@ -499,10 +503,13 @@ export default {
 
 <style lang="postcss" scoped>
 
-.loan-card-use,
 .loan-card-use:hover,
 .loan-card-use:focus {
-	@apply tw-text-primary tw-no-underline;
+	@apply tw-text-primary;
+}
+
+.loan-card-active-hover:hover .loan-card-use {
+	@apply tw-underline;
 }
 
 .loan-card-progress >>> [role=progressbar] {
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -6,7 +6,7 @@
 		:style="{ minWidth: '230px', maxWidth: cardWidth }"
 	>
 		<div class="tw-grow">
-			<div>
+			<div :class="{ 'loan-card-active-hover': !allSharesReserved }">
```
dyersituations, line 9: `allSharesReserved` doesn't exist on this component

roger-in-kiva: TY!

dyersituations:
There is an issue now where if you hover the loan tags, the loan use is still underlined, but clicking the loan tags doesn't navigate. So we should ask product what the desired behavior is. Do we include the loan tags in the click target (and keep the underline when it's hovered), or do we not underline the loan use when the loan tags are hovered? Having the underline to suggest clickable state and not having the loan tags clickable is confusing to the user.
image

roger-in-kiva:
> There is an issue now where if you hover the loan tags, the loan use is still underlined, but clicking the loan tags doesn't navigate. So we should ask product what the desired behavior is. Do we include the loan tags in the click target (and keep the underline when it's hovered), or do we not underline the loan use when the loan tags are hovered? Having the underline to suggest clickable state and not having the loan tags clickable is confusing to the user. image

Based on slack convo, loan tags should be clickable.

dyersituations:
There are still gaps above and below the loan tags that show the loan use statement underlined but clicking doesn't navigate to the borrower profile. We probably want to wrap the image, loan tags, and loan use statement in a single router-link

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -62,32 +62,38 @@
 				</div>
 
 				<!-- Loan tag -->
-				<loan-tag-v2 v-if="showTags && !isLoading" :loan="loan" :amount-left="amountLeft" />
+				<router-link
+					:to="customLoanDetails ? '' : `/lend/${loanId}`"
+					v-kv-track-event="['Lending', 'click-Read more', 'Photo', loanId]"
```
dyersituations, line 67: Instead of `Photo`, this should be `Tag`

roger-in-kiva: Thanks for the catch!

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -62,32 +62,38 @@
...
+					class="tw-flex hover:tw-no-underline"
+				>
+					<loan-tag-v2 v-if="showTags && !isLoading" :loan="loan" :amount-left="amountLeft" />
```
dyersituations, line 70: The loan tag will show an underline sometimes when clicked due to `focus`. I don't think we have `focus:tw-no-underline` available, so the parent `router-link` probably needs a new class and an entry added to the styles:

```
.loan-card-tag:focus {
  @apply tw-no-underline;
}
```

roger-in-kiva: Looks like `focus:tw-no-underline` is working fine

dyersituations (approved):

PR #4736, opened 2023-03-30
feat: add follow button to scroll to links on us loans

eddieferrer:
We want to replace the save button with a follow button for us loans that just scrolls down the page. See ACK-569 for more details. 

Screenshots:

Screen Shot 2023-03-30 at 2 34 13 PM
Screen Shot 2023-03-30 at 2 32 21 PM

ACK-569

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -228,11 +228,17 @@
 						:class="isLoggedIn ? 'tw-justify-between' : 'tw-justify-end'"
 					>
 						<loan-bookmark
-							v-if="isLoggedIn"
+							v-if="bookmarkVersion === 'bookmark'"
 							data-testid="bp-lend-cta-loan-bookmark"
 							:loan-id="loanId"
 							class="tw-hidden md:tw-inline-block lg:tw-hidden"
 						/>
+						<loan-follow
+							v-if="bookmarkVersion === 'follow'"
+							data-testid="bp-lend-cta-loan-follow"
+							class="tw-hidden md:tw-inline-block lg:tw-hidden tw-mr-2"
+						/>
+
 						<jump-links
 							class="tw-hidden md:tw-block lg:tw-mb-1.5 md:tw-mb-3"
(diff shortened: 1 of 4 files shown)


src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -46,7 +46,10 @@
 					>
 					</div>
 				</div>
-
+			</div>
+			<!-- "Follow" button for US loans links here from SummaryCard and LendCta -->
+			<a name="follow"></a>
```
eddieferrer, line 51: I had to add this anchor destination here, because the rest of this component hasn't loaded yet. This allows us to anchor link here even if the other elements are not in the dom yet

eddieferrer:
Merging this in will break some tests in `kiva-tests` I will modify those in a followup PR

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -793,6 +804,17 @@ export default {
 		},
 		isLendAmountButton() {
 			return (this.lendButtonVisibility || this.state === 'lent-to') && (isLessThan25(this.unreservedAmount)); // eslint-disable-line max-len
+		},
+		bookmarkVersion() {
+			// Display follow for all US loans no matter login state
+			if (this.distributionModel === 'direct') {
```
dyersituations, line 810: Doesn't necessarily need to happen now, but since these magic strings are being used in more than one file, might be worth creating a "fake" enum file for them like with `LoanStatusEnum.js`

dyersituations (approved):

eddieferrer:
Added a feature flag to control this. I believe I won't have to modify the `kiva-tests` tests? Or do they run with production environment feature flags?

mcstover:
@eddieferrer kiva-tests will run against stage so they would remain unaffected until the next db refresh that occurs after this has been turned on in prod. That said, prob worth getting the tests updated at some point in the next couple weeks to accomodate.

dyersituations (approved):
We also have plans to add a URL flag that disables experiments for automated tests.

PR #4738, opened 2023-03-30
fix: active experiment list should be split

dyersituations:
- Through functional testing found that I accidentally wasn't splitting the active experiments list
- Added unit test to cover this case


mcstover (approved):

roger-in-kiva (approved):
👍

PR #4741, opened 2023-03-31
fix: resolve $5 notes issues

dyersituations:
https://kiva.atlassian.net/browse/CORE-1253

- I forgot that the active experiments come back in a JSON string (extra double quotes) - parsing and splitting moved to util method - it meant that the first and last experiment in `active_experiments` was ignored due to extra double quote
- Selected option in `LendCtaExp` couldn't be a computed, since it was bound to a dropdown - an error was being shown locally, and it wasn't adding the right amount to the basket (defaulted to $25) - ported the `watch` code from `LendCta` component, since it should be the same logic needed
- While tracking down the active experiments issue, I went ahead and refactored the $5 notes mixin, since I initially thought it might have to do with timing of assignment (it didn't...), but I kept the refactor in - it makes it so all assignments now happen in `preFetch` for consistency

### DIFF
--- a/src/api/localResolvers/experiment.js
+++ b/src/api/localResolvers/experiment.js
@@ -40,7 +40,7 @@ export default ({ cookieStore, route }) => {
 					}
 
 					// Check if the requested experiment is active
-					if (!activeExperiments.split(',').includes(id)) {
+					if (!activeExperiments.includes(id)) {
 						logFormatter(`Experiment is not in active experiments list: ${id}`, 'warn');
 						return Experiment({ id });
 					}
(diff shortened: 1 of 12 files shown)


roger-in-kiva:
Looks like `Selected option` is not working for quick filter in LH for cases < $25

Screenshot 2023-03-31 at 11 03 03

mcstover (approved):
I'll try to spend some more time early next week directly verifying the updated experiment behaviors.

roger-in-kiva (approved):
Cool!

roger-in-kiva:
> > Looks like `Selected option` is not working for quick filter in LH for cases < $25
> 
> Good catch. I tested that page yesterday, but I must have only seen loans with $25+ left. The solution was to create a method that could be called both on component creation (when data is provided to a new component) and watch (when the component is initially in loading state). Both situations exist on the LH page.

Nice improvement!

PR #4747, opened 2023-04-06
Core 1174 new loan card filter page

roger-in-kiva:
- new loan card test added to filter page
- customized grid added to filter page to show new loan cards as expected
- state updated if test is enable to match number of loans per page

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -77,7 +77,10 @@
 						@updated="handleUpdatedFilters"
 						@reset="handleResetFilters"
 					/>
-					<div class="tw-flex tw-mt-1 tw-items-center tw-flex-wrap">
+					<div
+						class="tw-flex tw-mt-1 tw-items-center tw-flex-wrap"
+						:class="{ 'lg:tw-pl-4' : enableNewLoanCard }"
+					>
 						<p class="tw-hidden lg:tw-inline-block tw-mr-2">
 							{{ totalCount }} Loans
 						</p>
@@ -99,7 +102,11 @@
 						<a class="tw-cursor-pointer" @click="clickZeroLoansReset">start a new search.</a>
 					</p>
 				</template>
-				<kv-grid v-if="enableNewLoanCard" class="tw-grid-cols-1 md:tw-grid-cols-2 tw-mt-2">
+				<!-- eslint-disable max-len -->
+				<div
+					v-if="enableNewLoanCard"
+					class="tw-grid tw-grid-cols-1 md:tw-grid-cols-2 tw-mt-2 tw-mb-4 lg:tw-ml-1.5 lg:tw-px-2.5 tw-gap-x-6 tw-gap-y-4"
+				>
 					<kiva-classic-basic-loan-card-exp
 						v-for="(loan, index) in loans"
 						:key="`new-card-${index}`"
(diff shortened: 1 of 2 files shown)


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -303,6 +330,7 @@ export default {
 				this.trackLoans();
 			}
 		});
+		if (this.enableNewLoanCard) this.updateState({ ...this.loanSearchState, pageLimit: 10, pageOffset: 0 });
```
dyersituations, line 333: Instead of updating the state again, the pageLimit can be set in the data:

```
loanSearchState: { ...getDefaultLoanSearchState(), pageLimit: 10 },
```

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -130,6 +136,18 @@ export default {
 	},
 	created() {
 		this.initializeFiveDollarsNotes();
+
+		const loanCardExpData = getExperimentSettingCached(this.apollo, LOAN_CARD_EXP_KEY);
```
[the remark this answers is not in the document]
roger-in-kiva: Good to know!

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -303,6 +333,7 @@ export default {
 				this.trackLoans();
 			}
 		});
+		// if (this.enableNewLoanCard) this.updateState({ ...this.loanSearchState, pageLimit: 10, pageOffset: 0 });
```
dyersituations, line 336: remove

roger-in-kiva: TY!

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -388,6 +418,9 @@ export default {
 			const filters = convertQueryToFilters(query, allFacets, queryType, pageLimit);
 
 			await updateSearchState(apollo, filters, allFacets, queryType, loanSearchState);
+		},
+		addToBasket(payload) {
```
dyersituations, line 422: I don't think we need this tracking. Definitely check locally, but the new loan card should already be adding the needed event. In general, we also need the loan and loan amount info for ATB events as well (but I think this should be covered already by the new loan card component).

roger-in-kiva: I think Lauren mentioned this as a requirement in the ticket, https://kiva.atlassian.net/browse/CORE-1174
- ensure tracking for add to basket 

Similar like is happening in every section for lending home tho

dyersituations: That comment was referring to ensuring ATB events happened like expected, which they are inside the loan card. But having this extra event isn't bad either. I just think ATB events are generally less useful without the loan and amount info. Find to leave as is, but long-term we can probably remove this event and the similar ones from LH.

dyersituations (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
roger-in-kiva: Agreed, I think is for test metrics... thanks!

dyersituations: I think the reports just check for experiment assignment.

PR #4748, opened 2023-04-06
Campaign guest checkout 2

michelleinez:


src/components/Checkout/KivaCreditPayment.vue
```diff
@@ -1,34 +1,142 @@
...
-		type="submit"
-		id="kiva-credit-submit"
-		class="tw-mb-2"
-		v-kv-track-event="['payment.continueBtn']"
-		title="Checkout using your Kiva credit"
-		@click.native.prevent="validateCreditBasket"
-	>
-		<slot>Complete order</slot>
-	</kv-button>
+	<div>
```
mcstover, line 2: I think we need to keep this component (`KivaCreditPayment`) as is and create a new `KivaCreditGuestPayment` component to contain all of this new logic.

This will allow us to more easily show this button and it's associate inputs more intentionally as needed.

Additionally, The code below that changes the validation method is problematic in that it will break standard Kiva Credit checkout for an existing user.

PR #4752, opened 2023-04-07
feat: large loan card variation

dyersituations:
https://kiva.atlassian.net/browse/CORE-1168

- Large card variation that brings CTA higher
- See left card below, vs old version on the right
- Currently only used in LH recommended carousel

image


roger-in-kiva (approved):
Looks good!

PR #4754, opened 2023-04-10
Core 1180 solution for loan card tags that are truncated

dyersituations:
https://kiva.atlassian.net/browse/CORE-1180

- Turns out that removing the 50% limitation fixes this issue in most cases
- Fixed issue preventing storybook from loading due to folder structure change for portfolio

image


roger-in-kiva:
Nice!
Any example when issue is not fixed?

roger-in-kiva (approved):
Cool!

PR #4757, opened 2023-04-10
fix: use experiment key instead of name

dyersituations:
- Use experiment key instead of name for assignment to match BE


emuvente (approved):
Thanks!

mcstover:
Thank you both for finding and fixing this up!

PR #4758, opened 2023-04-10
refactor: refactor sharing functionality into shared mixin

eddieferrer:
There are currently 3 components which share to social media

- SocialShareV2 (ML thanks, thank-page-v2 layout on the thanks page)
- KvSocialShareButton (Borrower profile)
- ThanksPageShare (mars share layout on the thanks page)

This combines some of the shared functionality of these components into one place.

### DIFF
--- a/src/components/Thanks/ThanksPageShare.vue
+++ b/src/components/Thanks/ThanksPageShare.vue
@@ -97,7 +97,7 @@
 									<a
 										data-testid="share-facebook-button"
 										class="social__btn social__btn--facebook"
-										:href="facebookShareUrl"
+										:href="facebookShareUrl({utmCampaign, utmContent})"
 										v-kv-track-event="
 											['post-checkout', 'share', 'facebook', utmCampaign, loanId]"
 									>
@@ -111,7 +111,7 @@
 										:disabled="copyStatus.disabled"
 										v-kv-track-event="
 											['post-checkout', 'share', 'copy-link', utmCampaign, loanId]"
-										@click="copyLink"
+										@click="copyLink({utmCampaign, utmContent}, copyStatus.text)"
 									>
 										<kv-material-icon
 											name="clipboard"
@@ -123,7 +123,7 @@
 									<a
 										data-testid="share-twitter-button"
 										class="social__btn social__btn--twitter"
-										:href="twitterShareUrl"
+										:href="twitterShareUrl({utmCampaign, utmContent})"
 										target="_blank"
 										rel="noopener"
 										v-kv-track-event="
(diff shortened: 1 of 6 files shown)


src/components/Thanks/ThanksPageShare.vue
```diff
@@ -269,8 +269,6 @@ export default {
 		shareLink() {
 			const base = `https://${this.$appConfig.host}`;
 			const args = {
-				utm_campaign: this.utmCampaign,
```
eddieferrer, line 272: These will be added in the mixin and should not be added here

src/plugins/social-sharing-mixin.js
```diff
@@ -0,0 +1,136 @@
...
+			}
+			window.open(
+				destinationUrl,
+				'intent',
+				// eslint-disable-next-line max-len
+				`scrollbars=yes,resizable=yes,toolbar=no,location=yes,width=${width},height=${height},left=${left},top=${top}`
+			);
+			this.$showTipMsg(thanksText);
+		},
+		twitterShareUrl({ utmCampaign, utmContent }) {
```
eddieferrer, line 49: The 3 functions work the same way. Take the url to be shared, first, add utm params. Then run that url as the input param for adding the params required by the social media platform.

src/plugins/social-sharing-mixin.js
```diff
@@ -0,0 +1,136 @@
+import { getFullUrl } from '@/util/urlUtils';
+import clipboardCopy from 'clipboard-copy';
+
+/* This mixin depends on the component using it having the
+* following data or computed variables available:
+* this.shareMessage and this.shareLink
```
eddieferrer, line 6: I tried to make this as generic as possible for this first pass, but it still depends on shareMessage and shareLink from the component

src/util/urlUtils.js
```diff
@@ -48,10 +48,17 @@ export function isExcludedUrl(urlArray, route) {
  */
 export function getFullUrl(base, args) {
 	if (!args || Object.keys(args).length === 0) return base;
+
+	// remove hash portion of url if present
```
eddieferrer, line 52: Added functionality and additional tests. Function now supports args that are empty/null/undefined. These are skipped in the generation of the query params. Function also now supports hashes. For example `kiva.org#updates` should have query params inserted before the hashes, and produce something like `kiva.org?query1=true#updates`

dyersituations (approved):
The share link methods didn't have unit tests before, but those might be good to add for the new mixin at some point. Some generally complicated logic in the mixin.

PR #4759, opened 2023-04-10
feat: using fallback function when matched loans are less than 3 inst…

roger-in-kiva:
- Using fallback function when matched loans are less than 3 instead of less than 1 in lending home page

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -198,7 +198,7 @@ export default {
 		async getSecondCategoryData() {
 			this.secondCategoryLoans = await this.getMatchedLoans();
 			this.matchedLoansTotal = this.secondCategoryLoans.length;
-			if (this.matchedLoansTotal === 0) this.secondCategoryLoans = await this.getExpiringSoonAlmostFundedCombo(); // eslint-disable-line max-len
+			if (this.matchedLoansTotal < 3) this.secondCategoryLoans = await this.getExpiringSoonAlmostFundedCombo(); // eslint-disable-line max-len
 		},
 		async getExpiringSoonAlmostFundedCombo() {
 			const expiringSoonData = await runLoansQuery(
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -39,7 +39,7 @@
 
 			<!-- Second category row: Matched loans section -->
 			<lending-category-section
-				v-if="secondCategoryActive"
+				v-if="secondCategoryLoans.length > 0"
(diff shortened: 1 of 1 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -198,7 +198,7 @@ export default {
 		async getSecondCategoryData() {
 			this.secondCategoryLoans = await this.getMatchedLoans();
 			this.matchedLoansTotal = this.secondCategoryLoans.length;
-			if (this.matchedLoansTotal === 0) this.secondCategoryLoans = await this.getExpiringSoonAlmostFundedCombo(); // eslint-disable-line max-len
+			if (this.matchedLoansTotal < 3) this.secondCategoryLoans = await this.getExpiringSoonAlmostFundedCombo(); // eslint-disable-line max-len
```
dyersituations, line 201: We actually want to combine the matched with the fallback. So the first 1-2 loans are matched, and the rest are the fallback. Here we are replacing the matched loans results with the fallback.

roger-in-kiva: oh, I see the note 🤦‍♂️, ty

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -198,7 +199,11 @@ export default {
 		async getSecondCategoryData() {
 			this.secondCategoryLoans = await this.getMatchedLoans();
 			this.matchedLoansTotal = this.secondCategoryLoans.length;
-			if (this.matchedLoansTotal === 0) this.secondCategoryLoans = await this.getExpiringSoonAlmostFundedCombo(); // eslint-disable-line max-len
+			if (this.matchedLoansTotal < 3) {
+				const fallbackLoans = await this.getExpiringSoonAlmostFundedCombo();
+				this.secondCategoryLoans = [...this.secondCategoryLoans, ...fallbackLoans].slice(0, 9);
+			}
+			this.secondCategoryActive = true;
```
dyersituations, line 206: I don't think this new flag is needed, since the source of the loans is still `secondCategoryLoans`

roger-in-kiva: yeah, but if the first assignment is less than 9, this is rendered with them and not updated later... other way is to have `secondCategoryLoans` as a computed prop and remove the flag

dyersituations: Use a local variable in that method, then assign at the end.

roger-in-kiva: Good!

dyersituations (approved):

PR #4763, opened 2023-04-11
feat: make the supported button remove loans from the basket

michelleinez:


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -790,6 +790,7 @@ export default {
 				upcCreditRemaining: numeral(this.upcCreditRemaining).format('0.00'),
 				basketLoans: this.basketLoans,
 				promoName: this.campaignPartnerName,
+				removeLoanFromBasket: this.removeLoanFromBasket,
```
mcstover, line 793: Looks like this is defined as a function below. Prob don't need to dupe

src/components/LoanCards/KivaClassicBasicLoanCard.vue
```diff
@@ -300,6 +301,10 @@ export default {
 			type: String,
 			default: '/basket'
 		},
+		customCheckoutButtonAction: {
```
mcstover, line 304: Doesn't seem like this is used in this component. Do you plan to call it somewhere in this file later?

src/components/CorporateCampaign/CampaignLoanRow.vue
```diff
@@ -124,7 +125,7 @@ export default {
 		},
 		checkoutRoute: {
 			type: String,
-			default: '#show-basket'
+			default: ''
```
mcstover, line 128: I think this was previously used to keep the loan card from redirecting. I realize we're now using a new loan card so it may not be needed any longer?

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
dyersituations: Looks like this is a prop wrapper for passing into a component in the template, so all of these props have the same setup. Looks to be by design.

mcstover: oh yea. you're right sorry. This isn't the `data` obj!

src/components/CorporateCampaign/CampaignLoanRow.vue
michelleinez: I removed the redirect and added a @click function that would execute removal of the loan from the basket.  So the lend $25 button will be a toggle button that adds and removes loans from the basket.  This is all according to Sara's new design.

src/components/LoanCards/KivaClassicBasicLoanCard.vue
michelleinez: No, you're right this doesn't belong here anymore.  I just deleted it.

src/components/CorporateCampaign/CampaignLoanRow.vue
mcstover: Ok thanks for checking, this sounds like a new paradigm that we have not used on loan cards.

mcstover: How will one, "activate" the basket overlay if there is no longer a "checkout" button on the loan card or an automatic action to "show" the basket after adding a loan to it?

michelleinez: There is a checkout button in the progress bar section when a campaign is active, but when the campaign is not active there's also the "basket" link at the top nav.  I mentioned that the progress section of the page would not be shown during inactive campaign periods, and Sara was okay with this because this is not the majority use case.  Most people will be checking out loans via campaign pages only when a campaign is active.  And even when the campaign is inactive, accessing the basket is not blocked, in fact it's always visible in the top right section of the page.

dyersituations: It looks like the loan card now get the below button when ATB clicked, and clicking this button removes it from the basket. Is that the desired behavior?
image

michelleinez: Yes

michelleinez: That's the whole point of this PR and aligns with Sara's new design for the page.

dyersituations (approved):

mcstover (approved):

PR #4773, opened 2023-04-17
Core 1273 recommended loans prefetch

roger-in-kiva:
- code fixed in new loan card to get visible recommended loans data cached from parent component

### DIFF
--- a/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
+++ b/src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
@@ -526,6 +526,7 @@ export default {
 			this.loan = cachedLoan;
 			this.isLoading = false;
 		}
+		this.loanCallouts = loanCallouts(this.loan, this.categoryPageName);
 	},
 	watch: {
 		// When loan id changes, update watch query variables
--- a/src/graphql/fragments/loanCardFields.graphql
+++ b/src/graphql/fragments/loanCardFields.graphql
@@ -56,6 +56,12 @@ fragment loanCardFields on LoanBasic {
 	lenders(limit: 0) {
 		totalCount
 	}
+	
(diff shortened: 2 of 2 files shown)


src/graphql/fragments/loanCardFields.graphql
```diff
@@ -56,6 +56,12 @@ fragment loanCardFields on LoanBasic {
 	lenders(limit: 0) {
 		totalCount
 	}
+	
+	# for loan-progress component and lend cta component
+	unreservedAmount @client
```
dyersituations, line 61: We don't want to add these fields are ALL loan cards, just the recommended. This is a shared fragment that applies in many app components.

src/components/LoanCards/KivaClassicBasicLoanCardExp.vue
```diff
@@ -520,12 +515,13 @@ export default {
...
 			apollo: this.apollo,
 			loanId: this.loanId,
-			fragment: loanCardFieldsFragment,
+			fragment: loanCardFieldsExtendedFragment,
 		});
 		if (cachedLoan) {
 			this.loan = cachedLoan;
 			this.isLoading = false;
 		}
+		this.loanCallouts = loanCallouts(this.loan, this.categoryPageName);
```
dyersituations, line 524: This should probably be moved into a method and called in the two places it's needed in this component. Just in case the props of this method call changes, we can change in one place.

src/graphql/fragments/loanCardFieldsExtended.graphql
```diff
@@ -0,0 +1,72 @@
+fragment loanCardFieldsExtended on LoanBasic {
```
dyersituations, line 1: Can you import the other fragment and use it within this fragment?

dyersituations (approved):

PR #4778, opened 2023-04-19
fix: checkout pop up box bug fix

tiarascoleman:
Current error: 
image

I tried other attempts to define showbasket but it would cause the basket to disappear- curious to if this is a spelling mistake (regarding on if events should be kebab case or camel case) or error with linkage?

### DIFF
--- a/src/components/WwwFrame/TheBasketBar.vue
+++ b/src/components/WwwFrame/TheBasketBar.vue
@@ -10,7 +10,7 @@
 	>
 		<!--eslint-enable max-len -->
 		<kv-button
-			:to="basketLink"
+			@click="$emit('show-basket')"
 			v-kv-track-event="['BasketBar','click-Basket']"
 			:class="[
 				{ 'tw-w-full tw-shadow-sm' : !floating },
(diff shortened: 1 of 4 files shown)


src/components/WwwFrame/TheBasketBar.vue
```diff
@@ -10,7 +10,7 @@
 	>
 		<!--eslint-enable max-len -->
 		<kv-button
-			:to="basketLink"
```
mcstover, line 13: This cannot be removed as it powers the basket link on all UI pages outside of the CC page.

src/components/WwwFrame/TheBasketBar.vue
```diff
@@ -10,7 +10,7 @@
...
+			@click="$emit('show-basket')"
```
mcstover, line 13: You could change the basket button to be a js driven click by creating a new method that you call here. In the new method, you would handle emitting the event you need and blocking navigation if this.corporate is true.

You would still need to keep the `to` property in place and set to `/basket`.

src/components/WwwFrame/TheBasketBar.vue
mcstover: @michelleinez @tiarascoleman I did realize that this is shown on our `/cc` pages when in the mobile view. My apologies for thinking it is hidden previously. It would be be nice to get this tweaked so it works for your page and rest of the site.

michelleinez (commented):
This is already fixed on main now, this PR can be closed

PR #4781, opened 2023-04-19
feat: additions to improve lighthouse scores in lending home

roger-in-kiva:
- some aria unique ids, link aria names and `aria-hidden="true"` statements were added to improve lighthouse scores in lh

### DIFF
--- a/src/components/LoanCards/Buttons/LendCtaExp.vue
+++ b/src/components/LoanCards/Buttons/LendCtaExp.vue
@@ -1,9 +1,10 @@
 <template>
-	<div>
-		<form v-if="useFormSubmit" @submit.prevent="addToBasket" class="tw-w-full tw-flex">
+	<div aria-hidden="true">
+		<form v-if="useFormSubmit" @submit.prevent="addToBasket" class="tw-w-full tw-flex" aria-hidden="true">
 			<fieldset
 				class="tw-w-full tw-flex" :disabled="isAdding"
 				data-testid="bp-lend-cta-select-and-button"
+				aria-hidden="true"
 			>
 				<label
 					v-if="hideShowLendDropdown && !isLessThan25"
@@ -12,14 +13,15 @@
 				>
 					Lend amount
 				</label>
-				<div class="amountDropdownWrapper">
+				<div class="amountDropdownWrapper" aria-hidden="true">
 					<kv-ui-select
 						v-if="hideShowLendDropdown && !isLessThan25"
-						id="LoanAmountDropdown"
+						:id="`LoanAmountDropdown_${loan.id}`"
 						class="tw-min-w-12"
 						v-model="selectedOption"
 						v-kv-track-event="['Lending', 'click-Modify loan amount', 'open dialog', loanId, loanId]"
 						@update:modelValue="trackLendAmountSelection"
+						aria-hidden="true"
 					>
 						<option
 							v-for="priceOption in prices"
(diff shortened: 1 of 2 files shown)


src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -1,9 +1,10 @@
 <template>
-	<div>
-		<form v-if="useFormSubmit" @submit.prevent="addToBasket" class="tw-w-full tw-flex">
+	<div aria-hidden="true">
```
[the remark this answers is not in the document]
roger-in-kiva: testing is suggesting to have `aria-hidden` in parents of focusable elements, so, that is the rule... but same question from slack convo, do we want this at this level?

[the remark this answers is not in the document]
roger-in-kiva: We can preserve unique ids only if we want to avoid this...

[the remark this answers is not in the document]
dyersituations: Is there an alternative? I'm seeing this in w3c:
https://w3c.github.io/using-aria/#fourth
```
NOTE
Applying aria-hidden to a parent/ancestor of a visible interactive element will also result in the interactive element being hidden, so don't do this either:

<div aria-hidden="true">
<button>press me</button>
</div>
```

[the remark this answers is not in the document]
dyersituations: Maybe we can revert the `aria-hidden` changes for now and reevaluate later.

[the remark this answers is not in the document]
roger-in-kiva: Yeah, looks like there is not a difference in the score removing all of them
Screenshot 2023-04-19 at 17 17 43

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -7,15 +7,14 @@
 			>
 				<label
 					v-if="hideShowLendDropdown && !isLessThan25"
```
dyersituations, line 9: Do we still need the `for` entry?

roger-in-kiva: I tried to have this dynamically similar to id, but unique id error appears again, having an static? is for sr only.

dyersituations: Does removing `for` cause any issues?

roger-in-kiva: No issues, but not matching with every dynamic id... if I add a dynamic for like dynamic id to match them, score go down and unique id error appears again, do you agree having this static?  like `for="LoanAmountDropdown"` even they are not matching?

dyersituations: Sometimes labels give extra functionality, like click to select the connected radio button. Are we using it for anything like that in this case?

roger-in-kiva: No visible extra functionality, this is just for screen readers... https://tailwindcss.com/docs/screen-readers
it was inherited from original component, so, we can remove it if not needed as visible I think, wdyt?

dyersituations: How about remove the `label` and add `aria-label="Lend amount"` to the `kv-ui-select`

roger-in-kiva: Agreed!

dyersituations (approved):

PR #4788, opened 2023-04-20
CIT-109.75: new standardized function (isCCPage) 

tiarascoleman:
Hopefully the final PR in this eventful trilogy!

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -49,7 +49,7 @@
 			</div>
 		</div>
 		<div
-			v-if="leftoverCreditAllocationLoanId === String(loan.id) && isCorporateCampaignPage"
+			v-if="leftoverCreditAllocationLoanId === String(loan.id) && ccPageCheck($route)"
 			class="tw-w-full
 					md:tw-w-auto
 					md:tw-ml-3
@@ -114,6 +114,7 @@
 </template>
 
 <script>
+import { isCCPage } from '@/util/urlUtils';
 import CheckoutItemImg from '@/components/Checkout/CheckoutItemImg';
 import LoanMatcher from '@/components/Checkout/LoanMatcher';
 import LoanPromoCredits from '@/components/Checkout/LoanPromoCredits';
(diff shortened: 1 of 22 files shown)


src/components/Checkout/OrderTotals.vue
```diff
@@ -391,11 +392,11 @@ export default {
 				price: numeral(this.totals.donationTotal)
 			};
 		},
-		isCorporateCampaignPage() {
-			return this.$route.path.substring(0, 4) === '/cc/';
-		}
 	},
 	methods: {
+		ccPageCheck(route) {
+			return isCCPage(route);
```
dyersituations, line 398: I suggest changing this line to `return isCCPage(this.$route);`, so you can remove the `$route` usage in the templates. My guess is that that could then allow this to be converted to a computed value instead of a method, further simplifying usage to something like `v-if="!ccPageCheck"`

tiarascoleman: Will do! :)

src/components/Checkout/OrderTotals.vue
```diff
@@ -391,11 +392,11 @@ export default {
 				price: numeral(this.totals.donationTotal)
 			};
 		},
-		isCorporateCampaignPage() {
-			return this.$route.path.substring(0, 4) === '/cc/';
-		}
```
mcstover, line 396: I think you should keep this computed prop in place and replace it's contents with `return isCCPage(this.route);`. You would also leave in place all instances of this computed properties name.

The way you have this up now, makes it so the `ccPageCheck` + nested `isCCPage` function get called many times instead of using a computed property which will caluculate 1 time and be reused.

The pattern I'm describing here is also applicable to inContextCheckout and BasketItem components too.

src/components/Checkout/OrderTotals.vue
mcstover: Thanks @dyersituations, I would prefer we go with fewer changes as outlined in my comments.

src/components/Checkout/BasketItem.vue
```diff
@@ -160,6 +160,9 @@ export default {
 		};
 	},
 	computed: {
+		isCorporateCampaign() {
+			return isCCPage(this.route);
```
dyersituations, line 164: This probably needs to be `this.$route`

dyersituations (approved):

mcstover (approved):

PR #4789, opened 2023-04-21
feat: computed prop added for selected option in lend cta exp component

roger-in-kiva:
- computed prop added for selected option in `lendCtaExp` component to reset dropdown when page changing

### DIFF
--- a/src/components/LoanCards/Buttons/LendCtaExp.vue
+++ b/src/components/LoanCards/Buttons/LendCtaExp.vue
@@ -10,7 +10,7 @@
 						v-if="hideShowLendDropdown && !isLessThan25"
 						:id="`LoanAmountDropdown_${loan.id}`"
 						class="tw-min-w-12"
-						v-model="selectedOption"
+						v-model="defaultOption"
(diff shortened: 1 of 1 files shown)


dyersituations:
Here is a screenshot of the second page on LH locally. The second loan defaults to $5, where it should be $45.

image

roger-in-kiva:
@dyersituations best way is having a diff key in the loan card to be re-rendered based on this. 🖖

src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -28,7 +28,7 @@
 				<template v-for="(loan, index) in loans">
 					<kiva-classic-basic-loan-card-exp
 						v-if="enableLoanCardExp"
-						:key="`new-card-${index}`"
+						:key="`new-card-${loan.id + index}`"
```
dyersituations, line 31: While it would be super unlikely for the math to cause conflicts, we probably instead want just in case:

```
:key="`new-card-${loan.id}-${index}`"
```

dyersituations (approved):

PR #4791, opened 2023-04-24
Core 1296 enable flss experimentation on lending home

dyersituations:
https://kiva.atlassian.net/browse/CORE-1296

- Turns out that the recent experimentation refactor broke passing experiment header to gql for new users without an existing `uiab` cookie
- We didn't previously provide the header outside of LBC, and the data team requested it on: /lend/filter, /lending-home, /lend-by-category, and /lend-by-category/*
- The header value and experiment name for tracking were both updated following the data team's request, since this will be an ongoing experimentation setup for the ML backend


mcstover (approved):

roger-in-kiva (approved):
👍

PR #4794, opened 2023-04-25
Core 1276 ab setup user with balance

roger-in-kiva:
- a/b setup added to lending home for relending experiments for users having a positive balance

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -304,7 +304,7 @@ export default {
 
 		this.initializeFiveDollarsNotes();
 
-		const userBalance = Number(this.userInfo?.balance) ?? 0;
+		const userBalance = Number(this.userInfo?.balance ?? 0);
 		// Relending test for users with balance
 		if (userBalance) {
 			const { version } = trackExperimentVersion(
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -127,35 +127,27 @@ export default {
 	apollo: {
 		query: userInfoQuery,
 		preFetch(config, client) {
-			return client.query({
-				query: userInfoQuery
-			}).then(({ data }) => {
-				const userBalance = Number(data?.my?.userAccount?.balance) ?? null;
+			return Promise.all([
+				client.query({ query: experimentAssignmentQuery, variables: { id: EXP_KEY } }),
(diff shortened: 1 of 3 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -119,31 +121,41 @@ export default {
...
-			]).then(() => {
-				const userInfoPromise = client.query({
-					query: userInfoQuery,
-				});
-
-				const recommendedLoansPromise = client.query({
-					query: flssLoansQueryExtended,
-					variables: prefetchedRecommendedLoansVariables
-				});
+			return client.query({
```
dyersituations, line 130: Seeing this implemented, I think we can remove this first query and call `experimentAssignmentQuery` for the new experiment regardless of whether the user has balance. So this code can be removed, and the new `experimentAssignmentQuery` added below.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -299,6 +311,16 @@ export default {
 		}
 
 		this.initializeFiveDollarsNotes();
+
+		// Relending test for users with balance
+		const { version } = trackExperimentVersion(
```
dyersituations, line 316: Then with the change described in the other comment, we can check the balance before calling `trackExperimentVersion`

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -299,6 +303,19 @@ export default {
 		}
 
 		this.initializeFiveDollarsNotes();
+
+		const userBalance = Number(this.userInfo?.balance) ?? 0;
```
dyersituations, line 307: This needs to be `Number(this.userInfo?.balance ?? 0)` instead, since `Number(undefined)` is `NaN`, and `NaN ?? 0` is `NaN`.

roger-in-kiva: nice call!
balance is coming in a string, for validation we need a number I think. It's still working b/c NaN is false when validating. any tho?

dyersituations: `??` is null coalescing. So `NaN ?? 0` equates to `NaN.

The line should be: `const userBalance = Number(this.userInfo?.balance ?? 0);`

roger-in-kiva: thank you!

dyersituations (approved):

PR #4800, opened 2023-04-27
fix: add missing initial click event for braintree type checkout button

mcstover:
We currently don't have an initial click event for a braintree checkout. We have several subsequent events for different actions but not this one... It is a sibling [to this](https://github.com/kiva/ui/blob/main/src/components/Checkout/KivaCreditPayment.vue#L29) Kiva Credit button click.

### DIFF
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -156,6 +156,7 @@ export default {
 	},
 	methods: {
 		submit() {
+			this.$kvTrackEvent('basket', 'Braintree Checkout', 'Button Click');
 			if (this.isGuestCheckout) {
 				this.$v.$touch();
 				if (!this.$v.$invalid) {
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

SqueakAttack (approved):
thanks!

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -156,6 +156,7 @@ export default {
 	},
 	methods: {
 		submit() {
+			this.$kvTrackEvent('basket', 'Braintree Checkout', 'Button Click');
```
gwilson253, line 159: To confirm - we're deviating from the updated event tagging scheme to be consistent with the related existing tagging. This isn't unreasonable, but from an analytics standpoint the gains from consistency are countered by making new non-conforming events. I would propose `this.$kvTrackEvent('basket', 'click', 'braintree-checkout-button');`, but I don't feel strongly about it.

gwilson253 (approved):

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: I'm happy to update this one seeing as it's new!

PR #4801, opened 2023-04-27
chore: add .nvmrc to support nvm

hobbsh:
Adds `.nvmrc` to support `nvm`


.nvmrc
```diff
@@ -0,0 +1 @@
+v16.19.1
```
mcstover, line 1: We're currently [using 16.15.0](https://github.com/kiva/marketplace-web-ui-ci/blob/main/resources/org/kiva/marketplaceWebUiCi/ui/Dockerfile#L2) but it would be good to match these up with the latest version... so prob fine here to set this value so long as we update the linked dockerfile and verify it too.

mcstover (approved):

.nvmrc
mcstover: It actually looks like 16.20.0 is out now too..

mcstover: As an aside... I've been running 16.17.0 locally...

dyersituations: Looks like I'm running 16.15 to match docker...

mcstover:
https://github.com/kiva/marketplace-web-ui-ci/pull/76

.nvmrc
dyersituations: Looks like @mcstover just updated ui to 16.20

mcstover: I'm proposing we updated to the latest version and have linked the PR to update the Dockerfile for our Ui images below.

dyersituations (approved):

emuvente (approved):

mcstover (approved):
Please proceed!

PR #4806, opened 2023-05-02
fix: hotjar suppression classes for PII on impact dashboard MARS-389

emuvente:
This suppresses the same data as on the old portfolio page.


mcstover (approved):
ty!

christian14b (approved):

PR #4807, opened 2023-05-02
fix: email verification form not appearing for every user

michelleinez:
The basketLoans variable needs to be populated first before we can run all of these basket validation checks that open up the appropriate form.

### DIFF
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -1097,10 +1097,6 @@ export default {
 
 				this.setAuthStatus(this.kvAuth0?.user);
 
-				if (this.$route.hash === '#show-basket') {
-					this.showBasket();
-				}
-
 				this.updateBasketState();
 			});
 		},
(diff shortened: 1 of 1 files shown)


src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1241,6 +1237,10 @@ export default {
 				}
 			});
 
+			if (this.$route.hash === '#show-basket') {
+				this.showBasket();
+			}
+
```
mcstover, line 1243: This may be better placed within the `then` block of the `validateBasket` call below, I would suggest after `this.balanceLeftoverCredits();` on line 1268 as the auth status could be useful for the showBasket call as well. This way we'd have scaffolded all the basket data/user state and know we've validated the basket before potentially showing it.

michelleinez: ok done.

mcstover (approved):

mcstover:
Do note that this moves the showBasket call later in the sequence of events but will also make it part of other sequences so we may see some duplication of the call...

dyersituations (approved):

PR #4810, opened 2023-05-03
Core 1286 relending multiple atc logic

roger-in-kiva:
- relending experience added to lending home page to show and add multiple loans to cart based on conditions referred in this ticket
https://kiva.atlassian.net/browse/CORE-1286


src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -194,6 +202,14 @@ export default {
 			return this.loan?.minNoteSize ?? '';
 		},
 		unreservedAmount() {
+			if (this.enableRelendingExp) {
+				if (this.enableFiveDollarsNotes) {
+					if (this.userBalance > 20) return '25';
+					return '5';
+				}
+				if (this.loan.unreservedAmount > 25) return '25';
+				return this.loan?.unreservedAmount ?? '';
```
dyersituations, line 211: I think it's dangerous to change the meaning on `unreservedAmount` here. We have a decent amount of logic around what "unreserved" means. I think it would be safer to instead change the `amountLeft` passed to the `LendAmountButton` based on whether the experiment is applied.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -324,9 +318,33 @@ export default {
...
+			{ id: 0 }, { id: 0 },
+			{ id: 0 }, { id: 0 },
+			{ id: 0 }, { id: 0 },
+		];
+
+		let relendingArray = [
+			...cachedRecommendedLoans,
+		];
+
+		/* eslint-disable max-len */
```
dyersituations, line 334: Please use curlies instead of disabling the linting rule for the entire file

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -324,9 +318,33 @@ export default {
...
+		if (this.enableFiveDollarsNotes) {
+			if ((userBalance > 0 && userBalance < 10) || (userBalance > 20 && userBalance < 50)) relendingArray = relendingArray.slice(0, 2);
```
dyersituations, line 336: Unfortunately we aren't guaranteed that these loans will all have at least $25 remaining unreserved. This means that in `LendCtaExp.vue` we also need to check/fallback to the unreserved amount, like you have in the `multipleAmount` amount method. With the logic in `multipleAmount`, there could be a button with a larger amount than what is unreserved for a loan (and wouldn't align with the number calculated in `multipleAmount`.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -324,9 +318,33 @@ export default {
...
+			if ((userBalance > 10 && userBalance < 15) || (userBalance > 50 && userBalance < 75)) relendingArray = relendingArray.slice(0, 3);
+		} else {
+			if (userBalance > 0 && userBalance < 50) relendingArray = relendingArray.slice(0, 2);
+			if (userBalance > 50 && userBalance < 75) relendingArray = relendingArray.slice(0, 3);
+		}
+
+		this.userBalance = userBalance;
+		this.recommendedLoans = this.enableRelendingExp ? relendingArray : recommendedArray;
```
dyersituations, line 344: I would rename `recommendedLoans` to something more generic, like `firstRowLoans` so we don't confuse the two uses.

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -125,10 +133,39 @@ export default {
...
 			return '336px';
 		},
+		multipleAmount() {
+			let amount = 0;
+			for (let index = 0; index < this.totalLoans; index += 1) {
+				const loan = this.loans[index];
+				const { unreservedAmount } = loan;
+
+				if (this.enableFiveDollarsNotes) {
+					amount += this.userBalance > 20 ? 25 : 5;
```
dyersituations, line 143: The `unreservedAmount` needs to be checked here to, like in the `else`

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -125,10 +133,39 @@ export default {
...
 			this.$emit('add-to-basket', payload);
+		},
+		addMultipleLoans() {
+			for (let index = 0; index < this.totalLoans; index += 1) {
+				const { unreservedAmount } = this.loans[index];
+				const key = `relending-${index}`;
+
+				let amount = '';
+				amount = unreservedAmount > 25 ? '25' : unreservedAmount;
+				if (this.enableFiveDollarsNotes) amount = '5';
```
dyersituations, line 165: While unlikely to be under $5, we still need to check `unreservedAmount` here.

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -37,8 +38,11 @@
 						:use-full-width="true"
 						:large-card="isLargeCard"
 						:enable-five-dollars-notes="enableFiveDollarsNotes"
-						@add-to-basket="addToBasket"
 						class="tw-h-full"
+						:ref="`relending-${index}`"
```
dyersituations, line 42: This can be a more generic ref, like `loan-card-${index}`

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -125,10 +133,39 @@ export default {
...
+		}
 	},
 	methods: {
 		addToBasket(payload) {
 			this.$emit('add-to-basket', payload);
+		},
+		addMultipleLoans() {
+			for (let index = 0; index < this.totalLoans; index += 1) {
+				const { unreservedAmount } = this.loans[index];
+				const key = `relending-${index}`;
```
dyersituations, line 161: Extract this string template into a method that can be used here and in the template.

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -125,10 +133,39 @@ export default {
...
+
+				let amount = '';
+				amount = unreservedAmount > 25 ? '25' : unreservedAmount;
+				if (this.enableFiveDollarsNotes) amount = '5';
+
+				this.$refs[key][0].addToBasket(amount);
```
dyersituations, line 167: What does `[0]` here signify?

src/components/LoanFinding/LendingCategorySection.vue
roger-in-kiva: TY!

src/components/LoanFinding/LendingCategorySection.vue
roger-in-kiva: TY!

src/components/LoanFinding/LendingCategorySection.vue
roger-in-kiva: it means that data is wrapped in an array for every ref object.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -196,6 +204,17 @@ export default {
 		unreservedAmount() {
 			return this.loan?.unreservedAmount ?? '';
 		},
+		amountLeft() {
```
dyersituations, line 207: Looks like we still need to check the `unreservedAmount` before returning 5/25 in this method.

roger-in-kiva: You are so fast! 😛

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -196,6 +204,17 @@ export default {
...
+			if (this.enableRelendingExp) {
+				if (this.enableFiveDollarsNotes) {
+					if (this.userBalance > 20) return Number(this.unreservedAmount) > 25 ? '25' : this.unreservedAmount;
+					return this.unreservedAmount > 5 ? '5' : this.unreservedAmount;
+				}
+				if (this.unreservedAmount > 25) return Number(this.unreservedAmount) > 5 ? '5' : this.unreservedAmount;
```
dyersituations, line 213: Do we need this line? If so, what's the use case?

roger-in-kiva: yeah. my bad, this is for case outside $5 notes test ...

dyersituations (approved):

PR #4815, opened 2023-05-04
Core 1290 new matched loans position relending test

roger-in-kiva:
- matched loans section moved up to quick filters for relending test
- copy updated matched loans section for test
- data prefetched avoiding async behavior in this section

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -36,7 +36,6 @@
 
 				<!-- Second category row: Matched loans section -->
 				<lending-category-section
-					v-if="secondCategoryLoans.length > 0"
 					:title="secondCategoryTitle"
 					:subtitle="secondCategorySubtitle"
 					:loans="secondCategoryLoans"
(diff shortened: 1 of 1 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -147,9 +156,27 @@ export default {
 					variables: prefetchedRecommendedLoansVariables
 				});
 
+				const matchedLoansPromise = client.query({
```
dyersituations, line 159: Why are we prefetching the second row? Is it ever visible above the fold when the page first loads?

roger-in-kiva: I think it depends of screen resolution, its ever visible for the test for myself

Screenshot 2023-05-04 at 14 19 47

dyersituations: How much do the extra prefetch queries affect the Lighthouse scores locally?

roger-in-kiva: Nice call! Ill test it in a while...

dyersituations (approved):

PR #4816, opened 2023-05-04
feat: tracking event added to multiple add to cart button un lending …

roger-in-kiva:
- tracking event added to multiple add to cart button un lending home

### DIFF
--- a/src/components/LoanFinding/LendingCategorySection.vue
+++ b/src/components/LoanFinding/LendingCategorySection.vue
@@ -170,6 +170,7 @@ export default {
 
 				this.$refs[key][0].addToBasket(amount);
 			}
+			this.$kvTrackEvent('event-tracking', 'add-all-to-basket', 'relending-lending-home');
 		},
 		loanCardKey(index) {
 			return `loan-card-${index}`;
--- a/src/components/LoanFinding/LendingCategorySection.vue
+++ b/src/components/LoanFinding/LendingCategorySection.vue
@@ -160,6 +160,7 @@ export default {
 			this.$emit('add-to-basket', payload);
 		},
 		addMultipleLoans() {
+			const { multipleAmount } = this;
 			for (let index = 0; index < this.totalLoans; index += 1) {
 				const { unreservedAmount } = this.loans[index];
 				const key = this.loanCardKey(index);
@@ -170,7 +171,7 @@ export default {
 
 				this.$refs[key][0].addToBasket(amount);
 			}
-			this.$kvTrackEvent('loan-card', 'add-to-basket', 'relending-lending-home-add-all');
+			this.$kvTrackEvent('loan-card', 'add-to-basket', 'relending-lending-home-add-all', this.userBalance, multipleAmount); // eslint-disable-line max-len
 		},
 		loanCardKey(index) {
 			return `loan-card-${index}`;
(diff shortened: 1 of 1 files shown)


src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -170,6 +170,7 @@ export default {
 
 				this.$refs[key][0].addToBasket(amount);
 			}
+			this.$kvTrackEvent('event-tracking', 'add-all-to-basket', 'relending-lending-home');
```
eddieferrer, line 173: The action `add-all-to-basket` is not on our list of web event actions. I would suggest: 
'loan-card', 'add-to-basket', 'relending-lending-home-add-all'

But I don't know what this looks like, a screenshot would help, see: https://docs.google.com/spreadsheets/d/1pbxxYDfeOkrGhFi8HWl5K-vCPXaqO3bXS1Io3E_SDKc/edit?pli=1#gid=1801894844
for analytics actions and events

roger-in-kiva: Right! Lauren asked to be like that, but I think your suggestion should be useful for them. I'll ping her about it.

Thanks @eddieferrer !

roger-in-kiva: @eddieferrer only thing is that is not a loan card, is button in the top of the section adding multiple loans to the basket... agree having `event-tracking` or what do you suggest for this?

roger-in-kiva: Screenshot 2023-05-04 at 13 57 26

dyersituations: The action does trigger the loan card action, so using the loan-card signifier should be appropriate enough.

dyersituations (approved):

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -170,6 +170,7 @@ export default {
 
 				this.$refs[key][0].addToBasket(amount);
 			}
+			this.$kvTrackEvent('loan-card', 'add-to-basket', 'relending-lending-home-add-all');
```
dyersituations, line 173: Actually it would probably be good to add two more props, the comma-separated list of loan IDs, and the comma-separated list of loan amounts.

roger-in-kiva: Nice suggestion!

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -170,6 +170,8 @@ export default {
 
 				this.$refs[key][0].addToBasket(amount);
 			}
+			const { multipleAmount } = this;
```
dyersituations, line 173: I meant storing the constant before calling `addToBasket`

roger-in-kiva: Oh, Any concerns triggering event before `for` statement?

dyersituations: Not the tracking event, but storing `multipleAmount` in a constant above the loop before calling `addToBasket` in the loop. The `multipleAmount` return changes based on the `unreservedAmount`, which could be influenced by adding the loan to the basket.

eddieferrer (approved):

PR #4822, opened 2023-05-09
feat: fixes needed for relending experiment

dyersituations:
https://kiva.atlassian.net/browse/CORE-1355

Here are the notes from the ticket:
1. Extra “lend again” button if you lent to a loan that had more than $75 left and it shows up in the carousel again
2. Adding all loans to basket w/ $5 notes ends up adding them all at $5 instead of $25 (when your balance is higher, like $50)
3. Adding single when button says “Lend $5" will add a $25 dollar loan to the basket

Solution:
- Refactored logic around showing different CTA components, since it was somewhat confusing before - helped fix #1 
- Changed what `amountToLend` means - helped fix #2 + #3
- Fixed issue where carousel would cut off card drop shadow when only 2 loans + pager hidden
- Fixed random issue where `null` was passed to loan tags - might just a HMR issue, but it's addressed now


roger-in-kiva (approved):
Looks good, thanks!

mcstover (commented):
👍

PR #4826, opened 2023-05-09
feat: there was an edge case where a loan could be cleared

dyersituations:
- While testing relending exp, I would occasionally get a couple loan cards with blank data in the second row
- Looks like the initial load of the second row provides loan IDs of `0` due to lazy loading
- There seems to have been a race condition between the initial request and the one triggered by watching the loan ID


roger-in-kiva (approved):
Nice fix, thanks!

PR #4830, opened 2023-05-11
feat: add endorsements to borrower profile

eddieferrer:
ACK-610

If a loan has a comment, display a pill up top with the count. Then the why it matters section will now be a carousel, where each comment, with author is a slide, and the current why it matters will be the last slide. 

Loan with comments:
Screen Shot 2023-05-11 at 12 01 57 PM
Screen Shot 2023-05-11 at 12 01 50 PM
Screen Shot 2023-05-15 at 12 09 43 PM
hover state:
Screen Shot 2023-05-15 at 12 10 30 PM

Loan without comments (no change)
Screen Shot 2023-05-11 at 12 01 39 PM
Screen Shot 2023-05-11 at 12 01 26 PM

- We're planning on adding author avatars later. 
- This handles anonymous profiles gracefully, since the comments api call will return "Anonymous" as author name
- updated kv-ui-elements version as well
- This is currently the only use of KvPill, should this component be in styleguide?


src/components/BorrowerProfile/CommentsAndWhySpecial.vue
```diff
@@ -0,0 +1,145 @@
...
+							comments {
+								values {
+									id
+									authorName
+									body
+								}
+							}
+						}
+					}
+				}`,
```
mcstover, line 128: Why does this query need to differ from the one above? Could you define it once and reuse even if there are unused fields in one case vs the other?

src/components/BorrowerProfile/SummaryCard.vue
```diff
@@ -42,9 +42,31 @@
...
+						href="#bp-comments-jump-link"
+						class="tw-text-black hover:tw-text-white kvpill-wrapper"
+						v-kv-track-event="[
+							'borrower-profile',
+							'click',
+							'jump-link',
+							'comments-pill'
+						]"
+					>
+						<kv-pill
```
mcstover, line 55: This is similar to the KvCallout... not sure if you can override the bg color on that though.
https://main--608b4cf87f686c00213841b1.chromatic.com/?path=/story/kvloancallouts--default

So it "might" be usable, though it would be good to have something more reusable for this pill style.

Where did this BG color come from? Why?

mcstover (approved):

dyersituations (approved):

src/components/BorrowerProfile/SummaryCard.vue
eddieferrer: Yeah its kind of like KvCallout, I think it's more like the other component on the BP, the SummaryTag component. I will combine KvPill and Summary tag into 1, since the only change would be the background color.

eddieferrer: These blues are one-off colors 😖

mcstover: colors lolz ;) If you see opportunity to an appropriate kv-component please create a ticket and go for it, but it doesn't need to block your work here, could just be something to build and swap in once ready.

dyersituations (approved):

mcstover (approved):

PR #4831, opened 2023-05-11
Core 1363 race condition when adding all to basket

dyersituations:
https://kiva.atlassian.net/browse/CORE-1363

- Adding "all" loans in relending experiment synchronously
- Button shows "adding to basket" while waiting for async calls
- Add all no longer navigates to checkout
- Add a simple hover state to the "add all" button
- Fixed bug where lend amount button wouldn't appear if the loan had less than $25, wasn't in the relending experiment, and $5 notes was running


roger-in-kiva (approved):
Ty!

PR #4832, opened 2023-05-12
Fix loan cta again

dyersituations:
- Noticed that the CTA for $5 notes was incorrect after recent changes, it would show the lend amount button, where $5 notes should have the dropdown for all situations but the relending carousel


roger-in-kiva (approved):
Tested locally and seems like working fine. Nice catch!

PR #4835, opened 2023-05-15
feat: lend dropdown mobile update

egoroza:
updates mobile view to use the "Categories" dropdown instead of "Explore loans"

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -90,9 +90,10 @@
 				<template v-else>
 					<div
 						class="header
-							tw-grid tw-gap-x-2.5 lg:tw-gap-x-4 tw-items-center"
+							tw-grid lg:tw-gap-x-4 tw-items-center"
 						:class="{
-							'header-lend-menu-button-exp': lendMenuButtonExp,
+							'tw-gap-x-2.5 ': !lendMenuButtonExp,
+							'header-lend-menu-button-exp tw-gap-x-1': lendMenuButtonExp,
 							'header-lend-menu-button-exp-visitor': lendMenuButtonExp && isVisitor,
 							'header--mobile-open': searchOpen,
 							'mobile-lend-menu-button-exp': searchOpen && lendMenuButtonExp,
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -112,33 +112,34 @@
 						</div>
 
 						<template v-if="lendMenuButtonExp">
```
roger-in-kiva, line 114: Is this experiment disabled?

dyersituations: Yes it's currently disabled, but we'll enable it for this new variation.

dyersituations: It's not enabled on dev.

roger-in-kiva:
Looks like login button is a little bit tight... is that okay?

Screenshot 2023-05-15 at 14 35 26

dyersituations:
> Looks like login button is a little bit tight... is that okay?
> 
> Screenshot 2023-05-15 at 14 35 26

Good call, but what resolution is that? How does it look on iPhone setting?

roger-in-kiva:
> > Looks like login button is a little bit tight... is that okay?
> > Screenshot 2023-05-15 at 14 35 26
> 
> Good call, but what resolution is that? How does it look on iPhone setting?

Yeah is for iphone 12 setting
Screenshot 2023-05-15 at 14 54 03

dyersituations:
> Yeah is for iphone 12 pro setting

Ah good to know. Maybe we can reduce some of the padding between the items?

roger-in-kiva:
> > Yeah is for iphone 12 pro setting
> 
> Ah good to know. Maybe we can reduce some of the padding between the items?

I think so, have a smaller gap there and maybe consider to decrease space between logo and Categories menu. For logged in users, looks fine but we are not showing About menu and some hidden icons. So, ensure there is a correct spacing for both cases.

Screenshot 2023-05-15 at 15 03 05

roger-in-kiva:
Looking good for logged out users!
Screenshot 2023-05-18 at 9 09 49

But seems like for logged in users gap is smaller than screenshots above becoming a non-finger friendly ui...
Screenshot 2023-05-18 at 9 13 18

PR #4838, opened 2023-05-16
fix: upc credit bug fix

tiarascoleman:
Referenced ticket: https://kiva.atlassian.net/browse/CIT-130

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,7 @@
 		"memcached": "^2.2.2",
 		"memjs": "^1.2.2",
 		"minimist": "^1.2.6",
+		"node": "^16.15.0",
 		"numeral": "^2.0.6",
 		"passport": "^0.4.1",
 		"passport-auth0": "^1.3.1",
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -1424,7 +1424,8 @@ export default {
 			this.basketBalancing = true;
 			const LCALoanId = this.cookieStore.get('lcaid');
 			// Check if there is already a loan id that has unspent credit allocated to it
-			if (LCALoanId) {
+			// Also checks if the loan matches to the correct campaign
+			if ((LCALoanId) && (this.campaignPartnerName === this.promoFundDisplayName)) {
 				// Try getting that loan from the basket
 				const basketLCALoan = this.basketLoans.find(loan => String(loan?.id) === LCALoanId);
 				if (basketLCALoan?.price) {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -96,6 +96,7 @@
 		"memcached": "^2.2.2",
 		"memjs": "^1.2.2",
 		"minimist": "^1.2.6",
+		"node": "^16.15.0",
```
mcstover, line 99: please remove these changes to package + package-lock

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -1424,7 +1424,8 @@ export default {
 			this.basketBalancing = true;
 			const LCALoanId = this.cookieStore.get('lcaid');
 			// Check if there is already a loan id that has unspent credit allocated to it
-			if (LCALoanId) {
+			// Also checks if the loan matches to the correct campaign
+			if ((LCALoanId) && (this.campaignPartnerName === this.promoFundDisplayName)) {
```
mcstover, line 1428: The `promoFundDisplayName` prop doesn't exist in this component so you won't be able to check against it here. That said, we should not just create this b/c where it does exist it is derived from the same property as the `campaignPartnerName` computed property.

I'm not certain here but I think you mean to compare the active campaign `campaignPartnerName` in the page with the promo credits that are applied to the basket. If this is the case, the existing `prioritizedTargetCampaignCredit` computed property may have what you need. Once you have a campaign loaded in your browser and have added an item to you basket you can use the Vue inspector to see the values available on `prioritizedTargetCampaignCredit`.

tiarascoleman: Thanks Stover! After creating the draft PR, I was also presuming that `prioritizedTargetCampaignCredit` computed prop was the way to go and wanted to still see if the current `if-statement` was leading to the right solution.  Curious on if the current `if-statement` should transition to `if ( (LCALoanId) && (this.campaignPartnerName === this.prioritizedTargetCampaignCredit)` or something else? Also what are your thoughts on the current location for the conditional check- should it be moved later in LCALoanId function?

mcstover: I think `prioritizedTargetCampaignCredit` returns an object so you'll need to inspect the state and then use the item from that object that is appropriate to compare against `campaignPartnerName`.

mcstover: As for placement, if you move your check below where `basketLCALoan` is defined you can also check for and pull any campaign data off of this object instead.

PR #4842, opened 2023-05-22
fix: z-index fix for qf

egoroza:
adjusts z-index so that elements are stacked properly on pages with quick filters


dyersituations (approved):

egoroza:
also....  i wasn't seeing the other issue for the large loan cards since the background is the same color on mine 🤔 

Screen Shot 2023-05-22 at 5 07 51 PM

dyersituations:
> also.... i wasn't seeing the other issue for the large loan cards since the background is the same color on mine 🤔

Oh looks like the page background changed. I guess we're OK for now.

roger-in-kiva (approved):

PR #4843, opened 2023-05-23
refactor: remove foundation classes from thanks share

eddieferrer:
* Also standardize share button interactions -- make them buttons and enable small share pop up window on click and working toast messages

ACK-637

The result looks basically the same, the main content area is slightly wider because of the grid differences:

screencapture-localhost-8888-checkout-thanks-2023-05-23-11_46_56

### DIFF
--- a/src/components/Checkout/SocialShareV2.vue
+++ b/src/components/Checkout/SocialShareV2.vue
@@ -9,10 +9,12 @@
 			</p>
 		</div>
 		<div class="share__social social">
-			<a
+			<button
 				data-testid="share-facebook-button"
 				class="social__btn social__btn--facebook"
-				:href="facebookShareUrl({utmCampaign, utmContent})"
+				@click="showSharePopUp(
+					facebookShareUrl({utmCampaign, utmContent}),
+					'Thanks for sharing to Facebook!')"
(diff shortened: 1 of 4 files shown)


src/components/Checkout/SocialShareV2.vue
```diff
@@ -477,36 +477,6 @@ $loan-triangle-size: rem-calc(12);
 				transition: fill 0.25s ease-in;
 			}
 		}
-
-		&--success {
```
eddieferrer, line 481: These success/error classes are from a bygone era, they are not used anywhere

src/components/Checkout/SocialShareV2.vue
```diff
@@ -22,41 +24,39 @@
 			>
 				<kv-icon name="facebook-round" title="Facebook" class="social__icon" />
 				<span>Share</span>
-			</a>
-			<a
+			</button>
+			<button
```
eddieferrer, line 28: Here an on the Thanks share, I made these button actual button elements and they use the showSharePopUp utility so behavior is the same

dyersituations (approved):

mcstover (approved):

PR #4846, opened 2023-05-24
feat: guest checkout gated by feature flag in contentful

michelleinez:


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -168,7 +172,22 @@ export default {
...
+			// Set the default checkout validation method
+			let validationMethod = this.validateGuestBasket;
+			const validationPayload = {
+				email: this.email,
+				emailUpdates: this.emailUpdates,
+			};
+			// If promo guest checkout is enabled, use the promo guest checkout validation method.
+			// This method validates the lender email for promo first before running the guest checkout method
+			// in checkout utils.
+			if (promoGuestCheckoutEnabled) {
```
mcstover, line 184: need to set this as a prop here and pass it in from the parent.

mcstover: Oh or just use the `isGuestCheckout` which I think you're already passing this value to!

michelleinez: This is just a draft for you to see the direction I'm taking this in, but I know the code was not in a state where it was functioning perfectly quite yet.

I've since added the prop and done a bunch of other things, still just trying to ensure that the guest checkout is working while features are being gated by this flag.  Credit card payment seems to be going through just fine, but Kiva Credit payments are having trouble at the moment.

src/components/Checkout/KivaCreditGuestPayment.vue
```diff
@@ -0,0 +1,186 @@
...
+						this.showCheckoutError(errorResult);
+						this.$emit('checkout-failure', errorResult);
+					}
+				}).catch(errorResponse => {
+					this.$emit('updating-totals', false);
+					console.error(errorResponse);
+				});
+		},
+		validateCampaignEmail() {
+			const emailRegex = new RegExp(`^[A-Za-z0-9._%+-]@${promoFundEmail}`);
```
mcstover, line 181: We'll need to remove the `promoFundEmail` variable here and use a regex that validates for any email address instead of a partner specific domain.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -905,17 +908,25 @@ export default {
...
+			const promoCampaignNameFallback = this.promoData?.promoFund?.displayName ?? null;
+			return this.prioritizedTargetCampaignCredit?.promoFund?.displayName ?? promoCampaignNameFallback;
 		},
 		verificationRequired() {
-			if (this.promoData?.managedAccount?.isEmployee && this.promoData?.managedAccount?.formId) {
+			if (this.promoData?.managedAccount?.isEmployee
+				&& this.promoData?.managedAccount?.formId
+				// If promo guest checkout is enabled, we don't need formassembly verification
+				// because the promo guest checkout workflow will take care of email verification.
+				&& !this.promoGuestCheckoutEnabled) {
```
mcstover, line 919: @michelleinez I don't think this is entirely true and we may need to have an extra check here. The original check for formId was b/c that was a requirement in order to execute verification. If we're still going to allow form assembly for some accounts we'll probably need an additional computed property to signify which type of verification the account will use. ie. maybe something like `useFormAssebmly()` which checks for the formId. 

Additionally, the `isEmployee` fields tells us if verification is required. In otherwords we shouldn't be showing an email collection field for existing users if `isEmployee` is false. For Guest users we would need to skip the email verification check if `isEmployee` is false.

mcstover: I think we should begin to default to using the new email verification when `isEmployee` is true, activating our new inputs and api check when there is no formId returned on the managed account.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -168,7 +195,22 @@ export default {
...
+			// Set the default checkout validation method
+			let validationMethod = this.validateGuestBasket;
+			const validationPayload = {
+				email: this.email,
+				emailUpdates: this.emailUpdates,
+			};
+			// If promo guest checkout is enabled, use the promo guest checkout validation method.
+			// This method validates the lender email for promo first before running the guest checkout method
+			// in checkout utils.
+			if (this.isGuestCheckout) {
```
dyersituations, line 207: Should we be checking `this.isGuestCheckout` at the beginning of this method, and either wrapping the entire method content or returning early?

src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -69,21 +69,68 @@
...
+						v-if="promoGuestCheckoutEnabled"
+						variant="secondary"
+						:href="registerOrLoginHref"
+					>
+						Continue as an existing user
+					</kv-button>
+				</div>
+				<div class="tw-col-start-4 tw-col-end-7">
+					<p class="tw-grid-row">
+						For existing users, we recommend you log in with your work email address.
```
dyersituations, line 93: Is this component used in a place where a user might use a non-work-address? Asking because of the callout in this copy.

src/components/Checkout/KivaCreditGuestPayment.vue
```diff
@@ -0,0 +1,190 @@
...
+				id="email"
+				class="data-hj-suppress tw-mb-2 tw-w-full"
+				@focus="$kvTrackEvent(
+					'basket',
+					'click-email-receipt-field',
+					'Where should we email your receipt?'
+				)"
+			/>
+			<p v-if="$v.email.$error" class="input-error tw-text-danger tw-text-base tw-mb-2">
+				Valid email required.
```
dyersituations, line 26: One error message has a period, the other doesn't, is that expected?

src/components/Checkout/KivaCreditGuestPayment.vue
```diff
@@ -0,0 +1,190 @@
...
+		}
+	},
+	data() {
+		return {
+			email: null,
+			termsAgreement: false,
+			emailUpdates: false,
+		};
+	},
+	validations: {
```
dyersituations, line 141: Just out of curiosity, where are these `validations` used?

src/components/Checkout/InContext/InContextCheckout.vue
mcstover: Agree, we should condition this block on `promoGuestCheckoutEnabled`

src/graphql/mutation/checkout/validateLenderEmailForPromo.graphql
```diff
@@ -0,0 +1,9 @@
+mutation validateLenderEmailForPromo($lenderEmailAddress: String!, $managedAccountId: Int!, $promoFundId: Int!) {
+    validateLenderEmailForPromo (
```
dyersituations, line 2: Does the warning showing here in GH need to be addressed?

dyersituations:
Is there a way to test `promoGuestCheckoutEnabled` enabled locally? And if so, is there a particular checkout flow that we should be testing?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: I think this is supposed to be chekcout `promoGuestCheckoutEnabled` instead. If we're in this method, we already know we're in the guest checkout flow based on the method above.

This switch is between the mutation used to validate a promo with it's special conditions vs standard guest checkout.

src/graphql/mutation/checkout/validateLenderEmailForPromo.graphql
mcstover: Yes, I've created a ticket to make sure the new service is available before this code is operationalized...

mcstover:
> 

There's a feature flag passed down from the contentful page data. I'll ping you with info.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
michelleinez: Why can't we just disable the contentful flag to make the UI revert to the old behavior, in which case this check would fail because promoGuestCheckout would be disabled?

michelleinez: That's what would re-enable formassembly.  So if we need formassembly, we just do that...

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
michelleinez: I'm not sure I understand what you're asking me here...  What do you mean by wrapping?
And why would I return early?  This is to set the method that is used in the lines below it so returning early would mean not calling the function below...

src/components/Checkout/InContext/InContextCheckout.vue
michelleinez: I'll forward this question to Sara from design.  Her designs show this as the copy though.

src/components/Checkout/KivaCreditGuestPayment.vue
michelleinez: vuelidate -- I recommend taking a look at their jsfiddle to understand the usage: https://vuelidate.js.org/

src/graphql/mutation/checkout/validateLenderEmailForPromo.graphql
michelleinez: @mcstover have you seen this error before?  I don't really understand what the issue is exactly...  The code is making the appropriate API call it seems so I don't get why the linter is yelling at me.  It's just a warning though, not an error.

  2:5  warning  Cannot query field "validateLenderEmailForPromo" on type "Mutation"  graphql/template-strings

src/components/Checkout/KivaCreditGuestPayment.vue
michelleinez: Alright I added a period to the other sentence.

src/components/Checkout/KivaCreditGuestPayment.vue
dyersituations: Thanks for the info, I didn't realize we used vuelidate.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
dyersituations: I think @mcstover has more specific info that might be more appropriate. Basically what I was curious about was in this method for guest checkout, there is a check for `isGuestCheckout`. My assumption is that this method should only run if `isGuestCheckout` is `true`, so hence checking in the method before running any logic. But it sounds like maybe the check might just need to be a check on `promoGuestCheckoutEnabled`?

src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -69,21 +69,68 @@
...
+				:promo-name="promoName"
+				@refreshtotals="$emit('refresh-totals')"
 				@updating-totals="setUpdatingTotals"
 				@checkout-failure="handleCheckoutFailure"
 			/>
 
 			<checkout-drop-in-payment-wrapper
-				v-else
+				v-if="!showKivaCreditButton
+					&& continueAsGuest"
```
mcstover, line 126: This shouldn't be gated on `continueAsGuest`. With this in place the Braintree drop-in interface will now show up with payment is needed, blocks checkout.

src/components/Checkout/InContext/InContextCheckout.vue
michelleinez: So it sounds like for these cases where the sponsor is not requiring work addresses, we can omit this copy -- we would still need to have a check for this placed somewhere, which we can make sure we add before enabling the feature flag here but I don't think it should block this PR from going in right now.

michelleinez: This is what Sara said ^

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
michelleinez: So "isGuestCheckout" is the flag that enables the guest checkout user experience for the drop in payment wrapper.  The drop in payment wrapper is a component that gets used across kiva, not just for campaign pages.
promoGuestCheckoutEnabled is the flag that gets set in contentful that is **_specific to campaign pages_** so I don't want there to be a check for that in a generalized component.  I do set "isGuestCheckout" based on promoGuestCheckoutEnabled though in the parent component.

michelleinez: image
See here ^

dyersituations: Thanks for the info. It looks like the component has two props with the same value. Can those be consolidated?

src/components/Checkout/InContext/InContextCheckout.vue
michelleinez: We gate all payment methods with a button to select which payment method is shown.  In the case that "continue as a guest" is shown, THEN we show the checkout-drop-in-payment wrapper.  That's what this flag does.

michelleinez: It doesn't mean that there's a case where the drop in payment wrapper is inaccessible.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
michelleinez: oh, thanks yeah that's not supposed to be there.

michelleinez: oh wait, it is...
so sometimes is-guest-checkout will be the same check as promo-guest-checkout-enabled, but in other cases it might not be.  For example, if we are on kiva marketplace and not a corp campaign page, and a drop in payment wrapper is needed and it should follow the guest checkout, but not the promo-specific campaign copy.

michelleinez: If I were to consolidate this, then the component wouldn't be able to tell the difference between a promo guest checkout workflow vs. a kiva marketplace guest checkout workflow.

dyersituations: Thanks for the info. It seems like maybe `isGuestCheckout` is simply potentially a confusing name. If it only controls "guest checkout user experience for the drop in payment wrapper", the name might be more appropriate something along the lines of `isGuestCheckoutForDropIn`

mcstover (approved):

mcstover (approved):

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: We'll have a look at this in some follow up work we're doing to see if we can make things more clear.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
mcstover: We'll also do some more work here to more fully deprecate FA in the near future, but for now the feature flag gates the alternative.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -324,6 +366,10 @@ export default {
 					return kivaBraintreeResponse;
 				});
 		},
+		isValidEmailFormat() {
+			const emailRegex = new RegExp(`^[A-Za-z0-9._%+-]@${this.refs.email}`);
```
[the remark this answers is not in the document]
mcstover: We'll be refactoring this check before release as well. It's also behind the feature flag.

PR #4849, opened 2023-05-25
feat: make suggested loan comment a placeholder

eddieferrer:
We found that everybody was just leaving our prefilled comments and all the comments looked the same which made them look spammy. 

This leaves the comment empty so users can enter their unique comment. With a placeholder in place. Also adds validation incase the API returns a success but failed to add comment. Also disables the button until the comment contains text.


mcstover:
We used to have a small link/button in the share interface that would inject a suitable share message with the loan data prefilled. Maybe we could re-introduce something like that...

mcstover (approved):

dyersituations:
> We used to have a small link/button in the share interface that would inject a suitable share message with the loan data prefilled. Maybe we could re-introduce something like that...

Hmm... ChatGPT?

PR #4855, opened 2023-06-01
fix: styles and share link logic

christian14b:
Small fixes in design and replicate thanks page logic for share link


src/components/Thanks/ThanksPageDonationOnly.vue
```diff
@@ -21,9 +21,9 @@
 							allowfullscreen
 						></iframe>
 					</div>
-					<p class="tw-my-2">
+					<h2 class="tw-text-h2 tw-my-2">
```
mcstover, line 24: Is the class here needed seeing as we're using an h2?

mcstover (approved):

PR #4856, opened 2023-06-01
feat: set default dropdown on new loan cards based on session cookie

egoroza:
As a part of the $5 notes exp, this PR sets the dropdown default for users based on a session cookie, and depending on the `utm_campaign` query parameter:

- Top Lenders: set their default dropdown to $5
- Base Lenders: set their default dropdown to the highest amount of Kiva credit they can lend, rounded DOWN to nearest multiple of 5. (ex: if they have $18, it would round down to $15, so these lenders will have $15 as their default value in the dropdown)

Screen Shot 2023-05-31 at 9 26 26 AM
Screen Shot 2023-06-01 at 8 25 09 AM
Screen Shot 2023-06-01 at 8 48 41 AM
Screen Shot 2023-06-01 at 8 53 44 AM
Screen Shot 2023-05-31 at 9 26 18 AM


src/util/loanUtils.js
```diff
@@ -317,3 +317,50 @@ export function loanCallouts(loan, categoryPageName) {
...
+/**
+ * Enables session cookie
+ *
+ * @param	{String}	cookieName	The cookie name
+ * @param	{Object}	cookieStore The cookie store
+ * @param	{String}	expDate		The expiration
+ * @returns {void}
+ */
+export function enableCookie(cookieName, cookieStore, expDate) {
+	if (!cookieStore.get(cookieName)) {
```
dyersituations, line 330: We probably don't want this check here, since if the user clicks the email link again with the UTM, we'll want to set a new expiration date to 24 hours in the future.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -625,6 +633,12 @@ export default {
 					this.matchingHighlightExpShown = version === 'b';
 				}
 			}
+		},
+		getDropdownERL() {
```
dyersituations, line 637: This method can also be moved into the utils file, being passed the active cookie name.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -625,6 +633,12 @@ export default {
...
+			if (this.activeCookie !== '') {
+				return getCookieDropdown(this.activeCookie, this.userBalance, this.unreservedAmount);
+			}
+			return '25';
```
dyersituations, line 641: It looks like this default fallback could be incorrect in some cases, for example, if the unreserved amount is less than 25.

src/util/loanUtils.js
```diff
@@ -317,3 +317,50 @@ export function loanCallouts(loan, categoryPageName) {
...
+	return '';
+}
+
+/**
+ * Get CTA dropdown default value based on cookie
+ * @param 	{String} 	cookieName
+ * @param	{Number}	balance
+ * @returns	{Number}	Default selected value
+ */
+export function getCookieDropdown(cookieName, balance, unreservedAmount) {
```
dyersituations, line 360: We have a test file for this util, and these methods should get some new unit tests.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -635,6 +649,9 @@ export default {
 		},
 		unreservedAmount(newValue, previousValue) {
 			// set initial selected value for sub 25 loan if shown
+			if (this.enableFiveDollarsNotes && this.activeCookie !== '') {
```
dyersituations, line 652: It looks like this `watch` doesn't have the same logic as `mounted`. Here `selectedOption` is only affected by one of the UTM campaign use cases, where `mounted` is affect by both. Maybe the logic in `mounted` needs to be extracted and used in both the `watch` and the `mounted`.

src/util/loanUtils.js
```diff
@@ -317,3 +317,50 @@ export function loanCallouts(loan, categoryPageName) {
...
+	if (cookieName === 'erl-five-notes-base') {
+		const val = Math.floor(balance / 5) * 5;
+		return (unreservedAmount < balance || val === 0) ? 5 : val;
+	}
+	return 5;
```
dyersituations, line 365: We should check that `unreservedAmount`:

`return unreservedAmount > 5 ? 5 : unreservedAmount`

src/util/loanUtils.js
```diff
@@ -317,3 +317,50 @@ export function loanCallouts(loan, categoryPageName) {
...
+}
+
+/**
+ * Get CTA dropdown default value based on cookie
+ * @param 	{String} 	cookieName
+ * @param	{Number}	balance
+ * @returns	{Number}	Default selected value
+ */
+export function getCookieDropdown(cookieName, balance, unreservedAmount) {
+	if (cookieName === 'erl-five-notes-base') {
```
dyersituations, line 361: Instead of comparing against magic strings, we can store these constants in the top of this util file and use where needed.

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -798,6 +815,29 @@ export default {
 	},
 	mounted() {
 		this.createWrapperObserver();
+		const erlCampaign = this.$route.query.utm_campaign;
```
dyersituations, line 818: Most of this logic can be moved to the utils file and shared in both components. The new method can just return the value needed for `selectedOption`

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -798,6 +815,29 @@ export default {
...
+
+		const topUpLenders = 'topup-vb-balance-MPV1';
+		const baseLenders = 'base-vb_balance_MPV1';
+		this.activeCookie = isErlCookieActive(this.cookieStore);
+
+		if (this.enableFiveDollarsNotes) {
+			if (this.activeCookie !== '') {
+				this.selectedOption = this.getDropdownERL(this.userBalance, this.unreservedAmount);
+			} else if (erlCampaign) {
+				if (erlCampaign === topUpLenders) {
```
dyersituations, line 832: We can't guarantee the case of URLs, so compare these strings with `toUpperCase`

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -798,6 +815,29 @@ export default {
...
+					this.selectedOption = '5';
```
dyersituations, line 833: We'll want to check reserved amount here as well, just in case.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -204,12 +217,19 @@ export default {
 		},
 		amountToLend() {
 			if (this.enableRelendingExp) {
-				if (this.enableFiveDollarsNotes && this.userBalance <= 20) {
-					return Number(this.unreservedAmount) > 5 ? '5' : this.unreservedAmount;
+				if (this.enableFiveDollarsNotes) {
+					if (this.userBalance <= 5) {
+						return Number(this.unreservedAmount) > 5 ? '5' : this.unreservedAmount;
+					} if (this.activeCookie !== '') {
```
dyersituations, line 223: The `if` should be on a newline

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -204,12 +217,19 @@ export default {
...
+						return this.getDropdownERL();
+					}
 				}
 				return Number(this.unreservedAmount) > 25 ? '25' : this.unreservedAmount;
 			}
-			return (this.isLessThan25 && !this.enableFiveDollarsNotes) ? this.unreservedAmount : this.selectedOption;
+			if (this.enableFiveDollarsNotes && this.activeCookie !== '') {
+				return this.cookieDropdownValue;
```
dyersituations, line 230: Looks like `cookieDropdownValue` isn't defined.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -204,12 +217,19 @@ export default {
...
+			}
+			return this.isLessThan25 ? this.unreservedAmount : this.selectedOption;
```
dyersituations, line 232: Should this line still be the original with the $5 notes check? See below:

```
return (this.isLessThan25 && !this.enableFiveDollarsNotes) ? this.unreservedAmount : this.selectedOption;
```

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -173,6 +180,12 @@ export default {
...
 		},
+		getDropdownERL() {
+			if (this.activeCookie !== '') {
+				return getCookieDropdown(this.activeCookie, this.userBalance, this.unreservedAmount);
+			}
+			return '25';
+		}
 	},
 	watch: {
 		unreservedAmount(newValue, previousValue) {
```
dyersituations, line 191: Does the `unreservedAmount` `watch` need the same checks that the other CTA component has?

src/components/BorrowerProfile/LendCta.vue
egoroza: this should be the same logic as LendCtaExp right? that watch function is cleaner

dyersituations: It's hard to say without digging in. These two CTA components are being used in very different parent components.

dyersituations: Okay it looks like the two CTAs actually have very similar logic, once just uses a separate method. We couldn't make the logic identical between the two CTAs without some pretty significant functional testing, but it's definitely still an option. Because like you said, the logic should be similar.

src/util/loanUtils.js
egoroza: I might add to a separate ticket so we can hotfix tomorrow

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -798,6 +815,21 @@ export default {
 	},
 	mounted() {
 		this.createWrapperObserver();
+		const campaign = this.$route.query.utm_campaign;
+		const sessionTimestamp = new Date();
```
dyersituations, line 819: `sessionTimestamp` is no longer used in this method.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -293,6 +312,23 @@ export default {
 			return this.state === 'refunded' || this.state === 'expired';
 		},
 	},
+	mounted() {
+		const campaign = this.$route.query.utm_campaign;
+		const sessionTimestamp = new Date();
```
dyersituations, line 317: `sessionTimestamp` is no longer used in this method.

src/util/loanUtils.js
```diff
@@ -317,3 +325,68 @@ export function loanCallouts(loan, categoryPageName) {
...
+	if (cookieStore.get(TOP_LENDERS_COOKIE)) {
+		return TOP_LENDERS_COOKIE;
+	}
+	return '';
+}
+
+/**
+ * Get the CTA dropdown value based on cookie
+ */
+export function getDropdownERL(activeCookie, userBalance, unreservedAmount) {
```
dyersituations, line 347: Use `Erl` instead of `ERL` for consistency with the other method naming.

src/components/LoanCards/Buttons/LendCtaExp.vue
```diff
@@ -204,8 +218,13 @@ export default {
 		},
 		amountToLend() {
 			if (this.enableRelendingExp) {
-				if (this.enableFiveDollarsNotes && this.userBalance <= 20) {
-					return Number(this.unreservedAmount) > 5 ? '5' : this.unreservedAmount;
+				if (this.enableFiveDollarsNotes) {
+					if (34 <= 5) {
```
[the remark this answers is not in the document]
egoroza: `this.unreservedAmount <= 5`

was testing a use case lol forgot to replace back

PR #4859, opened 2023-06-06
feat: support kiva button on header

christian14b:
Add Support Kiva Button on header. This works comes along with 
image

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -318,7 +318,7 @@
 										</li>
 										<li>
 											<router-link
-												to="/about/due-diligence"
+												to="/donate/supportus"
 												v-kv-track-event="['TopNav','click-Support-Kiva']"
 											>
 												Support Kiva
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -222,7 +222,11 @@
 								v-show="!isMobile"
 								to="/borrow"
 								data-testid="header-borrow"
-								class="header__borrow header__button tw-flex"
+								class="header__borrow"
+								:class="{
+									'tw-hidden': !isVisitor,
+									'header__button !tw-hidden md:!tw-flex': isVisitor
+								}"
 								v-kv-track-event="['TopNav','click-Borrow']"
 							>
 								Borrow
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -365,28 +365,17 @@
 							</router-link>
 
 							<!-- Log in Link -->
-							<template v-if="isVisitor">
-								<kv-button
-									variant="secondary"
-									v-if="isMobile"
-									class="tw-bg-white tw-whitespace-nowrap"
-									:to="loginUrl"
-									data-testid="header-log-in"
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -748,8 +786,19 @@ export default {
 			hasLentBefore: this.cookieStore.get(hasLentBeforeCookie) === 'true',
 			hasDepositBefore: this.cookieStore.get(hasDepositBeforeCookie) === 'true',
 		});
+		window.addEventListener('resize', _throttle(() => {
+			this.determineIfMobile();
+		}, 200));
+	},
+	beforeDestroy() {
+		window.removeEventListener('resize', _throttle(() => {
```
dyersituations, line 794: This probably needs to use a method like in the other PR, rather than the throttle (unless we check that the throttle will be identified correctly as the same method, which it might, it depends on its implementation).

dyersituations:
Looks like the unit tests encountered errors. Please run locally and address.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -321,18 +316,25 @@
 												Due diligence
 											</router-link>
 										</li>
+										<li>
+											<router-link
+												to="/about/due-diligence"
```
emuvente, line 321: this should link to `/donate/supportus`

src/components/WwwFrame/TheHeader.vue
emuvente: +1 Put the `_throttle` call inside the `determineIfMobile` method instead, then just that method needs to be added/removed as the event listener.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -363,20 +365,32 @@
 							</router-link>
 
 							<!-- Log in Link -->
-							<kv-button
```
emuvente, line 366: Rather than swapping between a `kv-button` and an `a`, you could try changing the styling of the button using the `variant` prop:
```
:variant="isMobile ? 'secondary' : 'link'"
```
and apply whatever tailwinds classes are necessary to get the right styling after that.

src/components/WwwFrame/TheHeader.vue
christian14b: Thanks, I completely miss this

christian14b:
@mcstover @dyersituations @emuvente any final comments to these work before merging? I would like to QA this and solve any errors that might appear. Please, let me know!

mcstover (approved):

src/components/WwwFrame/TheHeader.vue
```diff
@@ -210,35 +212,29 @@
...
 							:class="{
-								'tw-gap-2.5': !lendMenuButtonExp,
+								'tw-gap-0.5': isMobile,
+								'tw-gap-2.5': !lendMenuButtonExp && !isMobile,
 								'tw-gap-1.5': lendMenuButtonExp,
 							}"
 						>
 							<!-- Borrow -->
 							<router-link
+								v-show="!isMobile"
```
dyersituations, line 222: Yesterday Lauren noticed that "Borrow" was appearing for logged in users on some pages (since CPS already has these changes). That isn't desired, even though it was in the mocks, like you mentioned.

dyersituations (approved):

PR #4860, opened 2023-06-06
feat: live loan card image Kiva Classic styling option MARS-385

emuvente:
Live Loan card styling option using `img2` in the request url to get a Kiva Classic card roughly matching what's on /lending-home.

- Adds support for drawing Material Design Icons in the live loan canvas
- Adds tests for new loan card loan use
- Refactors loan callouts so that the live loans can use them

live-loan-card-classic-3


mcstover (approved):
This looks great!

christian14b (approved):

server/util/live-loan/live-loan-draw.js
```diff
@@ -154,14 +186,180 @@ module.exports = async function draw(loanData) {
...
+			const useMaxLines = 2;
+			const useMaxLineHeight = 17 * 1.5 * classicResizeFactor;
+			const useText = getLoanUse({ ...loanData, version: 2 });
+			ctx.fillStyle = kivaColors.textPrimary;
+			ctx.font = `400 ${17 * classicResizeFactor}px "Kiva Post Grot"`;
+			wrapText(ctx, useText, useXPos, useYPos, useWidth, useMaxLines, useMaxLineHeight);
+		});
+
+		// Loan callouts
+		tracer.trace('loan-callouts', () => {
```
dyersituations, line 271: Theoretically a loan with two loan callouts might have the pills cutoff as they extend outside the card, but it's so uncommon that I couldn't quickly find an example of a current loan with 2 long pills. Just something worth mentioning, since it was an edge case we discussed in the Vue app. Probably something that can be addresses if it ever comes up...

dyersituations (approved):

server/util/live-loan/live-loan-draw.js
emuvente: Yeah, I think that's small enough of an edge case and such a minor visual issue that we can deal with it if it ever comes up and we decide it's a priority.

server/util/live-loan/live-loan-draw.js
```diff
@@ -1,32 +1,45 @@
...
+const classicCardHeight = 510 * classicResizeFactor;
 
 function fontFile(name) {
 	return path.join(__dirname, './fonts', name);
 }
 
 tracer.trace('registerFonts', () => {
 	/* eslint-disable max-len */
-	// registerFont(fontFile('PostGrotesk-Light.ttf'), { family: 'Kiva Post Grot', weight: '300' });
+	registerFont(fontFile('PostGrotesk-Light.ttf'), { family: 'Kiva Post Grot', weight: '300' });
```
emuvente, line 42: This font weight isn't actually being used, but when this line is commented out none of the other font weights load at all. It's also only this line that has that effect for me; no other lines being uncommented or commented made any difference to the loading behavior. It appears to be due to some quirk of the underlying OS that the server is running on, as this issue is happening on my local MacOS but is not happening on our Linux servers.
It doesn't hurt anything to leave this uncommented since all the font files are already on the server, so this doesn't use any additional resources.

dyersituations: When I worked with canvas and custom fonts, it took quite a few iterations to get the fonts loaded async before the rendering started, so not surprising to see the differences. Good to hear it's not an issue on the servers.

PR #4861, opened 2023-06-07
fix: mg page should show the already signed up box for modern subs

eddieferrer:
ACK-492

Now that the subscription service supports non logged in users, we can add this code back to show the already signed up box if a user has a modern sub

### DIFF
--- a/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
+++ b/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
@@ -152,12 +152,12 @@ const pageQuery = gql`
 		contentful {
 			entries(contentType: "page", contentKey: "monthlygood")
 		}
-		# mySubscriptions(includeDisabled: false) {
-		# 	values {
-		# 		id
-		# 		enabled
-		# 	}
-		# }
+		mySubscriptions(includeDisabled: false) {
+			values {
+				id
+				enabled
+			}
+		}
 	}
 `;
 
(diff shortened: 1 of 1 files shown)


src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -266,9 +266,8 @@ export default {
 		preFetch: true,
 		result({ data }) {
 			this.isMonthlyGoodSubscriber = data?.my?.autoDeposit?.isSubscriber ?? false;
-			// TODO! Add this back in when service supports non-logged in users
-			// const modernSubscriptions = data?.mySubscriptions?.values ?? [];
-			// this.hasModernSub = modernSubscriptions.length !== 0;
+			const modernSubscriptions = data?.mySubscriptions?.values ?? [];
+			this.hasModernSub = modernSubscriptions.length !== 0;
```
mcstover, line 270: So no more failures for non-logged-in or non-existent users in the subscriptions service?

mcstover (approved):

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
eddieferrer: Correct this was fixed a while ago, but the service now returns null and behaves the same as `my.account` -- See [PBJ-224](https://kiva.atlassian.net/browse/PBJ-224)

PR #4862, opened 2023-06-07
Core 1364 fix

dyersituations:
https://kiva.atlassian.net/browse/CORE-1367

**Work started by Emily and finished by Casey due to Emily being OOO**

- This experiment essentially changes the default lend CTA dropdown selection option in 2 ways for $5 notes experiment
- The "top up" campaign sets the default to $5
- The "base" campaign sets the default to the $5 increment under the user's current balance but not greater than $25
- The changes affects: BP, LH, category pages, and lend/filter


mcstover (approved):
How do the query params and cookie get set for use in this code?

dyersituations:
> How do the query params and cookie get set for use in this code?

The query params get set in ERL emails, the code is looking for the `utm_campaign` from those emails (a section of the campaign string specifically).

The cookie is set in the `getLendCtaSelectedOption` in the case of a new campaign "starting" for the user, as in identified in the query string.

PR #4864, opened 2023-06-08
fix: disable tooltip for team links on mobile

eddieferrer:
ACK-621

On mobile currently, you cannot click the contributing teams name or image to go to the team page, this is because KvPopper/KvTooltip is using the touchend to display the information pop up. Our priority is to allow users to navigate to the team pages from here, so this fix removes kvTooltip if mobile, allowing the link to work. 

Also fixes this error:
Screen Shot 2023-06-07 at 2 46 41 PM

It turns out that typeof is not the same as instance of


dyersituations (approved):

mcstover (approved):
Seems like we should think about how to show this information on mobile, unless it's clearly visible and discoverable after the click through...

PR #4870, opened 2023-06-13
feat: only check filesystem for static files when the /static route is requested

emuvente:
I noticed a trace for a live loan image that indicated waiting 7 seconds to do the fs.stat() call in express.static. I've seen this call in all our requests for a while now (usually only taking a few milliseconds), but I just realized that we could isolate that filesystem call to only static requests. This should improve server response time, especially in high-traffic situations.

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -52,7 +52,7 @@ function setHeaders(res, path) {
 	res.header('Access-Control-Allow-Headers', 'Origin, X-Requested-With, Content-Type, Accept');
 }
 
-app.use(express.static('dist', {
+app.use('/static', express.static('dist/static', {
 	setHeaders,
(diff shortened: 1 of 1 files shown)


mcstover:
This looks to namespace the files to /dist/static. Are there any other files served using this handler from just /dist?

emuvente:
> Are there any other files served using this handler from just /dist?

No, the only other files in /dist are vue-ssr-client-manifest.json and vue-ssr-server-bundle.json, which we don't serve publicly.

dyersituations (approved):

christian14b (approved):

mcstover (approved):
Nice Find! And thanks for verifying the absense of other files.

PR #4873, opened 2023-06-14
fix: remove duplicate loan query, use async await to control data and…

mcstover:
… loan loading sequence.

### DIFF
--- a/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
+++ b/src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
@@ -352,7 +352,6 @@ export default {
 				return !this.excludedTags.includes(tag.name);
 			});
 		});
-		this.setLoanDisplayMode(this.activeLoanDisplay);
 	},
 	computed: {
 		// Attributes are also known as LoanThemes
--- a/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
+++ b/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
@@ -698,8 +698,12 @@ export default {
 		// const basketItems = this.rawPageData?.shop?.basket?.items?.values ?? [];
 		this.loadingPage = basketItems.some(item => item.__typename === 'LoanReservation'); // eslint-disable-line no-underscore-dangle, max-len
 	},
-	mounted() {
+	async mounted() {
 		this.$root.$on('jumpToLoans', this.jumpToLoans);
+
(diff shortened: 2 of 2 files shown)


src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
```diff
@@ -352,7 +352,6 @@ export default {
 				return !this.excludedTags.includes(tag.name);
 			});
 		});
-		this.setLoanDisplayMode(this.activeLoanDisplay);
```
mcstover, line 355: This was triggering the `showLoans` mechanisms immediately upon page load. It the flows into the loan display and fired a default loan query before we knew anything about the promotion resulting in potentially ineligible loans to show on the page.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -698,8 +698,12 @@ export default {
 		// const basketItems = this.rawPageData?.shop?.basket?.items?.values ?? [];
 		this.loadingPage = basketItems.some(item => item.__typename === 'LoanReservation'); // eslint-disable-line no-underscore-dangle, max-len
 	},
-	mounted() {
+	async mounted() {
 		this.$root.$on('jumpToLoans', this.jumpToLoans);
+
+		// check for applied promo
+		await this.verifyOrApplyPromotion();
```
mcstover, line 705: This method starts our data aggregation for the promo state, we want to control this flow deliberately so we have all the promo information before proceeding with other steps.

src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue
```diff
@@ -985,35 +986,35 @@ export default {
 		},
 	},
 	methods: {
-		verifyOrApplyPromotion() {
+		async verifyOrApplyPromotion() {
```
mcstover, line 989: Using the async/await flow here will ensure we awat the promo application or getting promo info from the basket before any other steps.

src/components/CorporateCampaign/LoanSearch/LoanSearchFilters.vue
mcstover: I've verified that the default `rows` view and the contentful setting that swaps to `grid` both work and load up properly.

dyersituations (approved):

michelleinez (approved):

PR #4874, opened 2023-06-14
feat(DonationNudge): $1 donation upsell before no donation MARS-379

emuvente:
Depends on https://github.com/kiva/kv-ui-elements/pull/259

Screen Shot 2023-06-14 at 12 26 01 PM


mcstover (approved):
Is this going to cause an additional click to decline? If so, can we put some time limited cookie check like 2 weeks or a month to prevent that?

src/assets/icons/inline/heart-icon-circle.svg
dyersituations, line 1: Looks like this new SVG isn't being used?

dyersituations (approved):

christian14b (approved):

src/assets/icons/inline/heart-icon-circle.svg
emuvente: This is actually the old SVG. The viewbox was updated and the file was renamed. The old SVG wasn't being used at all, but I didn't want to make the decision to remove it in this PR.

PR #4877, opened 2023-06-16
chore: update verbiage on lending stats page to lending partner

sagetidwell:
Updates verbiage on lending stats page from field partner to lending partner in accordance with guidance from the Inclusion Committee. 

I wasn't able to validate this locally, so your support in that if you've already got your machine set up to do so would be greatly appreciated!  This updates the bottom section and paragraph of `/portfolio/lending-stats`


mcstover (approved):
Totally safe. Just a note that due to using `chore` instead of `fix` this will not trigger it's own build and deploy upon merge. You'll have to wait for a different commit with `fix` or `feat` to see this show up in dev.

dyersituations (approved):
If you ever want to get the FE apps running locally, just let me know!

PR #4879, opened 2023-06-21
fix: guest checkout controls not visible when logged in

michelleinez:


mcstover (approved):

dyersituations (approved):

mcstover (approved):

src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -96,7 +101,9 @@
 			</kv-grid>
 
 			<kiva-credit-payment
-				v-if="showKivaCreditButton && !promoGuestCheckoutEnabled"
+				v-if="showKivaCreditButton
+					&& !promoGuestCheckoutEnabled
```
mcstover, line 105: @michelleinez Testing this locally and I think this line should be `!continueAsGuest` instead otherwise an existing user on a guest checkout enabled account wouldn't be able to see the button.

mcstover: I realize it's prob fine for now as `promoGuestCheckoutEnabled` is gonna be false most cases, I can just add to my current work.

PR #4881, opened 2023-06-22
feat: improve experience of editing or cancelling your subscription

christian14b:


src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -255,10 +267,37 @@ export default {
...
 				this.subscriptionId = data?.my?.autoDeposit?.id;
+				this.firstName = data?.my?.userAccount?.firstName ?? '';
+				this.subscriptionsLoans = data?.mySubscriptions?.values?.length ?? 0;
+				const mySubscriptions = data?.mySubscriptions?.values?.history?.values ?? [];
+				this.subStartTimestamp = mySubscriptions?.[mySubscriptions.length - 1]?.timeStamp ?? null;
 			}
 		},
 	},
 	computed: {
+		subStartDate() {
```
eddieferrer, line 278: We use parseISO everywhere else in the ui app to convert UTC dates to formatted date strings. I'm not sure if there is a difference here, just wanted to point that out

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -10,54 +10,74 @@
 				</router-link>
 
 				<div v-if="isMonthlyGoodSubscriber">
-					<p class="tw-mb-2">
-						On the <button
-							class="tw-text-link tw-font-medium"
-							@click="showEditLightbox = true;"
+					<p>
+						<span class="tw-text-action">{{ firstName }},</span>
```
dyersituations, line 14: This element needs `data-hj-suppress`

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -189,6 +179,10 @@ import SubscriptionsMonthlyGoodCancellationFlow from
 import KvButton from '~/@kiva/kv-components/vue/KvButton';
 import KvLightbox from '~/@kiva/kv-components/vue/KvLightbox';
 
+const CHANGE_SUBSCRIPTION = 'change_subscription';
```
dyersituations, line 182: These constants could also be used in the template.

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -255,10 +267,37 @@ export default {
...
+				this.subscriptionsLoans = data?.mySubscriptions?.values?.length ?? 0;
+				const mySubscriptions = data?.mySubscriptions?.values?.history?.values ?? [];
+				this.subStartTimestamp = mySubscriptions?.[mySubscriptions.length - 1]?.timeStamp ?? null;
 			}
 		},
 	},
 	computed: {
+		subStartDate() {
+			if (!this.subStartTimestamp) return '';
+			const timestamp = new Date(this.subStartTimestamp); // Replace with your own timestamp
```
dyersituations, line 280: I'm guessing the trailing comment isn't needed.

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -370,7 +428,29 @@ export default {
...
+		onModifyCancel() {
+			this.modalStep = CHANGE_SUBSCRIPTION;
+			this.showEditLightbox = true;
+			this.showCancelLightbox = false;
+		},
+		onAbortCancel() {
+			this.showCancelLightbox = false;
+		},
+		setStep(step) {
+			console.log(step === CANCEL_SUBSCRIPTION);
```
eddieferrer, line 445: Remove log

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -255,10 +267,37 @@ export default {
...
 			}
 		},
 	},
 	computed: {
+		subStartDate() {
+			if (!this.subStartTimestamp) return '';
+			const timestamp = new Date(this.subStartTimestamp); // Replace with your own timestamp
+			return `since ${timestamp.toLocaleString('en-US', { month: 'long' })} ${timestamp.getFullYear()}!`;
+		},
+		subMonthCount() {
```
dyersituations, line 283: I recommend moving these new utility methods out of the component file so that they can unit tested. Either now or in a fast follow ticket.

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
dyersituations, line 445: Remove

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -10,54 +10,74 @@
...
+						<div>
+							<p class="tw-text-subhead tw-text-primary">
+								Payment Method
+							</p>
+							<p class="tw-text-secondary">
+								{{ cardDescription }}
+							</p>
+						</div>
+						<kv-button
+							variant="secondary"
```
eddieferrer, line 45: I think we would want analytics on the Edit/Update/Change buttons, no?

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -255,10 +267,37 @@ export default {
 				this.mgAmount = autoDepositAmount - this.donation;
 				this.paymentMethod = data?.my?.autoDeposit?.paymentMethod ?? {};
 				this.subscriptionId = data?.my?.autoDeposit?.id;
+				this.firstName = data?.my?.userAccount?.firstName ?? '';
+				this.subscriptionsLoans = data?.mySubscriptions?.values?.length ?? 0;
+				const mySubscriptions = data?.mySubscriptions?.values?.history?.values ?? [];
+				this.subStartTimestamp = mySubscriptions?.[mySubscriptions.length - 1]?.timeStamp ?? null;
```
eddieferrer, line 273: This timestamp gets updated every time the subscription gets updated, so if you are subscriber for 1 year, then update the payment method this will come back inaccurately

eddieferrer, line 273: subStartTimestamp needs to be added to data on the component

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -255,10 +267,37 @@ export default {
 				this.mgAmount = autoDepositAmount - this.donation;
 				this.paymentMethod = data?.my?.autoDeposit?.paymentMethod ?? {};
 				this.subscriptionId = data?.my?.autoDeposit?.id;
+				this.firstName = data?.my?.userAccount?.firstName ?? '';
+				this.subscriptionsLoans = data?.mySubscriptions?.values?.length ?? 0;
+				const mySubscriptions = data?.mySubscriptions?.values?.history?.values ?? [];
```
eddieferrer, line 272: This doesn't work at all. It should be: `data?.mySubscriptions?.values[0]?.history?.values` because data?.mySubscriptions?.values is an array

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
emuvente: The `LenderSubscription.timestamp` field is a `Long` integer timestamp. It probably should have been setup initially as a `Date` UTC string, but that's not what we're working with, unfortunately.

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -248,17 +269,39 @@ export default {
 		result({ data }) {
 			this.isMonthlyGoodSubscriber = data?.my?.autoDeposit?.isSubscriber ?? false;
 			if (this.isMonthlyGoodSubscriber) {
-				const autoDepositAmount = parseFloat(data?.my?.autoDeposit?.amount ?? 0);
-				this.donation = parseFloat(data?.my?.autoDeposit?.donateAmount ?? 0);
+				const autoDepositAmount = parseFloat(data?.mySubscriptions?.values?.[0]?.amount ?? 0);
```
eddieferrer, line 272: I don't think these changes are correct. 

The legacy service used to have the total defined by autoDepositAmount and a donation. This is the reason, why we had to do this: `this.mgAmount = autoDepositAmount - this.donation;`

With the new subscription service, this is not necessary. 

This line can be removed: `const autoDepositAmount = parseFloat(data?.mySubscriptions?.values?.[0]?.amount ?? 0);`

and `this.mgAmount = autoDepositAmount - this.donation;` can now become: 

`parseFloat(data?.mySubscriptions?.values?.[0]?.amount ?? 0);`

eddieferrer: Also, before setting the component variables in this result block, you should check to see that this subscription is a monthly good subscription and not just use the first item: data?.mySubscriptions?.values?.[0]

I think this should be refactored to be: 
`const activeMonthlyGoodSubscription = data?.mySubscriptions?.values.find(sub => sub.subscriptionType === "MG" === "MG")` - similar to what you're doing to get the MG history values. 

Then set all your other variables based on that subscription

eddieferrer:
> 



> > I pulled this locally and it is working. I think we should refactor some of this to better support the new subscriptions service.
> > For example, the amount and donation amount can now be pulled from here`"mySubscriptions": { "values": [ { "amount": 25, "donation": 3.75,`
> > instead of the `autoDeposit` response.
> > I also think we should not assume that the first subscription value is a monthly good subscription. We should be checking the values array for a subscription that is subscriptionType === "MG" and then target that subscription.
> 
> @eddieferrer thanks for the feedback regarding the new subscription service. This pr was updated using the `amount` and `donation` values from `mySubscriptions`. I wasn't completely sure if the first values of this array are the most recent subscriptions

See my other comment, you should filter on the subscription type for this array of values, like you are doing for the history values. A user should only have 1 subscription of type MG

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -370,7 +439,44 @@ export default {
...
+			this.showEditLightbox = true;
+			this.showCancelLightbox = false;
+		},
+		onAbortCancel() {
+			this.showCancelLightbox = false;
+		},
+		trackStep(step) {
+			switch (step) {
+				case CHANGE_SUBSCRIPTION:
+					this.$kvTrackEvent('monthlyGood', 'click-edit-monthly-good', 'Edit');
```
eddieferrer, line 458: These 3 events do not conform to our new web events framework. 

The category should be "subscription", the action "click", and the label something descriptive

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
eddieferrer: Yeah that would work 👍

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -248,17 +274,40 @@ export default {
...
+					.find(sub => sub?.category?.subscriptionType === 'MG');
+				this.donation = parseFloat(activeMonthlyGoodSubscription?.donation ?? 0);
 				this.dayOfMonth = data?.my?.autoDeposit?.dayOfMonth;
 				this.category = data?.my?.monthlyGoodCategory ?? '';
-				this.mgAmount = autoDepositAmount - this.donation;
+				this.mgAmount = parseFloat(activeMonthlyGoodSubscription?.amount ?? 0);
 				this.paymentMethod = data?.my?.autoDeposit?.paymentMethod ?? {};
 				this.subscriptionId = data?.my?.autoDeposit?.id;
+				this.firstName = data?.my?.userAccount?.firstName ?? '';
+				this.subscriptionsLoans = data?.mySubscriptions?.values?.length ?? 0;
```
eddieferrer, line 286: This is not correct. There will only be 1 value in `data?.mySubscriptions?.values` of type monthly good. I am not sure that we currently have a way to determine number of loans made. 

I think you'll have to take the subscription start date, the subscription amount and todays date, and manually count how many times it would have triggered for each $25 loan purchase to find out the number of loans being lent. This assumes that MG purchases only $25 loan share on different loans....

src/components/Subscriptions/SubscriptionsMonthlyGood.vue
```diff
@@ -248,17 +274,40 @@ export default {
...
+				// eslint-disable-next-line max-len
+				const mgSubs = activeMonthlyGoodSubscription?.history?.values ?? [];
+				this.subStartTimestamp = mgSubs?.[mgSubs.length - 1]?.timestamp ?? null;
```
eddieferrer, line 289: This is not working for me. I have a new subscription and it displays "12 months". Your conversion is working, but the timestamp on my last entry in `history?.values` is in fact 12 months ago. I would double check what events trigger these timestamps, and if you can rely on the last timestamp to be the subscription creation date. 

I am not confident on what events trigger timestamps. Maybe you want the first timestamp item?

eddieferrer (approved):
👍

PR #4884, opened 2023-06-23
feat: add team name to comments on BP

eddieferrer:
ACK-625
Screen Shot 2023-06-23 at 1 19 00 PM

### DIFF
--- a/src/components/BorrowerProfile/CommentsAndWhySpecial.vue
+++ b/src/components/BorrowerProfile/CommentsAndWhySpecial.vue
@@ -98,9 +98,14 @@
 								<h3>
 									{{ comment.authorName }}
 								</h3>
-								<!-- TODO add team name <h4>
-									Team Name
-								</h4> -->
+								<h4 v-if="comment.lenderTeam && comment.lenderTeamPublicId">
+									<router-link
+										:to="`/team/${comment.lenderTeamPublicId}`"
+										class="tw-text-primary"
+									>
+										{{ comment.lenderTeam }}
+									</router-link>
+								</h4>
 							</div>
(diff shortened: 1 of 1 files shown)


src/components/BorrowerProfile/CommentsAndWhySpecial.vue
```diff
@@ -98,9 +98,14 @@
 								<h3>
 									{{ comment.authorName }}
 								</h3>
-								<!-- TODO add team name <h4>
-									Team Name
-								</h4> -->
+								<h4 v-if="comment.lenderTeam">
+									<router-link
+										:to="`/team/${comment.lenderTeamPublicId}`"
```
dyersituations, line 103: Since `lenderTeamPublicId` can be `null`, should the `v-if` also check that prop?

src/components/BorrowerProfile/CommentsAndWhySpecial.vue
```diff
@@ -287,6 +297,17 @@ export default {
 									id
 									authorName
 									authorImageUrl
+									authorLendingAction {
```
dyersituations, line 300: When will this be available on prod?

eddieferrer: I can coordinate with merging this in. It is on dev now and should be in prod next release

mcstover: We'll just need to make sure we've deployed that to prod before we send out the UI release.

mcstover (approved):

dyersituations (approved):

PR #4886, opened 2023-06-26
Fix tablet design mars 410

christian14b:
Same error as found on cps: 
When items on the basket as a visitor, search should be a button and small fixes (Log in is always link, show dropdown on tablet, support kiva order in about dropdown)

image


mcstover (approved):

dyersituations:
Looks like some more adjustment might be needed. Here is a logged in user w/ basket. The searchbar is tiny.
image

dyersituations:
Last issue I'm seeing is really minor, but there's a slight spacing difference at a medium sized browser, for example 1102. It should be a quick fix, otherwise I wouldn't have mentioned it:
image

dyersituations (approved):

PR #4888, opened 2023-06-27
fix(KvDropdown): skip dropdown init and log error when controller missing

emuvente:
Fixes recent errors like `Cannot read properties of null (reading 'addEventListener')`

### DIFF
--- a/src/components/Kv/KvDropdown.vue
+++ b/src/components/Kv/KvDropdown.vue
@@ -59,7 +59,11 @@ export default {
 		},
 	},
 	mounted() {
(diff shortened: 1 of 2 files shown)


dyersituations (approved):

christian14b (approved):

mcstover (approved):
ty!

mcstover:
An alternate approach I just found to resolve the issue locally was to remove the `v-show` on line 416 and use `:class="{ 'tw-hidden': !isVisitor }"` instead.

PR #4890, opened 2023-06-27
feat: kv components version updated

roger-in-kiva:
- kv-components version updated

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "kvue",
-	"version": "2.612.1",
+	"version": "2.619.0",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "kvue",
-			"version": "2.612.1",
+			"version": "2.619.0",
 			"license": "UNLICENSED",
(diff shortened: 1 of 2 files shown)


dyersituations:
Is there a particular feature this is needed for?

roger-in-kiva:
> 



> Is there a particular feature this is needed for?

Are we using shared loan cards here? We have the same issue for clickable loan cards when loading... @dyersituations

dyersituations:
> > 
> 
> > Is there a particular feature this is needed for?
> 
> Are we using shared loan cards here? We have the same issue for clickable loan cards when loading... @dyersituations

We aren't using the new loan card yet in the `ui` repo. Maybe we should close this for now and wait for a feature that needs the updates?

roger-in-kiva:
> > > 
> > 
> > 
> > > Is there a particular feature this is needed for?
> > 
> > 
> > Are we using shared loan cards here? We have the same issue for clickable loan cards when loading... @dyersituations
> 
> We aren't using the new loan card yet in the `ui` repo. Maybe we should close this for now and wait for a feature that needs the updates?

Makes sense, thanks for checking!

PR #4892, opened 2023-06-28
fix: set donation tip to 0 when just 1 dollar donation modal is closed

christian14b:
Closing out donation modal wasn't updating the tip to $0, instead the value remained the same. This pr updates the tip to $0 when this modal is closed.

image

### DIFF
--- a/src/components/Checkout/DonationNudge/DonationNudgeLightbox.vue
+++ b/src/components/Checkout/DonationNudge/DonationNudgeLightbox.vue
@@ -4,7 +4,7 @@
 		:no-padding-sides="true"
 		:no-padding-bottom="true"
 		:no-padding-top="true"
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

src/components/Checkout/DonationNudge/DonationNudgeLightbox.vue
```diff
@@ -4,7 +4,7 @@
 		:no-padding-sides="true"
 		:no-padding-bottom="true"
 		:no-padding-top="true"
-		@lightbox-closed="closeNudgeLightbox"
+		@lightbox-closed="closeZeroUpsell(0)"
```
emuvente, line 7: We only want to do this when the zero donation upsell is open (i.e. the user has already selected $0), not all the time.

emuvente (approved):

PR #4896, opened 2023-07-06
feat: needed changes to replace lbc by lending home page

roger-in-kiva:
Required changes to replace lend-by-category page by lending-home page
- removing LendByCategoryPage.vue file and related unused files
- Change /lend-by-category route to use LoanFinding.vue
- Redirect /lending-home to /lend-by-category
- Remove canonical link from LoanFinding.vue header/metaInfo


src/components/LoansByCategory/CategoryRowHover.vue
```diff
@@ -1,651 +0,0 @@
...
-</template>
-
-<script>
-import _filter from 'lodash/filter';
-import _get from 'lodash/get';
-import _throttle from 'lodash/throttle';
-import LoanCardController from '@/components/LoanCards/LoanCardController';
-import categoryRowArrowsVisibleMixin from '@/plugins/category-row-arrows-visible-mixin';
-import KvExpandable from '@/components/Kv/KvExpandable';
-import detailedLoanCardFragment from '@/graphql/fragments/detailedLoanCard.graphql';
```
dyersituations, line 148: Can this file be removed?

src/components/LoansByCategory/FeaturedHeroLoanWrapper.vue
```diff
@@ -1,302 +0,0 @@
...
-			/>
-			<kv-loading-overlay v-if="loading" id="updating-overlay" />
-		</div>
-	</div>
-</template>
-
-<script>
-import _filter from 'lodash/filter';
-import _get from 'lodash/get';
-import featuredLoansQuery from '@/graphql/query/featuredLoansData.graphql';
```
dyersituations, line 38: Can this file be removed?

src/components/LoansByCategory/MGLightbox.vue
```diff
@@ -1,144 +0,0 @@
...
-	inject: ['apollo', 'cookieStore'],
-	components: {
-		KvButton,
-		KvLightbox,
-		KvMaterialIcon,
-	},
-	data() {
-		return {
-			imageRequire,
-			imageSource: imageRequire('./MG-Lightbox-11894_Paraguay_Pablina_45.jpg'),
```
dyersituations, line 57: Can we remove this image?

src/components/LoansByCategory/RecentlyViewedLoans.vue
```diff
@@ -1,154 +0,0 @@
...
-			</div>
-		</div>
-		<kv-loading-overlay v-if="!isLoaded && !zeroRecentLoans" id="updating-overlay" />
-	</div>
-</template>
-
-<script>
-import _get from 'lodash/get';
-import WebStorage from 'store2';
-import loansByIDQuery from '@/graphql/query/loansById.graphql';
```
dyersituations, line 24: Can we remove this file?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -1,901 +0,0 @@
...
-import { isLoanFundraising } from '@/util/loanUtils';
-import {
-	getExperimentSettingCached,
-	trackExperimentVersion
-} from '@/util/experiment/experimentUtils';
-import experimentQuery from '@/graphql/query/experimentAssignment.graphql';
-import basketItems from '@/graphql/query/basketItems.graphql';
-import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
-import lendByCategoryQuery from '@/graphql/query/lendByCategory/lendByCategory.graphql';
-import loanChannelQuery from '@/graphql/query/loanChannelData.graphql';
```
dyersituations, line 87: Can this file be removed?

src/pages/Lend/LendByCategoryPage.vue
```diff
@@ -1,901 +0,0 @@
...
-import _filter from 'lodash/filter';
-import _get from 'lodash/get';
-import _map from 'lodash/map';
-import _uniqBy from 'lodash/uniqBy';
-import _without from 'lodash/without';
-import { addCustomChannelInfo } from '@/util/categoryUtils';
-import logReadQueryError from '@/util/logReadQueryError';
-import { readJSONSetting } from '@/util/settingsUtils';
-import { indexIn } from '@/util/comparators';
-import { FLSS_ORIGIN_LEND_BY_CATEGORY } from '@/util/flssUtils';
```
dyersituations, line 77: Please ask Rob if we should be using this `FLSS_ORIGIN_LEND_BY_CATEGORY` on the new lending home page now that it lives at `/lend-by-category`, or if we should continue to use `FLSS_ORIGIN_LENDING_HOME`. We want to continue using `FLSS_ORIGIN_LENDING_HOME`, then `FLSS_ORIGIN_LEND_BY_CATEGORY` can be removed from `flssUtills`. My guess is that we want to use `FLSS_ORIGIN_LEND_BY_CATEGORY` in the new page, but we'll want to verify first.

src/pages/Lend/LendByCategoryPage.vue
roger-in-kiva: Removed in this commit

src/pages/Lend/LendByCategoryPage.vue
dyersituations: I saw the commit to change the origin. Is that what the data team prefers?

roger-in-kiva: Yeah, confirmation is in the ticket.

src/router/routes.js
```diff
@@ -353,10 +353,7 @@ module.exports = [
 	},
 	{
 		path: '/lending-home',
-		component: () => import('@/pages/LoanFinding/LoanFinding'),
-		meta: {
-			excludeFromStaticSitemap: true,
```
dyersituations, line 358: Do we still need this entry for the old URL?

roger-in-kiva: Good question. I'm not sure b/c `/lend-by-category` route was not including that, if we want to preserve the same is not required I think.

```
meta: {
      unbounceEmailCapture: true,
},
```

But other cases like category page, /lend-by-category/*, includes both

```
unbounceEmailCapture: true,
unbouncePopUp: true,
```

What do you think about this?

dyersituations: Right, I just think we might want to exclude the old URL in the static sitemap. Please take a look at where `excludeFromStaticSitemap` is used, since that might help determining the correct entry here. My assumption without testing is that `excludeFromStaticSitemap` will exclude `/lending-home` from the static site map. Or maybe the static sitemap excludes all redirects? That's generally what needs to be confirmed.

dyersituations: We want `/lend-by-category` in the static sitemap but not `/lending-home`.

mcstover: If `/lending-home` is now a redirect you could leave in the excludeFromStaticSitemap setting but leave off the others as they signify something loaded in the client and that page would never be allowed to load due to the redirect.

dyersituations: It looks like the other redirects don't have `excludeFromStaticSitemap`, but please confirm that the redirects are indeed excluded.

dyersituations: Thanks @mcstover! Sounds like this redirect should still have `excludeFromStaticSitemap`.

dyersituations: @mcstover Is there a reason why the other redirects don't have `excludeFromStaticSitemap`?

mcstover: I'm not sure. I think your point about running the script and verifying (or closely inspecting the sitemap generator code , NOT the UISiteMap.vue file) is the best way to know if it's required or not.

dyersituations: I just took a look, and the sitemap lives locally at `http://localhost:8888/sitemaps/ui.xml`, and I verified in the that we do indeed ignore `redirect` entries in the generator that pulls from the Vue Router:

```
// Remove excluded routes
.filter(route => !route.meta?.excludeFromStaticSitemap && !route.redirect)
```

dyersituations: @roger-in-kiva `excludeFromStaticSitemap` can be removed from this entry.

dyersituations (approved):

PR #4897, opened 2023-07-07
fix: minor issues with comments on bp

eddieferrer:
There was a bug with displaying the team name if a user is in many teams and the default limit of 20 is not enough. Added artificial limit of 100 

Also added word wrapping as some users are putting in very long urls in the comment

### DIFF
--- a/src/components/BorrowerProfile/CommentsAndWhySpecial.vue
+++ b/src/components/BorrowerProfile/CommentsAndWhySpecial.vue
@@ -45,7 +45,7 @@
 						</div>
 						<!-- comment -->
 						<h2>
-							<em>"{{ comment.body }}"</em>
+							<em class="tw-break-words">"{{ comment.body }}"</em>
 						</h2>
 						<!-- author -->
 						<div class="tw-float-right tw-flex tw-align-center tw-mt-1.5">
@@ -301,7 +301,7 @@ export default {
 										teams
 										lender {
 											id
-											teams {
+											teams(limit: 100) { #arbitrary limit for lenders that have a lot of teams
 												values {
 													id
 													name


src/components/BorrowerProfile/CommentsAndWhySpecial.vue
```diff
@@ -301,7 +301,7 @@ export default {
 										teams
 										lender {
 											id
-											teams {
+											teams(limit: 100) { #arbitrary limit for lenders that have a lot of teams
```
mcstover, line 304: Should we reduce this further? Also, can you please check to see if this comment is getting transported in the query POST? If so, please move the comment out of and place above the query.

eddieferrer: The comment is not transported in the request. 

I'm not sure. Is there any performance implications? We're finding that a small % of lenders do have more than 20 teams. I think 40 teams should cover most users, but we wanted to be sure to cover all users. If there are performance implications I could reduce this

mcstover: If this loads up team data for every visitor we should trend toward a lower threshold. If it's only loading this data for the single logged in user we can keep it higher.

Are we showing more than 1 team name per lender? Are we showing this information at all? Could we just use a totalCount instead of fetching data for each team? What is this being used for?

eddieferrer: So the way the API contract is currently, for a single comment on a loan, `authorLendingAction.teams` returns the team name that comment was attributed to. For example "nerdfighters". But in order to get the team url so the team name can link to the team, we have to cross reference this name with the complete team info from all the lenders teams, which is given by `authorLendingAction.lender.teams`

So yeah this is just used to get the team url. 

It would load for every visitor when they scroll down to the comments section.

mcstover: Hmmm. I can see now "why" you have to do this but I don't think we should be doing this long term. It's going to create a lot of unnessary load on requesting team data so should be considered temporary in this state.

Seems like lendingAction should be returning a team object instead of just a string with the name... Alternatively, if that had the team id along with the name you could then query for just the team you need data for.

Could you please plan some work to optimize this.

mcstover (approved):

PR #4902, opened 2023-07-12
feat: requests comments from guests on thank you page

eddieferrer:
Ask for comments from guests as well, since they are a large portion of Kiva US Checkout.

The flow for guests is: 
1. On thank you page ask for comment
2. Save comment to cookie, inform user that they have to finish creating an account to post their comment. 
3. Display account claim form (first Name, last name)
4. Follow account claim steps: auth0 password creation, redirect to `/register/guest-redirect`
5. If comment in progress cookies are found by `/register/guest-redirect` redirect to login with destination URL borrower profile with query params.
6. User lands on borrower profile. Mix-in checks for comment cookies, and submits comment and deletes cookies if successful.   
 
Refactors: 

- Refactored guest account creation form so we can use it in multiple places (normal upsell an this new comments upsell)
- Refactored some events to match new event analytics (post-checkout category, etc)

### DIFF
--- a/src/components/Thanks/CommentAsk.vue
+++ b/src/components/Thanks/CommentAsk.vue
@@ -3,9 +3,13 @@
 		<div class="tw-bg-secondary" v-if="showComments">
 			<kv-page-container>
 				<kv-grid class="tw-grid-cols-12">
-					<div class="tw-col-span-12 lg:tw-col-span-8 lg:tw-col-start-3 tw-pt-2 tw-mb-4">
+					<!-- comment form -->
+					<div
+						class="tw-col-span-12 lg:tw-col-span-8 lg:tw-col-start-3 tw-pt-2 tw-mb-4"
+						v-if="!showAccountCreation"
+					>
 						<h1	class="tw-mt-1 tw-mb-3 tw-text-left">
-							Tell others why you love this loan to {{ loanName }}.
+							Tell others why you love this loan to {{ loanName }}
 						</h1>
 						<p class="tw-m-0 tw-text-subhead">
 							Your comments can really help {{ loanName }} fully fund their loan.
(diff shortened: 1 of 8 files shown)


mcstover (approved):
This really neat. The abstraction of the guest account claim component may help us out in the campaign context as well!

src/components/Thanks/CommentAsk.vue
```diff
@@ -151,6 +179,22 @@ export default {
...
 		submitComment() {
+			if (!this.isGuest) {
+				this.submitCommentAsUser();
+			} else {
+				this.submitCommentAsGuest();
+			}
+		},
+		submitCommentAsGuest() {
+			// save comment to cookie
+			this.cookieStore.set('guestCommentComment', this.userComment, { path: '/' });
```
dyersituations, line 190: Consider exporting these magic strings from the mixin file.

src/plugins/guest-comment-mixin.js
```diff
@@ -0,0 +1,55 @@
...
+import logFormatter from '@/util/logFormatter';
+
+export default {
+	mounted() {
+		const guestCommentComment = this.cookieStore.get('guestCommentComment');
+		const guestCommentInProgress = this.cookieStore.get('guestCommentInProgress');
+		const guestCommentLoanId = this.cookieStore.get('guestCommentLoanId');
+		const currentLoanId = this.$route?.params?.id;
+		const isTargetBorrowerProfile = currentLoanId.toString() === guestCommentLoanId.toString();
+		if (this.$route.query.guest_comment
```
dyersituations, line 18: Could we instead just check the existence of the cookies and not worry about the query string?

src/plugins/guest-comment-mixin.js
```diff
@@ -0,0 +1,55 @@
...
+			}).then(({ data }) => {
+				// comment was added successfully
+				if (data.loan.addComment) {
+					this.$showTipMsg('Thank you for leaving a comment!');
+					// remove cookie values so we don't post the comment again
+					this.cookieStore.remove('guestCommentComment');
+					this.cookieStore.remove('guestCommentInProgress');
+					this.cookieStore.remove('guestCommentLoanId');
+				} else {
+					throw new Error('Comment not added');
```
dyersituations, line 46: Should we remove the cookie also if it was unsuccessful?

eddieferrer: Maybe? My thinking was that if there was some intermittent server error or something the comment would still be added if the user reloaded the page or visited the loan next time.

dyersituations (approved):

PR #4903, opened 2023-07-13
feat: ab setup added to lending home page for recommended row replace…

roger-in-kiva:
- ab setup added to lending home page for recommended row replacement test


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -427,6 +430,16 @@ export default {
 			FLSS_ONGOING_EXP_KEY,
 			'EXP-VUE-FLSS-Ongoing-Sitewide'
 		);
+
+		// Recommended row replacement test
+		const { version } = trackExperimentVersion(
```
dyersituations, line 435: The ticket mentions that this experiment should only run for new users, but this runs for all users. Are we indeed only targeting new users?

roger-in-kiva: I think this experiment for logged in and logged out users.
Here the case for logged out users, https://kiva.atlassian.net/browse/CORE-1343
And here the case for logged in users, with personalized copy: https://kiva.atlassian.net/browse/CORE-1342

dyersituations (approved):

PR #4905, opened 2023-07-14
feat: show captcha on /register/social page when required MARS-446

emuvente:
Add support for requiring a captcha at login/registration. Because captcha values can be large, I changed the form to submit a POST request with form encoded data instead of redirecting with GET query parameters. Support for the data coming in a POST request was added in kiva/auth0#338.

### DIFF
--- a/src/pages/LoginAndRegister/RegisterSocial.vue
+++ b/src/pages/LoginAndRegister/RegisterSocial.vue
@@ -10,8 +10,8 @@
 			<form
 				id="registerSocialTermsForm"
 				class="promptForm tw-text-left"
-				action="."
-				@submit.prevent.stop="postRegisterSocialForm"
+				:action="`https://${$appConfig.auth0.domain}/continue?state=${$route.query.state}`"
+				@submit="postRegisterSocialForm"
 			>
 				<kv-base-input
 					name="firstName"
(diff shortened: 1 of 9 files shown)


src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -10,8 +10,8 @@
 			<form
 				id="registerSocialTermsForm"
 				class="promptForm tw-text-left"
-				action="."
-				@submit.prevent.stop="postRegisterSocialForm"
+				:action="`https://${$appConfig.auth0.domain}/continue?state=${$route.query.state}`"
```
mcstover, line 13: Does the query on this url still need the `newsConsent` query param used elsewhere or is that just a fallback now?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -64,8 +64,19 @@
 				>
 					I want to receive updates about my loans, Kiva news, and promotions in my inbox
 				</kv-base-input>
+				<re-captcha-enterprise
+					:requried="needsCaptcha"
+					@update="captcha = $event"
```
mcstover, line 69: Wondering if we can tap into this update event to extract errors or rate limit failures for logging?

mcstover (approved):

src/pages/LoginAndRegister/RegisterSocial.vue
emuvente: Yeah, we might be able to. I'll look at that after verifying

PR #4912, opened 2023-07-18
feat: resolve issues with register social e2e

dyersituations:
- Two changes were needed to get the e2e tests passing in Chrome
- Looks like test setup + config issues rather than app issues

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -19,4 +19,6 @@ module.exports = defineConfig({
 		specPattern: 'test/e2e/specs/**/*.cy.{js,jsx,ts,tsx}',
 		supportFile: 'test/e2e/support/index.js',
 	},
+	// Required to enable cross-domain iframe access
+	chromeWebSecurity: false
 });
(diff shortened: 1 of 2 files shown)


mcstover (approved):
Is there a way to conditionally set the chrome exclusion based on the domain we're running tests from?

dyersituations:
> Is there a way to conditionally set the chrome exclusion based on the domain we're running tests from?

I guess the `chromeWebSecurity` requires a browser reload, so it cannot be toggled for a single test. We could split out the tests the need iframes to their own files and run those separately, according to the Cypress dev below.

Thoughts? Keep settings as is in this PR or investigate splitting out a separate test file and a new test run in the pipeline? While I'm not a huge fan of lowering the security, I don't expect there to be an issue, since we are explicitly choosing what the tests interact with.

https://github.com/cypress-io/cypress/issues/8160

mcstover:
@dyersituations I think we can just proceed for now. thank you!

PR #4914, opened 2023-07-18
feat: displayed loans event implemented for new featured loan test

roger-in-kiva:
- displayed loans event implemented for new featured loan test

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -366,7 +366,7 @@ export default {
 			}
 		},
 		trackFirstRowDisplayedLoans() {
-			this.trackDisplayedLoans(this.enableRelendingExp ? 'relending' : 'recommended', 1, this.firstRowLoans);
+			this.trackDisplayedLoans(this.enableRelendingExp && !this.enableRecommendedReplacementExp ? 'relending' : 'recommended', 1, this.enableRecommendedReplacementExp ? [this.featuredLoan] : this.firstRowLoans); // eslint-disable-line max-len
 		},
 		trackQuickFiltersDisplayedLoans({ data, pageOffset }) {
 			this.trackDisplayedLoans('quick-filters', this.enableRelendingExp ? 3 : 2, data, pageOffset);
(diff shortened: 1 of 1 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -366,7 +366,7 @@ export default {
 			}
 		},
 		trackFirstRowDisplayedLoans() {
-			this.trackDisplayedLoans(this.enableRelendingExp ? 'relending' : 'recommended', 1, this.firstRowLoans);
+			this.trackDisplayedLoans(this.enableRelendingExp && !this.enableRecommendedReplacementExp ? 'relending' : 'recommended', 1, this.enableRecommendedReplacementExp ? [this.featuredLoan] : this.firstRowLoans); // eslint-disable-line max-len
```
dyersituations, line 369: We will be removing the relending experiment, so you could simplify this line now.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -366,7 +366,7 @@ export default {
 			}
 		},
 		trackFirstRowDisplayedLoans() {
-			this.trackDisplayedLoans(this.enableRelendingExp ? 'relending' : 'recommended', 1, this.firstRowLoans);
+			this.trackDisplayedLoans('recommended', 1, this.enableRecommendedReplacementExp ? [this.featuredLoan] : this.firstRowLoans); // eslint-disable-line max-len
```
dyersituations, line 369: It actually looks like we also need a fixed fallback here: https://github.com/kiva/ui/blob/048d035a642cc44bb482cdc050104b52390c05c5/src/pages/LoanFinding/LoanFinding.vue#L449

`})?.fundraisingLoans?.values[0] ?? [{ id: 0 }];` -> notice the array

roger-in-kiva: Well, with `fundraisingLoans?.values[0]` we are just returning the loan object and assigning to this `featuredLoan: {},`.
Said that, we don't need the array.
If we  want to preserve the array in both, we need to update the assignment to the `featureLoan` var.

dyersituations: Oh yup, you're right.

PR #4915, opened 2023-07-19
feat: set up tip rate optimization experiment

christian14b:
Setup tip rate optimization experiment. Should be merged along with https://github.com/kiva/kiva/pull/11778

### DIFF
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -509,7 +509,7 @@ export default {
 		}
 
 		// MARS-452 tip rate optimization experiment
-		if (!this.checkingOutAsGuest) {
+		if (this.myId !== null && this.myId !== undefined) {
 			const tipRateOptimizationExperiment = this.apollo.readFragment({
 				id: `Experiment:${TIP_RATE_OPTIMIZATION_EXP}`,
 				fragment: experimentVersionFragment,


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -502,6 +504,21 @@ export default {
...
 		}
 
+		// MARS-452 tip rate optimization experiment
+		const tipRateOptimizationExperiment = this.apollo.readFragment({
+			id: `Experiment:${TIP_RATE_OPTIMIZATION_EXP}`,
+			fragment: experimentVersionFragment,
+		}) || {};
+
+		this.tipRateOptimizationExperimentVersion = tipRateOptimizationExperiment.version;
+		if (this.tipRateOptimizationExperimentVersion && !this.checkingOutAsGuest) {
```
mcstover, line 514: If guest users are to be excluded, please wrap all of this in `!this.checkingOutAsGuest` instead of just the tracking.

dyersituations (approved):

mcstover (approved):

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -502,6 +508,23 @@ export default {
 			);
 		}
 
+		// MARS-452 tip rate optimization experiment
+		if (!this.checkingOutAsGuest) {
```
emuvente, line 512: Because this is in the `created` hook, `this.checkingOutAsGuest` will always be `false`. It doesn't get set to `true` until the user clicks "Continue as guest". Maybe this should check for `this.myId` instead? I'm not sure if/how we can set this up before the user has logged in.

emuvente (approved):

PR #4916, opened 2023-07-19
feat: adding v-else for matchingCampaign type [CIT-160]

tiarascoleman:
For Matching Type Accounts there is no promo credit therefore the UI should not render promo credit meter. Curious if the button length should be reduced?

Screen Shot 2023-07-18 at 3 27 51 PM
(How it looks without a filled basket) 

Screen Shot 2023-07-18 at 3 49 10 PM
(With loan in basket)

Screen Shot 2023-07-18 at 3 50 04 PM
(Mobile screen)

### DIFF
--- a/src/components/CorporateCampaign/CampaignProgressBar.vue
+++ b/src/components/CorporateCampaign/CampaignProgressBar.vue
@@ -1,6 +1,6 @@
 <template>
 	<div
-		v-if="promoAmount && promoAmount > 0"
+		v-if="(promoAmount && promoAmount > 0) && isMatchingCampaign"
 		class="
 			tw-rounded
 			tw-bg-brand-50
@@ -42,6 +42,32 @@
 			<kv-ui-button
 				v-show="basketLoans.length > 0"
 				@click="$emit('show-basket')"
+			>
+				Checkout now
+			</kv-ui-button>
+		</div>
+	</div>
+	<div
+		v-else
+		class="
+			tw-rounded
+			tw-bg-brand-50
+			tw-w-full
+			tw-py-2
+			tw-px-1
+			tw-mb-2
+			tw-flex-col
+			tw-flex
+			tw-justify-start
+			lg:tw-justify-between
+			lg:tw-align-center
+			"
+	>
+		<div class="lg:tw-mr-1 lg:tw-ml-0 tw-ml-1 lg:tw-pt-1">
+			<kv-ui-button
+				v-show="basketLoans.length > 0"
+				@click="$emit('show-basket')"
+				class="tw-w-full"
 			>
 				Checkout now
 			</kv-ui-button>
--- a/src/components/CorporateCampaign/CampaignProgressBar.vue
+++ b/src/components/CorporateCampaign/CampaignProgressBar.vue
@@ -1,5 +1,54 @@
 <template>
 	<div
+		v-if="(promoAmount && promoAmount > 0) && !isMatchingCorporateCampaign"
+		class="
+			tw-rounded
+			tw-bg-brand-50
+			tw-w-full
+			tw-py-3
+			tw-px-1
+			tw-mb-2
+			tw-flex-col
+			tw-flex
+			lg:tw-flex-row
+			tw-justify-start
+			lg:tw-justify-between
+			lg:tw-align-center
+			"
+	>
+		<div class="tw-flex-grow">
+			<h4 v-if="creditLeft === promoAmount" class="tw-mb-1 tw-px-3">
+				Let's Get Started
+			</h4>
+			<h4 v-else-if="creditLeft > 0" class="tw-mb-1 tw-px-3">
+				Keep Going!
+			</h4>
+			<h4 v-else-if="creditLeft === 0" class="tw-mb-1 tw-px-3">
+				You Did It!
+			</h4>
+			<h3 class="tw-mb-2 tw-px-3">
+				You have ${{ creditLeft }} in credit left
+				<span v-if="campaignPartnerName"> from {{ campaignPartnerName }} </span>
+			</h3>
+			<kv-grid class="tw-grid-cols-2">
+				<kv-progress-bar
+					:value="percentageLeft"
+					class="tw-mb-1.5 lg:tw-mb-1 tw-ml-3 tw-col-span-1"
+					:aria-label="`You have $${ creditLeft } in credit left`"
+				/>
+			</kv-grid>
+		</div>
+		<div class="lg:tw-mr-3 lg:tw-ml-0 tw-ml-3 lg:tw-pt-3">
+			<kv-ui-button
+				v-show="basketLoans.length > 0"
+				@click="$emit('show-basket')"
+			>
+				Checkout now
+			</kv-ui-button>
+		</div>
+	</div>
+	<div
+		v-else-if="basketLoans.length > 0"
 		class="
 			tw-rounded
 			tw-bg-brand-50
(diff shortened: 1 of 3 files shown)


src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -47,6 +47,32 @@
...
+		<div class="lg:tw-mr-1 lg:tw-ml-0 tw-ml-1 lg:tw-pt-1">
+			<kv-ui-button
+				v-show="basketLoans.length > 0"
+				@click="$emit('show-basket')"
+				class="tw-w-full"
+			>
+				Checkout now
+			</kv-ui-button>
+		</div>
+	</div>
```
dyersituations, line 75: Does there need to be a 3rd UI fallback if the two cases in this template are not valid? So when it's not a matching campaign and there are no items in the basket?

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -47,6 +47,32 @@
 			</kv-ui-button>
 		</div>
 	</div>
+	<div
+		v-else
+		class="
+			tw-rounded
+			tw-bg-brand-50
+			tw-w-full
```
dyersituations, line 55: In response to your question about the button width, you could just only apple `tw-w-full` in mobile view, so the button doesn't stretch. Then the decision will have to be made on where the button sits (left, center, right) in desktop view.

tiarascoleman:
How `commit-6331c20` looks: 

(This refers to the third UI default, when there is no Corporate Campaign or Matching Campaign)
Screen Shot 2023-07-19 at 3 52 42 PM

(Refers to desktop mode)
Screen Shot 2023-07-19 at 3 49 40 PM

(Refers to mobile mode)
Screen Shot 2023-07-19 at 3 49 55 PM

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -47,6 +47,32 @@
 			</kv-ui-button>
 		</div>
 	</div>
+	<div
+		v-else
```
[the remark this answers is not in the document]
tiarascoleman: @dyersituations I was having trouble with the following code line `v-else-if="(basketLoans.length > 0) && isMatchingCampaign` so I changed it to `v-if=(promoAccount == promoAccount > 0) && !isMatchingCampaign` because I reread the ticket and I believe MatchingCampaigns are not supposed to have any promo credit at all ?

src/components/CorporateCampaign/CampaignProgressBar.vue
tiarascoleman: @dyersituations Yes I'm still unsure what the fallback should be and have improvised with just a empty progress bar section- I'll likely reach with Sara about it!

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -86,6 +122,9 @@ export default {
 			}
 			return this.promoData?.promoFund?.displayName ?? null;
 		},
+		isMatchingCampaign() {
+			return this.pageSettingData?.matcherAccountId !== undefined;
```
mcstover, line 126: This information will not be available here. Please just pass down a new prop into this component for `isMatchingCampaign` from the CCLanding so we don't duplicate the logic here.

mcstover: What I've suggested will make the other check in `campaignPartnerName` also function correctly which is currently referencing a missing prop...

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -15,30 +31,50 @@
...
+				<kv-ui-button
+					v-show="basketLoans.length > 0"
+					@click="$emit('show-basket')"
+				>
+					Checkout now
+				</kv-ui-button>
+			</div>
 		</div>
-		<div class="lg:tw-mr-3 lg:tw-ml-0 tw-ml-3 lg:tw-pt-3">
+		<!-- Only checkout button appears when there is a matching campaign with no promo credit -->
```
dyersituations, line 66: This else will trigger in other cases. Currently the if above shows if no matching campaign and promo credit. That means this else will show if there is no promo credit or it's a matching campaign, it doesn't check for both. Do we only want to show this button "when there is a matching campaign with no promo credit" or are we okay with the two cases I just mentioned?

mcstover: I think it's correct to show the basket button if there are loans in the basket and the other conditions are not met. This will ensure visibility of the Basket button during general usage of page without promo credit applied.

@tiarascoleman You can change `v-else` to `v-else-if="basketLoans.length > 0"` and remove the check on the nested button below which will prevent just the blank green box from showing up.

tiarascoleman: @dyersituations Thanks for the catch! Just updated Line 19, which is the v-if statement. The main focus is that when there is not any promo credit (usually in the case of there being a matching campaign), the progress bar should be hidden and when there is promo credit, the meter should appear (I believe). In other words, the focus is more on if promoCredit is present or not and hopefully this new iteration of the v-if statement reflects this! :)

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -86,6 +122,9 @@ export default {
 			}
 			return this.promoData?.promoFund?.displayName ?? null;
 		},
+		isMatchingCorporateCampaign() {
+			return this.isMatchingCampaign;
+		},
```
mcstover, line 127: @tiarascoleman my apologies for the confusion but this computed property is not needed.

In order to make your component inherit the existing `isMatchingCampaign` that we've established in CCLanding you'll need to do the following:
1. Update the ``  object [here](https://github.com/kiva/ui/blob/main/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue#L793), by adding `isMatchingCampaign: this.isMatchingCampaign` along side the other key value pairs.
2. Update [CampaignLoanWrapper](https://github.com/kiva/ui/blob/main/src/components/Contentful/CampaignLoanWrapper.vue#L23) to pass a new prop into the CampaignProgressBar component by adding this: `:is-matching-campaign="componentProps.isMatchingCampaign"`
3. Update [CampaignProgressBar](https://github.com/kiva/ui/blob/main/src/components/CorporateCampaign/CampaignProgressBar.vue#L80) to add a new incoming prop declaration to match the prop your passing in:
```
		isMatchingCampaign: {
			type: Boolean,
			default: false
		},
```

With the listed additions in place you can remove this property b/c you will now have `isMatchingCampaign` available for use in your template and `this.isMatchingCampaign` available for use in your code.

src/components/CorporateCampaign/CampaignProgressBar.vue
```diff
@@ -1,7 +1,23 @@
...
+			tw-flex-col
+			tw-flex
+			tw-justify-start
+			lg:tw-justify-between
+			lg:tw-align-center
+			"
+	>
+		<!-- Progress bar appears as long as there is promo credit -->
+		<div
+			v-if="(promoAmount && promoAmount > 0) && (isMatchingCorporateCampaign || !isMatchingCorporateCampaign)"
```
mcstover, line 19: Once the proper passing of props has been resolved as listed below, this should be `v-if="(promoAmount && promoAmount > 0) && !isMatchingCampaign"`

dyersituations (approved):

mcstover (approved):
Thanks for getting this fixed up!

PR #4917, opened 2023-07-19
feat: carousel updated for mobile version in lending home page

roger-in-kiva:
- carousel updated for mobile version in lending home page

New mobile version
Screenshot 2023-07-19 at 15 59 27

Desktop version stills as current
Screenshot 2023-07-19 at 16 05 01

### DIFF
--- a/src/components/LoanFinding/LendingCategorySection.vue
+++ b/src/components/LoanFinding/LendingCategorySection.vue
@@ -1,7 +1,8 @@
 <template>
 	<div class="tw-w-full">
-		<div class="tw-mx-auto tw-px-2.5 md:tw-px-4 lg:tw-px-8" style="max-width: 1200px;">
-			<div class="tw-flex tw-flex-col lg:tw-flex-row tw-justify-between tw-items-end lg:tw-items-center">
+		<div class="tw-mx-auto tw-px-0 md:tw-px-4 lg:tw-px-8" style="max-width: 1200px;">
+			<!-- eslint-disable-next-line max-len -->
+			<div class="tw-flex tw-flex-col lg:tw-flex-row tw-justify-between tw-items-end lg:tw-items-center tw-px-2.5 md:tw-px-0">
 				<div class="tw-w-full lg:tw-w-auto">
 					<h2 v-html="title" class="tw-text-h2 tw-text-primary"></h2>
 					<p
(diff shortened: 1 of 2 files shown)


dyersituations:
I'm seeing this for the mobile version (which starts at 733px). The carousel should have one loan per page being shown. I know this is behavior of the existing carousel, so we might need a new version of the carousel component?

image

roger-in-kiva:
> I'm seeing this for the mobile version (which starts at 733px). The carousel should have one loan per page being shown. I know this is behavior of the existing carousel, so we might need a new version of the carousel component?
> 
> image

Adding a new version of the carousel component is a great idea, mainly if we can incorporate TW features inside it. Like grid or flex classes to have responsive loan card sizes based on the tailwind breakpoints.
The current behavior is b/c we are hard-coding the single slide width in every case the carousel is implemented, like here: https://github.com/kiva/ui/blob/ce763c47df526b85b6f47b5481c13eb00d360926/src/components/LoanFinding/LendingCategorySection.vue#L106

For this specific case, lending home component, I see two options to fix it:
1. Update the hard-coded single slide width for the screen size you mentioned.
2. Try to have a reactive width based on the screen size; this can solve the issue of the slide width as the screen size is expanded. Ex. load the page in small screen size and then expand it; the slide will keep the width as it was loaded, small, showing a larger number of loans than expected. This is common in recommended loans row.

dyersituations:
> 1. Update the hard-coded single slide width for the screen size you mentioned.
> 2. Try to have a reactive width based on the screen size; this can solve the issue of the slide width as the screen size is expanded. Ex. load the page in small screen size and then expand it; the slide will keep the width as it was loaded, small, showing a larger number of loans than expected. This is common in recommended loans row.

Yeah I think either of those options seem valid. What is a simpler path to get the single loan card flush with the edges of the screen for mobile, according to mocks.

src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -197,6 +193,11 @@ export default {
 			return `loan-card-${index}`;
 		},
 	},
+	mounted() {
+		window.addEventListener('resize', () => {
```
dyersituations, line 197: The event listener needs to be removed in `beforeDestroy`. Also, the resize should be throttled. A good example of both is in CPS in the `QuickFiltersController`. Notice how the throttled method needs to be stored in a variable so that the remove event listener can find the method to remove.

dyersituations:
The last small UI tweak for the `kv-carousel` is to remove the left + right padding for this variation, so that the card goes all the way to the sides of the browser window.

roger-in-kiva:
> The last small UI tweak for the `kv-carousel` is to remove the left + right padding for this variation, so that the card goes all the way to the sides of the browser window.

Thanks for the comments!
I think that as a future work we can build a KvCarouselController to wrap all the existing carousels in `ui` including the listener and the tailwind breakpoints to have a consistent component for this. Every case is implemented in a different way.

dyersituations (approved):

PR #4920, opened 2023-07-26
feat: UI App Login Handler Updates

michelleinez:


mcstover (approved):
May want to update the comment to something like: "Signify ssoRedirect action, value should be a valid sso connection name"

PR #4921, opened 2023-07-27
fix: auth0 repo requires camel case for these variables

michelleinez:


server/auth-router.js
```diff
@@ -107,7 +107,7 @@ module.exports = function authRouter(config = {}) {
 		}
 		// Specify ssoRedirect url
 		if (req.query.ssoRedirect) {
-			options.sso_connection = req.query.ssoRedirect;
+			options.ssoRedirect = req.query.ssoRedirect;
```
dyersituations, line 110: Just out of curiosity, it seems that `login_hint` currently does work or is there is difference in how these two options are handled?

mcstover: right, is this linting or something from the Auth0 docs we didn't catch...

mcstover (approved):

server/auth-router.js
dyersituations: Oh it looks like it's because `login_hint` has custom handling in `entry.js`, where the `auth0` PR uses the field directly from the config. We tend to use camel case for parameters, so using this directly seems good.

dyersituations (approved):

PR #4924, opened 2023-07-28
feat: send uiab experiment assignments in x-uiab header to graphql api

emuvente:


mcstover (approved):

src/api/ExperimentIdLink.js
```diff
@@ -25,17 +25,24 @@ function buildExpHeaders(cache) {
...
+export default ({ cookieStore }) => {
 	return setContext((operation, previousContext) => {
-		// Fetch experiment header value
+		// Initialize new context
+		const newContext = { ...previousContext };
+		newContext.headers = newContext.headers ?? {};
+
+		// Fetch experiment header values
 		const experimentHeader = buildExpHeaders(previousContext?.cache);
+		const uiabValue = cookieStore.get(UIAB_COOKIE_NAME) ?? '';
```
dyersituations, line 36: Keep in mind that we don't actually store the assignments in the `uiab` cookie unless the assignment is forced via URL (or it's a user with an old `uiab` cookie from before the experimentation changes a few months ago).

emuvente: That's exactly what we need this for, passing the forced assignments to the api.

emuvente:
Pairs with kiva/kiva#11802

src/api/ExperimentIdLink.js
dyersituations: Cool, good to know!

dyersituations (approved):

PR #4925, opened 2023-07-28
feat: remove exp code for control version

christian14b:
Remove `control` version of share ask copy experiment. Winner `variant`


emuvente (approved):
👍

PR #4926, opened 2023-07-31
feat: 5 dollars banner added to lending home page

roger-in-kiva:
- $5 banner added to lending home page

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -1,6 +1,8 @@
 <template>
 	<www-page main-class="tw-bg-white" style="height: auto;">
 		<div class="tw-w-full">
+			<five-dollars-banner v-if="showFiveDollarsBanner" class="tw-mb-2" />
+
 			<!-- eslint-disable-next-line max-len -->
 			<div v-if="showWelcomeMsg" class="tw-mx-auto tw-p-2 tw-py-1 lg:tw-pt-3 tw-px-2.5 md:tw-px-4 lg:tw-px-8" style="max-width: 1200px;">
 				<h3 class="tw-text-h3 tw-text-primary">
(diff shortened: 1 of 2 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -396,6 +402,19 @@ export default {
 		trackSpotlightDisplayedLoans() {
 			this.trackDisplayedLoans('spotlight', 4, this.spotlightLoans);
 		},
+		check5DollarsBannerCookie() {
+			const currentDate = new Date();
+			const dateCookie = this.cookieStore.get('kvfivedollarsbanner');
```
dyersituations, line 407: Store magic string in a const at the top of this method.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -396,6 +402,19 @@ export default {
...
+			if (!dateCookie) {
+				this.cookieStore.set('kvfivedollarsbanner', currentDate);
```
dyersituations, line 409: In many situations we pass in `{ path: '/' }` as the third parameter to `.set`. Is that needed here?

dyersituations: What is the default cookie expiration here?

roger-in-kiva: From docs, `If omitted, the cookie becomes a session cookie.`

Should we add an expiration time?

dyersituations: Do we want it to be a session cookie? We check for 3 days elsewhere, so I'm guessing a session cookie would be long gone by then.

roger-in-kiva: Right! Agree by 6 months? it would be there for experiment now, but if becomes a feat we need to extend this expiration time

dyersituations: 6 months is probably fine, but maybe discuss with the team?

roger-in-kiva: Sounds good, thanks!

dyersituations (approved):

PR #4927, opened 2023-08-01
feat: handle ui login kiva_corp_partner param

dyersituations:


mcstover (approved):

emuvente (commented):
Have you tested this out yet? Because I don't think a custom option like that is going to be passed to the /login page on Auth0. That's why we've had to put custom config options in the login_hint as JSON. I assume that's a problem for the ssoRedirect parameter as well though, so if that's actually working then maybe Auth0 changed that behavior and we don't need to stuff custom options in the login_hint anymore.

dyersituations:
> Have you tested this out yet? Because I don't think a custom option like that is going to be passed to the /login page on Auth0. That's why we've had to put custom config options in the login_hint as JSON. I assume that's a problem for the ssoRedirect parameter as well though, so if that's actually working then maybe Auth0 changed that behavior and we don't need to stuff custom options in the login_hint anymore.

Oh yup, we did test the `ssoRedirect` parameter, hence why I used that simpler pattern. This URL will show you the new variation of the page (note there's a minor UI bug with the button solved by the latest small PR: https://www.dev.kiva.org/ui-login?kivaCorpPartner=true)

PR #4928, opened 2023-08-03
Productize mobile borrower profile mars 466

christian14b:
Move `LendCta` section to be at the top only on mobile and at the bottom when it's not visible.

## Mobile view:
image

image

### Logged in:

image

## Tablet:

image

## Desktop:
image

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -399,10 +399,6 @@ export default {
 			type: Boolean,
 			default: false,
 		},
-		isMobile: {
-			type: Boolean,
-			default: false
-		}
 	},
 	components: {
 		LendAmountButton,
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -36,7 +36,6 @@
 				>
 					<lend-cta
 						class="tw-pointer-events-auto"
-						:is-mobile="isMobile"
 						:loan-id="loanId"
 						:enable-five-dollars-notes="enableFiveDollarsNotes"
 					>


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -222,20 +227,32 @@
...
+								'tw-hidden': isSticky,
+							}
+						]"
 					>
 						<loan-bookmark
 							v-if="isLoggedIn"
 							data-testid="bp-lend-cta-loan-bookmark"
 							:loan-id="loanId"
-							class="tw-hidden md:tw-inline-block lg:tw-hidden"
+							class="tw-inline-block md:tw-hidden"
```
mcstover, line 252: Previously this could only be hidden here b/c it would be shown in the other component. But now it's removed in the other component. We need to make sure it is visible for a logged in user on all screen sizes.

mcstover: The code here looks like it's going to be hidden on all screens except mobile...

roger-in-kiva:
Looks like is blocking some links of the footer

Screenshot 2023-08-03 at 12 38 41

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -6,26 +6,29 @@
...
-					'tw-px-2.5',
-					'tw-bg-primary',
-					'tw-border-t tw-border-tertiary',
 					{
 						'md:tw-rounded-b': !isSticky,
 						'md:tw-border-none': !isSticky,
 						'md:tw-px-3': !isSticky,
 						'md:tw-px-4': isSticky,
+						'tw-bg-transparent': !isSticky && isMobile,
+						'tw-bg-primary': !isMobile || isSticky,
```
emuvente, line 30: Can we apply these styles entirely with classes instead of relying on `isMobile`? I think this will work:
```
'tw-bg-transparent md:tw-bg-primary': !isSticky,
'tw-bg-primary': isSticky,
```

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -222,20 +227,32 @@
...
+								'tw-hidden': isSticky,
+							}
+						]"
 					>
 						<loan-bookmark
 							v-if="isLoggedIn"
 							data-testid="bp-lend-cta-loan-bookmark"
 							:loan-id="loanId"
-							class="tw-hidden md:tw-inline-block lg:tw-hidden"
+							class="tw-inline-block lg:tw-hidden"
```
mcstover, line 252: It still seems we're just arbitrarily hiding the bookmark function on desktop... not sure why but I think this was supposed to be something temporary due to other features being added.

mcstover: @eddieferrer wasn't that part of the new share button work? Can we show this on desktop now that this PR also makes the share-button visible at the same code point.

mcstover (approved):
Pending questions over bookmark visibility, but good overall.

src/components/BorrowerProfile/LendCta.vue
mcstover: Yes

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -383,6 +399,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		isMobile: {
```
emuvente, line 402: I don't think this is needed anymore, right?

src/components/BorrowerProfile/LendCta.vue
emuvente: It makes sense to me to not hide this at all. From our last team chat it seemed clear that the intent for the bookmark button is to always display it for logged-in users.

emuvente (approved):
Looks great!

PR #4937, opened 2023-08-15
feat: initial exp setup including query/mutation files

mcstover:
- Initializes an experiment to control validation and rollout of the new query
- Includes new mutation files for async kiva credit and braintree checkouts
- Implements a switch between async vs standard mutation files based on exp version
- Includes a new query file for checkoutStatus

Todo:
- Create util method to poll for checkout status
- Handle checkout status completion + error return values


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -493,6 +498,21 @@ export default {
 		// TODO: Implement check against contentful setting
 		// to signify if holiday mode is enabled
 
+		// VUE-1725 Async Checkout Mutation Rollout
+		const asyncCheckoutExp = this.apollo.readFragment({
+			id: `Experiment:${ASYNC_CHECKOUT_EXP}`,
+			fragment: experimentVersionFragment,
+		}) || {};
+
+		this.asyncCheckoutExpVersion = asyncCheckoutExp?.version;
```
dyersituations, line 507: This could instead store `asyncCheckoutExp?.version === 'b'`

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -493,6 +498,21 @@ export default {
...
+		if (this.asyncCheckoutExpVersion && !this.emptyBasket) {
```
dyersituations, line 508: Do we only want to store whether the exp is enabled if `!this.emptyBasket`?

src/graphql/query/checkout/checkoutStatus.graphql
```diff
@@ -0,0 +1,11 @@
+query checkoutStatus($transactionId: String!, $visitorId: String!) {
+	checkoutStatus(transactionId: $transactionId, visitorId: $visitorId) {
```
dyersituations, line 2: Is this being used?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -159,6 +160,10 @@ export default {
 		promoGuestCheckoutEnabled: {
 			type: Boolean,
 			default: false,
+		},
+		useAsyncCheckout: {
```
dyersituations, line 164: Is there more UI work needed before we can roll out? Like polling status and whatnot? I'm assuming we won't know what's happening currently after this async checkout is triggered.

src/pages/Checkout/CheckoutPage.vue
mcstover: True I could convert that to `asyncCheckoutActive` instead...

src/pages/Checkout/CheckoutPage.vue
mcstover: Correct. no need to track or swap mutations if no checkout can take place. Adding something to the basket would activate this block.

src/graphql/query/checkout/checkoutStatus.graphql
mcstover: Follow commits will use this in the polling function. Unless I end up setting that up in kv-shop, in which case I can remove this file.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: The only UI work for this initial pass will be to ensure the "loading" state is active and to show and error if it's returned during polling.

src/pages/Checkout/CheckoutPage.vue
dyersituations: In that case, should we move setting the data prop into the if?

mcstover: Yea, I suppose we could move the readFragment and setting the version var into a check for `!emptyBasket`

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -493,6 +498,21 @@ export default {
...
+		}) || {};
+
+		this.asyncCheckoutActive = asyncCheckoutExp?.version === 'b';
+		if (asyncCheckoutExp?.version !== null) {
+			this.$kvTrackEvent(
+				'Basket',
+				'EXP-VUE-1725-Aug2023',
+				this.asyncCheckoutExpVersion,
+			);
+		}
```
mcstover, line 514: I've removed the gates here for initialization to ensure this takes affect from an empty basket state. The tracking is fine to fire at all times.

mcstover:
In order to keep this iterative, I'd like to move forward with getting this PR merged. Default state is "off/false".

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -493,6 +498,21 @@ export default {
...
 		// to signify if holiday mode is enabled
 
+		// VUE-1725 Async Checkout Mutation Rollout
+		const asyncCheckoutExp = this.apollo.readFragment({
+			id: `Experiment:${ASYNC_CHECKOUT_EXP}`,
+			fragment: experimentVersionFragment,
+		}) || {};
+
+		this.asyncCheckoutActive = asyncCheckoutExp?.version === 'b';
+		if (asyncCheckoutExp?.version !== null) {
```
dyersituations, line 508: Should this just be `if (asyncCheckoutExp?.version)`? I think the default "no assignment" is `undefined`

mcstover: I'm getting this for the version when there is no setting. Should I check against both null and undefined?
`asyncCheckoutExp Object { __typename: "Experiment", id: "async_checkout_rollout", version: null }`

dyersituations: We've just been comparing against truthy. But I was wondering if the default from apollo would be `null` instead.

dyersituations (approved):

mcstover:
Planning to merge this after the release next Monday.

PR #4938, opened 2023-08-15
feat: exclude countries from lending stats page based on an admin set…

eddieferrer:
…ting

Also refactors uses of lodash on the lending stats page

### DIFF
--- a/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
+++ b/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
@@ -88,8 +88,6 @@
 
 <script>
 import _differenceBy from 'lodash/differenceBy';
-import _get from 'lodash/get';
-import _map from 'lodash/map';
 import _sortBy from 'lodash/sortBy';
 import lendingStatsQuery from '@/graphql/query/myLendingStats.graphql';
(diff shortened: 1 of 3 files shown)


src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -136,23 +134,29 @@ export default {
 		query: lendingStatsQuery,
 		preFetch: true,
 		result({ data }) {
-			const allCountries = _sortBy(_map(_get(data, 'lend.countryFacets'), 'country'), 'name');
-			this.countriesLentTo = _sortBy(_get(data, 'my.lendingStats.countriesLentTo'), 'name');
+			const countriesWhereKivaDoesntWork = data?.general?.excluded_countries?.value ?? '';
```
dyersituations, line 137: Should the fallback be `[]`?

eddieferrer: No settings always come back as strings. This is actually just a string of comma separated values. Down below. `.includes` works on strings anyway

dyersituations (approved):

emuvente (approved):

PR #4939, opened 2023-08-16
feat: new loan card added for empty basket state in checkout page

roger-in-kiva:
- new loan card added for empty basket state in checkout page
- having 4 loans as current is very tight with new loan card

Screenshot 2023-08-16 at 17 32 12


src/components/RandomLoanSelector/RandomLoanSelector.vue
```diff
@@ -22,18 +21,18 @@
...
 				>
 					<template v-for="(loan, index) in randomLoans" #[`slide${index}`]>
-						<minimal-loan-card
-							:key="index"
-							:class="`minimal-loancard-${index}`"
-							:loan="loan"
-							category-set-id="random-loans"
-							:card-number="index"
-							@refreshtotals="$emit('refreshtotals')"
-							@updating-totals="$emit('updating-totals', $event)"
```
mcstover, line 36: Can we please restore this functionality by emitting an `updating-totals` event from the `addToBasket` method within KvClassicLoanCardContainer? This emit would be at the beginning of that method along with `this.isAdding = true;`, before the mutation occurs.

As it stands there is no loading state triggered until after the add to basket mutation has completed and emitted the addToBasket event. Ideally, we trigger a loading state for the page on click. The `refreshTotals` method you've set below will handle removing the loading state.

roger-in-kiva: ok, that makes sense. thanks for checking!

dyersituations (approved):

PR #4943, opened 2023-08-21
fix: unused var removed in saved search component

roger-in-kiva:
- var removed as is no longer being used


dyersituations (approved):
Thanks!

PR #4946, opened 2023-08-28
feat: test setup added to show three loans in recommended row for len…

roger-in-kiva:
- test setup added to show three loans in recommended row for lending home page
- there is another ticket to remove setup for featured loan test

### DIFF
--- a/src/pages/LoanFinding/LoanFinding.vue
+++ b/src/pages/LoanFinding/LoanFinding.vue
@@ -33,6 +33,7 @@
 				:enable-five-dollars-notes="enableFiveDollarsNotes"
 				:enable-relending-exp="enableRelendingExp"
 				:user-balance="userBalance"
+				:per-step="perStepRecommendedRow"
 				@add-to-basket="trackCategory($event, 'recommended')"
 				:class="{
 					'tw-pt-3 tw-pb-4 tw-bg-secondary' : enableRelendingExp,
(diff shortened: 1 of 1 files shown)


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -160,13 +162,14 @@ export default {
...
 			showFiveDollarsBanner: false,
+			enableThreeLoansRecommended: false,
 			HandOrangeIcon,
 		};
 	},
 	apollo: {
 		preFetch(config, client) {
 			return client.query({
-				query: experimentAssignmentQuery, variables: { id: RECOMMENDED_REPLACEMENT_EXP_KEY }
+				query: experimentAssignmentQuery, variables: { id: THREE_LOANS_RECOMMENDED_ROW_EXP_KEY }
```
dyersituations, line 172: Don't we still need to fetch `RECOMMENDED_REPLACEMENT_EXP_KEY`? I see it being used in the `.then` of this query. Although I guess we doing a global prefetch of assignments, and none of these assignment calls are needed in components? Please verify locally.

roger-in-kiva: We don't need to fetch `RECOMMENDED_REPLACEMENT_EXP_KEY` more.
For that reason we have other ticket to remove this setup as mentioned in the description above.

dyersituations (approved):

PR #4947, opened 2023-08-28
feat: add partner with us page to about dropdown

christian14b:


src/components/WwwFrame/TheHeader.vue
```diff
@@ -287,6 +287,14 @@
 												About us
 											</router-link>
 										</li>
+										<li>
+											<router-link
```
dyersituations, line 291: I think this needs to be an anchor tag, since the page is being served by CPS.

emuvente: Technically it's ok as a router-link, because this beforeEach handler in the client will change the window location for any router-links that don't have routes: https://github.com/kiva/ui/blob/f1abf78b3f49df954eff3d6d03deff92f6a704e5/src/client-entry.js#L167-L173

emuvente (approved):

src/components/WwwFrame/TheHeader.vue
emuvente: We've done other about pages as anchor links though, so makes sense to switch.

dyersituations (approved):

emuvente (approved):

PR #4950, opened 2023-08-29
feat: display flash messages for visitor in the tip message MARS-448

emuvente:
Checks for FlashMessages from the api while still using the TipMessage from the local resolver. Displays whichever has a higher level first (error being the highest, then warning, then tip), favoring the tipMessage in a tie. Displays the next available message after the current message is closed.

### DIFF
--- a/src/components/WwwFrame/TheTipMessage.vue
+++ b/src/components/WwwFrame/TheTipMessage.vue
@@ -1,15 +1,52 @@
 <template>
 	<kv-toast
 		ref="tip"
-		@close="$closeTipMsg"
+		@close="closeCurrentMessage"
 		class="tw-fixed tw-top-9 md:tw-top-11 tw-left-0 tw-right-0 tw-z-banner"
 	/>
 </template>
 
 <script>
+import { captureException } from '@sentry/vue';
+import { gql } from '@apollo/client';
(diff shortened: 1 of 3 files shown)


src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -21,37 +58,94 @@ export default {
...
+			// clear all flash messages
+			return this.apollo.mutate({
+				mutation: clearFlashMessagesMutation,
+				variables: {
+					visitorId: this.cookieStore.get('uiv'),
+				},
+			});
+		}).catch(err => {
+			try {
+				console.error(err);
```
dyersituations, line 96: `logFormatter`?

dyersituations (approved):

mcstover (approved):
👍

src/components/WwwFrame/TheTipMessage.vue
emuvente: I was going to say that this is only for the browser console, in which case I prefer to just spit out the error using the formatting provided by `console.error`, but I realized this is actually in created so it could get called on the server. I'm going to move this into mounted so that it's only called in the browser.

dyersituations (approved):

mcstover (approved):

PR #4951, opened 2023-08-30
fix(TheTipMessage): tipMessage missing from currentMessage and incorrect persist logic

emuvente:
Tip messages from $showTipMsg aren't appearing due to `type` being an empty string by default which the existing logic of `currentMessage` doesn't account for. This PR fixes that and also fixes a mistake in the `persist` logic that persisted TIP types instead of ERROR types.

### DIFF
--- a/src/components/WwwFrame/TheTipMessage.vue
+++ b/src/components/WwwFrame/TheTipMessage.vue
@@ -103,6 +103,10 @@ export default {
 	},
 	computed: {
 		currentMessage() {
+			// return tip message if there are no flash messages
+			if (!this.flashMessages?.length) {
+				return this.tipMessage;
+			}
 			// return tip message if it's level is higher than first flash message, otherwise return first flash message
 			return this.tipMessage?.type && compareLevels(this.tipMessage.type, this.flashMessages[0]?.messageType)
 				? this.tipMessage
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheTipMessage.vue
```diff
@@ -103,22 +103,26 @@ export default {
...
+			return !!this.currentMessage?.persist || type === ERROR || type === WARNING;
 		},
 	},
 	methods: {
 		showCurrentMessage() {
 			const message = this.currentMessage?.message ?? '';
 			const safeMessage = DOMPurify.sanitize(message, { ALLOWED_TAGS: ['b', 'i', 'em', 'strong', 'a'] });
 			const type = this.currentMessage?.type ?? this.currentMessage?.messageType ?? '';
-			this.$refs.tip?.show(safeMessage, type, this.persist);
+			this.$refs.tip?.show(safeMessage, type === TIP ? '' : type, this.persist);
```
emuvente, line 125: KvToast doesn't handle the 'tip' type yet, so passing an empty string to get the default instead for now.

mcstover (approved):

dyersituations (approved):

PR #4954, opened 2023-09-06
Core 1352 remove old loan card experiment

roger-in-kiva:
- new loan card test code removed
- unnecessary components removed
- loan card for experiment replaced with shared loan card
- query for tipMessage fixed to avoid a lint error about missed id


emuvente (approved):
Thanks!

mcstover (approved):

PR #4957, opened 2023-09-08
feat: implement optional async checkout muatations and status check

mcstover:
This PR implements the async mutations for kiva credit and braintree checkout and adds an additional polling function to check the status of the async checkout operation. Once the status is complete, we use the checkout/transaction id to move through the existing post-process and on to the thanks page.


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -339,39 +340,61 @@ export default {
...
-							'basket',
-							`${paymentType} Braintree DropIn Payment`,
-							'Success',
-							transactionId,
-							transactionId
-						);
-						// Complete transaction handles additional analytics + redirect
-						this.$emit('complete-transaction', transactionId);
-					}
+					this.handleSuccessfulCheckout(transactionResult, paymentType);
```
dyersituations, line 370: Should this code be in an `else`? It looks like the `if` above doesn't `await`, so this code will be hit while the async checkout is polling.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -339,39 +340,61 @@ export default {
 					},
 				})
 				.then(kivaBraintreeResponse => {
+					// extract transaction saga id or transaction id from response
+					const transactionResult = this.useAsyncCheckout
+						? kivaBraintreeResponse?.data?.shop?.doNoncePaymentDepositAndCheckoutAsync
+						: kivaBraintreeResponse?.data?.shop?.doNoncePaymentDepositAndCheckout;
+
+					if (this.useAsyncCheckout && typeof transactionResult !== 'object') {
+						pollForCheckoutStatus(this.apollo, transactionResult)
```
dyersituations, line 349: What happens if the user navigates away while this is polling? Will the complete transaction logic never run?

src/graphql/query/checkout/checkoutStatus.graphql
```diff
@@ -3,6 +3,9 @@ query checkoutStatus($transactionId: String!, $visitorId: String) {
 		basketId
 		errorCode
 		errorMessage
+		receipt {
+		checkoutId
```
dyersituations, line 7: Missing tab

src/util/checkoutUtils.js
```diff
@@ -120,3 +121,68 @@ export function removeCredit(apollo, creditType) {
...
+ * - STARTED
+ * - TRANSIENT_PAYMENT_METHOD_CHARGED
+ *
+ * @param {Object} apollo Apollo Client instance
+ * @param {Number} transactionId
+ * @param {Number} interval How often to poll
+ * @param {Number} timeout How long to allow polling to continue
+ * @returns {Promise}
+ */
+export async function pollForCheckoutStatus(
```
dyersituations, line 152: Need unit tests?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: This process uses the same page level "loading" state as the existing checkout mutation. So it's certainly possible for someone to decide not to wait on either flow to complete. In either case the checkout can complete in the background even if the post-process or thanks page are somehow blocked from occurring.

dyersituations: So nothing from the UI is needed to "complete" the transaction? It should complete fine in the background?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: The if block above doesn't contain any async operations and should fire the various tracking/logging methods and return if there are errors.

dyersituations: I meant the `if` on line 348. It calls the async checkout and has `handleSuccessfulCheckout` in the callback. But this line will also call `handleSuccessfulCheckout`, since there is no `else`.

dyersituations: Oh never mind, I missed the `return`. Resolving this comment.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: Correct, we've already called one of the checkout mutations at this point and gotten it's response. We are now just checking for the status IF we've used the async version.

mcstover: But yea I do see some potential cross over between the processes so will think on that some more.

src/util/checkoutUtils.js
```diff
@@ -120,3 +121,68 @@ export function removeCredit(apollo, creditType) {
...
+	apollo = null,
+	transactionSagaId = 0,
+	interval = 1000,
+	timeout = 60000
```
emuvente, line 156: The existing checking process already times out after 60 seconds, so it seems like this will have the same problem. What if we went with 5 minutes? That's the default connection timeout for most browsers.

src/util/checkoutUtils.js
mcstover: Working on moving this method into kv-shop and adding tests there. Will remove from this PR once that is in place

dyersituations:
@mcstover Pushing up the history from `main` complicated this PR a bit, since the branch was published. I created a new branch, pulled in the appropriate changes. Tomorrow I'll push up what I've done so far. The history of that branch will be a lot simpler. I'll leave this PR and branch up until we know we don't need them anymore.

mcstover:
Superceded by: https://github.com/kiva/ui/pull/5164

PR #4960, opened 2023-09-12
feat: create donation module in impact dashboard

christian14b:
- Add new "Your donations" module
- Remove the “average donations per $25” stat from “Your lending insights” module

image

### DIFF
--- a/src/pages/Portfolio/ImpactDashboard/YourDonations.vue
+++ b/src/pages/Portfolio/ImpactDashboard/YourDonations.vue
@@ -168,8 +168,7 @@ export default {
 						filters: transactionFilters,
 						limit: this.limit,
 					},
-				}).then(({ data }) => {
-					this.loading = false;
+				}).then(async ({ data }) => {
 					const currentYear = this.currentDate.getFullYear();
 					const transactions = data?.my?.transactions?.values ?? [];
 					const amountOfLoans = numeral(data?.my?.userStats?.amount_of_loans ?? 0);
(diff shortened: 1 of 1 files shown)


src/pages/Portfolio/ImpactDashboard/YourDonations.vue
```diff
@@ -0,0 +1,277 @@
...
+					if (amountOfLoans.value() > 0) {
+						donation.multiply(1 / amountOfLoans.value()).multiply(25);
+					}
+					this.averageDonation = donation.format('$0,0[.]00');
+
+					this.totalDonations = data?.my?.transactions?.totalCount ?? 0;
+					this.currentYearDonations = this.filterAndSumDonations(transactions, currentYear);
+					this.lastYearDonations = this.filterAndSumDonations(transactions, currentYear - 1);
+
+					if (this.totalDonations > this.limit) {
```
dyersituations, line 186: Is the goal to get the last X number of donations? It looks like if the total donations is greater than 100, we get the donations for 100+ transactions.

src/pages/Portfolio/ImpactDashboard/YourDonations.vue
```diff
@@ -0,0 +1,277 @@
...
+					}
+				}).finally(() => {
+					this.loadingPromise = null;
+				});
+			}
+		},
+		whenVisible() {
+			this.fetchAsyncData();
+		},
+		getRemainDonations(transactionFilters, offset) {
```
dyersituations, line 211: `getRemainingDonations`?

src/pages/Portfolio/ImpactDashboard/YourDonations.vue
```diff
@@ -0,0 +1,277 @@
...
+	}
+};
+</script>
+
+<style lang="postcss" scoped>
+.stat-placeholder {
+	margin-bottom: 6px;
+	@apply tw-mt-1 tw-h-4.5;
+
+	@screen md {
```
dyersituations, line 253: Do these nested `@screen` usages play nicely with postcss? I know we don't have the nesting module, but I wasn't sure if `@screen` counted as nesting here.

src/pages/Portfolio/ImpactDashboard/YourDonations.vue
dyersituations: Is this meant to be a slight performance boost?

emuvente: This is just paging, not a performance thing. We need all the donation transactions to be able to count the total donation amount, and the most results we can get back from the collect result in a single query is 100.

src/pages/Portfolio/ImpactDashboard/YourDonations.vue
```diff
@@ -0,0 +1,277 @@
...
+					this.averageDonation = donation.format('$0,0[.]00');
+
+					this.totalDonations = data?.my?.transactions?.totalCount ?? 0;
+					this.currentYearDonations = this.filterAndSumDonations(transactions, currentYear);
+					this.lastYearDonations = this.filterAndSumDonations(transactions, currentYear - 1);
+
+					if (this.totalDonations > this.limit) {
+						let offset = this.limit;
+						while (offset < this.totalDonations) {
+							this.getRemainingDonations(transactionFilters, offset)
```
emuvente, line 189: The way this works right now, the entire while loop will execute synchronously, kicking off a bunch of parallel requests for the rest of the results, and then complete before any of the results can come in. The `finally()` callback will be called before the rest of those requests complete.

You could maybe build an array of requests based on the totalCount and then return Promise.all(requests)

emuvente (approved):
Thanks for the updates, looks good!

PR #4961, opened 2023-09-13
feat: new process join team page

eddieferrer:
Simple page that requires auth that just take the query params and fire the join team mutation. This will allow us to have minimal clicks for joining a team for un-authenticated users in the new stack. 

I plan to link the "Join Team" buttons from the new team join pages `/team/join/<team>` and from the new `/teams` page to this page for un authenticated users (authenticated users can just fire the mutations directly from those pages)

Initially I was thinking of repurposing `/join-team` but found that that route is still firing a couple of events a month, so it is still in use, it was hard to track down how it is still in use (Something about a redirect from the legacy basket if a team invitation is pending in the monolith?)

### DIFF
--- a/src/pages/ProcessJoinTeam.vue
+++ b/src/pages/ProcessJoinTeam.vue
@@ -1,0 +1,183 @@
+<template>
+	<www-page>
+		<kv-page-container>
+			<kv-grid class="tw-grid-cols-12 tw-pt-3 md:tw-pt-5 lg:tw-pt-7 tw-mb-4 md:tw-mb-6">
+				<div
+					class="tw-col-span-12 md:tw-col-start-3 md:tw-col-span-8"
+				>
+					<div v-if="isLoading" class="tw-flex tw-items-center tw-flex-col">
+						<h2>Joining team {{ teamName }}</h2>
+						<div>
+							<kv-loading-spinner class="tw-mt-2" />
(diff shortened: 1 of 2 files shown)


src/pages/ProcessJoinTeam.vue
```diff
@@ -0,0 +1,183 @@
...
+					variables: {
+						teamPublicId: this.teamPublicId,
+					},
+					fetchPolicy: 'network-only'
+				});
+				// update the users membership status
+				this.userMembershipStatus = userMembershipData?.community?.team?.userProperties?.membershipStatus;
+				if (this.isMember) {
+					this.$showTipMsg('You have joined the team!');
+					this.memberRedirect();
```
dyersituations, line 157: Do we want a slight delay before redirect so that the user sees the toast?

mcstover: I was thinking the same! Delay 5-10 secs would at least allow someone to scan/read the page.

mcstover (approved):

src/pages/ProcessJoinTeam.vue
eddieferrer: I considered that approach, I think its fine, the delay will be the page loading :)

dyersituations (approved):

PR #4965, opened 2023-09-14
fix: filter out unhanded google ads code request errors from sentry

mcstover:
This should clear up the false alarms for UI checkout errors

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -81,6 +81,12 @@ export default function createApp({
 				// make sentry colleted event easy to compare to
 				const eventAsString = JSON.stringify(event);
 				// match specific 3rd party events for exclusion
+				// Skip sending failed to fetch error caused by unhandled promise rejection in google ads
+				// Sentry Event Link: https://kiva.sentry.io/issues/4413252219/events/726c65f507684f43b748e913d4793518/
+				// This url is unreachable: https://pagead2.googlesyndication.com/pagead/buyside_topics/set/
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

src/main.js
```diff
@@ -81,6 +81,12 @@ export default function createApp({
 				// make sentry colleted event easy to compare to
 				const eventAsString = JSON.stringify(event);
 				// match specific 3rd party events for exclusion
+				// Skip sending failed to fetch error caused by unhandled promise rejection in google ads
+				// Sentry Event Link: https://kiva.sentry.io/issues/4413252219/events/726c65f507684f43b748e913d4793518/
+				// This url is unreachable: https://pagead2.googlesyndication.com/pagead/buyside_topics/set/
+				if (eventAsString.indexOf('Failed to fetch') !== -1) {
```
emuvente, line 87: This will also filter out the failed to fetch error messages we sometimes see from Apollo. We haven't really done anything about those errors before, but it seems a little concerning to remove our signal about that entirely.

mcstover: I had some similar concerns but didn't know we'd seen this much before. Added a secondary string check that would focus this exclusion to the target error.

Note: I used `pagead` here b/c I've found instances of that as well as `pagead2`, thus we should be able to capture both.

emuvente (approved):
Looks good!

src/main.js
emuvente: Excellent, thank you!

PR #4966, opened 2023-09-15
fix: ensure icons are compressed and skip font compression MARS-288

emuvente:
Moving the compression to after the svg plugin in client base so that it runs on the icons file as well, and runs a little faster b/c it's not compressing the files for the server build. I also found that the font files were *larger* after compression, so I'm excluding those.


dyersituations (approved):

mcstover (approved):
Thanks for making all this better!

PR #4971, opened 2023-09-19
feat: support inviter param and recruitment on team process page

eddieferrer:
We can't generate an recruitmentId if the user is not logged in, so this page needs to also support the inviter query param. 

This allows us to go from un authenticated page > auth0 > team process page (with inviter param) 
Then we can use the inviter param to get the inviter id and pass that to the join team mutation as the recruitment id.

### DIFF
--- a/src/pages/ProcessJoinTeam.vue
+++ b/src/pages/ProcessJoinTeam.vue
@@ -162,8 +162,9 @@ export default {
 					mutation: joinTeam,
 					variables: {
 						team_id: this.teamId,
-						team_recruitment_id: this.recruitmentIdMutationVariable,
-						promo_id: this.promoId
+						...(this.recruitmentIdMutationVariable
+							&& { team_recruitment_id: this.recruitmentIdMutationVariable }),
+						...(this.promoId && { promo_id: this.promoId }),
 					}
 				});
 				// get updated team membership status to determine if joined or pending
--- a/src/pages/ProcessJoinTeam.vue
+++ b/src/pages/ProcessJoinTeam.vue
@@ -1,7 +1,7 @@
 <template>
 	<www-page>
 		<kv-page-container>
-			<kv-grid class="tw-grid-cols-12 tw-pt-3 md:tw-pt-5 lg:tw-pt-7 tw-mb-4 md:tw-mb-6">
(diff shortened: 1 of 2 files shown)


src/pages/ProcessJoinTeam.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<www-page>
 		<kv-page-container>
-			<kv-grid class="tw-grid-cols-12 tw-pt-3 md:tw-pt-5 lg:tw-pt-7 tw-mb-4 md:tw-mb-6">
+			<kv-grid class="tw-grid-cols-12 tw-pt-5 md:tw-pt-7 lg:tw-pt-9 tw-mb-4 md:tw-mb-6">
```
eddieferrer, line 4: Add a little space so the toast doesnt cover the page message

src/pages/ProcessJoinTeam.vue
```diff
@@ -11,7 +11,7 @@
 							<kv-loading-spinner class="tw-mt-2" />
 						</div>
 					</div>
-					<div v-else>
+					<div v-else class="tw-text-center">
```
eddieferrer, line 14: styling tweaks

src/pages/ProcessJoinTeam.vue
```diff
@@ -93,24 +103,28 @@ export default {
 		preFetchVariables({ route }) {
 			return {
 				teamPublicId: route.query.teamPublicId,
+				publicId: route.query.inviter ?? '',
```
eddieferrer, line 106: If the inviter query param is not present, this just sends an empty string as the param, the api just returns lender null and the page functions fine

src/pages/ProcessJoinTeam.vue
```diff
@@ -138,7 +162,7 @@ export default {
 					mutation: joinTeam,
 					variables: {
 						team_id: this.teamId,
-						team_recruitment_id: this.teamRecruitmentId,
+						team_recruitment_id: this.recruitmentIdMutationVariable,
```
dyersituations, line 165: Do we need to check if `recruitmentIdMutationVariable` ends up being `null` before sending of this mutation, or are we OK with it being `null`?

eddieferrer: It works, but no harm in making both of these conditional

dyersituations (approved):

emuvente (approved):

PR #4976, opened 2023-09-20
feat: add cps visibleUrls approach

christian14b:
Avoid double disclaimers in ui pages

Screen Shot 2023-09-20 at 15 18 58

### DIFF
--- a/src/components/WwwFrame/DisclaimersContentful.vue
+++ b/src/components/WwwFrame/DisclaimersContentful.vue
@@ -99,7 +99,7 @@ export default {
 					// check for visibility based on current route and hiddenUrls field
 					const hiddenUrls = globalBannerDenyList.concat(promoContent?.fields?.hiddenUrls ?? []);
 					const visibleUrls = promoContent?.fields?.visibleUrls ?? [];
-					if (!isExcludedUrl(hiddenUrls, visibleUrls, this.$route.path)) return false;
+					if (isExcludedUrl(hiddenUrls, visibleUrls, this.$route.path)) return false;
 					if (promoContent.fields.active) {
 						return false;
 					}


src/components/WwwFrame/DisclaimersContentful.vue
```diff
@@ -59,7 +59,7 @@ export default {
 		preFetch: true,
 		result({ data }) {
 			// Hide ALL banners on these pages
-			if (isExcludedUrl(globalBannerDenyList, this.$route.path)) return false;
+			if (isExcludedUrl(globalBannerDenyList, [], this.$route.path)) return false;
```
emuvente, line 62: To match the cps behavior, this line should be dropped and the `globalBannerDenyList` should be concatenated with the hiddenUrls field in the calls to `isExcludedUrl` below, like it is here https://github.com/kiva/cms-page-server/blob/50cc7ccab8a353516eccb795f9492327bcc101a0/composables/usePromoBanner.ts#L53

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -58,7 +58,7 @@ export default {
 		preFetch: true,
 		result({ data }) {
 			// Hide ALL banners on these pages
-			if (isExcludedUrl(globalBannerDenyList, this.$route.path)) return false;
+			if (isExcludedUrl(globalBannerDenyList, [], this.$route.path)) return false;
```
emuvente, line 61: Same as my comment above

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -126,7 +127,7 @@ export default {
 	computed: {
 		showAppeal() {
 			// Check if Appeal Banner is active and the user is not on a denied page URL
-			if (this.appealEnabled && !isExcludedUrl(globalBannerDenyList, this.$route.path)) return true;
+			if (this.appealEnabled && !isExcludedUrl(globalBannerDenyList, [], this.$route.path)) return true;
```
emuvente, line 130: isExcludedUrl will be called already in the prefetch result, so `showAppeal` can just be `return this.appealEnabled;` or the usages of it could be replaced with appealEnabled.

src/util/urlUtils.js
```diff
@@ -16,27 +16,32 @@ export const globalBannerDenyList = [
 	'/process-instant-lending/*',
 ];
 
+function matchRoute(route, pattern) {
```
dyersituations, line 19: It looks like this file already has a unit testing file, consider adding new unit tests for these two methods.

emuvente (approved):
Looks great!

PR #4981, opened 2023-09-25
Core 1513 thanks page ftds

roger-in-kiva:
- Message for FTDs added to thanks page
- Settings added to kiva admin to manage this feature

ftds_desktop_cut

ftds_mobile_cut

### DIFF
--- a/src/components/Thanks/ShareStepper.vue
+++ b/src/components/Thanks/ShareStepper.vue
@@ -1,6 +1,6 @@
 <template>
 	<div class="tw-my-2 tw-px-2">
-		<div style="width: 136px;" class="tw-mx-auto tw-px-1">
+		<div v-if="!isFirstLoan" style="width: 136px;" class="tw-mx-auto tw-px-1">
 			<div class="tw-flex tw-items-center">
 				<template v-for="step in steps">
 					<ShareStepperStep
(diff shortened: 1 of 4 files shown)


src/graphql/query/thanksPage.graphql
```diff
@@ -132,6 +132,14 @@ query checkoutReceipt($checkoutId: Int!, $visitorId: String) {
 			key
 			value
 		}
+		ftd_message_enable: uiConfigSetting(key: "thanks_page_ftd_message_enable") {
```
dyersituations, line 135: Is there a reason why were using a config instead of an experiment?

roger-in-kiva: This is not a test, but we want the ability to turn this off

src/components/Thanks/ShareStepper.vue
```diff
@@ -17,6 +17,10 @@
 				<span v-if="commentsMode">
 					<span class="data-hj-suppress">{{ lenderName }}</span>, go the extra mile
 				</span>
+				<span v-else-if="isFirstLoan">
+					<!-- eslint-disable-next-line max-len -->
+					<span class="data-hj-suppress">{{ lenderName }}</span>, You’re support means the world to <span class="data-hj-suppress">{{ loanName }}</span>. Let us help you change another life.
```
dyersituations, line 22: Does `loanName` need to be suppressed?

roger-in-kiva: Commonly this is the borrower name, should be?

dyersituations: It does look like we've hidden the loan name elswhere.

dyersituations (approved):

PR #4987, opened 2023-09-26
feat: add loanId filter

christian14b:
Creating new route to fetch loan data
Sample: `{{host}}/live-loan/lid/2639154/img2`

image

### DIFF
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -185,8 +185,22 @@ module.exports = function liveLoanRouter(cache) {
 		});
 	});
 
+	// Loan Id URL Router
+	router.use('/lid/:id([0-9]{0,})/url', async (req, res) => {
+		await tracer.trace('live-loan.loanid.redirectToUrl', { resource: req.path }, async () => {
+			await redirectToUrl('loanid', cache, req, res);
+		});
+	});
+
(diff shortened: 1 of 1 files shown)


server/live-loan-router.js
```diff
@@ -185,6 +185,13 @@ module.exports = function liveLoanRouter(cache) {
 		});
 	});
 
+	// Loan Id IMG Router (Kiva Classic)
+	router.use('/lid/:id([a-zA-Z0-9.%,_-]{0,})/img2', async (req, res) => {
```
emuvente, line 189: The regex for the id parameter should only allow numbers, `[0-9]{0,}`

server/live-loan-router.js
```diff
@@ -185,6 +185,13 @@ module.exports = function liveLoanRouter(cache) {
...
+		await tracer.trace('live-loan.loanid.serveImg', { resource: req.path }, async () => {
+			await serveImg('loanid', 'classic', cache, req, res);
+		});
+	});
```
emuvente, line 193: This is looking great! Please add similar `lid` route handlers for /img and /url

emuvente (commented):
It's looking

emuvente (approved):
Looks great! Thanks for the updates!

PR #4992, opened 2023-09-29
feat: remove lend_urgency exp code

eddieferrer:
🧹 

If we need a countdown, we can use the one in kv elements


mcstover (approved):

package.json
```diff
@@ -43,7 +43,6 @@
 		"@babel/eslint-parser": "^7.18.2",
 		"@babel/polyfill": "^7.10.4",
 		"@babel/runtime": "^7.12.5",
-		"@chenfengyuan/vue-countdown": "^1.1.5",
```
dyersituations, line 46: @eddieferrer We have a custom countdown component in the new loan card (since this library wasn't compatible with both versions of Vue at the same time), do we need to remove the countdown from that component as well?

PR #5002, opened 2023-10-02
feat: teams page additional features

eddieferrer:
Some additional features for the /teams page

* Leaderboards now change to filter on the category 
* Added info line indicating this. 
* Added placeholder loader while leaderboards load. 
* Added no teams to show when leaderboard query is empty
* Fixed some issues with query params not being removed when the "all" value is selected
* Added utility variable of team category values and friendly labels

Screenshot 2023-10-02 at 5 09 42 PM

### DIFF
--- a/src/pages/LendingTeams/TeamListing.vue
+++ b/src/pages/LendingTeams/TeamListing.vue
@@ -1,14 +1,12 @@
 <template>
 	<div>
-		<div
-			class="tw-flex tw-justify-between tw-items-baseline tw-mb-3"
-		>
-			<h3>
+		<div>
+			<h3 class="tw-mb-0">
 				Team Listing
 			</h3>
-			<p class="tw-text-small" v-if="!loading">
+			<p class="tw-text-small tw-mb-2">
 				<!-- eslint-disable-next-line max-len -->
-				{{ numeral(totalCount).format('0,0') }} lending teams in {{ teamCategory ? teamCategory : 'all categories' }}
+				<span v-if="!loading">{{ numeral(totalCount).format('0,0') }}</span> lending teams in {{ teamCategory ? teamCategoryFriendlyName(teamCategory) : 'all categories' }}
 			</p>
 		</div>
 		<team-search-bar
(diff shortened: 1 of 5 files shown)


src/pages/LendingTeams/TeamListing.vue
```diff
@@ -450,10 +398,21 @@ export default {
 				pushToRouter('teamCategory');
 				return;
 			}
+			if (this.teamCategory === '' && this.teamCategory !== this.$route.query?.teamCategory) {
+				const query = { ...this.$route.query };
+				delete query?.teamCategory;
+				this.$router.replace({ query });
+			}
+			console.log(this.teamOption, this.$route.query?.teamOption, this.teamOption === '');
```
dyersituations, line 406: Remove console.log?

src/pages/LendingTeams/TeamListing.vue
```diff
@@ -492,6 +451,7 @@ export default {
 		userIsTeamMember(teamId) {
 			return this.myTeams.some(team => team.id === teamId);
 		},
+		teamCategoryFriendlyName
```
dyersituations, line 454: Looks like an empty method or does this actually work for imported method? I thought we generally added those methods to the data object for the component.

eddieferrer: Its imported from the utils file

src/util/teamsUtil.js
```diff
@@ -1,6 +1,85 @@
...
+		label: 'Youth Groups',
+	},
+	{
+		value: 'Other',
+		label: 'Other',
+	}
+];
+
+export function teamCategoryFriendlyName(categoryValue) {
+	return teamCategories.find(category => category.value === categoryValue)?.label;
```
dyersituations, line 80: Do we want to make the comparison case insensitive?

src/util/teamsUtil.js
```diff
@@ -24,10 +103,13 @@ export async function fetchTeams(apollo, sortOption, category, membershipType, q
 	}
 }
 
-export async function fetchLeaderboard(apollo) {
+export async function fetchLeaderboard(apollo, category) {
```
dyersituations, line 106: Looks like this file needs unit tests in general. Doesn't need to happen now, but it would be a good addition.

src/pages/LendingTeams/TeamListing.vue
dyersituations: Interesting, I didn't realize this worked, since I've used the data object. Good to know!

src/util/teamsUtil.js
eddieferrer: I could go either way. The Enum list in the API is cased, so this matches up with that.

PR #5012, opened 2023-10-04
feat: prevent loading state while waiting for the post

christian14b:
In the past implementation, the module was still appearing while fetching the post and then disappeared if no posts were found. 

With this implementation the page will decide to show or not the whole education module, instead of the module.

The downside of this is that we will have to make an api call even if the module is not visible.

### DIFF
--- a/src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
+++ b/src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
@@ -11,7 +11,7 @@
 					<lending-insights />
 					<recent-loans-list />
 					<your-donations />
-					<education-module v-if="showEdModule" @hide-module="hideModule" />
+					<education-module v-if="post" :post="post" />
 					<kiva-credit-stats />
 					<account-updates />
 					<your-teams />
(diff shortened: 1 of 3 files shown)


dyersituations:
@christian14b Both this PR and https://github.com/kiva/cms-page-server/pull/693 seem related, and the recent changes seem good. Were you able to test locally? If so, I can approve, I just know we had issues with testing the previous iteration of the promo spot.

dyersituations (approved):

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -59,13 +60,31 @@ export default {
...
+					}`,
+				variables: {
+					customFields: 'metadata.tags.sys.id[in]=impact-page|order=-fields.originalPublishDate'
+				},
+			}).then(({ data }) => {
+				this.post = data?.contentful?.blogPosts?.items?.[0]?.fields ?? null;
+			});
 		}
+	},
+	created() {
```
emuvente, line 86: Let's use `mounted` instead so that this isn't started on the server-side. The server completes the request as soon as `created` finishes, so anything async started during that is ignored.

PR #5021, opened 2023-10-06
fix: [CP-548] Fixing API audience URL

kathrynlovett:
Screen Shot 2023-10-06 at 10 06 56 AM

### DIFF
--- a/config/dynamic.js
+++ b/config/dynamic.js
@@ -37,7 +37,7 @@ module.exports = {
 			},
 			enable: true,
 			checkFakeAuth,
-			apiAudience: `https://api.${formattedUrlEnv}kivaws.org/graphql`,
+			apiAudience: `https://api.${baseUrl}/graphql`,
 			mfaAudience: `https://kiva-${env}.auth0.com/mfa/`,
 			browserClientID,
 			serverClientID,


dyersituations (approved):

mcstover (approved):

kathrynlovett:
@dyersituations @mcstover I think our staging environment is relying on the versioned images for the `ui` project. How can I get these changes in a new versioned release? Do y'all create versions manually or is there another process for that?

mcstover:
@kathrynlovett Do you mean the new staging environment? The prior stage env just received the same tag deployed to dev. We then manually deployed the tag to QA for the testing cycle. The full release is currently created for prod only.

I think what you're talking about is moving to this new `-rc` pre-release process which we don't have an image build for yet in either UI or CPS.

kathrynlovett:
@mcstover no, the `ui` isn't setup for our new `-rc` process yet. It's just looking for the latest version for the project and I wasn't seeing any of the versions with the changes from this PR: https://github.com/kiva/ui/releases
Or are you saying that we only create a versino when deploying to prod?

mcstover:
Well we create a version "tag" on each merge...https://github.com/kiva/ui/tags

mcstover:
Release however, are only created when deploying to prod: https://github.com/kiva/ui/releases

kathrynlovett:
Ah, got it. Thank you! Looks like we do have the version tag we need, and I think we're pointing at it. I guess the only remaining mystery is why I'm still seeing us use the old audience. I'll keep digging. Thank you!

PR #5022, opened 2023-10-10
feat: introspection queries passed to a new query file

roger-in-kiva:
- Introspection queries passed to a new query file for filtering options.
- Genders bug fixed in quick filters when resetting selected filters. We are currently showing no option selected in gender dropdown after resetting the filters.
- Updates tested in `/lend/filter`, `/lend-by-category` and `/lend-by-category/*` pages

Screenshot 2023-10-10 at 12 27 47

### DIFF
--- a/src/util/loanSearch/dataUtils.js
+++ b/src/util/loanSearch/dataUtils.js
@@ -1,4 +1,5 @@
 import loanFacetsQuery from '@/graphql/query/loanFacetsQuery.graphql';
+import loanEnumsQuery from '@/graphql/query/loanEnumsQuery.graphql';
 import {
 	fetchFacets,
 	fetchLoans,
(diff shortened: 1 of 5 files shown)


src/util/loanSearch/dataUtils.js
```diff
@@ -61,13 +62,14 @@ export async function runLoansQuery(apollo, loanSearchState, origin) {
 export async function fetchLoanFacets(apollo) {
 	try {
 		const result = await apollo.query({ query: loanFacetsQuery, fetchPolicy: 'network-only' });
+		const resultEnums = await apollo.query({ query: loanEnumsQuery, fetchPolicy: 'network-only' });
```
dyersituations, line 65: These can be sent off at the same time with `Promise.all`

roger-in-kiva: Fixed, thanks!

dyersituations (approved):

PR #5023, opened 2023-10-10
feat: productize value prop experiment and removing old checkout modu…

christian14b:
…le experiment.

image

image


src/components/Checkout/DonationItem.vue
```diff
@@ -1,184 +1,13 @@
...
-						>
-							<kv-material-icon
-								class="remove-x tw-text-tertiary"
-								name="small-x"
-								:from-sprite="true"
-								title="Remove donation"
-							/>
-						</button>
-					</div>
-					<donate-repayments
```
mcstover, line 167: Why is this being completely removed? What other pathway do we expect lenders to use to activate this condition on their loans?

mcstover: Ok, I do see this in one of the code blocks below so looks like it will still remain. Sorry for the confusion.

mcstover (approved):
Nice Cleanup

PR #5025, opened 2023-10-11
feat: update default percentage to 0.18 and hide covid category to ne…

christian14b:
…w users in monthly good setup page

### DIFF
--- a/src/components/Forms/MonthlyGoodUpdateForm.vue
+++ b/src/components/Forms/MonthlyGoodUpdateForm.vue
@@ -273,6 +273,10 @@ export default {
 				{ label: 'Preserve existing settings', value: '', shortName: '' }
 			);
 		}
+		// Removing Covid 19 category option if it is not active for the user
+		if (this.category?.marketingName !== 'COVID-19') {
+			this.lendingCategories = this.lendingCategories.filter(category => category.marketingName !== 'COVID-19');
+		}
 		/** After initial value is loaded, setup watch to make form dirty on value changes
 		 * and emit values on value changes. Setting this in mounted prevents the form
 		 * from being dirty on initial load


src/pages/MonthlyGood/MonthlyGoodSetupPage.vue
```diff
@@ -540,6 +542,15 @@ export default {
 
 		// Sanitize and set initial form values.
 		// Initial group from prop
+		const hasLoggedInResult = this.apollo.readQuery({
+			query: hasEverLoggedInQuery
+		});
+		if (!hasLoggedInResult?.hasEverLoggedIn) {
```
emuvente, line 548: There shouldn't be an if statement around this because the COVID-19 category should be removed as an option for everybody.

src/plugins/loan-group-categories.js
```diff
@@ -58,15 +58,7 @@ export default {
...
-				},
-				{
-					value: 'disaster_relief_covid',
-					label: 'COVID-19 Coronavirus',
-					shortName: 'the global COVID—19 response',
-					marketingName: 'COVID-19',
-					marketingOrder: 7,
-					expLoansIds: [2319451, 2320833, 2320176]
-				},
+				}
```
mcstover, line 61: How will this removal affect the subscription/mg edit functionality under settings/subscriptions?

mcstover: Prob a mute point if we don't have any/many subs using this category. Just wanted to make sure it doesn't break that interface if it's the currently set category on an account...

emuvente: Yeah, curious about that as well. I think this is a good way to go about it, but it might not allow people who are already in that category to change their other settings without selecting one of the other categories. I think just removing the option from the setup page (like how you had it earlier) would be an ok solution if there are too many issues doing it this way.

dyersituations (approved):

mcstover (approved):
It may be  useful to add a "guard" of some sort in the subscriptions management to prevent saving the "Covid-19" category in that context. Maybe filter the option out if it's not currently "active" on the logged in user.

mcstover (approved):

PR #5036, opened 2023-10-19
feat: default sort option added to quick filters for almost funded ro…

roger-in-kiva:
- Ending soon option added as default to quick filters section for almost funded row test

### DIFF
--- a/src/components/LoanFinding/QuickFiltersSection.vue
+++ b/src/components/LoanFinding/QuickFiltersSection.vue
@@ -16,7 +16,7 @@
 				:targeted-loan-channel-url="targetedLoanChannelURL"
 				:with-categories="true"
 				:enable-qf-mobile="enableQfMobile"
-				default-sort="amountLeft"
+				:default-sort="defaultSort"
 				tracking-category="lending-home"
 				@update-filters="updateQuickFilters"
 				@reset-filters="resetFilters"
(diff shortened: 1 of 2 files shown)


src/components/LoanFinding/QuickFiltersSection.vue
```diff
@@ -110,7 +114,7 @@ export default {
 			loanSearchState: {
 				pageOffset: 0,
 				pageLimit: 6,
-				sortBy: 'amountLeft'
+				sortBy: this.enableAlmostFundedRow ? 'expiringSoon' : 'amountLeft'
```
dyersituations, line 117: I forget, can the `computed` be used here as well?

mcstover: I think it only gets set on initialization but would not be reactive for subsequent changes...

dyersituations: I think that would be the same as this line, but I forget if `computed` is accessible yet in `data`.

roger-in-kiva: Unfortunately not, this is the main reason:
`The computed property cannot be used in `data()` because it is before initialization.`

dyersituations: Okay then, I was wondering, thanks for checking!

dyersituations (approved):

PR #5047, opened 2023-10-26
feat: integrate double the donation plugin and wrap it behind feature…

christian14b:
… flag on only donation thanks page.

This is an experiment with double the donation so that's the reason behind the feature flag.

### DIFF
--- a/src/components/Thanks/ThanksPageDonationOnly.vue
+++ b/src/components/Thanks/ThanksPageDonationOnly.vue
@@ -1,24 +1,26 @@
 <template>
 	<div>
 		<div v-if="doubleDonationEnabled" class="tw-py-5 tw-bg-marigold-1">
-			<kv-grid class="tw-grid-cols-12">
-				<div class="tw-col-span-12 lg:tw-col-span-8 lg:tw-col-start-3 tw-pt-2 tw-mb-4 hide-for-print">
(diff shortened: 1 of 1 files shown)


src/components/Thanks/ThanksPageDonationOnly.vue
```diff
@@ -1,11 +1,29 @@
...
+				</div>
+			</kv-grid>
+		</div>
+		<kv-page-container>
+			<kv-grid class="tw-grid-cols-12 tw-my-8">
+				<div class="tw-col-span-12 lg:tw-col-span-8 lg:tw-col-start-3 tw-pt-2 tw-mb-4 hide-for-print">
+					<div v-if="doubleDonationEnabled" class="tw-mb-8">
+						<div id="dd-container">
+							<!--eslint-disable-next-line max-len-->
+							<a href="https://doublethedonation.com/matching-grant-resources/matching-gift-basics/">Matching Gift</a> and <a href="https://doublethedonation.com/matching-grant-resources/volunteer-grant-basics/">Volunteer Grant</a> information provided by <br><a href="https://doublethedonation.com"><img alt="Powered by Double the Donation" src="https://doublethedonation.com/api/img/powered-by.png"></a>
```
mcstover, line 18: Would it be ok to add `target="_blank"` to these external links?

emuvente (approved):

PR #5054, opened 2023-10-31
feat: teams page tweaks

roger-in-kiva:
- Change the amounts format  to show the full number in the leaderboards, seeing the exact number is more valuable.
- Add some helper text to the first two tabs in the leaderboard member counts like this: “Only members who have lent at least once are counted”
- Hide the text "We loan because:" if the field is empty

Screenshot 2023-10-31 at 16 36 41
Screenshot 2023-10-31 at 16 36 51
Screenshot 2023-10-31 at 16 37 01

### DIFF
--- a/src/pages/LendingTeams/TeamListing.vue
+++ b/src/pages/LendingTeams/TeamListing.vue
@@ -161,6 +161,7 @@
 			>
 				<p
 					class="tw-text-small tw-text-primary tw-flex-1"
+					v-if="shortLoanBecause(team.loanBecause)"
 				>
 					We loan because: {{ shortLoanBecause(team.loanBecause) }}
 				</p>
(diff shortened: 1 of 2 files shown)


src/pages/LendingTeams/TeamListing.vue
```diff
@@ -161,6 +161,7 @@
 			>
 				<p
 					class="tw-text-small tw-text-primary tw-flex-1"
+					v-if="shortLoanBecause(team.loanBecause)"
```
mcstover, line 164: I think you should just check `team.loanBecause`here instead of pushing it through the truncation method.

mcstover (approved):

dyersituations (approved):

eddieferrer (approved):
Looks good, thanks for making these changes!

PR #5056, opened 2023-11-01
fix: turn off apollo batching for ui in all evns including prod

mcstover:
Preparation for changes to api infra

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -2,7 +2,7 @@ var path = require('path')
 
 module.exports = {
 	app: {
-		apolloBatching: true,
+		apolloBatching: false,
 		host: 'www.kiva.org',
 		transport: 'https',
 		publicPath: 'https://www-kiva-org.freetls.fastly.net/ui/',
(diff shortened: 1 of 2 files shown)


dyersituations (approved):

kathrynlovett:
Does this need to be updated for the EKS side as well? It uses the dynamic config: https://github.com/kiva/ui/blob/main/config/dynamic.js#L3

mcstover:
> Does this need to be updated for the EKS side as well? It uses the dynamic config: https://github.com/kiva/ui/blob/main/config/dynamic.js#L3

i don't think dynamic config needs to change. We are passing in `false` via flux for the envs that use that and batching is already off there. I'll double check the prod config for ui in flux though!

kathrynlovett (approved):
LGTM! 🚀

JoeTravisKiva (approved):

mcstover:
@emuvente I think we need to go ahead with this update, please review.

emuvente (approved):
Let's do it!

PR #5059, opened 2023-11-06
fix: change query param on new teams page

eddieferrer:
Changes the query param for teams categories from `teamCategory=something` to `category=something` on the `/teams` page. 

This is needed because there are some links in the legacy teams pages that are linking to `/teams?category=something` so instead of fixing all the legacy link, lets just make this query param the same.

### DIFF
--- a/src/pages/LendingTeams/TeamLeaderboards.vue
+++ b/src/pages/LendingTeams/TeamLeaderboards.vue
@@ -41,10 +41,10 @@ export default {
 		};
 	},
 	watch: {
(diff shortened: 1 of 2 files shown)


mcstover (approved):
Guessing we haven't rolled out usage of the newer incorrect param key...

dyersituations (approved):

eddieferrer:
> Guessing we haven't rolled out usage of the newer incorrect param key...

Nope, just internally on the page, set from the dropdown

PR #5060, opened 2023-11-06
fix: reduce checkout query complexity

mcstover:
- Move the new ftd settings to a different query
- Add Sentry log for missing totals data. One key component of critical basket data.
- - Looks like our current apollo prefetch handler is not returning the `errors` out of the collected set of queries so a check at that level will need additional work.

### DIFF
--- a/config/dev-local.js
+++ b/config/dev-local.js
@@ -7,7 +7,7 @@ module.exports = merge(base, devVm, {
 		apolloBatching: false,
 		host: 'localhost',
 		publicPath: '/',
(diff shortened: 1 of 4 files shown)


roger-in-kiva (approved):
Looks good, thanks!

dyersituations (approved):

roger-in-kiva:
Should be nice to find a way to replicate the query complexity error in dev to prevent it in prod.

emuvente (approved):

JoeTravisKiva (approved):
Nice! Seems like we finally found a benefit of using Apollo Router.

PR #5061, opened 2023-11-07
feat: increase edit donation amount from modal

christian14b:
Add a feature flag to increase default tip rates in edit modal. This change will be enabled with the seasonal 20% default donation rate.


emuvente (approved):
👍 Looks great!

PR #5062, opened 2023-11-08
feat: run release and release build in same workflow

emuvente:


.github/workflows/semantic-release.yaml
```diff
@@ -1,32 +1,82 @@
...
+    env:
+      GITHUB_PAT: "${{ secrets.KIVA_ROBOT_GITHUB_PAT }}"
+      AWS_REGION: "us-west-2"
+      AWS_ACCESS_KEY_ID: "${{ secrets.AWS_ACCESS_KEY_ID }}"
+      AWS_SECRET_ACCESS_KEY: "${{ secrets.AWS_SECRET_ACCESS_KEY }}"
+      SLACK_BOT_TOKEN: "${{ secrets.SLACK_TOKEN }}"
+      SLACK_CHANNEL: "eng-build-failures"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
kathrynlovett, line 58: [nit] I see above we're using `actions/checkout@v4` - is there any reason we wouldn't want to use the same version here? 🤔

.github/workflows/semantic-release.yaml
```diff
@@ -1,32 +1,82 @@
...
+      - name: Run semantic-release
+        id: release
         uses: ./.github/actions/semantic-release
         env:
           GITHUB_PAT: ${{ secrets.KIVA_ROBOT_GITHUB_PAT }}
         with:
           config: "node"
+
+  # Build and push the image to ECR if the release was published
+  build-image:
```
mcstover, line 45: Can most of this code move back into the gh-actions repo at some point for easier reuse?

mcstover (approved):

.github/workflows/semantic-release.yaml
```diff
@@ -1,32 +1,82 @@
...
+          path: .docker
+          ref: main
+          patterns: |
+              resources/org/kiva/marketplaceWebUiCi/ui
+      - name: move files
+        run: |
+          mv .docker/resources/org/kiva/marketplaceWebUiCi/ui/Dockerfile .
+      - name: docker-build
+        uses: ./.github/actions/docker-build-push
+        if: github.event_name == 'pull_request'
```
kathrynlovett, line 79: When someone triggers a manual build, will the Github event name be `pull_request`? Or was this moved over from elsewhere and needs to be updated, perhaps?

kathrynlovett (approved):
This all looks great to me! Left a couple questions but nothing blocking. Do we want to update the [release-build](https://github.com/kiva/ui/blob/main/.github/workflows/release-build.yaml) action at all to prevent builds on tag pushes, now that we're building in the semantic release action?

.github/workflows/semantic-release.yaml
emuvente: Yeah, gha allows for reusable/shareable workflows. I think it would be great if we could eventually setup these workflows in the github actions repo and then reference them in all the app repos.

.github/workflows/semantic-release.yaml
emuvente: v3 is what is used in the docker build publish workflow. I wasn't sure if there was a reason we were using v3 instead of v4 there, so I didn't change it. I presume it would be better to be using v4 in all cases though.

.github/workflows/semantic-release.yaml
emuvente: Good catch! That needs to be updated

.github/workflows/semantic-release.yaml
kathrynlovett: That's what I would guess, too, though who knows! Guess we'll find out 😁

kathrynlovett (approved):
🚀

emuvente:
> Do we want to update the [release-build](https://github.com/kiva/ui/blob/main/.github/workflows/release-build.yaml) action at all to prevent builds on tag pushes, now that we're building in the semantic release action?

@kathrynlovett Sorry I missed this at first. Yes, we should update that too, thanks for catching that! I just pushed up a commit to do that.

kathrynlovett (approved):
Thanks for updating the `release-build` action! 🙏🏻

PR #5067, opened 2023-11-10
fix: allow passing in full api url

mcstover:
Activates the API_HOSTNAME env var in our dynamic config when passed.


kathrynlovett (approved):
🚀

PR #5075, opened 2023-11-17
feat: loan comment modal added to portfolio page

roger-in-kiva:
- Loan comment modal added to portfolio page
- Going to borrower profile after saving the comment
- Showing toast confirmation after saving the comment

Screenshot 2023-11-17 at 9 08 08

### DIFF
--- a/src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
+++ b/src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
@@ -1,0 +1,128 @@
+<template>
+	<kv-lightbox
+		:visible="loan.visible"
+		:title="modalTitle"
+		@lightbox-closed="closeModal"
+	>
+		<div class="tw-text-center">
+			<div class="tw-relative">
+				<textarea
+					class="tw-w-full tw-border tw-border-secondary tw-rounded-sm tw-h-7 tw-p-1 comment-input"
+					v-model="userComment"
+					:placeholder="commentPlaceholder"
+				>
+			</textarea>
+				<kv-material-icon
+					class="tw-w-2.5 tw-h-2.5 tw-absolute tw-bottom-2 tw-right-0"
+					:icon="mdiPencilOutline"
+				/>
+			</div>
+			<small
+				class="tw-text-left tw-block"
+			>Please follow our
+				<a
+					href="/kiva-community-guidelines" target="_blank"
+					class="tw-text-action hover:tw-text-action-highlight"
+				>
+					community guidelines
+				</a>
+			</small>
+		</div>
+		<kv-button
+			class="tw-mt-2"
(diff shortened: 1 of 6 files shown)


src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
```diff
@@ -0,0 +1,128 @@
...
+					this.closeModal();
+					this.$kvTrackEvent('portfolio', 'click', 'Leave a loan comment', this.loan.borrowerName, this.loan.loanId); // eslint-disable-line max-len
+					this.$showTipMsg(`Thank you for helping ${this.loan.borrowerName}!`, 'confirmation', true);
+					this.$router.push(`/lend/${this.loan.loanId}`);
+				} else {
+					throw new Error('Comment not added');
+				}
+			}).catch(e => {
+				logFormatter(e, 'error');
+				this.$showTipMsg('There was a problem commenting on this loan', 'error');
```
dyersituations, line 106: Since we don't close the modal on error, will the tip message work when shown with the open modal? Do we need to also close the modal on error?

src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+	<kv-lightbox
+		:visible="loan.visible"
+		:title="modalTitle"
+		@lightbox-closed="closeModal"
+	>
+		<div class="tw-text-center">
```
dyersituations, line 7: It looks like this is some duplicate code from the `CommentAsk` component. Would it make sense to abstract this out and use in both cases? I'm not sure how visually and/or functionally different these two components are.

src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
roger-in-kiva: Thanks for this catch!
A possible error here is b/c any API issue, so, I don't think closing the modal is a good UX.
I think we can add the error message below the `Save` button, adding a `try again` copy, in this way the user don't need to open de modal and write the comment again.

src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
roger-in-kiva: With the fix in the comment above, for the error message this is a little bit different, but we can abstract the submit method into a mixin to use it in both cases.

dyersituations: Probably isn't needed in this changeset, but it would probably be worth creating a new ticket to cover abstracting out this deplicate functionality.

dyersituations (approved):

PR #5078, opened 2023-11-22
feat: lend menu buttons test code removed

roger-in-kiva:
- lend menu buttons old test code removed

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -111,28 +111,26 @@
 							</a>
 						</div>
 
-						<template>
-							<!-- Lend -->
-							<router-link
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -114,39 +111,7 @@
...
-											:icon="mdiChevronDown"
-											:class="{'tw-rotate-180' : isLendMenuVisible}"
-										/>
-									</span>
-								</div>
-							</div>
-						</template>
-
-						<template v-else>
+						<template>
```
dyersituations, line 114: `template` can probably be removed, since it's always visible.

dyersituations (approved):

PR #5079, opened 2023-11-22
feat: homepage file cleaned up

roger-in-kiva:
- homepage file cleaned up

@mcstover should we remove support for New Yeah Share Campaign?


src/pages/Homepage/Homepage.vue
```diff
@@ -4,24 +4,11 @@
 
 <script>
 /* eslint-disable vue/multi-word-component-names */
-import { gql } from '@apollo/client';
-import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
-import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import { preFetchAll } from '@/util/apolloPreFetch';
 import numeral from 'numeral';
 
 const ContentfulPage = () => import('@/pages/ContentfulPage');
```
dyersituations, line 10: Can we remove this file entirely?

roger-in-kiva: This from a slack convo with Stover:

> It should be safe to clean up the old commented out route (leave the redirect in place for now) followed by cleaning up the Homepage.vue file and any homepage specific (non-contentful) components.
> We should keep the Contentful page intact for now

https://kivahq.slack.com/archives/CB7GRAFKK/p1700513842600929

mcstover: I don't think we should remove the Contentful page and related code just yet. But I think it's fine to remove the Homepage.vue file...

roger-in-kiva: Gotcha, thanks!

src/components/Homepage/MonthlyGoodExplained.vue
```diff
@@ -1,162 +0,0 @@
...
-				>
-					Learn more
-				</router-link>
-			</div>
-		</div>
-	</div>
-</template>
-
-<script>
-import MgStepOne from '@/assets/inline-svgs/hp-mg-1-signup.svg';
```
dyersituations, line 73: Remove these assets if they aren't used elsewhere (looks like they aren't?)

roger-in-kiva: Thanks!

dyersituations (approved):

PR #5092, opened 2023-12-05
feat: update events when user clicks No donation and enters 0 dollars …

christian14b:
…in the custom amount box


src/components/Checkout/DonationNudge/DonationNudgeLightbox.vue
```diff
@@ -187,17 +187,20 @@ export default {
 	methods: {
 		setDonationAndClose(amount, source) {
 			const zeroUpsellCookie = this.cookieStore.get('zero_upsell_visible') || true;
+			if (amount === 0) {
+				const expires = new Date();
+				expires.setTime(expires.getTime() + (24 * 60 * 60 * 1000));
```
dyersituations, line 192: While I know this is a fairly simple calculation, might be worth adding a comment to describe the expected expiration in simple words.

emuvente: You could also use date-fns for this:
```
import { add } from 'date-fns';

const expires = add(new Date(), { days: 1 });
```

dyersituations (approved):

PR #5095, opened 2023-12-08
feat: add prometheus metrics

hobbsh:
Adds Prometheus metrics via `express-prom-bundle` (wrapper for `prom-client`).

To test, spin up the dev server and hit `/metrics`.


mcstover (approved):
Cool, do you have a config pr for flux too? And markeplace-web-ui-ci once we verify, so we can add to current production deploys...

PR #5098, opened 2023-12-11
fix: styling issues with photos

christian14b:
image


mcstover:
I don't understand why we would obscure someones face instead of just scaling the image into the visible frame. Do we plan on activating a carousel there? Are we checking to see if folks try to scroll or interact with the image when it's not complete?

mcstover (approved):
Sounds like prep for a carousel, so ok.

PR #5100, opened 2023-12-13
fix: my teams link in portfolio

eddieferrer:


src/pages/Portfolio/ImpactDashboard/YourTeams.vue
```diff
@@ -69,7 +69,7 @@
 		<!-- all your teams link -->
 		<p v-if="totalTeams > 5" class="tw-text-center tw-mt-4">
 			<router-link
-				to="/blog/myloans"
```
eddieferrer, line 72: This link does not exist

mcstover: It should. are you saying it doesn't work?

mcstover: The page exists if you're logged in. Prob just needs to be an href.

mcstover (approved):
I think it's fine to remove from here, this link is correctly accessed from the Borrower Updates section of portfolio.

src/pages/Portfolio/ImpactDashboard/YourTeams.vue
eddieferrer: @mcstover Oh yeah I see, the link does exist in production. It seems to not resolve in dev: http://development.kiva.org/blog/myloans or in my tilt vm, but probably because of a missing entry in the tilt htaccess

eddieferrer:
> I think it's fine to remove from here, this link is correctly accessed from the Borrower Updates section of portfolio.

OK merging in this fix.

PR #5101, opened 2023-12-13
feat: ability to attribute a loan to a team from cookie value

eddieferrer:
We want the ability for a guest or logged in user to attribute a loan to a team from the challenge page. They may not be part of that team, and they may or may not be logged in at the time they add the loan to their basket

1. User goes to team challenge page, adds a loan to basket, cookie is created (Not in this PR, this will be on cps)
2. User eventually arrives at checkout. 
3. Read cookie 
4. Set the teamId on the loan reservation by calling the `updateLoanReservationTeam` mutation. 
5. If the user is not part of the team, append the team name and id to the drop down, user can manually change the attribution at this point. 
6. Checkout --> user will attribute loan to team and contribute to the challenge even if they are not part of the team. 

I have verified that you can in fact checkout with a loan attributed to a team you do not belong to.

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -38,10 +38,10 @@
 				/>
 				<team-attribution
 					class="tw-mb-1 tw-mt-0.5"
-					v-if="teams.length"
-					:teams="teams"
+					v-if="combinedTeams.length"
+					:teams="combinedTeams"
 					:loan-id="loan.id"
-					:team-id="loan.team ? loan.team.id : 0"
+					:team-id="loanTeamAttributionId"
 				/>
 				<loan-promo-credits
 					:applied-promo-credits="appliedPromoCredits"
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

src/components/Checkout/BasketItem.vue
```diff
@@ -193,5 +206,26 @@ export default {
...
+					// Is team not in the users list, append it
+					if (!this.combinedTeams.some(team => team.id === loan.teamId)) {
+						this.appendedTeams.push({
+							id: loan.teamId,
+							name: loan.teamName
+						});
+					}
+					this.forceTeamId = loan.teamId;
+				}
+			});
```
mcstover, line 227: Can we cleanup/delete the cookie at this point if it's been used?

roger-in-kiva (approved):

mcstover (approved):

PR #5103, opened 2023-12-14
fix: remove unnecessary package, reduce origins to only our domain

mcstover:
Verified the library update is working in Dev, webvitals emanate from Tracing and there are now some in the dashboard.

Minor tweaks
- increase sampling span to 25% which will cascade to prod in the next release, I'll also add these to flux and plan to turn dev up to 100%
- remove localhost and regex match on host
- remove unnecessary package!


mcstover:
I think we may be able to remove this in CPS too!

PR #5106, opened 2023-12-20
fix: add sentry trace rate config defaults

mcstover:
These couple with the recent updates to sentry to allow passing in an env var for the trace rate.


config/dev-local.js
```diff
@@ -23,7 +25,7 @@ module.exports = merge(base, devVm, {
 			enable: true,
 			browserCallbackUri: 'http://localhost:8888/process-browser-auth',
 			serverCallbackUri: 'http://localhost:8888/process-ssr-auth',
-			apiAudience: 'https://api.development.kiva.org/graphql',
+			apiAudience: 'https://gateway.development.kiva.org/graphql',
```
mcstover, line 28: This will help folks working locally

dyersituations (approved):

PR #5107, opened 2023-12-20
Pf 15 UI otel tracing

mcstover:
Baseline OTEL auto instrumentation for nodejs
- https://opentelemetry.io/docs/instrumentation/js/getting-started/nodejs/

Next steps will be to extend this with a tracer for the manually traced code sections.

Will be activated by: https://github.com/kiva/flux/pull/587


server/instrumentation.js
```diff
@@ -0,0 +1,21 @@
...
+const {
+	getNodeAutoInstrumentations,
+} = require('@opentelemetry/auto-instrumentations-node');
+const {
+	PeriodicExportingMetricReader,
+	ConsoleMetricExporter,
+} = require('@opentelemetry/sdk-metrics');
+
+const sdk = new NodeSDK({
+	traceExporter: new ConsoleSpanExporter(),
```
hobbsh, line 14: We want to use a `grpc` span exporter. `ConsoleSpanExporter` will not ship traces anywhere.

hobbsh: It should end up looking a lot like this: https://github.com/kiva/contentful-apollo-server/blob/master/tracing.js

mcstover: Added the [recommend otel exporters](https://opentelemetry.io/docs/instrumentation/js/exporters/#otlp-dependencies). They default to these endpoints:
`http://localhost:4318/v1/traces` + `http://localhost:4318/v1/metrics`.

Do we need to use a different endpoint?

mcstover: I see that `cas` has `OTEL_EXPORTER_OTLP_ENDPOINT` configured in the code but that doesn't appear to be set anywhere in flux, so it seems unused...

hobbsh: It's set when the value `opentelemetry.enabled: true` in the chart: https://github.com/kiva/deploy-helm/blob/main/charts/kiva-service/templates/configmap.yaml#L29

mcstover: Ok. I can enable the env var and have a default of the value you specified.

Any others I need to consider for this first pass? When I had the console logger in place I was able to see traces being output.

hobbsh: The endpoint var is good to use as a gate to enable tracing initialization as a whole, because without the endpoint set, traces won't ship anywhere. The rest of OTEL should be configurable via environment vars for the most part so it should be good. What nice is that the SDK across all languages respects the same environment vars: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

mcstover:
@hobbsh is this where it needs to be now? OTEL seems [to already be enabled for UI](https://github.com/kiva/flux/pull/587/files#diff-49283844424a64ddccd412225a7fd2c5e45edde70053849eef7e263b552958deR67)

eddieferrer (approved):

dyersituations (approved):

mcstover:
Going to merge this and activate only in dev

PR #5111, opened 2024-01-08
feat: use worker threads for vue app and sitemap rendering

emuvente:
This is a quick first pass at using worker threads for ui server requests.

I didn't find a way to share the vue bundle renderer between threads (functions can't be passed, as far as I can tell from the errors), so I just initialized one for each request. Not sure how that will affect performance, but I think threading will likely more than make up for it.

I setup a worker for live loans as well, but ran into issues with font registration. It sounds like node-canvas doesn't play nice with threads (according to https://github.com/Automattic/node-canvas/issues/1694) because of having to communicate between native C code and the nodejs process. I think it would be better to spin off live loans into their own service anyway, or maybe set it up as a lambda function.

### DIFF
--- a/src/util/KvAuth0.js
+++ b/src/util/KvAuth0.js
@@ -142,7 +142,7 @@ export default class KvAuth0 {
 	}
 
 	getFakeAuthCookieValue() {
-		if (!this.checkFakeAuth) return;
+		if (!this.fakeAuthAllowed()) return;
 
 		const cookieValue = this.cookieStore.get(FAKE_AUTH_NAME) ?? '';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -101,7 +101,8 @@
 				"web-vitals": "^2.1.4",
 				"webpack-merge": "^5.8.0",
 				"whatwg-fetch": "^3.6.2",
-				"winston": "^3.3.3"
+				"winston": "^3.3.3",
+				"workerpool": "^9.0.4"
 			},
 			"devDependencies": {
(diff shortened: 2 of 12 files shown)


src/util/KvAuth0.js
```diff
@@ -142,7 +142,7 @@ export default class KvAuth0 {
 	}
 
 	getFakeAuthCookieValue() {
-		if (!this.checkFakeAuth) return;
+		if (!this.fakeAuthAllowed()) return;
```
emuvente, line 145: Slipped in from a previous stash. This helps secure the fake auth process.

mcstover (approved):
Generally the setup looks good here.

Being that the vue middleware is also I/O intensive due to API data fetches in the renderToString method, I'm curious to see how much this is going to help. I suppose it will where we're rendering from mostly cached objects!

The docs also talk about using a pool of workers in practice and we don't have that hooked up yet. Is that something we need for this to be effective?

emuvente:
I haven't been able to test this at scale, but using a worker pool might solve the issue of needing to create the bundle renderer for each request. Each worker would still need to create a bundle renderer for themselves, but then they could respond to messages from the parent. I like that idea enough that I'll add a worker pool implementation to this PR.

emuvente:
I added workerpool and the request time was cut nearly in half after the first request (typically ~2.3 seconds w/o the pool and ~1.2 seconds with it)

dyersituations (approved):

mcstover (approved):

PR #5112, opened 2024-01-08
feat: clear challenge cookie after completed checkout

dyersituations:
https://kiva.atlassian.net/browse/ACK-895

- Created util for challenge cookie methods w/ tests
- Moved clearing cookie loans to checkout completion method
- Added script for test watch


emuvente (approved):

dyersituations:
> Nice! Should we update the cookie if the loan is removed from the basket?

@roger-in-kiva Good call, just pushed up that fix, thanks!

roger-in-kiva (approved):
👍

PR #5115, opened 2024-01-08
feat: goals service notify me in thanks page ack 864

christian14b:
New "Notify me" component within the standard thanks page that allows user to be added to an iterable list. To choose which team's goal to display (in the edge case a person belongs to two teams and make checkout to two different challenges) it'll select the team assigned to the last loan.

Mail icon will change in the future

Screenshot 2024-01-08 at 15 29 10

### DIFF
--- a/src/components/Thanks/NotifyMe.vue
+++ b/src/components/Thanks/NotifyMe.vue
@@ -130,11 +130,11 @@ export default {
 
 <style lang="postcss" scoped>
 .notify-button >>> span {
-  @apply tw-flex tw-items-center;
+	@apply tw-flex tw-items-center;
 }
 
 .container {
-  @apply tw-flex tw-flex-col tw-items-center tw-w-full tw-bg-white tw-rounded tw-px-1.5 tw-pt-1.5 tw-pb-2.5;
+	@apply tw-flex tw-flex-col tw-items-center tw-w-full tw-bg-white tw-rounded tw-px-1.5 tw-pt-1.5 tw-pb-2.5;
 }
 
 </style>
--- a/src/graphql/query/teamsGoals.graphql
+++ b/src/graphql/query/teamsGoals.graphql
@@ -1,35 +1,35 @@
-query GetGoals ($teamId: Int, $limit: Int) {
-  getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId}) {
-    totalCount
-    values {
-      id
-      description
-      name
-      participation {
-        totalCount
-        values {
-          lender {
-            id
-            image {
-              url
-            }
-          }
-        }
-      }
-      status
-      targets {
-        totalCount
-        values {
-          ... on LoanTarget {
-            loanId
-            status
-          }
-        }
-      }
-      endDate
-      ... on TeamGoal {
-        teamId
-      }
-    }
-  }
+query GetGoals($teamId: Int, $limit: Int) {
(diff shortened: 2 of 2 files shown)


src/components/Thanks/NotifyMe.vue
```diff
@@ -0,0 +1,132 @@
...
+	},
+	props: {
+		email: {
+			type: String,
+			default: '',
+		}
+	},
+	data() {
+		return {
+			iterableListIdString: '0daedc50-6b66-44de-898d-fc7365f64da5',
```
mcstover, line 83: Assuming this is for production, should we have as switch for an Iterable Dev version of this too to keep cruft from posting to the prod list? I guess if someone is manually managing that list it's ok...

src/assets/icons/inline/check-circle.svg
dyersituations, line 1: We could replace this with material UI `CheckCircleOutline`

src/graphql/query/teamsGoals.graphql
```diff
@@ -1,5 +1,5 @@
-query GetGoals {
-  getGoals(filters: { audienceType: TEAM, isActive: true}) {
+query GetGoals ($teamId: Int, $limit: Int) {
```
dyersituations, line 1: Will this query still work if these params are empty? We use this query on `/teams`. I'll merge my PR so you can have the latest changes locally, sorry for conflicts!

src/components/Thanks/NotifyMe.vue
dyersituations: Or do we just disable the button somehow for lower envs?

src/components/Thanks/NotifyMe.vue
```diff
@@ -0,0 +1,132 @@
...
+			// eslint-disable-next-line max-len
+			const response = await fetch(`//links.iterable.com/lists/publicAddSubscriberForm?publicIdString=${this.iterableListIdString}`, {
+				method: 'POST',
+				body: new URLSearchParams({
+					email: this.email,
+				})
+			});
+			if (response.status === 200) {
+				this.addedToIterable = true;
+				this.$showTipMsg(`We will notify ${this.email} when the challenge is announced!`);
```
dyersituations, line 113: `next challenge`

src/components/Thanks/NotifyMe.vue
```diff
@@ -0,0 +1,132 @@
...
+	props: {
+		email: {
+			type: String,
+			default: '',
+		}
+	},
+	data() {
+		return {
+			iterableListIdString: '0daedc50-6b66-44de-898d-fc7365f64da5',
+			addedToIterable: false,
```
dyersituations, line 84: Do we have a way to check if the user's email is already added? Otherwise, the notify button will be there each time the user checks out. That might be OK for the MVP...

src/graphql/query/teamsGoals.graphql
dyersituations: It's merged

src/graphql/query/teamsGoals.graphql
```diff
@@ -1,35 +1,35 @@
...
+    values {
+      id
+      description
+      name
+      participation {
+        totalCount
+        values {
+          lender {
+            id
 						image {
```
dyersituations, line 13: The tabbing should be fixed in this file.

dyersituations (approved):

dyersituations:
@christian14b Approved pending issue found in tests regarding formatting in `NotifyMe.vue`. See in the console of the `Ui Tests` action.

PR #5117, opened 2024-01-09
fix: add event to notify me button

christian14b:
Adding:
- Tracking event information to cta
- Adding secondary bg outside component

Fixing:
- Fetch goal information before page loads

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -129,6 +129,26 @@ import teamsGoalsQuery from '../../graphql/query/teamsGoals.graphql';
 const hasLentBeforeCookie = 'kvu_lb';
 const hasDepositBeforeCookie = 'kvu_db';
 
+const getLoans = receipt => {
+	const loansResponse = receipt?.items?.values ?? [];
+	const loans = loansResponse
+		.filter(item => item.basketItemType === 'loan_reservation')
+		.map(item => {
+			return {
+				...item.loan,
+				team: item.team,
+			};
+		});
+
+	return loans;
+};
+
+const getTeamId = loans => {
+	const teamsIds = loans.filter(loan => !!loan?.team?.id)
+		.map(loan => loan.team.id) ?? [];
+	return teamsIds?.[0] ?? null;
+};
+
 export default {
 	name: 'ThanksPage',
 	components: {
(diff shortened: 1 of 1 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -269,6 +271,15 @@ export default {
 		},
 	},
 	created() {
+		const filters = {
+			teamId: this.teamId,
+		};
+		const limit = 1;
+		fetchGoals(this.apollo, limit, filters)
```
emuvente, line 278: This will need to happen in the preFetch. Async functions in `created` won't complete on the server. The server immediately responds once `created` completes.

dyersituations: Good call, a fundamental difference from CPS/nuxt.

emuvente: > add the teamsGoalsQuery within the Promise.all, then use readQuery on created to assign the goal

Yeah, I think that's the best way to do this in this component.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -176,9 +179,32 @@ export default {
...
+				const teamId = teamsIds?.[0] ?? null;
+
+				const filters = {
+					teamId,
+				};
+				const limit = 1;
+
 				return Promise.all([
 					client.query({ query: experimentAssignmentQuery, variables: { id: 'share_ask_copy' } }),
+					fetchGoals(client, limit, filters),
```
dyersituations, line 207: This could be ran only if `teamId` is defined.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -176,9 +179,32 @@ export default {
...
+				const loans = loansResponse
+					.filter(item => item.basketItemType === 'loan_reservation')
+					.map(item => {
+						return {
+							...item.loan,
+							team: item.team,
+						};
+					});
+
+				const teamsIds = loans.filter(loan => !!loan?.team?.id)
```
dyersituations, line 196: Not critical to change, but the logic to get `loans` and `teamIds` is now duplicated in this component. Two small methods could be added to the component and shared.

dyersituations (approved):

PR #5120, opened 2024-01-10
feat: add global tracing for express + graphql, include trace and spa…

mcstover:
…n ids in winston logs

Couples with: https://github.com/kiva/flux/pull/607


server/util/tracer.js
```diff
@@ -0,0 +1,83 @@
...
+
+const { AlwaysOnSampler, SamplingDecision, SimpleSpanProcessor } = require('@opentelemetry/sdk-trace-base');
+const { GraphQLInstrumentation } = require('@opentelemetry/instrumentation-graphql');
+const { OTLPTraceExporter } = require('@opentelemetry/exporter-trace-otlp-proto');
+const { ConsoleSpanExporter, NodeTracerProvider } = require('@opentelemetry/sdk-trace-node');
+const { Resource } = require('@opentelemetry/resources');
+const { WinstonInstrumentation } = require('@opentelemetry/instrumentation-winston');
+const { SemanticResourceAttributes } = require('@opentelemetry/semantic-conventions');
+
+const Exporter = (process.env.EXPORTER || '').toLowerCase().startsWith('c') ? ConsoleSpanExporter : OTLPTraceExporter;
```
mcstover, line 16: I've included the ConsoleSpanExporter for local use. Starting the ui server locally with the "EXPORTER=console" env var will log spans for verification. We can remove after we refine this a bit more.

server/util/tracer.js
```diff
@@ -0,0 +1,83 @@
...
+const { ExpressInstrumentation } = require('@opentelemetry/instrumentation-express');
+const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http');
+
+const otlpDisabled = process.env?.OTEL_SDK_DISABLED || false;
```
hobbsh, line 20: The SDK should actually see this env var and disable things, we technically shouldn't need to wrap setup in a separate condition.

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#disabling-opentelemetrysdk

server/util/tracer.js
```diff
@@ -0,0 +1,83 @@
...
+// eslint-disable-next-line no-unused-vars
+function ignoreTheseSpans(spanName, spanKind, attributes) {
+	return spanKind !== opentelemetry.SpanKind.SERVER;
+}
+
+function setupTracing(serviceName) {
+	if (!otlpDisabled) {
+		const provider = new NodeTracerProvider({
+			resource: new Resource({
+				[SemanticResourceAttributes.SERVICE_NAME]: serviceName,
```
hobbsh, line 49: We can set both of these with `OTLP_` environment variables. Would that be acceptable?

server/util/tracer.js
```diff
@@ -0,0 +1,83 @@
...
+			return `FilterSampler(${parent.toString()})`;
+		}
+	};
+}
+
+// Basic filter function, currently excludes internal express spans
+// attributes are helpful if we want to inspect and omit requests such as health checks or timesync
+// eslint-disable-next-line no-unused-vars
+function ignoreTheseSpans(spanName, spanKind, attributes) {
+	return spanKind !== opentelemetry.SpanKind.SERVER;
```
hobbsh, line 42: Are we sure `SERVER` spans should be excluded?

server/util/tracer.js
hobbsh: Would it be possible to name this environment variable something more descriptive? Like `OPENTELEMETRY_EXPORTER_TYPE`?

server/util/tracer.js
mcstover: This is right from there example for express so maybe... Ultimately, I had problems getting the filter function to work on broader excludes so I trimmed it back down to what they had in their examples.

server/util/tracer.js
mcstover: What OTLP vars exists for the Ui pods at present?

server/util/tracer.js
mcstover: testing this out. I wrapped it to prevent initialization of those extra libs as I do have concerns about the performance implications of these libraries.

mcstover: Just using that ENV var does not work across all of the different collectors, it also doesn't disable the exporter... sticking with the wrapper for now.

server/util/tracer.js
mcstover: Sure, this is temporary but I can update

server/util/tracer.js
mcstover: Ok added these in. OTEL_SERVICE_NAME exists but OTEL_SERVICE_NAMESPACE does not.

hobbsh (approved):
looks good, thanks!

dyersituations (approved):

PR #5122, opened 2024-01-11
test: ensure teams are loaded on Checkout Page to get forceTeamId

christian14b:
Problem: I've been seeing duplicated teams in Teams dropdown because teams are not assigned at the moment forceTeamId is requested. With this change the component will wait until teams changes to call forceTeamId.

I'm still trying to fix the initial problem but this seems to be related.

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -191,6 +191,11 @@ export default {
 			return this.loan.team ? this.loan.team.id : 0;
 		}
 	},
+	watch: {
+		teams() {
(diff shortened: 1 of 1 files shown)


src/components/Checkout/BasketItem.vue
```diff
@@ -191,6 +191,11 @@ export default {
 			return this.loan.team ? this.loan.team.id : 0;
 		}
 	},
+	watch: {
+		teams() {
```
dyersituations, line 195: Looking at `ThanksPage`, it looks like teams should be prefetched, so I wouldn't think a `watch` was needed. Is that not the case?

dyersituations: Okay makes sense. Using the `watch` isn't a bad thing...

dyersituations (approved):

PR #5128, opened 2024-01-12
fix: tracking format in View challenge button

christian14b:


src/components/Teams/TeamGoal.vue
```diff
@@ -71,12 +71,12 @@
 					</span>
 				</div>
 				<kv-button
-					:v-kv-track-event="[
+					v-kv-track-event="`[
```
dyersituations, line 74: The ` can be removed

dyersituations (approved):

PR #5129, opened 2024-01-16
fix: apollo error

christian14b:
Injection needed for the tip message

image


emuvente (approved):

src/components/Thanks/NotifyMe.vue
```diff
@@ -73,6 +73,7 @@ import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
 
 export default {
 	name: 'NotifyMe',
+	inject: ['apollo'],
```
dyersituations, line 76: Which component uses `apollo` in this situation?

dyersituations (approved):

PR #5131, opened 2024-01-16
fix: ensure forceTeamId is not overwritten by undefined value

christian14b:
Ensure we override the teamId when there's a cookie present. 

As `combinedTeams` has the team from cookie already merged, at the moment forceTeamId is requested ` !combinedTeams.some(team => team.id === loan.teamId)` will always be false and forceTeamId will be undefined, causing `loanTeamAttributionId` to be `loan.team.id`.


dyersituations (approved):

dyersituations:
@christian14b Looks like we need unit test changes.

PR #5132, opened 2024-01-17
fix: mutate loan reservation on mounted

christian14b:
Calling `updateLoanReservationTeam` when `TeamAttribution` is mounted to assign the loan to a team and also removing `updateLoanReservation` condition because when you select the same option the method is not triggered by `@update:modelValue`, it's being called when you select a different option.

### DIFF
--- a/test/unit/specs/components/Checkout/BasketItem.spec.js
+++ b/test/unit/specs/components/Checkout/BasketItem.spec.js
@@ -29,6 +29,7 @@ describe('BasketItem loan', () => {
 						readFragment: () => {},
 						query: () => Promise.resolve({}),
 						readQuery: () => {},
+						mutate: () => Promise.resolve({}),
 					},
 					cookieStore: new CookieStore(),
 				},


src/components/Checkout/TeamAttribution.vue
```diff
@@ -74,41 +74,40 @@ export default {
 	},
 	mounted() {
 		this.selectedId = this.teamId || 0;
+		this.updateLoanReservation();
```
dyersituations, line 77: Is the team ID definitely populated at this point?

src/components/Checkout/TeamAttribution.vue
```diff
@@ -74,41 +74,40 @@ export default {
...
 	},
 	methods: {
 		updateLoanReservation() {
-			if (this.selectedId !== this.teamId) {
```
dyersituations, line 80: If we wait for the team ID to be defined in mounted, do we need to remove this check?

dyersituations: Or is this just in case?

dyersituations:
@christian14b Seeing an interesting/confusing unit test failure.

roger-in-kiva:
@dyersituations we are struggling with the thanks page locally, I think is monolity API, are you able to test this PR?

src/components/Checkout/TeamAttribution.vue
dyersituations: Sounds like just in case, which I think is fine.

dyersituations (approved):

roger-in-kiva (approved):

dyersituations (approved):

PR #5133, opened 2024-01-17
fix: calling updateLoanReservation when teamId changes

christian14b:


src/components/Checkout/TeamAttribution.vue
```diff
@@ -69,12 +69,12 @@ export default {
 		teamId(newId, prevId) {
 			if (newId !== prevId) {
 				this.selectedId = newId;
+				this.updateLoanReservation();
```
dyersituations, line 72: This gets called as expected when the page/component loads? Since `selectedId` gets set to `teamId` prop on `data`, I could see there potentially being a situation where this watch doesn't get called on load. Or will this watch also get called in that situation, where a prop is provided on load.

dyersituations: Sure, immediate true _should_ cover this situation. Are there repro steps that we can use on dev to make sure this works once deployed?

dyersituations: Sounds like we just need the `immediate` property. We use that property elsewhere, if you're looking for examples in the repo.

dyersituations (approved):

PR #5136, opened 2024-01-17
fix: check all loans instead of selected loan to show CTA on thanks page

christian14b:
Check all loans Instead of evaluating selectedLoan's id

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -363,11 +363,9 @@ export default {
 
 				this.goal = response.getGoals?.values.length ? response?.getGoals?.values[0] : null;
 
-				const teamsIds = this.loans.filter(loan => !!loan?.team?.id)
-					.map(loan => loan.team.id) ?? [];
-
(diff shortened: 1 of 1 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -362,8 +362,12 @@ export default {
 				});
 
 				this.goal = response.getGoals?.values.length ? response?.getGoals?.values[0] : null;
+
+				const teamsIds = this.loans.filter(loan => !!loan?.team?.id)
+					.map(loan => loan.team.id) ?? [];
+
 				this.showNotifyMe = this.goal && this.goal?.targets?.values
-					.findIndex(target => target.loanId === this.selectedLoan.id) !== -1;
+					.findIndex(target => teamsIds.includes(target.loanId)) !== -1;
```
dyersituations, line 370: Looks like this is comparing loan ID vs team ID, instead of loan ID vs loan ID.

dyersituations (approved):

PR #5139, opened 2024-01-19
fix: [ACK-959] filter goals that are not in progress for team query

1337jwilliams95:
I think this all we need once the back end change comes through ?

### DIFF
--- a/src/graphql/query/teamsGoals.graphql
+++ b/src/graphql/query/teamsGoals.graphql
@@ -1,5 +1,5 @@
 query GetGoals($teamId: Int, $limit: Int) {
-	getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId}) {
+	getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId, status: [IN_PROGRESS]}) {
 		totalCount
 		values {
 			id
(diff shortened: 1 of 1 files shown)


src/graphql/query/teamsGoals.graphql
```diff
@@ -1,5 +1,5 @@
 query GetGoals($teamId: Int, $limit: Int) {
-	getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId}) {
+	getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId, status: [IN_PROGRESS]}) {
```
dyersituations, line 2: We might want to future proof by adding `NONE`, since that status still might be used.

1337jwilliams95:
```suggestion
	getGoals(limit: $limit, filters: { audienceType: TEAM, isActive: true, teamId: $teamId, status: [IN_PROGRESS, NONE]}) {
```

1337jwilliams95: one thing to note, from what I've seen goals that are intended for the future (this is based on integration test cases) are in the status NONE. I'm wondering if we should/could get rid of the isActive field in the future and just use this status, as they seem a bit redundant and if we fail to update one or the other at some point we may face some data consistency issues.

dyersituations: The `NONE` status was created, but we didn't add code to change that to `IN_PROGRESS` in the service. So we started the concept of using `NONE`, but some work remained. One different is `isActive` is that it checks the dates on the goal, so that is still important.

dyersituations (approved):

emuvente (approved):

1337jwilliams95:
Screenshot 2024-01-22 at 11 11 17 AM
@dyersituations marked the aplus challenge complete in dev, and this is the result with the filter locally. it LGTM so I'm going to merge it

PR #5141, opened 2024-01-22
feat: show only the newest goal per team

christian14b:
Show newest goal to handle multiple active goals

### DIFF
--- a/src/components/Teams/TeamGoalsList.vue
+++ b/src/components/Teams/TeamGoalsList.vue
@@ -68,7 +68,7 @@ export default {
 		result(result) {
 			const teams = _groupBy(result?.data?.getGoals?.values ?? [], 'teamId');
 			this.activeGoals = _map(teams, goals => {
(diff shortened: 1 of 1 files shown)


src/components/Teams/TeamGoalsList.vue
```diff
@@ -63,7 +66,10 @@ export default {
 		query: teamsGoals,
 		preFetch: true,
 		result(result) {
-			this.activeGoals = result?.data?.getGoals?.values ?? [];
+			const teams = _groupBy(result?.data?.getGoals?.values ?? [], 'teamId');
+			this.activeGoals = _map(teams, goals => {
+				return _orderBy(goals, 'startDate', 'desc')[0];
```
dyersituations, line 71: Here `startDate` is a string. Will sorting work consistently? I'm assuming the string might sort like expected, but you might need to convert the date strings to actual date objects.

dyersituations (approved):

PR #5143, opened 2024-01-24
feat: setup stream api class

christian14b:
Get Stream API Class setup

### DIFF
--- a/src/util/streamService.js
+++ b/src/util/streamService.js
@@ -1,0 +1,78 @@
+import { connect } from 'getstream';
+
+/**
+ * A class that abstracts calls to the GetStream.io API.
+ *
+ * @class StreamService
+ */
+class StreamService {
+	/**
+	 * Creates an instance of StreamService.
+	 *
+	 * @param {string} apiKey - The GetStream.io API key.
+	 * @param {string} appId - The GetStream.io App id.
+	 * @param {string} authToken - The auth user token.
+	 * @memberof StreamService
+	 */
+	constructor(apiKey, authToken, appId) {
+		/**
+		 * The GetStream.io client instance.
+		 *
+		 * @type {Stream.Client}
+		 */
+		this.client = connect(apiKey, authToken, appId);
+	}
+
+	/**
+	 * Creates a user feed for the specified user ID.
+	 *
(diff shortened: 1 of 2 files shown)


mcstover:
@christian14b The env specific config files in this repo are no longer used. You'll want to create an env based switch for these new configs in the `dynamic.js` config file that assigns the value of a config passed in from the flux deployment for UI. You can add initial env vars for UI here: https://github.com/kiva/flux/blob/main/services/development/ui.yaml#L39

dyersituations:
@christian14b Do we need this API in the `ui` repo? Or do we just need it in CPS? If we need it in both repos, this should be built in `kv-elements`. My hope is that we could just build this in CPS for now, but it depends on the use cases.

src/util/streamService.js
```diff
@@ -0,0 +1,78 @@
...
+	/**
+	 * Adds an activity to the specified feed.
+	 *
+	 * @param {Stream.Feed} feed - The feed to which the activity will be added.
+	 * @param {object} activity - The activity object to add
+	 * Read: https://getstream.io/activity-feeds/docs/javascript/adding_activities/?language=javascript#fields
+	 * @returns {Promise<void>}
+	 * @memberof StreamService
+	 */
+	static addActivity(feed, activity) {
```
dyersituations, line 61: What is the shape of the `activity` object? This class should probably own what that means, so splitting this out into params would probably make sense.

OR, if we move this to CPS (see other comment), then we can use TypeScript for type safety here.

src/util/streamService.js
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Retrieves activities from a feed based on the provided parameters.
+	 *
+	 * @param {Stream.Feed} feed - The feed from which to retrieve activities.
+	 * @param {object} params - Parameters for the get request (e.g., limit).
+	 * @returns {Promise<Stream.GetStreamResponse>}
+	 * @memberof StreamService
+	 */
+	static getActivity(feed, params) {
```
dyersituations, line 73: Same comment about `params`. What shape is expected? A component/class using this should need to know the shape.

PR #5151, opened 2024-01-31
fix: set cookie headers not being added to response

emuvente:


dyersituations (approved):

server/vue-worker.js
```diff
@@ -47,22 +47,22 @@ async function render(context) {
...
-			context.setCookies = [...cookieInfo.setCookies];
+			// collect any newly fetched 'Set-Cookie' headers to send after the render
+			setCookies = [...cookieInfo.setCookies];
 			// render the app
 			return renderer.renderToString(context);
 		})
 		.then(html => {
 			// collect any cookies created during the app render
-			setCookies.concat(context.setCookies);
+			setCookies = [...setCookies, ...context.setCookies];
```
mcstover, line 65: will setCookies exist on the context if we're not setting them above? I'm getting a `uncaughtException: context.setCookies is not iterable` locally.

mcstover: I def see setCookies on the context and it's an array

emuvente: When did you see that? I haven't been able to reproduce that error. I can add a check to be safe though.

mcstover: I also see them set on the html response so It may be this only happened in the error case locally.

mcstover (approved):

dyersituations (approved):

server/vue-worker.js
mcstover: I think the error case was triggered b/c I still had a browser window open when I started the dev server after updating the branch... so that page was calling with hmr constantly until something was available to respond...

PR #5154, opened 2024-02-02
chore: setup in-context checkout stories

dyersituations:
- We have another use case of in-context checkout coming up for IWD
- In all likelihood we'll need a new component
- But we might as well get the existing component working in storybook for a starting point
- The prop defaults were incorrect and leading to errors in the story render

image


mcstover (approved):

dyersituations:
I realized this new page is only in CPS, so instead of making these changes, I'll just bring this component over to CPS and make it work in Vue 3.

PR #5156, opened 2024-02-05
feat: added flag for IWD to thanks page, created working stories

dyersituations:
https://kiva.atlassian.net/browse/CORE-1729

- First step in IWD thanks page changes -> make it more easily testable locally
- Created working stories with variations for with and without conditions for IWD
- Fixed apollo mixin for `readQuery` and `readFragment`
- Data contains "real" data w/ fake IDs and names (real image kept)

image

### DIFF
--- a/.storybook/mixins/apollo-story-mixin.js
+++ b/.storybook/mixins/apollo-story-mixin.js
@@ -5,13 +5,13 @@ export default ({
 	loading = false,
 } = {}) => {
 	// Add to basket button expects basketAddInterstitial
-	const queryData = { ...queryResult, data: { ...queryResult.data, basketAddInterstitial: {} } };
+	const queryData = { ...queryResult.data, data: { ...queryResult.data, basketAddInterstitial: {} } };
 
 	return {
 		provide: {
 			apollo: {
 				mutate() {
-					return loading ? new Promise(() => {}) : Promise.resolve(mutationResult);
+					return loading ? new Promise(() => { }) : Promise.resolve(mutationResult);
 				},
 				readQuery() {
 					return queryData;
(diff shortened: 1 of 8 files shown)


.storybook/mixins/apollo-story-mixin.js
```diff
@@ -5,13 +5,13 @@ export default ({
 	loading = false,
 } = {}) => {
 	// Add to basket button expects basketAddInterstitial
-	const queryData = { ...queryResult, data: { ...queryResult.data, basketAddInterstitial: {} } };
+	const queryData = { ...queryResult.data, data: { ...queryResult.data, basketAddInterstitial: {} } };
```
mcstover, line 8: This might affect some other stories...there seem to be several that try to pull their data off of `data`...

dyersituations: Are those instances using a `query` or `readQuery`? Queries pull from `data` and `readQuery` doesn't have `data` in the response.

dyersituations: I'm wondering if `readQuery` has never working in our stories...

dyersituations: For sure, will do, thanks.

mcstover (approved):

.storybook/mixins/apollo-story-mixin.js
dyersituations: Clicked through the stories. Added some quick resolutions with missing mixins. There are some failing stories, but those seem to be from different/specific issues.

dyersituations: And those failures exist in the deployed storybook instance as well.

mcstover: COol thank you!

PR #5157, opened 2024-02-07
feat: first IWD thanks page variation

dyersituations:
https://kiva.atlassian.net/browse/CORE-1729

- Implemented the first IWD thanks page variation
- This is when a user checks out after being invited by a user
- The query string in this situation will have `valet_inviter` as either `kiva` or the lender public ID
- There is a slight variation on whether the inviter is Kiva, but below is the situation where there's a lender public ID
- Implemented decorator to change a story iframe URL before it gets rendered

image

image

### DIFF
--- a/.storybook/stories/ThanksPage.stories.js
+++ b/.storybook/stories/ThanksPage.stories.js
@@ -2,26 +2,46 @@ import ThanksPage from '@/pages/Thanks/ThanksPage';
 import apolloStoryMixin from '../mixins/apollo-story-mixin';
 import cookieStoreStoryMixin from '../mixins/cookie-store-story-mixin';
 import kvAuth0StoryMixin from '../mixins/kv-auth0-story-mixin';
-import { maleLoanData, femaleLoanData, iwdExperiment } from '../mock-data/thanks-page-data-mock';
+import {
+	maleLoanData,
+	femaleLoanData,
+	femaleLoanDataWithInviter,
+	iwdExperiment,
+} from '../mock-data/thanks-page-data-mock';
+import VueRouter from 'vue-router';
+
+const routes = new VueRouter();
 
 export default {
 	title: 'Page/ThanksPage',
 	component: ThanksPage,
+	decorators: [
+		(story, { args }) => {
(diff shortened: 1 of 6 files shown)


src/components/Iwd/IwdThanksPageVariations.vue
```diff
@@ -0,0 +1,178 @@
...
+						md:tw-mt-2
+						tw-rounded-b
+						md:tw-rounded-t
+						tw-p-1.5
+						tw-gap-1
+						tw-items-center
+					"
+				>
+					<div class="tw-shrink-0">
+						<img :src="iWD2024Badge" alt="IWD Badge" class="tw-w-10 md:tw-w-6">
```
mcstover, line 20: Rounded corners?

mcstover (approved):

src/components/Iwd/IwdThanksPageVariations.vue
dyersituations: Are you referring to here (the badge) or the borrower image? The mocks didn't show rounded corners, but you are correct that we usually round our borrower images...

mcstover: Ya, the borrower image looks a like clunky + unadorned...

.storybook/stories/ThanksPage.stories.js
```diff
@@ -2,26 +2,46 @@ import ThanksPage from '@/pages/Thanks/ThanksPage';
...
+	iwdExperiment,
+} from '../mock-data/thanks-page-data-mock';
+import VueRouter from 'vue-router';
+
+const routes = new VueRouter();
 
 export default {
 	title: 'Page/ThanksPage',
 	component: ThanksPage,
+	decorators: [
```
mcstover, line 18: This is neat, thank you!

src/components/Iwd/IwdThanksPageVariations.vue
dyersituations: Agreed, added rounded corners to match other instances of the borrower image:

image

PR #5160, opened 2024-02-08
feat: add campaign progress bar to category page

dyersituations:
https://kiva.atlassian.net/browse/CORE-1724

- Updated the campaign progress bar to match the updated mocks + data
- Moved the IWD header into its own file
- Created a wrapper component for the shared progress bar usage
- Added the bar to the category page for IWD
- Small fix for padding on thanks page from previous PR

image

image

**UPDATE**

- Added the progress bar to the thanks page variations for IWD
- Created minimal version of the progress bar

image

image

image

image

### DIFF
--- a/.storybook/mock-data/thanks-page-data-mock.js
+++ b/.storybook/mock-data/thanks-page-data-mock.js
@@ -112,6 +112,9 @@ export const maleLoanData = {
 	}
 }
 
+const futureDate = new Date();
+futureDate.setDate(futureDate.getDate() + 20);
+
 export const femaleLoanData = {
 	"data": {
 		"shop": {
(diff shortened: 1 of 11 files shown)


mcstover (approved):
With a release planned today, we'll need to keep an eye on those missing queries... If active on server render they could fail the page load. Should break more gracefully if only in the client.

dyersituations:
> With a release planned today, we'll need to keep an eye on those missing queries... If active on server render they could fail the page load. Should break more gracefully if only in the client.

@mcstover Good call. The monolith is going out, so I think we should be OK. Also the new fields added to the server-side call are all existing fields. The new queries are called on mount. But just in case I wrapped the activity feed query in a try catch, because we weren't catching exceptions before. The other new stats query is already wrapped in a try catch. In the process I noticed that we were actually pulling the wrong campaign for the activity feed, so I updated that as well!

mcstover (approved):

PR #5164, opened 2024-02-13
feat: async checkout polling

mcstover:
Started a fresh branch off of `main` and cherry-picked the 2 outstanding original commits.

- Imported the kv-shop methods
- Cleaned up old unused method
- Updated kv-shop lib after merging https://github.com/kiva/kv-ui-elements/pull/341
- Backend Task for missing Status Enum (MPL-30) Fixed by: https://github.com/kiva/checkout-orchestrator/pull/159
- Split Error formatting on general Graphql error vs a checkoutStatus Error response

CURRENT STATE:
- Happy path async checkout works for both KivaCredit and Braintree checkouts, I was able to observe multiple network calls to the checkoutStatus mutation and transition to the thanks page after hitting the "COMPLETED" status.


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -338,40 +339,73 @@ export default {
 						visitorId: this.cookieStore.get('uiv') || null
 					},
 				})
-				.then(kivaBraintreeResponse => {
-					// Check for errors in transaction
+				.then(async kivaBraintreeResponse => {
+					// extract transaction saga id or transaction id from response
+					const transactionResult = this.useAsyncCheckout
```
dyersituations, line 344: The new endpoint also returns a transaction ID? The naming of this as "result" confused me at first, since we used "id" before.

src/components/Checkout/KivaCreditPayment.vue
```diff
@@ -47,29 +48,56 @@ export default {
...
-						this.$emit('complete-transaction', transactionResult);
+				.then(async transactionResult => {
+					if (this.useAsyncCheckout && typeof transactionResult !== 'object') {
+						await pollForFinishedCheckout({
+							apollo: this.apollo,
+							transactionSagaId: transactionResult,
+						})
+							.then(checkoutStatusResponse => {
+								// eslint-disable-next-line max-len
+								this.handleSuccessfulCheckout(checkoutStatusResponse?.data?.checkoutStatus?.receipt?.checkoutId);
```
dyersituations, line 64: I see that `checkoutId` has optional chaining, does there need to be a fallback other than `undefined`?

src/graphql/query/checkout/checkoutStatus.graphql
```diff
@@ -3,6 +3,9 @@ query checkoutStatus($transactionId: String!, $visitorId: String) {
 		basketId
 		errorCode
 		errorMessage
+		receipt {
+		checkoutId
```
dyersituations, line 7: Missing indentation.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
mcstover: Changed to `transactionResponse`

dyersituations (approved):

PR #5167, opened 2024-02-16
chore: update configs for custom domain and dev mode with caddy

mcstover:
This is an alternate approach to developing UI locally which allows for both login and checkout while on the VPN.

It uses a simple Caddy server to proxy requests to a locally running UI instance targeting our dev environment.

Setup:
- Add `127.0.0.1 kiva-ui.local` to your `/etc/hosts` file
- Auth0 configs are already in place
- `caddy start` from the root of the ui repo
- `npm run dev -- --config=dev-custom-host` in a separate terminal at the root of the ui repo

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -1,0 +1,7 @@
+{
+	http_port 8989
+}
+
+kiva-ui.local {
+	reverse_proxy localhost:8888
+}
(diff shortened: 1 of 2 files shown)


dyersituations:
I wonder if it would simplify things to write a small script that does `caddy start`, then runs in the new dev mode? Just so we can start the local dev env with a single command?

mcstover:
Feel free to add a flag or new command to `ui-server.sh` in the root of this repo... that script can prob be cleaned up and refined to a more modern usage too...

dyersituations:
> Feel free to add a flag or new command to `ui-server.sh` in the root of this repo... that script can prob be cleaned up and refined to a more modern usage too...

I'll circle back after IWD...

dyersituations (approved):

PR #5182, opened 2024-03-05
feat: visible url array added to global promotional banner component

roger-in-kiva:
- visible url array added to global promotional banner component from contentful

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
+++ b/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
@@ -107,7 +107,7 @@ export default {
 				if (activePromoBanner && activePromoBanner?.fields) {
 					// check for visibility based on current route and hiddenUrls field
 					const hiddenUrls = globalBannerDenyList.concat(activePromoBanner?.fields?.hiddenUrls ?? []);
-					const visibleUrls = [];
+					const visibleUrls = activePromoBanner?.fields?.visibleUrls ?? [];
 					if (isExcludedUrl(hiddenUrls, visibleUrls, this.$route.path)) return false;
 
 					// check for visibility on promo session override


dyersituations (approved):

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -107,7 +107,7 @@ export default {
 				if (activePromoBanner && activePromoBanner?.fields) {
 					// check for visibility based on current route and hiddenUrls field
 					const hiddenUrls = globalBannerDenyList.concat(activePromoBanner?.fields?.hiddenUrls ?? []);
-					const visibleUrls = [];
+					const visibleUrls = activePromoBanner?.fields?.visibleUrls ?? [];
```
mcstover, line 110: This may have unintended consequences, hence. How do you plan to test and verify this doesn't affect behavior of our other visibility controls?

dyersituations: @roger-in-kiva is this 100% matching the functionality in `ui`?

mcstover: Wasn't the need for this on legacy team pages?

mcstover: It's good to sync up across apps though, just want to make sure we understand when conflicting (ie hidden urls) are also blended in.

mcstover: Teams is in CPS too lolz

mcstover: It may not be here primarily b/c @christian14b was working on the Donate banner type and the "visibleUrls" approach wasn't necessarily specced out for the standard banner. So, again, nice to have this added.

dyersituations: @roger-in-kiva Once this is test on dev, let us know here if it helped with one of the use cases, and which URL specifically. Thanks!

roger-in-kiva: In fact, we have teams in legacy and CPS, but adding `ui-global-promo` stuff, was also affecting `ui` repo

There is some work left on legacy to make it work as expected...

mcstover: I think one of the main things here is that we are able to have one of these specialized team display banners on at the same time as a global promo banner and the global promo banner shows everywhere except where the customized visibleUrl banner shows...

dyersituations: Oh right, it was appearing in `ui` pages because visible URLs wasn't working... Let us know if this helps on dev!

dyersituations: @roger-in-kiva is it true that we are using the global banner for team challenges?

roger-in-kiva: Before this updated, we were showing the banner across the `ui` pages, right now is filtered by the visible url setting in contentful.

Shehaam reported this case, https://www.development.kiva.org/lend-by-category

Hidden now!

roger-in-kiva: It's true b/c the query for promo banners in CPS is having that type: 
https://github.com/kiva/cms-page-server/blob/main/graphql/query/bannerContentful.graphql

Global but filtered, lolz

PR #5184, opened 2024-03-06
feat: setup test for activity feed on borrower profile

christian14b:


src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -487,6 +489,7 @@ export default {
 						client.query({ query: experimentAssignmentQuery, variables: { id: SHARE_LANGUAGE_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: FIVE_DOLLARS_NOTES_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: EDUCATION_PLACEMENT_EXP } }),
+						client.query({ query: experimentAssignmentQuery, variables: { id: ACTIVITY_FEED_EXP } }),
```
roger-in-kiva, line 492: I think is needed no more...
@dyersituations could have more context.

dyersituations: Correct, this line isn't needed, since we assign all active experiments globally based on the Kiva Admin setting.

roger-in-kiva: Maybe a tech debt removing all those cases...

dyersituations: Yeah we can leave the extra ones for now, so we don't have to functional verify everything.

dyersituations (approved):

PR #5188, opened 2024-03-07
fix: ensure prefetched channel data is correctly pulled from apollo cache

dyersituations:
https://kiva.atlassian.net/browse/CORE-1783

- We were getting an apollo cache miss due to using `undefined` as the fallback sort enum value


roger-in-kiva (approved):
Thanks!

mcstover (approved):
Thanks, was just noticing that console in my local ;)

PR #5190, opened 2024-03-07
feat: activities were not shown due to lendCta being mounted before k…

christian14b:
…nowing if experiment was enabled. Also, adding remaining props to kvLoanActivities

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -423,7 +423,7 @@ export default {
 		},
 		activities: {
 			type: Object,
-			default: () => ({}),
+			default: null,
 		}
 	},
 	components: {
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -434,7 +434,6 @@ export default {
 				'Asia',
 				'Europe'
 			],
-			enableActivityFeed: false,
 			activities: null,
 		};
 	},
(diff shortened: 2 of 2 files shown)


roger-in-kiva (changes requested):
addToBasket emit is not working as expected...

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -570,8 +574,9 @@ export default {
 		},
 	},
 	methods: {
-		async addToBasket() {
+		async addToBasket(lendAmount = 0) {
```
roger-in-kiva, line 577: Working fine from the modal but BP lend cta is broken...

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -585,11 +590,11 @@ export default {
 				}
 			}).catch(e => {
 				this.isAdding = false;
-				const msg = e[0]?.extensions?.code === 'reached_anonymous_basket_limit' && e[0]?.message
+				this.errorMsg = e[0]?.extensions?.code === 'reached_anonymous_basket_limit' && e[0]?.message
```
roger-in-kiva, line 593: We need to clean this when addToBasket method is called.

roger-in-kiva (approved):

src/components/BorrowerProfile/LendCta.vue
```diff
@@ -417,9 +421,9 @@ export default {
 			type: Boolean,
 			default: false,
 		},
-		enableActivityFeed: {
-			type: Boolean,
-			default: false,
+		activities: {
+			type: Object,
+			default: () => ({}),
```
dyersituations, line 426: If we are checking on whether this is defined, then this default should also be `null`

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -594,6 +596,12 @@ export default {
 		);
 		if (activityFeedExpData?.version === 'b') {
 			this.enableActivityFeed = true;
```
dyersituations, line 598: Can this flag be removed now that we aren't checking it?

dyersituations (approved):

PR #5191, opened 2024-03-08
fix: add an apollo retry link to handle network errors

mcstover:
Adds a [retry link](https://www.apollographql.com/docs/react/api/link/apollo-link-retry/) which only operates on network errors.

I've not been able to replicate the errors we see in prod, locally. But hope we can try to replicate it in dev. We'll still have logging in place for when this happens for the time being. If this works, we can reduce the verbosity of that log.

Activate in dev: https://github.com/kiva/flux/pull/789/files


src/api/NetworkErrorRetryLink.js
```diff
@@ -0,0 +1,5 @@
+import { RetryLink } from '@apollo/client/link/retry';
+
+export default () => {
+	return new RetryLink();
```
hobbsh, line 4: Do we need to add any defaults (and configurable via env vars so changes are easier)?

mcstover: I don't believe so. They document their defults in the link above.

mcstover: There's a max of 5 and also delay mechanisms...

hobbsh: OK I only quickly read it and didn't catch that those were defaults.

hobbsh: With max of 5 I'm wondering if that's too much, because it could tie up workers long enough for pending tasks to pile up. Maybe we should reduce it to 1 or two or make it configurable?

mcstover: Looking at adding a couple configs to allow us to control the attempts and activate/deactivate retries

dyersituations (approved):

config/dynamic.js
```diff
@@ -1,6 +1,8 @@
 const path = require("path");
 
 const apolloBatching = process.env.APOLLO_BATCH !== 'false';
+const apolloNetworkErrorRetryActive = process.env.APOLLO_NETWORK_RETRY_ACTIVE !== 'false';
```
dyersituations, line 4: `process.env.APOLLO_NETWORK_RETRY_ACTIVE === 'true'` might be safer?

hobbsh (approved):

JoeTravisKiva (approved):

PR #5196, opened 2024-03-12
feat: powered by lenders in borrower profile

christian14b:
image

### DIFF
--- a/src/components/BorrowerProfile/PoweredByLenders.vue
+++ b/src/components/BorrowerProfile/PoweredByLenders.vue
@@ -1,26 +1,15 @@
 <template>
 	<div class="tw-flex tw-items-center tw-mb-1.5">
-		<div class="tw-flex tw-shrink-0">
-			<img
-				v-for="(lender, i) in participationLendersDisplayed"
+		<div class="tw-flex tw-shrink-0 tw-items-end" data-testid="participation-lenders">
+			<kv-user-avatar
 				:key="lender.id"
-				:src="lender.image.url"
-				alt="Lender photo"
-				class="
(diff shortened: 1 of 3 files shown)


test/unit/specs/components/BorrowerProfile/PoweredByLenders.spec.js
```diff
@@ -0,0 +1,92 @@
+import { render } from '@testing-library/vue';
+import PoweredByLenders from '@/components/BorrowerProfile/PoweredByLenders';
+
+const participants = {
```
dyersituations, line 4: Could this also use the new mock data?

src/components/BorrowerProfile/PoweredByLenders.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+	<div class="tw-flex tw-items-center tw-mb-1.5">
+		<div class="tw-flex tw-shrink-0">
+			<img
```
dyersituations, line 4: Can we use `KvUserAvatar`?

test/unit/specs/components/BorrowerProfile/PoweredByLenders.spec.js
roger-in-kiva: We already have contributing lender on the page, why using mock data?

dyersituations: I was referring to the storybook mock data in this changeset, though I guess we don't typically share mocks between storybook and jest, so disregard.

dyersituations (approved):

PR #5200, opened 2024-03-14
feat: progress bar in challenge header

christian14b:
Updating campaign progress bar to handle amounts instead of people and adding an Icon at the end.

image

### DIFF
--- a/src/components/Lend/LoanSearch/ChallengeHeader.vue
+++ b/src/components/Lend/LoanSearch/ChallengeHeader.vue
@@ -1,29 +1,100 @@
 <template>
-	<div>
-		<kv-progress-campaign
-			:funded-amount="fundedAmount"
-			:total-amount="totalAmount"
-			:days-left="daysLeft"
-		/>
+	<div class="tw-flex tw-flex-col lg:tw-flex-row tw-w-full tw-gap-1 lg:tw-gap-5 tw-mb-4 lg:tw-mb-6">
+		<div class="lg:tw-basis-3/5 tw-flex tw-flex-col tw-gap-y-2">
+			<div class="tw-flex tw-items-center tw-gap-2">
+				<img
+					v-if="teamImageUrl"
+					class="tw-w-4 tw-h-4 lg:tw-w-8 lg:tw-h-8 tw-flex-none tw-rounded-sm"
+					:src="teamImageUrl"
+				>
+				<div>
+					<h4
+						v-if="teamName"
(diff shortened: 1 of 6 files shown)


roger-in-kiva:
@christian14b Missing shadow for the icon placeholder.

src/components/Kv/KvProgressCampaign.vue
```diff
@@ -2,46 +2,51 @@
...
-			:bg-variant="bgVariant"
-		/>
-		<div v-if="!minimalStats">
-			<p>
-				{{ daysLeft }} days left
-			</p>
+		<div class="tw-relative tw-mt-1">
+			<kv-progress-bar
+				class="tw-my-0.5"
+				aria-label="Percent the loan has funded"
```
dyersituations, line 14: `Percent the campaign has funded`

src/components/Kv/KvProgressCampaign.vue
```diff
@@ -2,46 +2,51 @@
...
 export default {
 	name: 'KvProgressCampaign',
 	components: {
-		KvProgressBar
+		KvProgressBar,
+		HeartOutFromBox
 	},
 	props: {
-		fundedBorrowers: {
+		fundedAmount: {
```
dyersituations, line 37: With these prop changes, make sure to remove it from the existing usage, `IwdProgressCampaign`. We still might be user testing the landing page, so we need to make sure it still loads, but the progress bar won't be shown without an active campaign, so we can remove it from that component.

src/components/Lend/LoanSearch/ChallengeHeader.vue
```diff
@@ -1,17 +1,42 @@
...
+		fundedAmount() {
+			// TODO: Change this when we have the correct data name
+			return this.challengeData?.amountGoal ?? 0;
+		},
+		totalAmount() {
+			return this.challengeData?.participation?.amountLent ?? 0;
+		},
+		daysLeft() {
+			return intervalToDuration({
+				start: this.challengeData?.endDate ?? new Date(),
```
dyersituations, line 36: `startDate`?

src/components/Kv/KvProgressCampaign.vue
dyersituations: Let's keep that component for now, just comment out the `kv-progress-campaign`. We still don't know the future of that IWD page.

dyersituations (approved):

PR #5206, opened 2024-03-18
feat: team picks handling for challenge header

roger-in-kiva:
- team picks handling for challenge header

### DIFF
--- a/src/components/Lend/LoanSearch/LoanSearchInterface.vue
+++ b/src/components/Lend/LoanSearch/LoanSearchInterface.vue
@@ -85,6 +85,7 @@
 					<team-picks-switch
 						v-if="showChallengeHeader"
 						class="tw-mb-2"
+						@handle-team-picks="handleTeamPicks"
 					/>
 					<loan-search-filter
 						:extend-flss-filters="extendFlssFilters"
@@ -142,6 +143,7 @@
 						:show-tags="true"
 						:enable-five-dollars-notes="enableFiveDollarsNotes"
 						:user-balance="userBalance"
+						:is-team-pick="showTeamPicks"
 						@add-to-basket="addToBasket"
 					/>
 				</div>
@@ -253,6 +255,7 @@ export default {
 			trackedHits: undefined,
 			userId: null,
 			userBalance: undefined,
+			showTeamPicks: false,
 		};
 	},
 	apollo: {
@@ -351,6 +354,11 @@ export default {
 				this.trackLoans();
 			}
 		});
+
+		// Verify challenge header and show team picks
+		if (this.showChallengeHeader) {
+			this.showTeamPicks = true;
+		}
 	},
 	computed: {
 		defaultPageLimit() {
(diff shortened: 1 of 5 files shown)


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -436,6 +444,9 @@ export default {
 		},
 		addToBasket(payload) {
 			if (payload.success) this.$kvTrackEvent('loan-card', 'add-to-basket', 'filter-page-new-card');
+		},
+		handleTeamPicks(payload) {
+			this.showTeamPicks = payload;
```
dyersituations, line 449: I don't think we want to set this to the payload here, since it's the toggle on whether to display the new icon.

src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -351,6 +354,11 @@ export default {
 				this.trackLoans();
 			}
 		});
+
+		// Verify challenge header and show team picks
+		if (this.showChallengeHeader) {
+			this.showTeamPicks = true;
```
dyersituations, line 360: Can we just use `showChallengeHeader` and remove `showTeamPicks`?

roger-in-kiva: This is the initial state for showTeamPicks param if we are showing the header.
About your comment above, I understand that we can remove recommended loans with the team picks switch; if not, should we block the switch to be always on if is the page is under a challenge?

src/components/Lend/LoanSearch/LoanSearchInterface.vue
dyersituations: @roger-in-kiva still curious if this is necessary.

roger-in-kiva: I guess that removing recommended loans with the team picks switch, accordingly with comment below, we should remove the label... makes sense?

dyersituations: Oh wait, is the payload here a boolean on whether to show the team picks?

roger-in-kiva: Oh yes, the default value is on if we show the challenge,
https://github.com/kiva/ui/blob/c80f11ea26927f992c1e0320fabb0cbf110b2e43/src/components/Lend/LoanSearch/TeamPicksSwitch.vue#L32

dyersituations: Sounds good. I was thinking this was getting loan data or some other data based on interaction with the team picks.

dyersituations (approved):

src/components/Lend/LoanSearch/LoanSearchInterface.vue
roger-in-kiva: As not having BE well defined yet, difficult to do that. We can have a next iteration on this if needed.

PR #5218, opened 2024-03-22
fix: ensure exporter is created

mcstover:
My theory here is that the check for the exporter type if failing with a js error b/c it wasn't set. Trying to fix the values the conditional is operating on first after which I'll also define a better default in flux.

Also wrapped the debug init in a flag I can set in flux if further debugging is needed...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
 				"@opentelemetry/api": "^1.7.0",
 				"@opentelemetry/auto-instrumentations-node": "^0.40.2",
 				"@opentelemetry/exporter-metrics-otlp-proto": "^0.46.0",
+				"@opentelemetry/exporter-trace-otlp-grpc": "^0.49.1",
 				"@opentelemetry/exporter-trace-otlp-proto": "^0.46.0",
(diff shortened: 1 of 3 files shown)


hobbsh:
What are you trying to accomplish?

mcstover:
> What are you trying to accomplish?

I don't think the UI traces are actually getting exported/activated from UI in development. I can see them fine using the console exporter locally, including trace ids on page request logs...

hobbsh:
> > What are you trying to accomplish?
> 
> I don't think the UI traces are actually getting exported/activated from UI in development. I can see them fine using the console exporter locally, including trace ids on page request logs...
> 
> I was also able to see the actual exporter active in Tilt applying similar updates.

Ok I recall this exporter being super verbose so it might not be a great idea to enable it in anything but local.

hobbsh:
I think the problem could be the exporter type - it should be GRPC, as long as these are not originating from the client. If they are originating from the client, we'll need to setup a special otel ingress to receive them.

mcstover:
This setup is only for server side to start with

eddieferrer (approved):

PR #5230, opened 2024-04-01
feat: show "added to cart" message for lend filter challenge

dyersituations:
https://kiva.atlassian.net/browse/ACK-1070

- New tooltip/message shown when a user ATB from lend/filter with new challenge variation

image

image

### DIFF
--- a/src/components/Lend/LoanSearch/ChallengeCallout.vue
+++ b/src/components/Lend/LoanSearch/ChallengeCallout.vue
@@ -5,27 +5,17 @@
 				<div class="tw-col-span-12 tw-w-full">
 					<div class="info tw-w-full">
 						<div class="tw-flex tw-gap-1 tw-items-center">
-							<div class="tw-shrink-0">
-								<img
+							<div class="tw-flex tw-shrink-0">
+								<kv-user-avatar
 									v-for="(p, i) in participants"
 									:key="p.id"
-									:src="p.image.url"
-									alt="Lender photo"
-									class="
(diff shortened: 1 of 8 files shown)


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -239,6 +239,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		showChallengeHeader: {
```
roger-in-kiva, line 242: Why do we need this instead of the computed?

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -1,15 +1,68 @@
 <template>
-	<div v-if="lenderName">
+	<div v-if="showAddedToCartMessage">
+		<kv-page-container class="container">
+			<kv-grid class="tw-grid-cols-12">
+				<div class="tw-col-span-12 tw-w-full">
+					<div class="info tw-w-full">
+						<div class="tw-flex tw-gap-1 tw-items-center">
+							<div class="tw-shrink-0">
+								<img
```
roger-in-kiva, line 9: Should we use KvUserAvatar here?
Looks like hiding users with no image...

dyersituations: True, I was just re-using code from another component. Changed to avatar component.

roger-in-kiva (approved):

PR #5234, opened 2024-04-02
feat: add fade out to activity feed on challenge header

christian14b:
image

### DIFF
--- a/src/components/Lend/LoanSearch/ChallengeHeader.vue
+++ b/src/components/Lend/LoanSearch/ChallengeHeader.vue
@@ -59,7 +59,7 @@
 			</div>
 
 			<div class="tw-relative tw-flex tw-items-center tw-px-0">
-				<div class="tw-absolute tw-right-0 tw-w-10 tw-h-8 tw--mr-0.5 loan-activity-overlay"></div>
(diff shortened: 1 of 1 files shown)


src/components/Lend/LoanSearch/ChallengeHeader.vue
```diff
@@ -154,6 +157,17 @@ export default {
...
 
 <style scoped lang="postcss">
+
+.loan-activity-overlay {
+	background: linear-gradient(90deg, rgba(245, 245, 245,0) 0%,rgba(245, 245, 245, 1) 90%);
+}
+
+@screen lg {
+	.loan-activity-overlay {
+		background: linear-gradient(90deg, rgba(245, 245, 245,0) 0%,rgba(245, 245, 245, 1) 70%);
```
dyersituations, line 167: Just out of curiosity, why is the slight opacity difference needed between desktop and mobile?

roger-in-kiva: Also verify this on tablet, we are using same condition for mobile but is larger than mobile and desktop of course.
Using a fixed size instead of a percentage for all cases could be a better approach maybe...

PR #5236, opened 2024-04-03
fix: copyUrl calling getFullUrl twice and mixing up different params

christian14b:
Currently, with `social-share-mixin` and `shareLink` computed definition, `getFullUrl` is being called twice and making some of the `shareLink` arguments invalid (Browser doesn't recognize them).

For instance, in `ShareChallenge` component when you hit the `Copy Link` button the value is: `https://www.development.kiva.org/lend/filter?team=aplus&lender=christian78848470?utm_source=social_share_link&utm_medium=referral&utm_campaign=teamschallenge&utm_content=christian78848470`. When you visit this link, it should show a callout at the top of the page due to lender param, but it only show the challenge header because of the double `?`.

### DIFF
--- a/src/util/urlUtils.js
+++ b/src/util/urlUtils.js
@@ -56,8 +56,11 @@ export function getFullUrl(base, args) {
 
 	// remove hash portion of url if present
 	const baseUrlWithoutHash = base.split('#')[0];
+	// remove params portion of url if present
+	const baseUrl = baseUrlWithoutHash.split('?')[0];
 	const hashParams = base.split('#')[1] ? `#${base.split('#')[1]}` : '';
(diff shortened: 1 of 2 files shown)


src/util/urlUtils.js
```diff
@@ -56,16 +56,24 @@ export function getFullUrl(base, args) {
 
 	// remove hash portion of url if present
 	const baseUrlWithoutHash = base.split('#')[0];
+	// remove params portion of url if present
+	const baseUrl = baseUrlWithoutHash.split('?')[0];
```
dyersituations, line 60: Have you tested this with the existing default share module?

PR #5237, opened 2024-04-03
feat: team picks unchecked by default

roger-in-kiva:
- team picks unchecked by default on filter page


src/components/Lend/LoanSearch/LoanSearchInterface.vue
```diff
@@ -309,9 +309,6 @@ export default {
 		} catch (e) {
 			logReadQueryError(e, 'LoanSearchInterface itemsInBasketQuery');
 		}
-
-		// Get challenge filters and loans
-		this.getChallengeFilters();
```
dyersituations, line 314: Will this still get called when team picks are enabled? Or do we need these filters already queried and ready?

roger-in-kiva: This is called when toggling the team picks switch here:
https://github.com/kiva/ui/blob/43a6d52d506c43314253383a44280b3dc9dc6383/src/components/Lend/LoanSearch/LoanSearchInterface.vue#L484

dyersituations (approved):

PR #5238, opened 2024-04-03
feat: add basket total to header for large baskets

dyersituations:
https://kiva.atlassian.net/browse/CORE-1811

- Ported the recent PR from CPS
- Fixed unrelated bug with corporate header login button wrapping
- Fixed unrelated error with other story

**Small basket:**
image
image

**Large basket:**
image
image

### DIFF
--- a/.storybook/stories/TheHeader.stories.js
+++ b/.storybook/stories/TheHeader.stories.js
@@ -31,13 +31,39 @@ const loggedIn = {
 
 const itemInCart = {
 	shop: {
-		nonTrivialItemCount: 1234,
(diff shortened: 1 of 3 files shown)


roger-in-kiva (approved):
Not fully related with this work, but should we show the $0 credit amount? Maybe making more room if we hide it.

dyersituations:
> Not fully related with this work, but should we show the $0 credit amount? Maybe making more room if we hide it.

Good question. That's existing functionality, so I'll leave for now, but we should discuss a different UI fallback.

mcstover (approved):

PR #5239, opened 2024-04-03
feat: emit tracking event when progress bar shows

christian14b:


dyersituations (approved):

src/components/Lend/LoanSearch/ChallengeHeader.vue
```diff
@@ -153,6 +153,11 @@ export default {
 			return this.challengeData?.participation ?? {};
 		}
 	},
+	mounted() {
+		this.$nextTick(() => {
```
roger-in-kiva, line 157: Wondering why nextTick?

roger-in-kiva (approved):

src/components/Lend/LoanSearch/ChallengeHeader.vue
dyersituations: Right, but why is it needed instead of just called in `mounted` directly?

PR #5240, opened 2024-04-04
feat: header showed for active user team challenge

roger-in-kiva:
- header showed for active user team challenge

### DIFF
--- a/src/pages/Lend/LoanSearchPage.vue
+++ b/src/pages/Lend/LoanSearchPage.vue
@@ -91,11 +91,11 @@ const SHOW_LOANS_ACTIVITY_FEED_EXP = 'filter_loans_activity_feed';
 
 const getHasEverLoggedIn = client => !!(client.readQuery({ query: hasEverLoggedInQuery })?.hasEverLoggedIn);
 
-const reduceAmountLentTeamArray = userTeams => {
-	return userTeams?.reduce((prev, current) => {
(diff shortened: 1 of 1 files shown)


src/pages/Lend/LoanSearchPage.vue
```diff
@@ -301,14 +317,25 @@ export default {
...
 				const userTeams = userTeamsData.my?.teams?.values ?? [];
+				const activeGoals = goalsData?.goals?.values ?? [];
 
-				let maxAmountLentTeam = {};
+				let sortedAmountLentTeams = [];
+				let userChallengeTeam = {};
 				if (userTeams.length > 0) {
-					maxAmountLentTeam = reduceAmountLentTeamArray(userTeams);
+					sortedAmountLentTeams = sortAmountLentTeamArray(userTeams);
+					for (let index = 0; index < sortedAmountLentTeams.length; index += 1) {
```
dyersituations, line 328: This duplicate code can be moved into a shared method, like we have for `sortAmountLentTeamArray`. It looks like this version is slightly safer than the one above, since it checks `goal` as well on line 331.

roger-in-kiva: Oh yeah, I missed that check there...
Sounds good having a shared method, let's refactor this.

roger-in-kiva: Done!

dyersituations (approved):

PR #5241, opened 2024-04-04
feat: track when challenge header appears on ty page

christian14b:
Same tracking when progressbar shows in `lend/filter` but for sharing module in thanks page

### DIFF
--- a/src/components/Thanks/ChallengeHeader.vue
+++ b/src/components/Thanks/ChallengeHeader.vue
@@ -86,11 +86,20 @@ export default {
 			type: String,
 			default: '',
 		},
+		teamName: {
+			type: String,
+			default: '',
+		},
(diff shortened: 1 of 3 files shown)


src/components/Thanks/ChallengeHeader.vue
```diff
@@ -86,11 +86,20 @@ export default {
...
+			default: '',
+		},
 	},
 	computed: {
 		teamChallengePath() {
 			return this.teamPublicId ? `/team/challenge/${this.teamPublicId}` : '/teams';
 		},
 	},
+	mounted() {
+		this.$nextTick(() => {
```
dyersituations, line 100: Why is next tick necessary?

dyersituations (approved):

PR #5243, opened 2024-04-04
fix: revert changes for challenge header and apply them in ShareChall…

christian14b:
…enge component. Plus, rounding percentage value.

### DIFF
--- a/src/components/Thanks/ShareChallenge.vue
+++ b/src/components/Thanks/ShareChallenge.vue
@@ -194,7 +194,7 @@ export default {
 				return sum + (value?.amountLent ?? 0);
 			}, 0) ?? 0;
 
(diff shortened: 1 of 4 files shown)


src/components/Thanks/ShareChallenge.vue
```diff
@@ -194,7 +194,7 @@ export default {
 				return sum + (value?.amountLent ?? 0);
 			}, 0) ?? 0;
 
-			return (fundedAmount / targetAmount) * 100 || 0;
+			return Math.round((fundedAmount / targetAmount) * 100) || 0;
```
dyersituations, line 197: This won't work in the 99.8% situation. Please add a check to round down in that situation, if we keep rounding in.

dyersituations (approved):

PR #5249, opened 2024-04-08
feat: sticky tooltip on lend filter and copy change

christian14b:
image

### DIFF
--- a/src/components/Lend/LoanSearch/ChallengeCallout.vue
+++ b/src/components/Lend/LoanSearch/ChallengeCallout.vue
@@ -21,7 +21,7 @@
 								<span class="tw-whitespace-nowrap">Added to cart!</span>
 								<span :class="{'tw-whitespace-nowrap': borrowerName}">{{ participantsMessage }}</span>
 								<span
-									v-if="borrowerName"
+									v-if="participants.length > 1 && borrowerName"
 									class="data-hj-suppress tw-whitespace-nowrap"
 								>{{ borrowerName }}.</span>
 								<a


roger-in-kiva (approved):
Looks good!
Should we add a way to close it?

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -109,6 +106,7 @@ export default {
 	data() {
 		return {
 			mdiArrowTopRight,
+			participantsMessage: 'You are on your way to supporting the team challenge!'
```
dyersituations, line 109: This message is only when there are no other participants. So this message should be in the else of the tertiary that used to be in `participantsMessage`

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -1,8 +1,9 @@
 <template>
-	<www-page id="lend-filter">
+	<www-page id="lend-filter" class="tw-bg-secondary">
 		<challenge-callout
 			v-if="showChallengeHeader && !!teamData"
-			class="tw-bg-secondary tw-pb-1.5"
+			:class="{'tw-sticky tw-top-0 tw-z-sticky' : showAddedToCartMessage}"
+			class="tw-pb-1.5"
```
dyersituations, line 6: Did this solve the Safari issue mentioned in the ticket?

dyersituations (changes requested):
I think it's OK not being dismissible, since this is meant to be a CTA of sorts.

@christian14b does it sit above or below the header?

src/components/Lend/LoanSearch/ChallengeCallout.vue
dyersituations: I think your latest changes are close, but I'm guessing we only want to show the borrower name if there are more than one participant, since the new message probably doesn't make sense with the borrower name just at the end.

dyersituations:
@christian14b also note that the tests failed.

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -19,10 +19,11 @@
 							</div>
 							<div class="tw-flex tw-gap-0.5 tw-flex-wrap">
 								<span class="tw-whitespace-nowrap">Added to cart!</span>
-								<template v-if="borrowerName">
-									<span class="tw-whitespace-nowrap">{{ participantsMessage }}</span>
-									<span class="data-hj-suppress tw-whitespace-nowrap">{{ borrowerName }}.</span>
-								</template>
+								<span :class="{'tw-whitespace-nowrap': borrowerName}">{{ participantsMessage }}</span>
+								<span
+									v-if="borrowerName"
```
dyersituations, line 24: I think we also need to check `this.participants.length > 1`, otherwise the message might not make sense when there's one participant (borrower name just sitting after the `!` in the new message.

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -1,8 +1,9 @@
 <template>
-	<www-page id="lend-filter">
+	<www-page id="lend-filter" class="tw-bg-secondary">
```
dyersituations, line 2: Does this change the appearance of the page?

dyersituations (approved):

mcstover:
> image

@christian14b @dyersituations If this floats over the interface controls and loan cards, it should be dismmissable. What if they are trying to add other loans or use the filter interface?

dyersituations:
> @christian14b @dyersituations If this floats over the interface controls and loan cards, it should be dismmissable. What if they are trying to add other loans or use the filter interface?

Yeah @roger-in-kiva recommended that as well, but we're operating with zero design support at the moment. I'll create a ticket to add a dismiss button.

PR #5251, opened 2024-04-09
feat: use default invite url in Share Challenge component

christian14b:


src/components/Thanks/ShareChallenge.vue
```diff
@@ -209,7 +209,11 @@ export default {
 				team: this.teamPublicId,
 				lender: this.lender?.publicId ? `${this.lender.publicId}` : '',
 			};
-			return getFullUrl(`${base}/lend/filter`, args);
+			let invitedBy = '';
+			if (this.loan.id && this.lender?.inviterName) {
```
dyersituations, line 213: Just to check, are these two values defined like expected for this challenge header?

dyersituations (approved):
@christian14b make sure to test the invite count on dev once this is deployed.

PR #5253, opened 2024-04-10
feat: goal mixin updated

roger-in-kiva:
- goal mixin updated
- goal mixin implemented to related challenge components

### DIFF
--- a/test/unit/specs/components/Thanks/ChallengeHeader.spec.js
+++ b/test/unit/specs/components/Thanks/ChallengeHeader.spec.js
@@ -49,7 +49,7 @@ describe('ChallengeHeader', () => {
 		const LoansFunded = getByText('1/2 loans funded');
 		expect(getByTestId('loans-funded')).toBe(LoansFunded);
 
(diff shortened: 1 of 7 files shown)


test/unit/specs/components/Thanks/ChallengeHeader.spec.js
```diff
@@ -49,7 +49,7 @@ describe('ChallengeHeader', () => {
 		const LoansFunded = getByText('1/2 loans funded');
 		expect(getByTestId('loans-funded')).toBe(LoansFunded);
 
-		const PercentageFunded = getByText('50% complete');
+		const PercentageFunded = getByText('0% complete');
```
dyersituations, line 52: Shouldn't this be "50%" if 1/2 the funds are funded, like mentioned here on line 49?

dyersituations: Oh right, this is the old component that is no longer active.

dyersituations (approved):

PR #5254, opened 2024-04-10
fix: fe showing empty tooltip when back from filter page

dyersituations:
https://kiva.atlassian.net/browse/MP-91

- Added simple check to make sure tooltip doesn't show empty message
- Added missing ID from gql query (required to commit)


roger-in-kiva (approved):
Thanks!

emuvente (approved):

PR #5255, opened 2024-04-10
feat: place callout on top of borrower profile

christian14b:
## Desktop 
User invite bubble
Screenshot 2024-04-10 at 15 17 31

## Mobile 
User invite bubble
Screenshot 2024-04-10 at 15 19 08

### DIFF
--- a/src/pages/BorrowerProfile/BorrowerProfile.vue
+++ b/src/pages/BorrowerProfile/BorrowerProfile.vue
@@ -517,15 +517,14 @@ export default {
 						client.query({ query: experimentAssignmentQuery, variables: { id: SHARE_LANGUAGE_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: FIVE_DOLLARS_NOTES_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: EDUCATION_PLACEMENT_EXP } }),
-						teamPublicId,
 						activeChallengeHeaderExp,
 						teamPublicId && activeChallengeHeaderExp
 							? client.query({ query: TeamInfoFromId, variables: { team_public_id: teamPublicId } })
 							: null,
 					]);
 				}).then(response => {
-					const teamId = response[5]?.data?.community?.team?.id ?? null;
-					const activeChallengeHeaderExp = response[4];
+					const teamId = response[4]?.data?.community?.team?.id ?? null;
+					const activeChallengeHeaderExp = response[3];
 					const lenderPublicId = route?.query?.lender ?? '';
 
 					if (teamId && activeChallengeHeaderExp) {


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -613,6 +613,7 @@ export default {
 
 				this.$showTipMsg(this.errorMsg, 'error');
 			});
+			this.$emit('add-to-basket', { loanId: this.loanId, lendAmount: this.selectedOption });
```
dyersituations, line 616: As mentioned in the ticket, this work is just to add a bubble to the BP when a user accepts an invite. So I think a lot of this code and extra querying in BP component can be removed.

We only need to show the bubble when the user accepts an invite, and it should take the form of the screenshot in the ticket.

The idea here is that it's a fix for allowing the invite process to be counted, but also show the user that they accepted an invite to a challenge. We don't want to add too much extra page load to the BP page.

mcstover: There does seem to be a lot of extra info required for this. Can we move it out of the prefetch and into it's on method within mounted instead?

dyersituations: There's a good chance we don't need most of this extra info. But agreed that using mounted for this is a good approach.

dyersituations, line 616: Can this line be removed now?

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -138,6 +138,9 @@ export default {
 				? `You & ${this.participants.length - 1} other members are supporting`
 				: 'You are on your way to supporting the team challenge!';
 		},
+		teamLink() {
+			return `/lend/filter?team=${this.$route.query.team}`;
```
dyersituations, line 142: Does this need to be `this.$route?.query?.team`?

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -138,6 +138,9 @@ export default {
 				? `You & ${this.participants.length - 1} other members are supporting`
 				: 'You are on your way to supporting the team challenge!';
 		},
+		teamLink() {
+			return `/lend/filter?team=${this.$route?.query?.team ?? ''}`;
```
dyersituations, line 142: Can we use the `teamName` prop here instead? It's also a required component.

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -123,8 +123,8 @@ export default {
 		},
 		headerCallout() {
 			return this.shareLenderName
-				? `Support ${this.shareLenderName} and help ${this.teamName} hit their goal`
-				: `Help ${this.teamName} hit their goal`;
+				? `Support ${this.shareLenderName} and help `
+				: 'Help ';
```
dyersituations, line 127: Does this trailing space create a double space in the message since a space also exists in the HTML?

test/unit/specs/components/Lend/LoanSearch/ChallengeCallout.spec.js
```diff
@@ -4,14 +4,22 @@ import ChallengeCallout from '@/components/Lend/LoanSearch/ChallengeCallout';
...
 describe('ChallengeCallout', () => {
-	it('should display share lender in message', () => {
+	it('should display team name', () => {
 		const { getByText } = render(ChallengeCallout, {
-			props: { shareLender, teamName }
+			props: { shareLender, teamName },
+			mocks,
 		});
 
-		getByText(`Support ${shareLender.name} and help ${teamName} hit their goal`);
```
dyersituations, line 14: Don't we still want to check the entire message? The lender name should still be displayed.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -437,64 +452,107 @@ export default {
...
-				.then(({ data }) => {
-					const expCookieSignifier = cookieStore.get('kvlendborrowerbeta');
-					if (expCookieSignifier === 'a' || expCookieSignifier === 'c') {
-						const { query } = route;
-						return Promise.reject({
-							path: `/lend-classic/${route.params.id}`,
-							query,
-						});
-					}
+			return Promise.all([
```
dyersituations, line 465: `Promise.all` not needed here anymore. Reverting that might make this diff slightly easier to read.

src/components/Lend/LoanSearch/ChallengeCallout.vue
```diff
@@ -45,7 +45,7 @@
 			</kv-grid>
 		</kv-page-container>
 	</div>
-	<div v-else-if="shareLenderName">
+	<div v-else>
```
dyersituations, line 48: It looks like this check can be moved to line 56. The `img` tag uses the name for the `alt`.

```
v-if="shareLenderImage && shareLenderName"
```

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -489,11 +507,49 @@ export default {
...
+						teamPublicId,
+						activeChallengeHeaderExp,
 					]);
+				}).then(response => {
+					const teamPublicId = response[3];
+					const activeChallengeHeaderExp = response[4];
+
+					let teamDataPromise = Promise.resolve();
+					if (teamPublicId && activeChallengeHeaderExp) {
+						teamDataPromise = client.query({ query: TeamInfoFromId, variables: { team_public_id: teamPublicId } }); // eslint-disable-line max-len
```
dyersituations, line 530: Can't this query can be ran in the `Promise.all` on line 517? I don't see anything that this `.then` needs from that `Promise.all`.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -489,11 +507,49 @@ export default {
...
+					}
+
+					return Promise.all([
+						teamDataPromise,
+						activeChallengeHeaderExp,
+					]);
+				}).then(responseTeams => {
+					const teamId = responseTeams[0]?.data?.community?.team?.id ?? null;
+					const activeChallengeHeaderExp = responseTeams[1];
+					const lenderPublicId = route?.query?.lender ?? '';
```
dyersituations, line 540: This already is saved to `getPublicId`

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -718,6 +777,40 @@ export default {
...
+			if (teamPublicId) {
+				const teamInfo = this.apollo.readQuery({ query: TeamInfoFromId, variables: { team_public_id: teamPublicId } }); // eslint-disable-line max-len
+				teamId = teamInfo?.community?.team?.id ?? null;
+			}
+			if (teamId) {
+				const teamData = this.apollo.readQuery({ query: TeamInfoFromId, variables: { team_id: teamId } });
+				this.teamData = teamData?.community?.team || {};
+
+				try {
+					const data = this.apollo.readQuery({
```
dyersituations, line 799: This is the only `readyQuery` here within a try catch. I don't think `readQuery` is supposed to throw an exception, but I also know that we often wrap `readQuery` in a try catch. It would be worth investigating to see if an exception is thrown. If not, this try catch can be removed. If so, then the other `readQuery` usages in this method also need a try catch.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -489,11 +506,39 @@ export default {
...
+						id: `Experiment:${CHALLENGE_HEADER_EXP}`,
+						fragment: experimentVersionFragment,
+					}) ?? {};
+					const activeChallengeHeaderExp = challengeHeaderExpData?.version === 'b';
+
 					return Promise.all([
 						client.query({ query: experimentAssignmentQuery, variables: { id: SHARE_LANGUAGE_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: FIVE_DOLLARS_NOTES_EXP } }),
 						client.query({ query: experimentAssignmentQuery, variables: { id: EDUCATION_PLACEMENT_EXP } }),
+						teamPublicId,
```
dyersituations, line 520: It looks like this is no longer used in the `.then` and can be removed.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -489,11 +506,39 @@ export default {
...
+						teamPublicId && activeChallengeHeaderExp
+							? client.query({ query: TeamInfoFromId, variables: { team_public_id: teamPublicId } })
+							: null,
 					]);
+				}).then(response => {
+					const teamId = response[5]?.data?.community?.team?.id ?? null;
+					const activeChallengeHeaderExp = response[4];
+					const lenderPublicId = route?.query?.lender ?? '';
+
+					if (teamId && activeChallengeHeaderExp) {
```
dyersituations, line 531: This should also check `lenderPublicId`

dyersituations (approved):

PR #5256, opened 2024-04-10
feat: remove active login check at checkout MP-87

emuvente:
We have better security measures than this already in-place, and this just causes friction. A similar change is being made in the api, and this will have to wait for that change to be deployed before this is merged.

### DIFF
--- a/src/components/Checkout/InContext/InContextCheckout.vue
+++ b/src/components/Checkout/InContext/InContextCheckout.vue
@@ -57,10 +57,10 @@
 		<div
 			class="in-context-login"
 			:class="{ 'tw-text-right' : !isCorporateCampaign }"
-			v-if="!isActivelyLoggedIn && !promoGuestCheckoutEnabled"
+			v-if="!isLoggedIn && !promoGuestCheckoutEnabled"
 		>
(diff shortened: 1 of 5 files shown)


mcstover (approved):
Oh my, thank you for working through this!

dyersituations (approved):

mcstover:
It appears we do this in [CC pages too](https://github.com/kiva/ui/blob/main/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue#L892), any thoughts on if we should also remove here? Seems like it couldn't really hurt...I can create a maintenance task in CIT if need be.

emuvente:
> It appears we do this in [CC pages too](https://github.com/kiva/ui/blob/main/src/pages/LandingPages/CorporateCampaign/CCLandingPage.vue#L892), any thoughts on if we should also remove here? Seems like it couldn't really hurt...

Yeah, I'll do that here too, thanks for pointing that out @mcstover !

dyersituations (approved):

mcstover (approved):
Thank you for making the updates in the CC/in-context sections too!

PR #5257, opened 2024-04-12
feat: make challenge bubble dismissable

christian14b:
Added MP-103 work

Screenshot 2024-04-12 at 8 11 05 AM

Screenshot 2024-04-12 at 8 11 14 AM

### DIFF
--- a/src/components/Lend/LoanSearch/ChallengeCallout.vue
+++ b/src/components/Lend/LoanSearch/ChallengeCallout.vue
@@ -41,7 +41,7 @@
 							</div>
 							<button
 								class="tw-flex"
-								@click="$emit('hide')"
+								@click="$emit('close')"
 							>
 								<kv-material-icon
 									class="tw-h-3 tw-w-3"
@@ -54,7 +54,7 @@
 			</kv-grid>
 		</kv-page-container>
 	</div>
-	<div v-else>
+	<div v-else-if="!hideMsg">
 		<kv-page-container class="container">
 			<kv-grid
 				class="tw-grid-cols-12"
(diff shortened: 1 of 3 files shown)


roger-in-kiva:
Looks like this is pushing down the page. Could we have a toast behavior, floating over page?
Page is jumping up when bubble is closed.

mcstover:
> Looks like this is pushing down the page. Could we have a toast behavior, floating over page? Page is jumping up when bubble is closed.

I'm also confused as to why we didn't use our global toast component here which can be set to remain visible as well as contain custom html... is it just b/c of the secondary icon injection at the end? If so, please work with design to remove that...

roger-in-kiva:
> > Looks like this is pushing down the page. Could we have a toast behavior, floating over page? Page is jumping up when bubble is closed.
> 
> 
> 
> I'm also confused as to why we didn't use our global toast component here which can be set to remain visible as well as contain custom html... is it just b/c of the secondary icon injection at the end? If so, please work with design to remove that...

Agreed! I think we can also pass the secondary icon as custom toast content template...

roger-in-kiva:
> @roger-in-kiva @mcstover I think the main problem will be moving the user's avatar inside a template. I'm not sure how to add a custom component here. Is there any example of this in the past?

I guess you can follow this: https://github.com/kiva/kv-ui-elements/blob/5e7199e37777f95d09134f1e25745472896021e4/%40kiva/kv-components/vue/stories/KvToast.stories.js#L99

As global toast component is dismissable, we can close this PR though.

mcstover:
@christian14b @roger-in-kiva It looks like KvToast supports a slot so which is what I think Roger was pointing out...

PR #5263, opened 2024-04-24
feat: deposit incentive banner component and story

christian14b:
`amountToLend` field based on https://github.com/kiva/kiva/pull/12464

No balance variant:

image
With Balance variant:

image

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
+++ b/src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
@@ -34,7 +34,6 @@ import GenericPromoBanner from '@/components/WwwFrame/PromotionalBanner/Banners/
 import DonationBannerContainer from '@/components/WwwFrame/PromotionalBanner/Banners/Donation/DonationBannerContainer';
 import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import DepositIncentiveBanner from '@/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner';
-import experimentAssignmentQuery from '@/graphql/query/experimentAssignment.graphql';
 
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -145,5 +156,15 @@ export default {
 			}
 		}
 	},
+	mounted() {
```
emuvente, line 159: This could be done in `created` instead of `mounted` so that the banner would render on the server too

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -145,5 +156,15 @@ export default {
...
+		const { version } = this.apollo.readFragment({
```
emuvente, line 160: The experimentAssignment query will need to be called during prefetch for this to work

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -142,8 +154,23 @@ export default {
 						this.isPromoEnabled = true;
 					}
 				}
+
+				return this.apollo.query({
```
emuvente, line 158: Since this is in the result function it won't finish before `created` is called. The experiment assignment could also be fetched by adding the experiment field to the `bannerQuery` above:
```graphql
query bannerQuery {
	contentful {
		entries(contentType: "uiSetting", contentKey: "ui-global-promo")
	}
	experiment(id: "deposit_incentive_banner") @client {
		id
		version
	}
}
```

dyersituations (approved):

emuvente (approved):
Looks great!

PR #5290, opened 2024-04-29
feat: make sure redirect after passwordless login goes to the appropriate impact dashboard

michelleinez:


server/auth-router.js
```diff
@@ -101,6 +101,13 @@ module.exports = function authRouter(config = {}) {
 		if (req.query.loginHint) {
 			options.login_hint = req.query.loginHint;
 		}
+
+		// Specify partnerContentId & set default doneUrl in case one isn't set:
+		if (req.query.partnerContentId) {
+			options.partnerContentId = req.query.partnerContentId;
+			options.doneUrl = `impact-dashboard/${req.query.partnerContentId}`;
```
dyersituations, line 108: On a few lines below, we use `req.session.doneUrl`. Do we want to do the same here?

mcstover: @michelleinez  We do not need to handle the `doneUrl` in this location as it's already handled. Rather, we'll be "defining" and setting the `doneUrl` in our impact dashboard share links. What we define there will be handled by the code below.

mcstover: With that in mind, I think this one can be closed as you're already handling the `partnerContentId` option.

PR #5294, opened 2024-05-01
fix: stop adding uiab header to api calls pending header approval

emuvente:
Seeing a CORS failure on all api calls in prod when uiab cookie is present, because the x-uiab header is not is the allowed list of headers. commenting out this code until we can sort that out.


dyersituations (approved):

mcstover (approved):
The alternative would be to add that allowed header right? Wondering if that is in flux yet...or if it's still hard-coded in the gateway...

emuvente:
> The alternative would be to add that allowed header right? Wondering if that is in flux yet...or if it's still hard-coded in the gateway...

@mcstover yes that's right, but I didn't want to block any users while we figured out what needs to be updated.

PR #5295, opened 2024-05-01
feat: add exp tracking to users that see the deposit incentive banner

christian14b:
Fire event only when banner is visible


dyersituations (approved):

emuvente (approved):

emuvente (changes requested):
We should track it when it's not visible too. We should track whenever we have an experiment assignment

src/components/WwwFrame/PromotionalBanner/GlobalPromotionalBannerContentful.vue
```diff
@@ -166,6 +166,14 @@ export default {
 			fragment: experimentVersionFragment,
 		}) ?? {};
 
+		if (version) {
```
dyersituations, line 169: Look to the usage of the `trackExperimentVersion` method.

emuvente (approved):

PR #5296, opened 2024-05-01
feat: add progress bar to incentive banner exp

christian14b:
image

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
@@ -47,17 +47,10 @@ export default {
 	apollo: {
 		query: amountToLendQuery,
 		preFetch: true,
-		variables() {
-			return {
-				basketId: this.cookieStore.get('kvbskt')
-			};
-		},
 		result({ data }) {
 			this.amountToLend = data?.my?.depositIncentiveAmountToLend ?? 0;
 			this.isLoggedin = !!data?.my?.id ?? false;
-			this.basketTotal = data.shop?.basket?.items?.values?.reduce((sum, item) => {
-				return sum + +(item?.price ?? 0);
-			}, 0) ?? 0;
+			this.basketTotal = parseFloat(data.shop?.basket?.totals?.loanReservationTotal ?? 0);
 		},
 	},
 	computed: {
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
```diff
@@ -27,15 +40,24 @@ export default {
...
 		preFetch: true,
+		variables() {
+			return {
+				basketId: this.cookieStore.get('kvbskt')
+			};
+		},
 		result({ data }) {
 			this.amountToLend = data?.my?.depositIncentiveAmountToLend ?? 0;
 			this.isLoggedin = !!data?.my?.id ?? false;
+			this.basketTotal = data.shop?.basket?.items?.values?.reduce((sum, item) => {
```
emuvente, line 58: This should use the loan reservation total like you do below. However, that total is a string, so it will need to be parsed into a number first:
```
this.basketTotal = parseFloat(data.shop?.basket?.totals?.loanReservationTotal ?? 0);
```

src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
```diff
@@ -27,15 +40,24 @@ export default {
...
+			basketTotal: 0,
 		};
 	},
 	inject: ['apollo', 'cookieStore'],
 	apollo: {
 		query: amountToLendQuery,
 		preFetch: true,
+		variables() {
+			return {
+				basketId: this.cookieStore.get('kvbskt')
```
emuvente, line 52: The basket id variable is added to all queries automatically, so this isn't needed.

src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
```diff
@@ -55,5 +77,29 @@ export default {
 			};
 		}
 	},
+	methods: {
+		activateBasketWatchQuery() {
```
emuvente, line 81: The apollo plugin does this watch automatically, so this isn't needed. Any time a different query gets a new result for a field that is used in the `amountToLendQuery` the `result` function above will be called with the updated value.

emuvente (approved):

PR #5301, opened 2024-05-07
feat: content changes to registration form for passwordless case

michelleinez:


dyersituations (approved):

roger-in-kiva (approved):

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -48,10 +48,10 @@
 					v-model="newAcctTerms"
 					:validation="$v.newAcctTerms"
 				>
-					I have read and agree to the Kiva
+					I have read and agree to the {{ !passwordless ? 'Kiva' : '' }}
 					<a href="/legal/terms" target="_blank">Terms of Use</a>
 					and
-					<a href="/legal/privacy" target="_blank">Privacy Policy</a>
+					<a href="/legal/privacy" target="_blank">Privacy Policy</a> {{ !passwordless ? '' : '(required)' }}
```
mcstover, line 54: I don't think we should remove this b/c the field is required and we need to state that.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -63,7 +63,11 @@
 					v-show="needsNews"
 					v-model="newsConsent"
 				>
-					I want to receive updates about my loans, Kiva news, and promotions in my inbox
+					{{ !passwordless
+						? 'I want to receive updates about my loans, Kiva news, and promotions in my inbox'
+						: `Receive email updates from Kiva (including borrower updates and promos).
+							You can unsubscribe anytime. (optional)`
```
mcstover, line 69: Again, I think we need to keep the `You can unsubscribe anytime. (optional)` portion of the this copy.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -48,10 +48,10 @@
 					v-model="newAcctTerms"
 					:validation="$v.newAcctTerms"
 				>
-					I have read and agree to the Kiva
+					I have read and agree to the {{ !passwordless ? 'Kiva' : '' }}
```
mcstover, line 51: This looks like it should be a global change for all users.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -136,10 +140,14 @@ export default {
...
 			newsConsent: false,
 			showSsoTerms: false,
+			passwordless: false,
 		};
 	},
 	computed: {
 		registrationMessage() {
+			if (this.passwordless) {
+				return '';
+			}
```
mcstover, line 150: I don't think we can bypass any of this...it looks related to the validations we need to have in place...Also, we won't be able to control the captcha aspect here so that def needs to stay in place...

src/pages/LoginAndRegister/RegisterSocial.vue
mcstover: Also, this is an addition. I think it should be global.

src/pages/LoginAndRegister/RegisterSocial.vue
mcstover: Oh wait, this is a new addition. Sorry.

mcstover: We probably should add the `You can unsubscribe anytime. (optional)` portion to the existing copy as well.

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: reinstated / done

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: won't do since comment sounds like it was made in error

mcstover (approved):

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: done

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: done

src/pages/LoginAndRegister/RegisterSocial.vue
mcstover: I suspect the additional copy is coming from Legal recommendations which is why I think we might want to always show for all users.

PR #5302, opened 2024-05-08
feat: activity support on filter page

roger-in-kiva:
- URL and filter pills support added for activities on filter page

### DIFF
--- a/src/util/loanSearch/queryParamUtils.js
+++ b/src/util/loanSearch/queryParamUtils.js
@@ -10,7 +10,6 @@ import filterConfig from '@/util/loanSearch/filterConfig';
 export function hasExcludedQueryParams(query) {
 	// Handle temporary query param exclusions
 	const excludedParams = [
-		'activity',
 		'city_state',
 		'loanTags',
 		'state',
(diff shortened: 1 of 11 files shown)


src/util/loanSearch/queryParamUtils.js
```diff
@@ -10,7 +10,6 @@ import filterConfig from '@/util/loanSearch/filterConfig';
 export function hasExcludedQueryParams(query) {
 	// Handle temporary query param exclusions
 	const excludedParams = [
-		'activity',
```
dyersituations, line 13: I don't think we can change this exclusion, since it will affect the search bar navigation for users not in the experiment. For the experiment, we should probably replicate this method for the filter page.

roger-in-kiva: Nice call...

roger-in-kiva: On this particular case, we are not showing Activities in search bar results, even that, with activities support on filter page, we should be able to show them and don't redirect to legacy filters that is the case we are currently using this function. thoughts?

roger-in-kiva: That's right, we are not showing results in the search bar with activities. Also, the usage to exclude that query param is to redirect to legacy filter if activity included, a  feature we are enabling now on filter page.

dyersituations: Sounds good. I was suggesting to create a new version of this method for the experiment, so it would help with new navigation, but if there's no search with activities, we're good. We just need to make sure to not affect existing functionality, and it sounds like we're good here.

dyersituations (approved):

PR #5306, opened 2024-05-08
feat: logo rendering for passwordless reg form + styles

michelleinez:


src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -221,7 +259,27 @@ export default {
 				this.$kvTrackEvent('Register', 'error-register-social-form-invalid-input');
 			}
 		},
+		fetchStrategicPartnerLoginInfoByPageId() {
+			this.apollo.query({
+				query: StrategicPartnerLoginInfoByPageId,
+				variables: { pageId: this.partnerContentId }
+			}).then(response => {
+				const { logo } = response.data.strategicPartnerLoginInfoByPageId;
```
dyersituations, line 267: Is `logo` guaranteed to be defined? Do we need optional chaining here and below where `logo` is used?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -221,7 +259,27 @@ export default {
...
+			});
+		}
 	}
 
 };
 </script>
+
+<style scoped>
+.logo {
+	box-shadow: 0 0 18px rgba(0, 0, 0, 0.2);
```
dyersituations, line 283: Can we use our pre-defined shadows in TW?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -208,6 +238,14 @@ export default {
 			this.needsNews = true;
 		}
 	},
+	mounted() {
+		if (this.passwordless && this.partnerContentId) {
+			this.fetchStrategicPartnerLoginInfoByPageId();
```
dyersituations, line 243: Do we want to prefetch this data? Or are we not worried about image pop-in?

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: None of them look right for this design.  They don't provide adequate contrast against the black and make the logos look washed out.

michelleinez: Sara gave the shadow a deeper color for this case in figma than anything supported by tailwind.

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: The whole component won't show up until the data is available.

dyersituations: That means the logos will pop-in and move the page elements around? Or is there no content below this in this use case?

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: Optional chaining added.  Done

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -221,7 +262,28 @@ export default {
...
+		fetchStrategicPartnerLoginInfoByPageId() {
+			this.apollo.query({
+				query: StrategicPartnerLoginInfoByPageId,
+				variables: { pageId: this.partnerContentId }
+			}).then(res => {
+				const spLoginInfo = res?.data?.strategicPartnerLoginInfoByPageId;
+				const logo = spLoginInfo?.contentful?.entry?.fields?.primaryLogo;
+				this.fetchedLogo = {
+					url: logo?.fields?.file?.url || '',
+					alt: logo?.fields.title || '',
```
dyersituations, line 274: Does this need to be `logo?.fields?.title`?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -1,6 +1,31 @@
...
+							tw-border-white tw-border-4 logo"
+					>
+						<img
+							v-if="strategicPartnerLogo"
+							:src="strategicPartnerLogo.url"
+							:alt="strategicPartnerLogo.alt"
+							class="tw-w-full tw-h-full tw-object-contain"
+						>
+					</div>
+					<div
```
mcstover, line 18: We could keep the Kiva logo if the partner is missing. That would make this more reusable later...

src/pages/LoginAndRegister/RegisterSocial.vue
mcstover: you could try to do this at the bottom of created...may need to wrap in nextTick if the passwordless and partnerContentId aren't getting set quickly enough...

michelleinez: Actually yeah you're right the logos are popping in.
Prefetch implemented.  Done.

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: done

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: done

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -221,7 +277,28 @@ export default {
 				this.$kvTrackEvent('Register', 'error-register-social-form-invalid-input');
 			}
 		},
+		fetchStrategicPartnerLoginInfoByPageId() {
```
mcstover, line 280: Should be able to remove this and the call in mounted now...

michelleinez: Ack.  I thought it was the prefetch that was successfully fetching the logo but it looks like it is still this function.  When I delete it, it seems like the graphql call doesn't even get made.  I'm not sure why.

mcstover: YOu need a "result(result) {}" block after your prefetch function where you can set values on `this`. `this` cannot be set in the direct response/callback of the prefetch query.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -208,6 +242,28 @@ export default {
...
+				const spLoginInfo = res?.data?.strategicPartnerLoginInfoByPageId;
+				const logo = spLoginInfo?.contentful?.entry?.fields?.primaryLogo;
+				this.fetchedLogo = {
+					url: logo?.fields?.file?.url || '',
+					alt: logo?.fields?.title || '',
+				};
+			}).catch(error => {
+				console.error('Error fetching strategic partner info:', error);
+			});
+		}
```
mcstover, line 265: Need to add result block here
```
result(result) {
  
}
```

src/pages/LoginAndRegister/RegisterSocial.vue
michelleinez: Does this look right to you?  I get a 500 server error when I do this.
image

michelleinez: image
This too

michelleinez: This also doesn't work
image

michelleinez: oh, it looks like this.partnerContentId can't be accessed.

mcstover: Oh.. also, looks like you need a prefechVariables block... here's a good example: https://github.com/kiva/ui/blob/main/src/pages/BorrowerProfile/BorrowerProfile.vue#L517

mcstover: you'll prob want the `variables` block too. prefetchVariables are for the server fetch before the vue functionality is initialized and variables are in the clicet for an in-app navigation...

michelleinez: 500 here too 
image

michelleinez: pushed this latest non-working version just now.

mcstover: Can't use `this.` in the variables. you'll need to pass in the `{ route }` object to both variable blocks and extract the partnerContentId directly from the route.

michelleinez: okay so this makes the graphql fetch call, but variables are not set for the query.

image


Then this following version results in an error that says [Vue warn]: Error in created hook: "TypeError: Cannot read properties of undefined (reading 'route')"

image

And then this appears to make no graphql call at all.
image

michelleinez: In LoanStory.vue I see an implementation here that uses route in the preFetchVariables and uses this for variables.
image

So it seems like that's possible to do, but not sure then why the variable isn't getting set in the first of the three.

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -1,6 +1,31 @@
 <template>
 	<system-page>
 		<div class="page-content" style="max-width: 20rem;">
+			<div v-if="passwordless">
+				<div class="tw-flex tw-justify-center tw-items-center tw-relative tw-mb-2">
+					<div
+						v-if="fetchedLogoUrl"
```
mcstover, line 7: Moved this up. So now, along with the guards for partnerContentId, the Kiva logo shows up for passwordless even if there is no partner associated with the flow.

mcstover (approved):
I think this is good to go. I tested both server and client renders locally as well as with and without the passwordless and partnerContentId params. Things seem in order.

PR #5315, opened 2024-05-14
feat: assign team to loan in bp

christian14b:
Sharing challenge from TY page sends you to borrower profile but it doesn't assign the team with the loan.


roger-in-kiva:
Should we add the experiment flag for this?

emuvente (approved):

PR #5316, opened 2024-05-15
fix: count month days for teams challenge days left

christian14b:
`IntervalToDuration` method returns an interval object with ({years, months, days}) we were just getting the days between two dates causing an issue if the difference between them is 1 month at least.

Note: I'm assuming challenge's days are less than 1 year for this fix

Given these dates:
Start: today
End: `2024-07-17T15:37:46.711Z`

## Issue
image

## Fix:
image

### DIFF
--- a/src/plugins/team-goal-mixin.js
+++ b/src/plugins/team-goal-mixin.js
@@ -1,4 +1,4 @@
-import { differenceInCalendarDays } from 'date-fns';
+import { formatDistanceToNowStrict } from 'date-fns';
 
 export default {
 	props: {
@@ -46,11 +46,13 @@ export default {
 			return this.goal?.targets?.values?.[0]?.targetLendAmount ?? 0;
 		},
 		daysLeft() {
-			const start = this.goal?.startDate ? new Date(this.goal?.startDate) : new Date();
 			const end = this.goal?.endDate ? new Date(this.goal?.endDate) : new Date();
-			return differenceInCalendarDays(
+			return formatDistanceToNowStrict(
 				end,
-				start,
+				{
+					unit: 'day',
+					roundingMethod: 'ceil'
+				}
 			);
 		},
 		challengeDescription() {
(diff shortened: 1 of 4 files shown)


src/plugins/team-goal-mixin.js
```diff
@@ -48,10 +48,17 @@ export default {
...
-			return intervalToDuration({
+			const interval = intervalToDuration({
 				start,
 				end,
-			}).days;
+			});
+
+			if (interval?.months) {
+				const daysInMonth = getDaysInMonth(start);
+				return interval.days + (interval.months * daysInMonth);
```
roger-in-kiva, line 58: Not sure if we should cover this right now, but what happens if we have more than 1 month?

src/plugins/team-goal-mixin.js
```diff
@@ -48,10 +48,17 @@ export default {
 		daysLeft() {
 			const start = this.goal?.startDate ? new Date(this.goal?.startDate) : new Date();
 			const end = this.goal?.endDate ? new Date(this.goal?.endDate) : new Date();
-			return intervalToDuration({
+			const interval = intervalToDuration({
```
mcstover, line 51: I think we may want to use one of the "days" based methods instead. Maybe [differenceInDays](https://date-fns.org/v3.6.0/docs/differenceInDays)

mcstover: That said, there is still opportunity with both your current method and the new one to have an offset in hours due to timezones when using dates from the goal service which are static...

emuvente:
I think formatDistanceToNowStrict will do what you want here, no matter the dates. https://date-fns.org/v3.6.0/docs/formatDistanceToNowStrict

mcstover:
This exploration of date-fns methods will help use refine our clientside date handling in the impact-dashboard too ;)

mcstover (approved):
Neat!

PR #5320, opened 2024-05-17
fix: amount lent value

christian14b:
Only limited participants where considered to set funded amount value.

### DIFF
--- a/src/plugins/team-goal-mixin.js
+++ b/src/plugins/team-goal-mixin.js
@@ -28,9 +28,7 @@ export default {
 		},
 		percentageFunded() {
 			const targetAmount = this.goal?.targets?.values?.[0]?.targetLendAmount ?? 0;
-			const fundedAmount = this.goal?.participation?.values?.reduce((sum, value) => {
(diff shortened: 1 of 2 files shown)


1337jwilliams95 (approved):
LGTM

src/plugins/team-goal-mixin.js
```diff
@@ -28,19 +28,15 @@ export default {
...
 			return Math.floor((fundedAmount / targetAmount) * 100) || 0;
 		},
 		participationTotalCount() {
 			return this.goal?.participation?.totalCount ?? 0;
 		},
 		fundedAmount() {
-			return this.goal?.participation?.values?.reduce((sum, value) => {
-				return sum + (value?.amountLent ?? 0);
-			}, 0) ?? 0;
+			return this.goal?.participation?.amountLent ?? 0;
```
emuvente, line 39: `amountLent` is a string, but fundedAmount is expected to be a number by the components that use it. This should be converted to a number here to avoid unexpected results.

PR #5326, opened 2024-05-21
fix: add check for new access_denied error

dyersituations:
https://kiva.atlassian.net/browse/MP-245

### DIFF
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -165,8 +165,8 @@ module.exports = function authRouter(config = {}) {
 
 			// Handle errors
 			if (req.query.error && !silentAuth) {
-				// Re-attempt login with the login form forced to display if unauthorized error happened
-				if (req.query.error === 'unauthorized') {
+				// Re-attempt login with the login form forced to display if unauthorized/access_denied error happened
+				if (req.query.error === 'unauthorized' || req.query.error === 'access_denied') {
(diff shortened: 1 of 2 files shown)


mcstover (approved):

roger-in-kiva (approved):

server/auth-router.js
```diff
@@ -165,8 +165,8 @@ module.exports = function authRouter(config = {}) {
 
 			// Handle errors
 			if (req.query.error && !silentAuth) {
-				// Re-attempt login with the login form forced to display if unauthorized error happened
-				if (req.query.error === 'unauthorized') {
+				// Re-attempt login with the login form forced to display if unauthorized/access_denied error happened
+				if (req.query.error === 'unauthorized' || req.query.error === 'access_denied') {
```
emuvente, line 169: I don't think we want to always reattempt login when there is an access_denied. There may be some cases where we do want to go to the error page

emuvente: Yeah, I found that we still want to show the error page when the error description contains force_password_reset, and there are probably others. I think this should parse the error description and only return to login in the specific cases we expect, and otherwise show the error page.

dyersituations: I guess we did return the unauthorized error in two cases. Just pushed an update to check specifically for the timeout message.

emuvente (approved):

mcstover (approved):

PR #5328, opened 2024-05-22
feat: opt-in language exp mp-270

christian14b:
Checkout and Register Social

Experiment:
image

### DIFF
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -81,7 +81,7 @@
 					>Privacy {{ enableCommsExperiment ? 'Notice' : 'Policy' }}</a>.
 					<p v-if="$v.termsAgreement.$error" class="input-error tw-text-danger tw-text-base">
 						You must agree to the Kiva Terms of service & Privacy
-						{{ enableCommsExperiment ? 'notice' : 'policy' }}.
+						{{ enableCommsExperiment ? 'Notice' : 'Policy' }}.
 					</p>
 				</kv-checkbox>
 				<kv-checkbox
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -800,15 +800,13 @@ export default {
 			}
 		}
 
-		if (this.cookieStore.get(COMMS_OPT_IN_EXP_KEY) !== 'true') {
-			const { version } = this.apollo.readFragment({
-				id: `Experiment:${COMMS_OPT_IN_EXP_KEY}`,
-				fragment: experimentVersionFragment,
-			}) ?? {};
+		const { version } = this.apollo.readFragment({
+			id: `Experiment:${COMMS_OPT_IN_EXP_KEY}`,
+			fragment: experimentVersionFragment,
+		}) ?? {};
 
-			if (version === 'b') {
-				this.cookieStore.set(COMMS_OPT_IN_EXP_KEY, true, { path: '/' });
-			}
+		if (version === 'b') {
+			this.cookieStore.set(COMMS_OPT_IN_EXP_KEY, true, { path: '/' });
 		}
 
 		userHasLentBefore(this.cookieStore.get(hasLentBeforeCookie) === 'true');
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -126,7 +126,7 @@ module.exports = function authRouter(config = {}) {
 
 		// Opt-In Communication Exp MP-271
 		if (cookies.opt_in_comms) {
-			options.optInComms = 'b';
+			options.optInComms = true;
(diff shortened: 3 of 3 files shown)


src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -77,11 +77,11 @@
...
 						:href="`https://${this.$appConfig.host}/legal/privacy`"
 						target="_blank"
-						title="Open Privacy Policy in a new window"
-					>Privacy Policy</a>.
+						:title="`Open Privacy ${enableCommsExperiment ? 'Notice' : 'Policy' } in a new window`"
+					>Privacy {{ enableCommsExperiment ? 'Notice' : 'Policy' }}</a>.
 					<p v-if="$v.termsAgreement.$error" class="input-error tw-text-danger tw-text-base">
 						You must agree to the Kiva Terms of service & Privacy
-						policy.
+						{{ enableCommsExperiment ? 'Notice' : 'Policy' }}.
```
dyersituations, line 84: It looks like "policy" was lowercase before, is having it always uppercase expected?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -148,6 +152,11 @@ export default {
 		KvTextInput,
 	},
 	inject: ['apollo', 'cookieStore'],
+	apollo: {
```
dyersituations, line 155: This should already be handled globally here: https://github.com/kiva/ui/blob/4085cf13460d40fc9771fd793128d222847b246a/src/util/experiment/experimentUtils.js#L343

Does this still work like expected with this `preFetch` removed?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -198,6 +207,16 @@ export default {
 			const inner = parts.length ? `${parts.join(', ')} and ${last}` : last;
 			return `To finish creating your account, please ${inner} below.`;
 		},
+		emailUpdatesCopy() {
+			if (this.enableCommsExperiment) {
```
dyersituations, line 211: Will we have a situation where `passwordless` and `enableCommsExperiment` are both enabled?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -258,18 +277,41 @@ export default {
...
-				query: strategicPartnerLoginInfoByPageIdQuery,
-				variables: { pageId: route.query.partnerContentId ?? '' }
-			});
-		},
+			return Promise.all([
+				client.query({
+					query: strategicPartnerLoginInfoByPageIdQuery,
+					variables: { pageId: route.query.partnerContentId ?? '' }
+				}),
+				client.query({
```
dyersituations, line 309: See other comment, this probably isn't needed.

src/pages/LoginAndRegister/RegisterSocial.vue
mcstover: I would say if `passwordless` is active the exp should be bypassed. The corporate partner context isn't something we want to "mix" with general experimentation and passwordless is specifically for that context at this time.

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -77,11 +77,11 @@
...
 						:href="`https://${this.$appConfig.host}/legal/privacy`"
 						target="_blank"
-						title="Open Privacy Policy in a new window"
-					>Privacy Policy</a>.
+						:title="`Open Privacy ${enableCommsExperiment ? 'Notice' : 'Policy' } in a new window`"
+					>Privacy {{ enableCommsExperiment ? 'Notice' : 'Policy' }}</a>.
 					<p v-if="$v.termsAgreement.$error" class="input-error tw-text-danger tw-text-base">
 						You must agree to the Kiva Terms of service & Privacy
-						policy.
+						{{ enableCommsExperiment ? 'notice' : 'policy' }}.
```
dyersituations, line 84: I see this was reverted to be lowercase, but do we actually want it uppercase to match other usages in the page?

src/components/WwwFrame/TheHeader.vue
```diff
@@ -798,6 +800,17 @@ export default {
 			}
 		}
 
+		if (this.cookieStore.get(COMMS_OPT_IN_EXP_KEY) !== 'true') {
```
dyersituations, line 803: Why check the cookie value here?

src/pages/LoginAndRegister/RegisterSocial.vue
dyersituations: It looks like this still needs to be bypassed if `passwordless` is `true`

dyersituations (approved):

PR #5333, opened 2024-05-28
feat: second variant of opt-in experiment

christian14b:
Adding new variant to opt-in experiment using radio buttons:

image

image

### DIFF
--- a/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
+++ b/src/components/Checkout/CheckoutDropInPaymentWrapper.vue
@@ -194,9 +194,9 @@ export default {
 			enableCheckoutButton: false,
 			isClientReady: false,
 			paymentTypes: ['paypal', 'card', 'applePay', 'googlePay'],
-			enableCommsExperiment: true,
+			enableCommsExperiment: false,
 			selectedComms: '',
-			enableRadioBtnExperiment: true,
+			enableRadioBtnExperiment: false,
 		};
 	},
 	validations: {


src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -65,40 +65,99 @@
...
-					@update:modelValue="$kvTrackEvent(
-						'Login',
-						'click-marketing-updates',
-						emailUpdatesCopy,
-						$event ? 1 : 0
-					)"
-				>
-					{{ emailUpdatesCopy }}
-				</kv-base-input>
+				<template v-if="enableRadioBtnExperiment">
```
dyersituations, line 68: Could this new `template` be extracted and reused between these two components?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -54,51 +54,104 @@
...
-					)"
-				>
-					{{ emailUpdatesCopy }}
-				</kv-checkbox>
+					<p
+						v-if="selectedComms === '0'"
+						class="tw-border-brand-200 tw-border tw-bg-brand-100 tw-p-1.5 tw-rounded tw-text-small"
+					>
+						Can we ask you to reconsider? This borrower and others like them will need your
+						help to change their lives. You can unsubscribe at any time.
```
dyersituations, line 103: Does design want to adjust the spacing or copy so that "time" doesn't sit on a new line by itself?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -237,6 +300,12 @@ export default {
 				checked: val => val,
 			};
 		}
+		if (this.enableRadioBtnExperiment) {
+			validations.selectedComms = {
+				required,
+				checked: val => val !== '',
```
dyersituations, line 306: Interesting, so we're making them click one more time before submitting their order? This is a new click, right?

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -54,51 +54,104 @@
...
+							value="0"
+							name="reportComms"
+							v-model="selectedComms"
+							:class="{'radio-error': $v.selectedComms.$error}"
+							v-kv-track-event="[
+								'basket',
+								'click',
+								'marketing-updates',
+								// eslint-disable-next-line max-len
+								'Dont receive email updates from Kiva (including borrower updates and promos). You can unsubscribe anytime.',
```
mcstover, line 85: Do we need "You can unsubscribe at any time." in this section? It's confusing b/c this is the opt-out so I shouldn't be subscribed and thus shouldn't have to "unsubscribe" at ANY TIME!

src/components/Checkout/CheckoutDropInPaymentWrapper.vue
```diff
@@ -149,6 +158,12 @@ export default {
 		BraintreeDropInInterface: () => import('@/components/Payment/BraintreeDropInInterface'),
 		KvCheckbox,
 		KvTextInput,
+		UserUpdatesPreference,
+	},
+	provide() {
+		return {
+			$v: this.$v
```
dyersituations, line 165: What is this for?

dyersituations: Oh this is from `vuelidate`?

dyersituations (approved):

mcstover (approved):

PR #5336, opened 2024-05-30
fix: radio button was not displayed and validation issues

christian14b:
Seems like timing issue with $v, it was not adding a validation for `selectedComms` and that was causing 500 server error on userUpdates component. As a workaround, setting the flag to true as default so it can be added to validations and then if it's not the radio button experiment, remove that validation.

Still looking some ways to do this without the needsComms flag.

### DIFF
--- a/src/pages/LoginAndRegister/RegisterSocial.vue
+++ b/src/pages/LoginAndRegister/RegisterSocial.vue
@@ -151,7 +151,7 @@
 
 <script>
 import { validationMixin } from 'vuelidate';
-import { requiredIf } from 'vuelidate/lib/validators';
+import { required, requiredIf } from 'vuelidate/lib/validators';
 import logReadQueryError from '@/util/logReadQueryError';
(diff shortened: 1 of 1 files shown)


dyersituations:
@christian14b Is this something that you cannot test locally?

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -159,6 +159,7 @@ import SystemPage from '@/components/SystemFrame/SystemPage';
 import strategicPartnerLoginInfoByPageIdQuery from '@/graphql/query/strategicPartnerLoginInfoByPageId.graphql';
 import experimentVersionFragment from '@/graphql/fragments/experimentVersion.graphql';
 import { trackExperimentVersion } from '@/util/experiment/experimentUtils';
+import UserUpdatesPreference from '@/components/Checkout/UserUpdatesPreference';
```
dyersituations, line 162: This component isn't being used.

mcstover (approved):

PR #5337, opened 2024-05-30
fix: fetch experiment data

christian14b:
Needed to query the experiment, since `assignAllActiveExperiments` was only used for pages under wwwPage.


src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -339,7 +340,7 @@ export default {
 		preFetch(config, client, { route }) {
 			const pageId = route?.query?.partnerContentId;
 			if (!pageId) {
-				return Promise.resolve();
+				return client.query({ query: experimentQuery, variables: { id: COMMS_OPT_IN_EXP_KEY } });
```
dyersituations, line 343: Any issues not running this when `pageId` is defined?

dyersituations (approved):

PR #5341, opened 2024-06-03
feat: lent amount removed in activity feed

roger-in-kiva:
- lent amount removed in activity feed


mcstover (approved):

src/components/LoanCards/KvClassicLoanCardContainer.vue
```diff
@@ -345,7 +345,7 @@ export default {
 						dataObject?.data.push({
 							lenderName: action.lender.name,
 							lenderImage: action.lender?.image?.url,
-							text: `${action.lender.name} lent $${parseFloat(action.shareAmount).toFixed()}`,
```
dyersituations, line 348: Should we also remove `shareAmount` from the gql query? If not isn't being used anywhere else.

dyersituations (approved):

PR #5342, opened 2024-06-05
feat: repayment email settings update

roger-in-kiva:
- repayment email settings update

Screenshot 2024-06-04 at 12 31 05
Screenshot 2024-06-04 at 12 31 16

### DIFF
--- a/src/pages/Settings/EmailSettings.vue
+++ b/src/pages/Settings/EmailSettings.vue
@@ -81,30 +81,6 @@
 								</kv-checkbox>
 
 								<label
-									class="tw-block tw-mt-4 tw-mb-1"
-									for="repayment-updates-input"
-								>Repayment notifications</label>
-								<kv-select
-									id="repayment-updates-input"
-									v-model="form.repaymentUpdates"
-								>
-									<option value="none">
-										Do not send
-									</option>
-									<option value="nightly">
-										Send a nightly digest
-									</option>
-									<option value="monthly">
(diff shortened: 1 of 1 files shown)


src/pages/Settings/EmailSettings.vue
```diff
@@ -627,8 +679,6 @@ export default {
 								}
 							}
 						`,
-						awaitRefetchQueries: true,
```
dyersituations, line 630: Why were these two lines removed?

roger-in-kiva: Because we have the mutation below doing the same.

dyersituations: Right, but these are different mutations. Don't we also need to refetch after unsubscribing here?

roger-in-kiva: Refetching is to get the updated data on the page, `pageQuery` and show it as mutations are done.
Unsubscribing implies repayments too, so for the case `Only legally required emails` we need the second mutation if we have a different value than default for repayments, so we need to refetch the page data once. makes sense?

dyersituations: Oh I clicked to expand the methods and see that both mutations are in the same method, makes sense, I missed that before.

dyersituations (approved):

mcstover (approved):

PR #5343, opened 2024-06-05
feat: mp-267 thanks page for opted in users

christian14b:
[WIP] New thank you page for opt-in

image

### DIFF
--- a/src/components/Thanks/LoanNextSteps.vue
+++ b/src/components/Thanks/LoanNextSteps.vue
@@ -55,9 +55,6 @@ export default {
 				animationDelay: `${index * 1}s`
 			};
 		}
-	},
-	async mounted() {
-		await this.$nextTick();
 	}
 };
 </script>


src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,369 @@
...
+		};
+	},
+	mounted() {
+		confetti({
+			origin: {
+				y: 0.2
+			},
+			particleCount: 150,
+			spread: 200,
+			colors: ['#6AC395', '##223829'],
```
dyersituations, line 329: Double `#`

src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,369 @@
...
+							class="tw-mb-5"
+							:weeks-to-repay="weeksToRepay"
+						/>
+						<div class="tw-flex tw-flex-col tw-gap-2 tw-pb-3">
+							<template v-if="userOptedOut && isGuest">
+								<h3 class="tw-text-center">
+									<!-- eslint-disable-next-line max-len -->
+									Want to hear how you’re impacting {{ borrowerName }}’s life and more ways to help people like them?
+								</h3>
+								<kv-button>
```
dyersituations, line 50: Do these two buttons need click events?

src/router/routes.js
```diff
@@ -676,4 +676,11 @@ module.exports = [
 			excludeFromStaticSitemap: true,
 		}
 	},
+	{
+		path: '/thanks/not-opt-in',
```
dyersituations, line 680: This path doesn't really match the name of the component.

src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,369 @@
...
+		isGuestUsCheckout() {
+			// Is a guest checking out only with US loans?
+			// eslint-disable-next-line no-underscore-dangle
+			return this.isGuest && this.loans.every(loan => loan?.__typename === 'LoanDirect');
+		},
+		hash() {
+			return this.selectedLoan.image?.hash ?? '';
+		},
+		weeksToRepay() {
+			const date = Date.parse(this.selectedLoan?.disbursalDate) ?? new Date();
```
emuvente, line 278: The date that a loan will start to be repaid is different depending on the loan. For loans that have a fixed repayment plan you can get the date from the first element of the expectedPayments array:
```graphql
... on LoanBasic {
  terms {
    expectedPayments {
      dueToKivaDate
    }
  }
}
```
For loans that don't have any expected payments yet, we can assume that the first repayment will start 1 month after the loan finishes fundraising. We don't know when the loan will finish fundraising, just that it will be between now and the `plannedExpirationDate`. So it's actually a range of time that the first repayment could happen, and we need to calculate the beginning and end of that range and translate it to something like "2-3 weeks" or "3-5 weeks".

src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,369 @@
...
+			const days = formatDistanceToNowStrict(
```
emuvente, line 279: The number of weeks can be calculated with [differenceInWeeks](https://date-fns.org/v3.6.0/docs/differenceInWeeks):
```js
// Get the number of weeks between the first repayment date (in the future) and now
return differenceInWeeks(date, new Date())
```

src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,407 @@
...
+			const date = this.selectedLoan?.terms?.expectedPayments?.[0]?.dueToKivaDate ?? null;
+			const today = new Date();
+			if (date) {
+				// Get the number of weeks between the first repayment date (in the future) and now
+				return `${differenceInWeeks(Date.parse(date), today)} weeks`;
+			}
+
+			// Calculating a possible range of weeks between the planned expiration date and a month after
+			const expDate = Date.parse(this.selectedLoan?.plannedExpirationDate);
+			const minDate = differenceInWeeks(expDate, today);
```
emuvente, line 320: The minDate will be 1 month after today, so `differenceInWeeks(addMonths(today, 1), today)`

src/pages/Thanks/ThanksPageOptIn.vue
```diff
@@ -0,0 +1,407 @@
...
+			const maxDate = differenceInWeeks(addMonths(expDate, 1), today);
+
+			return `${minDate} - ${maxDate} weeks`;
```
emuvente, line 323: It's possible that minDate and maxDate will be the same number of weeks, due to rounding, so this should check for that case and only return `${minDate} weeks` if they're the same.

src/graphql/query/thanksPage.graphql
```diff
@@ -85,6 +86,12 @@ query checkoutReceipt($checkoutId: Int!, $visitorId: String) {
 							unreservedAmount @client
 							inPfp
 							gender
+							... on LoanDirect {
+								disbursalDate
```
dyersituations, line 90: Is `disbursalDate` currently being used? I'm not seeing it in the changeset.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -473,6 +487,13 @@ export default {
 			fragment: experimentVersionFragment,
 		}) || {};
 		this.enableMayChallengeHeader = shareChallengeExpData?.version === 'b';
+
+		const landedLoanId = this.cookieStore.get('optin-landed-bp');
```
dyersituations, line 491: What if the user doesn't add to the basket from the BP? Does this experiment only work if the user goes through the BP?

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+							>
+								Go to my kiva
+							</kv-button>
+						</template>
+					</div>
+				</div>
+			</div>
+			<div class="tw-py-5 tw-px-3">
+				<div class="tw-mb-2">
+					<!-- eslint-disable-next-line max-len -->
```
dyersituations, line 99: This comment can be removed.

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+					<div
+						v-if="userOptedOut && isGuest"
+						class="
+								tw-w-full tw-border tw-rounded
+								tw-flex tw-justify-between tw-cursor-pointer
+								tw-py-2 tw-px-3
+							"
+						@click="openCreateAccount = !openCreateAccount"
```
dyersituations, line 107: For ease of migration to Vue 3, this probably needs to be `() => openCreateAccount = !openCreateAccount`. But please verify that it still works in Vue 2 after changing.

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+							<guest-account-creation
+								class="tw-pt-3 account-creation"
+								event-category="thanks"
+								event-label="open-account-creation-drawer"
+							/>
+						</div>
+					</kv-expandable>
+				</div>
+				<div class="tw-mb-2">
+					<!-- eslint-disable-next-line max-len -->
```
dyersituations, line 140: This comment can be removed.

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+					<div
+						class="
+								tw-w-full tw-border tw-rounded
+								tw-flex tw-justify-between tw-cursor-pointer
+								tw-py-2 tw-px-3
+							"
+						@click="openOrderConfirmation = !openOrderConfirmation"
```
dyersituations, line 147: Same comment about turning into a function.

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+					</kv-expandable>
+				</div>
+				<div>
+					<div
+						class="
+								tw-w-full tw-border tw-rounded
+								tw-flex tw-justify-between tw-cursor-pointer
+								tw-py-2 tw-px-3
+							"
+						@click="openShareModule = !openShareModule"
```
dyersituations, line 183: Same.

src/components/Thanks/OptInExpVariation.vue
```diff
@@ -0,0 +1,417 @@
...
+
+.borrower-container {
+	animation: fadein ease-in 1s;
+	width: 150px;
+	@apply tw-block tw-relative tw-mx-auto tw-z-4;
+}
+
+.borrower-image {
+	width: 150px;
+	position: absolute;
```
dyersituations, line 380: `tw-absolute`

src/pages/Thanks/ThanksPage.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-	<www-page data-testid="thanks-page">
+	<www-page data-testid="thanks-page" :class="{'tw-bg-eco-green-1': !showOldThanksPage && !isOnlyDonation}">
 		<template v-if="isOnlyDonation">
 			<thanks-page-donation-only
 				:monthly-donation-amount="monthlyDonationAmount"
 			/>
 		</template>
-		<template v-else>
+		<template v-else-if="showOldThanksPage">
```
dyersituations, line 8: Are we sure this won't hide some behavior that we want? This part of the template covers a number of different thanks page variations.

src/pages/Thanks/ThanksPage.vue
dyersituations: Removed for this PR or future ticketed work?

src/pages/Thanks/ThanksPage.vue
```diff
@@ -473,6 +493,29 @@ export default {
 			fragment: experimentVersionFragment,
 		}) || {};
 		this.enableMayChallengeHeader = shareChallengeExpData?.version === 'b';
+
+		// Only show the old page if the first page is a borrower page
+		if (this.$appConfig?.firstPage?.indexOf('/lend/') !== -1) {
+			const url = this.$appConfig.firstPage?.split('/')[2];
+			const firstVisitloanId = url?.[2] ?? null;
```
emuvente, line 500: it might be also good to check that url[1] is 'lend' so that we know that url[2] is a loan id and not something else.

src/pages/Thanks/ThanksPage.vue
emuvente: It seems like this is what is being requested by product. However, I agree that it doesn't feel right. @christian14b I think it would be good to also put this new thanks page behind a feature flag so that we can safely rollout the code changes to production while we test the different checkout scenarios in development and review them with product.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -473,6 +493,31 @@ export default {
 			fragment: experimentVersionFragment,
 		}) || {};
 		this.enableMayChallengeHeader = shareChallengeExpData?.version === 'b';
+
+		// Only show the old page if the first page is a borrower page and feature flag is enabled
+		const newThanksPageEnabled = data?.general?.newThanksPageEnabled?.value === 'true';
+
+		if (newThanksPageEnabled && this.$appConfig?.firstPage?.indexOf('/lend/') !== -1) {
```
dyersituations, line 500: The comment mentions only showing the new thanks page if the first page of the session is the BP, but this would also target the `/lend/filter` page, which I also think is a common landing page for users. You could do a regex match here and make sure the match is /lend/number.

src/graphql/query/thanksPage.graphql
```diff
@@ -146,6 +147,10 @@ query checkoutReceipt($checkoutId: Int!, $visitorId: String) {
 			key
 			value
 		}
+		newThanksPageEnabled: featureSetting(key: "new_thanks_page.enabled") {
```
dyersituations, line 150: Is there a reason we're using a feature setting instead of an experiment setting?

src/components/Thanks/OptInSteps.vue
```diff
@@ -0,0 +1,120 @@
...
+	},
+	methods: {
+		getDelayStyle(index) {
+			return {
+				animationDelay: `${index * 1}s`
+			};
+		}
+	},
+	mounted() {
+		setTimeout(() => {
```
dyersituations, line 60: Why is `setTimeout` needed here? Would `nextTick` work? Using `nextTick` might be a more specific solution, if we just need to wait for the next render.

src/components/Thanks/WhatIsNextTemplate.vue
```diff
@@ -0,0 +1,514 @@
...
+			return loan?.image?.hash ?? '';
+		},
+		updateOptIn(value) {
+			this.selectOption = true;
+			this.openCreateAccount = true;
+			if (value) {
+				try {
+					this.apollo.mutate({
+						mutation: gql`
+									mutation updateCommunicationSettings(
```
dyersituations, line 412: Extra indentation here could be removed.

dyersituations:
@christian14b There are a lot of comments not responded to and/or unresolved. Please respond and resolve so we know if other work is in progress.

dyersituations:
> This should be the new TY page for all lenders who just made their first transaction and didn’t opt-in. I added those conditions to the flag and removed the feature flag

Is there still work in progress or is this ready for review again?

src/pages/Thanks/ThanksPage.vue
```diff
@@ -473,6 +491,33 @@ export default {
...
+		// New Thanks Page Experiment
+		if (this.showNewTYPage) {
+			const { version } = trackExperimentVersion(
+				this.apollo,
+				this.$kvTrackEvent,
+				'thanks',
+				NEW_THANKS_PAGE_EXP,
+				'EXP-MP-267-Jun2024',
+			);
+			if (version === 'b') {
```
dyersituations, line 517: So this new experiment is only shown for certain users AND has two variations?

src/components/Thanks/LoanNextSteps.vue
```diff
@@ -0,0 +1,119 @@
...
+	methods: {
+		getDelayStyle(index) {
+			return {
+				animationDelay: `${index * 1}s`
+			};
+		}
+	},
+	async mounted() {
+		await this.$nextTick();
+		this.currentStep = 1;
```
dyersituations, line 61: Why is this needed after a `nextTick`? The `data` value is already `1`.

dyersituations (approved):

PR #5346, opened 2024-06-11
feat: add banner link back to impact dash if user came from impact dash

eddieferrer:
When a user arrives to the BP with a query param like: `fromContext=/impact-dashboard/hitachi` Generate banner with link back to `/impact-dashboard/hitachi`

This banner state should trump all other states. 

Waiting on copy confirmation in ticket, so there could be a copy  update here

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/PromoCreditBanner.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/PromoCreditBanner.vue
@@ -1,7 +1,25 @@
-<!-- eslint-disable vue/no-useless-template-attributes -->
 <template>
 	<div
-		v-if="lendingRewardOffered"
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/PromoCreditBanner.vue
```diff
@@ -22,7 +40,6 @@
 		</template>
 		<template
 			v-else
-			v-kv-track-event="['TopNav','click-Promo','Lending Reward Banner']"
```
eddieferrer, line 25: You can't have a directive on a temporary template element that vue is going to strip out anyway

mcstover (approved):

dyersituations (approved):

PR #5353, opened 2024-06-20
fix: firstLoan condition was false in guest checkout

christian14b:
Set `isFirstLoan` to true if loans exist and it was a guest checkout

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -450,7 +450,7 @@ export default {
 		userHasDepositBefore(hasDepositBefore);
 
 		const totalLoans = data?.my?.loans?.totalCount ?? 0;
-		const isFirstLoan = this.loans.length && (totalLoans === this.loans.length || this.isGuest);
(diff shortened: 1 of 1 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -450,7 +450,7 @@ export default {
 		userHasDepositBefore(hasDepositBefore);
 
 		const totalLoans = data?.my?.loans?.totalCount ?? 0;
-		const isFirstLoan = this.loans.length && totalLoans === this.loans.length;
+		const isFirstLoan = this.loans.length && (totalLoans === this.loans.length || this.isGuest);
```
dyersituations, line 453: This is affecting all guest checkout, not just this experiment. Is that expected? And is this because `data?.my?.loans?.totalCount` is `undefined` for guests?

dyersituations: Yeah it we only want this to affect the experiment, a small adjustment is needed.

dyersituations (approved):

PR #5356, opened 2024-06-21
fix: always evaluate if user landed on a us loan to prevent rendering new ty page

christian14b:


src/pages/Thanks/ThanksPage.vue
```diff
@@ -367,6 +366,20 @@ export default {
...
+			if (bpPattern.test(this.$appConfig.firstPage)) {
+				const url = this.$appConfig.firstPage?.split('/');
+				const firstVisitloanId = url?.[2] ?? null;
+				const landedLoan = this.loans.find(loan => loan.id === Number(firstVisitloanId));
+				return landedLoan?.geocode?.country?.isoCode === 'US';
+			}
+			return false;
+		},
+		showNewTYPage() {
+			return !this.landedOnUSLoan && (this.isFirstLoan || this.isGuest) && !this.optedIn;
```
dyersituations, line 381: How is this solution different? Was data not completed loaded on `created`?

emuvente: Because of [this if statement](https://github.com/kiva/ui/pull/5356/files#diff-c3f6821900e71ed9c13104928b17481990a08fb8bc9c8a5073a62a4adeee0b1bL498) `showNewTYPage` would only be evaluated when the user started their session on a borrower profile, preventing anyone else from seeing the new thanks page.

dyersituations: Ah, we're showing this for non US loans, got it, thanks.

dyersituations (approved):

PR #5357, opened 2024-06-24
fix: copy change

christian14b:


src/components/Thanks/LoanNextSteps.vue
```diff
@@ -43,7 +43,7 @@ export default {
 			currentStep: 1,
 			steps: [
 				{ text: 'Your contribution is received ' },
-				{ text: 'Their loan is fully funded' },
+				{ text: 'Their is funded' },
```
dyersituations, line 46: Is there a word missing? Should it be `Their loan is funded`?

dyersituations (approved):

PR #5358, opened 2024-06-26
feat: pie chart design for distribution tables MP-434

emuvente:
Pie chart was requested for new stack, but no designs provided. The design of the component in this PR has been briefly reviewed, so I'm creating this as a story for final review. There will be follow-up work to actually use this pie chart after it has been approved by prod/design.

Screenshot 2024-06-25 at 5 00 48 PM

### DIFF
--- a/src/components/Charts/PieChartFigure.vue
+++ b/src/components/Charts/PieChartFigure.vue
@@ -1,0 +1,201 @@
+<template>
+	<figure
+		class="pie-chart tw-flex tw-items-center tw-justify-center tw-gap-2"
+		@mouseleave="activeSlice = null"
+	>
+		<!-- pie chart -->
+		<div class="tw-relative tw-h-full">
+			<!-- eslint-disable-next-line max-len -->
+			<div v-if="loading" class="pie-placeholder tw-h-full tw-p-2.5">
+				<div class="tw-overflow-hidden tw-rounded-full tw-h-full">
+					<kv-loading-placeholder />
+				</div>
+			</div>
+			<svg
+				v-else
+				class="tw-h-full"
+				viewBox="0 0 32 32"
+				xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
+			>
+				<path
+					v-for="(slice, index) in slices"
+					:key="index"
+					class="tw-origin-center tw-transition-transform"
+					:style="activeSlice === slice ? { transform: 'scale(1.1)' } : {}"
+					:d="slice.path"
+					:stroke="slice.color"
+					:stroke-width="lineWidth"
+					fill="none"
+					@mouseenter="activeSlice = slice"
+					@click="activeSlice = slice"
+				/>
+			</svg>
+			<!-- active slice -->
+			<div
+				v-if="activeSlice"
+				class="
+					tw-absolute tw-top-1/2 tw-left-1/2 -tw-translate-x-1/2 -tw-translate-y-1/2
+					tw-flex tw-flex-col tw-items-center tw-text-center"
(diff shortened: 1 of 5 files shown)


mcstover (approved):
Looks fun, would be nice to have in kv-components and with colors mapped from ecosystem once the design stabilizes...

src/components/Charts/PieChartFigure.vue
```diff
@@ -0,0 +1,201 @@
+<template>
+	<figure
+		class="pie-chart tw-flex tw-items-center tw-justify-center tw-gap-2"
+		@mouseleave="activeSlice = null"
+	>
+		<!-- pie chart -->
+		<div class="tw-relative tw-h-full">
+			<!-- eslint-disable-next-line max-len -->
```
dyersituations, line 8: Looks like this comment can be removed.

src/components/Charts/PieChartFigure.vue
```diff
@@ -0,0 +1,201 @@
...
+		</div>
+	</figure>
+</template>
+
+<script>
+import numeral from 'numeral';
+import Alea from '@/util/experiment/Alea';
+import KvLoadingPlaceholder from '~/@kiva/kv-components/vue/KvLoadingPlaceholder';
+
+// convenience function to get point on circumfrence of a given circle (from https://codepen.io/grieve/pen/xwGMJp)
```
dyersituations, line 95: `circumference`

src/components/Charts/PieChartFigure.vue
```diff
@@ -0,0 +1,201 @@
...
+export default {
+	name: 'PieChartFigure',
+	props: {
+		loading: {
+			type: Boolean,
+			default: true,
+		},
+		values: {
+			type: Array,
+			default: () => [],
```
dyersituations, line 112: default: () => ([])

src/components/Charts/PieChartFigure.vue
```diff
@@ -0,0 +1,201 @@
...
+			for (let i = 0; i < this.values.length; i += 1) {
+				const value = this.values[i];
+				const end = start + value.percent;
+				const [startX, startY] = circumPointFromAngle(cX, cY, r, start * Math.PI * 2);
+				const [endX, endY] = circumPointFromAngle(cX, cY, r, end * Math.PI * 2);
+				const largeArc = value.percent > 0.5 ? 1 : 0;
+				// Draw just the outer arc of the slice
+				const path = `M ${startX} ${startY} A ${r} ${r} 0 ${largeArc} 1 ${endX} ${endY}`;
+				// Draw the entire slice
+				// const path = `M ${cX} ${cY} L ${startX} ${startY} A ${r} ${r} 0 ${largeArc} 1 ${endX} ${endY} Z`;
```
dyersituations, line 156: Remove?

src/components/Charts/PieChartFigure.vue
emuvente: good call

src/components/Charts/PieChartFigure.vue
emuvente: I had left it in there in case we did want to draw a full slice, but I think that's probably easy enough to figure out again in the future if we do want to change the design. I'll pull it out to avoid confusion.

dyersituations (approved):

emuvente:
> Looks fun, would be nice to have in kv-components and with colors mapped from ecosystem once the design stabilizes...

@mcstover Thanks! Yeah, I agree this would be good to have in kv-components. I'm going to make a follow-up ticket for that. For the colors though, Design has already weighed in and they said they like the random colors for now and they will work on building out a set of colors to use in the future if/when they have time for it.

PR #5359, opened 2024-06-26
feat: disable sticky checkout and log experiment assignment in desktop

christian14b:


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -683,9 +683,15 @@ export default {
 		if (this.showCheckoutStickyExperiment) {
 			fireHotJarEvent('checkout_sticky_experiment');
 		}
+
+		// MP-445:
```
dyersituations, line 687: I think we can remove this comment.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -683,9 +683,15 @@ export default {
...
+		this.determineIfMobile();
+		window.addEventListener('resize', this.throttledResize);
+		this.initializeCheckoutStickyExperiment();
 	},
 	beforeDestroy() {
 		window.removeEventListener('scroll', this.checkIsAboveCheckoutActionsThrottled);
+		window.addEventListener('resize', this.throttledResize);
```
dyersituations, line 694: `removeEventListener`

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1201,6 +1212,13 @@ export default {
 				this.checkoutStickyExperimentEnabled = checkoutStickyExperiment.version === 'b';
 			}
 		},
+		// MP-445: Following two methods are defined to avoid sticky checkout experiment to be enabled in desktop
```
dyersituations, line 1215: Same with this comment.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1190,6 +1196,11 @@ export default {
 					fragment: experimentVersionFragment,
 				}) || {};
 
+				if (!this.isMobile) {
```
dyersituations, line 1199: This if statement can be removed.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -683,9 +683,15 @@ export default {
 		if (this.showCheckoutStickyExperiment) {
 			fireHotJarEvent('checkout_sticky_experiment');
 		}
+
+		// MP-445:
+		this.determineIfMobile();
+		window.addEventListener('resize', this.throttledResize);
+		this.initializeCheckoutStickyExperiment();
```
dyersituations, line 690: Wrap this in `if (this.isMobile)`.

src/pages/Checkout/CheckoutPage.vue
dyersituations: Yeah we don't need that new log

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -683,9 +682,16 @@ export default {
 		if (this.showCheckoutStickyExperiment) {
 			fireHotJarEvent('checkout_sticky_experiment');
 		}
+
+		this.determineIfMobile();
+		window.addEventListener('resize', this.throttledResize);
```
dyersituations, line 687: Actually we probably don't need these event listeners at all. We really only care if the user loads the checkout page in a specific viewport. Mounted won't run when the viewport changes anyways. Why don't we just use this code and simplify these changes:

```
		if (window?.innerWidth) {
			this.initializeCheckoutStickyExperiment();
		}
```

dyersituations (approved):

PR #5360, opened 2024-06-26
fix: prevent failed reciept query thanks

mcstover:
Folks are consistently getting to, or trying to load thanks without a transaction id which spews a mass query failure error for the reciept. This prevents us from executing an unnessesary query and restores the Login link instead of showing an empty page frame.

Also adds notes on dev setup with caddy...

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,52 @@ The Kiva UI project is bound by a [Code of Conduct](code_of_conduct.md).
 
 Kiva welcomes outside contributions to our UI repository. If you have any ideas for a feature or improvement, create an issue and we can discuss whether it makes sense to create a pull request. Thanks for the help!
 
-## Build Setup for localhost (outside of a Kiva VM or Server)
+# Local Development Setup with Caddy
+
+> IMPORTANT NOTE: Turn off Docker if it's running! We have a perpetually running network related to Tilt that will prevent Caddy from starting. I did consider using Tilt and docker-compose but it would require rewriting alot of the Monolith TiltVM setup so this just bypasses it all.
+
+### Required Dependencies
+
+1. Add `127.0.0.1 kiva-ui.local` to your `/etc/hosts` file on your mac
+	- Auth0 configs are already in place to support this domain in the dev tenant
+
+2. Install Caddy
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,13 +12,52 @@ The Kiva UI project is bound by a [Code of Conduct](code_of_conduct.md).
...
+1. Add `127.0.0.1 kiva-ui.local` to your `/etc/hosts` file on your mac
+	- Auth0 configs are already in place to support this domain in the dev tenant
+
+2. Install Caddy
+
+	 `brew install caddy`
+
+3. Start Caddy from the root of the ui repo
+
+	`caddy start`
```
dyersituations, line 30: Doesn't need to change, but I use `caddy run` so I can watch for failures (not usually important though...).

src/pages/Thanks/ThanksPage.vue
```diff
@@ -387,6 +397,12 @@ export default {
 		const transactionId = this.$route.query?.kiva_transaction_id
 			? numeral(this.$route.query?.kiva_transaction_id).value()
 			: null;
+
+		// Check if transactionId is null, exit if missing
+		if (!transactionId) {
+			return false;
```
dyersituations, line 403: What happens when in this case? What does the user end up seeing?

mcstover: Continuing beyond this point without a transaction id causes a readQuery error, similar to the prefetch error. Visitors would see the same message to login.

dyersituations (approved):

src/pages/Thanks/ThanksPage.vue
mcstover: From my read, nothing below this can function properly without successful query which will fail without the transaction id.

emuvente (approved):

mcstover:
@emuvente Do you think we should still log a more basic error, or possibly a warning if we hit this scenario?

emuvente:
@mcstover yeah, having a warning for this case would be good. There might be an opportunity later to make this experience better.

PR #5366, opened 2024-06-28
fix: design tweaks

christian14b:
Currently:
image

After changes:
image

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -1,5 +1,5 @@
 <template>
-	<www-page data-testid="thanks-page" :class="{'tw-bg-eco-green-1 auto-height': showNewTYPage && !isOnlyDonation}">
+	<www-page data-testid="thanks-page" :class="{'tw-bg-eco-green-1 !tw-h-auto': showNewTYPage && !isOnlyDonation}">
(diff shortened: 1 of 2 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<www-page data-testid="thanks-page" :class="{'tw-bg-eco-green-1': showNewTYPage && !isOnlyDonation}">
+	<www-page data-testid="thanks-page" :class="{'tw-bg-eco-green-1 auto-height': showNewTYPage && !isOnlyDonation}">
```
dyersituations, line 2: Depending on TW compilation (sometimes TW misses classes in `:class` it seems), this could be `!tw-h-auto`. If it doesn't work, try commenting out code in `tailwind.config.js`, then uncommenting, and seeing if the class is included in the TW output.

dyersituations: Yeah there's a known bug with local TW compilation, and I found that trick in a random GitHub thread recently.

emuvente (approved):
Looks great! This also gives us a few new examples of ecosystem being pushed a little beyond what has already been defined, so we can use this to help make a full ecosystem styleguide.

PR #5379, opened 2024-07-11
fix: mp 517 paypal credit update config

mcstover:
- ensures we update the config for paypalCredit like we do for paypal, google and applepay
- makes the amount watcher use the immediate flag
- extend the overlay visibility so the braintree widget is covered until it gets the updated values

### DIFF
--- a/src/components/Checkout/DonationItem.vue
+++ b/src/components/Checkout/DonationItem.vue
@@ -353,9 +353,8 @@ export default {
 						'basket',
 						'Update Donation',
 						'Update Success',
-						// pass raw donation amount to property field
-						numeral(this.amount).format('0,0.00'),
-						// pass donation amount as whole number in value field
+						// pass donation amount as whole number
+						numeral(this.amount).value() * 100,
 						numeral(this.amount).value() * 100
 					);
 					this.amount = numeral(this.amount).format('$0,0.00');
--- a/src/components/Checkout/DonationItem.vue
+++ b/src/components/Checkout/DonationItem.vue
@@ -353,8 +353,9 @@ export default {
 						'basket',
 						'Update Donation',
 						'Update Success',
-						// pass donation amount as whole number
-						numeral(this.amount).value() * 100,
+						// pass raw donation amount to property field
+						numeral(this.amount).format('0,0.00'),
+						// pass donation amount as whole number in value field
 						numeral(this.amount).value() * 100
 					);
 					this.amount = numeral(this.amount).format('$0,0.00');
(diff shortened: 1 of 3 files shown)


src/components/Checkout/DonationItem.vue
```diff
@@ -353,8 +353,9 @@ export default {
 						'basket',
 						'Update Donation',
 						'Update Success',
-						// pass donation amount as whole number
-						numeral(this.amount).value() * 100,
+						// pass raw donation amount to property field
+						numeral(this.amount).format('0,0.00'),
+						// pass donation amount as whole number in value field
```
dyersituations: Should the value also be the raw donation?

mcstover: No, we pass an augmented value to ensure the google events can recieve it. Google only supports a whole number, not floats.

I've reverted this anyhow b/c it's converted so that when we see 1 it's 1 cent... which is correct, just confusing.

src/components/Payment/BraintreeDropInInterface.vue
```diff
@@ -104,17 +104,26 @@ export default {
...
-				);
-				this.btDropinInstance?.updateConfiguration?.('applePay', 'paymentRequest', this.applePaymentRequest);
-			}
-		}
+		// Watch the amount property for changes and update each payment method configuration to ensure correct amount.
+		amount: {
+			handler(next, prev) {
+				if (next !== prev) {
+					this.btDropinInstance?.updateConfiguration?.('paypal', 'amount', this.formattedAmount);
+					this.btDropinInstance?.updateConfiguration?.('paypalCredit', 'amount', this.formattedAmount);
```
dyersituations, line 112: What is `paypalCredit` specifically compared to `paypal`?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -940,7 +940,10 @@ export default {
 				if (hasFreeCredits && refreshEvent === 'kiva-card-applied') {
 					this.disableGuestCheckout();
 				}
-				this.setUpdatingTotals(false);
+				const overlayTimeout = setTimeout(() => {
+					this.setUpdatingTotals(false);
+					clearTimeout(overlayTimeout);
```
dyersituations, line 945: This `clearTimeout` shouldn't be needed.

src/components/Payment/BraintreeDropInInterface.vue
mcstover: Ability to use a paypal credit is part of the paypal modal if you select that option specifically. We don't show a separate button for it though.

dyersituations (approved):

JoeTravisKiva (approved):

PR #5380, opened 2024-07-11
feat: add to basket experiment mp-346

christian14b:
Bubble up an small borrower image to the new basket and show cart modal

https://github.com/user-attachments/assets/51629278-7564-45e0-9fac-71b9bfd01041

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -1,5 +1,11 @@
 <template>
-	<header :class="{ 'tw-fixed tw-w-full tw-z-1': enableBasketExperiment }">
+	<header
+		class="tw-transition-all tw-duration-1000 tw-ease-in-out"
+		:class="{
+			'tw-absolute tw-z-1 tw-w-full' : enableAddToBasketExp ,
+			'sticky-header !tw-fixed tw-w-full tw-z-1': enableAddToBasketExp && hasBasket
+		}"
+	>
 		<nav
 			aria-label="Primary navigation"
 			class="tw-bg-primary tw-border-b tw-border-tertiary tw-relative"
--- a/src/pages/Lend/LoanChannelCategoryControl.vue
+++ b/src/pages/Lend/LoanChannelCategoryControl.vue
@@ -1,7 +1,7 @@
 <template>
 	<div
-		class="tw-relative tw-transition-all tw-duration-1000 tw-ease-in-out"
-		:class="{'sticky-header': enableAddToBasketExp && hasBasket}"
+		class="tw-relative"
+		:class="{'sticky-header': enableAddToBasketExp }"
 	>
 		<div
 			class="tw-mx-auto tw-px-2.5 md:tw-px-4 lg:tw-px-8"
--- a/src/pages/Lend/LoanSearchPage.vue
+++ b/src/pages/Lend/LoanSearchPage.vue
@@ -17,9 +17,9 @@
 			@close="closeChallengeCallout"
 		/>
 		<article
-			class="tw-bg-secondary tw-relative tw-transition-all tw-duration-1000 tw-ease-in-out"
+			class="tw-bg-secondary"
 			:class="{'tw-pt-6': !showChallengeHeader, 'tw-pt-3 lg:tw-pt-5': showChallengeHeader,
-				'sticky-header': enableAddToBasketExp && hasBasket
+				'sticky-header': enableAddToBasketExp
 			}"
 		>
 			<kv-page-container>


src/components/WwwFrame/TheHeader.vue
```diff
@@ -340,41 +340,65 @@
 								}"
 							>
 								<router-link
+									v-if="enableBasketExperiment"
```
dyersituations, line 343: Does this new basket icon/button also need the `data-testid="header-basket"` attribute? Please check how robot + cypress tests are using that test id.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -589,16 +614,17 @@ import PromoCreditBanner from './PromotionalBanner/Banners/PromoCreditBanner';
 const hasLentBeforeCookie = 'kvu_lb';
 const hasDepositBeforeCookie = 'kvu_db';
 const COMMS_OPT_IN_EXP_KEY = 'opt_in_comms';
+const NEW_ADD_TO_BASKET_EXP = 'new_add_to_basket';
 
 const optimizelyUserDataQuery = gql`query optimizelyUserDataQuery {
-  	my {
+		my {
 		id
```
dyersituations, line 621: Looks like with the formatting fix here, `id` also should be indented.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -647,6 +674,8 @@ export default {
 			basketTotal: 0,
 			teams: null,
 			teamsMenuEnabled: false,
+			newAddToBasketExpVersion: 'a',
```
dyersituations, line 677: Would it be clearer to store whether the experiment is enabled, like we often do, or will we have a version other than a + b?

src/components/WwwFrame/TheHeader.vue
```diff
@@ -765,6 +797,16 @@ export default {
 			}, 0) ?? 0;
 			this.teams = data?.my?.teams ?? {};
 			this.teamsMenuEnabled = readBoolSetting(data, 'general.teamsMenuEnabled.value');
+
+			// Add To Basket Experiment MP-346
+			const limit = this.basketCount < 3 ? this.basketCount : 3;
+			this.loansInBasket = data?.shop?.basket?.items?.values?.slice(0, limit).map(item => {
```
dyersituations, line 803: `slice` can just take `this.basketCount` here

src/components/WwwFrame/WwwPage.vue
```diff
@@ -90,6 +92,7 @@ export default {
 			return [
 				this.mainClass,
 				{ 'tw-bg-secondary': this.grayBackground },
+				{ 'md:tw-mt-9 tw-mt-8': this.enableBasketExperiment }
```
dyersituations, line 95: This solution is causing page sizing pop-in (see below for initial page rendering). Please investigate alternatives to get the server render correct.
image

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -547,6 +555,22 @@ export default {
...
+		// MP-346 New Add To Basket
+		const newAddToBasketExpData = this.apollo.readFragment({
+			id: `Experiment:${NEW_ADD_TO_BASKET_EXP}`,
+			fragment: experimentVersionFragment,
+		}) || {};
+
+		this.newAddToBasketExpVersion = newAddToBasketExpData.version;
+
+		if (this.newAddToBasketExpVersion) {
+			this.$kvTrackEvent(
```
dyersituations, line 568: Do we want to rather track the experiment assignment from the header component, since that will run for all pages?

src/pages/Lend/LoanChannelCategoryControl.vue
```diff
@@ -330,6 +334,7 @@ export default {
 			helpMeChooseSort: null,
 			helpMeChooseLoans: [],
 			isLoadingHC: true,
+			newAddToBasketExpVersion: 'a',
```
dyersituations, line 337: Same comment about storing flag rather than assignment.

src/components/WwwFrame/WwwPage.vue
dyersituations: This is with the experiment enabled, since the header is sticky. The sticky header is why the content is higher up until the loans load, enable experiment, and push down with margin.

src/components/LoanCards/KvClassicLoanCardContainer.vue
```diff
@@ -446,5 +523,24 @@ export default {
 			}
 		},
 	},
+	computed: {
+		lenderName() {
```
eddieferrer, line 527: I think these should be `borrowerName` and `borrowerImageUrl`

eddieferrer:
Do we need to account for Anonymous loans in this work? I can't remember if Anonymous loans return "Anonymous" in the name field. I know they are few and far between, I can't even find one at the moment. @emuvente

emuvente:
> Do we need to account for Anonymous loans in this work?

Loans can only become anonymous after the fundraising period, so it shouldn't be a problem for this work. When loans are anonymous then the name does only come through from the API as "Anonymous".

src/components/WwwFrame/TheHeader.vue
```diff
@@ -799,8 +800,8 @@ export default {
 			this.teamsMenuEnabled = readBoolSetting(data, 'general.teamsMenuEnabled.value');
 
 			// Add To Basket Experiment MP-346
-			const limit = this.basketCount < 3 ? this.basketCount : 3;
-			this.loansInBasket = data?.shop?.basket?.items?.values?.slice(0, limit).map(item => {
+			const loans = data?.shop?.basket?.items?.values?.filter(loan => loan?.__typename === 'LoanReservation');
```
dyersituations: This needs a linting disable for the `__typename`

dyersituations:
I'm now seeing the page content always under the header in `/lend/filter`:
image

dyersituations:
So it looks like we're just enabling this on the category pages? Or do we also want it enabled on `/lend/filter`? If we want it on `/lend/filter`, create a mixin like we have for other experiments, and enable the experiment on the filter page. Otherwise, if we only want the experiment on the category page, move the experiment tracking back to the channel component. Sorry if that's the case, I assumed we were rolling this checkout experience to as many pages as possible.

dyersituations:
While testing on the category page, I also noticed that sometimes the animated bubbles go behind the other loan cards on the page. Make sure the z indexing is correct to make the bubbles always be on the top. I mostly noticed when I clicked a button on the left side of a row.

dyersituations:
Is it expected that the tablet view is top left of the screen (not full-width or right-side)?
image

dyersituations:
I noticed that the header doesn't become sticky until you add a loan to the basket. This means that there is a weird jump in the page layout when you are scrolled down a bit and add a loan to the basket. The jump can be seen more in mobile view. Would we rather want to have the header be sticky always for the experiment (whether there are loans in the basket or not)?

emuvente:
> > I noticed that the header doesn't become sticky until you add a loan to the basket. This means that there is a weird jump in the page layout when you are scrolled down a bit and add a loan to the basket. The jump can be seen more in mobile view. Would we rather want to have the header be sticky always for the experiment (whether there are loans in the basket or not)?
> 
> Yeah.. your solution seems like the most straightforward. I tried to make it sticky after the viewport but it looks weird because the cart modal appears with margin top value of the header's height on desktop. @emuvente what do you think?

I don't think we want to test making the header sticky always as part of this experiment. Could we try using a transition to slide the header in from the top instead of a sudden jump?

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -9,8 +17,10 @@
 			@close="closeChallengeCallout"
 		/>
 		<article
-			class="tw-bg-secondary tw-relative"
-			:class="{'tw-pt-6': !showChallengeHeader, 'tw-pt-3 lg:tw-pt-5': showChallengeHeader }"
+			class="tw-bg-secondary tw-relative tw-transition-all tw-duration-1000 tw-ease-in-out"
```
dyersituations, line 20: Can these new classes be moved below to line 22?

dyersituations: Or does the transition need to already exist?

src/pages/Lend/LoanSearchPage.vue
```diff
@@ -354,6 +371,14 @@ export default {
 				}
 			}
 		}
+
+		const data = this.apollo.readQuery({
+			query: basketCountQuery,
```
dyersituations, line 376: Is see in this changeset that we're pulling from the Apollo cache here. Where is this query being ran originally?

dyersituations (approved):

PR #5381, opened 2024-07-11
fix: cit 1728 exclude corps radio exp

mcstover:
- use `partnerContentId` for exclusion from experiment state
- use `partnerContentId` to show partner logo if present

### DIFF
--- a/src/pages/LoginAndRegister/RegisterSocial.vue
+++ b/src/pages/LoginAndRegister/RegisterSocial.vue
@@ -313,7 +313,7 @@ export default {
 			this.needsNews = true;
 		}
 
-		if (!this.passwordless && this.needsNews) {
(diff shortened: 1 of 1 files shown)


mac-arrap (approved):

src/pages/LoginAndRegister/RegisterSocial.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<system-page>
 		<div class="page-content" style="max-width: 20rem;">
-			<div v-if="passwordless">
```
dyersituations, line 4: It looks like `passwordless` isn't being used in the component anymore?

mcstover: That's true now for the time being. I'd still like to keep this var initialized so we know it's present and available.

PR #5396, opened 2024-07-16
feat: use different qr code that's vue 2 compatible

dyersituations:
https://kiva.atlassian.net/browse/MP-467

- Turns out that the Vue QR code library we were using was trying to call Vue `defineComponent` -> we're using Vue 2.6, and this method looks to have gotten backwards support in Vue 2.7 (it's a Vue 3 method)
- This changeset introduces a new QR Code package that works in old school Vue 2
- You can't get to the dialog with the code in dev (we don't send the emails to validate the account to get there), but the first photo is the dialog in prod, and the second photo is a hacked in code on a random `ui` page. The size seems to be good.

image

image


package.json
```diff
@@ -119,6 +119,7 @@
 		"popper.js": "^1.16.1",
 		"prom-client": "^15.0.0",
 		"qrcode": "^1.4.4",
+		"qrcode.vue": "^1.7.0",
```
mcstover, line 122: Do we still need the baseline `qrcode` package? Or does the new `qrcode.vue` rely on it?

mcstover (approved):

package.json
dyersituations: I was wondering the same thing, it's not referenced anywhere. Do you know where we used it?

mcstover: Didn't see it used anywhere...thought maybe it was a dep of the prior library...

dyersituations: Oh looks like it was installed the same time as the old package, I'll try removing it!

dyersituations: Done, turns out that the `qrcode` package isn't needed by this new library.

mcstover: Cool, thank you!

emuvente (approved):

PR #5398, opened 2024-07-17
chore: remove ddTrace converting that tracer implementation to a pass…

mcstover:
… through for now.

This is a preliminary step to remove ddTrace without refactoring out all of the locations we have tracer wrapping functions.

The new tracer setup is still in progress, only using "global" automatic tracing for now but we can eventually begin to trace specific methods or sections, possibly by extending the "mockTracer" methods.


emuvente (approved):
🧹🙌

eddieferrer (approved):

PR #5399, opened 2024-07-17
feat: get deposit incentive banner campaign ID from settings manager

dyersituations:
https://kiva.atlassian.net/browse/MP-539

- The deposit incentive campaign ID is now pulled from the Settings Manager
- The settings key is `config.deposit_incentive_active_campaign_id`
- There's currently an error in the dev API for pulling campaign info, so I can't test end-to-end, but I verified that the campaign ID is returned, parsed, then the query to get the data for the banner is returned
- Also cleaned up a variable name in the same file

### DIFF
--- a/src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
+++ b/src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
@@ -1,9 +1,9 @@
 <template>
-	<div v-if="!isLoggedin || !hasCampaignReward">
+	<div v-if="!isLoggedIn || !hasCampaignReward">
 		<generic-promo-banner
 			class="tw-text-center"
 			:promo-banner-content="promoBannerContent"
-			:enable-deposit-incentive-exp="isLoggedin"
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/PromotionalBanner/Banners/DepositIncentiveBanner.vue
```diff
@@ -46,31 +49,38 @@ export default {
...
+			});
 		},
 	},
+	created() {
+		const campaignId = JSON.parse(this.apollo.readQuery({ query: configSettingQuery, variables: { key } })
+			?.general?.configSetting?.value ?? '""');
+		const data = this.apollo.readQuery({
+			query: amountToLendQuery,
+			variables: { campaignId, basketId: this.cookieStore.get('kvbskt') }
+		});
```
mcstover, line 71: Should we try/catch these reads, incase the setting is missing or something fails...

dyersituations: Oh right, `readQuery` doesn't gracefully fail. Fixed.

mcstover (approved):
Thanks!

PR #5406, opened 2024-07-22
fix: missing props and emits for components and design tweaks

christian14b:
image


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -58,6 +60,7 @@ export default {
 			},
 			pageLayout: 'control',
 			enableLoanTags: false,
+			modalPosition: {},
```
dyersituations, line 63: Looks like this can be removed?

dyersituations (approved):

PR #5407, opened 2024-07-23
fix: make header sticky in add to basket exp

christian14b:
image

### DIFF
--- a/src/components/WwwFrame/WwwPage.vue
+++ b/src/components/WwwFrame/WwwPage.vue
@@ -100,7 +100,6 @@ export default {
 @import 'settings';
 
 .www-page {
-	height: 100%;
 	display: flex;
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/WwwPage.vue
```diff
@@ -103,6 +103,7 @@ export default {
 	height: 100%;
 	display: flex;
 	flex-flow: column nowrap;
+	min-height: 100vh;
```
emuvente, line 106: I see that the `height: 100%` rule is still included. Is this working to solve the issue you were seeing?

emuvente (approved):
Looks good! Please be sure to check a few pages in UI to confirm that min-height change hasn't broken anything, like /lend-by-category and /error

PR #5413, opened 2024-07-25
feat: dedications added to lender loans list

roger-in-kiva:
- dedications added to lender loans list

Screenshot 2024-07-25 at 10 30 36 a m

Kiva dedication:
Screenshot 2024-07-25 at 10 35 24 a m

One recipient dedication:
Screenshot 2024-07-25 at 10 35 33 a m

Multiple recipients dedication:
Screenshot 2024-07-25 at 10 35 42 a m

### DIFF
--- a/src/components/LoanCards/NewHomePageLoanCard.vue
+++ b/src/components/LoanCards/NewHomePageLoanCard.vue
@@ -122,6 +122,15 @@
 				</figure>
 			</div>
 		</router-link>
+		<div v-if="dedications.length > 0">
+			💚
+			<router-link
+				:to="`/dedication/${loanId}`"
+				class="data-hj-suppress"
+			>
+				{{ dedicationsCopy }}
+			</router-link>
+		</div>
 	</div>
 </template>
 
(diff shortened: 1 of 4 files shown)


eddieferrer:
Maybe it should say "Anonymous" instead of "anonymized" so it looks a little more formatted like the other dedications

roger-in-kiva:
> Maybe it should say "Anonymous" instead of "anonymized" so it looks a little more formatted like the other dedications

Agreed but this is coming from data, you can see current case here: https://www.kiva.org/lender/alexandra6339?pageID=4

Maybe just capitalizing that text.

src/components/LoanCards/NewHomePageLoanCard.vue
```diff
@@ -249,6 +269,21 @@ export default {
...
+				dedications.push({
+					recipientName: 'Kiva',
+				});
+			}
+			return dedications;
+		},
+		dedicationsCopy() {
+			return this.dedications.length > 1
+				? 'Multiple recipients'
+				: this.dedications[0].recipientName;
```
dyersituations, line 285: Looks like we still need to check length here, since `dedications` can be empty.

src/util/loanUtils.js
```diff
@@ -270,7 +271,7 @@ export function watchLoanCardData({
 }
 
 export function readLoanFragment({
-	apollo, loanId, fragment
+	apollo, loanId, fragment, publicId
```
dyersituations, line 274: Does adding this new variable cause any issues with the other usages of these utils? As in, is loan data still successfully pulled from the cache and watched?

src/components/LoanCards/NewHomePageLoanCard.vue
roger-in-kiva: If `dedications` is empty the section is not rendered in line 125 so this computed prop is not called.

src/util/loanUtils.js
roger-in-kiva: This is also used in `KvClassicLoanCardContainer` that I tested and working fine.

PR #5416, opened 2024-07-25
fix: auto lending advance settings incorrectly displays partners

dyersituations:
https://kiva.atlassian.net/browse/MP-276

- The partners shown were limited to the 100 returned by default from the monolith -> this must have been a bug since that limitation was added to the monolith endpoints
- Now all non-direct partners are shown in the autolending settings modal
image


mcstover (approved):
"Most" of the partners ;)

PR #5418, opened 2024-07-26
fix: lender profile components fix

roger-in-kiva:
- lender profile components fix

### DIFF
--- a/src/components/LenderProfile/LenderTeamsList.vue
+++ b/src/components/LenderProfile/LenderTeamsList.vue
@@ -17,13 +17,13 @@
 					:href="`/team/${team.teamPublicId}`"
 				>
 					<kv-material-icon
-						v-if="!team.image.url"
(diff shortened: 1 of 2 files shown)


src/components/LenderProfile/LenderTeamsList.vue
```diff
@@ -17,13 +17,13 @@
 					:href="`/team/${team.teamPublicId}`"
 				>
 					<kv-material-icon
-						v-if="!team.image.url"
+						v-if="!team.image?.url"
```
dyersituations, line 20: I thought optional chaining didn't work in a Vue 2 template and that we had to use a computed value. Is this working locally when you open the page?

roger-in-kiva: Oh nice call... As this is a loop I'll add a method to get the image url.

dyersituations (approved):

PR #5419, opened 2024-07-26
feat: remove clickable tags exp

christian14b:
Removing clickable tags experiment related code


dyersituations (approved):
Approved assuming the component library PR is also coming 😄

mcstover (approved):

PR #5423, opened 2024-07-30
feat: lender profile invitees

roger-in-kiva:
- Lender invitees added to lender profile

Screenshot 2024-07-30 at 12 59 39 p m

### DIFF
--- a/src/components/LenderProfile/LenderInviteesList.vue
+++ b/src/components/LenderProfile/LenderInviteesList.vue
@@ -1,0 +1,158 @@
+<template>
+	<section v-if="lenderInvitees.length > 0" class="tw-my-8">
+		<h4 class="data-hj-suppress tw-mb-1">
+			{{ lenderInviteesTitle }}
+		</h4>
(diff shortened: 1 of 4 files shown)


src/components/LenderProfile/LenderInviteesList.vue
```diff
@@ -0,0 +1,158 @@
...
+				const { data } = await this.apollo.query({
+					query: lenderInviteesQuery,
+					variables: {
+						publicId: this.publicId,
+						limit: this.inviteesLimit,
+						offset: this.inviteesOffset
+					},
+				});
+
+				this.lenderInvitees = data.community?.lender?.invitees?.values ?? [];
```
dyersituations, line 122: Does `data` need optional chaining here or is it always defined? Same with the line after this.

roger-in-kiva: it's always defined above from the query, like other cases, loans, teams, dedications.

dyersituations (approved):

PR #5424, opened 2024-07-30
feat: badges section for lender page

christian14b:
image

### DIFF
--- a/src/components/LenderProfile/LenderBadges.vue
+++ b/src/components/LenderProfile/LenderBadges.vue
@@ -21,7 +21,7 @@ export default {
 	props: {
 		completedAchievements: {
 			type: Array,
-			default: () => []
+			default: () => ([])
 		},
 		totalPossibleBadges: {
 			type: Number,
--- a/src/pages/LenderProfile/LenderProfile.vue
+++ b/src/pages/LenderProfile/LenderProfile.vue
@@ -19,7 +19,17 @@
 				:lender-info="lenderInfo"
 			/>
 
+			<lender-dedications-list
+				:public-id="publicId"
+				:lender-info="lenderInfo"
+			/>
+
 			<lender-teams-list
+				:public-id="publicId"
+				:lender-info="lenderInfo"
+			/>
+
+			<lender-invitees-list
 				:public-id="publicId"
 				:lender-info="lenderInfo"
 			/>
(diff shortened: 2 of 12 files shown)


src/components/LenderProfile/LenderBadges.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+	<div>
+		<h4>{{ lenderName }}'s Badges</h4>
```
roger-in-kiva, line 3: We are adding a `no-name` value like here: https://github.com/kiva/ui/blob/0d5bf7ba307a528a7c94956dc3a24869867d104e/src/components/LenderProfile/LenderTeamsList.vue#L89

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -21,6 +21,12 @@
 			<lender-stats
 				:lender-info="lenderInfo"
 			/>
+
+			<lender-badges
```
roger-in-kiva, line 25: Looks like badges is between Loans and Dedications: https://github.com/kiva/kiva/blob/2fbbc6e05c4d0649e618bb499c4bb5c6789216bc/httpdocs/views/Lender/LenderView.tpl#L66

src/pages/Portfolio/LendingStats/BadgesList.vue
```diff
@@ -0,0 +1,119 @@
...
+export default {
+	name: 'BadgesList',
+	inject: ['apollo'],
+	components: {
+		KvLoadingPlaceholder,
+	},
+	props: {
+		completedAchievements: {
+			type: Array,
+			default: () => []
```
dyersituations, line 55: default: () => ([])

src/pages/Portfolio/LendingStats/BadgesList.vue
```diff
@@ -0,0 +1,119 @@
...
+				}
+			},
+			immediate: true
+		}
+	},
+	methods: {
+		loadBadgesContent() {
+			if (this.completedChallengeKeys !== '') {
+				const filterFields = `fields.key[in]=${this.completedChallengeKeys}`;
+				this.apollo.query({
```
dyersituations, line 95: Try catch around `query`

dyersituations: Or maybe `.catch`, since you're using chaining instead of async await. Just make sure that a thrown exception here is handled.

src/components/LenderProfile/LenderBadges.vue
```diff
@@ -0,0 +1,49 @@
...
+
+export default {
+	name: 'LenderBadges',
+	components: {
+		BadgesList
+	},
+	props: {
+		completedAchievements: {
+			type: Array,
+			default: () => []
```
dyersituations, line 24: default: () => ([])

dyersituations (approved):

PR #5431, opened 2024-08-02
fix: redirect to /checkout/thanks if optedIn is set

christian14b:


src/pages/Checkout/PostPurchase.vue
```diff
@@ -41,7 +41,7 @@ export default {
 						let successPath = route.query.challenge
 							? `/checkout/thanks/${route.query.challenge}` : '/thanks';
```
mcstover, line 42: @emuvente do you know of any reason we're still redirecting through `/thanks` instead of using `/checkout/thanks` as the default these days?

mcstover: Yep and it's certainly fine to leave for now, just wanted to check in on the topic as we seem to be making interesting carve outs that are prob not really necessary if we weren't passing through that old url which does eventually redirect back to `/checkout/thanks` anyhow...

mcstover (approved):

PR #5436, opened 2024-08-07
fix: mfa api token auth requires using authorize call, not checkSession MP-479

emuvente:
Auth0 will require MFA authentication via redirect after Aug 30 for MFA API tokens, so silent auth calls like checkSession need to be replaced. Since it's a redirect for an implicit grant, we need to parse the token from the hash on an approved callback url (like /process-browser-auth). /process-browser-auth wasn't being used anymore (previously only used for popup window auth) so I repurposed that to redirect to a destination after confirming state.

Depends on https://github.com/kiva/kiva/pull/12870

### DIFF
--- a/src/pages/ProcessBrowserAuth.vue
+++ b/src/pages/ProcessBrowserAuth.vue
@@ -5,42 +5,21 @@
 <script>
 import store2 from 'store2';
 
-function checkHashSuccess(hash) {
-	if (hash.indexOf('error') > -1) {
-		return false;
-	}
-	if (hash.indexOf('access_token') === -1
-		&& hash.indexOf('id_token') === -1
-		&& hash.indexOf('refresh_token') === -1) {
-		return false;
-	}
-	return true;
-}
-
 export default {
 	name: 'ProcessBrowserAuth',
 	inject: ['kvAuth0'],
 	mounted() {
-		const hashKey = 'auth0.browser_hash';
+		const { state } = this.$route.query;
+		if (state) {
+			const auth0State = store2.session('auth0.state');
+			if (auth0State === state) {
+				const redirect = store2.session('auth0.redirect');
 
-		if (window.location.hash) {
-			const { hash } = window.location;
+				store2.session.remove('auth0.state');
+				store2.session.remove('auth0.redirect');
 
-			if (checkHashSuccess(hash)) {
-				// store hash for after post-auth redirect
-				store2.session(hashKey, hash);
-				// post-auth redirect
-				window.location = '/authenticate/ui?doneUrl=/process-browser-auth';
-			} else {
-				// some problem occured, so close the window and let normal error handling take over
-				this.kvAuth0.popupCallback({ hash });
+				this.$router.push(`${redirect}${window.location.hash}`);
 			}
-		} else {
-			// fetch & erase stored hash
-			const hash = store2.session(hashKey);
-			store2.session.remove(hashKey);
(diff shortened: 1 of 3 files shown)


src/util/KvAuth0.js
```diff
@@ -236,27 +272,31 @@ export default class KvAuth0 {
...
+				.then(() => this[parseMfaHash]())
+				.then(() => {
+					if (this.mfaManagementToken) {
 						resolve(this.mfaManagementToken);
+					} else {
+						// If we still don't have a token, try to get one
+						// Ensure browser clock is correct before fetching the token
+						syncDate().then(() => {
+							// Store the current URL in session storage to redirect back to it after MFA
+							const state = storeRedirectState();
```
dyersituations, line 289: Do we need to ensure this async method completes before calling `authorize` on the next line?

src/pages/ProcessBrowserAuth.vue
```diff
@@ -5,42 +5,21 @@
...
-	}
-	return true;
-}
-
 export default {
 	name: 'ProcessBrowserAuth',
 	inject: ['kvAuth0'],
 	mounted() {
-		const hashKey = 'auth0.browser_hash';
+		const { state } = this.$route.query;
```
dyersituations, line 12: Will this page just check called automatically with our current auth flow? I didn't see it being called explicitly in this change set.

src/util/KvAuth0.js
emuvente: Yes, good catch

src/pages/ProcessBrowserAuth.vue
emuvente: Yeah, this page is the current `redirectUri`

dyersituations (approved):

dyersituations (approved):

src/pages/ProcessBrowserAuth.vue
```diff
@@ -5,42 +5,21 @@
 <script>
```
mcstover, line 5: This endpoint is currently set as the clientEntry redirect uri and shows up in the login authorize params during login. Will these updates continue to serve that setup?

src/util/KvAuth0.js
```diff
@@ -285,7 +326,11 @@ export default class KvAuth0 {
 		this[sessionPromise] = new Promise(resolve => {
 			// Ensure browser clock is correct before checking session
 			syncDate().then(() => this[initWebAuth]()).then(() => {
-				this.webAuth.checkSession({}, (err, result) => {
+				this.webAuth.checkSession({
+					audience: this.audience,
+					responseType: 'token id_token',
+					scope: this.scope,
+				}, (err, result) => {
```
mcstover, line 333: Will we also need to update CPS, how might this differ in that world where it's client side auth only?

emuvente: CPS does not have any MFA management pages or features, so there's nothing to update there. MFA has always client-side auth only, so nothing is changing there.

src/pages/ProcessBrowserAuth.vue
emuvente: This was only used by the popup login, which wasn't used. Server-side authentication goes to the /process-ssr-auth route, not this one. The checkSession calls appear to work by passing messages through the iframes, not relying on the redirect, but that's hard to confirm with looking at the code. I want to merge this into dev to test that out.

PR #5438, opened 2024-08-12
feat: loans map added to lender profile

roger-in-kiva:
- loans map added to lender profile

Some cases:
Screenshot 2024-08-12 at 1 53 37 p m
Screenshot 2024-08-12 at 1 53 50 p m
Screenshot 2024-08-12 at 1 54 03 p m
Screenshot 2024-08-12 at 2 04 41 p m

Mouse over country:
Screenshot 2024-08-12 at 1 54 36 p m

Mouse over fundraising loans dot:
Screenshot 2024-08-12 at 1 54 46 p m

Hidden fundraising loans dots:
Screenshot 2024-08-12 at 2 00 18 p m

### DIFF
--- a/src/components/Kv/KvMap.vue
+++ b/src/components/Kv/KvMap.vue
@@ -12,6 +12,9 @@
 <script>
 import { createIntersectionObserver } from '@/util/observerUtils';
 import testDelayedGlobalLibrary from '@/util/timeoutUtils';
+import countriesBorders from '@/assets/data/components/LenderProfile/countries-borders.json';
+import { getIntervals, mapColors } from '@/components/LenderProfile/LenderMap';
+import kvTokensPrimitives from '~/@kiva/kv-tokens/primitives.json';
 
 export default {
 	name: 'KvMap',
@@ -124,6 +127,44 @@ export default {
 			type: Number,
 			default: 4
 		},
+		/**
+		 * Show the zoom control
+		 */
+		showZoomControl: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		 * Allow dragging of the map
+		 */
+		allowDragging: {
+			type: Boolean,
+			default: false
+		},
+		/**
+		 * Show labels on the map
+		 * Working for leaflet only
+		 */
+		showLabels: {
+			type: Boolean,
+			default: true
+		},
+		/**
+		 * Lender data for the map
+		 * Working for leaflet only
+		 */
+		countriesData: {
+			type: Array,
+			default: () => ([]),
+		},
+		/**
+		 * Show fundraising loans
+		 * Working for leaflet only
+		 */
+		showFundraisingLoans: {
+			type: Boolean,
+			default: false
+		},
 	},
 	data() {
 		return {
@@ -162,7 +203,13 @@ export default {
 			if (prev === null && this.lat && !this.mapLibreReady && !this.leafletReady) {
 				this.initializeMap();
 			}
-		}
+		},
+		showFundraisingLoans() {
+			if (this.mapInstance) {
+				this.mapInstance.remove();
+				this.initializeLeaflet();
+			}
+		},
 	},
 	mounted() {
 		if (!this.mapLibreReady && !this.leafletReady) {
@@ -260,8 +307,8 @@ export default {
 				center: [this.lat, this.long],
 				zoom: this.initialZoom || this.zoomLevel,
 				// todo make props for the following options
-				dragging: false,
-				zoomControl: false,
+				dragging: this.allowDragging,
+				zoomControl: this.showZoomControl,
 				animate: true,
 				scrollWheelZoom: false,
 				doubleClickZoom: false,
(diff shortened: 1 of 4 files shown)


src/components/LenderProfile/LenderMap.vue
```diff
@@ -0,0 +1,199 @@
...
+		</div>
+	</section>
+</template>
+
+<script>
+import numeral from 'numeral';
+import KvMap from '@/components/Kv/KvMap';
+import KvCheckbox from '~/@kiva/kv-components/vue/KvCheckbox';
+
+export const getIntervals = (min, max, nbIntervals) => {
```
dyersituations, line 75: It would be worth adding a ticket to the epic to extract this to a util and unit test it.

src/components/LenderProfile/LenderMap.vue
```diff
@@ -0,0 +1,199 @@
...
+		lenderMapTitle() {
+			return this.lenderInfo?.name
+				? `${this.lenderInfo.name}'s Lending Activity by Country`
+				: 'Lending Activity by Country';
+		},
+		countriesData() {
+			return (this.lenderInfo?.statsPerCountry?.values ?? []).map(stat => ({
+				label: stat.country?.name ?? '',
+				value: stat.loanCount ?? '',
+				lat: stat.country?.geocode?.latitude ?? 0,
```
dyersituations, line 142: Are we generally guaranteed the the stat data exists or should we be filtering the array before we map the data?

src/components/LenderProfile/LenderMap.vue
dyersituations: Also, what specifically are the intervals here? It would probably be good to add a jsdoc comment to this method to explain what it's generating.

src/components/Kv/KvMap.vue
```diff
@@ -12,6 +12,9 @@
 <script>
 import { createIntersectionObserver } from '@/util/observerUtils';
 import testDelayedGlobalLibrary from '@/util/timeoutUtils';
+import countriesBorders from '@/assets/data/components/LenderProfile/countries-borders.json';
+import { getIntervals, mapColors } from '@/components/LenderProfile/LenderMap';
```
dyersituations, line 16: It looks like these two methods might better exist here (or in a shared util file).

src/components/Kv/KvMap.vue
```diff
@@ -260,21 +307,64 @@ export default {
...
+				).addTo(this.mapInstance);
+
+				this.countriesData.forEach(country => {
+					if (country.numLoansFundraising > 0 && this.showFundraisingLoans) {
+						const circle = L.circle([country.lat, country.long], {
+							color: kvTokensPrimitives.colors.black,
+							weight: 1,
+							fillColor: kvTokensPrimitives.colors.brand[900],
+							fillOpacity: 1,
+							radius: 130000,
```
dyersituations, line 356: Why a large number for the `radius`?

src/components/Kv/KvMap.vue
```diff
@@ -12,6 +12,9 @@
 <script>
 import { createIntersectionObserver } from '@/util/observerUtils';
 import testDelayedGlobalLibrary from '@/util/timeoutUtils';
+import countriesBorders from '@/assets/data/components/LenderProfile/countries-borders.json';
+import { getIntervals, mapColors } from '@/components/LenderProfile/LenderMap';
+import kvTokensPrimitives from '~/@kiva/kv-tokens/primitives.json';
```
mcstover, line 17: I think everything in this file should be in https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-components/vue/KvMap.vue instead of UI. Havint the country borders json file would be helpful at a global context roo.

src/components/Kv/KvMap.vue
```diff
@@ -308,6 +398,82 @@ export default {
...
+				}
+			}
+
+			return kvTokensPrimitives.colors.gray[300];
+		},
+		onEachCountryFeature(feature, layer) {
+			const loansString = feature.lenderLoans
+				? `${feature.lenderLoans} loan${feature.lenderLoans > 1 ? 's' : ''}`
+				: '0 loans';
+			const countryString = `${feature.properties.ADMIN} <br/> ${loansString}`;
```
dyersituations, line 443: What is `feature.properties.ADMIN`?

src/components/Kv/KvMap.vue
```diff
@@ -308,6 +398,82 @@ export default {
...
+		resetHighlight(e) {
+			const layer = e.target;
+			const { feature } = layer;
+
+			layer.setStyle({
+				fillColor: this.getCountryColor(feature.lenderLoans),
+			});
+		},
+		circleMapClicked(countryIso) {
+			this.$router.push({
```
dyersituations, line 470: Should we also be emitting here so we can send analytics events? Or call analytics here directly?

src/components/Kv/KvMap.vue
dyersituations: The related intervals util could be moved as well and unit tested in the library.

mcstover: In other words KvMap has been migrated to kv-ui-elements and all enhancements and additions should go there, then that file imported for use in the page instead of a UI only file.

mcstover:
@roger-in-kiva @dyersituations I don't think we can use the country-borders json file in it's current state. I'm showing that file is 7.5 mb in size... Was there not a way to use the dynamic data from the map library to highlight countries as needed instead of requiring a file of this size?

roger-in-kiva:
> @roger-in-kiva @dyersituations I don't think we can use the country-borders json file in it's current state. I'm showing that file is 7.5 mb in size... Was there not a way to use the dynamic data from the map library to highlight countries as needed instead of requiring a file of this size?

We need the polygons of every country for the case of leaflet usage, for that reason this file is huge...
Let me see if we can use some dynamic data from the map library, if not we need to follow this approach as well.

src/components/Kv/KvMap.vue
roger-in-kiva: Agreed, but the suggestion in the ticket is to use this UI file: https://kiva.atlassian.net/browse/MP-415

src/components/Kv/KvMap.vue
roger-in-kiva: This radius is in meters based on the map scale:
https://leafletjs.com/reference.html#circle-radius

src/components/Kv/KvMap.vue
roger-in-kiva: Yeah, for this scope we have other ticket for analytics events... but if we port this feature to the shared library, we need to add an emit for this.

src/components/LenderProfile/LenderMap.vue
roger-in-kiva: There are some cases for private/anonymous returning null, when we are returning an empty array.
In other cases we are guaranteed that stat data exists.

src/components/LenderProfile/LenderMap.vue
```diff
@@ -0,0 +1,180 @@
+<template>
```
dyersituations, line 1: This file can be removed now.

dyersituations (approved):

mcstover (approved):
👍

PR #5442, opened 2024-08-13
feat: click blurred badge to fullscreen mp 597

christian14b:
First iteration of fullscreen when clicking reveal button:

https://github.com/user-attachments/assets/ff71c90d-7333-4474-a6a0-de5cd8d0cb0d

### DIFF
--- a/src/components/Thanks/BadgesCustomization.vue
+++ b/src/components/Thanks/BadgesCustomization.vue
@@ -11,8 +11,8 @@
 				<h1
 					class="tw-mb-1 tw-transition-all tw-duration-1000 tw-ease-in-out"
 					:class="{
-						'tw-relative tw-z-2 tw-text-black': newBgMounted,
-						'tw-text-white': !newBgMounted
+						'tw-relative tw-z-2 tw-text-black': badgeBlurRevealCompleted,
+						'tw-text-white': !badgeBlurRevealCompleted
 					}"
 				>
 					{{ headerTitle }}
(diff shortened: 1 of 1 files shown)


dyersituations:
@christian14b looking good, but based on the Figma, I think the badge is supposed to fade in, rather than immediate show unblurred. See this part of the mocks. Notice that the badge is partially blurred still as the full-page element animates open. I think both animations need to happen at the same time, unblurring and opening.

.storybook/stories/AnimatedStars.stories.js
```diff
@@ -0,0 +1,19 @@
...
+};
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  components: { AnimatedStars },
+  template: '<div class="tw-bg-stone-1 tw-w-12 tw-h-12 tw-relative"><animated-stars v-bind="$props" /></div>',
+});
+
+export const Default = Template.bind({});
+Default.args = {
```
dyersituations, line 15: Looks like this line can be removed.

src/components/Thanks/BadgesCustomization.vue
```diff
@@ -252,6 +346,8 @@ export default {
 			isBlurred: true,
 			isMobileLayout: false,
 			wiggle: false,
+			badgeRevealed: false,
+			newBgMounted: false,
```
dyersituations, line 350: I think a name like `badgeBlurRevealCompleted` might make more sense. Using "Mounted" makes me think it's related to Vue `mounted`, so I had to read through to figure out what this meant.

dyersituations (approved):

PR #5451, opened 2024-08-19
feat: loading states adding to lender profile sections

roger-in-kiva:
- Loading states adding to lender profile sections

Reqs: https://kiva.atlassian.net/browse/MP-618

### DIFF
--- a/src/components/LenderProfile/LenderBadges.vue
+++ b/src/components/LenderProfile/LenderBadges.vue
@@ -44,7 +44,8 @@ export default {
 		},
 		isLoading: {
 			type: Boolean,
-			default: false
+			required: true,
+			default: true
 		}
 	},
 	computed: {
--- a/src/components/LenderProfile/LenderBadges.vue
+++ b/src/components/LenderProfile/LenderBadges.vue
@@ -1,5 +1,5 @@
 <template>
-	<async-lender-section @visible="isLoading = false;">
+	<async-lender-section @visible="() => isLoading = false">
 		<section v-if="completedAchievements.length > 0">
 			<h4
 				v-if="!isLoading"
--- a/src/components/LenderProfile/LenderBadges.vue
+++ b/src/components/LenderProfile/LenderBadges.vue
@@ -1,6 +1,15 @@
 <template>
 	<section v-if="completedAchievements.length > 0">
-		<h4>{{ badgesTitle }}</h4>
+		<h4
+			v-if="!isLoading"
+			class="data-hj-suppress"
+		>
+			{{ badgesTitle }}
+		</h4>
+		<kv-loading-placeholder
+			v-else
+			style="height: 30px; width: 250px;"
+		/>
 		<badges-list
 			class="tw-my-4"
 			:completed-achievements="completedAchievements"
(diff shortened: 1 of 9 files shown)


src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -11,35 +11,42 @@
 			<lender-loans-list
 				:public-id="publicId"
 				:lender-info="lenderInfo"
+				@loans-loaded="isLoading = false"
```
dyersituations, line 14: `() => isLoading = false` for the Vue migration.

src/components/LenderProfile/LenderLoansList.vue
```diff
@@ -1,18 +1,25 @@
 <template>
 	<section v-if="lenderLoans.length > 0" class="tw-my-8" id="lender-loans">
-		<h4 class="data-hj-suppress tw-mb-1">
-			{{ lenderLoansTitle }}
-		</h4>
-		<p class="tw-mb-2">
-			{{ showedLoans }}
-		</p>
+		<div v-if="totalCount > 0">
```
dyersituations, line 3: Should this be a loading flag instead of `totalCount`? Since what if the lender has no loans, the loaders will never disappear.

src/components/LenderProfile/LenderDedicationsList.vue
```diff
@@ -135,12 +153,14 @@ export default {
...
-		this.fetchLenderDedications();
-	},
 	created() {
 		this.pageQuery = _get(this.$route, 'query');
 		this.updateFromParams(this.pageQuery);
-	}
+	},
+	watch: {
+		isLoading() {
+			if (!this.isLoading) this.fetchLenderDedications();
```
dyersituations, line 163: It looks like the lazy loading pattern here isn't quite lazy loading but instead other components all load once the `LenderLoansList` is done loading. As mentioned in the ticket, you can use the `@visible="fetchAsyncData"` such as in `YourTeams.vue` to do lazy loading.

src/components/LenderProfile/LenderInviteesList.vue
```diff
@@ -81,10 +94,15 @@ export default {
 			type: Object,
 			required: true,
 		},
+		isLoading: {
+			type: Boolean,
+			required: true,
+			default: true,
```
dyersituations, line 100: The `default` props aren't needed on these props, since `required` is now there.

src/components/LenderProfile/LenderLoansList.vue
roger-in-kiva: From line 2, we are not showing the whole section if the lender has no loans.

src/components/LenderProfile/LenderBadges.vue
```diff
@@ -1,33 +1,37 @@
...
-			style="height: 30px; width: 250px;"
-		/>
-		<badges-list
-			class="tw-my-4"
-			:completed-achievements="completedAchievements"
-			:total-possible-badges="totalPossibleBadges"
-			:is-loading="isLoading"
-		/>
-	</section>
+	<async-lender-section @visible="isLoading = false;">
```
dyersituations, line 2: I think these need to be `() => isLoading = false` for the Vue migration.

roger-in-kiva: Gotcha. Just curious where is the vue3 spec for this.

dyersituations: It was just something I ran into during the migration.

dyersituations (approved):

PR #5453, opened 2024-08-19
fix: change refugee badge name to plural

christian14b:


dyersituations (approved):
Thanks for the quick response to the product review!

PR #5456, opened 2024-08-21
feat: badge details screen mp 599

christian14b:
- New Screen for badge's detail (Missing carousel work in another ticket)
- Refactoring so transition between the revealed badge state (New component) and discover all badges screen be smooth. Before this, there was a jump between these two screens

https://github.com/user-attachments/assets/b1714088-407a-498a-910d-9387349035d7

### DIFF
--- a/src/components/Thanks/BadgesCustomization.vue
+++ b/src/components/Thanks/BadgesCustomization.vue
@@ -20,11 +20,11 @@
 			<RevealedBadge
 				key="revealed-badge-screen"
 				@show-discover-badges="showDiscoverBadges"
-				v-if="newScreenSteps === 1"
+				v-else-if="newScreenSteps === 1"
 				:is-guest="isGuest"
 			/>
 			<DiscoverSection
-				v-if="newScreenSteps === 2"
+				v-else-if="newScreenSteps === 2"
 				key="discover-screen"
 				:selected-loan-region="selectedLoanRegion"
 				:is-guest="isGuest"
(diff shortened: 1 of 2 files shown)


src/components/Thanks/BadgesCustomization.vue
```diff
@@ -1,330 +1,63 @@
...
+				key="discover-screen"
 				:selected-loan-region="selectedLoanRegion"
 				:is-guest="isGuest"
-				@back-to-earned-badge="() => discoverBadges = false"
+				:badges="randomSortedBadges"
+				@back="() => newScreenSteps -= 1"
+				@select-badge="selectBadge"
+			/>
+			<DetailSection
+				v-if="newScreenSteps === 3"
```
dyersituations, line 36: `v-else-if` on these transition children. Vue 3 throws errors when it thinks there could be more than one child of a transition.

src/components/Thanks/BadgesCustomization.vue
```diff
@@ -351,88 +84,79 @@ export default {
 	},
 	data() {
 		return {
-			openCreateAccount: false,
-			openOrderConfirmation: false,
-			openShareModule: false,
-			mdiChevronDown,
-			imageRequire,
 			isBlurred: true,
 			isMobileLayout: false,
```
dyersituations, line 88: Looks like this can be removed now?

dyersituations (approved):

PR #5461, opened 2024-08-27
feat: update user preferences with goal mp 601

christian14b:
Create user preference if user does not have. Update it with goal information and redirect to portfolio to show tip message.

### DIFF
--- a/src/components/Thanks/Badges/DetailSection.vue
+++ b/src/components/Thanks/Badges/DetailSection.vue
@@ -237,7 +237,7 @@ export default {
 					this.$route.push({ path: '/portfolio', query: { goal_saved: true } });
 				}
 			} catch (error) {
-				throw new Error(error);
+				console.error(error);
 			}
 		},
 	}


src/components/Thanks/Badges/DetailSection.vue
```diff
@@ -164,7 +169,75 @@ export default {
...
+			return response?.data?.my?.createUserPreferences?.id;
+		},
+		async setAsGoal() {
+			try {
+				let updateUserPreferencesId = this.userPreferences?.id ?? null;
+
+				if (!updateUserPreferencesId) {
+					updateUserPreferencesId = await this.createUserPreferences();
+				}
+				const currentPreferences = this.userPreferences?.preferences ?? {};
```
dyersituations, line 202: Should this default object be `{ id:  updateUserPreferencesId }`?

src/components/Thanks/Badges/DetailSection.vue
```diff
@@ -164,7 +169,75 @@ export default {
...
+				if (response.errors) {
+					throw new Error(
+						response.errors[0].extensions.code
+						|| response.errors[0].message
+					);
+				} else {
+					this.$route.push({ path: '/portfolio', query: { goal_saved: true } });
+				}
+			} catch (error) {
+				throw new Error(error);
```
dyersituations, line 238: What happens if an error is thrown here? Is a tooltip shown to the user?

dyersituations: I see there's now a tooltip, but what happens here when the `error` is thrown a second time? How does the app handle it?

dyersituations: In that case, this throw should probably be a console.error then.

src/components/Thanks/Badges/DetailSection.vue
dyersituations: Are more changes needed here?

dyersituations (approved):

PR #5467, opened 2024-08-28
feat: lender stats data import updated

roger-in-kiva:
- Lender stats data import updated


src/components/LenderProfile/LenderStats.vue
```diff
@@ -35,11 +35,13 @@ import AsyncLenderSection from './AsyncLenderSection';
 
 export default {
 	name: 'LenderStats',
-	inject: ['apollo', 'cookieStore'],
 	props: {
-		lenderInfo: {
+		publicId: {
```
dyersituations, line 39: Looks like this prop can be removed.

dyersituations (approved):

dyersituations (approved):

PR #5471, opened 2024-08-29
feat: handle goal in portfolio for all users in experiment

christian14b:
Moving logic to portfolio to save goals for guests and logged in users.

### DIFF
--- a/src/components/Thanks/Badges/DetailSection.vue
+++ b/src/components/Thanks/Badges/DetailSection.vue
@@ -86,7 +86,6 @@
 
 <script>
 import { mdiChevronLeft, mdiCheckCircleOutline } from '@mdi/js';
-import { gql } from '@apollo/client';
 import KvMaterialIcon from '~/@kiva/kv-components/vue/KvMaterialIcon';
 import KvButton from '~/@kiva/kv-components/vue/KvButton';
 import KvCarousel from '~/@kiva/kv-components/vue/KvCarousel';
--- a/src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
+++ b/src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
@@ -76,17 +76,13 @@ export default {
 			showTeamChallenge: false,
 			teamsChallengeEnable: false,
 			allowedTeams: [],
+			userPreferences: null,
 		};
 	},
 	mixins: [badgeGoalMixin],
 	apollo: {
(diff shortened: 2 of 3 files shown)


src/components/Thanks/Badges/DetailSection.vue
dyersituations, line 1: I can't comment on the line directly, but this file has an unused import now.

src/plugins/badge-goal-mixin.js
```diff
@@ -0,0 +1,63 @@
...
+							updateUserPreferences(id: $updateUserPreferencesId, userPreferences: {
+								preferences: $preferences
+						}) {
+								id
+								preferences
+							}
+						}
+				}`;
+
+				const updateUserPreferences = this.apollo.mutate({
```
dyersituations, line 49: Looks like an `await` is needed here.

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -115,7 +122,14 @@ export default {
 		this.showTeamChallenge = teamsChallengeEnable && this.allowedTeams.length > 0;
 
 		if (this.$route?.query?.goal_saved) {
-			this.$showTipMsg('Goal saved');
+			const badgeName = this.route?.query?.goal_saved ?? '';
+			const userPreferences = portfolioQueryData.my?.userPreferences ?? null;
+
+			this.storeGoal({ userPreferences, badgeName }).then(() => {
+				this.$showTipMsg('Goal saved');
```
dyersituations, line 129: Should this logic be happening on `mounted`? Due to the tooltip not being needed server-side.

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -77,9 +78,15 @@ export default {
 			allowedTeams: [],
 		};
 	},
+	mixins: [badgeGoalMixin],
 	apollo: {
-		async preFetch(config, client) {
-			return client.query({ query: portfolioQuery });
+		async preFetch(config, client, { route }) {
```
dyersituations, line 83: Related to another comment, if we moved the saving logic to the client, this can also be moved to mounted.

dyersituations (approved):
@christian14b sorry, I thought I had approved this already. Thanks for the reminder!

PR #5472, opened 2024-08-30
feat: tracking events added to lender profile

roger-in-kiva:
- tracking events added to lender profile
 
Reqs: https://kiva.atlassian.net/browse/MP-570

### DIFF
--- a/src/components/LoanCards/NewHomePageLoanCard.vue
+++ b/src/components/LoanCards/NewHomePageLoanCard.vue
@@ -124,12 +124,12 @@
 		</router-link>
 		<div v-if="dedications.length > 0">
 			💚
-			<router-link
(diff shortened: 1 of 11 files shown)


src/components/LoanCards/NewHomePageLoanCard.vue
```diff
@@ -339,6 +339,10 @@ export default {
 				this.isLoading = false;
 			}
 		},
+		dedicationClickEvent() {
+			this.$emit('dedication-click', { loanId: this.loanId, dedicationCopy: this.dedicationsCopy });
```
dyersituations, line 343: Does this get logged successfully with the router push right after?

roger-in-kiva: Works fine locally.

PR #5475, opened 2024-09-04
fix: add number of loans within badge category in anaylitics

christian14b:
Number of loans within the badge category. Filters used based on `loan-channel-query-map.js` file

### DIFF
--- a/src/components/Thanks/BadgesCustomization.vue
+++ b/src/components/Thanks/BadgesCustomization.vue
@@ -204,11 +204,11 @@ export default {
 					let matches = true;
 
 					if (badge.category.themes) {
-						matches = matches && badge.category.themes.every(theme => loan.themes.includes(theme));
+						matches = matches && badge.category.themes.some(theme => loan.themes.includes(theme));
 					}
 
 					if (badge.category.tags) {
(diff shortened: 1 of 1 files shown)


src/components/Thanks/BadgesCustomization.vue
```diff
@@ -176,6 +198,37 @@ export default {
 		showNewBg() {
 			this.newBgActive = true;
 		},
+		countLoansInCategories() {
+			this.defaultSortBadges.forEach((badge, idx) => {
+				this.loans.forEach(loan => {
+					let matches = true;
```
dyersituations, line 204: I think `matches` should probably default to `false`

dyersituations: Is there a situation where a loan will match none of the conditions? In that case, the default would be `true`, which would be incorrect, right? Maybe I'm reading the logic wrong.

dyersituations: Ah got it, so we'll never currently encounter a situation where we'd get `true` to fall through accidentally?

dyersituations (approved):

PR #5479, opened 2024-09-05
feat: use FLSS for live loan image filter searches when supported MP-732

emuvente:
Use FLSS for all filter searches that it supports, with a fallback to the legacy loan search if using a sort option that only works there.

### DIFF
--- a/server/util/live-loan/live-loan-fetch.js
+++ b/server/util/live-loan/live-loan-fetch.js
@@ -244,7 +244,7 @@ const findFilterOption = (options, name, value) => {
 
 // Takes a string like: "sort_expiringSoon,gender_female"
 // and extracts the sort option, e.g. "expiringSoon"
-async function parseSortString(sortString, sortOptions) {
+function parseSortString(sortString, sortOptions) {
 	let sort = null;
 
 	// only try parsing if the input is valid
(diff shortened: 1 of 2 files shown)


server/util/live-loan/live-loan-fetch.js
```diff
@@ -197,12 +233,47 @@ const getFilterArrays = filterString => {
...
+	const option = options.find(o => o?.name?.toLowerCase() === value);
+	if (!option) {
+		warn(`Unknown ${name} "${value}"`);
+	}
+	return option;
+};
+
+// Takes a string like: "sort_expiringSoon,gender_female"
+// and extracts the sort option, e.g. "expiringSoon"
+async function parseSortString(sortString, sortOptions) {
```
dyersituations, line 247: Does this method need `async`?

emuvente: No, looks like it does not

emuvente: https://github.com/kiva/ui/pull/5479/commits/662bb2adc063e9ca7268ccda98ac6bc94bcc9724

dyersituations (approved):
Approved pending passing tests.

roger-in-kiva (approved):

mcstover (approved):

PR #5488, opened 2024-09-09
feat: exclude users that uses kiva card on checkout

christian14b:
Exclude users with gift cards from experiment

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -410,6 +410,17 @@ export default {
 		showNewTYPage() {
 			return !this.landedOnUSLoan && !this.optedIn && this.loans.length > 0;
 		},
+		receiptValues() {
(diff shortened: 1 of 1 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -546,9 +546,10 @@ export default {
 		this.enableMayChallengeHeader = shareChallengeExpData?.version === 'b';
 
 		this.optedIn = data?.my?.communicationSettings?.lenderNews || this.$route.query?.optedIn === 'true';
+		const usedKivaCard = Number(this.receipt?.totals?.kivaCardTotal);
```
dyersituations, line 549: Does this cover only the situation Shehaam mentioned? I thought it was for print cards. Maybe we can't call out the specific situation that she mentioned in Slack.

dyersituations (approved):

PR #5502, opened 2024-09-11
fix: walk up component lib version

dyersituations:
- Turns out the memory leak was in the component library
- Starting the slow process of walking up the version until we get the issue again


mcstover (approved):

mcstover:
Thanks @dyersituations for working to track this down! Looks like we're getting closer!

dyersituations:
> Thanks @dyersituations for working to track this down! Looks like we're getting closer!

It's been a wild ride. Gonna be interesting to see what it was...

PR #5534, opened 2024-09-24
fix: update static asset path for docker build action

emuvente:


dyersituations (approved):

mcstover (approved):
I noticed we aren't filing the newly built assets into the `/js`, `/css` and `/fonts` folders.

It should be ok but we will be just dumping everything in the root of `/static` for Ui now like we do for CPS.

PR #5536, opened 2024-09-24
build: client path should be set in source directory so that /static is the public asset root

emuvente:


emuvente:
@mcstover no worries!

PR #5538, opened 2024-09-24
fix: load lending stats even with gql error

dyersituations:


mcstover (approved):
Handling no data gracefully is def something we need in several parts of the app. Luckily, we've done much better with newer functionality though the `null` loan result can still 500 LBC too.

PR #5545, opened 2024-09-25
chore: remove unused script include

mcstover:
Noticed this was still trying to load and failing in dev. It's been off a while in staging in prod so this removes the code that would previously load the script.


kathrynlovett (approved):
Looks good to me!

dyersituations (approved):

emuvente (approved):
🙌

PR #5556, opened 2024-09-27
feat: adding new condition to redirect to challenge page

christian14b:
As equity badge challenge is live, we needed to add a new condition to show the new thanks page badge experiment


emuvente (changes requested):
I think we should do this in the same if as `if (!this.iwdExpEnabled) {` so that we don't execute the achievements query if we don't need to.

Also, instead of checking if the thanks badge experiment is version b, we should check if the thanks badge experiment is active at all. That way, users in the control group who qualify for the equity badge don't see the challenge page either.

emuvente (approved):
Thanks!

PR #5559, opened 2024-09-27
fix: mp 877 client auth guard

mcstover:
Without this, the auth guard redirect was not being observed during client side navigation. This ensures the redirect occurs based on the route contained in the error.


eddieferrer (approved):

mcstover:
Gonna merge and verify in dev as this may fix a couple oddities with login behavior.

PR #5565, opened 2024-09-30
fix: iwd has false value causing the redirect

christian14b:
OR condition let IWD flag to redirect to old ty page


emuvente (approved):
good catch!

PR #5568, opened 2024-10-01
fix: pass null instead of all, which is not an enum option, in team l…

mcstover:
…eaderboard category field


eddieferrer (approved):
Looks good, I think there might still be a bug on this page when changing to a category and then changing back to "All Categories" the query params do not seem to update

mcstover:
@eddieferrer that's likely as the TeamListing file has a large chunk of if statements which all `return` so some query params may not get updated if earlier conditions are hit...

PR #5571, opened 2024-10-01
fix: remove duplicate id=app

eddieferrer:


mcstover (approved):
Looks like this was the prior state and as you've mentioned that id is now in the template.

I haven't been able to find anything in this code base but externals such as Optimizely and GTM might be using this id in combination with that hydration flag to understand page readiness...

PR #5574, opened 2024-10-02
feat: improve ui-site-map MP-880

emuvente:
This separates the redirect routes into their own list, showing both the original path and the destination.
Screenshot 2024-10-02 at 2 53 16 PM

This also extracts and replaces path parameters with values from inputs, allowing clicking on the link with whatever test value you want. However, this doesn't yet work with parameters that have pattern limitations (like `/lend/:id(\d+)`).
Screenshot 2024-10-02 at 2 53 28 PM


mcstover (approved):
Nice update!

PR #5580, opened 2024-10-03
fix: dynamic rich text rendering

eddieferrer:
Fixes rendering of dynamic components in the built production mode on /monthlygood

### DIFF
--- a/src/components/Contentful/DynamicRichText.vue
+++ b/src/components/Contentful/DynamicRichText.vue
@@ -27,8 +27,6 @@ export default {
 					KvContentfulImg: defineAsyncComponent(() => import('@kiva/kv-components/vue/KvContentfulImg')),
 					ButtonWrapper: defineAsyncComponent(() => import('#src/components/Contentful/ButtonWrapper')),
 				},
-				props: {
-				},
 			};
 		},
 		elementClasses() {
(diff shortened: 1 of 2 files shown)


vite.config.js
```diff
@@ -27,95 +27,104 @@ const imagesRegex = /images\/.+\.(png|jpe?g|gif|webp|avif|svg|ico)/;
...
+		resolve: {
+			alias: {
+				// alias src directory
+				'#src': resolve('src'),
+				// alias promise module to handle timesync calling require('promise')
+				promise: resolve('build/promise.js'),
+				// this alias is required for the rendering of src/components/Contentful/DynamicRichText.vue
+				// it should only be used on the client build in production
+				// eslint-disable-next-line max-len
+				...(isSsrBuild === false && mode === 'production' && { vue: path.resolve(__dirname, 'node_modules', 'vue', 'dist', 'vue.esm-bundler.js') })
```
eddieferrer, line 113: This is the main change in this PR. 
We have to use 'vue' as the alias target, and the 'vue.esm-bundler.js' build. And we can only have this alias for the client side production build.

src/components/Contentful/DynamicRichText.vue
```diff
@@ -52,16 +50,7 @@ export default {
...
-			'div', /* Wrap output in a div since Vue templates must have 1 root element */
-			[
-				h(this.dynamicComponent, {
-					class: this.elementClasses,
-					props: {
-					},
-				})
-			]
-		);
+		return h(this.dynamicComponent, { class: this.elementClasses });
```
mcstover, line 53: Do we need to blend in props here or is that automatic?

mcstover (approved):

src/components/Contentful/DynamicRichText.vue
eddieferrer: Should just work

PR #5582, opened 2024-10-04
fix: hydration and missing content on mg page

eddieferrer:
It seems we need to wrap 'v-html' and components which use 'v-html' with 'isClient' to avoid hydration errors and missing content in the built app

### DIFF
--- a/src/components/MonthlyGood/AutomaticallySupportNotice.vue
+++ b/src/components/MonthlyGood/AutomaticallySupportNotice.vue
@@ -14,7 +14,7 @@
 			<p v-html="valueBody"></p>
 		</div>
 		<div
-			class="md:tw-flex-1 tw-block
+			class="md:tw-flex-1
 			tw-order-first md:tw-order-last
 			md:tw-px-4
 			tw-mb-2 md:tw-mb-4
--- a/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
+++ b/src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
@@ -4,6 +4,7 @@
 		<kv-hero
 			class="tw-text-center"
 			style="margin-bottom: 0;"
+			v-if="isClientReady"
 		>
 			<template #images>
 				<kv-contentful-img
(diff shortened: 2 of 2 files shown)


src/components/MonthlyGood/AutomaticallySupportNotice.vue
```diff
@@ -14,7 +14,7 @@
 			<p v-html="valueBody"></p>
 		</div>
 		<div
-			class="md:tw-flex-1 tw-block
```
eddieferrer, line 17: This class is overwritten by tw-inline-flex directly below

src/pages/MonthlyGood/MonthlyGoodLandingPage.vue
```diff
@@ -238,24 +240,14 @@ export default {
...
-			return element.value === this.category;
-		});
-		this.selectedChannelLoanIds = resultsArray?.[0]?.expLoansIds ?? [];
-	},
 	watch: {
 		selectedGroup() {
 			const resultsArray = this.lendingCategories.filter(element => {
 				return element.value === this.selectedGroup;
 			});
-			this.selectedChannelLoanIds = resultsArray?.[0]?.expLoansIds ?? [];
```
eddieferrer, line 258: this selectedChannel stuff is for an experiment that is no longer active

emuvente (approved):
This is surprising, as it suggests that v-html does not work for server renders. Or is this only for content that is loaded client-side only?

eddieferrer:
> @emuvente We can currently see the server render showing the full contentful content so that is working. Also noting the full content is present when executing an in-app navigation. Some of the nested content is disappearing on a full page refresh/load. So weird that it works on both the client and server but wipes itself out with the hydration mismatch.
> 
> @eddieferrer Did you try to create a readQuery in created and re-populate `pageData` to ensure it was available in the client?

I did not try the readQuery specifically. 

The content is available on the client and on the server. @emuvente my takeaway here is the same as yours that v-html is not working for server renders in conjunction with our preFetch mechanism.

PR #5599, opened 2024-10-09
feat: thanks view tracking MP-921

emuvente:
The thanks page is complex enough that it's hard to know what version of the page a user actually saw without digging into the details of the URL parameters and what was in their basket. This refactors the logic for selecting which thanks view is visible into a computed value, and adds tracking for the visible view.

### DIFF
--- a/src/components/Thanks/ThanksLayoutV2.vue
+++ b/src/components/Thanks/ThanksLayoutV2.vue
@@ -403,9 +403,9 @@ export default {
 
 		this.determineIfMobile();
 		this.$kvTrackEvent(
-			'thanks',
-			'view',
-			`${this.visibleSection} v2 view`,
+			'post-checkout',
+			'show',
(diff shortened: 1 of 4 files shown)


mcstover (approved):

eddieferrer:
We don't currently have 'thanks' and 'view' as values for snowplow categories. And the label and properties should all be lower case kebab cased.

Since these are new events they should fit the web event guidelines. 

We do have the 'post-checkout' category. And 'show' action. So we could just use those.  


For example this event: 
`'thanks', 'view', 'May challenge view', 'guest' `

Could be: 
`'post-checkout', 'show', 'thanks-may-challenge', 'guest' `

mcstover (approved):

roger-in-kiva (approved):

mcstover (approved):

PR #5607, opened 2024-10-17
feat: my kiva badges section

roger-in-kiva:
- MyKiva badges section added
- reqs: https://kiva.atlassian.net/browse/MP-837

Desktop:
Screenshot 2024-10-17 at 2 43 25 p m

Tablet:
Screenshot 2024-10-17 at 2 43 40 p m

Mobile:
Screenshot 2024-10-17 at 2 44 02 p m


src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -133,6 +142,30 @@ const handleSelectedLoan = loan => {
...
+			logReadQueryError(e, 'MyKivaPage userAchievementProgressQuery');
+		});
+};
+
+const fetchBadgesData = () => {
+	apollo.query({
+		query: contentfulEntriesQuery,
+		variables: {
+			contentType: 'challenge',
+			limit: 100,
```
dyersituations, line 159: How many badges do we currently have? I wonder when this will not be enough...

src/components/MyKiva/BadgesSection.vue
```diff
@@ -0,0 +1,127 @@
...
+
+			if (!userAchievement) {
+				badgeFound = {
+					...badgeFound,
+					hasStarted: false,
+					level: 0,
+				};
+			} else {
+				const hasStarted = userAchievement.status !== 'NO_PROGRESS';
+				const level = userAchievement.totalProgressToAchievement;
```
dyersituations, line 99: Is this the correct level to show? I thought `totalProgressToAchievement` was the number of loans to that level of the badge. Here I wonder if instead we need the highest badge level from the achievement service that is completed.

src/pages/Portfolio/MyKiva/MyKivaPage.vue
roger-in-kiva: 5 leves per badge, having 5 badges there are 25.
But as this a query for the whole challege set, we have `"total": 45`

We can change this to 200 to extra ensure.

dyersituations: 200 should be fine for now.

src/components/MyKiva/BadgesSection.vue
roger-in-kiva: Nice call! I guess yeah, we need the level from the service for this... or to clarify where the level is coming from.

roger-in-kiva: We can use this as dummy and change it when having more data... I'll add a comment for that.

dyersituations: Sounds good, add a `TODO` comment and add a ticket to the epic, thanks!

dyersituations (approved):
Looks good. Approving without testing locally, since we'll be doing be doing lots of testing with real data soon.

PR #5610, opened 2024-10-18
feat: add metadata content in lender profile for badge share

christian14b:
Share metadata for badge share

### DIFF
--- a/test/unit/specs/composables/useBadgeContentfulData.spec.js
+++ b/test/unit/specs/composables/useBadgeContentfulData.spec.js
@@ -1,0 +1,69 @@
+import logReadQueryError from '#src/util/logReadQueryError';
+import useBadgeContentfulData from '#src/composables/useBadgeContentfulData';
+import { defaultBadges } from '#src/util/achievementUtils';
+
+jest.mock('#src/util/logReadQueryError');
+
+describe('useBadgeContentfulData', () => {
+	let apollo;
+	let route;
+
+	beforeEach(() => {
+		route = {
+			query: {
+				utm_campaign: `social_share_badge_${defaultBadges[0]}`
+			}
+		};
+	});
+
+	test('loads badge info correctly', async () => {
+		apollo = {
+			readQuery: jest.fn().mockReturnValue({
+				contentful: {
+					entries: {
+						items: [
+							{
+								fields: {
+									badgeImage: {
+										fields: {
+											file: {
+												url: 'http://example.com/badge.png'
+											}
+										}
+									},
+									challengeName: 'Test Challenge'
+								}
+							}
+						]
+					}
+				}
+			}),
+		};
+
+		const { loadBadgeInfo, isBadgeKeyValid } = useBadgeContentfulData(apollo, route);
+
+		const result = await loadBadgeInfo();
+
+		expect(result.badgeImage).toBe('http://example.com/badge.png');
+		expect(result.badgeCategory).toBe('Test Challenge');
+		expect(isBadgeKeyValid).toBe(true);
+		expect(apollo.readQuery).toBeCalledTimes(1);
+	});
+
+	test('handles error in loading badge info', async () => {
+		apollo = {
+			readQuery: jest.fn().mockImplementation(() => {
+				throw new Error('Query error');
+			}),
+		};
+
+		const { loadBadgeInfo } = useBadgeContentfulData(apollo, route);
+
+		const result = await loadBadgeInfo();
+
+		expect(result.badgeImage).toBeNull();
+		expect(result.badgeCategory).toBe('');
+		expect(logReadQueryError).toHaveBeenCalled();
+		expect(apollo.readQuery).toBeCalledTimes(1);
+	});
+});


src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -95,18 +103,25 @@ export default {
...
 			const publicId = currentRoute.params?.publicId ?? '';
 
-			return client.query({
-				query: lenderPublicProfileQuery,
-				variables: { publicId }
-			});
-		},
+			const utmCampaign = currentRoute?.query?.utm_campaign ?? '';
+			const loadBadgeInfo = utmCampaign.includes('badge_') && utmCampaign.includes('social_share');
+			const badgeKey = utmCampaign.split('badge_')[1];
```
dyersituations, line 118: Should we be checking if `badgeKey` is defined and acting different below if it's not defined?

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -119,23 +134,47 @@ export default {
...
-			return this.lenderInfo?.seoImage?.url ?? '';
+			return this.enableBadgeContent ? this.badgeImage : this.lenderInfo?.seoImage?.url ?? '';
 		},
 		completedAchievements() {
 			return this.allAchievements.filter(achievement => achievement.status === 'COMPLETE');
 		},
 	},
+	methods: {
+		loadBadgeInfo(badgeKey) {
+			const data = this.apollo.readQuery({
```
dyersituations, line 165: We'll want a try catch around this, with `logReadQueryError` in the `catch`

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -151,6 +190,12 @@ export default {
 		this.lenderInfo = cachedLenderInfo?.community?.lender ?? {};
 		this.allAchievements = cachedLenderInfo?.userAchievementProgress?.achievementProgress ?? [];
 		this.lenderIsPublic = !!this.lenderInfo?.id;
+
+		const utmCampaign = this.$route?.query?.utm_campaign ?? '';
+		if (utmCampaign.includes('badge_') && utmCampaign.includes('social_share')) {
+			const badgeKey = utmCampaign.split('badge_')[1];
+			this.loadBadgeInfo(badgeKey);
```
dyersituations, line 197: Do we need `await` this query so that the data is guaranteed to be there on server render? I see that we're pulling data from this query to add to the page title and whatnot.

dyersituations:
@christian14b @emuvente I see in the ticket that this is also needed on the home page, but with the home page being cached, I don't see a great way to update the page title and whatnot on server render. Maybe we can chat with product about using a different, non-cached page?

emuvente:
> I see in the ticket that this is also needed on the home page, but with the home page being cached, I don't see a great way to update the page title and whatnot on server render. Maybe we can chat with product about using a different, non-cached page?

@dyersituations Fastly treats URLs with different query parameters as separate cache objects unless we explicitly tell it not to https://docs.fastly.com/en/guides/making-query-strings-agnostic. So each different utm_campaign value will have a separate cached page.

dyersituations:
> > I see in the ticket that this is also needed on the home page, but with the home page being cached, I don't see a great way to update the page title and whatnot on server render. Maybe we can chat with product about using a different, non-cached page?
> 
> @dyersituations Fastly treats URLs with different query parameters as separate cache objects unless we explicitly tell it not to https://docs.fastly.com/en/guides/making-query-strings-agnostic. So each different utm_campaign value will have a separate cached page.

Ah yup, that's right, I forgot about the query params existing, thanks!

src/pages/LenderProfile/LenderProfile.vue
dyersituations: Looks like try catch still needed here.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -151,6 +192,14 @@ export default {
 		this.lenderInfo = cachedLenderInfo?.community?.lender ?? {};
 		this.allAchievements = cachedLenderInfo?.userAchievementProgress?.achievementProgress ?? [];
 		this.lenderIsPublic = !!this.lenderInfo?.id;
+
+		const utmCampaign = this.$route?.query?.utm_campaign ?? '';
+		const isUtmValid = utmCampaign.includes('badge_') && utmCampaign.includes('social_share');
+		const badgeKey = utmCampaign.split('badge_')[1];
+		const isBadgeKeyValid = defaultBadges.includes(badgeKey);
```
dyersituations, line 199: There's some duplicated logic here to check the badge, can it be extracted to a method and be used in both places? Similarly, will this logic be needed in the home page as well? If so, most of the new code here can be extracted to a new composable (yay Vue3!) and used here and in the home page.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -24,6 +24,7 @@ import lenderPublicProfileQuery from '#src/graphql/query/lenderPublicProfile.gra
 import LenderProfileWrapper from '#src/components/LenderProfile/LenderProfileWrapper';
 import NotFoundWrapper from '#src/components/NotFound/NotFoundWrapper';
 import KvPageContainer from '@kiva/kv-components/vue/KvPageContainer';
+import useBadgeContentfulData from '#src/composables/useBadgeContentfulData';
```
dyersituations, line 27: Looks like the composable file wasn't added.

src/composables/useBadgeContentfulData.js
```diff
@@ -0,0 +1,47 @@
...
+import { gql } from 'graphql-tag';
+import logReadQueryError from '#src/util/logReadQueryError';
+
+export const badgeQuery = gql`query contentfulBadgeImage ($badgeKey: String!) {
+	contentful {
+		entries(contentKey: $badgeKey, contentType: "challenge")
+	}
+}`;
+
+export default function useBadgeContentfulData(client, route) {
```
dyersituations, line 11: Ideally this composable gets unit testing, but it can be a follow-up task. Please log a ticket if it isn't done now. A great benefit of composables is that it is a great pattern for allowing testing in the FE, much like utils.

dyersituations (approved):

dyersituations (approved):

PR #5611, opened 2024-10-21
feat: connect badge modal to real data

dyersituations:
https://kiva.atlassian.net/browse/MP-838

- Added the badge modal back to the MyKiva page
- Connected with real Contentful + achievement service data
- Added unit testing for composables in previous PR
- Fixed issue with unit test package.json command
- Other minor cleanup

Next steps (same ticket):
- Different badge states in journey
- Caption + button in journey

image

image


src/components/MyKiva/BadgeModalContentJournal.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div>
 		<p>
-			{{ description }}
+			{{ badge.fields.shareFact }}
```
roger-in-kiva, line 4: Is this a required field?

dyersituations: Nope, but we need to discuss further what field to use. I just chose one that we could use for now.

roger-in-kiva (approved):

PR #5615, opened 2024-10-24
feat: user achievements query updated

roger-in-kiva:
- user achievements query updated


src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -157,18 +157,15 @@ export default {
...
 	created() {
 		this.apollo.query({
-			query: userAchievementsProgress,
-			variables: {
-				userId: this.userId.toString(),
-			},
+			query: userAchievementProgressQuery,
 		}).then(({ data }) => {
-			this.allAchievements = data?.userAchievementProgress?.achievementProgress ?? [];
+			this.allAchievements = data?.userAchievementProgress?.lendingAchievements ?? [];
```
dyersituations, line 162: The lender profile doesn't currently show tiered badges, correct? It looks like these changes utilize the old `lendingAchievements` query.

roger-in-kiva: Not sure what you mean. is this helpful?
https://kiva.atlassian.net/browse/MP-923?focusedCommentId=359210

src/graphql/query/userAchievementProgress.graphql
```diff
@@ -1,9 +1,19 @@
 query UserAchievementProgress {
   userAchievementProgress {
     id
+    lendingAchievements {
```
dyersituations, line 4: These two queries were combined, but I only see locations where we use either `lendingAchievements` or `tieredLendingAchievements`. Do we need this combined version or separate queries?

roger-in-kiva: This will be helpful for this my kiva ticket: https://kiva.atlassian.net/browse/MP-951

src/pages/Portfolio/LendingStats/LendingStatsPage.vue
dyersituations: Does `lendingAchievements` include the new tiered badges as well? Or are those not included yet? And the comment mentions checking all of the milestones, but it doesn't look like we're doing that yet. And I don't know if this code covers the climate badge situation.

roger-in-kiva: Oh no, they are separated as you see in your comment about the query below.

src/graphql/query/userAchievementProgress.graphql
dyersituations: Okay that makes sense. I hadn't seen that ticket before, but I see that it's a new ticket.

src/pages/Portfolio/LendingStats/LendingStatsPage.vue
dyersituations: Okay makes sense. What about checking all milestones of an old achievement, like John mentioned? It looks like we're just checking the first milestone. And do we need special casing for the climate badge?

roger-in-kiva: Based on John's comment:

> for the lendingMilestones (the non-tiered ones) they are complete if all the milestones are complete

The first one will be complete if all are complete, for the edge case as climate badge.

dyersituations: Ah gotcha. Do we have special casing already for the climate badge?

roger-in-kiva: Seems like not. Do you see smt different?

dyersituations: Not seeing anything, but if we haven't had it, maybe we don't need the special casing here.

dyersituations:
@roger-in-kiva changes here look good, but like we discussed, I don't think we can merge at the moment. Could you post in the team slack channel, asking the BE engineers the timeline on getting the achievement service to prod?

PR #5620, opened 2024-10-28
feat: composable for tiered badge data

dyersituations:

- Partial implementation to allow for other engineers to utilize this work
- Created composable for combining badge tiered achievements and contentful data
- Work to come: comments, unit tests, using on MyKiva correctly, etc.
- This doesn't break the page more than it already is currently 😄

### DIFF
--- a/src/composables/useBadgeData.js
+++ b/src/composables/useBadgeData.js
@@ -1,0 +1,126 @@
+import { onMounted, ref, computed } from 'vue';
+import userAchievementProgressQuery from '#src/graphql/query/userAchievementProgress.graphql';
+import contentfulEntriesQuery from '#src/graphql/query/contentfulEntries.graphql';
+import logReadQueryError from '#src/util/logReadQueryError';
+import { defaultBadges } from '#src/util/achievementUtils';
+
+/**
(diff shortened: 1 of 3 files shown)


src/composables/useBadgeData.js
```diff
@@ -0,0 +1,126 @@
...
+	 *   "level": undefined
+	 * }
+	 */
+	const badgeData = computed(() => {
+		const badges = [];
+
+		// Ensure data loaded from both achievement service and Contentful
+		if (badgeAchievementData.value && badgeContentfulData.value) {
+			// Currently only targeting specific tiered badges
+			defaultBadges.forEach(badgeKey => {
```
roger-in-kiva, line 84: I guess we need a more generic data return and not for default ones only, as in my achievements section we need to verify all of them and show the earned tiers as well.

dyersituations: Ah, I didn't realize we were showing everything in that section. Okay, I'll change this real quick.

dyersituations: Done.

roger-in-kiva (approved):
Looks good, thanks!

PR #5622, opened 2024-10-28
feat: continue expanding use of tiered badge data

dyersituations:

- Next step is getting the badges section of MyKiva to show the correct badge title, level, and icon based on the current in-progress tier
- Added unit testing for new composable (mostly adding lots of mock data to make a realistic test)
- Fixed small spacing issue on MyKiva badge section badges
- Moved tier date cleanup to composable
- Moved Vue lifecycle methods out of composable so it could be reused as a util
- Other minor cleanup, like commenting

image

### DIFF
--- a/src/composables/useBadgeData.js
+++ b/src/composables/useBadgeData.js
@@ -1,4 +1,4 @@
-import { onMounted, ref, computed } from 'vue';
+import { ref, computed } from 'vue';
 import userAchievementProgressQuery from '#src/graphql/query/userAchievementProgress.graphql';
 import contentfulEntriesQuery from '#src/graphql/query/contentfulEntries.graphql';
 import logReadQueryError from '#src/util/logReadQueryError';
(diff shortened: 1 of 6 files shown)


src/composables/useBadgeData.js
```diff
@@ -1,26 +1,35 @@
...
 	});
 
-	const fetchAchievementData = () => {
+	/**
+	 * Calls Apollo to get the badge achievement service data
+	 */
+	const fetchAchievementData = apollo => {
 		apollo.query({ query: userAchievementProgressQuery })
 			.then(result => {
 				badgeAchievementData.value = result.data?.userAchievementProgress?.tieredLendingAchievements ?? [];
```
roger-in-kiva, line 35: Can we add `lendingAchievements` here?

dyersituations: Wouldn't we need to use both to get all achievements (old and new)?

roger-in-kiva: We need both, for the journey and earned badges, these last ones includes both.
Probably we can also just call the new field `achievements` on the query, to get both in a call.

dyersituations: What do you mean by the new field `achievements`? I'm not seeing that in the gql sandbox.

roger-in-kiva: For real? That one includes both cases.

Screenshot 2024-10-28 at 5 08 25 p m

dyersituations: Ah, on the parent, got it, thanks.

I'll do this in a follow-up PR.

roger-in-kiva: Thanks!

roger-in-kiva (approved):

PR #5626, opened 2024-10-29
feat: connect journey tiered data

dyersituations:

- The badge section now shows the correct badge icon: in-progress tier or next tier if no progress
- Badge journey now shows the correct badge icons per tier
- Badge journey now shows the level names from Contentful
- Updated stories with correct data format
- Fixed issue with badge modal opening to incorrect state
- Fixed size of locked badge icon
- Resolved Vue warning related to badge title prop being `null`
- Removed `isBadgeImageLoaded` since images are now the same square size and smaller + faster loading
- Generally cleaned up how badge data is combined and passed from the MyKiva page
- Fixed issues with the small progress circle in the badge journey
- Entire badge is now clickable in MyKiva badge section

image

image


src/components/MyKiva/BadgeInProgress.vue
```diff
@@ -200,11 +176,11 @@ const onInteractCarousel = interaction => {
 	$kvTrackEvent('carousel', 'click-carousel-horizontal-scroll', interaction);
 };
 
-const badgeName = computed(() => props.badge?.fields?.challengeName ?? '');
+const badgeName = computed(() => props.badge?.challengeName ?? '');
 // TODO: Replace once the work of using the icon of the correct in-progress badge level is complete
 const badgeImage = computed(() => props.badge?.fields?.badgeImage?.fields?.file?.url ?? '');
 const badgeLevel = computed(() => props.badge?.fields?.level ?? '');
-const totalProgress = computed(() => props.badge?.totalProgressToAchievement ?? 0);
+const totalProgress = computed(() => props.badge?.achievementData?.totalProgressToAchievement ?? 0);
```
roger-in-kiva, line 183: Do we need reactive props here?

dyersituations: Probably, since the modal can be re-opened with another badge.

roger-in-kiva: Same for other cases...

dyersituations: Right, I'm assuming we generally want reactivity here for when the badge modal is re-opened with another badge.

roger-in-kiva: It's some upcoming work for that?

dyersituations: What do you mean, these fields are already reactive, `computed` based on `props`?

roger-in-kiva: True! I mean if passing toRefs is needed...

dyersituations: It shouldn't be needed, since we're not deconstructing the props.

PR #5628, opened 2024-10-29
feat: minor changes to badges so we can start functional testing

dyersituations:


src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -88,7 +88,7 @@
 					>
 						My impact journeys
 					</h3>
-					<BadgesSection :badge-data="badgeData" @badge-clicked="handleBadgeClicked" />
+					<BadgesSection :badge-data="badgeData" @badge-clicked="handleBadgeSectionClicked" />
 
 					<BadgeModal
```
roger-in-kiva, line 93: I guess we can move BadgeModal outside journeys section to use it from others like My Achievements and header badge tile; so, at some point we can hide some of them w/o affecting the BadgeModal rendering.

dyersituations: `BadgeModal` currently works for the 3 use cases (with some work left). What are you referring to specifically?

dyersituations: Those other use cases should work with some small changes to code.

roger-in-kiva: I'm just referring that is wrapped by the Journeys section in MyKivaPage component right now.
So, for a more generic structure could be move outside that section, probably below the www-page element.

I can also do that in my work, but should be nice if agree having on this PR.

dyersituations: Ah, it doesn't really matter where that component lives in the page, as it is a modal. Go ahead and move it with your changes, if that makes more sense somewhere else.

roger-in-kiva (approved):

mcstover (approved):

PR #5630, opened 2024-10-30
fix: various MyKiva fixes

emuvente:


src/components/BorrowerProfile/HeroBackground.vue
```diff
@@ -29,6 +29,12 @@ export default {
 	components: {
 		KvContentfulImg,
 	},
+	props: {
+		loanId: {
```
dyersituations, line 33: Are we using this prop anywhere currently?

dyersituations (approved):

PR #5632, opened 2024-10-30
feat: save flag if user was part of my kiva experiment in preferences

christian14b:
Store in user preferences if user was part of my kiva experiment

### DIFF
--- a/src/composables/useUserPreferences.js
+++ b/src/composables/useUserPreferences.js
@@ -1,4 +1,5 @@
 import { gql } from 'graphql-tag';
+import logReadQueryError from '#src/util/logReadQueryError';
 
 export default function useUserPreferences(apollo) {
 	const createUserPreferences = () => {
@@ -20,8 +21,8 @@ export default function useUserPreferences(apollo) {
 					preferences: '',
 				},
 			});
-		} catch (error) {
-			throw new Error(error);
+		} catch (e) {
+			logReadQueryError(e, 'useUserPreferences createUserPreferences');
 		}
 	};
 
@@ -51,8 +52,8 @@ export default function useUserPreferences(apollo) {
 					preferences,
 				},
 			});
-		} catch (error) {
-			throw new Error(error);
+		} catch (e) {
+			logReadQueryError(e, 'useUserPreferences updateUserPreferences');
 		}
 	};
 
@@ -67,8 +68,8 @@ export default function useUserPreferences(apollo) {
 			const response = await updateUserPreferences({ userPreferences: currentPreferences, newPreference });
 
 			return response;
-		} catch (error) {
-			throw new Error(error);
+		} catch (e) {
+			logReadQueryError(e, 'useUserPreferences saveUserPreferences');
 		}
 	};
 
--- a/src/pages/Portfolio/MyKiva/MyKivaPage.vue
+++ b/src/pages/Portfolio/MyKiva/MyKivaPage.vue
@@ -217,14 +217,14 @@ const fetchMyKivaData = () => {
 		});
 };
 
-const fetchUserPreferences = async () => {
+const saveMyKivaToUserPreferences = () => {
 	const preferences = userInfo.value?.userPreferences?.preferences;
 	const formattedPreference = typeof preferences === 'string'
 		? JSON.parse(userInfo.value?.userPreferences?.preferences)
 		: preferences;
 
 	if (!formattedPreference?.myKivaPageExp) {
-		await saveUserPreferences({
+		saveUserPreferences({
 			userPreferences: userInfo.value?.userPreferences ?? null,
 			newPreference: {
 				myKivaPageExp: 1,
@@ -247,6 +247,6 @@ onMounted(async () => {
 	await fetchMyKivaData();
 	fetchAchievementData(apollo);
 	fetchContentfulData(apollo);
-	fetchUserPreferences();
+	saveMyKivaToUserPreferences();
 });
 </script>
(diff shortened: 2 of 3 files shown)


src/composables/useUserPreferences.js
```diff
@@ -0,0 +1,70 @@
+import { gql } from 'graphql-tag';
+
+export default function useUserPreferences(apollo) {
+	const	createUserPreferences = () => {
```
dyersituations, line 4: Remove extra spaces after `const`

src/composables/useUserPreferences.js
```diff
@@ -0,0 +1,70 @@
...
+				my {
+					createUserPreferences(userPreferences: {preferences: $preferences}) {
+						id
+						preferences
+					}
+				}
+			}
+		`;
+
+		return apollo.mutate({
```
dyersituations, line 16: Exceptions from this `mutate` need to be handled.

src/composables/useUserPreferences.js
```diff
@@ -0,0 +1,70 @@
...
+	};
+
+	const saveUserPreferences = async ({ userPreferences, newPreference }) => {
+		let currentPreferences = userPreferences;
+		if (!userPreferences?.id) {
+			const createPreferences = await createUserPreferences();
+			currentPreferences = createPreferences?.data?.my?.createUserPreferences ?? {};
+		}
+
+		return updateUserPreferences({ userPreferences: currentPreferences, newPreference });
```
dyersituations, line 62: Do we need to ensure that this updated is awaited or is it OK if it happens in the background?

src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -207,6 +210,20 @@ apollo.query({ query: myKivaQuery })
 			// eslint-disable-next-line prefer-destructuring
 			activeLoan.value = loans.value[0];
 			fetchLoanUpdates(activeLoan.value.id);
+
+			const preferences = userInfo.value?.userPreferences?.preferences;
```
dyersituations, line 214: We can probably move this new code to `onMounted`.

src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -207,6 +210,20 @@ apollo.query({ query: myKivaQuery })
...
+			const formattedPreference = typeof preferences === 'string'
+				? JSON.parse(userInfo.value?.userPreferences?.preferences)
+				: preferences;
+
+			if (!formattedPreference?.myKivaPageExp) {
+				saveUserPreferences({
+					userPreferences: userInfo.value?.userPreferences ?? null,
+					newPreference: {
+						myKivaPageExp: 1,
```
dyersituations, line 223: We also need to ensure that we update the `ImpactDashboard` component with getting the user preferences, so the user gets navigated to MyKiva like expected. I think that work would be part of this changeset, unless you plan on a quick follow-up.

src/pages/Portfolio/MyKiva/MyKivaPage.vue
dyersituations: Make sure they calls are awaited.

src/pages/Portfolio/MyKiva/MyKivaPage.vue
dyersituations: Yes, since right now we only take the user to MyKiva is they have less that 4 loans. We need to have that check OR if they have this user preference saved (they've seen MyKiva before and should continue seeing it).

test/unit/specs/composables/useUserPreferences.spec.js
```diff
@@ -0,0 +1,67 @@
...
+			);
+
+			expect(newUserPreference).toEqual({
+				data: updatedPreferenceDataMock
+			});
+		});
+	});
+
+	describe('saveUserPreferences', () => {
+		it('should create user preferences expected', async () => {
```
dyersituations, line 45: `should save`

src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -198,21 +201,39 @@ const handleSelectedLoan = loan => {
...
+};
+
+const fetchUserPreferences = async () => {
+	const preferences = userInfo.value?.userPreferences?.preferences;
+	const formattedPreference = typeof preferences === 'string'
+		? JSON.parse(userInfo.value?.userPreferences?.preferences)
+		: preferences;
+
+	if (!formattedPreference?.myKivaPageExp) {
+		await saveUserPreferences({
```
dyersituations, line 227: This `await` isn't necessary.

src/pages/Portfolio/MyKiva/MyKivaPage.vue
```diff
@@ -198,21 +201,39 @@ const handleSelectedLoan = loan => {
...
-onMounted(() => {
+				activeLoan.value = loans.value[0];
+				fetchLoanUpdates(activeLoan.value.id);
+			}
+		}).catch(e => {
+			logReadQueryError(e, 'MyKivaPage myKivaQuery');
+		});
+};
+
+const fetchUserPreferences = async () => {
```
dyersituations, line 220: This method would make more sense named something like `saveMyKivaToUserPreferences`

src/composables/useUserPreferences.js
```diff
@@ -0,0 +1,80 @@
...
+			`;
+
+			return apollo.mutate({
+				mutation: createUserPreferencesMutation,
+				variables: {
+					preferences: '',
+				},
+			});
+		} catch (error) {
+			throw new Error(error);
```
dyersituations, line 24: We don't want to re-throw the exceptions in these 3 methods. We want to instead use `logReadQueryError`

dyersituations (approved):

PR #5635, opened 2024-10-31
feat: stellate api apollo link switch

mcstover:


src/api/HttpLink.js
```diff
@@ -19,10 +27,34 @@ export default ({
...
+		uri: stellateGraphqlUri ?? uri,
+	};
+
+	const link = split(
+		operation => {
+			// check to see if we have a Stellate uri available
+			if (!stellateGraphqlUri) {
+				return false;
+			}
+			console.log('- - - - - - - - - -');
```
emuvente, line 41: Could we use logFormatter for these logs?

mcstover: I don't intend to leave these in. It's just for running locally and verifying the behavior. I think these in particular would just be noisy but if there is something you think we should log intentionally I can add those.

src/api/HttpLink.js
```diff
@@ -1,12 +1,20 @@
 import { BatchHttpLink } from '@apollo/client/link/batch-http/index';
 import { HttpLink } from '@apollo/client/link/http/index';
+import { split } from '@apollo/client/core/index';
+
+const cachableQueryOperationNames = [
+	'configSetting',
```
hobbsh, line 6: What do you think about setting these via environment variables? Or does temporary nature of this addition negate the need?

mcstover: Operation names correlate to the named queries/mutaions in our .graphql files so I think these should remain here as they will relate to queries that are safe to cache.

I think it will take a pretty extensive review and configuration effort before we could run all queries through their graph. There are a lot of things that can't or shouldn't be cached and controlling that at the type layer isn't granular enough. As I reviewed more the general tab I see that we wouldn't want to cache much of that for very long and definitiely not wholesale as the config is currently set. We'll need to go through all of the queries and their fields and detail what the appropriate rules for them.

hobbsh: That makes sense. I just meant setting this list as a comma separated env var (like `CACHEABLE_OPERATION_NAMES: operationOne,operationTwo` and `join` it in the code) to make additions/subtractions easy to test. Whatever works best though.

hobbsh (approved):
looks good based on my limited JS knowledge, thanks for getting this done!

mcstover:
This should be ready for some basic testing in staging. I can add the configs to activate it there after merging and doing some basic verificaiton without stellate on it dev.

config/dynamic.js
```diff
@@ -87,6 +87,8 @@ export const app = {
 	sentryURI: process.env.SENTRY_URI || 'https://7ce141b23c4a4e6091c206d08442f0e9@o7540.ingest.sentry.io/1201287',
 	sentryTraceSampleRate: process.env.SENTRY_TRACE_RATE || 0.25,
 	snowplowUri: process.env.SNOWPLOW_URI || 'events.fivetran.com/snowplow/v5qt54ocr2nm',
+	stellateGraphqlUri: process.env.STELLATE_GRAPHQL_URI || null,
```
dyersituations, line 90: Which config is `dynamic`?

dyersituations (approved):

config/dynamic.js
mcstover: This is the file we now use for all real environments.

emuvente (approved):

PR #5636, opened 2024-10-31
feat: using totalProgressToAchievement to evaluate if journey started

christian14b:
`hasStarted` didn't evaluate `totalProgressToAchievement` so journeys without completed tiers were showing the "Start journey" copy 

image
image

### DIFF
--- a/src/composables/useBadgeData.js
+++ b/src/composables/useBadgeData.js
@@ -135,7 +135,7 @@ export default function useBadgeData() {
 					// Get specific properties used in the UI
 					const completedTiers = sortedTiers.filter(t => !!t.completedDate);
 					const hasStarted = completedTiers.length > 0 || achievementData?.totalProgressToAchievement > 0;
-					const level = hasStarted ? completedTiers?.[completedTiers.length - 1]?.level ?? 1 : undefined;
+					const level = hasStarted ? completedTiers?.[completedTiers.length - 1]?.level : undefined;
 
 					// Clean up milestone progress date format
 					const { milestoneProgress } = achievementData;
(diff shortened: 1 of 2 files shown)


src/composables/useBadgeData.js
```diff
@@ -134,8 +134,8 @@ export default function useBadgeData() {
 
 					// Get specific properties used in the UI
 					const completedTiers = sortedTiers.filter(t => !!t.completedDate);
-					const hasStarted = completedTiers.length > 0;
-					const level = hasStarted ? completedTiers[completedTiers.length - 1].level : undefined;
+					const hasStarted = completedTiers.length > 0 || achievementData?.totalProgressToAchievement > 0;
+					const level = hasStarted ? completedTiers?.[completedTiers.length - 1]?.level ?? 1 : undefined;
```
dyersituations, line 138: The `level` should be `undefined` if a level has not been completed. So I think this line can be reverted.

src/composables/useBadgeData.js
```diff
@@ -134,8 +134,8 @@ export default function useBadgeData() {
 
 					// Get specific properties used in the UI
 					const completedTiers = sortedTiers.filter(t => !!t.completedDate);
-					const hasStarted = completedTiers.length > 0;
-					const level = hasStarted ? completedTiers[completedTiers.length - 1].level : undefined;
+					const hasStarted = completedTiers.length > 0 || achievementData?.totalProgressToAchievement > 0;
```
dyersituations, line 137: Please add a unit test that covers this case.

dyersituations: Did that test fail before your change to the mock data?

dyersituations: In that case, I think we're OK then.

dyersituations (approved):

PR #5640, opened 2024-10-31
fix: guest account flow store name in userAccount instead of lender

christian14b:
Within `GuestAccountCreation` component, mutation  `startGuestAccountClaim` store `firstName` and `lastName` in `userAccount` instead of `lender`. That's why `lenderName` is `empty` causing first letter not being shown

### DIFF
--- a/src/components/MyKiva/MyKivaProfile.vue
+++ b/src/components/MyKiva/MyKivaProfile.vue
@@ -40,16 +40,20 @@ const props = defineProps({
 		type: Object,
 		default: () => ({}),
 	},
+	userInfo: {
+		type: Object,
+		default: () => ({}),
+	},
 	isLoading: {
 		type: Boolean,
 		default: false,
 	},
 });
 
-const { lender } = toRefs(props);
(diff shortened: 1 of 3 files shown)


emuvente (approved):

src/components/MyKiva/MyKivaProfile.vue
```diff
@@ -40,16 +40,20 @@ const props = defineProps({
...
+		default: () => ({}),
+	},
 	isLoading: {
 		type: Boolean,
 		default: false,
 	},
 });
 
-const { lender } = toRefs(props);
+const { lender, userInfo } = toRefs(props);
```
dyersituations, line 53: For a guest in this situation, I'm assuming the `lender` image doesn't exist? Does the avatar show the first/last name initials like expected?

dyersituations: If the avatar exists with initials, all is well (so I'll approve the PR for now), but if the initials are missing, please investigate.

dyersituations (approved):

PR #5642, opened 2024-11-01
feat: change contentful level to a string and update code accordingly

dyersituations:
https://kiva.atlassian.net/browse/MP-980

- Created new field on challenges in Contentful, `levelName` short string, and removed unused `level` field -> simplifies differentiating challenge and level names
- Updated field for all 7 levels of the current 5 badges on dev in Contentful
- Updated `useBadgeData` to utilize this new field (allows us to have `levelNames` as strings)
- Updated UI to use the new field as required

### DIFF
--- a/src/components/MyKiva/BadgeModalContentJourney.vue
+++ b/src/components/MyKiva/BadgeModalContentJourney.vue
@@ -145,7 +145,10 @@ const getBadgeStatus = index => {
 
 const getTierName = index => {
 	const contentfulData = badgeWithVisibleTiers.value.contentfulData[index];
-	return `${contentfulData.challengeName} ${contentfulData.levelName}`;
+	if (contentfulData.challengeName && contentfulData.levelName) {
+		return `${contentfulData.challengeName} ${contentfulData.levelName}`;
+	}
+	return `Level ${index + 1}`;
 };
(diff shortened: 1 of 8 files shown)


src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -143,6 +143,11 @@ const getBadgeStatus = index => {
 	return BADGE_LOCKED;
 };
 
+const getTierName = index => {
+	const contentfulData = badgeWithVisibleTiers.value.contentfulData[index];
+	return `${contentfulData.challengeName} ${contentfulData.levelName}`;
```
mcstover, line 148: I think we should guard these to return an empty string if either are undefined...

dyersituations: Sure, can do, though if data is undefined, it means we messed up adding contentful data.

dyersituations: Both should be defined to have a understandable level name, so I added a fallback to a simple level name.

mcstover (approved):
Ok cool. I didn't know if those fields were required or not in Contentful

PR #5646, opened 2024-11-04
feat: badge modal header buttons mp 979

christian14b:
Also, fixing some padding in modal header 

UPDATED:
image

image

### DIFF
--- a/src/components/MyKiva/BadgeModal.vue
+++ b/src/components/MyKiva/BadgeModal.vue
@@ -1,11 +1,11 @@
 <template>
 	<KvLightbox
-		:class="{'badge-modal': !isJourneyActive}"
+		:class="{'badge-modal': !isJourneyActive && !isEarnedSection}"
 		:visible="show"
 		:title="title"
 		@lightbox-closed="closeLightbox"
 	>
-		<template v-if="!isJourneyActive" #header>
+		<template v-if="!isJourneyActive && !isEarnedSection" #header>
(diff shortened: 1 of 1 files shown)


dyersituations:
@christian14b what is the little gray line?
image

src/components/MyKiva/BadgeModal.vue
```diff
@@ -1,5 +1,21 @@
 <template>
-	<KvLightbox :visible="show" :title="title" @lightbox-closed="closeLightbox">
+	<KvLightbox
+		:class="{'badge-modal': !isJourneyActive}"
+		:visible="show"
+		:title="title"
+		@lightbox-closed="closeLightbox"
+	>
+		<template v-if="!isJourneyActive" #header>
```
dyersituations, line 8: I think we also need to check `isEarnedSection` somewhere in this new template code. The reason being is that the modal should close and not return to the journey if a completed badge is clicked in the bottom section of MyKiva, where the completed badges are shown.

src/components/MyKiva/BadgeModal.vue
```diff
@@ -1,5 +1,21 @@
 <template>
-	<KvLightbox :visible="show" :title="title" @lightbox-closed="closeLightbox">
+	<KvLightbox
+		:class="{'badge-modal': !isJourneyActive && !isEarnedSection}"
```
dyersituations, line 3: Does the earned badge state of the modal look correct in regards to the header buttons without these new styles?

dyersituations (approved):

PR #5652, opened 2024-11-05
feat: using useBadgeData composable to get share content

christian14b:
Removing old composable logic to get badge data from the composable in my kiva page

### DIFF
--- a/src/pages/LenderProfile/LenderProfile.vue
+++ b/src/pages/LenderProfile/LenderProfile.vue
@@ -23,8 +23,6 @@ import LenderProfileWrapper from '#src/components/LenderProfile/LenderProfileWra
 import NotFoundWrapper from '#src/components/NotFound/NotFoundWrapper';
 import KvPageContainer from '@kiva/kv-components/vue/KvPageContainer';
 import useBadgeData from '#src/composables/useBadgeData';
-import userAchievementProgressQuery from '#src/graphql/query/userAchievementProgress.graphql';
-import contentfulEntriesQuery from '#src/graphql/query/contentfulEntries.graphql';
 import lenderProfileBadgeDataQuery from '#src/graphql/query/lenderProfileBadgeData.graphql';
 
 export default {


src/components/MyKiva/BadgeCompleted.vue
```diff
@@ -118,7 +118,12 @@ const badgeImage = computed(() => {
 });
 
 const badgeLevel = computed(() => {
-	return badgeData.value?.achievementData?.target ?? 0;
+	return badgeData.value?.achievementData?.level ?? 0;
+});
+
+const shareUrl = computed(() => {
+	const base = isPublic.value ? `/lender/${lender.value?.publicId}` : '';
```
dyersituations, line 125: If we're taking the user to the home page do we need to add the query string on the next line?

src/composables/useBadgeData.js
```diff
@@ -308,6 +309,20 @@ export default function useBadgeData() {
 		return visibleData;
 	};
 
+	/**
+	 * Get the badge key and check if it has a valid format
+	 *
+	 * @param {*} badgeKey The badge key to validate
```
dyersituations, line 315: `{*}` can be removed here

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -33,6 +43,55 @@ export default {
 		LenderProfileWrapper,
 		NotFoundWrapper,
 	},
+	setup() {
```
dyersituations, line 46: I see this is `setup` within an older Vue component, does `onMounted` and everything work OK?

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -33,6 +43,55 @@ export default {
...
+		const apollo = inject('apollo');
```
dyersituations, line 47: For example, this component already has `apollo` injected.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -33,6 +43,55 @@ export default {
...
+
+		const {
+			isBadgeKeyValid, fetchAchievementData, fetchContentfulData, badgeData, getTierBadgeDataByLevel
+		} = useBadgeData();
```
dyersituations, line 51: Is this `setup` format needed because of this `composable` usage?

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -33,6 +43,55 @@ export default {
...
+
+		const route = useRoute();
+
+		const utmCampaign = route.query?.utm_campaign ?? '';
+		const badgeLevel = route.query?.badge_level ?? 0;
+		const badgeKey = utmCampaign.split('badge_')[1];
+		const enableBadgeContent = ref(false);
+
+		onMounted(async () => {
+			if (isBadgeKeyValid(utmCampaign)) {
```
dyersituations, line 61: This data used to be queried server-side, but it's now clientside in mounted. Is that an issue?

dyersituations: Ah yes, we do need the page title populated on server render.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -93,21 +95,26 @@ export default {
...
 			const currentRoute = route.value ?? route ?? {};
 			const publicId = currentRoute.params?.publicId ?? '';
-			const { isBadgeKeyValid, badgeQuery, badgeKey } = useBadgeContentfulData(client, currentRoute);
+			const badgeKey = currentRoute.query?.utm_campaign;
+			const { isBadgeKeyValid } = useBadgeData();
 
 			return Promise.all([
 				client.query({ query: lenderPublicProfileQuery, variables: { publicId } }),
-				isBadgeKeyValid ? client.query({ query: badgeQuery, variables: { badgeKey } }) : null,
+				isBadgeKeyValid(badgeKey) ? client.query({ query: userAchievementProgressQuery }) : null,
```
dyersituations, line 110: Can this two prefetched queries be combined?

src/components/MyKiva/BadgeCompleted.vue
dyersituations: Looks like this question is still open.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -93,21 +95,27 @@ export default {
...
 			const currentRoute = route.value ?? route ?? {};
 			const publicId = currentRoute.params?.publicId ?? '';
-			const { isBadgeKeyValid, badgeQuery, badgeKey } = useBadgeContentfulData(client, currentRoute);
+			const badgeKey = currentRoute.query?.utm_campaign;
+			const { isBadgeKeyValid } = useBadgeData();
 
 			return Promise.all([
 				client.query({ query: lenderPublicProfileQuery, variables: { publicId } }),
-				isBadgeKeyValid ? client.query({ query: badgeQuery, variables: { badgeKey } }) : null,
+				isBadgeKeyValid(badgeKey) ? [
```
dyersituations, line 110: It looks like this still runs the queries separately. Also, does this work? I think you would still need to spread the array created here. Is it possible to combine these queries to reduce the # of queries called?

dyersituations (approved):

PR #5657, opened 2024-11-07
fix: status card too wide and padding on large screens

christian14b:
image

Also removing padding for large screen to fit the second card end with the container.

image

### DIFF
--- a/src/components/MyKiva/MyKivaContainer.vue
+++ b/src/components/MyKiva/MyKivaContainer.vue
@@ -23,7 +23,7 @@ import KvGrid from '@kiva/kv-components/vue/KvGrid';
 }
 
 .container > div {
-	@media screen and (width <= 67rem) {
+	@media screen and (width >= 67rem) {
 		padding-left: 0;
 		padding-right: 0;
 	}


src/components/MyKiva/MyKivaContainer.vue
```diff
@@ -21,4 +21,11 @@ import KvGrid from '@kiva/kv-components/vue/KvGrid';
 	max-width: 67rem;
 	@apply tw-mx-auto;
 }
+
+.container > div {
```
dyersituations, line 25: This is applying to the entire page, what was the issue and does this have any other affects?

dyersituations:
@christian14b Looks good other than linting errors in action.

src/components/MyKiva/MyKivaContainer.vue
```diff
@@ -21,4 +21,11 @@ import KvGrid from '@kiva/kv-components/vue/KvGrid';
...
+	@media screen and (width <= 67rem) {
```
dyersituations, line 26: Was this supposed to be greater than?

dyersituations (approved):

PR #5666, opened 2024-11-11
feat: opt-in module mp 1024

christian14b:
image

image

image

After clicking a button:
image

image

### DIFF
--- a/src/components/Thanks/MyKiva/OptInNotification.vue
+++ b/src/components/Thanks/MyKiva/OptInNotification.vue
@@ -1,6 +1,6 @@
 <template>
 	<div
-		class="tw-flex tw-mx-auto tw-items-center tw-justify-start tw-gap-0.5 md:tw-gap-2
+		class="tw-flex tw-mx-auto tw-items-center tw-justify-start tw-gap-1 md:tw-gap-2
 			tw-bg-white tw-rounded tw-px-1.5 md:tw-px-3 tw-py-1.5 container"
 		:style="{maxWidth: '620px'}"
 	>
@@ -13,12 +13,12 @@
 				:key="loan.id"
 				:lender-name="loan?.name"
 				:lender-image-url="loan?.image?.url"
-				class="avatar tw-rounded-full tw-shadow tw-border-white tw-border-2 tw-w-auto"
+				class="avatar tw-rounded-full tw-shadow tw-w-auto"
 				:class="{'smaller-borrower-avatar' : loansToDisplay.length > 2 && index !== 1 }"
 				:style="{
 					zIndex: index === 1 ? 2 : 1,
-					marginRight: loansToDisplay.length > 2 && index === 0 ? '-30px' : '0',
-					marginLeft: loansToDisplay.length > 1 && index === loansToDisplay.length - 1 ? '-30px' : '0',
+					marginRight: loansToDisplay.length > 2 && index === 0 ? '-18px' : '0',
+					marginLeft: loansToDisplay.length > 1 && index === loansToDisplay.length - 1 ? '-18px' : '0',
 				}"
 			/>
 		</div>
@@ -59,9 +59,13 @@ const notificationMsg = computed(() => {
 
 <style lang="postcss" scoped>
 
-.smaller-borrower-avatar :deep(img) {
-	height: 36px;
-	width: 36px;
+.avatar :deep(img) {
+	@apply tw-w-4.5 tw-h-4.5 md:tw-h-6 md:tw-w-6 tw-border-2 tw-border-white;
+}
+
+.smaller-borrower-avatar, .smaller-borrower-avatar :deep(img) {
+	height: 30px;
+	width: 30px;
 }
 
 .message {


mcstover:
@dyersituations @emuvente Can we offer more options here such as checkboxes for the different sub-types of emails that would be sent? Alternatively, can we include a link to the email settings and messaging allowing folks to see the different types of emails they are agreeing to and choose accordingly?

dyersituations:
> @dyersituations @emuvente Can we offer more options here such as checkboxes for the different sub-types of emails that would be sent? Alternatively, can we include a link to the email settings and messaging allowing folks to see the different types of emails they are agreeing to and choose accordingly?

That would be a question for product. I'll ask in our team channel.

.storybook/stories/ThanksBadges.stories.js
```diff
@@ -26,6 +26,8 @@ const story = (args = {}) => {
 	return template;
 };
 
+const loans = mockedReceiptData.items.values?.map((item) => item?.loan).filter(loan => !!loan?.id);
```
dyersituations, line 29: Is this the same as the already defined `mockLoans`?

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,253 @@
...
+.borrower-image {
+	width: 80px !important;
+	height: 80px !important;
+
+	@screen md {
+		width: 160px !important;
+		height: 160px !important;
+	}
+
+	box-shadow: '0px 4.42px 22.1px 0px #D1DCD6';
```
dyersituations, line 223: This `box-shadow` isn't working (there's also an error in the CSS panel in Chrome). Use our existing drop shadow like in my recent PR.

dyersituations:
The circle of the borrower image doesn't look entire smooth in Chrome. Any way to improve this? The drop-shadow fix might be enough.

image

dyersituations:
In Figma these two elements are wider for mobile

image

dyersituations:
In the mocks these two are overlapped. See my recent PR for a similar example.

image

dyersituations:
There needs to be story (and potentially code) to support the 3+ loan use case. See my recent PR for a similar example.
image

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,253 @@
...
+							:aspect-ratio="1"
+							:default-image="{ width: 200, faceZoom: 30 }"
+							:hash="hash(loan)"
+							:images="[
+								{ width: 200, faceZoom: 30 },
+							]"
+						/>
+					</div>
+				</div>
+				<p>{{ description }}</p>
```
dyersituations, line 30: It looks like this should be `h3`

src/components/Thanks/MyKiva/ThanksBadges.vue
```diff
@@ -167,6 +175,14 @@ const props = defineProps({
 		type: Object,
 		default: () => ({}),
 	},
+	optedIn: {
```
dyersituations, line 178: This prop already exists above.

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,261 @@
+<template>
+	<transition-group name="collapse" tag="div">
+		<template v-if="!newConsentAnswered">
+			<div class="module-container">
+				<h2>{{ title }}</h2>
+				<div class="tw-flex tw-items-center tw-justify-center">
+					<BorrowerImage
```
dyersituations, line 7: Why is this using `BorrowerImage` instead of `KvUserAvatar`? That might be where some of the rendering differences I'm seeing are, such as the little dark line around the rounded edge of the image. Also these 1-3 borrower avatars need a shadow. For the shadow, you could use `tw-shadow`.

Current:
image

Figma:
image

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,261 @@
...
+});
+
+const description = computed(
+	() => `Want to hear how you're impacting ${borrowerName.value}'s life and more ways to help people like them?`
+);
+
+const loansToDisplay = computed(() => {
+	const loans = [...props.loans];
+	if (props.loans.length === 3) {
+		const indexToRemove = loans.indexOf(loan => loan.id === props.selectedLoan.id);
```
dyersituations, line 129: Why is a loan removed here?

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,261 @@
...
+							)
+						}
+					}
+				`,
+				variables: {
+					lenderNews: value,
+				},
+			});
+		} catch (error) {
+			logReadQueryError(error, 'error');
```
dyersituations, line 184: logReadQueryError(error, 'OptInModule updateCommunicationSettings');

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+.ghost-button :deep(span) {
+	@apply tw-bg-transparent;
+}
+
+.collapse-enter-active,
+.collapse-leave-active {
+  transition: max-height 1s ease, opacity 1s ease;
```
dyersituations, line 253: When you click the "Yes..." button, you will see the padding immediately disappear from inside the white box, then animate. That makes for a rough animation, so the padding should disappear like it does.

dyersituations: This happens with the "No.." link as well

src/components/Thanks/MyKiva/OptInNotification.vue
```diff
@@ -0,0 +1,73 @@
...
+	<div
+		class="tw-flex tw-mx-auto tw-items-center tw-justify-start tw-gap-0.5 md:tw-gap-2
+			tw-bg-white tw-rounded tw-px-3 tw-py-1.5"
+		:style="{maxWidth: '620px'}"
+	>
+		<div
+			v-if="receiveNews"
+			class="avatars-container tw-flex tw-items-center tw-w-full"
+		>
+			<KvUserAvatar
```
dyersituations, line 11: Images get stretched when there are more than one loan:

image
image

src/components/Thanks/MyKiva/ThanksBadges.vue
```diff
@@ -167,6 +175,10 @@ const props = defineProps({
 		type: Object,
 		default: () => ({}),
 	},
+	selectedLoan: {
```
dyersituations, line 178: What is this selected loan?

dyersituations: Yeah I didn't see it called out in the requirements, so I didn't worry about having a loan being spotlighted

dyersituations:
Avatars are stretched in all use cases.

image

image

dyersituations:
Seeing this error on animation:
KvButton.vue:215 Uncaught TypeError: Cannot read properties of null (reading 'removeChild')
    at HTMLSpanElement.animationComplete (KvButton.vue:215:27)

dyersituations:
Avatar size doesn't match Figma for any of the use cases (1, 2, 3 avatars)

image

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,220 @@
...
+						class="borrower-image tw-rounded-full tw-shadow tw-border-white tw-border-2"
+						:class="{'centered-borrower-image' : index === 1 && loansToDisplay.length === 3}"
+						:style="{
+							marginRight: getMarginRight(index),
+							marginLeft: getMarginLeft(index),
+							zIndex: index === 1 ? 2 : 1,
+						}"
+					/>
+				</div>
+				<h3>{{ description }}</h3>
```
dyersituations, line 21: It looks like this also needs `tw-font-book`

dyersituations:
The 3-loan version is closer but it's still not quite like figma. The two loans in the background should be more visible. Please try to match the Figma.

Figma:
image

VS

Actual:
image

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,220 @@
...
+						@click="() => updateOptIn(true)"
+						v-kv-track-event="[
+							'thanks',
+							'click',
+							'accept-opt-in-request',
+						]"
+					>
+						Yes, keep me updated
+					</kv-button>
+					<kv-button
```
dyersituations, line 34: Remove left/right padding so that the text is one line in mobile

dyersituations:
Figma shows this as one line in mobile, can it be done?

image

src/components/Thanks/MyKiva/OptInModule.vue
dyersituations: Looks like this change is still needed (removes the bold styling).

src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -0,0 +1,247 @@
...
+		height: 200px;
+	}
+}
+
+.borrower-image, .borrower-image :deep(img) {
+	width: 124px;
+	height: 124px;
+
+	@screen md {
+		width: 160px;
```
dyersituations, line 204: Looking at Figma, the center borrower image should be `200px` even when there is a single loan, right now the single loan use case shows a smaller image at `160px`.

src/components/Thanks/MyKiva/OptInModule.vue
dyersituations: This is still an issue.

dyersituations:
There some padding that needs to be removed for the module container for mobile. The padding needs to be changes. This screenshot shows how it probably should be `1rem` vs `1.5rem`. Otherwise, there's a single word that gets wrapped at the end.
image

dyersituations:
The mobile version of this notification has word-wrap. It looks like there should be space for the text to be one line in this use case. Also the avatar is only `36px` in Figma, which is smaller than it currently is.

image

src/components/Thanks/MyKiva/OptInNotification.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+	<div
+		class="tw-flex tw-mx-auto tw-items-center tw-justify-start tw-gap-0.5 md:tw-gap-2
```
dyersituations, line 3: I think `tw-gap-0.5` here should be `tw-gap-1`

dyersituations:
It still looks like these avatars should be `36px`. I'm seeing them as `36px` throughout Figma.

image

PR #5671, opened 2024-11-11
fix: badge tile width fixed

roger-in-kiva:
- badge tile width fixed


src/components/MyKiva/BadgeTile.vue
```diff
@@ -6,13 +6,13 @@
 		/>
 		<div
 			v-else
-			class="tw-cursor-pointer"
+			class="tw-cursor-pointer md:tw-w-1/2"
```
dyersituations, line 9: The request from design was just the white box, so the click and cursor should be moved to the div under the h4.

dyersituations (approved):

PR #5672, opened 2024-11-12
feat: add the level name to the top tile

dyersituations:
https://kiva.atlassian.net/browse/MP-1028

- Added level/tier name to top badge tile
- Resolved some console warnings

image


roger-in-kiva:
Based on the ticket example, should we add level to text?
`badge e.g. Women’s equality level 2`

dyersituations:
> Based on the ticket example, should we add level to text? `badge e.g. Women’s equality level 2`

Good question, but we don't do that anywhere else. Probably fine to leave "level" off for now and see if they request it.

roger-in-kiva (approved):

PR #5681, opened 2024-11-13
fix: match tracking events to epic

christian14b:


src/components/MyKiva/BadgeInProgress.vue
```diff
@@ -27,6 +27,7 @@
 			slides-to-scroll="visible"
 			:slide-max-width="singleSlideWidth"
 			@interact-carousel="onInteractCarousel"
+			@change="onInteractCarousel"
```
dyersituations, line 30: With `@change` and `@interact-carousel` both calling the same method, do we get double analytics events?

dyersituations: What was the issue with `@interact-carousel`?

dyersituations (approved):

PR #5682, opened 2024-11-14
feat: add tracking events to typage components

christian14b:
tracking events

### DIFF
--- a/src/components/Thanks/MyKiva/OptInModule.vue
+++ b/src/components/Thanks/MyKiva/OptInModule.vue
@@ -77,7 +77,7 @@ const props = defineProps({
 		type: Boolean,
 		default: false,
 	},
-	numberOfBadgesEarned: {
+	numberOfBadges: {
 		type: Number,
 		default: 0,
 	}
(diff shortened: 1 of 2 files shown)


src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -82,10 +72,19 @@ const props = defineProps({
 	optedIn: {
 		type: Boolean,
 		default: false
+	},
+	isGuest: {
+		type: Boolean,
+		default: false,
+	},
+	numberOfBadgesEarned: {
```
dyersituations, line 80: It looks like we aren't passing this property to the component in `ThanksBadges` yet.

src/components/Thanks/MyKiva/ThanksBadges.vue
```diff
@@ -183,9 +199,9 @@ const props = defineProps({
 		type: Object,
 		default: () => ({}),
 	},
-	selectedLoan: {
-		type: Object,
-		default: () => ({})
+	numberOfBadgesEarned: {
```
dyersituations, line 202: Instead of a new prop, you can create a new `computed` that returns the length of `props.badgesAchieved`

dyersituations (approved):

PR #5691, opened 2024-11-18
fix: gear icon width fixed

roger-in-kiva:
- gear icon width fixed


mcstover (approved):
We've had to do similar in some other contexts as well.

PR #5694, opened 2024-11-19
feat: filter loans in carousel by status

christian14b:
Filter loans which status are different from:  `FUNDED`, `FUNDRAISING`, `PAYING_BACK`, `RAISED`

### DIFF
--- a/src/graphql/query/myKiva.graphql
+++ b/src/graphql/query/myKiva.graphql
@@ -2,7 +2,7 @@ query myKivaQuery {
   my {
     id
     isBorrower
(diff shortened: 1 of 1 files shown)


dyersituations:
> > With a 10 limit in the query for these loans, are we ensuring that w'ell have 9 after filtering?
> 
> We probably need to increase that limit

Agreed, how much data is pulled back on each loan? If it's minimal, and the query is just used on MyKiva, increasing to 50 might be OK.

emuvente (approved):

dyersituations (approved):

PR #5704, opened 2024-11-20
fix: change event handler to transition

christian14b:
`CSS` uses transition for the bubble, so the correct directive is `transitioned`, instead of `animationend`


src/components/LoanCards/KvClassicLoanCardContainer.vue
```diff
@@ -39,7 +39,7 @@
 				:lender-name="borrowerName"
 				:lender-image-url="borrowerImageUrl"
 				:class="{'animate': isAnimating}"
-				@animationend="resetBubble"
+				@transitionend="resetBubble"
```
dyersituations, line 42: What issue does this resolve?

dyersituations (approved):

PR #5707, opened 2024-11-22
feat: add tiered badges to BadgesList component

christian14b:
Adding tiered badges to Lending Stats (old portfolio) and Lender Badges (Lender profile) using MyKiva logic

Lender Profile:
Screenshot 2024-11-21 at 4 07 11 PM

Old Portfolio [UPDATED]:
image

### DIFF
--- a/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
+++ b/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
@@ -23,7 +23,7 @@
 					</p>
 					<hr class="tw-border-tertiary tw-my-4">
 					<badges-section
-						:total-possible-badges="totalPossibleBadges"
+						:is-loading="badgesLoading"
 						:completed-achievements="completedAchievements"
 					/>
 					<hr class="tw-border-tertiary tw-my-4">
@@ -178,10 +178,8 @@ export default {
 
 			return completedBadgesArr;
 		},
-		totalPossibleBadges() {
-			return this.badgesData.reduce((acc, badge) => {
-				return acc + (badge?.contentfulData?.length ?? 0);
-			}, 0);
+		badgesLoading() {
+			return !this.badgesData.length;
 		}
 	},
 	methods: {
--- a/src/pages/Portfolio/LendingStats/BadgesSection.vue
+++ b/src/pages/Portfolio/LendingStats/BadgesSection.vue
@@ -38,10 +38,9 @@ export default {
 			type: Array,
 			default: () => []
 		},
-		/* total number of possible badges */
-		totalPossibleBadges: {
-			type: Number,
-			default: 0
+		isLoading: {
+			type: Boolean,
+			default: false
 		},
 	},
 	components: {
@@ -51,9 +50,6 @@ export default {
 	computed: {
 		badgesObtained() {
 			return this.completedAchievements.length ?? 0;
-		},
-		isLoading() {
-			return this.totalPossibleBadges === 0;
 		},
 	},
 };


src/components/LenderProfile/LenderBadges.vue
```diff
@@ -13,92 +13,48 @@
...
+	badgeData,
+	fetchAchievementData,
+	fetchContentfulData,
+	completedBadges,
+} = useBadgeData(apollo);
+const isLoading = ref(true);
+const badgesTitle = computed(() => (props.lenderInfo?.name ? `${props.lenderInfo.name}'s badges` : 'Badges'));
+
+const fetchUserAchievements = async () => {
+	await fetchAchievementData(apollo, props.publicId);
```
dyersituations, line 54: Use `Promise.all`

src/components/LenderProfile/LenderBadges.vue
```diff
@@ -13,92 +13,48 @@
...
+	await fetchContentfulData(apollo);
+	isLoading.value = false;
 };
+
+const totalPossibleBadges = computed(() => badgeData.value?.length ?? 0);
```
dyersituations, line 59: Looking at the screenshot, I don't think this number is correct, it shows `15`. Could you ask product in our team channel if the # of badges should include all of the tiers? If so the number will be much larger.

src/composables/useBadgeData.js
```diff
@@ -388,6 +389,53 @@ export default function useBadgeData() {
...
+						level: 0,
+					});
+				}
+			}
+		});
+
+		return completedBadgesArr;
+	};
+
+	const completedBadges = computed(() => {
```
dyersituations, line 429: Add jsdoc comment

src/pages/Portfolio/LendingStats/BadgesList.vue
```diff
@@ -63,63 +67,30 @@ export default {
...
-			return challenge?.dateTagline?.toLowerCase() !== 'n/a';
+		getBadgeImgUrl(badge) {
+			if (badge.level === 0) {
+				return badge?.contentfulData?.[0]?.imageUrl ?? '';
+			}
+			const badgeData = badge?.contentfulData?.find(data => data.level === badge.level);
+			return badgeData?.imageUrl ?? '';
+		},
+
+		getBadgeDate(badge) {
```
dyersituations, line 91: It looks like this new method is missing some logic from `shouldShowBadgeDate`, do we need the `n/a` check?

src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -23,7 +23,7 @@
 					</p>
 					<hr class="tw-border-tertiary tw-my-4">
 					<badges-section
-						:total-possible-badges="allAchievements.length"
+						:total-possible-badges="badgesData.length"
```
dyersituations, line 26: Similarly, is this the correct count? Chat with product in the team channel.

src/composables/useBadgeData.js
```diff
@@ -388,6 +389,53 @@ export default function useBadgeData() {
...
+	 * Get completed badges of a user
+	 *
+	 * @param badges The badges to get the completed badges from
+	 * @returns Completed badges
+	 */
+	const getCompletedBadges = badges => {
+		const completedBadgesArr = [];
+
+		badges?.forEach(badge => {
+			if (badge.achievementData?.tiers?.length) {
```
dyersituations, line 402: Does this need `badge?` like the other `if` in this method?

test/unit/specs/composables/useBadgeData.spec.js
```diff
@@ -502,4 +502,52 @@ describe('useBadgeData.js', () => {
 			expect(getHighestPriorityDisplayBadge(badges)).toEqual({ id: ID_CLIMATE_ACTION, level: 3 });
 		});
 	});
+
+	describe('getCompletedBadges', () => {
+		it('should return the completed badges', () => {
```
dyersituations, line 507: Would probably be good to add a test or two for bad states, like when badges not defined.

src/composables/useBadgeData.js
dyersituations: Looks like this change is still needed.

src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -178,6 +178,11 @@ export default {
 
 			return completedBadgesArr;
 		},
+		totalPossibleBadges() {
```
dyersituations, line 181: This is no longer being used.

dyersituations: Actually it is still being used by the other component. Is there an issue passing the larger number? How is it used in the other component?

dyersituations: I think loading will always be true now.

dyersituations: If we aren't showing that value, shouldn't we just remove it entirely and use `completedAchievements` as the loading indicator?

dyersituations: If we're using a field no longer shown anywhere (or used for a displayed value) as the loading indicator, we need to find a more accurate indicator.

src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -157,19 +159,46 @@ export default {
...
+			const completedBadgesArr = this.getBadgesData();
+
+			completedBadgesArr.sort((a, b) => {
+				return new Date(a.earnedAtDate) - new Date(b.earnedAtDate);
+			});
+
+			return completedBadgesArr;
 		},
+		badgesLoading() {
+			return !this.badgesData.length;
```
dyersituations, line 182: What happens if the user is new and has no badges?

dyersituations: Ah, so it's the data for the badges themselves and will always contain data for users?

dyersituations (approved):

PR #5715, opened 2024-11-26
feat: hide badge tile when all journeys are complete

dyersituations:
https://kiva.atlassian.net/browse/MP-1108


roger-in-kiva (approved):
I guessed if there is not a selected tier we can hide it 😝

PR #5718, opened 2024-12-02
fix: tracking event trigger fixed for loans number

roger-in-kiva:
- Tracking event trigger fixed for MyKiva loans number
- As `hasActiveLoans` is a computed prop, was always false on mounted hook

### DIFF
--- a/src/components/MyKiva/BorrowerCarousel.vue
+++ b/src/components/MyKiva/BorrowerCarousel.vue
@@ -126,7 +126,8 @@ import {
 	toRefs,
 	inject,
 	onMounted,
-	onBeforeUnmount
+	onBeforeUnmount,
+	watch,
 } from 'vue';
 import {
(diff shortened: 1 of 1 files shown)


src/components/MyKiva/BorrowerCarousel.vue
```diff
@@ -257,13 +258,15 @@ const onInteractCarousel = interaction => {
 	tabs.value.tabContext.selectedIndex = interaction.value;
 };
 
-onMounted(() => {
+watch(() => loans.value, () => {
 	if (!hasActiveLoans.value) {
```
dyersituations, line 262: Do we need to also check `loans.value.length`? Or is the first change of this computed when the data loads?

roger-in-kiva: Is the first change when data loads...

PR #5722, opened 2024-12-04
fix: more issues resolved for the MyKiva TY flow

dyersituations:
https://kiva.atlassian.net/browse/MP-1017

- Fixed issue where badge module needed to be shown for guests (with equality badge)
- Fixed issue where badge rays animation was displayed while the badge data was being loaded
- Fixed issue where `getLastCompletedBadgeLevelData` didn't sort completed date desc
- Fixed issue where `getLastCompletedBadgeLevelData` didn't handle when two tiers were achieved at the same time
- Fixed issue where `getLastCompletedBadgeLevelData` returned the wrong achievement data
- Fixed issue where MyKiva TY would be shown when no loan was checked out (Kiva Card)
- Fixed issue where MyKiva TY was being shown when loans advanced a tier but didn't achieve the tier
- Fixed issue where loan/borrower images on MyKiva TY were pixelated

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -67,6 +67,7 @@ export default function createApp({
 	app.provide('kvAuth0', kvAuth0);
 	app.provide('locale', locale);
 	app.provide('$kvTrackEvent', app.config.globalProperties.$kvTrackEvent); // provide kvTrackEvent for composition api
+	app.provide('$appConfig', appConfig); // provide appConfig for composition api
 
 	// Provide application config to all components
 	app.config.globalProperties.$appConfig = appConfig;
--- a/src/components/Thanks/MyKiva/OptInModule.vue
+++ b/src/components/Thanks/MyKiva/OptInModule.vue
@@ -8,7 +8,7 @@
 						v-for="loan, index in loansToDisplay"
 						:key="loan.id"
 						:lender-name="loan?.name"
-						:lender-image-url="loan?.image?.url"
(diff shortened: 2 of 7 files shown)


src/main.js
```diff
@@ -67,6 +67,7 @@ export default function createApp({
 	app.provide('kvAuth0', kvAuth0);
 	app.provide('locale', locale);
 	app.provide('$kvTrackEvent', app.config.globalProperties.$kvTrackEvent); // provide kvTrackEvent for composition api
+	app.provide('$appConfig', appConfig); // provide appConfig for composition api
```
mcstover, line 70: Does this also cover the use case below for components? Seems like it might?

mcstover: How will this manifest in the compiled app? Maybe a globally loaded file?

mcstover (approved):

src/main.js
dyersituations: I just tested locally, and provide properties aren't added to the `this` object for components. My guess is that we could convert `this.` usages to inject, if we desired at some point.

Regarding where it exists when compiled, it doesn't look like it's any different than the exist use of the app config. The config is passed to `createApp` when the app starts, and it seems like the data remains in memory for the components. I tried compiling the app, and I'm not seeing app config props being defined, only used.

mcstover (approved):

PR #5728, opened 2024-12-09
chore: add request/response metadata to logs

hobbsh:
Making inferences from the logs is difficult at the moment because the request/response metadata is not structured (just dumped to `message`). This makes the logs much more searchable/sortable by changing the log format from this:

```
{"meta":{},"level":"info","message":"HTTP 200 GET /UI_REVISION 663ms"}
```

To this:

```
{"meta":{"req":{"url":"/UI_REVISION","headers":{"host":"localhost:8888"}},"res":{"statusCode":200},"responseTime":221},"level":"info","message":"HTTP 200 GET /UI_REVISION 221ms"}
```

### DIFF
--- a/server/util/errorLogger.js
+++ b/server/util/errorLogger.js
@@ -8,7 +8,6 @@ export const requestLogger = expressWinston.logger({
 			stringify: true
 		})
 	],
-	meta: false,
 	requestWhitelist: ['url', 'responseTime', 'level', 'message', 'headers.host', 'headers.method'],
 	msg: 'HTTP {{res.statusCode}} {{req.method}} {{req.url}} {{res.responseTime}}ms',
 	ignoreRoute: req => {


mcstover (approved):
Noting, the format we currently use was set to match other log formats across our stack. Are we / should we change those as well? CPS and Monolith come to mind.

There's also a [util method](https://github.com/kiva/ui/blob/main/src/util/logFormatter.js) that is used widely that we may want to start using differently in cases where this type of `meta` is available.

hobbsh:
> Noting, the format we currently use was set to match other log formats across our stack. Are we / should we change those as well? CPS and Monolith come to mind.
> 
> There's also a [util method](https://github.com/kiva/ui/blob/main/src/util/logFormatter.js) that is used widely that we may want to start using differently in cases where this type of `meta` is available.

So the monolith has this metadata accessible since the Caddy logs are fully structured JSON with individual request/response fields, which we can easily pull out in Loki. I am not overly concerned about everything being in the same exact format, more that we are able to do JSON field extraction in a straightforward way. So short answer, yes I think we should also update CPS so we can more easily extract request/response fields. Our Kotlin services could also probably use an update.

The reason this came up at all was because I was trying to correlate specific request urls with extended slowness in the application last night.

PR #5738, opened 2024-12-17
feat: integrate recommended loans endpoint with lending home row

michelleinez:


src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,16 +193,26 @@ export default {
...
+			enableLoanRecommendations: false,
 		};
 	},
 	apollo: {
 		preFetch(config, client) {
 			return client.query({
 				query: experimentAssignmentQuery, variables: { id: THREE_LOANS_RECOMMENDED_ROW_EXP_KEY }
-			}).then(() => {
+			}).then(async () => {
+				const loanRecommendationsExp = await client.query({
```
dyersituations, line 204: Does the await here work as expected? Asking because this `preFetch` logic is a custom implementation that we built, and we don't use async await with `preFetch` anywhere else in `ui`. I know, for example, that async await doesn't work as expected in the created method.

dyersituations: One thing to try here is actually removing this `experimentAssignmentQuery` entirely and using `readQuery`. We assign all active experiments globally on `preFetch`, so we might not need to run the assignment here.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -262,22 +284,31 @@ export default {
...
 		async getRecommendedLoans() {
-			const { loans } = await runLoansQuery(
-				this.apollo,
-				{ pageLimit: 12 },
-				FLSS_ORIGIN_LEND_BY_CATEGORY
-			);
+			let loans = [];
+			if (this.enableLoanRecommendations) {
+				const response = await runRecommendationsQuery(this.apollo, {
+					userId: this.userInfo?.id,
```
dyersituations, line 290: Do we also need to pass in `null` here if there is no ID (like in `created`)?

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,16 +193,26 @@ export default {
...
+				});
+				const useRecommendations = loanRecommendationsExp?.data?.experiment?.version === 'b';
+
 				const recommendedLoansPromise = client.query({
-					query: flssLoansQueryExtended,
-					variables: prefetchedRecommendedLoansVariables
+					query: useRecommendations ? loanRecommendationsQueryExtended : flssLoansQueryExtended,
+					variables: useRecommendations ? {
+						...prefetchedRecommendationsVariables,
+						userId: config.userId || null
```
dyersituations, line 214: I don't think we use `config.userId` elsewhere, is this the Kiva user ID? What happens if it's a guest user?

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -424,16 +455,43 @@ export default {
...
+					...prefetchedRecommendationsVariables,
+					userId: this.userInfo?.id || null
+				}
+			});
+			cachedRecommendedLoans = recommendedLoansData
+				?.loanRecommendations?.values?.filter(loan => loan !== null) ?? [];
+		} else {
+			const flssLoansData = this.apollo.readQuery({
+				query: flssLoansQueryExtended,
+				variables: {
```
dyersituations, line 488: This can be `prefetchedFlssVariables`

src/util/loanSearch/dataUtils.js
```diff
@@ -53,6 +54,37 @@ export async function runLoansQuery(apollo, loanSearchState, origin) {
...
+ * @param {Object} apollo The Apollo client instance
+ * @param {Object} options Query options including userId and origin
+ * @param {Number} options.userId The user ID to get recommendations for (optional)
+ * @param {String} options.origin Origin of query formatted as web:##page-context##
+ * @param {Object} options.filterObject Optional filters to apply to recommendations
+ * @param {String} options.sortBy Sort option for recommendations (defaults to 'personalized')
+ * @param {Number} options.limit Limit the number of recommendations (optional)
+ * @returns {Object} The results of the recommendations query
+ */
+export async function runRecommendationsQuery(apollo, {
```
dyersituations, line 69: This new method could use some unit tests as well.

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -132,12 +133,23 @@ import HandOrangeIcon from '#src/assets/images/hand_orange.svg';
...
 import KvCartModal from '#kv-components/KvCartModal';
 
-const prefetchedRecommendedLoansVariables = { pageLimit: 4, origin: FLSS_ORIGIN_LEND_BY_CATEGORY };
+const prefetchedFlssVariables = {
+	pageLimit: 4,
+	origin: FLSS_ORIGIN_LEND_BY_CATEGORY
+};
+
+const prefetchedRecommendationsVariables = {
+	origin: FLSS_ORIGIN_LEND_BY_CATEGORY,
```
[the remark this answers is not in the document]
michelleinez: Rob says: "I think the same origin is fine...the main purpose of that param is to track the caller, not the API method (which we can get from our own logging)"

src/pages/LoanFinding/LoanFinding.vue
michelleinez: okay, after fixing the chaining, it still works - so I will push up that version in a bit, but I'm a little confused by the last comment since this IS the prefetch function...?

mcstover: I think he just means using the `result` from the query in another `then` statement so the 2 queries are separated. `exp query` -> `.then(result => recommendationsQuery` with data from the result

src/pages/LoanFinding/LoanFinding.vue
mcstover: Does the new loanRecommendations endpoint not handle non-logged in users? That's typically been handled by the service using the information from the auth header, or lack of information if not logged in...

I also couldn't find anywhere we reference the `userId` on `config` but if you can verify it's there after login that would be cool! Otherwise, you can add a `my > id` query somewhere above this to check for login state, where `id` would be the value you need if present.

If you need to pass `uiv` for non-logged in users, you can extract the `{ cookieStore }` in the `preFetch` params and get it from the cookie...

src/pages/LoanFinding/LoanFinding.vue
michelleinez: Ignore the config access...  I just added a userInfo query into the prefetch function.  There's a cache read in the created() function as well which I'm hesitant to remove because I don't actually know how to test user state in dev.  Do either of you know how to do that?

src/pages/LoanFinding/LoanFinding.vue
michelleinez: Ignore the config access... I just added a userInfo query into the prefetch function. There's a cache read in the created() function as well which I'm hesitant to remove because I don't actually know how to test user state in dev. Do either of you know how to do that?

michelleinez: A lack of user id still appears to return loans from the loansRecommended endpoint.

src/pages/LoanFinding/LoanFinding.vue
michelleinez: done

src/pages/LoanFinding/LoanFinding.vue
michelleinez: Okay I've read this all over a few times but still scratching my head.  @mcstover  are you talking about the promise chaining comment where casey is talking about using .then chaining, or the last comment Casey made about not needing to run an experiment assignment in the prefetch function because we assign all active experiments globally on preFetch?

The part I am confused about is the idea that "we don't need to run assignment query" on prefetch because we are "running an assignment query on prefetch".

dyersituations: I actually meant that we run `assignAllActiveExperiments` in `WwwPage`. If I remember right, you still need to run the individual assign queries if you need the assignment in a child `preFetch`, but you also might not, since the parent `preFetch` should run first. The hierarchy of pre fetch calls might run simultaneously though, that's the part I'm forgetting.

src/pages/LoanFinding/LoanFinding.vue
michelleinez: done

src/util/loanSearch/dataUtils.js
michelleinez: done

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,26 +193,49 @@ export default {
...
 		preFetch(config, client) {
 			return client.query({
-				query: experimentAssignmentQuery, variables: { id: THREE_LOANS_RECOMMENDED_ROW_EXP_KEY }
+				query: experimentAssignmentQuery,
+				variables: { id: THREE_LOANS_RECOMMENDED_ROW_EXP_KEY }
 			}).then(() => {
-				const recommendedLoansPromise = client.query({
-					query: flssLoansQueryExtended,
-					variables: prefetchedRecommendedLoansVariables
+				return client.query({
```
dyersituations, line 205: If the pre fetches run simultaneously (see my other comment a min ago), then we can keep these queries here, but we can also use Promise.all, for example: https://github.com/kiva/ui/blob/a1f117681edcf26b6bbc5cbda05d4c79a89d1085/src/pages/Lend/LoanChannelCategoryControl.vue#L451

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,26 +193,49 @@ export default {
...
-				return Promise.all([
-					client.query({ query: userInfoQuery }),
-					client.query({ query: experimentAssignmentQuery, variables: { id: FIVE_DOLLARS_NOTES_EXP } }),
-					client.query({ query: experimentAssignmentQuery, variables: { id: FLSS_ONGOING_EXP_KEY } }),
-					recommendedLoansPromise
-				]);
+			}).then(loanRecommendationsExp => {
+				const useRecommendations = loanRecommendationsExp?.data?.experiment?.version === 'b';
+
+				return client.query({ query: userInfoQuery })
```
dyersituations, line 212: Actually since the result of the previous `experimentAssignmentQuery` call isn't needed for this query, all 3 queries can be ran in a single `Promise.all`

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,26 +193,49 @@ export default {
...
+					.then(userInfoResult => {
+						const userId = userInfoResult?.data?.my?.userAccount?.id || null;
```
dyersituations, line 214: Based on the question below, this might need to be `|| this.cookieStore.get('uiv'`

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -262,22 +297,31 @@ export default {
...
 		async getRecommendedLoans() {
-			const { loans } = await runLoansQuery(
-				this.apollo,
-				{ pageLimit: 12 },
-				FLSS_ORIGIN_LEND_BY_CATEGORY
-			);
+			let loans = [];
+			if (this.enableLoanRecommendations) {
+				const response = await runRecommendationsQuery(this.apollo, {
+					userId: this.userInfo?.id || null,
```
dyersituations, line 303: Same question about visitor ID

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -424,16 +468,40 @@ export default {
...
-			query: flssLoansQueryExtended,
-			variables: prefetchedRecommendedLoansVariables
-		})?.fundraisingLoans?.values?.filter(loan => loan !== null) ?? [];
+		let cachedRecommendedLoans = [];
+		if (this.enableLoanRecommendations) {
+			const recommendedLoansData = this.apollo.readQuery({
+				query: loanRecommendationsQueryExtended,
+				variables: {
+					...prefetchedRecommendationsVariables,
+					userId: this.userInfo?.id || null
```
dyersituations, line 493: Same question about visitor ID

src/pages/LoanFinding/LoanFinding.vue
```diff
@@ -181,26 +193,49 @@ export default {
...
+
+				return client.query({ query: userInfoQuery })
+					.then(userInfoResult => {
+						const userId = userInfoResult?.data?.my?.userAccount?.id || null;
+
+						const recommendedLoansPromise = client.query({
+							query: useRecommendations ? loanRecommendationsQueryExtended : flssLoansQueryExtended,
+							variables: useRecommendations ? {
+								...prefetchedRecommendationsVariables,
+								userId
```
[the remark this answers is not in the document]
michelleinez: Visitor id is not supported by this endpoint so I should pass null.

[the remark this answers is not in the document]
dyersituations: Do we have a general fallback recommendation when `null` is provided? I know we've had that behavior for other endpoints.

src/pages/LoanFinding/LoanFinding.vue
michelleinez: Yes, Rob says null is fine.

michelleinez: I got the answer to this question by asking Rob, he says the endpoint does not support visitorId and to use null.

dyersituations: Sounds great!

src/pages/LoanFinding/LoanFinding.vue
michelleinez: @dyersituations does this promise.all version look any better?

src/pages/LoanFinding/LoanFinding.vue
michelleinez: done

src/pages/LoanFinding/LoanFinding.vue
michelleinez: done

dyersituations (approved):

src/pages/LoanFinding/LoanFinding.vue
dyersituations: Yup, looks good, queries are now ran in parallel where we can!

mcstover:
Thanks @michelleinez and @dyersituations for working through this and getting the updates in order!

PR #5744, opened 2024-12-23
feat: update opt-in module with guest mutation

christian14b:
Use new mutation to store guest selected `opt-in` in MyKiva ty page

### DIFF
--- a/src/components/Thanks/MyKiva/OptInModule.vue
+++ b/src/components/Thanks/MyKiva/OptInModule.vue
@@ -198,7 +198,7 @@ const updateOptIn = value => {
 
 	if (value) {
 		const visitorId = cookieStore.get('uiv') || null;
-		if (props.isGuest) {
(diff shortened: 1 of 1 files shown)


src/components/Thanks/MyKiva/OptInModule.vue
```diff
@@ -158,12 +150,58 @@ const updateOptIn = value => {
...
+		'post-checkout',
+		'click',
+		`${value ? 'accept' : 'reject'}-opt-in-request`,
+		props.isGuest ? 'guest' : 'signed-in',
+		props.numberOfBadges,
+	);
+
+	if (value) {
+		const visitorId = cookieStore.get('uiv') || null;
+		if (props.isGuest) {
```
dyersituations, line 201: We should probably check `visitorId` here as well.

dyersituations (approved):

PR #5748, opened 2025-01-03
feat: my kiva title should only count active loans

dyersituations:
https://kiva.atlassian.net/browse/MP-1219

- Title should use active loans in count (to match loans we display in carousel)
- Added optional chaining to address data edge case that currently only happens on dev

### DIFF
--- a/src/components/MyKiva/BorrowerCarousel.vue
+++ b/src/components/MyKiva/BorrowerCarousel.vue
@@ -168,9 +168,11 @@ const windowWidth = ref(0);
 const openWhatIsNext = ref(false);
 const lastVisitedLoanIdx = ref(0);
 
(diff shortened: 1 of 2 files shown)


roger-in-kiva:
Should we consider filtered loans? (Sliced) Those are we are currently showing in carousel.

dyersituations:
> Should we consider filtered loans? (Sliced) Those are we are currently showing in carousel.

Good question. I think showing the total pre-sliced is probably good.

roger-in-kiva:
> > Should we consider filtered loans? (Sliced) Those are we are currently showing in carousel.
> 
> 
> 
> Good question. I think showing the total pre-sliced is probably good.

Would work fine for new lenders...

roger-in-kiva (approved):

PR #5776, opened 2025-02-05
fix: badge module for new ty page

dyersituations:
https://kiva.atlassian.net/browse/MP-1318
https://kiva.atlassian.net/browse/MP-1319
https://kiva.atlassian.net/browse/MP-1313

- Ported badge module over from MyKiva TY page
- Added confetti to page load
- Created generic continue method that will be needed for several modules

image

image

image

image

image

### DIFF
--- a/src/components/Thanks/ThanksPageSingleVersion.vue
+++ b/src/components/Thanks/ThanksPageSingleVersion.vue
@@ -33,15 +33,37 @@
 				:show-receipt="showReceipt"
 			/>
 		</div>
+		<KvLightbox
+			:visible="showGuestAccountModal"
+			title="Finish creating your account to see what's next"
+			@lightbox-closed="showGuestAccountModal = false"
+		>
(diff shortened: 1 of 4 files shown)


src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -72,10 +102,18 @@ const props = defineProps({
 		type: Array,
 		default: () => ([]),
 	},
+	router: {
```
roger-in-kiva, line 105: Should we use `useRouter` composable instead of passing router function? Sometimes an anti-pattern in vue...

dyersituations: Oh yup, good call. I was following the old Vue 2 pattern in the TY page parent component. Fixed!

roger-in-kiva (approved):

PR #5781, opened 2025-02-06
feat: work in progress - initial UI for Loans Beta Page – Backend integration pending

michelleinez:


dyersituations:
@michelleinez cool to see progress on the migration work! Let us know when this is ready for review 😄

michelleinez:
> @michelleinez cool to see progress on the migration work! Let us know when this is ready for review 😄

@dyersituations @eddieferrer @emuvente
I'm trying to figure out why this build issue is blocking me from pushing my code up.
I am importing KvFlag in what appears to be the same exact way it is done in other component files.
Would either of you have ideas about why this is getting blocked?

emuvente:
> I am importing KvFlag in what appears to be the same exact way it is done in other component files.
Would either of you have ideas about why this is getting blocked?

@michelleinez I recently replaced all usages of KvFlag with the one provided from kv-components. If you rebase your branch on the latest `main`, you should see the new import syntax in the other component files:
```js
import { KvFlag } from '@kiva/kv-components';
```

michelleinez:
WIP: Initial UI for Loans Beta Page – Backend integration pending

- Added basic front-end code for the new page.
- Backend functionality is still in progress.
- This commit is for incremental review and to avoid a massive single commit.

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,234 @@
...
+									<div class="tw-mb-1">
+										${{ loan.userProperties.loanBalance.amountRepaidToLender }} repaid to you
+									</div>
+								</div>
+							</td>
+							<td class="tw-max-w-[150px] tw-whitespace-normal tw-break-words tw-px-2">
+								<div class="tw-items-center">
+									<template v-if="loan.teams?.values?.[0]">
+										<img
+											v-if="loan.teams.values[0].image?.url"
```
michelleinez, line 140: Just want to leave a comment here noting that this is still pending a backend decision on exactly where we will get the team data...  Using the [0]th element here is not the intended solution, but it does allow the component to display a dummy team for the time being.

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,234 @@
...
+				{ span: 2, marginLeft: true },
+				{ span: 2 }
+			]
+		};
+	}
+};
+</script>
+
+<style scoped>
+.scroll-gradient {
```
michelleinez, line 225: Another pending decision that needs to be run by product/design is how we want to handle the issue of this table being wider than the container containing it for the latest table component design.

src/pages/Portfolio/Loans/LoansPage.vue
```diff
@@ -0,0 +1,92 @@
...
+	},
+	created() {
+		this.apollo.query({
+			query: myLoansQuery,
+			fetchPolicy: 'network-only'
+		}).then(({ data }) => {
+			if (data?.my?.loans) {
+				console.log('LoansPage fetched data:', data.my.loans);
+				this.loans = data.my.loans.values || [];
+				this.totalLoans = data.my.loans.totalCount || this.loans.length;
```
eddieferrer, line 83: Is there a use case where `totalCount` is not available here and we need to fall back to loans.length?

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,234 @@
+<template>
+	<div class="loan-list-container tw-mt-4">
+		<div class="loan-list-scroll-container tw-relative">
+			<div class="loan-list-inner-container tw-overflow-x-auto tw-min-w-full">
```
eddieferrer, line 4: Are these classes `loan-list-container`, `loan-list-scroll-container`, and `loan-list-inner-container` being used?

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,234 @@
...
+	methods: {
+		formatDate(date) {
+			if (!date) return '';
+			return new Date(date).toLocaleDateString('en-US', {
+				month: 'short',
+				day: 'numeric',
+				year: 'numeric'
+			});
+		},
+		getPrintableStatus(rawStatus) {
```
eddieferrer, line 188: I think we can bring in these definitions from: src/api/fixtures/LoanStatusEnum.js

src/components/Portfolio/LoanStatsTable.vue
```diff
@@ -0,0 +1,165 @@
+<template>
+	<div class="tw-flex tw-justify-between tw-mb-2 tw-bg-[#efefef] tw-p-2">
```
eddieferrer, line 2: #efefef is a named color, $platinum in: src/assets/scss/_settings.scss

src/components/Portfolio/LoanStatsTable.vue
```diff
@@ -0,0 +1,165 @@
...
+			</div>
+			<div class="tw-col-span-4 tw-text-right">
+				{{ formatValue(avgStats[row.key], row.type) }}
+			</div>
+		</div>
+	</div>
+
+	<hr class="tw-border-tertiary tw-my-4">
+
+	<div class="tw-grid tw-grid-cols-1 md:tw-grid-cols-2 tw-gap-x-2">
```
eddieferrer, line 49: In general this table is not following best practices for accessibility. Is there any way we can make this an actual table? If not, I think we should at least add aria table roles to these elements: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role

src/pages/Portfolio/Loans/LoansPage.vue
michelleinez: No, that was leftover from something else.  Removed.

src/components/Portfolio/LoanStatsTable.vue
michelleinez: Okay...  I just went with tw-bg-gray-200 instead so it can be done within tailwind instead of css, if thats ok.

src/components/Portfolio/LoanList.vue
michelleinez: Not anymore, removed.

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,234 @@
+<template>
+	<div class="loan-list-container tw-mt-4">
+		<div class="loan-list-scroll-container tw-relative">
+			<div class="loan-list-inner-container tw-overflow-x-auto tw-min-w-full">
+				<table class="tw-w-full tw-min-w-[900px]">
```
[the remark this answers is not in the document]
michelleinez: Okay, I think the width setting helped since the table almost fits, but not quite. That said, even with the right-hand gradient, we’re still not sure if this is the direction we want to go because design still needs to look at this. So, I’ve removed it for now—no point in adding something new if we’re not certain we’ll use it.

src/components/Portfolio/LoanStatsTable.vue
michelleinez: Agreed for the first section -- I've converted that to a table.
The second section however wants to split into multiple columns at larger screen sizes, and I'm not sure tables support this kind of behavior the way grids do...  So I've opted for setting an aria role for the second section to address this.  Any thoughts on that?

eddieferrer (approved):

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -0,0 +1,82 @@
...
+					Filter loans
+				</div>
+				<kv-button
+					class="tw-text-sm"
+					variant="primary"
+				>
+					Export {{ totalLoans }} loans
+				</kv-button>
+			</div>
+			<kv-text-input
```
dyersituations, line 15: It looks like we have a search box but it's not connected, are we adding search functionality?

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -0,0 +1,82 @@
...
+		<div class="tw-text-secondary tw-mt-2">
+			Sorted by date posted on Kiva
+		</div>
+	</div>
+</template>
+
+<script>
+import { KvSelect, KvTextInput, KvButton } from '@kiva/kv-components';
+
+export default {
```
dyersituations, line 60: This doesn't need to change for this changeset, but keep in mind that we can (and should) use Vue 3 composition API components moving forward.

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -0,0 +1,82 @@
...
+	name: 'LoanFilterBar',
+	components: {
+		KvSelect,
+		KvTextInput,
+		KvButton,
+	},
+	data() {
+		return {
```
dyersituations, line 68: It looks like filters in general aren't implemented yet?

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,245 @@
...
+							v-for="loan in loans"
+							:key="loan.id"
+							class="tw-border-b tw-border-tertiary"
+						>
+							<td class="tw-px-2 tw-py-2">
+								<div class="tw-flex tw-items-center">
+									<img
+										:src="loan.image.url"
+										alt="Loan image"
+										class="tw-w-[50px] tw-h-[50px] tw-mr-2"
```
dyersituations, line 62: We try to avoid TW dynamic classes, since it adds to the compiled CSS file size, so it would be good to just inline styling or a class for these two dynamic classes.

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,245 @@
...
+								</div>
+							</td>
+							<td class="tw-text-right tw-px-2">
+								<div>
+									<div class="tw-mb-1">
+										${{ loan.userProperties.loanBalance.amountRepaidToLender }} repaid to you
+									</div>
+								</div>
+							</td>
+							<td class="tw-max-w-[150px] tw-whitespace-normal tw-break-words tw-px-2">
```
dyersituations, line 135: Same comment about dynamic TW class

src/components/Portfolio/LoanList.vue
```diff
@@ -0,0 +1,245 @@
...
+			]
+		};
+	}
+};
+</script>
+
+<style scoped>
+.scroll-gradient {
+	pointer-events: none;
+	position: absolute;
```
dyersituations, line 238: A lot of these can be replaced with TW classes, if you add `lang="postcss"` to the `style` tag

src/graphql/query/portfolio/myLoans.graphql
```diff
@@ -0,0 +1,61 @@
+query myLoans {
+	my {
+		id
+		loans {
```
dyersituations, line 4: It looks like `my > loans` has a default limit of `20`, so this query is probably returning only `20` loans. I'm guessing we want a way to get all loans?

src/pages/Portfolio/Loans/LoansPage.vue
```diff
@@ -0,0 +1,91 @@
...
+					repaidWithCurrencyLoss: 0,
+					defaulted: 0,
+					refunded: 0,
+					expired: 0
+				}
+			}
+		};
+	},
+	created() {
+		this.apollo.query({
```
dyersituations, line 76: I'm assuming we only want to load the loans on client side, is that correct? If so, we want to move this query to mounted.

src/pages/Portfolio/Loans/LoansPage.vue
```diff
@@ -0,0 +1,91 @@
...
+			query: myLoansQuery,
+			fetchPolicy: 'network-only'
+		}).then(({ data }) => {
+			if (data?.my?.loans) {
+				this.loans = data.my.loans.values || [];
+				this.totalLoans = data.my.loans.totalCount;
+			}
+		}).catch(error => {
+			console.error('Error fetching loans:', error);
```
dyersituations, line 85: We can use `logFormatter` here, which handles some cleaning of the message.

src/graphql/query/myLendingStats.graphql
```diff
@@ -34,6 +34,27 @@ query myLendingStats {
 		}
 		userStats {
 			amount_of_loans
+			number_of_loans
```
dyersituations, line 37: How much extra latency does this add to the query? If it's noticeable, we should create a new query for the new page, so that we don't slow down the old/existing page that uses this query.

src/components/Portfolio/LoanFilterBar.vue
michelleinez: Yes, not sure if you saw my comment about the purpose of this PR: https://github.com/kiva/ui/pull/5781#issuecomment-2655132250

The search functionality backend doesn't exist yet.

src/components/Portfolio/LoanFilterBar.vue
michelleinez: Right.

src/components/Portfolio/LoanFilterBar.vue
dyersituations: Sounds good, I wasn't sure how much the old comments still applied, thanks.

src/graphql/query/portfolio/myLoans.graphql
dyersituations: It's built into the query, so set in the backend, and there's an `offset` parameter. You can see the fields in the sandbox for dev or prod: https://gateway.development.kiva.org/graphql

dyersituations: It would be worth testing with an account to see if the limit is indeed respected. But all of our queries that I've seen with `values` returned are setup for paging with `limit` and `offset`.

src/components/Portfolio/LoanFilterBar.vue
michelleinez: Updated anyway.

src/components/Portfolio/LoanList.vue
michelleinez: Just did what I think you mean by this, lmk if anything looks off

src/graphql/query/portfolio/myLoans.graphql
michelleinez: I am currently pulling 20 from my account so it does seem like the limit is being respected.  I can add pagination in a future commit if that works...

src/pages/Portfolio/Loans/LoansPage.vue
michelleinez: yes, done.

src/graphql/query/portfolio/myLoans.graphql
dyersituations: Yeah we'll need paging to make this a replacement for the old feature. No worries if it's a following ticket.

src/components/Portfolio/LoanList.vue
dyersituations: Yup that's what I meant, thanks.

src/graphql/query/myLendingStats.graphql
dyersituations: If filtering and paging are follow-up tickets, this is the remaining open question.

michelleinez: Request Timing
Request Timing

dyersituations: Thanks for the details. I think we might as well create a new query, since we do tend to create unique queries for specific pages, and creating a new query doesn't negatively affect our app maintainability.

michelleinez: Request Timing
Actually, after reducing the query down to what this new page actually needs, we also brought the response time down significantly on the new page to ~200ms instead of 1.25 seconds.

dyersituations: Nice! Yeah some of those gql fields have a lot of monolith processing/code behind them.

dyersituations (approved):
Approved with the mindset that there are follow-up tickets for filtering and paging.

PR #5795, opened 2025-02-11
feat: add borrowers avatars to badge module mp 1336

christian14b:
image

### DIFF
--- a/.storybook/stories/ThanksPageSingleVersion.stories.js
+++ b/.storybook/stories/ThanksPageSingleVersion.stories.js
@@ -141,6 +141,15 @@ export const Badge = story({
 	myKivaEnabled: true,
 });
 
+export const BadgeLoansNotInPfpOptedIn = story({
+	isOptedIn: true,
+	lender: mockLender,
+	loans: mockLoans.slice(1),
+	receipt: receiptWithMultipleLoans,
+	badgesAchieved: [mockTieredBadge],
+	myKivaEnabled: true,
+});
+
 export const BadgeSingleLoanNotOptedIn = story({
 	lender: mockLender,
 	loans: [mockLoans[0]],
--- a/.storybook/stories/ThanksPageSingleVersion.stories.js
+++ b/.storybook/stories/ThanksPageSingleVersion.stories.js
@@ -141,7 +141,7 @@ export const Badge = story({
 	myKivaEnabled: true,
 });
 
-export const BadgeLoansNotInPfpOptedIn = story({
+export const BadgeLoansInPfp = story({
 	isOptedIn: true,
 	lender: mockLender,
 	loans: mockLoans.slice(1),
--- a/src/components/Thanks/SingleVersion/BadgeMilestone.vue
+++ b/src/components/Thanks/SingleVersion/BadgeMilestone.vue
@@ -89,7 +89,7 @@ const props = defineProps({
 		type: Array,
 		default: () => ([]),
 	},
-	showLoanComment: {
+	loanCommentModuleShown: {
 		type: Boolean,
 		default: false,
 	}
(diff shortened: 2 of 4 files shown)


src/components/Thanks/SingleVersion/BadgeMilestone.vue
```diff
@@ -7,6 +7,23 @@
...
 		<template v-else>
+			<!-- Borrower images -->
+			<div v-if="showAvatars" class="tw-flex tw-items-center">
+				<KvUserAvatar
+					v-for="(loan, index) in avatars"
+					:key="loan.id"
+					:lender-name="loan?.name"
+					:lender-image-url="loan?.image?.url"
+					class="tw-rounded-full tw-shadow tw-border-white tw-border-2 tw-w-auto"
+					:class="{ 'smaller-borrower-avatar': avatars.length > 2 && index !== 1 }"
```
roger-in-kiva, line 18: I don't see this class defined below.

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -159,12 +161,20 @@ const loanForComment = computed(() => {
...
+
+const askforComments = computed(() => hasPfpLoan.value || hasTeamAttributedPartnerLoan.value);
+
 const showOptInModule = computed(() => !props.isOptedIn);
 const showKivaCardsModule = computed(() => printableKivaCards.value.length);
 const showBadgeModule = computed(() => props.myKivaEnabled && props.badgesAchieved.length > 0);
 const showJourneyModule = computed(() => props.myKivaEnabled && !showBadgeModule.value);
 const showControlModule = computed(() => !props.myKivaEnabled);
-const showLoanComment = computed(() => !!loanForComment.value);
+const showLoanComment = computed(() => !!loanForComment.value && askforComments.value);
```
roger-in-kiva, line 177: I guess we are not using `askforComments` for this version, wondering why avatars should be hidden if there are comments here...

roger-in-kiva: I saw that in mocks, thanks clarifying!

roger-in-kiva (approved):

.storybook/stories/ThanksPageSingleVersion.stories.js
```diff
@@ -141,6 +141,15 @@ export const Badge = story({
 	myKivaEnabled: true,
 });
 
+export const BadgeLoansNotInPfpOptedIn = story({
```
dyersituations, line 144: I think is actually `BadgeLoansInPfp`?

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -159,12 +161,20 @@ const loanForComment = computed(() => {
 	return orderedLoans[0];
 });
 
+const hasPfpLoan = computed(() => props.loans.some(loan => loan?.inPfp));
+
+const hasTeamAttributedPartnerLoan = computed(
+	() => props.loans.some(loan => loan?.distributionModel === 'fieldPartner' && loan?.team?.name)
```
dyersituations, line 167: I know the ticket said to copy this functionality over, but looking at the code now, I think we probably need to instead check the specific loan, which we have here as `loanForComment`. Otherwise, the first loan might not match what we're checking for.

.storybook/stories/ThanksPageSingleVersion.stories.js
dyersituations: @christian14b one more small change, I'd recommend renaming this story, since isn't the name I suggested more accurate (opposite the original name)?

dyersituations (approved):

PR #5802, opened 2025-02-12
feat: refactor borrowers avatars component for thank you page single …

christian14b:
…version variant

This component will contain these two variants:

image

image


dyersituations (approved):

roger-in-kiva:
Wondering if we should add the same to MyKiva ThanksBadge component, not sure about the future of this... @dyersituations

dyersituations:
> Wondering if we should add the same to MyKiva ThanksBadge component, not sure about the future of this... @dyersituations 

Good question, but we can ignore the MyKiva components for now. If this TY page goes well, that code will be deleted.

roger-in-kiva (approved):

PR #5811, opened 2025-02-15
feat: hide name, and facts related to badge when no loans

christian14b:
image

image

### DIFF
--- a/src/components/Thanks/SingleVersion/BadgeMilestone.vue
+++ b/src/components/Thanks/SingleVersion/BadgeMilestone.vue
@@ -4,23 +4,23 @@
 		class="tw-rounded md:tw-rounded-lg tw-bg-white tw-shadow-lg tw-px-3 md:tw-px-8 tw-py-4 tw-flex tw-flex-col
 			tw-gap-2 print:tw-hidden tw-items-center tw-text-center tw-overflow-hidden tw-relative"
 	>
-		<BgRays v-show="!isLoading" />
 		<KvLoadingPlaceholder v-if="isLoading" class="!tw-h-9 !tw-rounded" />
 		<template v-else>
 			<!-- Borrower images -->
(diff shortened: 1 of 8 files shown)


src/components/Thanks/SingleVersion/BadgeMilestone.vue
```diff
@@ -106,13 +108,15 @@ const showEqualityBadge = computed(() => props.isGuest || props.onlyKivaCardsAnd
...
 	let title = '';
 	if (props.isOptedIn && !props.kivaCardsModuleShown) {
 		title += 'Thank you!<br />';
 	}
 
-	if (!badgeDataAchieved.value?.length) {
+	if (!loansCount.value) {
 		title += 'Take the next step on your impact journey.';
 	} else {
 		title += badgeDataAchieved.value?.length === 1
```
roger-in-kiva, line 122: Should we also check for `badgeDataAchieved` array lenght?

roger-in-kiva: Probably smt like `else if (!badgeDataAchieved.value?.length)` ...

roger-in-kiva: Makes sense, just considering a case having loans with no badges, should we show Equality?
I'm confused with those uses cases...

roger-in-kiva: We are rendering this component if there are badges only, so we are safe with this condition, thanks checking!

roger-in-kiva (approved):

src/components/Thanks/SingleVersion/BadgeMilestone.vue
dyersituations: If there are loans but no badges we should `JourneyGeneralPrompt`

dyersituations (commented):
👍

PR #5814, opened 2025-02-19
feat: adjustments to ATB experiment and removing old exp code

christian14b:
- Remove bubble cart
- Ensure desktop animation works correctly

https://github.com/user-attachments/assets/ca7f0a79-9d55-4e22-ae22-8b0e270621c1

https://github.com/user-attachments/assets/3280e19b-ccbb-4be0-8d42-9555470c6f71

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -588,7 +588,7 @@ import { readBoolSetting } from '#src/util/settingsUtils';
 import experimentVersionFragment from '#src/graphql/fragments/experimentVersion.graphql';
 import addToBasketExpMixin from '#src/plugins/add-to-basket-exp-mixin';
 import {
-	KvUserAvatar, KvButton, KvMaterialIcon, KvPageContainer
+	KvButton, KvMaterialIcon, KvPageContainer
 } from '@kiva/kv-components';
 import SearchBar from './SearchBar';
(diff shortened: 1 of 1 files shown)


src/components/LoanCards/KvClassicLoanCardContainer.vue
```diff
@@ -435,7 +435,8 @@ export default {
 			}
 		},
 		getTargetPosition() {
-			const target = document.getElementById('basket-exp');
+			const targets = [...document.querySelectorAll('[data-testid="header-basket"]')];
```
dyersituations, line 438: It looks like this will be good code to abstract out when we port this over to CPS, so we don't have duplication.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -345,66 +345,41 @@
 								}"
 							>
 								<router-link
-									v-if="enableBasketExperiment"
```
dyersituations, line 348: Instead of removing the fallback, we should be using the feature flag here. We want a way to roll this back on prod immediately if there are issues, like we have in place for the TY page. Rolling out new features to new users without fallback is dangerous when it's such a critical part of the app. Here's the recent ticket that added the feature flag: https://kiva.atlassian.net/browse/MP-1322

dyersituations: Oh, right, I misread the code, this code isn't specific to the experiment.

dyersituations (approved):

PR #5815, opened 2025-02-19
feat: guest username passed to auth flow

roger-in-kiva:
- guest username passed to auth flow

### DIFF
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -135,6 +135,11 @@ export default function authRouter(config = {}) {
 			options.login_hint = 'signUp';
 		}
 
+		// Guest Username
+		if (req.query.username) {
+			options.username = req.query.username;
+		}
+
 		info(`LoginUI: attempt login, session id:${req.sessionID}, cookie:${getSyncCookie(req)}, done url:${req.query.doneUrl}`); // eslint-disable-line max-len
 		passport.authenticate('auth0', options)(req, res, next);
 	});
--- a/src/components/Forms/GuestAccountCreation.vue
+++ b/src/components/Forms/GuestAccountCreation.vue
@@ -138,7 +138,7 @@ export default {
 						throw new Error('Missing reset url');
 					}
 					if (this.guestUsername) {
-						resetUrl += `?username=${this.guestUsername}`;
+						resetUrl += `&username=${this.guestUsername}`;
 					}
 
 					window.location = resetUrl;
--- a/src/components/Forms/GuestAccountCreation.vue
+++ b/src/components/Forms/GuestAccountCreation.vue
@@ -86,7 +86,11 @@ export default {
 		eventValue: {
 			type: Number,
 			default: 0
-		}
+		},
+		guestUsername: {
+			type: String,
+			default: ''
+		},
 	},
 	data() {
 		return {
@@ -129,10 +133,14 @@ export default {
 					if (result?.errors?.length > 0) {
 						throw result.errors;
 					}
-					const resetUrl = result?.data?.general?.startGuestAccountClaim;
+					let resetUrl = result?.data?.general?.startGuestAccountClaim;
 					if (!resetUrl) {
 						throw new Error('Missing reset url');
 					}
+					if (this.guestUsername) {
+						resetUrl += `?username=${this.guestUsername}`;
+					}
+
 					window.location = resetUrl;
 				}).catch(err => {
 					this.serverError = true;
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -876,7 +876,7 @@ export default {
 				if (this.checkingOutAsGuest) {
 					checkoutAdditionalQueryParams += `&optedIn=${this.userOptedIn}`;
 				}
-				if (payload.username) {
+				if (payload?.username) {
 					checkoutAdditionalQueryParams += `&username=${payload.username}`;
 				}
 
--- a/src/pages/Checkout/CheckoutPage.vue
+++ b/src/pages/Checkout/CheckoutPage.vue
@@ -850,7 +850,8 @@ export default {
 				this.setUpdatingTotals(false);
 			});
 		},
-		completeTransaction(transactionId) {
+		completeTransaction(payload) {
+			const transactionId = typeof payload === 'object' ? payload.transactionId : payload;
 			// compile transaction data
 			const transactionData = formatTransactionData(
 				numeral(transactionId).value(),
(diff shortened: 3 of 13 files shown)


server/auth-router.js
```diff
@@ -135,6 +135,11 @@ export default function authRouter(config = {}) {
 			options.login_hint = 'signUp';
 		}
 
+		// Guest Username
+		if (req.query.username) {
+			options.username = req.query.username;
```
dyersituations, line 140: Should we be adding this to the `login_hint` JSON?

src/pages/LoginAndRegister/GuestAccountRedirect.vue
```diff
@@ -23,15 +24,18 @@ export default {
 						}
 					}
 				}`,
-			}).then(({ data }) => {
+				}),
+				route,
```
dyersituations, line 28: Why does `route` need to be passed here to `Promise.all`, can't it just be accessed already inside this method?

src/pages/LoginAndRegister/GuestAccountRedirect.vue
```diff
@@ -44,14 +48,18 @@ export default {
...
+					const username = currentRoute.query?.username ?? '';
+
 					return Promise.reject({
 						path: '/ui-login',
 						query: {
 							loginHint: `login|${JSON.stringify({
 								guest: true,
 							})}`,
 							doneUrl: `${path}?${queryString}`,
+							username,
```
dyersituations, line 62: Related to the other comment, maybe this can be passed above in the `loginHint` JSON.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -30,6 +31,7 @@
 				:receipt="receipt"
 				:badges-achieved="badgesAchieved"
 				:router="$router"
+				:guest-username="guestUsername"
```
dyersituations, line 34: I think we can just focus on the new single TY page, so we could rollback these old components to simplify this changeset.

src/components/Forms/GuestAccountCreation.vue
```diff
@@ -129,10 +133,14 @@ export default {
 					if (result?.errors?.length > 0) {
 						throw result.errors;
 					}
-					const resetUrl = result?.data?.general?.startGuestAccountClaim;
+					let resetUrl = result?.data?.general?.startGuestAccountClaim;
```
dyersituations, line 136: Does this URL ever have other query parameters?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -850,7 +850,8 @@ export default {
 				this.setUpdatingTotals(false);
 			});
 		},
-		completeTransaction(transactionId) {
+		completeTransaction(payload) {
+			const transactionId = typeof payload === 'object' ? payload.transactionId : payload;
```
dyersituations, line 854: Are there situations where this won't be an `object` now?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -875,6 +876,9 @@ export default {
 				if (this.checkingOutAsGuest) {
 					checkoutAdditionalQueryParams += `&optedIn=${this.userOptedIn}`;
 				}
+				if (payload.username) {
```
dyersituations, line 879: Given the check above for whether an object, might be good to have optional chaining here just in case.

server/auth-router.js
roger-in-kiva: i think `login_hint` has a different purpose, so adding an extra param for `username` is good to clarifying this new purpose.

src/pages/LoginAndRegister/GuestAccountRedirect.vue
roger-in-kiva: Same as previous comment, different purpose... also, this will complicate driving query as `username` will be part of the `loginHint` param....

src/pages/Thanks/ThanksPage.vue
roger-in-kiva: This work is already done, said that, if for some reason, product feedback or whatever, we need to turn off single version this feature stills working in other cases... Otherwise this we'll easy to remove if single version is a winner having a clean up of the page in the future.

src/components/Forms/GuestAccountCreation.vue
roger-in-kiva: This URL is coming from BE response and only query param we have is `login_hint=isGuest`

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Yep, here: https://github.com/kiva/ui/blob/1766cb68be5e5105e50802dde6a177a6e6dd01e2/src/pages/Checkout/CheckoutPage.vue#L100

src/components/Forms/GuestAccountCreation.vue
dyersituations: Right, but below we are adding the new query string with `?`, can there be another query string there? If so we'll need `&`.

src/pages/LoginAndRegister/GuestAccountRedirect.vue
roger-in-kiva: This is under `prefetch`, we don't have access to `this` method...

src/components/Forms/GuestAccountCreation.vue
roger-in-kiva: Agreed, thanks clarifying.

src/pages/LoginAndRegister/GuestAccountRedirect.vue
dyersituations: Right, but `route` is being passed into `preFetch`. I see we are already using `cookieStore` below without adding it to the `Promise` resolution. Does `route` work similarly in this case?

src/components/Forms/GuestAccountCreation.vue
roger-in-kiva: Yeah, mutation exclusive for guest account claim, always returning `?login_hint=isGuest`

src/pages/LoginAndRegister/GuestAccountRedirect.vue
roger-in-kiva: Nice call, checking now...

roger-in-kiva: Works fine, thanks!

src/components/Forms/GuestAccountCreation.vue
roger-in-kiva: Should be fine if mutation could return `username` instead of passing across different components and query params... worth a try in the future.

dyersituations (approved):

PR #5817, opened 2025-02-19
feat: export button functionality

michelleinez:


src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -55,18 +57,45 @@
...
+
+		const queryParams = new URLSearchParams({
+			iDisplayStart: '0',
+			iDisplayLength: props.totalLoans.toString(),
+			user_id: userId.toString()
+		});
+
+		window.location.href = `/portfolio/loans/export?${queryParams.toString()}`;
+	} catch (error) {
+		logReadQueryError(error, 'LoanFilterBar userIdQuery');
```
eddieferrer, line 98: Should we display something to the user here?

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -55,18 +57,45 @@
...
+			return;
+		}
+
+		const queryParams = new URLSearchParams({
+			iDisplayStart: '0',
+			iDisplayLength: props.totalLoans.toString(),
+			user_id: userId.toString()
+		});
+
+		window.location.href = `/portfolio/loans/export?${queryParams.toString()}`;
```
dyersituations, line 96: The old export button downloaded a CSV, this one navigates away, is that an expected change?

michelleinez: No, it's not supposed to take the user anywhere - just download a file.  When I tested it using https://www.kiva.org/portfolio/loans/export?${queryParams.toString()} it behaved fine for me.  But when it uses the local url, there's no file returned so it probably navigates to a blank page...

michelleinez: When it goes into prod, it should use kiva.org/portfolio....

dyersituations: Ah makes sense, it looks like we used `window.location.assign` before, does `window.location.href` similarly to trigger the download?

michelleinez: Yeah it also triggers the download.

Uploading Screen Recording 2025-02-19 at 4.10.24 PM.mov…

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -55,18 +57,45 @@
...
+
+const handleExportClick = async () => {
+	try {
+		const { data } = await apollo.query({
+			query: userIdQuery
+		});
+
+		const userId = data?.my?.userAccount?.id;
+		if (!userId) {
+			console.error('No user ID available for export');
```
dyersituations, line 86: This can be replaced with `logFormatter`.

src/components/Portfolio/LoanFilterBar.vue
michelleinez: done

src/components/Portfolio/LoanFilterBar.vue
michelleinez: done

src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -83,7 +84,8 @@ const handleExportClick = async () => {
 
 		const userId = data?.my?.userAccount?.id;
 		if (!userId) {
-			console.error('No user ID available for export');
+			logFormatter('No user ID available for export');
+			this.$showTipMsg('There was a problem preparing your loan export.', 'error');
```
dyersituations, line 88: `this.$showTipMsg` won't exist for components that use composition API. If we want to show a tip message, that functionality will need to be made available and usable by the composition API. See how we `provide` different functionality in `main.js`. Might be a good conversation for the business owner/stakeholder for this migration. Or maybe we just match what we used to do in the monolith, did we show a notification on failure? If not, we can just use `logFormatter` and leave it there, since these errors shouldn't happen.

dyersituations (approved):

PR #5819, opened 2025-02-20
fix: remove active loans switch from distribution in old portfolio page

christian14b:


dyersituations:
@christian14b I thought that the new fields existing for the `my` query? Did we just not add the new fields to the public profile yet?

dyersituations:
> Oh yeah, it exists for `my` query and its working well here, Lender profile is the one missing, this is the PR #5820 . Do you want me to leave this code? The ticket referred to lender profile that's why I open this PR removing this code from old portfolio

Ah, so we added it to the wrong page in general, correct (lender profile instead of lender public profile)?

dyersituations (approved):

PR #5820, opened 2025-02-20
feat: active loans switch in lender profile

christian14b:
Active loans switch in lender profile page


dyersituations (commented):
Changes look good, but we'll need to wait to merge until monolith is released to prod. I'll wait to approve, but definitely ping us if we miss when that release goes out.

dyersituations:
@christian14b sounds good to merge!

PR #5826, opened 2025-02-24
feat: atb modal added to BP

roger-in-kiva:
- atb modal added to BP

Screenshot 2025-02-24 at 5 04 00 p m


roger-in-kiva:
> KvCartModal last update from kv-elements doesn't support the case of showing the loan, just showing some content. This was removed because of new designs. I think we're good just showing the kvCartModal without any content for now

Oh that is good to know. But we need to update all cases on `ui`, so let's add a new ticket for that.

dyersituations:
@christian14b could you add a ticket for that to the current sprint with technical details?

dyersituations (approved):

PR #5830, opened 2025-02-25
feat: use new field for amountFunded value

christian14b:
Also, adding `null` as a default value for stats to trigger fallback cases


dyersituations (approved):

dyersituations (changes requested):
We also need to update the gql query.

dyersituations (commented):
Looks good, thanks, I'll merge when the monolith changes are deployed.

dyersituations (approved):

PR #5832, opened 2025-02-26
fix: adjust monthly good page text

dyersituations:
https://kiva.atlassian.net/browse/MP-1373

### DIFF
--- a/src/pages/MonthlyGood/KivaAsExpert.vue
+++ b/src/pages/MonthlyGood/KivaAsExpert.vue
@@ -114,13 +114,13 @@ export default {
 			return `${this.yearsSinceStartCalc} years, ${this.amountFunded ? `${this.amountFunded} billion in impact` : 'billions of dollars in impact'}`;
 		},
 		description() {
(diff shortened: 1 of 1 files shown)


src/pages/MonthlyGood/KivaAsExpert.vue
```diff
@@ -114,13 +114,13 @@ export default {
 			return `${this.yearsSinceStartCalc} years, ${this.amountFunded ? `${this.amountFunded} billion in impact` : 'billions of dollars in impact'}`;
 		},
 		description() {
-			let desc = 'With billions of dollars in loans funded';
+			let desc = 'With billions of dollars';
 			if (this.amountFunded) {
 				desc = `With nearly ${this.amountFunded} billion`;
```
roger-in-kiva, line 119: Could we have billions here?

dyersituations: `billion` should be correct in the cases I can think of, it's also what we showed before.

roger-in-kiva (approved):

PR #5838, opened 2025-02-27
fix: force login screen to load for sso redirects

emuvente:
This should make it so that we always attempt an sso connection via the login screen when using the ssoRedirect parameter.

### DIFF
--- a/server/auth-router.js
+++ b/server/auth-router.js
@@ -110,6 +110,7 @@ export default function authRouter(config = {}) {
 		// Specify ssoRedirect url
 		if (req.query.ssoRedirect) {
 			options.ssoRedirect = req.query.ssoRedirect;
+			options.prompt = 'login';
 		}
 		// Enable Kiva Corp Partner version of the login UI
 		if (req.query.kivaCorpPartner) {


JoeTravisKiva (approved):
🎉

meganmcmillan (approved):
Will this keep the behavior we have now to extend the sessions? Or will this prompt a login each time? 

Edit: nevermind, I'm now understanding your comment in slack. 

> The ssohandler is on our auth0 login screen, so we need to actually load to the login page for the connect sso method to be called. Since you have an existing auth0 session, the login screen isn't shown at all. So, if we force the login screen to display when using the ssoredirect parameter, then the sso handler will actually run, and hopefully that will fix the problem.

emuvente:
> Will this keep the behavior we have now to extend the sessions? Or will this prompt a login each time?

@meganmcmillan let's try this out to confirm how it behaves. This will attempt a new oidc authorization each time instead of using the existing auth0 session, but that new authorization with the oidc provider likely will reuse the existing oidc session, so no additional manual logging in should be required. The only difference I would expect is briefly seeing the Kiva login loading page where previously it would be blank.

dyersituations (approved):

PR #5852, opened 2025-03-04
fix: encoding the full loginHint value instead of just the username

christian14b:
Encoding the whole loginHint value to avoid double encoding of the username


roger-in-kiva:
Ensure decoding before splitting by pipe on the auth0 side.

dyersituations:
@christian14b I think we should be cautious with encoding the entire login_hint, since we have a number of use cases we'd need to test. Can we instead prevent double encoding? We want it to be encoded, and if it is already getting encoded...

dyersituations (approved):

PR #5857, opened 2025-03-04
fix: showing ty page for monthly donation amount

christian14b:
- Since there's a use case (Monthly Donation from `donate/supportus`) that doesn't have a receipt, I'm moving it to the top to show its thank you page.
- Also, as `transactionId` is `null`, `monthlyDonationAmount` variable is never going to be set with the param value.

You can test with: https://kiva-ui.local/checkout/thanks?monthly_donation_amount=50.00
image

### DIFF
--- a/src/pages/Thanks/ThanksPage.vue
+++ b/src/pages/Thanks/ThanksPage.vue
@@ -445,10 +445,8 @@ export default {
 			return this.kivaCards.filter(card => card.kivaCardObject.deliveryType === 'print');
 		},
 		activeView() {
-			// Show the donation only view if the user has only donated and not lent
-			if (this.showDafThanks
-				|| (this.receipt && this.receipt?.totals?.itemTotal === this.receipt?.totals?.donationTotal)
-				|| this.monthlyDonationAmount?.length) {
+			// Show the donation only view if the user has only subscribed to a donation
+			if (this.showDafThanks || this.monthlyDonationAmount) {
 				return DONATION_ONLY_VIEW;
 			}
 			// Show the login required view if we couldn't get the receipt
@@ -458,6 +456,10 @@ export default {
 			// Show the single version view if the experiment is enabled
 			if (this.thanksSingleVersionEnabled) {
 				return SINGLE_VERSION_VIEW;
+			}
+			// Show the donation only view if the user has only donated and not lent
+			if (this.receipt?.totals?.itemTotal === this.receipt?.totals?.donationTotal) {
+				return DONATION_ONLY_VIEW;
 			}
 			// Show the MyKiva view if qualifications are met
 			if (this.myKivaEnabled) {


roger-in-kiva:
Should we avoid logging this error for this use cases?
`{"meta":{},"level":"warning","message":"Thanks page readQuery skipped due to missing transaction_id."}`

dyersituations:
> Should we avoid logging this error for this use case? `{"meta":{},"level":"warning","message":"Thanks page readQuery skipped due to missing transaction_id."}`

Good question, but since we've been logging it forever, we can probably ignore for now.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -445,6 +445,12 @@ export default {
 			return this.kivaCards.filter(card => card.kivaCardObject.deliveryType === 'print');
 		},
 		activeView() {
+			// Show the donation only view if the user has only donated and not lent
+			if (this.showDafThanks
+				|| (this.receipt && this.receipt?.totals?.itemTotal === this.receipt?.totals?.donationTotal)
```
dyersituations, line 450: Won't this line make it so that donation-only transactions won't see the new TY page? We still want the user to see the new TY page if there is a receipt and the receipt has only non-monthly donations.

roger-in-kiva:
> Should we avoid logging this error for this use cases? `{"meta":{},"level":"warning","message":"Thanks page readQuery skipped due to missing transaction_id."}`



> > Should we avoid logging this error for this use case? `{"meta":{},"level":"warning","message":"Thanks page readQuery skipped due to missing transaction_id."}`
> 
> Good question, but since we've been logging it forever, we can probably ignore for now.

Well this is the first condition in created hook and returning false. So, yeah, we were logging it forever for a good reason but exit the flow; now having a condition before, probably we should return false before checking if there is a `transaction_id`

dyersituations:
> Well this is the first condition in created hook and returning false. So, yeah, we were logging it forever for a good reason but exit the flow; now having a condition before, probably we should return false before checking if there is a `transaction_id`

True. @christian14b definitely take a look at this, maybe there's more we can skip for the monthly donation use case.

roger-in-kiva:
> > > Well this is the first condition in created hook and returning false. So, yeah, we were logging it forever for a good reason but exit the flow; now having a condition before, probably we should return false before checking if there is a `transaction_id`
> > 
> > 
> > True. @christian14b definitely take a look at this, maybe there's more we can skip for the monthly donation use case.
> 
> Sounds good. I'll take a look in the morning, but are we ok with `monthly_donation_amount` taking precedence over `transaction_id`? I think most of the `created` code is related to get the data from the receipt. I'll move

Agreed! We'll also have this consideration for auto deposit case...

roger-in-kiva:
> > > > Well this is the first condition in created hook and returning false. So, yeah, we were logging it forever for a good reason but exit the flow; now having a condition before, probably we should return false before checking if there is a `transaction_id`
> > > 
> > > 
> > > True. @christian14b definitely take a look at this, maybe there's more we can skip for the monthly donation use case.
> > 
> > 
> > Sounds good. I'll take a look in the morning, but are we ok with `monthly_donation_amount` taking precedence over `transaction_id`? I think most of the `created` code is related to get the data from the receipt. I'll move
> 
> Agreed! We'll also have this consideration for auto deposit case...

Well, probably for current `monthly_donation_amount` we need to verify if there is also a receipt, as initially this was after `transaction_id` check...

dyersituations:
Yeah it general sounds like we need a new if statement at the top of the logic flow that catches the monthly donation only scenario

dyersituations:
> Oh I saw in description `donate/supportus` and this is coming from CPS, so looks like this is not an MG lending, this is a donation case that could also being Monthly, so, worth a try asking for this use case in unified TY. Said that, makes sense having `monthly_donation_amount` at the top of the logic, as mentioned before.

Oh yup, the naming is confusing, this is a unique edge case, monthly donation. So yeah a new use case at the top to show the old page is the simplest for now.

dyersituations (approved):

PR #5863, opened 2025-03-11
feat: integrate new graphql endpoints for average kiva lender stats that were recently added for portfolio/loans-beta page

michelleinez:


emuvente (approved):

src/components/Portfolio/LoanStatsTable.vue
```diff
@@ -161,7 +161,22 @@ export default {
 		query: lendingStatsQuery,
 		result({ data }) {
 			this.stats = data?.my?.userStats ?? {};
-			this.avgStats = {};
+			this.avgStats = {
+				amount_of_loans: data?.general?.kivaStats?.avgAmountLent ?? null,
```
dyersituations, line 165: I'm not seeing `avgAmountLent` in the updated query. And is `avgAmountLent` and same as `amount_of_loans`?

michelleinez: It's not the same, but it is analogous...  The value for the user stats column that corresponds to avgAmountLent in the average kiva lender stats column is amount_of_loans.

michelleinez: Regarding avgAmountLent that is true, it's technically not an issue since it'll default to null and this is the value that it will be once the backend errors are resolved, but I can remove it for now until that happens.

dyersituations (approved):

PR #5866, opened 2025-03-11
feat: mykiva experiment flag added to atb modal container

roger-in-kiva:
- MyKiva experiment flag added to ATB modal container

### DIFF
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -87,6 +87,8 @@ const modalPosition = computed(() => {
 	return { right, top };
 });
 
+const isGuest = computed(() => !userData.value?.my);
+
 const fetchUserData = async () => {
 	await apollo.query({
 		query: userAtbModalQuery,
@@ -104,7 +106,7 @@ onMounted(async () => {
 		apollo,
 		$kvTrackEvent,
 		userData.value?.my?.userPreferences,
-		userData.value?.my?.loans?.totalCount || 0,
+		!isGuest.value ? userData.value?.my?.loans?.totalCount : 0,
 	);
 });
 </script>
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -10,17 +10,30 @@
 		:photo-path="PHOTO_PATH"
 		:basket-count="basketCount"
 		@cart-modal-closed="closeCartModal"
-	/>
+	>
+		<template
+			#content
+			v-if="myKivaExperimentEnabled"
+		>
+			<!-- ATB content -->
+		</template>
+	</KvCartModal>
 </template>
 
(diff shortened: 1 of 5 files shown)


src/util/myKivaUtils.js
```diff
@@ -157,8 +157,12 @@ export const getIsMyKivaEnabled = (apollo, $kvTrackEvent, userPreferences, loanT
 		);
 
 		// Ensure that the user continues to see MyKiva after passing the loan limit
-		// Only update the user preferences if running client-side
-		if (isMyKivaExperimentEnabled && !hasSeenMyKiva && typeof window !== 'undefined') {
+		// Only update the user preferences if running client-side and defined
+		if (isMyKivaExperimentEnabled
+			&& !hasSeenMyKiva
+			&& typeof window !== 'undefined'
+			&& typeof userPreferences !== 'undefined'
```
dyersituations, line 164: In what case is this `undefined`?

roger-in-kiva: A guest user could be in test group, mainly for this: https://kiva.atlassian.net/browse/MP-1409

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -70,6 +86,27 @@ const modalPosition = computed(() => {
...
+};
+
+onMounted(async () => {
+	await fetchUserData();
+
+	myKivaExperimentEnabled.value = getIsMyKivaEnabled(
+		apollo,
+		$kvTrackEvent,
+		userData.value?.my?.userPreferences,
+		userData.value?.my?.loans?.totalCount || 0,
```
dyersituations, line 107: We don't want to fallback to `0` but allow `undefined` to be provided in case of API error. If we need to pass in `0` for guests, we should only pass for guests.

dyersituations: For example, on TY we use `!this.isGuest ? data?.my?.loans?.totalCount : 1,`

roger-in-kiva: Technically if `userData.value?.my` is undefined is a guest user otherwise will be 0 if a new logged in user.

A difference with TY page is that you have lent instead that for ATB modal you haven't, you are 1 step before, technically 0 makes sense.

For this case, we are not using `!this.isGuest ? data?.my?.loans?.totalCount : 1` to show a different view as TY page, we are using this just to pass this condition here: https://github.com/kiva/ui/blob/a09349509d087419b6aa9e2d820e8c30167f494d/src/util/myKivaUtils.js#L144

Makes sense? Otherwise I can pass 1 as fallback, but doesn't makes sense to me as you haven't lent yet.

dyersituations: Yeah makes sense, in that case `!this.isGuest ? data?.my?.loans?.totalCount : 0` might make sense, only using the fallback for guests.

dyersituations (approved):

PR #5867, opened 2025-03-12
feat: payment history modal

michelleinez:


src/components/Portfolio/PaidAmountModal.vue
```diff
@@ -0,0 +1,66 @@
...
+			</div>
+		</kv-lightbox>
+	</div>
+</template>
+
+<script>
+import numeral from 'numeral';
+import { KvLightbox } from '@kiva/kv-components';
+
+export default {
```
dyersituations, line 35: While this component doesn't necessarily need to change now, we want to use `script setup` and composition API for new Vue components.

src/components/Portfolio/PaidAmountModal.vue
```diff
@@ -0,0 +1,66 @@
...
+	name: 'PaidAmountModal',
+	components: {
+		KvLightbox
+	},
+	props: {
+		amount: {
+			type: [Number, String],
```
dyersituations, line 42: Is this prop really two possible types?

src/components/Portfolio/PaidAmountModal.vue
michelleinez: okay, I updated it anyway.

src/components/Portfolio/PaidAmountModal.vue
michelleinez: ah no, I was testing both types and meant to change this back.  should be ok now.

src/components/Portfolio/PaidAmountModal.vue
```diff
@@ -0,0 +1,57 @@
...
+</template>
+
+<script setup>
+import { ref, computed } from 'vue';
+import numeral from 'numeral';
+import { KvLightbox } from '@kiva/kv-components';
+
+const props = defineProps({
+	amount: {
+		type: [String],
```
dyersituations, line 38: This can just be `type: String`

dyersituations (approved):

PR #5870, opened 2025-03-12
feat: first approach to my kiva atb modal

roger-in-kiva:
- First approach to MyKiva ATB modal

**Note: Upcoming work for 1 additional loan away edge case**

Screenshot 2025-03-12 at 4 07 53 p m
Screenshot 2025-03-12 at 4 08 26 p m

### DIFF
--- a/src/components/BorrowerProfile/LendCta.vue
+++ b/src/components/BorrowerProfile/LendCta.vue
@@ -5,7 +5,7 @@
 		:style="wrapperStyle"
 	>
 		<kv-atb-modal-container
-			:basket-size="basketSize"
+			:added-loan="addedLoan"
 			:cart-modal-visible="cartModalVisible"
 			@close-cart-modal="closeCartModal"
 		/>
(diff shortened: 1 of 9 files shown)


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -5,7 +5,7 @@
 		:style="wrapperStyle"
 	>
 		<kv-atb-modal-container
-			:basket-size="basketSize"
+			:added-loan="addedLoan"
```
dyersituations, line 8: It doesn't look like `addedLoan` is defined.

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -99,6 +125,25 @@ const fetchUserData = async () => {
...
 };
 
+const fetchPostCheckoutAchievements = async loanIds => {
+	await apollo.query({
+		query: postCheckoutAchievementsQuery,
+		variables: { loanIds }
+	}).then(({ data }) => {
+		const loanAchievements = data.postCheckoutAchievements?.overallProgress ?? [];
+		contributingAchievements.value = loanAchievements.filter(achievement => achievement.postCheckoutTier !== achievement.preCheckoutTier); // eslint-disable-line max-len
+		showModalContent.value = myKivaExperimentEnabled.value && contributingAchievements.value.length;
```
dyersituations, line 135: You don't need to check `myKivaExperimentEnabled.value` here since it's checked where the method is called.

dyersituations (approved):

PR #5871, opened 2025-03-13
fix: atb modal position when banner is present mp 1417

christian14b:
There was not a problem around the banner, but header position was never being updated. that's why the first value never changed.

https://github.com/user-attachments/assets/53539c6b-5253-4e71-a670-ef61b803a42a

### DIFF
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -94,6 +94,8 @@ const updateHeaderPosition = () => {
 	}
 };
 
+const updateHeaderPositionThrottled = _throttle(updateHeaderPosition, 100);
+
 const getTargetsPosition = () => {
 	const targets = [...document.querySelectorAll('[data-testid="header-basket"]')];
 	const target = targets.find(t => t?.clientHeight);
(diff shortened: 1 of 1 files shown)


roger-in-kiva:
We usually use a throttle for this events and removing listeners when unmounting.

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -153,6 +161,9 @@ onMounted(async () => {
 		userData.value?.my?.userPreferences,
 		!isGuest.value ? userData.value?.my?.loans?.totalCount : 0,
 	);
+
+	updateHeaderPosition();
```
dyersituations, line 165: Like @roger-in-kiva mentioned, we want to throttle this and remove on unmount

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -163,8 +165,18 @@ onMounted(async () => {
 	);
 
 	updateHeaderPosition();
-	window.addEventListener('scroll', updateHeaderPosition);
+
+	window.addEventListener('scroll', _throttle(() => {
```
dyersituations, line 169: The throttle call returns a new instance, so it won't be removed like expected. You need to use a variable, like here: https://github.com/kiva/ui/blob/c85874e0adc6abc2e8bd4d271185104862703175/src/composables/useIsMobile.js#L18

dyersituations (approved):

PR #5879, opened 2025-03-20
feat: add loan avatars to journey badge

christian14b:
- Adding loans that belong to badge journey as avatars
- Loans' names to description

image

### DIFF
--- a/src/composables/useBadgeData.js
+++ b/src/composables/useBadgeData.js
@@ -525,7 +525,6 @@ export default function useBadgeData() {
 	 * @param loans The loans to filter
 	 * @returns The filtered loans
 	 */
-
 	const getFilteredLoansByJourney = (badge, loans) => {
 		return loans.filter(loan => {
 			if (badge.id === ID_US_ECONOMIC_EQUALITY) {
@@ -544,10 +543,11 @@ export default function useBadgeData() {
 				return loan?.themes?.some(theme => theme?.toLowerCase() === refugeeTheme);
 			}
 			if (badge.id === ID_BASIC_NEEDS) {
-				const sector = [6, 10];
-				const themeId = 8;
+				const basicNeedsSectors = [6, 10];
+				const basicNeedsTheme = 'water and sanitation';
 
-				return sector.includes(loan?.sector?.id) || loan?.themes?.includes(themeId);
+				return basicNeedsSectors.includes(loan?.sector?.id)
+					|| loan?.themes?.some(theme => theme?.toLowerCase() === basicNeedsTheme);
 			}
 			if (badge.id === ID_CLIMATE_ACTION) {
 				const climateActionTheme = 'clean energy';
--- a/test/unit/specs/composables/useBadgeData.spec.js
+++ b/test/unit/specs/composables/useBadgeData.spec.js
@@ -895,12 +895,24 @@ describe('useBadgeData.js', () => {
 						'Agriculture'
 					]
 				},
+				{
+					id: 3,
+					themes: [
+						'Clean Energy'
+					]
+				}
 			];
 			expect(getFilteredLoansByJourney({ id: ID_CLIMATE_ACTION }, loans)).toEqual([
 				{
 					id: 2,
 					tags: [
 						'#Eco-friendly'
+					]
+				},
+				{
+					id: 3,
+					themes: [
+						'Clean Energy'
 					]
 				}
 			]);
@@ -970,6 +982,12 @@ describe('useBadgeData.js', () => {
 					sector: {
 						id: 7,
 					}
+				},
+				{
+					id: 3,
+					themes: [
+						'Water and Sanitation'
+					]
 				}
 			];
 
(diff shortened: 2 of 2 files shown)


src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -157,10 +175,29 @@ const handleBadgeClick = index => {
 		});
 	}
 };
+
+const journeyLoans = computed(() => getFilteredLoansByJourney(badgeWithVisibleTiers.value, props.loans));
```
roger-in-kiva, line 179: Why not to add slice here?

roger-in-kiva: Nevermind, you use the length for the legend...

src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -1,7 +1,21 @@
...
+				class="avatar tw-border-white tw-rounded tw-border-2"
+				:key="loan.id"
+				v-for="(loan, i) in journeyLoans.slice(0, 3)"
+				:lender-name="loan.name"
+				:lender-image-url="loan.image.url"
+				:class="{ 'tw--ml-2.5': i > 0 }"
+				:style="{ 'z-index': journeyLoans.length }"
+			/>
+			<p v-if="journeyLoans.length - 3 > 0">
+				+{{ journeyLoans.length - 3 }}
```
roger-in-kiva, line 14: Wondering if we should use `totalLoans` instead of `journeyLoans.length`...

src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -8,10 +8,10 @@
 				:lender-name="loan.name"
 				:lender-image-url="loan.image.url"
 				:class="{ 'tw--ml-2.5': i > 0 }"
-				:style="{ 'z-index': journeyLoans.length }"
+				:style="{ 'z-index': journeyTotalLoans }"
 			/>
-			<p v-if="journeyLoans.length - 3 > 0">
-				+{{ journeyLoans.length - 3 }}
+			<p v-if="extraLoanCount">
+				+{{ extraLoanCount }}
```
roger-in-kiva, line 14: Sorry not specifying, I meant this: https://github.com/kiva/ui/blob/e568547f9620b2eadd89d4cc231c633fef0d2e46/src/pages/Portfolio/MyKiva/MyKivaPage.vue#L306

roger-in-kiva (approved):

src/components/MyKiva/BadgeModalContentJourney.vue
roger-in-kiva: NVM, just related to specific journey.

src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -1,7 +1,21 @@
...
+				class="avatar tw-border-white tw-rounded tw-border-2"
+				:key="loan.id"
+				v-for="(loan, i) in journeyLoans.slice(0, 3)"
+				:lender-name="loan.name"
+				:lender-image-url="loan.image.url"
+				:class="{ 'tw--ml-2.5': i > 0 }"
+				:style="{ 'z-index': journeyTotalLoans }"
+			/>
+			<p v-if="extraLoanCount">
+				+{{ extraLoanCount }}
```
dyersituations, line 14: We'll want to check if `extraLoanCount` is greater than 0.

src/composables/useBadgeData.js
```diff
@@ -518,6 +518,31 @@ export default function useBadgeData() {
 		}
 	};
 
+	const getFilteredLoansByJourney = (badge, loans) => {
+		const filter = getFilteredUrl(badge);
```
dyersituations, line 522: I don't think this covers the basic needs journey, since the filter URL doesn't include all of the filters. Check the channel filters to make sure that's the case, and see if there's an alternative mapping we can use. Actually the map file we use for channels on the category page might be sufficient.

dyersituations: Oh right, we needed those filters to match for the limitations on `/lend/filter`. But in the achievement service, we have one more filter. So we need to make sure that we also have an OR on the theme ID (shown below). We might need to add a new prop to the map file, maybe called `achievementFilters` or something along those lines. Please check the achievements in the Admin V2 UI on dev to make sure that the filters are covered correctly for all achievements.

```
"filters": [
    {
      "sectorId": {
        "any": [
          6,
          10
        ]
      }
    },
    {
      "themeId": {
        "any": [
          8
        ]
      }
    }
  ]
```

image

test/unit/specs/composables/useBadgeData.spec.js
```diff
@@ -822,4 +822,192 @@ describe('useBadgeData.js', () => {
...
+						'Refugees/Displaced'
+					]
+				}
+			]);
+		});
+
+		it('should return expected filtered loans for basic-needs', () => {
+			const { getFilteredLoansByJourney } = useBadgeData();
+
+			const loans = [
```
dyersituations, line 961: Add a loan here with theme as well, to cover this more unusual filter requirement when compared to the other milestones.

dyersituations (approved):

PR #5883, opened 2025-03-25
feat: check loan ids in basket instead of only added loan when atb

christian14b:
Evaluating milestone progress with whole basket

Jira: https://kiva.atlassian.net/browse/MP-1424


dyersituations:
@christian14b please add Jira ticket URL to PR desc so that it gets connected automatically to the ticket

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -91,6 +92,7 @@ const categoryNames = {
 
 const $kvTrackEvent = inject('$kvTrackEvent');
 const apollo = inject('apollo');
+const cookieStore = inject('cookieStore');
```
dyersituations, line 95: Is this component used in any stories? I don't think so, but if it is, we'll need to ensure cookie store is available.

dyersituations (approved):

PR #5885, opened 2025-03-26
feat: deprecate borrow precheck page

eddieferrer:
Confirmed with Eli in CIT-2299 - this is part of the work to move /borrow to be a contentful driven page.


mcstover (approved):
Can you please spin up a task to add redirects for this deprecated section? Where should it go, we probably don't want the links to 404...

eddieferrer:
> Can you please spin up a task to add redirects for this deprecated section? Where should it go, we probably don't want the links to 404...

I will add a task. Where should we add those redirects? In flux? Or CPS?

PR #5890, opened 2025-03-26
fix: check if loan is group first for pill message

christian14b:
https://kiva.atlassian.net/browse/MP-1440


src/components/BorrowerProfile/LendCta.vue
```diff
@@ -677,6 +681,8 @@ export default {
 				id: this.loanId,
 				name: this.name ?? '',
 				gender: this.loan?.gender ?? '',
+				borrowerCount: this.loan?.borrowerCount ?? [],
```
dyersituations, line 684: Should the fallback here be `1`?

dyersituations (approved):

PR #5897, opened 2025-04-01
fix: use options api to ensure perfetch works on new donations page

dyersituations:
https://kiva.atlassian.net/browse/MP-1308

- Turns out we need to use options API to ensure child components with options API prefetch like expected -> there was pop-in with user status in header (which shouldn't happen in `ui`)
- Fixed another minor pop-in issue with loader found while testing

### DIFF
--- a/src/pages/Portfolio/Donations/DonationsPage.vue
+++ b/src/pages/Portfolio/Donations/DonationsPage.vue
@@ -12,13 +12,13 @@
 						<p class="tw-text-gray-500 tw-font-medium">
 							Donor name:
 						</p>
-						<h2 class="tw--mt-1 tw-mb-2 tw-break-words">
+						<h2 class="tw-mb-2 tw-break-words">
 							<KvLoadingPlaceholder
 								v-if="loading"
-								class="tw-mt-1 !tw-w-1/4 tw-inline-block"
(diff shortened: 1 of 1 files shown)


src/pages/Portfolio/Donations/DonationsPage.vue
```diff
@@ -272,69 +266,89 @@ import myDonationsQuery from '#src/graphql/query/portfolio/myDonations.graphql';
 import { mdiHelpCircle } from '@mdi/js';
 import KvLoadingSpinner from '#src/components/Kv/KvLoadingSpinner';
```
mcstover, line 267: There's a version of this in kv-components as well. If we move to refactor the child components at some point we could migrate this too.

mcstover (approved):

src/pages/Portfolio/Donations/DonationsPage.vue
dyersituations: Funny thing is that I looked for a spinner and couldn't find it. Haha oops. I replaced with the shared component, it looks better anyways.

dyersituations: Thanks for the callout!

PR #5900, opened 2025-04-01
feat: adding new condition when my kiva flag is enabled

christian14b:
Removing previous conditions from my kiva for a feature flag

### DIFF
--- a/test/unit/specs/util/myKivaUtils.spec.js
+++ b/test/unit/specs/util/myKivaUtils.spec.js
@@ -361,6 +361,15 @@ describe('myKivaUtils.js', () => {
 
 			expect(result).toBe(false);
 		});
+
+		it('should return true if myKivaFlagEnabled is true', () => {
+			apolloMock.readFragment.mockReturnValue({ version: 'b' });
+			myKivaFlagEnabled = true;
+
+			const result = getIsMyKivaEnabled(apolloMock, $kvTrackEventMock, preferencesMock, 4, myKivaFlagEnabled);
+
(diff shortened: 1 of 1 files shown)


test/unit/specs/util/myKivaUtils.spec.js
```diff
@@ -352,6 +354,13 @@ describe('myKivaUtils.js', () => {
 				},
 			});
 		});
+
+		it('should return false if myKivaFlagEnabled is false', () => {
```
dyersituations, line 358: Can we get a test for when `myKivaFlagEnabled` is `true`?

src/pages/Thanks/ThanksPage.vue
```diff
@@ -540,6 +541,9 @@ export default {
 
 		this.loans = getLoans(this.receipt);
 
+		// Enable My Kiva Page for all users
+		this.myKivaFlagEnabled = readBoolSetting(data, MY_KIVA_FOR_ALL_USERS_KEY);
```
dyersituations, line 545: Doesn't the portfolio/MyKiva page still need this change as well?

roger-in-kiva: That route doesn't exist, as we'll create a new one probably is fine for now...

dyersituations: Ah that's right, we'll only need the experiment on the home page and here.

dyersituations (approved):

roger-in-kiva (approved):

PR #5906, opened 2025-04-02
feat: roll out new donations migration

dyersituations:
https://kiva.atlassian.net/browse/MP-1445

- Migrated page passes QA, now to start release to prod
- Flux changes coming once this is live in prod

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -464,7 +464,7 @@ export default [
 		}
 	},
 	{
-		path: '/portfolio/donations-beta',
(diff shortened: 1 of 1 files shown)


roger-in-kiva (approved):

mcstover (approved):
This will allow UI render this route once recieved.

For the flux (an tilt) ingress configs we'll need to be careful to point only this route and not the sub paths/route to ui, leaving the existing sub-paths pointing to monolith.

PR #5910, opened 2025-04-03
fix: loading placeholder size in user-avatar

christian14b:
- Fixing loading sizes
- Image infinite loading issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
 				"@godaddy/terminus": "^4.11.0",
 				"@graphql-tools/load": "^7.7.0",
 				"@graphql-tools/url-loader": "^7.17.18",
-				"@kiva/kv-components": "^6.8.1",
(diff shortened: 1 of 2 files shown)


roger-in-kiva (approved):

dyersituations (approved):

src/components/Thanks/BorrowerAvatarsContainer.vue
```diff
@@ -92,6 +95,31 @@ const getLoanImageUrl = loan => {
...
+			}
+		};
+		check();
+	});
+};
+
+onMounted(() => {
+	avatarsRefs.value.forEach(async el => {
+		const img = el.imageRef;
+		await waitForImageToComplete(img);
```
dyersituations, line 118: Is this something we can do within the avatar component? Otherwise we'll have to do this change with each use case.

dyersituations (approved):

PR #5912, opened 2025-04-04
feat: my kiva page added

roger-in-kiva:
- My Kiva page added
- Portfolio page cleaned up
- Reqs: https://kiva.atlassian.net/browse/MP-1452
- `onServerPrefetch` hook doc: https://vuejs.org/api/composition-api-lifecycle.html#onserverprefetch


src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -378,7 +376,24 @@ const fetchContentfulHeroData = () => {
 		});
 };
 
+onServerPrefetch(async () => {
+	await apollo.query({
```
dyersituations, line 380: Does the hook await promises like expected? I know there are otherwise issues with waiting for promises in Vue 3, for example created doesn't wait.

roger-in-kiva: Yeah, works like expected, I tested both cases, just await promises is that we want, otherwise doesn't work prefetching apollo cache.

roger-in-kiva: Good approach for future pages on vue3...

mcstover: Thanks for verifying, it's structured as they instruct in the docs. I think it's essentially the replacement for created in vue3.

mcstover (approved):
This looks good. Please let us know when this needs to be released to prevent breaking anything currently running around the MyKiva exps.

src/pages/MyKiva/MyKivaPage.vue
roger-in-kiva: Not precisely a replacement, if you have a vue3 environment without SSR support this doesn't work, basically from docs:
**Registers an async function to be resolved before the component instance is to be rendered on the server.**

dyersituations:
@roger-in-kiva we need to think through how we're going to roll this out. If this PR is needed and released, MyKiva will be disabled. I think we might need to add back the code to the portfolio page and remove it after release.

dyersituations (changes requested):
Making as request changes, since we can't release this as is.

roger-in-kiva:
> @roger-in-kiva we need to think through how we're going to roll this out. If this PR is needed and released, MyKiva will be disabled. I think we might need to add back the code to the portfolio page and remove it after release.

Same thought, wondering if we should have this first? https://kiva.atlassian.net/browse/MP-1427

Otherwise, we need to prefetch My Kiva flag in portfolio page and redirect to this new page...

dyersituations:
> > @roger-in-kiva we need to think through how we're going to roll this out. If this PR is needed and released, MyKiva will be disabled. I think we might need to add back the code to the portfolio page and remove it after release.
> 
> Same thought, wondering if we should have this first? https://kiva.atlassian.net/browse/MP-1427
> 
> Otherwise, we need to prefetch My Kiva flag in portfolio page and redirect to this new page...

Yeah I think the tasks are connected. I think you could revert the portfolio page changes here, keeping in MyKiva for the portfolio page. So basically just adding in the new MyKiva page, ready for that task you linked.

dyersituations:
>  As prefetching for `isHeroEnabled` is now different, not as prop, reverting portfolio changes here means reverting all. Probably we can just wait to merge this, or as I mentioned redirecting to new my kiva route from portfolio page for user under my kiva flag.

Sounds good, let's just wait to merge!

dyersituations:
Added solution that allows us to rollout the /mykiva route AND keep the old portfolio/MyKiva experiment

mcstover (approved):

roger-in-kiva:
@dyersituations looks like working fine both cases.
I don't see this prop in my kiva component: https://github.com/kiva/ui/blob/7ed78416ae002c926f59869403c0f7e000ecba43/src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue#L5

We could probably remove that logic from portfolio.

roger-in-kiva:
Also, we need to find a way to show this work under those new conditions: https://kiva.atlassian.net/browse/MP-1434

dyersituations:
> We could probably remove that logic from portfolio.

@roger-in-kiva good call, I already added a link to the commit that removes this code to the ticket to remove the feature flag, so we can remove that code then.

And for the badge section, when the flag to show MyKiva to all users is enabled, the old portfolio page is shown, so we can load the MyKiva experiment assignment and show that section.

dyersituations (approved):

roger-in-kiva:
@dyersituations already removed for consistency, thanks!

PR #5913, opened 2025-04-04
fix: optin notification avatar sizes

christian14b:
Before:
image

After:
image


roger-in-kiva (approved):

dyersituations (approved):
@christian14b have you tested all other usages of the shared component? I forget how often we've used this particular avatar.

PR #5922, opened 2025-04-07
feat: filter only public tags

christian14b:
Only show public tags in loan card


roger-in-kiva:
@christian14b can you verify this is also happening for live loans data here? https://github.com/kiva/ui/blob/main/server/util/live-loan/live-loan-fetch.js

Have you checked other loan cards we have in this repo?

mcstover (approved):
Good catch on adding the vocabularyId!

dyersituations:
@christian14b please see the question added by @roger-in-kiva about other loan cards and live loans. I forget if they have loan tags displayed. Any other changes needed in this repo?

PR #5930, opened 2025-04-09
feat: persist apollo cache in session storage

emuvente:
This allows for two things that noticeably improve performance:

- The apollo cache is maintained when the page is refreshed
- The apollo cache is **shared between ui and cps!** (see kiva/cms-page-server#1927)

I tried this out in the tilt VM and navigating between pages I had already visited was *much* faster. I didn't spot any problems with stale data or anything else, though I haven't tried checking out. We've already used the 'network-only' fetch policy for many of the queries that need to be up-to-date, so I don't expect that this would cause problems in any important user journeys.

I've set this up with a feature flag so that we can test it out in development and staging for a while before moving to production. The main difference is that data will stay cached much longer than it has in the past, so we should be looking out for any data that appears stale or appears different with a new session.


dyersituations (approved):

mcstover (approved):
Other notes in the CPS repo apply here too

PR #5940, opened 2025-04-11
fix: show milestone in checkout right after a non-tiered milestone achieved

dyersituations:
https://kiva.atlassian.net/browse/MP-1479


roger-in-kiva (approved):
Nice!

PR #5943, opened 2025-04-14
feat: reduce gap between slides

christian14b:
image

### DIFF
--- a/src/components/Contentful/JourneyCardCarousel.vue
+++ b/src/components/Contentful/JourneyCardCarousel.vue
@@ -14,12 +14,11 @@
 			:key="orderedSlides.length"
 			:embla-options="{
 				loop: false,
-				align: slidesAligmment,
-				slidesToScroll,
+				align: 'start',
 			}"
 			:slide-max-width="singleSlideWidth"
 			:multiple-slides-visible="true"
-			class="journey-card-carousel tw-w-full md:tw-overflow-visible"
+			class="journey-card-carousel tw-w-full tw-overflow-visible"
 		>
 			<template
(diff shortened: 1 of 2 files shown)


src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -269,6 +269,10 @@ watch(orderedSlides, (newSlides, oldSlides) => {
 	@apply tw-hidden;
 }
 
+.journey-card-carousel:deep(div:first-child) {
+	@apply tw-gap-2;
```
roger-in-kiva, line 273: This won't work for smaller screen size like iPhone SE.
My guess is better to follow smt like here: https://github.com/kiva/ui/blob/0082a12e6f75a0bec89db9671d3e6dca76de6348/src/components/MyKiva/BorrowerCarousel.vue#L231

Based on viewport size.

src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -269,6 +275,10 @@ watch(orderedSlides, (newSlides, oldSlides) => {
...
```
roger-in-kiva, line 279: Should we keep previous gap?

roger-in-kiva: Makes sense. Ensure happens what Nathan mentioned here: https://kivahq.slack.com/archives/C06BYUF84L8/p1744664806625809?thread_ts=1744405224.499089&cid=C06BYUF84L8

dyersituations (approved):

roger-in-kiva (approved):

PR #5945, opened 2025-04-15
feat: ty page query refactor

roger-in-kiva:
- TY page query split to separate receipt data from user, experiments and other data
- This is required for some upcoming work related to non-transactional TY page doesn't include transaction id
- Reqs: https://kiva.atlassian.net/browse/MP-1393

Upcoming work:
https://kiva.atlassian.net/browse/MP-1341
https://kiva.atlassian.net/browse/MP-1342
https://kiva.atlassian.net/browse/MP-1391
https://kiva.atlassian.net/browse/MP-1500


emuvente (approved):
Great refactor!

src/pages/Thanks/ThanksPage.vue
```diff
@@ -301,46 +302,55 @@ export default {
...
-			return client.query({
-				query: thanksPageQuery,
-				variables: {
-					checkoutId: transactionId,
-					visitorId: cookieStore.get('uiv') || null,
+			return Promise.all([
+				client.query({ query: thanksPageQuery }),
+				client.query({ query: experimentAssignmentQuery, variables: { id: 'share_ask_copy' } }),
+			]).then(() => {
+				const currentRoute = route.value ?? route ?? {};
```
dyersituations, line 309: Do we still want to do this check before running the two queries? It seems like they don't rely on the results.

src/pages/Thanks/ThanksPage.vue
```diff
@@ -301,46 +302,55 @@ export default {
...
-				};
-				const limit = 1;
 
-				return Promise.all([
-					client.query({ query: experimentAssignmentQuery, variables: { id: 'share_ask_copy' } }),
-					teamId ? fetchGoals(client, limit, filters) : null,
-					fetchPostCheckoutAchievements(client, getLoanIds(loans)),
-				]);
+				return client.query({
+					query: thanksPageReceiptQuery,
```
dyersituations, line 323: If we can move the transaction ID check above, can this query be combined in the `Promise.all` above?

roger-in-kiva: For this and comment above, to resolve promise on line 319 I think this is a good approach.

dyersituations: Right, but we resolved before running the other queries before the refactor. Does that not work after the refactor?

dyersituations: Ah gotcha, makes sense, thanks for the explanation!

dyersituations (approved):

PR #5958, opened 2025-04-17
feat: set /mykiva page users in experiment when they hit homepage links

christian14b:
https://kiva.atlassian.net/browse/MP-1427

Update routes that send users to homepage to `/mykiva` if they are in the experiment and flag is enabled

### DIFF
--- a/src/composables/useMyKivaHome.js
+++ b/src/composables/useMyKivaHome.js
@@ -35,13 +35,13 @@ export default function useMyKivaHome(apollo, $kvTrackEvent) {
 	onMounted(async () => {
 		await fetchUserData();
 
-		redirectToMyKivaHomepage.value = userData.value?.id && getIsMyKivaEnabled(
+		redirectToMyKivaHomepage.value = getIsMyKivaEnabled(
 			apollo,
 			$kvTrackEvent,
 			userData.value?.userPreferences,
 			userData.value?.loans?.totalCount ?? 0,
 			myKivaFlagEnabled.value,
-		);
+		) && userData.value?.id;
 	});
 
 	return {
--- a/src/plugins/my-kiva-homepage-mixin.js
+++ b/src/plugins/my-kiva-homepage-mixin.js
@@ -24,12 +24,12 @@ export default {
 		const userData = data?.my ?? null;
 		const myKivaFlagEnabled = readBoolSetting(data, MY_KIVA_FOR_ALL_USERS_KEY);
 
-		this.redirectToMyKivaHomepage = userData?.id && getIsMyKivaEnabled(
+		this.redirectToMyKivaHomepage = getIsMyKivaEnabled(
 			this.apollo,
 			this.$kvTrackEvent,
 			userData?.userPreferences,
 			userData?.loans?.totalCount ?? 0,
 			myKivaFlagEnabled,
-		);
+		) && userData?.id;
 	}
 };


src/components/NotFound/NotFoundWrapper.vue
```diff
@@ -31,3 +31,16 @@
 		</p>
 	</div>
 </template>
+
+<script setup>
+import {
+	inject,
+} from 'vue';
+import useMyKivaHome from '#src/composables/useMyKivaHome';
```
dyersituations, line 39: Looks like `useMyKivaHome` composable is missing.

src/composables/useMyKivaHome.js
```diff
@@ -0,0 +1,43 @@
...
+			myKivaFlagEnabled.value = readBoolSetting(data, MY_KIVA_FOR_ALL_USERS_KEY);
+		}).catch(e => {
+			logFormatter(e, 'Modal ATB User Data');
+		});
+	};
+
+	onMounted(() => {
+		fetchUserData();
+
+		console.log('hello from useMyKivaHome');
```
dyersituations, line 29: Remove.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -53,7 +53,7 @@
 						<router-link
 							v-show="!isVisitor"
 							:id="myKivaMenuId"
-							to="/portfolio"
+							:to="redirectToMyKivaHomepage ? '/mykiva' : '/portfolio'"
```
dyersituations, line 56: Sorry I didn't call it out specifically, but we want the nav portfolio link to still go to the old portfolio page. This work is to make the homepage MyKiva basically for logged-in users.

src/components/Thanks/MyKiva/ThanksBadges.vue
```diff
@@ -236,6 +236,9 @@ const props = defineProps({
 	},
 });
 
+const $kvTrackEvent = inject('$kvTrackEvent');
```
dyersituations, line 239: This file can be reverted to reduce churn.

src/components/WwwFrame/TheHeader.vue
dyersituations: So with that in mind, we want the header "Kiva" logo to take the user to `/mykiva`, if they are logged in.

src/plugins/my-kiva-homepage-mixin.js
```diff
@@ -0,0 +1,27 @@
...
+	},
+	created() {
+		const myKivaData = this.apollo.readQuery({
+			query: myKivaForAllUsersQuery,
+		});
+
+		const userData = myKivaData?.my ?? null;
+		const myKivaFlagEnabled = readBoolSetting(myKivaData, MY_KIVA_FOR_ALL_USERS_KEY);
+
+		this.redirectToMyKivaHomepage = getIsMyKivaEnabled(
```
dyersituations, line 19: We only want to redirect the user if they are logged in, since visitors can't see MyKiva.

src/components/WwwFrame/WwwPage.vue
```diff
@@ -68,6 +69,7 @@ export default {
 		preFetch(_, client) {
 			return Promise.all([
 				client.query({ query: hasEverLoggedInQuery }),
+				client.query({ query: myKivaForAllUsersQuery }),
```
dyersituations, line 72: While I like the idea of caching this once, I think we should be cautious with adding this to ALL page requests. We have many uses that will never need this flag (search bots, blog pages, campaigns, etc.). Try to identify a way to share some caching but not run this query on all pages. If we need to instead load the settings directly on a few pages, we should be fine losing some code simplification for better performance.

src/components/LoanCollections/KivaClassicLoanCategorySelector.vue
```diff
@@ -41,7 +41,7 @@ export default {
 			const categoryId = category?.id ?? null;
 			const categoryShortName = category?.shortName ?? '';
 			// build event category from url, special case for homepage, clean page path otherwise
-			const eventContext = this.$route.path === '/'
+			const eventContext = this.$route.path === '/' || this.$route.path === '/mykiva'
```
dyersituations, line 44: I don't think this component is used on mykiva? Do we need this change?

src/components/NotFound/NotFoundWrapper.vue
```diff
@@ -8,7 +8,7 @@
 		</p>
 
 		<div class="tw-flex tw-flex-col tw-text-center tw-my-5 tw-gap-y-2 tw-text-subhead">
-			<router-link to="/">
+			<router-link :to="redirectToMyKivaHomepage ? '/mykiva' : '/'">
```
dyersituations, line 11: The use of the composable could be simplified through all components by instead providing a method that returns the URL.

src/plugins/my-kiva-homepage-mixin.js
```diff
@@ -0,0 +1,27 @@
...
+import { getIsMyKivaEnabled, MY_KIVA_FOR_ALL_USERS_KEY } from '#src/util/myKivaUtils';
+import { readBoolSetting } from '#src/util/settingsUtils';
+
+export default {
+	data() {
+		return {
+			redirectToMyKivaHomepage: false,
+		};
+	},
+	async created() {
```
dyersituations, line 11: Can this be moved to mounted and still work as expected? Otherwise this will run server-side on all page requests in the header.

dyersituations:
@christian14b like mentioned in the ticket, we do want to take the user to mykiva when they are navigating to the portfolio from the TY page.

dyersituations:
> Yeah, I reverted all changes in TY page actions, basically just touched header and not found components here

Right, but we do want to take the user to mykiva from the TY page, instead of the portfolio page.

roger-in-kiva:
> > > Yeah, I reverted all changes in TY page actions, basically just touched header and not found components here
> > 
> > 
> > Right, but we do want to take the user to mykiva from the TY page, instead of the portfolio page.
> 
> Got it! but Isn't this code doing that?
> 
> https://github.com/kiva/ui/blob/2714bb97903f88ad20ce942a2932ff731376b17e/src/components/Thanks/ThanksPageSingleVersion.vue#L234
> 
> Seems like it's just using the flag value.. should I change it to use `myKivaExperimentEnabled` value instead?

Ideally, we might keep this as mi kiva is rolled out to all user and prevent prod behavior is we have a release before.

dyersituations:
> > > Yeah, I reverted all changes in TY page actions, basically just touched header and not found components here
> > 
> > 
> > Right, but we do want to take the user to mykiva from the TY page, instead of the portfolio page.
> 
> Got it! but Isn't this code doing that?
> 
> https://github.com/kiva/ui/blob/2714bb97903f88ad20ce942a2932ff731376b17e/src/components/Thanks/ThanksPageSingleVersion.vue#L234
> 
> Seems like it's just using the flag value.. should I change it to use `myKivaExperimentEnabled` value instead?

Looks like the TY page change was removed from the PR.

src/composables/useMyKivaHome.js
```diff
@@ -0,0 +1,51 @@
...
+	});
+
+	const portfolioPath = computed(() => {
+		return redirectToMyKivaHomepage.value ? '/mykiva' : '/portfolio';
+	});
+
+	onMounted(async () => {
+		await fetchUserData();
+
+		redirectToMyKivaHomepage.value = userData.value?.id && getIsMyKivaEnabled(
```
dyersituations, line 38: Changes are looking good! Quick question for validation, when testing locally, are the experiment assignments being logged like expected:
- Any page where these updated links are being used
- Only showing assignment analytics when new feature flag is enabled

dyersituations: With the latest commit are experiment assignments looking correct on the pages that need to log MyKiva assignment? And the assignment is correctly not being logged when not part of the experiment bucket?

roger-in-kiva:
Just a reminder to consider same on CPS header...

src/composables/useMyKivaHome.js
roger-in-kiva: A bit lost here, `getIsMyKivaEnabled` is to get experiment flag under those conditions is what we are expecting? I mean, not rolling out to all users. But repeat, a bit lost here...

dyersituations: The last parameter is passing in the override for getting all users. Does that make sense?

roger-in-kiva: Looks like is passed to get the experiment:
https://github.com/kiva/ui/blob/e029fb84ceab9762e2a4e877c3439fae23a3b1d4/src/util/myKivaUtils.js#L144

dyersituations: Right, as expected I think. Are you seeing issues with the current changeset?

roger-in-kiva: Just ensure that is expected.
Thanks clarifying!

dyersituations (approved):

PR #5964, opened 2025-04-21
fix: change hp path only if my kiva flag is enabled

christian14b:
- Only change hp path if My Kiva flag is enabled


dyersituations (approved):
Thanks!

PR #5967, opened 2025-04-22
fix: remove twitter from social share

The-Best-Joke:
Jira ticker: [MP-1446](https://kiva.atlassian.net/browse/MP-1446) | [MP-1475](https://kiva.atlassian.net/browse/MP-1475) | [MP-1447](https://kiva.atlassian.net/browse/MP-1447)

Screenshots:
MyKiva Share
Screenshot 2025-04-24 at 4 57 21 PM

Thank You Share
Screenshot 2025-04-24 at 4 57 58 PM

[MP-1446]: https://kiva.atlassian.net/browse/MP-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-1475]: https://kiva.atlassian.net/browse/MP-1475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MP-1447]: https://kiva.atlassian.net/browse/MP-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Kv/KvSocialShareButton.vue
```diff
@@ -42,17 +42,17 @@
...
 					@click="showSharePopUp(
-						twitterShareUrl({utmCampaign, utmContent}),
-						'Thanks for tweeting!')"
+						blueskyShareUrl({utmCampaign, utmContent}),
+						'Thanks for sharing!')"
 				>
 					<kv-material-icon
-						class="social-button__icon social-button__icon--twitter"
-						:icon="mdiTwitter"
+						name="bluesky" title="Bluesky"
```
dyersituations, line 52: Can you use the icon in the same fashion as in SocialShareV2? I see that icon appearing correctly in storybook locally.

src/components/Kv/KvSocialShareButton.vue
```diff
@@ -42,17 +42,17 @@
...
+						class="social-button__icon social-button__icon--bluesky"
```
dyersituations, line 53: It doesn't look like this class currently exists, but maybe it will need to after using the `kv-icon` component here like in `SocialShareV2`.

src/components/Thanks/ShareChallenge.vue
```diff
@@ -202,7 +217,7 @@ export default {
 			}
 			return getFullUrl(`${base}/${invitedBy}`, args);
 		},
-		// Expected by social-sharing-mixin (used by X/Twitter and "copy link")
+		// Expected by social-sharing-mixin (used by X/Twitter/Bluesky and "copy link")
```
dyersituations, line 220: You can just remove X and Twitter here.

src/plugins/social-sharing-mixin.js
```diff
@@ -46,19 +46,18 @@ export default {
 			);
 			this.$showTipMsg(thanksText);
 		},
-		twitterShareUrl({ utmCampaign, utmContent }) {
+		blueskyShareUrl({ utmCampaign, utmContent }) {
```
dyersituations, line 49: I know sometimes non-deployed behavior with social medias works oddly, but I did notice that this link only opens the root of bluesky. Should we be seeing a new post locally when testing? Does bluesky have that functionality in general?

src/components/Kv/KvSocialShareButton.vue
The-Best-Joke: Unfortunately it's not quite as simple as using it as in `SocialShareV2`.

The `kv-material-icon` component is expecting the raw `d` path from an `mdi` style icon import as is typical of the icon libraries. Because of the current way that we're importing svgs into the project, the only way to obtain the d path from imported svgs is by doing a very, very hackish method by rendering the svg on the template with no visibility and then using references to obtain the path of said svg to pass it into the `kv-material-icon` icon parameter. Not a very clean solution.

The other alternative that I also tried was creating the mdi-esque icon in the `kv-ui-elements/@kiva/kv-components` repo and importing from there, but that led into the whole separate rabbit hole of trying to wrangle that local repo as the package for the `ui` repo too.

I still think that the best solution IS to make the icon part of the `@kiva/kv-components` package, but it's more work than expected making that whole thing work locally and part of me is tempted to just get permission to create the icon as a formal package release, and then locally we can just update the `@kiva/kv-components` package and leverage the icon without going through dependency hell.

mcstover: @dyersituations I wonder if this is a good time to install and use directly the new mdi icons package. I know we haven't discussed this much yet but it's likely we end up installing that in each app where it's needed like we do today with mdi/js...

dyersituations: Release the icon first, then importing is indeed a fine path.

Alternatively, my suggestion was more trying to use `<kv-icon name="bluesky" title="Bluesky" class="social__icon" />` here instead of `kv-material-icon` as a temporary solution while we update the library. But if adding the icon to the library, then testing locally seems like a fine path, good to try that instead first.

mcstover: Package in question: https://github.com/marella/material-design-icons/tree/main/svg#readme

dyersituations: @mcstover we have a separate ticket for that work, and unfortunately that library doesn't have the bluesky icon either.

dyersituations: But yes, adding it eventually to the components lib does make sense!

mcstover: We should then be able in use individual svgs and our existing svg import handle it like we do any other `.svg` import. It appears we are doing something like `.svg#url` in UI so there may a little more config needed.

src/plugins/social-sharing-mixin.js
The-Best-Joke: I went the extra mile or hardcoding some urls based on their own documentation to see the behavior (just in case that the local code had an error – which incidentally it did because I was adding an extra `?text=` bit in the url), but it seems that bluesky's default behavior when a user visits the compose link to share when not authenticated is not to prompt for a login but instead just redirects to the homepage.

src/components/Kv/KvSocialShareButton.vue
The-Best-Joke: @dyersituations Added a halfway solution that introduces an identical button to the `kv-button` s  present (simply adding the `kv-icon` to the `kv-button` did not work). I think it's close enough to be a good middle solution, at least for now.

dyersituations: @The-Best-Joke sounds good, go ahead and add a new ticket to the same sprint and epic for replacing the icon with an SVG from the component library. And feel free to dig into that now.

dyersituations:
@The-Best-Joke it looks like the shared component also updates the BP, so you can add this ticket to the PR description, so it gets linked in Jira: https://kiva.atlassian.net/browse/MP-1447

src/components/Thanks/ThanksPageDonationOnly.vue
```diff
@@ -78,16 +78,16 @@
...
 							@click="showSharePopUp(
-								twitterShareUrl({utmCampaign, utmContent}),
-								'Thanks for tweeting!')"
+								blueskyShareUrl({utmCampaign, utmContent}),
+								'Thanks for sharing!')"
 						>
-							<kv-icon name="twitter" title="Twitter" class="social__icon" />
-							<span>Tweet your followers</span>
+							<kv-icon name="bluesky" title="Bluesky" class="social__icon" />
+							<span>Share to your followers</span>
```
dyersituations, line 90: While testing this locally, I found something funny that has been broken on this page for ages. It turns out that the `message` sent to the social share is always empty, so the share dialog never opens the social media website on this page. This page is going away soonish, but for how could you give `message` in `data` a default of `Kiva is an easy way to make a real difference in someone's life.` to match other sharing?

The-Best-Joke: Yup, will do.

src/assets/inline-svgs/logos/bluesky-logo.svg
dyersituations, line 1: Do we need the second SVG?

src/components/BorrowerProfile/MoreAboutLoan.vue
```diff
@@ -139,7 +139,7 @@ export default {
 				facebook: '',
 				instagram: '',
 				linkedin: '',
-				twitter: '',
+				bluesky: '',
```
dyersituations, line 142: This line probably needs to be reverted for now as well.

src/assets/inline-svgs/logos/bluesky-logo.svg
The-Best-Joke: I could try and fill in the color in the svg added to `icons/inline` but I figured I may as well maintain current project structure: All the other icons have their svg logo counterpart with that specific greenish color of the Borrower Page.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -350,36 +350,36 @@ export default {
 					content: this.facebookPageLabel
 				},
 			] : []).concat([
-				// Twitter Tags
-				{ name: 'twitter:title', vmid: 'twitter:title', content: this.shareTitle },
+				// bluesky Tags
```
dyersituations, line 353: Are these same tags recognized by bluesky?

src/pages/ContentfulPage.vue
```diff
@@ -182,8 +182,8 @@ export default {
 					content: this.title?.replace('| Kiva', ''),
 				},
 				{
-					name: 'twitter:title',
-					vmid: 'twitter:title',
+					name: 'bluesky:title',
```
dyersituations, line 185: Same question about whether bluesky recognizes these tags.

src/pages/BorrowerProfile/BorrowerProfile.vue
dyersituations: Comment applies to the other files too.

The-Best-Joke: I'll look into it 😅

src/components/BorrowerProfile/BorrowerBusinessDetails.vue
```diff
@@ -60,7 +60,7 @@ import * as Sentry from '@sentry/vue';
 import { KvTextLink } from '@kiva/kv-components';
 import { defineAsyncComponent } from 'vue';
 
-const TwitterLogo = defineAsyncComponent(() => import('#src/assets/inline-svgs/logos/twitter-logo.svg'));
+const BlueskyLogo = defineAsyncComponent(() => import('#src/assets/inline-svgs/logos/bluesky-logo.svg'));
```
dyersituations, line 63: So this file covers the businesses we work with, so we should leave this as twitter for now. We're just changing the share dialog on the borrower profile, which your changes to the share component already does.

dyersituations: That means we probably don't need the new icon after all.

src/pages/BorrowerProfile/BorrowerProfile.vue
dyersituations: I'm wondering if we can leave the tags as twitter for now, across the pages in the changeset.

The-Best-Joke: Yeah I was being dumb: Bluesky simply leverages the twitter tags

dyersituations: Cool good to know!

src/components/BorrowerProfile/BorrowerBusinessDetails.vue
The-Best-Joke: Ahh! Ok.

I'll revert these changes.

src/pages/BorrowerProfile/BorrowerProfile.vue
```diff
@@ -350,7 +350,7 @@ export default {
 					content: this.facebookPageLabel
 				},
 			] : []).concat([
-				// Twitter Tags
+				// twitter Tags
```
dyersituations, line 353: Uppercase to reduce file history churn.

src/pages/LenderProfile/LenderProfile.vue
```diff
@@ -64,7 +64,7 @@ export default {
 					content: this.pageTitle,
 				},
 			]).concat([
-				// Twitter Tags
+				// twitter Tags
```
dyersituations, line 67: Uppercase to reduce file history churn.

src/components/Thanks/ThanksPageDonationOnly.vue
dyersituations: Otherwise, I think adding this string default to `message` in this file is that last change needed!

dyersituations (approved):
Nice! Feel free to merge this when the checks are complete and ticket the icon library work (releasing with an icon, then testing it locally in `ui` without `npm link`).

src/components/Thanks/ThanksPageDonationOnly.vue
The-Best-Joke: I added the default message, but looking at the code, I'm not entirely sure it's ever even used?

dyersituations: `shareMessage` is used in the mixin

dyersituations: So it should be good.

PR #5969, opened 2025-04-22
fix: redirect to my kiva hp only if flag is enabled

christian14b:
https://kiva.atlassian.net/browse/MP-1515

- For being able to recognize `nonContributing` loans, added a new object to work as a "hash table" which will work as
`table['womens-equality'] = currentTierInBasketForWomensEquality`. This table will be updated after the logic to save new reached tiers.
- FIxing how to separate `contributing` and `nonContributing` loan achievements using `tierTable`: Current issue was that when basket reached a new tier, `nonContributing` was always empty since `preCheckoutTier` will always be different than `postCheckoutTier`, that's why we needed a new variable (`tierTable`) to be able to identify that loan achievements where `preCheckoutTier !== postCheckoutTier` could be `nonContribuingLoans` 
- Now OneLoanAway achievement will search in all tiers target instead of `target` since target will change after `checkout` and if we're only adding loans to the basket `target` will not change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,120 @@
+## [3.17.0](https://github.com/kiva/ui/compare/v3.16.0...v3.17.0) (2025-04-22)
+
+### 🎉 New Features
+
+* add ID field for LoanFundraisingInfo queries ([3cde42e](https://github.com/kiva/ui/commit/3cde42e846160b2eb2e6c75759ff38a42663746f))
+* add tracking to hero carousel ([9afb3c0](https://github.com/kiva/ui/commit/9afb3c0249acac54475c321c40875604e0174574))
+* adjust carousel options ([73b75bf](https://github.com/kiva/ui/commit/73b75bf14a08607ab57212d4ce03e67fe47726a7))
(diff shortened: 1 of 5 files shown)


src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -241,23 +242,44 @@ const pillMsg = computed(() => {
...
+
 const fetchPostCheckoutAchievements = async loanIds => {
 	await apollo.query({
 		query: postCheckoutAchievementsQuery,
 		variables: { loanIds }
 	}).then(({ data }) => {
 		const loanAchievements = data.postCheckoutAchievements?.overallProgress ?? [];
-		contributingAchievements.value = loanAchievements.filter(achievement => achievement.postCheckoutTier !== achievement.preCheckoutTier); // eslint-disable-line max-len
-		const nonContributingAchievements = loanAchievements.filter(achievement => achievement.postCheckoutTier === achievement.preCheckoutTier); // eslint-disable-line max-len
+		const contributingLoanAchievements = [];
```
dyersituations, line 266: The new changes are looking good, but could you add a new ticket to the sprint and epic to extract out this logic and write unit testing? Ideally we pull out ~line266-334 and write testing for all expected user paths/cases. There's enough going on here that I think we want to stabilize it with testing. Once you create the ticket, please add a comment to this thread so we can see the follow-up from this PR. Thanks!

dyersituations (approved):

PR #5970, opened 2025-04-23
fix: adjustments to how mykiva page is loaded

dyersituations:
https://kiva.atlassian.net/browse/MP-1520

- We need options API parent for components with options API `preFetch` -> the MyKiva page at the new location did not have an options API parent, so the child header component didn't prefetch data correctly
- On vue router navigation, the `onServerPrefetch` wasn't working -> I realized the hero carousel doesn't get shown until mount anyways due to waiting for lender data, so I removed `onServerPrefetch` and called query in mounted


roger-in-kiva:
`onServerPrefetch` was added for hero modal flag only… did you test adding whole mykiva query to prefetch that data as well?

Weird we need to use options api for these pages…

dyersituations:
> `onServerPrefetch` was added for hero modal flag only… did you test adding whole mykiva query to prefetch that data as well?
> 
> Weird we need to use options api for these pages…

I don't think we need to prefetch the mykiva query, since everything is loaded in mounted. The options API parent is needed to ensure that the options API `preFetch` method gets called. Keep in mind that `preFetch` is a custom method we built on options API.

roger-in-kiva (approved):

PR #5973, opened 2025-04-24
fix: local dev change requirement for newer vite version

dyersituations:
- Newer version of `vite` now requires `allowedHosts` for security reasons
- NPM auto installs this version with our current package configs


mcstover (approved):
Thanks!

PR #5978, opened 2025-04-24
feat: prevent users from loading mykiva if they are not qualified to see mykiva

dyersituations:
https://kiva.atlassian.net/browse/MP-1530

- While testing I encountered that our new `/mykiva` path is accessible to all logged-in users on dev, not just users in MyKiva experiment
- The MyKiva page now redirects during `preFetch` as necessary

### DIFF
--- a/src/graphql/query/myKivaRedirect.graphql
+++ b/src/graphql/query/myKivaRedirect.graphql
@@ -1,0 +1,23 @@
+query MyKivaRedirect {
+	my {
+		id
+		userPreferences {
+			id
+			preferences
+		}
+		lender {
+			id
+			loanCount
+		}
+	}
+	general {
+		my_kiva_all_users: uiConfigSetting(key: "my_kiva_all_users_enable") {
+			key
+			value
+		}
+	}
+	experiment(id: "my_kiva_jan_2025") @client {
+		id
+		version
+	}
+}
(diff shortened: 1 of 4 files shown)


mcstover:
@dyersituations How do this mesh with your [terraform/fastly pr](https://github.com/kiva/terraform/pull/436/files). Seems like it could create an infinite loop b/c of the simple check for `kvls` at the fastly layer.

mcstover:
I'm also wondering why we need to bother guarding the `/mykiva` route at all. Is there any real problem with someone who happens upon that route and visits it? I'm assuming there will be no links unless they are in the experiment or we redirect them to `/mykiva` intentionally.

dyersituations:
> @dyersituations How does this mesh with your [terraform/fastly pr](https://github.com/kiva/terraform/pull/436/files). Seems like it could create an infinite loop b/c of the simple check for `kvls` at the fastly layer.

@mcstover I was fixing parallel issues, and testing Fastly live on dev today for a different use case. Basically I realized that we cannot match the experiment/preferences at the CDN layer that we have in the app. But you're right that I missed the infinite loop aspect as I solved issues in parallel and isolation. Now that I think about it again, I have a simpler solution in mind. And you're right that protecting the route isn't necessary, it was mostly due to the Fastly redirect. If I find time during my nearly 4 hours of meetings tomorrow, I'll put together the simpler solution that came to mind 😝.

PR #5980, opened 2025-04-25
feat: create simple cookie for redirecting at cdn layer

dyersituations:
https://kiva.atlassian.net/browse/MP-1530

- Sets a cookie when a user sees MyKiva that can then be used at the CDN layer
- Goes along with: https://github.com/kiva/terraform/pull/438

### DIFF
--- a/src/util/myKivaUtils.js
+++ b/src/util/myKivaUtils.js
@@ -145,6 +145,18 @@ export const checkGuestAssignmentCookie = cookieStore => {
 };
 
 /**
+ * Sets a cookie to redirect the user to the MyKiva homepage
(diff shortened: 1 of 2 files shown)


roger-in-kiva (approved):

mcstover (approved):
Def need to keep in mind the cookie expiration date herein if we remove the experiment based routing.

dyersituations:
> Def need to keep in mind the cookie expiration date herein if we remove the experiment based routing.

@mcstover for sure, I'm creating a bunch of tickets for the work that happens when MyKiva is rolled out!

PR #5987, opened 2025-04-29
chore: replace twitter with bluesky in MyKiva

The-Best-Joke:
Screenshot 2025-04-25 at 5 55 32 PM


src/components/Kv/KvSocialShareButton.vue
```diff
@@ -91,14 +95,17 @@ import {
 	mdiLink,
 } from '@mdi/js';
 import socialSharingMixin from '#src/plugins/social-sharing-mixin';
-import { KvButton, KvLightbox, KvMaterialIcon } from '@kiva/kv-components';
-import KvIcon from '#src/components/Kv/KvIcon';
+import {
```
dyersituations, line 98: Can the old bluesky SVG file now be removed from the repo?

The-Best-Joke: Determining right now how easy it is to reutilize the new `KvBlueskyIcon` from the `@kiva/kv-components` package in the regular buttons as opposed to just the `KvButton`s.

The-Best-Joke: @dyersituations 

Closest we can do for now without this PR increasing in complexity would be to replace all instances of the previous Bluesky svg with a direct svg component that makes direct path use of the introduced `KvBlueskyIcon` path.

Let me know if you think this is a preferable alternative.

Screenshot 2025-04-29 at 4 48 50 PM

The-Best-Joke: As is the `KvButton` is not properly configured to fulfill the style guide for the multiple share components, and the `KvMaterialIcon` component does not play nice outside of a `KvButton` environment, which means that within a regular button a direct svg has to be leveraged.

dyersituations: We do use `KvMaterialIcon` outside of buttons elsewhere, but OK to merge this as is. The important aspect as starting the pattern of importing SVGs from the component library.

dyersituations (approved):

PR #5989, opened 2025-04-30
feat(ui): add amountLeft filter to Live Loans

The-Best-Joke:
# WIP

### DIFF
--- a/server/util/live-loan/live-loan-fetch.js
+++ b/server/util/live-loan/live-loan-fetch.js
@@ -322,6 +322,7 @@ const supportedFilterFLSS = name => {
 		case 'sector':
 		case 'theme':
 		case 'tag':
+		case 'amountleft':
 			return true;
 		default:
 			warn(`Unsupported FLSS filter "${name}"`);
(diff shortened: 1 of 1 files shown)


server/util/live-loan/live-loan-fetch.js
```diff
@@ -355,7 +364,13 @@ const parseFilterStringFLSS = async filterString => {
 		.filter(([name]) => name !== 'sort')
 		.filter(([name]) => supportedFilterFLSS(name))
 		.forEach(([name, value]) => {
-			if (name === 'country') {
+			// Special case for amountLeft with "gte" operation
+			// Value format would be "gte100" or "gte50", etc.
+			if (name === 'amountleft' && value.startsWith('gte')) {
```
dyersituations, line 369: We've used a different query string format in filtering, for example `1,2` for the min and max, but that didn't cover the equal/not-equal use case, so this should work for this test.

Do you have a working example URL for a live loan we can try locally?

The-Best-Joke: Not yet. To be frank I put the code out to get more eyes on it because I'm still not super duper sure how it's all stringed together. This was more of an intuition based on the how the `fetchRecommendationsByFilter` function was structured.

dyersituations: Cool, yeah I think this should work. Once you get the URL put together, you'll know for sure if it's doing what you want.

dyersituations (approved):

mcstover:
@The-Best-Joke These docs should give some insight into how testing can be done locally and in a live environment. Please update them to include your new filter option: https://kiva.atlassian.net/wiki/spaces/LW/pages/1800503320/Live+Loans

dyersituations:
> @The-Best-Joke These docs should give some insight into how testing can be done locally and in a live environment. Please update them to include your new filter option: https://kiva.atlassian.net/wiki/spaces/LW/pages/1800503320/Live+Loans

Thanks for the callout, already getting it covered.

PR #6000, opened 2025-05-01
fix: combo page redirect fix

roger-in-kiva:
- combo page redirect fix

### DIFF
--- a/src/pages/Lend/LoanChannelCategoryPage.vue
+++ b/src/pages/Lend/LoanChannelCategoryPage.vue
@@ -56,7 +56,7 @@ export default {
 		preFetch(config, client, args) {
 			return client.query({ query: uiConfigSettingQuery, variables: { key: CATEGORY_REDIRECT_KEY } })
 				.then(({ data }) => {
-					const query = args?.route?.query ?? {};
+					const query = args?.route?.value?.query ?? {};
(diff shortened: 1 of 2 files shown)


src/pages/Lend/LoanChannelCategoryPage.vue
```diff
@@ -56,7 +56,7 @@ export default {
 		preFetch(config, client, args) {
 			return client.query({ query: uiConfigSettingQuery, variables: { key: CATEGORY_REDIRECT_KEY } })
 				.then(({ data }) => {
-					const query = args?.route?.query ?? {};
+					const query = args?.route?.value?.query ?? {};
```
dyersituations, line 59: Do we instead need `args?.route?.value ?? args?.route`? We do something similar in other places.

roger-in-kiva: I don't see that approach in other places, also `route?.value` is a vue3 req.

dyersituations: We use `const currentRoute = route.value ?? route ?? {};` throughout `ui`

dyersituations: It's related to full refresh vs router, I think.

dyersituations (approved):

PR #6004, opened 2025-05-01
fix: fixes the FLSS multiple filters bug

The-Best-Joke:
Introduces a single filters set as opposed to an array of filter objects in order to properly OR the filters together within a single GraphQL query.

### DIFF
--- a/server/util/live-loan/live-loan-fetch.js
+++ b/server/util/live-loan/live-loan-fetch.js
@@ -338,22 +338,23 @@ const parseFilterStringFLSS = async filterString => {
 		return null;
 	}
 
-	const filters = {};
+	// Create a single filter object with all properties
+	const combinedFilter = {};
 
-	// Helper function to add to a value to an array filter
+	// Helper function to add an array filter
 	const addArrayFilterValue = (name, value) => {
-		// Make sure existing value is an array if it isn't already
-		filters[name] = filters[name] || { [name]: { any: [] } };
-		// Add the new value to the existing array
-		filters[name][name].any.push(value);
+		// Create the array if it doesn't exist
+		combinedFilter[name] = combinedFilter[name] || { any: [] };
+		// Add the value to the existing array
+		combinedFilter[name].any.push(value);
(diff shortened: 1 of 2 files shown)


server/util/live-loan/live-loan-fetch.js
```diff
@@ -385,8 +386,8 @@ const parseFilterStringFLSS = async filterString => {
 			}
 		});
 
-	const filterValues = Object.values(filters);
-	return filterValues.length ? filterValues : null;
+	// Return the combined filter in an array if any filters were added
+	return Object.keys(combinedFilter).length ? [combinedFilter] : null;
```
dyersituations, line 390: Based on the fact that we have unit tests for the previous iteration (filters as SET), I wonder if the OR behavior was previously expected for live loans. With that in mind, we either need to hold off on this, or design a way to support SETs and non-SETs. My recommendation is to maybe hold off on this work for now.

The-Best-Joke: Yeah, the existence of the previous automated tests makes me second guess the stability of this fix, even though it is working quite nicely as far as manual testing goes.

Still, the fix is here should it be found worthy.

mcstover: I'd like to second using tests to help fortify these incoming updates.

dyersituations: @The-Best-Joke how about reverting these changes and reusing this PR for testing the `amountLeft` filter?

The-Best-Joke: Due to the existence of the previous automated tests that strictly enforced the idea of every incoming filter living in its own SET, the code as-is will simply fail unless automated tests are modified.

Until such behavior can be clarified as either desired or a bug, it's best to lay off on any fix that requires changing the automated tests.

As a result, this PR will be changed to simply adding automated tests for the single filter `amountleft` use case.

dyersituations (approved):

PR #6016, opened 2025-05-07
fix: provide current route ref value from server entry

emuvente:
I'm starting to prep UI to render pages that could be cached in the CDN, and I've come across a few unrelated issues that were quick to address, so I made this PR for them.

- updated dependency install directions
- updated kv-components and kv-shop
- replaced a lodash/map usage with Object.keys().map
- made server-entry provide the current route ref value instead of the ref, fixing a couple issues where the ref value was not being accessed correctly
- got rid of the contentful preview cookie in favor of reading the preview query param from the route directly. If there is some other reason to keep this, let me know, but it looks like that cookie is deleted if the preview param isn't present, so just checking for the preview param should be equivalent.


mcstover (approved):
Should be good if the `route` is now normalized across server + client.

emuvente:
> Should be good if the route is now normalized across server + client.

Yeah, they're the same object, it's just that one is wrapped in a Ref. In components, this.$route and useRoute() are equivalent and return a [RouteLocationNormalizedLoadedGeneric](https://router.vuejs.org/api/functions/useRoute.html#Function-useRoute-). That's what is used for the prefetch hydration on the client. The server entry is using router.currentRoute, which is a [Ref\<RouteLocationNormalizedLoadedGeneric\>](https://router.vuejs.org/api/interfaces/Router.html#currentRoute). The main change I made was making server-entry always use router.currentRoute.value to unwrap the ref and provide the route object.

PR #6019, opened 2025-05-12
feat: redirect /loans-that-are-almost-funded to /almost-funded

christian14b:
https://kiva.atlassian.net/browse/MP-1576


mcstover (approved):
With both of these routes residing in UI this is ok to keep here and will handle any links from within or coming into UI ok.

`pathPrefix: "/lend-by-category/"` is set in flux so these will both continue to work.

PR #6021, opened 2025-05-12
feat: track borrower card on swipe and while using controls

christian14b:
Using `previousLastIndex` value since it's only updated when borrower avatars are clicked to distinguish tracking from carousel and avatars


roger-in-kiva (approved):
We have a confluence doc to log tracking events, can you add these pls?

PR #6022, opened 2025-05-13
chore: clean up unused configs

emuvente:
This removes all the config files that are not being used. The only configs I could find in use are `dynamic` for the deployed environments, `k8s-local` for the tilt vm, `local` for developing on localhost, and `dev-custom-host` for developing locally with caddy. I refactored `k8s-local` to inherit from `dynamic`, refactored `local` to inherit from `k8s-local`, and refactored `dev-custom-host` to inherit from `k8s-local` so that the configs can mostly be controlled by environment variables with sensible defaults for development. I was able to remove the `index` config because that only has prod values, and I was able to remove the `dev-vm` config because it had the same default development values as `dynamic` except for fbApplicationId, fbOgNameSpace, gaId, & hotjarId, which hav vm-specific values. I placed those in `k8s-local` instead.

I also found we still had some lingering perimeterX code and configs, so I removed those.


mcstover:
@emuvente We need to keep the dev-custom-host as that is used for our local caddy run mode. kiva-ui.local

config/dynamic.js
```diff
@@ -41,7 +41,7 @@ export const app = {
 		},
 		enable: true,
 		checkFakeAuth,
-		apiAudience: process.env.ALTERNATE_API_AUDIENCE || `https://api.${baseUrl}/graphql`,
+		apiAudience: process.env.ALTERNATE_API_AUDIENCE || `https://gateway.${baseUrl}/graphql`,
```
mcstover, line 44: Noting that the changed value was a default for prod, but we're already overriding with the process.env in the prod environment config so this should be ok.

emuvente: Most of the other default values in this file are for development instead of production.

dyersituations (approved):

PR #6028, opened 2025-05-19
fix: missing new hp events

christian14b:
https://kiva.atlassian.net/browse/MP-1598

Old portfolio events are present in development, new hp events were missing

### DIFF
--- a/src/components/MyKiva/BadgesSection.vue
+++ b/src/components/MyKiva/BadgesSection.vue
@@ -55,7 +55,7 @@ import useBadgeData from '#src/composables/useBadgeData';
 import { getBadgeShape, BADGE_COMPLETED, BADGE_IN_PROGRESS } from '#src/composables/useBadgeModal';
 import BadgeContainer from './BadgeContainer';
 
-const emit = defineEmits(['badge-clicked', 'next-step-event-sent']);
+const emit = defineEmits(['badge-clicked']);
 
 const $kvTrackEvent = inject('$kvTrackEvent');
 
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/BadgesSection.vue
```diff
@@ -118,6 +122,10 @@ watch(selectedJourney, () => {
 		const badge = visibleBadges.value.find(b => b.id === selectedJourney.value);
 		if (badge) {
 			badgeClicked(badge);
+			if (props.nextStepSectionClicked) {
```
dyersituations, line 125: Why is the prop needed?

src/components/MyKiva/JournalUpdatesCarousel.vue
```diff
@@ -145,8 +145,10 @@ const loadMoreUpdates = () => {
 watch(
 	() => updates,
 	() => {
-		if (updates.value.length > 0 && updates.value.length < 3) {
+		if (updates.value.length) {
```
dyersituations, line 148: In which situation was the event not being logged previously?

src/components/MyKiva/BadgesSection.vue
```diff
@@ -118,6 +122,10 @@ watch(selectedJourney, () => {
...
+				$kvTrackEvent('portfolio', 'view', 'Earn a badge', badge.challengeName, badge.level || 0);
```
dyersituations, line 126: Reading the ticket, I think this is when a hero carousel item is shown (one event per journey card), not based on a click event. This component also already has a click event logged. Could you ask Shehaam when specifically this event should be logged?

dyersituations (approved):

PR #6030, opened 2025-05-20
feat: mykiva data prefetched

roger-in-kiva:
- MyKiva page data prefetched
- https://kiva.atlassian.net/browse/MP-1603

### DIFF
--- a/src/pages/MyKiva/MyKivaPage.vue
+++ b/src/pages/MyKiva/MyKivaPage.vue
@@ -5,6 +5,12 @@
 </template>
 
 <script>
+import { readBoolSetting } from '#src/util/settingsUtils';
+import logReadQueryError from '#src/util/logReadQueryError';
+import { CONTENTFUL_CAROUSEL_KEY, MY_KIVA_HERO_ENABLE_KEY } from '#src/util/myKivaUtils';
+import myKivaQuery from '#src/graphql/query/myKiva.graphql';
+import contentfulEntriesQuery from '#src/graphql/query/contentfulEntries.graphql';
+import uiConfigSettingQuery from '#src/graphql/query/uiConfigSetting.graphql';
 import WwwPage from '#src/components/WwwFrame/WwwPage';
 import MyKivaPageContent from '#src/pages/MyKiva/MyKivaPageContent';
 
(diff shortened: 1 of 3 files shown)


src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -14,9 +20,32 @@ import MyKivaPageContent from '#src/pages/MyKiva/MyKivaPageContent';
  */
 export default {
 	name: 'MyKivaPage',
+	inject: ['apollo', 'cookieStore'],
```
dyersituations, line 23: Do we need `cookieStore`?

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -300,38 +297,32 @@ const loadMoreUpdates = () => {
...
-			};
-			loans.value = result.data?.my?.loans?.values ?? [];
-			totalLoans.value = result.data?.my?.loans?.totalCount ?? 0;
-			if (loans.value.length > 0) {
-				showLoanFootnote.value = loans.value.some(l => hasLoanFunFactFootnote(l));
-			}
-		}).catch(e => {
-			logReadQueryError(e, 'MyKivaPage myKivaQuery');
-		});
+	const result = apollo.readQuery({ query: myKivaQuery });
```
dyersituations, line 300: We probably want a try catch around the `readQuery` here and in `fetchContentfulHeroData`

dyersituations:
When I test locally I see the carousel loading initially. Can that data be prefetched?
image

src/pages/MyKiva/MyKivaPage.vue
roger-in-kiva: Looks like that, otherwise we get an error
`Error: Missing required injections! Add "inject: ['apollo', 'cookieStore']" to this component definition`

roger-in-kiva:
> When I test locally I see the carousel loading initially. Can that data be prefetched? image

Data is already prefetched in this change set, but we have a data process in the component: https://github.com/kiva/ui/blob/15a83b11cddfb3ee076678ac7ac0421aa05aaf28/src/components/Contentful/JourneyCardCarousel.vue#L357

Also, looks like a dev issue having this slow.

roger-in-kiva:
> > Probably moving badges fetch to `created`, wdyt?
> 
> Yeah it looks like in general that `watch` and loading are no longer needed with the prefetch.

I can test a refactor, as we are merging contentful with achievements data and ordering as required before showing slides.
https://github.com/kiva/ui/blob/15a83b11cddfb3ee076678ac7ac0421aa05aaf28/src/components/Contentful/JourneyCardCarousel.vue#L181

But as I mentioned, we are not prefetching achievements data, so probably we need to do that as well.
As we are using a composable in multiple components for that, we need to separate that fetching in this case.

dyersituations:
> But as I mentioned, we are not prefetching achievements data, so probably we need to do that as well. As we are using a composable in multiple components for that, we need to separate that fetching in this case.

In that case, we can merge this PR as is, and could you create a follow-up ticket? You can start working on it now, if you're interested. Thanks!

dyersituations (approved):

PR #6033, opened 2025-05-27
fix: remove bluesky svg from ui repo

The-Best-Joke:
Link to story: https://kiva.atlassian.net/browse/MP-1558

### DIFF
--- a/src/components/Thanks/ShareChallenge.vue
+++ b/src/components/Thanks/ShareChallenge.vue
@@ -84,8 +84,8 @@
 									utmCampaign,
 									selectedLoanId]"
 							>
-								<kv-material-icon name="bluesky" :icon="KvBlueskyIcon" style="width: 28px; fill: #1185FE" />
-								<span>Share</span>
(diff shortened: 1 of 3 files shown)


src/components/Checkout/SocialShareV2.vue
```diff
@@ -38,7 +38,7 @@
 					utmCampaign,
 					selectedLoanId]"
 			>
-				<kv-icon name="bluesky" title="Bluesky" class="social__icon" />
+				<kv-material-icon :icon="KvBlueskyIcon" class="social__icon"/>
```
dyersituations, line 41: Linter wants a space before `/>`

src/components/Thanks/ShareChallenge.vue
```diff
@@ -84,7 +84,7 @@
 									utmCampaign,
 									selectedLoanId]"
 							>
-								<kv-icon name="bluesky" title="Bluesky" />
+								<kv-material-icon name="bluesky" :icon="KvBlueskyIcon" style="width: 28px; fill: #1185FE" />
```
dyersituations, line 87: Why is the fill and width needed here but not the other components?

dyersituations (approved):

PR #6035, opened 2025-05-27
build: migrate vite from v5 to v6

emuvente:
Migrating vite from v5 to v6 required updating sass and storybook as well. This PR includes those updates and updates for related dependencies. The code changes for these migrations include:

- replacing `@import` with `@use` in sass
- removing unused sass variables
- reordering css rules to occur before nested media queries
- renaming `rgba` to `rgb` (`rgba` is an alias for `rbg` and `rbg` is recommended)


src/components/15Years/15YearsIndividualsProfile.vue
```diff
@@ -432,13 +432,4 @@ export default {
...
 }
-
-// Firefox looks like crap when animating blurs + transforms
-// https://bugzilla.mozilla.org/show_bug.cgi?id=925025
-@-moz-document url-prefix() {
-	.profile__portrait-img,
-	.profile__body-wrapper-wrapper {
-		filter: blur(0) !important;
-	}
-}
```
emuvente, line 443: This bug has been fixed.

src/components/15Years/15YearsTimeline.vue
```diff
@@ -241,18 +241,15 @@ export default {
...
 
-	color: $mint;
-
-	// &:hover {
-	// 	background-color: $offwhite;
-	// 	color: $twilight;
-	// }
+	& {
+		color: $mint;
+	}
```
emuvente, line 252: This is the one case where the rule still needed to come after the include, so it uses the suggested `& { ... }` syntax.

src/assets/scss/_settings.scss
```diff
@@ -204,54 +209,57 @@ $block-grid-max: 8;
...
-	),
-	large: (
-		'h1': ('font-size': rem-calc(36), 'line-height': math.div(rem-calc(44), rem-calc(36))),
-		'h2': ('font-size': rem-calc(28), 'line-height': math.div(rem-calc(36), rem-calc(28))),
-		'h3': ('font-size': rem-calc(18)),
-		'h4': ('font-size': rem-calc(16)),
-	),
-);
-$header-text-rendering: optimizelegibility;
+// $header-font-family: $body-font-family;
```
dyersituations, line 212: Do we want to keep this commented-out code?

emuvente: I'm not sure about that. The original settings file is provided by Foundation, and they expect you to copy it into your source code and modify it from there, which means that when Foundation is updated we have to go through their updated settings file and copy over any updates we might need. It would be nice to remove all the lines we aren't using entirely, but it's also nice to have the lines around still so that it's easier to spot the updates when diffing between our settings and the updated settings file in Foundation.

Ideally we would be removing Foundation and Sass entirely, but the different breakpoints between our Foundation settings and our Tailwind primitives means that converting isn't simple.

PR #6037, opened 2025-05-27
fix: change card size in my kiva carousel

christian14b:
https://kiva.atlassian.net/browse/MP-1621

We should also remove the secondary button in badge cards, but this can be done in contentful.

image


dyersituations (approved):
When this get merged in, go ahead and update the contentful settings on dev. And please create a ticket for updating the settings in prod, with the dev settings in the description. It can be in the same sprint and in the blocked column. Thanks!

PR #6059, opened 2025-06-02
feat: new badges shapes and new badges size from contentful

christian14b:
image

image


src/components/MyKiva/BadgeContainer.vue
```diff
@@ -278,8 +278,4 @@ onMounted(() => {
 		@apply tw-top-full tw-left-full;
 	}
 }
-
-.svg-background {
```
dyersituations, line 282: This is no longer needed for the patterned background?

src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -44,7 +44,7 @@
 							<BadgeContainer
 								:status="getBadgeStatus(index)"
 								:shape="getBadgeShape(badgeWithVisibleTiers.id)"
-								style="width: 133px; height: 133px;"
+								style="width: 128px; height: 128px;"
```
dyersituations, line 47: Do we also need to adjust the size for the dashed border to be closer to the outside of the badge image?

dyersituations: Or do we want the dashed line to be within the badge image now?

PR #6064, opened 2025-06-03
feat: handling gradient and completed journey

christian14b:
https://kiva.atlassian.net/browse/MP-1641

### DIFF
--- a/src/components/MyKiva/BadgeModalContentJourney.vue
+++ b/src/components/MyKiva/BadgeModalContentJourney.vue
@@ -271,7 +271,7 @@ const handleToggleGradient = e => {
 	}
 };
 
-const throttledHandleToggleGradient = _throttle(handleToggleGradient, 100);
+const throttledHandleToggleGradient = _throttle(handleToggleGradient, 300);
 
 onMounted(() => {
 	scrollEl.value = document.getElementById('sidesheet-content');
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -257,6 +262,31 @@ const journeyDescription = computed(() => {
...
+		toggleGradient.value = atBottom;
+		emit('toggle-gradient', toggleGradient.value);
+	}
+};
+
+const throttledHandleToggleGradient = _throttle(handleToggleGradient, 100);
+
+onMounted(() => {
+	// Sidesheet scrollable section
+	scrollEl.value = document.querySelector('.tw-overflow-y-auto.tw-overscroll-y-contain');
```
roger-in-kiva, line 278: As classes could change in the future, probably adding an id is better approach.

roger-in-kiva:
Is there a way to add some transition to gradient change?

src/components/MyKiva/BadgeModalContentJourney.vue
```diff
@@ -257,6 +262,30 @@ const journeyDescription = computed(() => {
 
 const RightLeaningLine = defineAsyncComponent(() => import('#src/assets/images/right-leaning-line.svg'));
 const LeftLeaningLine = defineAsyncComponent(() => import('#src/assets/images/left-leaning-line.svg'));
+
+const handleToggleGradient = e => {
+	const atBottom = e.target.scrollTop + e.target.clientHeight + 1 >= e.target.scrollHeight;
```
roger-in-kiva, line 267: We could probably increase this threshold to have a smoother change.

PR #6072, opened 2025-06-10
feat: recommended loans added to mykiva page

roger-in-kiva:
- Recommended loans carousel added to mykiva page
- https://kiva.atlassian.net/browse/MP-1635

Screenshot 2025-06-10 at 5 32 21 p m
Screenshot 2025-06-10 at 5 32 31 p m

### DIFF
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -57,6 +57,16 @@
 				/>
 			</div>
 		</section>
+		<section class="tw-my-4">
+			<LendingCategorySection
+				id="recommended-loans"
+				:title="recommendeLoansTitle"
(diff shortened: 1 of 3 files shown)


src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -57,6 +57,16 @@
 				/>
 			</div>
 		</section>
+		<section class="tw-my-4">
```
dyersituations, line 60: We probably want a loader here while the recommended loans are being loaded. All users will see this carousel.

dyersituations: Oh right, you created the placeholder empty loan data objects, never mind then!

dyersituations (approved):

PR #6075, opened 2025-06-11
test: spacing changes in sidesheet journey

christian14b:
https://kiva.atlassian.net/browse/MP-1673

As there are no designs, pushing some spacing tweaks so product can test live how the sidesheet looks

image

### DIFF
--- a/src/components/Badges/JourneySideSheet.vue
+++ b/src/components/Badges/JourneySideSheet.vue
@@ -1,5 +1,6 @@
 <template>
 	<KvSideSheet
+		class="sidesheet-journey"
 		:visible="visible"
 		:show-back-button="false"
 		:show-headline-border="false"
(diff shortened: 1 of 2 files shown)


src/components/Badges/JourneySideSheet.vue
```diff
@@ -123,3 +124,15 @@ const journeyCtaBtn = computed(() => {
 });
 
 </script>
+
+<style lang="postcss" scoped>
+
+.sidesheet-journey :deep(h2) {
```
dyersituations, line 130: Maybe we should be making these changes to the side sheet component? Is the journey the only place where these two changed components are visible for the side sheet currently?

dyersituations: The impact dashboard should be just the BP, so if the BP looks good there in the side sheet, changes should be safe. In that case, keep the ticket in progress until the changes are reviewed on dev with product, so you can finish up the shared component changes. Thanks!

PR #6079, opened 2025-06-11
fix: missing actions on click events in lending stats sidesheet

christian14b:
https://kiva.atlassian.net/browse/MP-1633

### DIFF
--- a/src/pages/Portfolio/LendingStats/BadgesList.vue
+++ b/src/pages/Portfolio/LendingStats/BadgesList.vue
@@ -176,8 +176,7 @@ export default {
 			if (this.isSelectedJourneyComplete) {
 				return this.handleSideSheetClosed();
 			}
-			console.log(this.$router);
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

src/pages/Portfolio/LendingStats/BadgesList.vue
```diff
@@ -115,6 +123,13 @@ export default {
 				...this.sortedEventBadges
 			];
 		},
+		allBadgesCompleted() {
+			const tieredBadges = this.badgesData?.filter(b => defaultBadges.includes(b?.id));
+			return tieredBadges?.every(b => !b.achievementData?.tiers?.find(t => !t?.completedDate));
```
roger-in-kiva, line 128: This could be probably simplified using `some` js array function.

dyersituations: Good call, but it might be similarly complex. I'll merge this PR because of coveralls...

PR #6091, opened 2025-06-18
feat: cacheable server-side rendering

emuvente:
This makes it possible to selectively render pages to be cached in a CDN that contain no private or user-specific data, by using [Edge Side Includes (ESI)](https://www.fastly.com/documentation/reference/vcl/statements/esi/) to get user-specific information separately from the rest of the page.

I tried to add unit tests for most of the methods I added or refactored, and for other methods in files I touched. I used copilot to generate the test code initially and updated it from there, so please check the tests for any AI oddities I may have missed.

This is made up of several major interconnected parts:

### Simulating CDN behavior for local development

> Commits: 25ce5b5, cde9b01, fb69635, 6b3b243

Processing and replacing ESI tags is normally handled by the CDN, but we don't have a CDN in local or VM environments. These commits use the `nodesi` module to process ESI tags in the rendering output when in those environments. There are other pull requests to enable ESI processing in our CDN.

These commits also remove the start-ui-session request, as that behavior is covered by the `head` ESI tag below.

### Rendering for Edge Side Includes requests and useCDNCaching mode

> Commits: a5f2652, 9224025, 2356095, 9eabdec, 8517d65, e266046

This mainly refactors `server-entry` to split between an ESI rendering mode and an App rendering mode that is handled by separate dedicated files. This also changes `main` to provide a useCDNCaching mode that is based on the flag of the same name in the meta data of the current route. This allows the CDN caching to be enabled on a per-route basis. It removes access any specific info about the user or device that made the request, so that the page can be rendered generically. The only information that is provided is whether the logic sync cookie indicates that the user making the request is logged in, which can be used to render a loading state for user-specific components. The CDN cache will be varied on that login sync cookie value, so there will be multiple cached versions of a page, one for logged-in users and one for logged-out users.

### `head` ESI tag for basket and user data

> Commits: 3e8b4c4, df95018, ce899fe, 407c83d, 1e9f273, b8536e0, daff04b, bc5b6f1, c8f65b7, 0c6b8af, 9f793ac, 1f89833

This is a replacement for all the user-specific information that is required for all rendering modes. This handles creating the visitor id if needed, creating a basket if needed, setting up external libraries based on user cookie opt-out, assigning active experiments, and providing CSS variables to render loading states.

### Apollo plugin improvements

> Commits: 4b2158e, bcf708b

To make it easier to implement all these component changes, the apollo plugin now supports multiple queries per component, and has a `shouldPreFetch` option that can be used to selectively do prefetching based on request details.

### Promo credit banner logic refactor

> Commits: c756833, d4455ed, 9d9366c, 739c8b7

This moved the logic for the promo credit banner into utility methods that could be tested independently and reused for the `head` ESI tag rendering.

### Header and other frame component changes for useCDNCaching mode

> Commits: 5a59f0d, b8536e0, 48a5f9d, 29b66f9, e4667b6, dedd1e8, 40870d0, b2a58c9, 7eb4deb, 48b4e8c, 55690de, 647f7a7, 8929f2d

These frame components were updated to render loading states or not render at all when using CDN caching. These take advantage of CSS variables with default values to allow the `head` ESI tag to configure the visibility of basket and user related elements. This even allows the user avatar to be visible during loading, as images can be defined with the css content property.

Screenshot 2025-06-18 at 11 53 55 AM
Screenshot 2025-06-18 at 11 53 38 AM

### useCDNHeaders composable for per-page caching rules

> Commits: 621ea8b

This composable has the same API as the useCDNHeaders composable from `nuxt-multi-cache` being used in CPS, and it sets the appropriate CDN headers for Fastly. This includes setting caching rules for the UI site map page to use as the first example UI page cached in the CDN.

### Other minor improvements

- [test: add test for promise module](https://github.com/kiva/ui/commit/3d7d4020356cfba864bf5d20373b620635b0c913)
- [feat: add server directory alias](https://github.com/kiva/ui/commit/ad4f0bdc6c4508d07fd045e8aed99200cbc35bf8)
- [refactor: loanid -> loanId](https://github.com/kiva/ui/commit/f5af4fe936447ad24112de78bf286312b26b3fb9)
- [refactor: replace webpack-merge with deepmerge](https://github.com/kiva/ui/commit/9d30a532a32c0b62593da8cfd2cb44da4f278f16)
- [fix: remove web manifest in local development](https://github.com/kiva/ui/commit/09ec054018cba9c4e250ac3d4f567e16be0020ad)


server/util/protectedRoutes.js
```diff
@@ -14,6 +14,5 @@ export default [
 	'register',
 	'settings',
 	'start-verification',
-	'UI_REVISION',
-	'ui-site-map'
```
dyersituations, line 18: Why was site map removed?

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -73,15 +78,19 @@ export default {
...
 			isRegionsLoading: true,
 			isChannelsLoading: true,
+			isUserDataLoading: false,
 			showMGUpsellLink: false,
 		};
 	},
 	apollo: {
 		query: pageQuery,
 		preFetch: true,
+		shouldPreFetch(config, { renderConfig }) {
```
dyersituations, line 90: Do we need to remember to use this on all components now that prefetch? And is this just for options API components?

src/components/WwwFrame/LendMenu/TheLendMenu.vue
```diff
@@ -73,15 +78,19 @@ export default {
 				'Asia',
 				'Oceania'
 			],
-			loadingSemaphore: 0,
 			isRegionsLoading: true,
 			isChannelsLoading: true,
+			isUserDataLoading: false,
```
dyersituations, line 83: Similarly, do we need to do this for all components loading user data now?

src/components/WwwFrame/TheHeader.vue
```diff
@@ -39,45 +34,35 @@
...
+									<kv-loading-placeholder />
+								</div>
+								<template v-else>
+									{{ basketCount - lcaLoanCount }}
+								</template>
 							</span>
 							Basket
 						</span>
-						<router-link
+						<my-kiva-button
```
dyersituations, line 57: I think we might want to rename this button, since MyKiva is no longer a portfolio replacement. So maybe this component should be renamed to `PorfolioButton`?

src/components/WwwFrame/LendMenu/TheLendMenu.vue
emuvente: Apollo plugin still only works for the options api. We could probably refactor the functionality so that it could be provided with a composable as well for the composition api.

src/components/WwwFrame/TheHeader.vue
emuvente: I was considering that, but I'm not sure. We've always called this button the "my kiva" button even before the recent MyKiva experience. We have similar naming with the My Kiva dropdown that hovering over this button reveals, and the My Kiva sections of the lend menu.

dyersituations: This morning I saw some early mocks for the nav changes coming, and there might be a MyKiva button in the nav. So with that in mind, what are your thoughts?

src/components/WwwFrame/LendMenu/TheLendMenu.vue
emuvente: The only thing that was prefetched in this component was the user id. Before the lend menu is opened, the user id is only used to conditionally render the "My Kiva" sections of the mega menu and the list menu. I've changed those menus to also use the isUserDataLoading flag to show those sections, so that the user id is not required for the server render at all, and this prefetch can be skipped.

dyersituations (approved):

PR #6093, opened 2025-06-18
fix: import svg locked container in component to reduce compiled file…

christian14b:
… size

`useIsMobile` compiled file used to be 76MB.
dist/server/static/useIsMobile-DQcSDJlW.js                               76,229.18 kB

With this change it will be reduced:
image

### DIFF
--- a/src/components/MyKiva/BadgeContainer.vue
+++ b/src/components/MyKiva/BadgeContainer.vue
@@ -3,7 +3,7 @@
 		class="tw-relative tw-inline-block tw-w-auto"
 		:class="{
 			'tw-grayscale': isInProgress,
-			'invisible-badge tw-cursor-pointer': isLocked,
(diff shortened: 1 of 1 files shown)


src/components/MyKiva/BadgeContainer.vue
```diff
@@ -154,10 +154,6 @@ onMounted(() => {
 	@apply tw-h-full tw-w-full;
 }
 
-.invisible-badge :deep(img) {
-	@apply tw-invisible;
```
dyersituations, line 158: Why is this no longer needed?

dyersituations (approved):

src/components/MyKiva/BadgeContainer.vue
dyersituations: So there is no longer an image that needs to be hidden? Was the styling not really doing anything?

PR #6099, opened 2025-06-23
feat: huge lend amount added as feature

roger-in-kiva:
- huge lend amount added as feature
- https://kiva.atlassian.net/browse/MP-1664


src/util/loanUtils.js
```diff
@@ -138,15 +138,15 @@ export function getDropdownPriceArray(
 		? build5DollarsPriceArray(parsedAmountLeft).slice(0, 28)
 		: buildPriceArray(parsedAmountLeft, minAmount).slice(0, 20);
 
-	const showHugeAmount = enableHugeAmount && parsedAmountLeft > 500;
+	const showHugeAmount = isLoggedIn && parsedAmountLeft > 500;
```
dyersituations, line 141: I forgot that we only showed the large amounts when logged in, makes sense to keep that in place.

PR #6102, opened 2025-06-23
fix: index issue with carousel updates

christian14b:
Fixing initial index of carousel without loading more updates

https://github.com/user-attachments/assets/3ef26770-939e-4d15-84c4-b02628c27f07

### DIFF
--- a/src/components/MyKiva/JournalUpdatesCarousel.vue
+++ b/src/components/MyKiva/JournalUpdatesCarousel.vue
@@ -194,7 +194,7 @@ watch(
 		}
 
 		if (prevUpdatesLength > 0 && prevUpdatesLength !== newUpdates.value?.length) {
-			carouselIndex.value = prevUpdatesLength - 1;
+			carouselIndex.value = prevUpdatesLength;
 		}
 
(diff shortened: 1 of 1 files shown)


src/components/MyKiva/JournalUpdatesCarousel.vue
```diff
@@ -187,16 +188,16 @@ const loadMoreUpdates = () => {
...
 			$kvTrackEvent('portfolio', 'view', 'At least one journal update viewed');
 		}
-		if (updates.value.length > 0 && updates.value.length < 3) {
-			carouselIndex.value = 0;
-		}
-		if (updates.value.length > 3) {
-			carouselIndex.value = updates.value.length - 2;
+
+		if (prevUpdatesLength > 0 && prevUpdatesLength !== newUpdates.value?.length) {
+			carouselIndex.value = prevUpdatesLength - 1;
```
dyersituations, line 197: It looks like this should be `carouselIndex.value = prevUpdatesLength` to move to the first new item, correct?

dyersituations: I think maybe moving to the first new item might be clearer.

PR #6105, opened 2025-06-24
fix: remove header gap caused by non-existent grid area

emuvente:
There is an extra-wide gap in the header between the logo and the lend menu button for large screens when a user is logged in with a balance greater than $0. This gap is caused by a grid area that isn't being used but is configured to appear.

Before
Screenshot 2025-06-24 at 4 51 29 PM

After
Screenshot 2025-06-24 at 4 51 55 PM


mcstover (approved):
Makes sense

PR #6108, opened 2025-06-26
fix: mp-1721 / Make Nav Header Sticky

The-Best-Joke:
Adds style changes to make the nav header component sticky throughout the UI repo. This applies to the lend-by-category, MyKiva and BP pages.


src/components/WwwFrame/TheHeader.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<header class="tw-transition-all tw-duration-1000 tw-ease-in-out">
+	<header class="tw-transition-all tw-duration-1000 tw-ease-in-out tw-sticky tw-top-0" style="z-index: 999;">
```
dyersituations, line 2: We actually only want this for the pages with the ATB feature enabled.

dyersituations: That's why it was special-cased before.

roger-in-kiva: We have a class for z index.

PR #6110, opened 2025-06-26
feat: journeys carousel update

roger-in-kiva:
- last transaction considered for my kiva journeys sort
- https://kiva.atlassian.net/browse/MP-1683

### DIFF
--- a/src/components/Contentful/JourneyCardCarousel.vue
+++ b/src/components/Contentful/JourneyCardCarousel.vue
@@ -202,7 +202,7 @@ const orderedSlides = computed(() => {
 		loanJourneys = getJourneysByLoan(transactionLoan);
 	}
 
-	defaultBadges.forEach(async badgeKey => {
+	defaultBadges.forEach(badgeKey => {
 		const achievementContent = badgesData.value.find(achievement => badgeKey === achievement.id);
 
 		if (achievementContent) {
--- a/src/components/Contentful/JourneyCardCarousel.vue
+++ b/src/components/Contentful/JourneyCardCarousel.vue
@@ -96,6 +96,7 @@
 </template>
 
 <script setup>
+import { parseISO, differenceInDays } from 'date-fns';
 import {
 	computed,
 	ref,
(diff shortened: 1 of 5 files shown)


src/graphql/query/myKiva.graphql
```diff
@@ -79,5 +79,38 @@ query myKivaQuery {
       id
       preferences
     }
+    transactions(
```
dyersituations, line 82: Is this different data from the transaction updates?

src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -177,9 +189,22 @@ const isNonBadgeSlide = slide => {
...
+		return t.type !== AVOID_TRANSACTION_LOANS_KEY && diffInDays <= TRANSACTION_DAYS_LIMIT;
+	});
+
+	if (transactionLoans?.length) {
+		const transactionLoan = transactionLoans?.[0]?.loan ?? {};
+		loanJourneys = getJourneysByLoan(transactionLoan);
+	}
 
-	defaultBadges.forEach(badgeKey => {
+	defaultBadges.forEach(async badgeKey => {
```
dyersituations, line 205: Looks like `async` isn't needed?

src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -177,9 +189,22 @@ const isNonBadgeSlide = slide => {
 
 const orderedSlides = computed(() => {
 	const achievementSlides = [];
+	let loanJourneys = [];
+	let sortedSlides = [];
+
+	const transactionLoans = props.userInfo?.transactions?.values?.filter(t => {
```
dyersituations, line 195: Is `transactions` returning all transactions for the user? What happens if the users has 1000s of transactions?

src/graphql/query/myKiva.graphql
roger-in-kiva: Probably related, but better approach as we are already prefetching user data, so including this is simplest solution.

roger-in-kiva:
> When the screen is full-width, we can now make sure the 3 cards take full-width:
> 
> image

I didn't to keep same gap in all sections we have on this page, but we can increase that gap so. agree?

dyersituations:
> I didn't to keep same gap in all sections we have on this page, but we can increase that gap so. agree?

Consistent gap probably makes sense, but the issue here is more that there shouldn't be gap/padding on the right when the page is full width (having peak in mobile is still important).

roger-in-kiva:
> > I didn't to keep same gap in all sections we have on this page, but we can increase that gap so. agree?
> 
> Consistent gap probably makes sense, but the issue here is more that there shouldn't be gap/padding on the right when the page is full width (having peak in mobile is still important).

Fixed!
Screenshot 2025-06-27 at 10 59 54 a m

PR #6119, opened 2025-07-02
feat: compact borrower status card mp 1652

christian14b:
https://kiva.atlassian.net/browse/MP-1652

- Adding borrower sheet and steps only when user is in `/portfolio` or `/mykiva`
- Shrinking borrower card 
- Add options to card
- ~~We probably want more improvements on next steps component. I didn't want to give it more time since it's not the relevant part of this changeset, but happy to work on it in a different ticket!~~)

image

image

### DIFF
--- a/src/components/BorrowerProfile/BorrowerSideSheetContent.vue
+++ b/src/components/BorrowerProfile/BorrowerSideSheetContent.vue
@@ -54,6 +54,7 @@
 				:kv-track-category="'borrower-profile'"
 				:external-links="true"
 				:unreserved-amount="unreservedAmount"
+				:is-visitor="isGuest"
 				@add-to-basket="addToBasket"
 			/>
 		</div>
@@ -143,6 +144,7 @@ export default {
 		const teamCount = computed(() => borrowerProfile?.teamCount?.value);
 		const lenderCount = computed(() => borrowerProfile?.lenderCount?.value);
 		const loading = computed(() => borrowerProfile?.loading?.value);
+		const isGuest = computed(() => borrowerProfile?.isGuest?.value);
 		const fundraisingPercent = computed(() => {
 			if (borrowerProfile.unreservedAmount.value === '0') return '0';
 			return borrowerProfile.fundraisingPercent.value ?? undefined;
@@ -203,6 +205,7 @@ export default {
 			currentRoute,
 			fundraisingPercent,
 			inPfp,
+			isGuest,
 			lenderCount,
 			loading,
 			loan,
--- a/src/components/LoanFinding/LendingCategorySection.vue
+++ b/src/components/LoanFinding/LendingCategorySection.vue
@@ -40,6 +40,7 @@
 						@add-to-basket="addToBasket"
 						@show-cart-modal="showCartModal"
 						@show-loan-details="showLoanDetails"
+						@mouseenter="$emit('mouse-enter-loan-card', loan?.id)"
 					/>
 				</template>
 				<template v-if="showViewMoreCard" #[`slide${loans.length}`]>
@@ -66,7 +67,7 @@ export default {
 		KvClassicLoanCardContainer,
 		ViewMoreCard,
 	},
-	emits: ['add-to-basket', 'show-loan-details'],
+	emits: ['add-to-basket', 'show-loan-details', 'mouse-enter-loan-card'],
 	props: {
 		title: {
 			type: String,
--- a/src/components/LoanFinding/PartnerSpotlightSection.vue
+++ b/src/components/LoanFinding/PartnerSpotlightSection.vue
@@ -24,6 +24,7 @@
 			@add-to-basket="$emit('add-to-basket', $event)"
 			@show-cart-modal="showCartModal"
 			@show-loan-details="showLoanDetails"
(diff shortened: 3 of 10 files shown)


src/components/MyKiva/BorrowerCarousel.vue
```diff
@@ -298,6 +319,93 @@ const loadMore = () => {
...
+	$kvTrackEvent('borrower-profile', 'go-to-old-bp', undefined, `${this.selectedLoan?.id}`);
+	window.open(`lend/${this.selectedLoan?.id}`);
+};
+
+const handleCloseSideSheet = () => {
+	showSideSheet.value = false;
+	selectedLoan.value = undefined;
+};
+
+const addToBasket = lendAmount => {
```
dyersituations, line 332: This code is duplicated from LBC, right? Would it make sense to move the `KvSideSheet` and `BorrowerSideSheetContent` into a shared component, and then move this util method there as well?

dyersituations:
@christian14b what improvements to next step? The components looks pretty good in the side sheet locally.

src/components/MyKiva/BorrowerCarousel.vue
```diff
@@ -58,7 +58,7 @@
 					ref="carousel"
 					class="borrower-carousel tw-w-full tw-overflow-visible"
 					:multiple-slides-visible="true"
-					:slide-max-width="singleSlideWidth"
+					:slide-max-width="'336px'"
```
dyersituations, line 61: This probably isn't new to this changeset, but I'm seeing this error currently locally when clicking the carousel arrows. Maybe it's a quick fix?

```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'tabContext')
    at onInteractCarousel (BorrowerCarousel.vue:251:13)
    at callWithErrorHandling (chunk-TJ4MTCWI.js?v=f1e4d7fa:2487:19)
    at callWithAsyncErrorHandling (chunk-TJ4MTCWI.js?v=f1e4d7fa:2494:17)
    at emit (chunk-TJ4MTCWI.js?v=f1e4d7fa:7657:5)
    at chunk-TJ4MTCWI.js?v=f1e4d7fa:9185:45
    at chunk-7HKXJEJN.js?v=f1e4d7fa:2457:9
```

src/components/MyKiva/BorrowerStatusCard.vue
```diff
@@ -158,31 +117,24 @@ import {
...
 
 const COMMENT_ID = 'comment';
-const DETAILS_ID = 'details';
 const SHARE_ID = 'share';
 
-const router = useRouter();
 const $kvTrackEvent = inject('$kvTrackEvent');
 
-const emit = defineEmits(['toggle-what-is-next', 'open-comment-modal', 'open-share-modal']);
+const emit = defineEmits(['open-comment-modal', 'open-share-modal', 'open-side-sheet']);
```
dyersituations, line 124: Adding a comment is navigating the user to the BP, can we instead open the BP side sheet for mykiva and portfolio carousels? Let me know if this should be a follow-up ticket.

dyersituations: I was referring to the app redirecting me to the BP after a comment was successfully added. Could we change the behavior to instead open the side sheet after a comment is successfully added from the borrower carousel?

src/pages/Portfolio/ImpactDashboard/LoanCommentModal.vue
```diff
@@ -101,7 +101,6 @@ export default {
 					this.closeModal();
 					this.$kvTrackEvent('portfolio', 'click', 'Leave a loan comment', this.loan.borrowerName, this.loan.loanId); // eslint-disable-line max-len
 					this.$showTipMsg(`Thank you for helping ${this.loan.borrowerName}!`, 'confirmation', true);
-					this.$router.push(`/lend/${this.loan.loanId}`);
```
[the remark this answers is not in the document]
dyersituations: We do show them in the side sheet with the comment component, but I wasn't seeing a comment being added.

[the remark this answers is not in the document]
dyersituations: Sounds good, please add a new ticket to the next sprint and same epic with the details you found. For now here, we probably should only show the add comment option if the loan is fundraising. Could you make that change in this PR?

src/components/MyKiva/BorrowerStatusCard.vue
```diff
@@ -213,6 +150,23 @@ const title = computed(() => `${borrowerName.value} in ${borrowerCountry.value}`
...
 const isFundraising = computed(() => loan.value?.status === FUNDRAISING);
+const isFunded = computed(() => loan.value?.status === FUNDED);
+
+const menuOptions = computed(() => {
+	const options = [{
+		id: SHARE_ID,
+		label: 'Share',
+	}];
+
+	if (!isFunded.value) {
```
dyersituations, line 161: Would it be more accurate to instead remove the option if the loan isn't fundraising? I don't know off hand which statuses don't allow comments, but we know it works for fundraising.

dyersituations (approved):

PR #6121, opened 2025-07-03
fix: add BP emitters and listeners

The-Best-Joke:


src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default {
+	methods: {
+		loadBPData(loanId) {
+			this.apollo.query({
+				query: borrowerProfileSideSheetQuery,
+				variables: {
+					loanId
+				}
+			});
```
dyersituations, line 12: Probably be good to add a .catch just in case:

```
		.catch(e => {
			logReadQueryError(e, 'borrowerProfileSideSheetQuery');
		});
```

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -0,0 +1,22 @@
...
+		},
+
+		getCustomHref(loanId) {
```
dyersituations, line 15: Not used, can it be removed?

dyersituations:
@The-Best-Joke changes look good locally, the BP data was preloaded when the side sheet opened, like expected!

PR #6123, opened 2025-07-04
fix: mp-1717 / Add BP SideSheet to MyKiva Recommended Loans Carousel

The-Best-Joke:


src/components/LoanFinding/LendingCategorySection.vue
```diff
@@ -40,6 +40,7 @@
 						@add-to-basket="addToBasket"
 						@show-cart-modal="showCartModal"
 						@show-loan-details="showLoanDetails"
+						@mouseenter="$emit('mouse-enter-loan-card', loan?.id)"
```
dyersituations, line 43: Looks like some changes from the other PR. Once this PR gets updated with that merged PR, I'll take another look.

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -99,344 +97,372 @@
...
+				:is-adding="isAdding"
+				:basket-items="basketItems"
+				@add-to-basket="addToBasket"
+			/>
+		</KvSideSheet>
 	</MyKivaContainer>
 </template>
 
-<script setup>
+<script>
```
dyersituations, line 130: Why the conversion from composition to options API?

roger-in-kiva: Looks like to use the mixin... but this should be added at page level and then passed as a prop, my guess.

dyersituations: We don't want to pass methods as props, so this is a bit more complicated.

The-Best-Joke: @dyersituations Roger is correct: This refactorization to the Options API was in order to leverage the code as a more maintainable mixin given the current structure of the project elsewhere.

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -32,6 +43,103 @@ export default {
...
+						}
+					});
+				} else {
+					try {
+						// track facebook add to basket
+						if (typeof window !== 'undefined' && typeof fbq === 'function') {
+							window.fbq('track', 'AddToCart', { content_category: 'Loan' });
+						}
+					} catch (e) {
+						console.error(e);
```
The-Best-Joke, line 121: I'm thinking a `logFormatter` would go better here?

The-Best-Joke (commented):
To review

src/plugins/borrower-profile-exp-mixin.js
dyersituations: Yup, we can use `logFormatter` for any logging.

src/pages/MyKiva/MyKivaPageContent.vue
dyersituations: Ideally we use composition API moving forward, but honestly we'll need to refactor no matter what when/if we move this page to CPS, so I'm not overly concerned about this current refactor.

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -99,344 +97,372 @@
...
 				:slides="moreWaysToHelpSlides"
 				:lender="lender"
 				:user-in-homepage="userInHomepage"
 				:hero-contentful-data="heroContentfulData"
 				:hero-tiered-achievements="heroTieredAchievements"
 				@update-journey="updateJourney"
-				class="tw-mt-2"
 			/>
 		</section>
+		<KvSideSheet
```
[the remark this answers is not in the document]
dyersituations: Actually seeing on LBC locally as well.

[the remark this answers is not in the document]
The-Best-Joke: I'm still seeing the warning even after updating to the latest version of the package. Gonna have to go back and take a look at my previous fix of `KvSideSheet`.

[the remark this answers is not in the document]
The-Best-Joke: Nevermind, that did the trick!

dyersituations (approved):

The-Best-Joke:
@christian14b When you can, please give this PR a quick review and if possible testing round to make sure I didn't break any of your recently introduced functionality in the refactorization of the component to Options API.

src/components/BorrowerProfile/BorrowerSideSheetWrapper.vue
```diff
@@ -10,156 +10,74 @@
...
 		type: Boolean,
 		default: false
 	},
+	widthDimensions: {
+		type: Object,
+		default: () => {}
+	}
 });
 
-const $kvTrackEvent = inject('$kvTrackEvent');
```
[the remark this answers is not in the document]
roger-in-kiva: I mentioned about initializing the experiment b/c we are tracking the page where is loaded and we have two cases, `my kiva` and `LBC`.

One options to don't migrate the whole component and also use the mixin we already have, is similar `vue-demi` case we have on shared components, so we can try this and check if works fine in our stack.

Here an example of the code.

```
<script>
import { ref } from 'vue'

export default {
  data() {
    return {
      otherMsg: this.changeMsg()  // Accessing a method from setup()
    }
  },
  setup() {
    const msg = ref('hello')
    const changeMsg = () => {
      return 'again'  // Define a method in setup()
    }
    return { msg, changeMsg } // Expose the reactive variable and the method
  },
  methods: {
    click() {
      this.msg = 'ok'  // Accessing a reactive variable from setup()
    },
  },
}
</script>
```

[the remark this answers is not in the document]
roger-in-kiva: Other option is to add `mixin` to `mykiva` page, and use `prop-emit` approach to pass and request data.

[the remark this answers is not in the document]
The-Best-Joke: @roger-in-kiva The `MyKivaPageContent` in this PR's current state already includes the `borrowingProfileExpMixin`.

@dyersituations that's one solution, although it would re-introduce the redundant existence from the `addToBasket` `goToLink` and `initializeBasketItems` functions in both the mixin and the wrapper.

Thing is… I don't want to get ahead of myself or anything but I think I've just about finished the refactorization? I've been testing the introduced functionality by @christian14b with MP-1652 in the `MyKivaPageContent` as well as the changes to the sidesheet's next steps bit… And everything seems to be in order. No Vue warnings, no major bugs.

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -32,6 +44,103 @@ export default {
...
+					},
+					fetchPolicy: 'network-only'
+				});
+				this.basketItems = data?.shop?.basket?.items?.values || [];
+			} catch (error) {
+				logFormatter(error, 'error');
+				this.basketItems = [];
+			}
+		},
+		addToBasket(lendAmount) {
```
dyersituations, line 73: Not sure why `addToBasket` is triggering `PostCheckoutAchievements`, but that query locally is being called after ATB and is returning the error below. It looks like this query isn't being triggered on dev after, so not sure specifically the source of this error, but it is a difference from dev.

```[
  {
    "message": "invalid type for variable: 'loanIds'",
    "extensions": {
      "name": "loanIds",
      "code": "VALIDATION_INVALID_TYPE_VARIABLE"
    }
  }
]
```

src/components/LoanCards/KvClassicLoanCardContainer.vue
```diff
@@ -288,7 +288,12 @@ export default {
 					}, 1000);
 				}
 				this.isAdding = false;
-				this.$emit('add-to-basket', { loanId: this.loanId, name: this.loan?.name, success: true });
+				this.$emit('add-to-basket', {
+					loanId: this.loanId,
+					lendAmount,
```
dyersituations, line 293: Why was this change needed? We haven't changed the ATB for loan cards for a while.

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -32,6 +44,103 @@ export default {
...
+					},
+					fetchPolicy: 'network-only'
+				});
+				this.basketItems = data?.shop?.basket?.items?.values || [];
+			} catch (error) {
+				logFormatter(error, 'error');
+				this.basketItems = [];
+			}
+		},
+		addToBasket({ loanId, lendAmount }) {
```
dyersituations, line 73: In general it looks like there's currently an issue with the ATB flow, so we'll need to dig into that before merging. Locally I'm seeing a network error whenever adding to basket from a loan card for either LBC or mykiva. For the `main` branch, I'm seeing the error on LBC but not mykiva, so mykiva must now be getting it in this branch with the merged work.

Specifically it's an error from `PostCheckoutAchievements` where it's getting a null `loanId`. But I think that's a result of an issue that occurs before. Locally I'm seeing the `updateLoanReservation` call twice. That shouldn't be happening. So please dig into the ATB flow and track down the core issue.

The-Best-Joke: I *think* I got it. Whenever the direct `lendCTA` buttons are triggered they themselves handle the `updateLoanReservation` query, the same that the `borrower-profile-exp-mixin` does, which is where the redundant calls were coming from.

Decoupling the borrower profile `addToBasket` from the `lendCta` ATB does the trick. 

The `PostCheckoutAchievements` query comes from the `loadBP` function that we're using on hover of each card to pre-load the card's information. I don't think that query is necessary for the pre-load function but I could be wrong as it does load (I think) some badge/achievement data?

dyersituations: I'm seeing the error resolved for the loan cards now on both LBC and MyKiva. But I'm still seeing the error for `PostCheckoutAchievements` when clicking the lend CTA in the side page. It's still showing a `null` loan ID, for example:
image

dyersituations: Testing on dev for LBC, I don't see this error.

dyersituations:
Linting errors now for other files where `src/plugins/basket-modal-mixin` is expected.

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -29,9 +54,108 @@ export default {
...
+						if (typeof window !== 'undefined' && typeof fbq === 'function') {
+							window.fbq('track', 'AddToCart', { content_category: 'Loan' });
+						}
+					} catch (e) {
+						logFormatter(e, 'error');
+					}
+					const basketId = this.cookieStore.get('kvbskt');
+					setTimeout(() => {
+						this.formatAddedLoan();
+					}, 1000);
```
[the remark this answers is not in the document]
The-Best-Joke: Took it straight from the `lendCta` button for the same operation. Since that works I figured best not to tweak it.

src/plugins/borrower-profile-exp-mixin.js
The-Best-Joke: Update: I was wrong about the `basketModalMixin`. Whoops!

Gonna revert that bit and accommodate the code to account for the decoupling of the mixins. My bad.

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -29,9 +53,107 @@ export default {
...
+					try {
+						// track facebook add to basket
+						if (typeof window !== 'undefined' && typeof fbq === 'function') {
+							window.fbq('track', 'AddToCart', { content_category: 'Loan' });
+						}
+					} catch (e) {
+						logFormatter(e, 'error');
+					}
+					const basketId = this.cookieStore.get('kvbskt');
+					setTimeout(() => {
```
dyersituations, line 136: The other usages have a comment above, `// Show modal after 1s (Defined in CSS)`, so we know later why the timeout is needed.

src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -210,6 +210,7 @@ const fetchAchievementFromBasket = async () => {
 };
 
 watch(addedLoan, async () => {
+	if (!addedLoan?.value?.id) return;
 	if (myKivaExperimentEnabled.value && !isGuest.value) {
 		await fetchBasketData();
 		fetchPostCheckoutAchievements([...loansIdsInBasket.value, addedLoan.value?.id]);
```
dyersituations, line 216: Here we are still adding duplicate loan IDs. Because there is an `await` on `fetchBasketData`, I think the line here needs to be just `fetchPostCheckoutAchievements(loansIdsInBasket.value);`.

PR #6124, opened 2025-07-07
fix: add tracking to nav search bar

dyersituations:
https://kiva.atlassian.net/browse/MP-1719

### DIFF
--- a/src/components/WwwFrame/SearchBar.vue
+++ b/src/components/WwwFrame/SearchBar.vue
@@ -188,6 +188,7 @@ export default {
 					]);
 				}
 			});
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/SearchBar.vue
```diff
@@ -188,6 +188,7 @@ export default {
 					]);
 				}
 			});
+			this.$kvTrackEvent('event-tracking', 'click', 'Nav-search-bar');
```
roger-in-kiva, line 191: We already have a `search` category, applies?

PR #6125, opened 2025-07-07
feat: mp-1744 / add huge lending amounts to LBC page

The-Best-Joke:
Adds the `isGuest` prop to the `LoanFinding` component and the `my { id }` graphql request to the `borrowerProfileSidesheet.graphql` file in order to toggle the huge amounts lending feature for non-guest users of the site.

### DIFF
--- a/src/composables/useBorrowerProfileData.js
+++ b/src/composables/useBorrowerProfileData.js
@@ -23,7 +23,7 @@ export default function useBorrowerProfileData(apolloClient, cookieStore) {
 	// Loading state and computed properties
 	const loading = computed(() => !bpData.value);
 
-	const isGuest = computed(() => bpData.value?.my);
+	const isGuest = computed(() => !bpData.value?.my);
 	const loan = computed(() => bpData.value?.lend?.loan);
 	const loanId = computed(() => loan.value?.id ?? 0);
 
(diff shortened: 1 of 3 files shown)


src/composables/useBorrowerProfileData.js
```diff
@@ -23,30 +23,32 @@ export default function useBorrowerProfileData(apolloClient, cookieStore) {
...
 	const loading = computed(() => !bpData.value);
 
+	const isGuest = computed(() => bpData.value?.my);
 	const loan = computed(() => bpData.value?.lend?.loan);
 	const loanId = computed(() => loan.value?.id ?? 0);
-	const userBalance = computed(() => bpData.value?.my?.userAccount?.balance);
-	const name = computed(() => loan.value?.name ?? '');
-	const hash = computed(() => loan.value?.image?.hash ?? '');
+
+	const anonymizationLevel = computed(() => loan.value?.anonymizationLevel ?? '');
```
dyersituations, line 30: The result of this auto sort is... strange. Why did it sort some correctly but not others? The newlines in between?

The-Best-Joke: Honestly kind of dumb of me since it only adds to the overhead of the review unnecessarily. I sorted them by the existing blocks of code but without merging them all into a single block.

src/composables/useBorrowerProfileData.js
```diff
@@ -23,30 +23,32 @@ export default function useBorrowerProfileData(apolloClient, cookieStore) {
 	// Loading state and computed properties
 	const loading = computed(() => !bpData.value);
 
+	const isGuest = computed(() => bpData.value?.my);
```
dyersituations, line 26: This needs to be `const isGuest = computed(() => !bpData.value?.my);`

PR #6129, opened 2025-07-10
feat: sunsetting /get-started page

tiarascoleman:


src/pages/GetStarted/GetStartedResults.vue
```diff
@@ -1,464 +0,0 @@
...
-		RecommendedLoanCard,
-	},
-	data() {
-		return {
-			loanIds: [],
-			totalPreferredCount: 0,
-			totalFundRaisingCount: 0,
-			howItWorksImgs: {
-				loan: [
-					['small', imgRequire('how-it-works-loan.png')],
```
roger-in-kiva, line 144: Looks like these images are no longer needed, remove?

dyersituations: @tiarascoleman yeah any import in these files that aren't used elsewhere can be removed.

src/components/GetStarted/EditPreferences.vue
```diff
@@ -6,138 +6,6 @@
 		<p class="tw-text-subhead tw-mb-4">
 			Change your preferences and find more borrowers to support.
 		</p>
-		<div class="row">
```
dyersituations, line 9: It actually looks like this component can be removed entirely? Definitely look into what was being used by the route that is being removed, and remove any components from there that can be removed that aren't used elsewhere.

src/router/routes.js
```diff
@@ -115,33 +115,6 @@ export default [
 		path: '/funded/:id',
 		redirect: '/lend/:id'
 	},
-	{
-		path: '/get-started',
```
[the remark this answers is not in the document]
dyersituations: Yeah homepage might be the best now that I think about it again.

src/router/routes.js
```diff
@@ -117,30 +117,7 @@ export default [
...
-			},
-			{
-				path: 'results',
-				component: () => import('#src/pages/GetStarted/GetStartedResults'),
-				meta: {
-					excludeFromStaticSitemap: true,
-				}
-			},
-		]
+		redirect: '/homepage-classic',
```
dyersituations, line 120: Technically this will work due to another redirect, but we can simplify to `redirect: '/',`

PR #6130, opened 2025-07-10
fix: apply style fixes

The-Best-Joke:


src/components/BorrowerProfile/BorrowerSideSheetContent.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tw-bg-eco-green-1 tw--mx-2">
+	<div class="tw--mx-2" style="background-color: rgb(245 245 245)">
```
mcstover, line 2: Do we not have a `tw-` class for [this under the "static" tab](https://main--608b4cf87f686c00213841b1.chromatic.com/?path=/story/base-styling-primitives--primitives)?

If not, we should push back and get this assigned a color class.

dyersituations: That's a good call out, I forgot this section had a custom bg color.

@The-Best-Joke let's use `tw-bg-secondary`

dyersituations:
Per ticket, loan details should match version in CPS/combo page:

Expected:
image

Current:
image

dyersituations (changes requested):
See comments, thanks!

dyersituations:
@The-Best-Joke also pushed up a small commit of a fix that I needed locally, unrelated to your changes. Not sure why this was needed all of a sudden, maybe my browser was in a weird state, but it would be good to get in, and I was already on your branch.

PR #6131, opened 2025-07-14
feat: add lazy loading and portfolio navigation to JournalUpdatesCaro…

infante-jaime:
> git commit -am "feat: add lazy loading and portfolio navigation to JournalUpdatesCarousel

- Implemented lazy loading for updates section using a wrapper component.
- Updated carousel to show 'Load more' until the end, then display 'View portfolio' button as a router-link.
- Ensured layout remains consistent and improved navigation experience for users."
⧗   input: feat: add lazy loading and portfolio navigation to JournalUpdatesCarousel

- Implemented lazy loading for updates section using a wrapper component.
- Updated carousel to show 'Load more' until the end, then display 'View portfolio' button as a router-link.
- Ensured layout remains consistent and improved navigation experience for users.

### DIFF
--- a/src/graphql/query/myKiva.graphql
+++ b/src/graphql/query/myKiva.graphql
@@ -35,7 +35,7 @@ query myKivaQuery {
       filter: {
         category: loan
       }
-      limit: 100
+      limit: 15
     ) {
       values {
         category
--- a/src/pages/MyKiva/AsyncMyKivaSection.vue
+++ b/src/pages/MyKiva/AsyncMyKivaSection.vue
@@ -1,0 +1,18 @@
+<template>
+	<div>
+		<slot></slot>
+	</div>
+</template>
+
+<script>
+import delayUntilVisibleMixin from '#src/plugins/delay-until-visible-mixin';
+
+export default {
+	name: 'AsyncMyKivaSection',
+	mixins: [delayUntilVisibleMixin],
+	emits: ['visible'],
+	mounted() {
+		this.delayUntilVisible(entry => this.$emit('visible', entry));
+	},
+};
+</script>
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -59,13 +59,16 @@
 					:total-loans="totalLoans"
 					show-menu
 				/>
-				<JournalUpdatesCarousel
-					v-if="!updatesLoading && mergedUpdates.length"
-					:updates="mergedUpdates"
-					:lender="lender"
-					:total-updates="totalUpdates"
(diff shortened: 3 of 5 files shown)


src/graphql/query/myKiva.graphql
```diff
@@ -35,7 +35,7 @@ query myKivaQuery {
       filter: {
         category: loan
       }
-      limit: 100
+      limit: 15
```
roger-in-kiva, line 38: As transactions query is also returning `loan_repayment` cases, are we guaranteed that we'll show at least 9 active loans in the carousel?

dyersituations: The problem is that it sounds like the transactions query is fairly expensive for the monolith. So we can try 15 here and see how it works for a few users. Sound good?

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -54,13 +54,16 @@
...
-				<JournalUpdatesCarousel
-					v-if="!updatesLoading && mergedUpdates.length"
-					:updates="mergedUpdates"
-					:lender="lender"
-					:total-updates="totalUpdates"
-					@load-more-updates="loadMoreUpdates"
-				/>
+				<async-my-kiva-section @visible="fetchInitialUpdates">
+					<JournalUpdatesCarousel
+						v-if="!updatesLoading && visibleUpdates.length"
```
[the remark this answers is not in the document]
dyersituations: Yeah it looks like a change is missing for defining `visibleUpdates`.

src/graphql/query/myKiva.graphql
roger-in-kiva: Probably makes sense, what about default limit of 20?
Issue I see is that an user who is not lending but receiving repayments, this query could start hiding loans in the carousel at some point.
Also, should be useful to know what is the default value for `endDate` filter.
Example: old loans query limit was 50; for Shua's fake auth case, having a limit of 100 here, we were able to get 36 loans after filtering. As showing 9 is not really needed, but useful to ensure we are able to show 9 at least.

src/pages/MyKiva/AsyncMyKivaSection.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+	<div>
+		<slot></slot>
+	</div>
+</template>
+
+<script>
+import delayUntilVisibleMixin from '#src/plugins/delay-until-visible-mixin';
+
+export default {
```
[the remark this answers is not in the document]
infante-jaime: This appears to have lots of issues with the merge to main so please stop the review for now and I'll create a new PR with a clean main and cherry pick from there closing this one fr now and I'll add the new limit

PR #6139, opened 2025-07-15
feat: add header query to combo page redirect

christian14b:
https://kiva.atlassian.net/browse/MP-1640

### DIFF
--- a/src/components/MyKiva/JournalUpdateCard.vue
+++ b/src/components/MyKiva/JournalUpdateCard.vue
@@ -199,7 +199,8 @@ onMounted(() => {
 
 const useFunds = () => {
 	$kvTrackEvent('portfolio', 'click', 'repayment-update-read-more', update.value.id);
-	window.location.href = `/lend/filter?header=You-can-impact-${update.value.livesImpacted}-lives-right-now`;
+	// eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/JournalUpdateCard.vue
```diff
@@ -199,7 +199,7 @@ onMounted(() => {
 
 const useFunds = () => {
 	$kvTrackEvent('portfolio', 'click', 'repayment-update-read-more', update.value.id);
-	window.location.href = '/lend/filter';
+	window.location.href = `/lend/filter?header=You-can-impact-${update.value.livesImpacted}-lives-right-now`;
```
dyersituations, line 202: Per ticket, this value actually needs to be `Where X = the user’s current balance divided by $25 + 1`

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -419,6 +419,8 @@ export default {
 		},
 		formatRepaymentCards(repayments) {
 			if (repayments.length === 0) return [];
+			const livesImpacted = Math.floor(this.userBalance / 25) + 1;
```
dyersituations, line 422: It might be good to change this variable to `livesToImpact`. Using `livesImpacted` is really similar to the text on the page for the borrower carousel.

PR #6142, opened 2025-07-16
feat: fix href to match all the card and open a new window, fix the t…

infante-jaime:
feat(blog-carousel): make entire card clickable, open in new tab, and improve tracking

- Updated BlogCard so the entire card is clickable and opens the blog post in a new browser tab, improving user experience and accessibility.
- Refactored tracking to emit only the blog slug, reducing coupling between BlogCard and LatestBlogCarousel.
- Ensured tracking events are fired consistently when a card is clicked.

### DIFF
--- a/src/components/MyKiva/BlogCard.vue
+++ b/src/components/MyKiva/BlogCard.vue
@@ -1,6 +1,10 @@
 <template>
 	<div
-		class="blog-card tw-w-full tw-relative tw-rounded tw-bg-cover tw-bg-center tw-bg-white"
+		class="blog-card tw-w-full tw-relative tw-rounded tw-bg-cover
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/BlogCard.vue
```diff
@@ -1,6 +1,10 @@
 <template>
 	<div
-		class="blog-card tw-w-full tw-relative tw-rounded tw-bg-cover tw-bg-center tw-bg-white"
+		class="blog-card tw-w-full tw-relative tw-rounded tw-bg-cover
```
dyersituations, line 3: On card hover, there be a cursor pointer and the title should underline.

dyersituations (approved):

PR #6143, opened 2025-07-16
feat: update datasource and add lazy loading to the updates carousel

infante-jaime:
added lazy loading plus a change to the data source to have more controlled results

### DIFF
--- a/src/components/MyKiva/JournalUpdatesCarousel.vue
+++ b/src/components/MyKiva/JournalUpdatesCarousel.vue
@@ -21,7 +21,7 @@
 					@share-loan-clicked="shareLoanClicked"
 				/>
 			</template>
-			<template v-if="showLoadMore || atEndOfCarousel" #view-more>
+			<template v-if="!updatesLoading" #view-more>
 				<div
 					:key="`view-more-card`"
 					class="tw-flex tw-items-center tw-h-full tw-pl-4"
@@ -114,7 +114,7 @@ const props = defineProps({
 		type: Number,
 		default: 0,
 	},
-	atEndOfCarousel: {
+	updatesLoading: {
 		type: Boolean,
 		default: false,
 	},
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/JournalUpdatesCarousel.vue
```diff
@@ -21,18 +21,28 @@
 					@share-loan-clicked="shareLoanClicked"
 				/>
 			</template>
-			<template v-if="showLoadMore" #view-more>
+			<template v-if="showLoadMore || atEndOfCarousel" #view-more>
```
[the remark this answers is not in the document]
infante-jaime: You're correct: I initially tried to use only `showLoadMore`, but the button rendered before the updates finished loading. By passing the existing `updatesLoading` variable as a prop, we can ensure the button only appears after the updates are fully loaded, making the UI more reliable and user-friendly.

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -56,13 +56,16 @@
...
-			<JournalUpdatesCarousel
-				v-if="!updatesLoading && mergedUpdates.length"
-				:updates="mergedUpdates"
-				:lender="lender"
-				:total-updates="totalUpdates"
-				@load-more-updates="loadMoreUpdates"
-			/>
+			<AsyncMyKivaSection @visible="fetchInitialUpdates">
+				<JournalUpdatesCarousel
+					v-if="!updatesLoading && visibleUpdates.length"
```
dyersituations, line 61: It looks like this component is already hidden if `updatesLoading`, do we also need to pass in the prop?

infante-jaime: for some weird reason if I don't it renders the show more as the first item of the carousel

PR #6150, opened 2025-07-17
fix: mp-1753 / Refactorization of BorrowerCarousel & LoanCards component to leverage BP mixin and SideSheet properly

The-Best-Joke:
This PR aims to refactor both the `BorrowerCarousel` and `LoanCards` component in order to have each properly leverage the props of the `borrower-profile-exp-mixin` which is more or less intrinsic to the proper functioning of the `KvSideSheet` component as a drawer for a Borrower Profile-esque page.

The original issue of this ticket sought to fix the bug where a brand new comment would not appear when the SideSheet appeared after the comment was made in either the MyKiva or Portfolio pages. This was simply a GraphQL caching issue, it's been fixed.

However, closer inspection revealed that the Borrower Profile SideSheet component as previously leveraged had many bugs: the `goToLink` call to the old-school borrower profile did not work, nor did the basket items mechanism or any of the lending. This is because we weren't using any of the props from the `borrower-profile-exp-mixin` in the `BorrowerSideSheetWrapper` component leveraged by `BorrowerCarousel` component.


src/components/MyKiva/BorrowerCarousel.vue
```diff
@@ -92,7 +92,8 @@
 		</div>
 		<!-- Loan Comment Component -->
 		<LoanCommentModal
-			:loan="commentLoanData"
+			:loan="selectedLoan"
```
dyersituations, line 95: Part of this bug was also that the comments couldn't be added for loans in statuses other than fundraising. We are currently hiding the comment from the borrower card menu if the loan isn't fundraising. With these changes if you show the comment menu item again for funded loans, does adding comments work like expected? If you need an account with funded loans in the carousel, you can use mine: `497437`.

Option missing here:
image

dyersituations: @The-Best-Joke if there are issues with saving comments to non-fundraising loans, that can also be a follow-up ticket/PR.

The-Best-Joke: @dyersituations confirmed that there is still an issue, very likely from the backend, when saving comments for loans not in fundraising status. This is despite changing the values in the frontend to allow for the comment modal to pop back up with funded loans.

dyersituations:
> Is old BP page working as expected?

So it would be the monolith page, and I just checked that it does indeed work on dev for a funded loan. Maybe an issue with the graphql version of adding a comment?

dyersituations:
@The-Best-Joke how about we move forward like this:

1. rollback change in this PR to show adding a loan comment in the menu
2. investigate this one issue I encountered locally where the ATB modal showed (2) incorrectly -> couldn't repro in dev LBC page, which is also in the same project
image
3. finish up this PR
4. keep ticket open and investigate different with monolith `lend-classic` page that allows comments on non-fundraising loans

Sounds good?

dyersituations:
> Probably, we also have an option to add a comment on TY page, works fine?

@roger-in-kiva I don't know if we've ever tested that we a loan that was just completed, since most cases the loan will be fundraising, but good call. @The-Best-Joke I'll add that use case to check to the ticket.

dyersituations:
> @dyersituations I'll rollback to the minimum required code changes necessary to fix the caching/rendering issue with newly added comments.
> 
> Question: Do we wanna keep the other BP Sidesheet improvements present such as having a functioning `goToLink` and `basketItems` and checkout? It sounds like an obvious yes but it does add a bit of the overall code changes in the repo so I dunno.

I think you could keep all of the current changes in this PR, since there are lots of good fixes. The screenshot is the only issue I noticed, which is hopefully a quick fix in this branch?

dyersituations (approved):

PR #6151, opened 2025-07-17
feat: badge journey modal update

roger-in-kiva:
- Badge modal update to show entire journey
- Images optimized and replaced
- No more needed function and test removed
- Some refactor in components
- https://kiva.atlassian.net/browse/MP-1828

Screenshot 2025-07-17 at 2 39 11 p m


src/components/MyKiva/BadgeContainer.vue
```diff
@@ -20,17 +20,6 @@
 				:src="solidComponent"
 			>
 		</div>
-		<div
-			v-if="isLocked"
-			class="tw-absolute tw-flex"
-			:class="{ 'animate-wiggle': animateLock }"
```
dyersituations, line 26: Oh this is one thing we miss with using the image with the lock. Could you check with Nathan in the Slack thread if we are OK not having the animated lock? If so, looks like we can remove other code related to `animateLock`.

roger-in-kiva: Removed

PR #6153, opened 2025-07-18
fix: my kiva loans query reverted

roger-in-kiva:
- My Kiva loans query reverted to show them correctly
- https://kiva.atlassian.net/browse/MP-1835

### DIFF
--- a/src/graphql/query/myKiva.graphql
+++ b/src/graphql/query/myKiva.graphql
@@ -4,6 +4,54 @@ query myKivaQuery {
     isBorrower
     loans {
       totalCount
+      values {
+        id
+        name
+        gender
+        status
+        use
+        image {
+          id
+          url
+          hash
+        }
+        loanAmount
+        plannedExpirationDate
+        terms {
+          currency
+          currencyFullName
+          lossLiabilityNonpayment
+          lossLiabilityCurrencyExchange
+          loanAmount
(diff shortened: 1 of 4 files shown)


src/graphql/query/myKiva.graphql
```diff
@@ -35,40 +83,26 @@ query myKivaQuery {
       filter: {
         category: loan
       }
```
dyersituations, line 85: The default limit is 20?

dyersituations: The user still sees their repayments like expected?

roger-in-kiva: Yeah, for both cases, transactions and loans.

roger-in-kiva: Repayments are merged with updates and still working as expected.

dyersituations: I know we've been talking about needing a filter for transactions, but what are the current limitations of this refactor?

roger-in-kiva: For sure!
Probably a temporary if we include new filter in transactions query soon.

dyersituations: Yeah definitely mention the temporary nature, maybe a month-ish?

PR #6156, opened 2025-07-22
feat: mykiva card added

roger-in-kiva:
- MyKiva card added for multiple purposes at mykiva page
- Part of https://kiva.atlassian.net/browse/MP-1751

Screenshot 2025-07-22 at 11 09 34 a m

Screenshot 2025-07-22 at 11 09 41 a m

Screenshot 2025-07-22 at 11 09 48 a m


src/components/MyKiva/MyKivaCard.vue
```diff
@@ -0,0 +1,252 @@
...
+			"
+			:class="{
+				'slide-gradient': hasGradient,
+				'tw-absolute': !imageUrl,
+			}"
+			:style="[
+				{ 'height': overlayHeight },
+			]"
+		>
+			<div class="tw-flex tw-flex-col tw-justify-end tw-h-full !tw-gap-1.5">
```
dyersituations, line 41: It looks like this content needs to be pushed to the bottom. The title and content move down as the image loads. The pop-in only affects the new variation.

dyersituations: Yes, if it happens in the app as well, it would be worth a fix. Just wanted to call it out. Otherwise everything looks good!

roger-in-kiva: Thanks, I'll check it out.
There is more upcoming work, this PR is mainly to create a new component and replace in current carousel we have (journeys).

Current new variation in Storybook is not precisely final version, we need to replace that image in the app, and some extra tweaks are coming...

Just to be sure, can you record a video just to check what probably I don't see?

And good to ask you, should we consider carousel version as a variation of this?

dyersituations: You mean add this new card to a carousel? I don't think we need that work yet.

When it comes to repro, just delete the image from the page, and you'll see the content push up. It's the same general affect of the image not being loaded, since there isn't a default image on the img tag.

dyersituations: I was testing by deleting the element, not the image URL, but I saw the same thing as the image was loading.

Oh good question on the image carousel, I don't remember seeing that previously. Could you ask Nathan and Lauren in the team channel if we're only showing one borrower image for now?

roger-in-kiva: Ok. I'll try to replicate what you mean about it in upcoming work, thanks!

PR #6158, opened 2025-07-22
feat: add animation to region checkmarks

christian14b:
Adding animations to checkmarks

https://github.com/user-attachments/assets/4e316fa5-6132-433d-87a6-24b0382f7ecc

### DIFF
--- a/src/components/MyKiva/RoundCheckbox.vue
+++ b/src/components/MyKiva/RoundCheckbox.vue
@@ -1,4 +1,4 @@
-3<template>
+<template>
 	<div
 		class="tw-block tw-relative"
 		:class="{


src/components/MyKiva/RoundCheckbox.vue
```diff
@@ -26,11 +26,13 @@
 				]"
 				style="aspect-ratio: 1 / 1;"
 			>
-				<KvMaterialIcon
-					v-if="inputValue"
-					:icon="mdiCheckCircle"
-					class="tw-w-3.5 tw-h-3.5 tw-text-brand-550"
-				/>
+				<transition name="fade">
```
dyersituations, line 29: Animations look good! Can we start the animations when the lending stats component enters the viewport? Not important for desktop, but it would probably needed for mobile, right?

src/components/MyKiva/RoundCheckbox.vue
```diff
@@ -1,4 +1,4 @@
-<template>
+3<template>
```
dyersituations, line 1: Remove `3`

src/components/MyKiva/RoundCheckbox.vue
```diff
@@ -68,3 +70,15 @@ watch(() => props.checked, val => {
...
 </script>
+
+<style lang="postcss" scoped>
+
+.fade-enter-active, .fade-leave-active {
+	transition: opacity 0.5s;
+}
+
+.fade-enter-from, .fade-leave-to {
+	opacity: 0;
```
dyersituations, line 81: `@apply tw-opacity-0`

src/components/MyKiva/LendingStats.vue
```diff
@@ -90,4 +98,32 @@ const pillHeader = computed(() => {
 	}
 	return `${loanRegions.value}/${totalRegions.value} Regions supported`;
 });
+
+// Local checked state for fade effect
+const checkedArr = ref(props.regions.map(() => false));
+
+onMounted(() => {
+	delayUntilVisible(() => {
```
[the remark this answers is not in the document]
dyersituations: True, but I could see us using these as a new checkbox variation long-term.

dyersituations:
Animation itself looks good, but it looks like the position of the circles changes slightly as the checks are animated in. Can we keep the sizing/positioning consistent? So the grays just change to the green checks without the elements moving? You can see it in the animation, but here I tried to call it out in a screenshot as well:

image

src/components/MyKiva/RoundCheckbox.vue
dyersituations, line 1: Oh there's still an extra `3` here that can be removed.

dyersituations (changes requested):
One small change before merge.

dyersituations (approved):

PR #6159, opened 2025-07-23
feat: resolve issue with TY page redirect and simplify MyKiva experiment logic

dyersituations:
https://kiva.atlassian.net/browse/MP-1854

- Simplified logic for MyKiva experiment assignment
- Removed user preferences usage -> MyKiva experiment no longer needs check for number of loans
- Moved user preferences utils to dedicated file -> these aren't being used but seem useful to keep around
- Fixed issue where all users were taken to MyKiva from TY page regardless of experiment assignment

### DIFF
--- a/src/graphql/query/shared/myKivaForAllUsers.graphql
+++ b/src/graphql/query/shared/myKivaForAllUsers.graphql
@@ -1,4 +1,7 @@
 query myKivaForAllUsers {
+	my {
+		id
+	}
 	general {
 		my_kiva_all_users: uiConfigSetting(key: "my_kiva_all_users_enable") {
 			key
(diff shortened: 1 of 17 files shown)


src/graphql/query/shared/myKivaForAllUsers.graphql
```diff
@@ -1,13 +1,6 @@
 query myKivaForAllUsers {
 	my {
 		id
```
roger-in-kiva, line 3: Needed?

mcstover (approved):
Nice to see this distilled a bit!

src/graphql/query/shared/myKivaForAllUsers.graphql
dyersituations: Making sure user is logged in.

roger-in-kiva (approved):

PR #6160, opened 2025-07-23
feat: carousel added to mykiva card

roger-in-kiva:
- carousel added to MyKiva card
- Part of https://kiva.atlassian.net/browse/MP-1751

Screenshot 2025-07-23 at 10 42 34 a m

### DIFF
--- a/src/components/MyKiva/MyKivaCard.vue
+++ b/src/components/MyKiva/MyKivaCard.vue
@@ -5,20 +5,30 @@
 		:style="{ backgroundImage: `url(${bgImage})` }"
 	>
 		<div
+			v-if="images.length"
+			class="tw-p-1"
+			style="min-height: 220px;"
+		>
+			<KvCarousel
+				class="carousel"
+				:embla-options="{ loop: false, align: 'center' }"
+				:is-dotted="true"
+			>
+				<template v-for="(image, index) in images" #[`slide${index+1}`] :key="index">
+					<img
+						:src="image"
+						class="tw-rounded tw-w-full"
+						loading="lazy"
+					>
+				</template>
+			</KvCarousel>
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/MyKivaCard.vue
```diff
@@ -5,20 +5,30 @@
 		:style="{ backgroundImage: `url(${bgImage})` }"
 	>
 		<div
-			v-if="tagText"
-			class="tw-absolute tw-bg-secondary tw-rounded tw-px-1.5 tw-py-0.5 tw-text-small tw-left-1.5 tw-top-1.5
-				tw-drop-shadow-sm tw-font-medium"
+			v-if="images.length"
```
dyersituations, line 8: What does the carousel look like with a single image?

roger-in-kiva: Nice call!
Difficult with same image size and controls dots, we can make it larger for single image case so.
But at least from design, same height as, for multiple case, having dots there just one row of text, and for single image case two rows of text. They are not intentionally to be shown together.

Screenshot 2025-07-23 at 11 39 51 a m
Screenshot 2025-07-23 at 11 39 57 a m

mcstover: Ok I just figured this could end up in a carousel alongside other cards.

I think you could use flex and grid ensure the heights are the same across all cards like we did in story cards. CTA/Buttons remain pinned to the bottom.

roger-in-kiva: Oh yeah, in that case we can fix it with a parent wrapper, not needed for now. Thanks!

dyersituations (approved):

PR #6167, opened 2025-07-25
feat: add card for regions - data - add tracker

infante-jaime:
- The parent now tracks when a specific part of the child (`loanRegionsElement` in `LendingStats`) is visible.
- This is done by exposing the ref from the child, referencing it in the parent, and using Intersection Observer for accurate, one-time event tracking.
- The observer is properly cleaned up on component unmount.

### DIFF
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -24,6 +24,7 @@
 		/>
 		<section v-if="isLendingStatsExp" class="tw-mt-4">
 			<LendingStats
+				ref="lendingStats"
 				:regions="lendingStats.regionsWithLoanStatus"
 				:loans="loans"
 			/>
@@ -264,6 +265,7 @@ export default {
 		} = useBadgeData(this.apollo);
 
 		return {
+			lendingStatsObserver: null,
 			badgeData,
 			blogCards: [],
 			blogCategories,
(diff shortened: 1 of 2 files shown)


src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -594,6 +596,30 @@ export default {
...
 		this.initializeIsBpModalEnabledExp('my-kiva-page-content');
+
+		this.$nextTick(() => {
+			const loanRegionsEl = this.$refs.lendingStats?.loanRegionsElement;
+			if (loanRegionsEl) {
+				const observer = new window.IntersectionObserver(
+					(entries, obs) => {
+						entries.forEach(entry => {
+							if (entry.isIntersecting) {
+								this.$kvTrackEvent('event-tracking', 'show', 'regions-lent-to');
```
roger-in-kiva, line 607: Why not doing this in LendingStats mounted hook?

mcstover:
Is there a reason we can't use our existing paradigm and utility for this as we do in our[ Borrower Profile components](https://github.com/kiva/ui/blob/main/src/components/BorrowerProfile/CommentsAndWhySpecial.vue#L213)?

src/pages/MyKiva/MyKivaPageContent.vue
infante-jaime: Updated it to match the way is done on the borrower profile but closing this one and making a new PR just to keep things clean.

infante-jaime:
moved to https://github.com/kiva/ui/pull/6170

PR #6171, opened 2025-07-27
feat: add variant checkbox

infante-jaime:
### Checkbox Variant Refactor

- **Added a `variant` prop** to support both `square` and `round` checkbox styles in a single component.
- The `variant` prop controls which visual style (classic square or modern round) is rendered.

### Disabled State

- The `disabled` prop disables the checkbox and applies a visual style to indicate it’s inactive.
- **For both variants:**  
  - When `disabled` is `true`, pointer events are blocked and the checkbox appears faded or blurred (depending on the `blurOnDisabled` prop).
  - The `blurOnDisabled` prop (default `true`) lets you choose between a blur effect or a standard faded/gray disabled look.

### DIFF
--- a/.storybook/stories/KvCheckbox.stories.js
+++ b/.storybook/stories/KvCheckbox.stories.js
@@ -167,6 +167,7 @@ export const RoundVariant = () => ({
                 id="round-1"
                 v-model="round1"
                 variant="round"
+				size="large"
             >
                 Round Option 1
             </kv-checkbox>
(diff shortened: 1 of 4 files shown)


dyersituations:
@infante-jaime we actually want to update the shared component here: https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-components/src/vue/KvCheckbox.vue. How tough would it be to port the changes there?

infante-jaime:
> @infante-jaime we actually want to update the shared component here: https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-components/src/vue/KvCheckbox.vue. How tough would it be to port the changes there?

it appears to be straight forward but will check it out tomorrow to see what extra moving parts we might have keeping this as reference but labeling  as do not merge

PR #6172, opened 2025-07-28
feat: my kiva all regions use case mp 1750

christian14b:
Show two achievements card when user lent to all regions

image

### DIFF
--- a/src/components/BorrowerProfile/BorrowerCountry.vue
+++ b/src/components/BorrowerProfile/BorrowerCountry.vue
@@ -66,8 +66,8 @@ export default {
 		this.loadCoordinates();
 	},
 	methods: {
-		loadCoordinates(loanId) {
-			if (!loanId) return;
+		loadCoordinates() {
+			if (!this.loanId) return;
 			return this.apollo.query({
 				query: gql`query borrowerCountryCoords($loanId: Int!) {
 					lend {
@@ -87,7 +87,7 @@ export default {
 						}
 					}
 				}`,
-				variables: { loanId }
+				variables: { loanId: this.loanId }
 			}).then(({ data }) => {
 				const geocode = data?.lend?.loan?.geocode;
 				this.mapLat = geocode?.latitude ?? null;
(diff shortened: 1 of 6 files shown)


src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -122,6 +141,10 @@ export default {
 					});
 					return { name: region, hasLoans };
 				});
+
+				this.userLentToAllRegions = regionsWithLoanStatus
```
dyersituations, line 145: Can this duplicate logic be shared?

src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -6,23 +6,12 @@
...
-		>
-			<template
-				v-for="(slide, index) in orderedSlides"
-				#[`slide${index}`]
-				:key="index"
-			>
+		<template v-if="inLendingStats">
+			<div class="tw-w-full tw-h-full tw-flex tw-flex-col md:tw-flex-row lg:tw-flex-nowrap tw-gap-4">
 				<MyKivaCard
+					v-for="(slide, index) in orderedSlides"
```
dyersituations, line 12: I think we might want to keep this in the carousel regardless, since we're now getting wrap in some screen sizes. The carousel should be able to handle all screen sizes:
image

dyersituations: Isn't the goal to have the 3 cards in a row? Using the carousel for all 3 might be a simple solution to ensure there's no wrap for different screen sizes.

dyersituations: For Shua's account, I'm no longer seeing the next step cards (shown in previous screenshot). Is that expected? I assumed I'd still be seeing the same 3 cards.
image

dyersituations: Sent in Slack

dyersituations (approved):

PR #6190, opened 2025-08-05
fix: mp-1900 reduce margin lending stats

infante-jaime:
Qa feedback fix


src/components/MyKiva/LendingStats.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tw-mb-3.5">
+	<div class="tw-mb-2.5">
```
dyersituations, line 2: `tw-mb-2` would match other sections.

PR #6192, opened 2025-08-05
fix: use top category according to achievement progress

dyersituations:
https://kiva.atlassian.net/browse/MP-1897
https://kiva.atlassian.net/browse/MP-1903

- Previously we were choosing the "top category" and getting the loan count based on the loans query that has a limit
- Now the data is being pulled from the achievement service data, which we are already using on MyKiva, I just created a new extended query
- Fixed a bug where we were displaying "persons" instead of "people"

image


mcstover (approved):
Looks good

PR #6193, opened 2025-08-06
feat: new kv header test integration

christian14b:
New header component integration under experiment flag for product testing

https://kiva.atlassian.net/browse/MP-1905

https://github.com/user-attachments/assets/fe092538-1199-41a3-9a33-bae734941150

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -96,7 +96,7 @@
 				<template v-else>
 					<div
 						class="header
-								tw-grid xl:tw-gap-x-4 tw-items-center"
+							tw-grid xl:tw-gap-x-4 tw-items-center"
 						:class="{
 							'tw-gap-x-1 ': isMobile,
 							'tw-gap-x-2.5': !isMobile,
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -103,442 +34,525 @@
...
+									data-testid="header-my-kiva-dropdown-list"
+									id="my-kiva-dropdown"
+									ref="userDropdown"
 								>
 									<ul>
+										<template v-if="isBorrower">
+											<li>
+												<router-link
+													to="/my/borrower"
+													v-kv-track-event="['TopNav','click-Portfolio-My borrower dashboard']"
```
dyersituations, line 443: Another long line here.

src/components/WwwFrame/TheHeader.vue
```diff
@@ -772,6 +791,10 @@ export default {
 			hasDepositBefore,
 		});
 		window.addEventListener('resize', this.determineIfMobile());
+
+		if (this.$refs?.newExpHeader) {
+			this.$refs?.newExpHeader?.getSuggestions?.(this.apollo);
```
dyersituations, line 796: In general would it be simpler to pass apollo to the component then have to use the ref in the parent?

src/components/WwwFrame/WwwPage.vue
```diff
@@ -1,10 +1,9 @@
 <template>
 	<div class="www-page">
 		<global-promo-contentful v-show="!isKivaAppReferral" />
-		<the-new-header v-if="isNavUpdateExp" />
 		<the-header
-			v-else
 			v-show="!isKivaAppReferral"
+			:is-nav-update-exp="isNavUpdateExp"
```
dyersituations, line 6: Would it be worth moving the experiment assignment/tracking to the header? It would eliminate one prop drilling.

PR #6196, opened 2025-08-06
fix: add missing dot to update card (MP-1829)

infante-jaime:
Just adding the missing dot here the filter issue will be addressed as a separate task
https://kiva.atlassian.net/browse/MP-1829


dyersituations (approved):

dyersituations:
@infante-jaime also note the transaction card subject line change mentioned in that ticket.

PR #6209, opened 2025-08-08
feat: prefetch exp query using client

christian14b:
`preFetching` experiment with client to forced variant

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -762,29 +762,29 @@ export default {
 		},
 		{
 			query: experimentAssignmentQuery,
-			preFetch: true,
-			variables() {
-				return {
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -762,29 +762,29 @@ export default {
 		},
 		{
 			query: experimentAssignmentQuery,
-			preFetch: true,
-			variables() {
-				return {
-					id: NAV_UPDATE_EXP_KEY
-				};
+			preFetch(_, client) {
+				return client.query({
```
emuvente, line 766: This isn't going to work for pages that use cacheable SSR (currently only UI-Site-Map). @christian14b @dyersituations How long do we expect to run this experiment? if it's going to be for a while, this experiment assignment should be handled with Fastly instead and the tracking should only happen in the `mounted` callback.

PR #6213, opened 2025-08-13
fix: atb modal showed based on user balance and basket total

roger-in-kiva:
- ATB modal showed based on user balance and basket total
- https://kiva.atlassian.net/browse/MP-1933

### DIFF
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -75,13 +75,13 @@ const myKivaFlagEnabled = ref(false);
 const tierTable = ref({});
 const milestonesProgress = ref({});
 const hasEverLoggedIn = ref(false);
+const basketTotal = ref(0);
 
 const basketCount = computed(() => {
 	return addedLoan.value?.basketSize ?? 0;
 });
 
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -116,6 +116,9 @@ const fetchBasketData = async () => {
 		},
 	}).then(({ data }) => {
 		basketData.value = data?.shop?.basket?.items?.values ?? [];
+		basketTotal.value = basketData.value.reduce((total, item) => {
```
dyersituations, line 119: This looks good for mount, but does this get ran each time the modal is opened? We'd need the updated basket total each time.

roger-in-kiva: For sure, we already have that here:
https://github.com/kiva/ui/blob/58dd2e27aba159216cbb15c633301421b7646cf2/src/components/WwwFrame/Header/KvAtbModalContainer.vue#L211

dyersituations: Ah yup, I missed the `addedLoan` part of the watcher, thanks!

dyersituations (approved):

PR #6228, opened 2025-08-21
feat: update header component and fix determineMobile method

christian14b:
https://kiva.atlassian.net/browse/MP-1696

Also fixing `determineIfMobile` not being called in `mounted` hook

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -647,6 +647,7 @@ export default {
 			trusteeId: null,
 			userId: null,
 			isNavUpdateExp: false,
+			throttledDetermineIfMobile: null,
 		};
 	},
 	emits: ['show-basket'],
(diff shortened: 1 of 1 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -810,20 +811,24 @@ export default {
...
+			this.determineIfMobile();
+		}, 200));
 
 		if (this.$refs?.newExpHeader) {
 			this.$refs?.newExpHeader?.getSuggestions?.(this.apollo);
 		}
 	},
 	beforeUnmount() {
-		window.removeEventListener('resize', this.determineIfMobile());
+		window.removeEventListener('resize', _throttle(() => {
```
dyersituations, line 825: Throttle can't be in the remove event listener, since it creates a new instance of the method.

dyersituations (approved):

PR #6229, opened 2025-08-22
feat: [MP-1916] - fix false positive cases in Vue Components as well

cristhianDt:


dyersituations (approved):

mcstover (approved):
Should be good, though we could probably allow any of these opt_outs to activate this var.

PR #6234, opened 2025-08-26
fix: [MP-1962] - fix issue with loader on funded borrower profile `/lend/:id`

cristhianDt:


mcstover (approved):

cristhianDt:
> > Do we need `isLoading` true as default value?
> > https://github.com/kiva/ui/blob/cc555271ee944b437e81b40ccf5f63ef2d6bf2d8/src/components/BorrowerProfile/FundedBorrowerProfile.vue#L185
> 
> Yeah it does look like having a default `true` value would prevent potential pop-in.

@dyersituations @roger-in-kiva changed

roger-in-kiva (approved):

PR #6240, opened 2025-09-01
fix: [MP-1980] - dont load gtm script when anyof X_opted_out is enable

cristhianDt:
https://kiva.atlassian.net/browse/MP-1980

### DIFF
--- a/src/head/script.js
+++ b/src/head/script.js
@@ -154,7 +154,7 @@ export default (config, globalOneTrustEvent) => {
 				if (config.enableGA && !optout && !piiOptout) {
 					OneTrust.InsertHtml('', 'head', insertGoogleAnalytics, null, 'C0004');
 				}
-				if (config.enableGTM && !optout) {
+				if (config.enableGTM && !optout && !piiOptout) {
 					OneTrust.InsertHtml('', 'head', insertGTM, null, 'C0004');
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/head/script.js
```diff
@@ -154,7 +154,7 @@ export default (config, globalOneTrustEvent) => {
 				if (config.enableGA && !optout && !piiOptout) {
 					OneTrust.InsertHtml('', 'head', insertGoogleAnalytics, null, 'C0004');
 				}
-				if (config.enableGTM && !optout) {
+				if (config.enableGTM && !optout && !piiOptout) {
```
dyersituations, line 157: @mcstover @emuvente I was curious, do we definitely want to exclude GTM when we opt out of GA or are there use cases I'm not thinking of? The trigger for these changes was that I noticed GTM scripts being loaded on UI pages when testing the cookie changes on dev when the PII opt out was enabled only (new requirement for opting out of signals).

emuvente: GTM is loading things besides GA. I thought the PII opt out was only for GA itself?

dyersituations: I was curious though since we're also checking `!optout` here from the opt-out cookie. Should we also not check the new PII output requirement for signals?

cristhianDt:
closing as it's not necessary anymore

PR #6251, opened 2025-09-04
feat: mp-1826 / Redirect Lend Menu Link to Combo Page with New Countries Haven't Tent to Experiment

The-Best-Joke:
fix: replace old countries not lent to link with new redirect

### DIFF
--- a/src/components/WwwFrame/LendMenu/LendListMenu.vue
+++ b/src/components/WwwFrame/LendMenu/LendListMenu.vue
@@ -150,7 +150,7 @@
 						</li>
 						<li>
 							<a
-								href="/lend/countries-not-lent"
+								href="/lend-category-beta?countries-not-lent-to=true"
 								class="lend-link tw-text-primary"
 								v-kv-track-event="['TopNav','click-Lend-Countries_Not_Lent']"
 							>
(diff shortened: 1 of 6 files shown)


src/components/WwwFrame/LendMenu/LendListMenu.vue
```diff
@@ -150,7 +150,7 @@
 						</li>
 						<li>
 							<a
-								href="/lend/countries-not-lent"
+								href="/lend-category-beta?countries-not-lent-to=true"
```
dyersituations, line 153: For this experiment, we can just change the lend menu, so this component and `LendMegaMenu.vue`.

And we only want to use the new URL if the experiment is enabled.

src/components/WwwFrame/LendMenu/LendMegaMenu.vue
```diff
@@ -121,7 +121,9 @@
 									</li>
 									<li>
 										<a
-											href="/lend/countries-not-lent"
+											:href="isCountriesNotLentToEnabled ?
+												'/lend-category-beta?countries-not-lent-to=true'
```
dyersituations, line 125: `/lend/filter?countries-not-lent-to=true`

PR #6258, opened 2025-09-09
feat: create goal setting modal and form first step

christian14b:
https://kiva.atlassian.net/browse/MP-2010

image


mcstover (approved):
Thank you for revisting with the global component!

PR #6266, opened 2025-09-12
feat: mp-2042 / Store Goal Setting to User Preferences When Modal is Saved

The-Best-Joke:
fix: apply initial code for setting goals

### DIFF
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -191,7 +191,7 @@ import { defaultBadges } from '#src/util/achievementUtils';
 import { fireHotJarEvent } from '#src/util/hotJarUtils';
 import { runRecommendationsQuery } from '#src/util/loanSearch/dataUtils';
 import logReadQueryError from '#src/util/logReadQueryError';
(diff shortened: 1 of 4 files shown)


src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -190,6 +191,7 @@ import { defaultBadges } from '#src/util/achievementUtils';
 import { fireHotJarEvent } from '#src/util/hotJarUtils';
 import { runRecommendationsQuery } from '#src/util/loanSearch/dataUtils';
 import logReadQueryError from '#src/util/logReadQueryError';
+import updateUserPreferences from '#src/util/userPreferencesUtils';
```
dyersituations, line 194: `userPreferenceUtils`

The-Best-Joke:
@dyersituations Testing was done. Turns out a fair bit more work was still pending both to handle instances of create vs update user preferences as well as handling proper code logic to not simply override but instead correctly update existing user preferences.

That's all done now.

dyersituations (approved):
Good catch on needing to create, I forgot about that step!

PR #6275, opened 2025-09-17
fix: [MP-2046] issues with experiment header for nav with cache

cristhianDt:
Sync experiment handler with CPS logic, following the order:

1. fastly header variant/value 
2. query param from `setuiab` cookie
3. cookieStore

### DIFF
--- a/server/vue-middleware.js
+++ b/server/vue-middleware.js
@@ -104,7 +104,7 @@ export default function createMiddleware({ config, vite }) {
 
 		// Handle a single home page experiment
 		if (homePageHeader) {
-			forceHeader = true;
+			forceHeader = homePageHeader;
 		}
 
 		// Setup rendering context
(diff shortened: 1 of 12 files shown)


cristhianDt:
The tests aren't ready yet, I'm working on it. Since I can't test the Fastly header locally I did some tests with hardcoded headers and works so I'd like to get feedback while I work on the unit tests.

server/vue-middleware.js
```diff
@@ -106,6 +118,7 @@ export default function createMiddleware({ config, vite }) {
 			locale: req.locale,
 			device,
 			cdnNotedLoggedIn,
+			forceHeader,
```
dyersituations, line 121: In CPS we just pass the headers through and use those in the experiment utils. Is that not a good option here?

PR #6278, opened 2025-09-18
feat: store current loan in category and calculate goal card progress

christian14b:
https://kiva.atlassian.net/browse/MP-2065

Support all
image

### DIFF
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -178,7 +178,7 @@ import AsyncMyKivaSection from '#src/pages/MyKiva/AsyncMyKivaSection';
 import MyKivaBorrowerCarousel from '#src/components/MyKiva/BorrowerCarousel';
 import JournalUpdatesCarousel from '#src/components/MyKiva/JournalUpdatesCarousel';
 import MyKivaStats from '#src/components/MyKiva/MyKivaStats';
-import useBadgeData from '#src/composables/useBadgeData';
+import useBadgeData, { ID_SUPPORT_ALL } from '#src/composables/useBadgeData';
 import LatestBlogCarousel from '#src/components/MyKiva/LatestBlogCarousel';
 import LendingCategorySection from '#src/components/LoanFinding/LendingCategorySection';
 import JourneySideSheet from '#src/components/Badges/JourneySideSheet';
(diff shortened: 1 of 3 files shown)


src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -366,7 +366,17 @@ export default {
...
 			const parsedPreferences = JSON.parse(preferences.preferences);
 			const existingGoals = parsedPreferences?.goals || [];
-			return existingGoals.length ? existingGoals[0] : null;
+			let goal = existingGoals.length ? existingGoals[0] : null;
+			// eslint-disable-next-line max-len
+			const achievementProgress = this.heroTieredAchievements.find(ach => ach.id === goal?.category)?.totalProgressToAchievement ?? 0;
+			if (goal) {
+				goal = {
+					...goal,
+					currentProgress: (achievementProgress - (goal?.currentLoanCount || 0)),
```
dyersituations, line 375: I don't think this gives us the number we need. What we need to know is how many loans they've lent to since setting this goal.

So we probably need to set the loan count when the goal is set, then we can compare starting loan count against new loans, or something along those lines.

Another option is using the start date to compare against transactions, but if a user is really active, there might be too many transactions to pull without creating a special endpoint.

dyersituations: Oh I see now, pulling from the category data. Yeah that works for the categories that match achievements. And yeah, I think maybe `loanTotalAtStart` might be clearer.

src/components/MyKiva/GoalCard.vue
```diff
@@ -145,18 +156,30 @@ const ctaHref = computed(() => {
 const achievementGoalImg = computed(() => {
 	const contentfulCategory = getContentfulKey(props.userGoal?.category) || '';
 	if (!contentfulCategory) return '';
-	const key = `my-kiva-${contentfulCategory}-journey`;
+	const key = `my-kiva-${'women'}-journey`;
```
dyersituations, line 159: Does the logic here need to change? Should this be `contentfulCategory`? What happens for "all loans" goal?

dyersituations: What happens for the "all loans" category? Do we need a new contentful entry?

dyersituations (approved):

PR #6281, opened 2025-09-19
feat: mp-2014 / Display Thank You Page and Prompt for New Achievement After Goal Completion

The-Best-Joke:
[JIRA Ticket](https://kiva.atlassian.net/jira/software/projects/MP/boards/275?selectedIssue=MP-2014)

feat: add preliminary work for MP-2014

### DIFF
--- a/src/components/Thanks/SingleVersion/GoalCompleted.vue
+++ b/src/components/Thanks/SingleVersion/GoalCompleted.vue
@@ -18,8 +18,8 @@
 					</h3>
 				</div>
 				<div class="tw-relative">
-					<BgRays style="top: -50px;" />
-					<BadgeContainer :show-shine="true">
+					<BgRays style="top: -90px;" />
+					<BadgeContainer :show-shine="true" style="top: -50px">
 						<HighFive alt="Hi five icon" style="width: 250px; height: 250px;" />
 					</BadgeContainer>
 				</div>
--- a/src/components/Thanks/ThanksPageSingleVersion.vue
+++ b/src/components/Thanks/ThanksPageSingleVersion.vue
@@ -16,8 +16,15 @@
 				class="tw-mb-2.5"
 				@view-pdf-clicked="scrollToReceipt"
 			/>
+			<GoalCompleted
+				v-if="isNextStepsExpEnabled && userGoalAchieved"
+				:current-goal="userGoal"
+				:get-goal-display-name="getGoalDisplayName"
+				:loading="goalDataLoading"
+				class="tw-mb-2.5"
+			/>
 			<BadgeMilestone
-				v-if="showBadgeModule || achievementsCompleted"
+				v-else-if="showBadgeModule || achievementsCompleted"
 				:is-guest="isGuest"
 				:is-opted-in="isOptedIn"
 				:badge-achieved-ids="badgeAchievedIds"
@@ -27,13 +34,6 @@
 				:kiva-cards-module-shown="showKivaCardsModule"
 				:achievements-completed="achievementsCompleted"
 				@continue-clicked="handleContinue"
-				class="tw-mb-2.5"
-			/>
-			<GoalCompleted
-				v-if="isNextStepsExpEnabled && userGoalAchieved"
-				:current-goal="userGoal"
-				:get-goal-display-name="getGoalDisplayName"
-				:loading="goalDataLoading"
 				class="tw-mb-2.5"
 			/>
 			<JourneyGeneralPrompt
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
 				"@godaddy/terminus": "^4.11.0",
 				"@graphql-tools/load": "^7.7.0",
 				"@graphql-tools/url-loader": "^7.17.18",
-				"@kiva/kv-components": "^6.62.3",
+				"@kiva/kv-components": "^6.62.5",
 				"@kiva/kv-shop": "^3.3.38",
 				"@kiva/kv-tokens": "^3.4.1",
 				"@mdi/js": "^7.4.47",


src/components/Thanks/SingleVersion/GoalCompleted.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+	<div
+		class="tw-rounded md:tw-rounded-xl tw-bg-white
+			tw-shadow-lg tw-px-3 md:tw-px-8
+			tw-py-4 tw-flex tw-flex-col
+			tw-gap-2 print:tw-hidden tw-items-center
+			tw-text-center tw-overflow-hidden tw-relative"
+	>
+		<KvLoadingPlaceholder v-if="isLoading" class="!tw-h-9 !tw-rounded" />
```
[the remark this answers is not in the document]
dyersituations: isLoading is indeed not defined.

src/graphql/query/userPreferences.graphql
```diff
@@ -0,0 +1,9 @@
+query userPreferencesQuery {
```
dyersituations, line 1: This query isn't being used, but we do need to load the preferences. You might be able to include this on the TY page query.

src/graphql/query/thankYouPage.graphql
```diff
@@ -0,0 +1,52 @@
+query thankYouPageQuery {
```
dyersituations, line 1: See other comment, I think this can be replaced with a simpler query.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,152 @@
...
+		const parsedPrefs = await loadPreferences();
+		initializeGoalState(parsedPrefs);
+		countLoansTowardGoals();
+		// Auto-update if active goal achieved
+		if (currentGoalAchieved.value && activeGoal.value) {
+			await storeGoalPreferences({
+				goalName: activeGoal.value.goalName,
+				dateStarted: activeGoal.value.dateStarted,
+				target: activeGoal.value.target,
+				count: activeGoal.value.count,
```
dyersituations, line 137: The `count` currently isn't being set when a goal is created. It might not be necessary if we use the user's total loan count for all loans and their achievement progress for goals related to achievements. See how we use `currentProgress` for goals elsewhere.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,152 @@
...
+	function countLoansTowardGoals() {
+		const currentYear = new Date().getFullYear();
+		loans
+			.filter(
+				loan => loan.disbursalDate
+				&& new Date(loan.disbursalDate).getFullYear() >= currentYear
+			)
+			.forEach(loan => {
+				(loan.tags || []).forEach(tag => {
+					getCategoriesForTag(tag).forEach(category => {
```
dyersituations, line 97: This check actually won't work, since the goals/achievements aren't related to just tags. See the filters in `useBadgeData`. Like mentioned in Slack, you should be able to utilize achievement service endpoints, like `fetchPostCheckoutAchievements`. Not sure if all the data needed is already surfaced, but we should leverage the achievement service if we can. There is also some timing around when the achievement service updates the totals on its side, hence why we use the `fetchPostCheckoutAchievements` method, it doesn't need to wait for results to change in achievement service. We can always extend the achievement service as needed.

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -228,18 +252,28 @@ const handleContinue = () => {
...
+		'event-tracking',
+		NEXT_STEPS_EXP_KEY,
+		'EXP-MP-1984-Sept2025'
+	);
+	isNextStepsExpEnabled.value = nextStepsExpData.version === 'b';
+};
 
+onMounted(async () => {
+	determineNextStepsExpEnabled();
+	await runGoalComposable();
```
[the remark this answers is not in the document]
The-Best-Joke: That's a good catch

src/composables/useGoalData.js
```diff
@@ -0,0 +1,152 @@
...
+			});
+	}
+
+	// Count loans toward goals (only current-year loans)
+	function countLoansTowardGoals() {
+		const currentYear = new Date().getFullYear();
+		loans
+			.filter(
+				loan => loan.disbursalDate
+				&& new Date(loan.disbursalDate).getFullYear() >= currentYear
```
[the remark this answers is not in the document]
dyersituations: Take a look at `ThanksPage`, the loans are only from the receipt. There really isn't a great way to load all loans for a user in general, since some users have 1000s of loans, too much data to load paged, so we don't really do it anywhere (except maybe the monolith old pages).

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -27,8 +27,16 @@
...
+				v-if="isNextStepsExpEnabled && currentGoalAchieved"
+				:active-goal="activeGoal"
+				:current-goal-achieved="currentGoalAchieved"
+				:get-goal-display-name="getGoalDisplayName"
+				:loading="goalDataLoading"
+				class="tw-mb-2.5"
+			/>
 			<JourneyGeneralPrompt
-				v-if="showJourneyModule"
+				v-if="!isNextStepsExpEnabled && showJourneyModule"
```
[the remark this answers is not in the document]
dyersituations: This comment is still applicable.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,152 @@
...
+export default function useGoalData(loans) {
+	const apollo = inject('apollo');
+
+	const loading = ref(true);
+	const goalState = ref({});
+	const userPreferences = ref(null);
+
+	async function loadPreferences() {
+		try {
+			const response = await apollo.query({ query: thankYouPageQuery });
```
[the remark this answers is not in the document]
dyersituations: This comment is still applicable.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,165 @@
...
+	const userPreferences = ref(null);
+	const totalLoans = ref(0);
+	const heroTieredAchievements = ref([]);
+
+	async function loadPreferences() {
+		try {
+			const response = await apollo.query({ query: thankYouPageQuery });
+			const prefsData = response.data?.my?.userPreferences || null;
+			totalLoans.value = response.data?.my?.loans?.totalCount ?? 0;
+			heroTieredAchievements.value = response?.data?.userAchievementProgress?.tieredLendingAchievements ?? [];
```
dyersituations, line 65: When testing locally was `userAchievementProgress` update right after the checkout transaction? I can't test locally currently due to the dev env being down. I ask because `userAchievementProgress` gets updated via a Kafka listener after the transaction, so there is some delay between transaction and progress being updated. The reason we use `fetchPostCheckoutAchievements` in other TY page modules related to badges is that that query doesn't wait for the Kafta events to run, it just checks "what if" the user checked out with the loans for the current user.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,127 @@
...
+		const goals = parsedPrefs.goals || [];
+		const goalIndex = goals.findIndex(g => g.goalName === updates.goalName);
+		if (goalIndex !== -1) goals[goalIndex] = { ...goals[goalIndex], ...updates };
+		else goals.push(updates);
+		await updateUserPreferences(apollo, userPreferences.value, parsedPrefs, { goals });
+		setGoalState({ goals }); // Refresh local state after update
+	}
+
+	const goalProgress = computed(() => {
+		const currentProgress = overallProgress.value.find(
```
dyersituations, line 91: This works for goals that match achievements but not the "all loans" goal. We'll need to also check the loan total for the current user, which should also already be in the TY page query.

src/composables/useGoalData.js
```diff
@@ -0,0 +1,127 @@
...
+			entry => entry.achievementId === currentGoal.value?.category
+		);
+		return (currentProgress?.contributionLoanIds || []).length;
+	});
+
+	const currentGoalActive = computed(() => currentGoal.value.status === 'in-progress');
+	const currentGoalAchieved = computed(() => goalProgress.value >= currentGoal.value?.target);
```
dyersituations, line 98: I think this needs to check against `loanTotalAtStart` + `target` on the goal

src/composables/useGoalData.js
```diff
@@ -0,0 +1,127 @@
...
+		else goals.push(updates);
+		await updateUserPreferences(apollo, userPreferences.value, parsedPrefs, { goals });
+		setGoalState({ goals }); // Refresh local state after update
+	}
+
+	const goalProgress = computed(() => {
+		const currentProgress = overallProgress.value.find(
+			entry => entry.achievementId === currentGoal.value?.category
+		);
+		return (currentProgress?.contributionLoanIds || []).length;
```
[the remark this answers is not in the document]
The-Best-Joke: Based on the behavior I noticed locally after creating a goal with no contributing loans and then doing one such loan, I believe so, yes, which is the main reason why the `currentGoalAchieved` computed variable does not rely on the `totalLoansAtStart` value either. Admittedly it's an assumption.

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -27,8 +27,16 @@
 				@continue-clicked="handleContinue"
 				class="tw-mb-2.5"
 			/>
+			<GoalCompleted
```
dyersituations: Add a story in the storybook file for this use case, this component is pretty well covered in storybook.

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -228,18 +256,28 @@ const handleContinue = () => {
...
+		'event-tracking',
+		NEXT_STEPS_EXP_KEY,
+		'EXP-MP-1984-Sept2025'
+	);
+	isNextStepsExpEnabled.value = nextStepsExpData.version === 'b';
+};
 
+onMounted(async () => {
+	determineNextStepsExpEnabled();
+	if (isNextStepsExpEnabled.value) await runGoalComposable();
```
dyersituations: I'm not seeing progress towards the goal when I test locally (setting goal to women and checking out locally). There is progress towards the badge journey, so the achievement service is getting updated with the checkouts.

dyersituations: I can't test the TY page until this works.

dyersituations (changes requested):
Critical changes needed, see comments, thanks!

src/components/MyKiva/LendingStats.vue
```diff
@@ -263,6 +290,14 @@ const regionImages = {
 	'South America': SouthAmerica,
 };
 
+const {
+	runComposable: runGoalComposable,
+	loading: goalProgressLoading,
+	userGoal,
+	userGoalAchieved,
+	goalProgress,
+} = useGoalData({ loans: props.loans });
```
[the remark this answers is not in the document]
The-Best-Joke: Resolved. Thanks for picking up on it.

src/components/Thanks/ThanksPageSingleVersion.vue
The-Best-Joke: Fixed. There was a silly little bug in the way `goalProgress` was being calculated in the composable but it's sorted out now.

src/components/MyKiva/LendingStats.vue
```diff
@@ -362,8 +380,13 @@ export default {
...
 		},
-		setGoal(preferences) {
-			this.$emit('store-goals-preferences', preferences);
+		async setGoal(preferences) {
+			const {
+				storeGoalPreferences,
+			} = useGoalData({
+				loans: this.loans,
+			});
+			await storeGoalPreferences(preferences);
```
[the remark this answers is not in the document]
dyersituations: Fixed.

src/components/Thanks/ThanksPageSingleVersion.vue
dyersituations: We can resolve in follow-up PR.

PR #6292, opened 2025-09-26
feat: portfolio lending stats exclusions

roger-in-kiva:
- portfolio lending stats exclusions
- https://kiva.atlassian.net/browse/MP-1804

### DIFF
--- a/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
+++ b/src/pages/Portfolio/LendingStats/LendingStatsPage.vue
@@ -99,6 +99,24 @@ import useBadgeData from '#src/composables/useBadgeData';
 import BadgesSection from './BadgesSection';
 import StatsSection from './StatsSection';
 
+const convertToNumberOrString = value => {
+	const num = Number(value);
+
+	if (!Number.isNaN(num) && value !== '') {
+		return num;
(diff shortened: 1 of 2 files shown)


src/pages/Portfolio/LendingStats/LendingStatsPage.vue
```diff
@@ -135,24 +153,38 @@ export default {
 		preFetch: true,
 		result({ data }) {
 			const countriesWhereKivaDoesntWork = data?.general?.excluded_countries?.value ?? '';
+			/* eslint-disable max-len */
+			const activitiesWhereKivaDoesntWork = (data?.general?.excluded_activities?.value ?? '').split(',').map(item => item.replaceAll('"', '')).map(item => convertToNumberOrString(item));
```
dyersituations, line 157: Why is the `replaceAll` needed for these but not `excluded_countries`?

roger-in-kiva: For countries we are not splitting based on a comma, said that, as a result we have an array including `"` in elements we need to clean to match with data.

We are doing this to have support to use names or IDs, for countries we use ISO codes, so we don't need same approach.

PR #6298, opened 2025-10-01
fix: use jpeg image instead of webp to match node-canvas support

mcstover:
Fixes broken image loading through node-canvas loadImage which doesn't support webp.


mac-arrap (approved):

emuvente (approved):
Looks good, just noting that this would work with .jpg extensions as well

PR #6299, opened 2025-10-01
feat: update lender profile send a kiva card functionality

cfowles:
- Use new buy page URL with correct query params for pre-filling.

### DIFF
--- a/src/components/LenderProfile/LenderSummary.vue
+++ b/src/components/LenderProfile/LenderSummary.vue
@@ -55,6 +55,7 @@
 					>
 						Send message
 					</kv-button>
+					<!-- eslint-disable max-len -->
 					<kv-button
 						class="tw-w-full lg:tw-w-auto"
 						variant="secondary"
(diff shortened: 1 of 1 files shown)


src/components/LenderProfile/LenderSummary.vue
```diff
@@ -55,10 +55,11 @@
 					>
 						Send message
 					</kv-button>
+					<!-- eslint-disable max-len -->
```
dyersituations, line 58: Doesn't this disable it for the entire file?

dyersituations (approved):

cfowles:
@dyersituations Unsure on the timing for when I merge this. Will that cause issues since it points to the -beta route or will the deploys be synced up ok?

dyersituations:
> @dyersituations Unsure on the timing for when I merge this. Will that cause issues since it points to the -beta route or will the deploys be synced up ok?

Oh I missed this was outside the experiment. We don't want to merge this as is. We'll be updating the new page to sit where the old is. Please update the URL accordingly.

dyersituations (changes requested):
See comment.

PR #6305, opened 2025-10-02
fix: impact goals mykiva cleanup

dyersituations:
https://kiva.atlassian.net/browse/MP-2112

I started testing the impact goal experiment use cases locally and found some holes, so I worked to cleanup what is shown on MyKiva in different stats of goals, achievements, etc.

The "top category" and lending experiment was also removed (confirmed with product) to simply the components.


src/components/Contentful/JourneyCardCarousel.vue
```diff
@@ -251,13 +259,13 @@ const orderedSlides = computed(() => {
 		];
 	}
 
-	if (props.slidesNumber) {
-		sortedSlides = sortedSlides.slice(0, props.slidesNumber);
+	if (shouldShowGoalCard.value) {
+		// Add empty slide at start for goal card
```
roger-in-kiva, line 263: Looks like GoalCard condition is based on an index `showGoalCard(index)`
Are we guaranteed always visible, or we could be adding this slide at start but not showing a Goal Card?

PR #6311, opened 2025-10-03
feat: [MP-1855] - show cta to next percentile for current year

cristhianDt:
# MP-1855

## Summary

- show cta to next percentile for current year

https://kiva.atlassian.net/browse/MP-1855

### Case: Next25 don't change

use `amount/currentThreshold - nextPercentileThreshold` as amount to lend

Screenshot 2025-10-03 at 00 15 17

--------------

### Case: Max level reached

Screenshot 2025-10-03 at 00 13 08

--------------

### Case: without Next25 (zero loans)

Screenshot 2025-10-03 at 00 12 05


src/pages/Portfolio/ImpactDashboard/LendingInsights.vue
```diff
@@ -513,6 +530,18 @@ export default {
 			}
 		},
 	},
+	mounted() {
+		this.unwatch = this.$watch('currentYearPercentile', newVal => {
```
cristhianDt, line 534: i find it more readable to use a watcher to register the event when we show the YTD panel to the user as first time by default, mainly because we have a 'lateinit' of the current percentile value. It could also be moved to `fetchCurrentYearStats`. let me know if you have any thoughts on this

dyersituations: Is this to trigger when the second tab is clicked?

dyersituations: Looks like fetch is only called when async component is visible. But you can verify locally by scrolling away from component and back.

dyersituations: If it only gets called once, it would be good to move this to where the value is being set for simplicity.

src/pages/Portfolio/ImpactDashboard/LendingInsights.vue
```diff
@@ -165,25 +166,30 @@
...
+									{{ nextPercentileMsg }}
 									<kv-material-icon
 										class="tw-ml-0.5 tw-w-2 tw-h-2"
 										:icon="mdiArrowRight"
 									/>
 								</router-link>
+								<span
+									v-else-if="currentYearPercentile === 99"
+									class="stat-link tw-text-eco-green-2 tw-font-medium"
+									style="display: inline-flex; align-items: center;"
```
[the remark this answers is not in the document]
infante-jaime: updated

PR #6315, opened 2025-10-06
fix: refresh totals for other basket errors in upsell module

christian14b:
https://kiva.atlassian.net/browse/MP-2121


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1062,6 +1062,7 @@ export default {
 									`Failed: ${error.message.substring(0, 40)}...`
 								);
 								Sentry.captureMessage(`Add to Basket: ${error.message}`);
+								this.refreshTotals();
```
dyersituations, line 1065: Does this both clear the loader on the page and get a new upsell?

dyersituations (approved):

PR #6800, opened 2026-04-02
feat: fixed url redirect

tiarascoleman:
JIRA Ticket: https://kiva.atlassian.net/jira/software/projects/MP/boards/275?selectedIssue=MP-2603


src/router/routes.js
```diff
@@ -435,6 +435,15 @@ export default [
 			excludeFromStaticSitemap: true,
 		},
 		beforeEnter(to, from, next) {
```
christian14b: Instead of this, is it possible to add a /mykiva/ route and redirect to /mykiva instead?

dyersituations (changes requested):

dyersituations (changes requested):
See comments.

christian14b (approved):

dyersituations (approved):
Nice cleanup!

PR #6802, opened 2026-04-06
feat: mp-2576-lower-equity

infante-jaime:
Jira ticket 
[MP-2576](https://kiva.atlassian.net/browse/MP-2576)

Updated Thanks badge selection so equity is treated as a fallback, and any earned non-equity badge is preferred for display when both are present.
Added composable helpers to pick the fallback badge and to override equity when non-equity badge metadata is only available from Contentful.
Adjusted BadgeMilestone wiring and tests to cover guest fallback, override behavior, and naming consistency with the new non-equity-first logic.

[MP-2576]: https://kiva.atlassian.net/browse/MP-2576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Thanks/SingleVersion/BadgeMilestone.vue
+++ b/src/components/Thanks/SingleVersion/BadgeMilestone.vue
@@ -52,7 +52,7 @@ import {
 	onMounted,
 } from 'vue';
 import { mdiArrowRight } from '@mdi/js';
-import useBadgeData, { ID_EQUITY } from '#src/composables/useBadgeData';
+import useBadgeData from '#src/composables/useBadgeData';
 import {
 	KvMaterialIcon, KvButton, KvLoadingPlaceholder
 } from '@kiva/kv-components';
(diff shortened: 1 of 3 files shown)


dyersituations (changes requested):
See comments about guest and first time logged in user use cases.

src/components/Thanks/SingleVersion/BadgeMilestone.vue
```diff
@@ -150,6 +151,10 @@ const displayedBadgeData = computed(() => {
 			return badgeDataAchieved.value[0];
 		}
 		const displayedBadge = getHighestPriorityDisplayBadge(badgeDataAchieved.value);
+		// Contentful-only badges (non-equity badges not in achievement service) have no achievementData
+		if (!displayedBadge.achievementData) {
```
dyersituations: Why is this change needed? Non-tiered badges displayed correctly before.

infante-jaime: getNonEquityBadgeOverride can add a badge that only comes from Contentful — it doesn't have achievementData. Without this check, getLastCompletedBadgeLevelData doesn't find anything useful and returns empty, so the badge breaks. This just makes sure those badges show up correctly since they already have what they need.

dyersituations (approved):
@infante-jaime for testing this on dev, please enable the IWD 2026 through the day of testing, try checking out with a new user to ensure IWD 2026 badge is on TY instead of equity.

PR #6805, opened 2026-04-06
feat: mp-2626-next steps bugs

infante-jaime:
https://kiva.atlassian.net/browse/MP-2626
https://kiva.atlassian.net/browse/MP-2627
https://kiva.atlassian.net/browse/MP-2628
https://kiva.atlassian.net/browse/MP-2629

Prevented repeated cards by aligning top-row and CYLA card selection logic so cards shown in one section are excluded from the other.
Applied non-repeated card behavior for both desktop and mobile flows.
Added consistent loading placeholders for goal/CYLA card areas so cards render in stable slots after goal data is ready.
Introduced a shared goal-card placeholder component to keep loading UI consistent across paths.

### DIFF
--- a/src/components/MyKiva/LendingStats.vue
+++ b/src/components/MyKiva/LendingStats.vue
@@ -161,6 +161,14 @@
 				</div>
 			</div>
 		</template>
+		<div
+			v-else-if="goalProgressLoading"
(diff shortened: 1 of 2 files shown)


dyersituations (changes requested):
See comments.

dyersituations (approved):

PR #6822, opened 2026-04-10
feat(apollo): add lazy query option to Apollo plugin

emuvente:
## Summary
- Add `lazy` option to Apollo operation config that defers query execution until the component enters the viewport via `IntersectionObserver`, reducing unnecessary data fetching for below-the-fold content
- Handle Vue 3 fragment components where `$el` may be a comment/text node by walking siblings to find the first actual `Element`
- Extract `setupWatchQuery` helper for reuse across eager and lazy paths
- Split `config/local.js` into `local-prod.js` and `local-dev.js` for clearer local dev configuration

## Usage

Add `lazy: true` to any Apollo operation to defer it until the component is near the viewport (default 500px root margin):

```js
apollo: {
  lazy: true,
  query: myQuery,
  variables() { return { id: this.id }; },
  result({ data }) { this.myData = data; },
}
```

Custom `IntersectionObserver` options (any valid option can be overridden):

```js
apollo: {
  lazy: { rootMargin: '200px', threshold: 0.5 },
  query: myQuery,
  // ...
}
```

Custom observation target via `$refs` (defaults to `$el`):

```js
apollo: {
  lazy: { target: 'lazyRoot' },
  query: myQuery,
  // ...
}
```
```html
<template>
  <div ref="lazyRoot">...</div>
</template>
```
If the operation has `preFetch: true` and data was already fetched server-side, the lazy option is skipped and the watch query is set up immediately so no observer is created.

## Example refactor

Here's what a refactor looks like using `WhySpecial.vue` as an example (pulled from the [feature/borrower-profile-migration-ui](https://github.com/kiva/ui/tree/feature/borrower-profile-migration-ui) branch):

**Before:** manual observer setup, teardown, and a separate `loadData` method:

```vue
<script>
import { gql } from 'graphql-tag';
import { createIntersectionObserver } from '#src/util/observerUtils';

export default {
  name: 'WhySpecial',
  // ...
  apollo: {
    query: gql`query whySpecial($loanId: Int!) { ... }`,
    variables() { return { loanId: this.loanId }; },
    result(result) {
      this.whySpecial = result?.data?.lend?.loan?.whySpecial ?? '';
    },
  },
  methods: {
    createObserver() {
      this.observer = createIntersectionObserver({
        targets: [this.$el],
        rootMargin: '500px',
        callback: entries => {
          entries.forEach(entry => {
            if (entry.target === this.$el && entry.intersectionRatio > 0) {
              this.loadData();
            }
          });
        }
      });
      if (!this.observer) {
        this.loadData();
      }
    },
    destroyObserver() {
      if (this.observer) {
        this.observer.disconnect();
      }
    },
    loadData() {
      this.apollo.query({
        query: gql`query whySpecial($loanId: Int!) { ... }`,
        variables: { loanId: this.loanId },
      }).then(result => {
        this.whySpecial = result?.data?.lend?.loan?.whySpecial ?? '';
        this.loading = false;
      });
    },
  },
  mounted() { this.createObserver(); },
  beforeUnmount() { this.destroyObserver(); },
};
</script>
```

**After:** one `lazy: true` declaration replaces observer setup, teardown, loadData, and the duplicated query:

```vue
<script>
import { gql } from 'graphql-tag';

export default {
  name: 'WhySpecial',
  // ...
  apollo: {
    lazy: true,
    query: gql`query whySpecial($loanId: Int!) { ... }`,
    variables() { return { loanId: this.loanId }; },
    result(result) {
      this.whySpecial = result?.data?.lend?.loan?.whySpecial ?? '';
      this.loading = false;
    },
  },
  // No methods, mounted, or beforeUnmount needed for lazy loading
};
</script>
```

This eliminates the `createIntersectionObserver` import, three methods, two lifecycle hooks, and a duplicated query definition per component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ $ npm run dev -- --config=local (any context outside of kiva vm)
 # visit localhost:8888/ui-site-map to explore some pages (/styleguide or /lend-by-category may be of interest)
 
 # Alternate configs:
-# use `dev-local` to run localhost mode against development Environments
-# use `dev-vm-mac` to run ui on your mac against the kiva vm
+# use `local-dev` to run localhost mode against development Environments
+# use `dev-custom-host` to run with Caddy at https://kiva-ui.local (enables Auth0)
 
 # COMPILED/PROD MODE
 
--- a/config/local-dev.js
+++ b/config/local-dev.js
@@ -1,0 +1,21 @@
+import merge from 'deepmerge';
+import base from './k8s-local.js';
+
+export default merge(base, {
+	app: {
+		apolloBatching: false, // Remove this once batching is enabled in prod
+		auth0: {
+			enable: false,
+		},
+		graphqlUri: 'https://gateway.development.kiva.org/graphql',
+		host: 'localhost',
+		photoPath: 'https://www.development.kiva.org/img/',
+		publicPath: '/',
+	},
+	server: {
+		graphqlUri: 'https://gateway.development.kiva.org/graphql',
+		gzipEnabled: true,
+		memcachedEnabled: false,
+		memcachedServers: '',
+	},
+});
--- a/config/local-prod.js
+++ b/config/local-prod.js
@@ -1,0 +1,21 @@
+import merge from 'deepmerge';
+import base from './k8s-local.js';
+
+export default merge(base, {
+	app: {
+		apolloBatching: false, // Remove this once batching is enabled in prod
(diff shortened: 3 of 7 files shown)


eddieferrer (approved):
very cool!

mcstover (approved):
Awesome!

dyersituations (approved):

PR #6826, opened 2026-04-13
refactor: remove opt in cookie logic

christian14b:
https://kiva.atlassian.net/browse/MP-2658

Removing cookie for opted-in logic since it's no longer necessary with latest changes

### DIFF
--- a/src/composables/useOptIn.js
+++ b/src/composables/useOptIn.js
@@ -1,9 +1,7 @@
 import logReadQueryError from '#src/util/logReadQueryError';
 import { gql } from 'graphql-tag';
 
-export const MAIL_UPDATES_OPT_COOKIE_NAME = 'kvemailopt';
-
-export default (apollo, cookieStore) => {
+export default apollo => {
 	const updateCommunicationSettings = async (lenderNews, loanUpdates, globalUnsubscribed) => {
 		try {
 			await apollo.mutate({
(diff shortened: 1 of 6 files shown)


src/composables/useOptIn.js
```diff
@@ -1,9 +1,7 @@
 import logReadQueryError from '#src/util/logReadQueryError';
 import { gql } from 'graphql-tag';
 
-export const MAIL_UPDATES_OPT_COOKIE_NAME = 'kvemailopt';
-
-export default (apollo, cookieStore) => {
+export default apollo => {
```
mcstover: Looks like cookieStore is still passed in the usage.

mcstover: Probably good to just keep it passed through incase some of the apollo methods expect it

src/composables/useOptIn.js
```diff
@@ -1,8 +1,6 @@
 import logReadQueryError from '#src/util/logReadQueryError';
 import { gql } from 'graphql-tag';
 
-export const MAIL_UPDATES_OPT_COOKIE_NAME = 'kvemailopt';
-
 export default (apollo, cookieStore) => {
```
dyersituations: `cookieStore` not defined so encountering linting error. If we want it here for future reference, use `_cookieStore`.

dyersituations (approved):

PR #6827, opened 2026-04-13
feat: fix on live loan redirect

tiarascoleman:
JIRA Ticket: https://kiva.atlassian.net/jira/software/projects/MP/boards/275?selectedIssue=MP-2602


roger-in-kiva (approved):

cristhianDt (approved):

christian14b (approved):

server/util/live-loan/live-loan-fetch.js
```diff
@@ -297,15 +297,15 @@ function parseSortString(sortString, sortOptions) {
...
+		// start parsing the string
 		getFilterArrays(sortString)
 			// remove any filter that isn't "sort"
 			.filter(([name]) => name === 'sort')
 			// return just the value of the sort option
 			.map(array => array?.[1])
 			// if the sort option value is valid, set it as the sort to be returned
 			.forEach(value => {
-				const sortOption = sortOptions.find(o => o?.toLowerCase() === value);
+				const sortOption = sortOptions.find(o => o?.toLowerCase() === value?.toLowerCase());
```
mcstover: We should be able to [add a check in the tests](https://github.com/kiva/ui/blob/main/test/unit/specs/server/live-loan-fetch.spec.js#L159) that fails without your update, then succeeds with your update...

dyersituations: Agreed, what specifically was the issue with the links? It looks like the casing matched.

dyersituations (changes requested):
See comment about use case and adding tests.

server/util/live-loan/live-loan-fetch.js
dyersituations: Above this `getFilterArrays` already lower cases the filters, so this is a no-op. The core issue will likely be elsewhere.

dyersituations (changes requested):
See latest comment about casing.

server/util/live-loan/live-loan-fetch.js
```diff
@@ -557,6 +557,38 @@ const shouldUseFLSS = async filterString => {
 	return true;
 };
 
+// Convert legacy filter string to modern query params
+// e.g., "gender_female,sort_amountLeft" -> { gender: 'female', sortBy: 'amountLeft' }
+export async function convertFiltersToQueryParams(filterString) {
```
dyersituations, line 562: What problem is this specifically solving? It looks like this code is being applied to the URL that the user gets redirected to when they click the live loan. The bug in the ticket is that the rendered live loan images are sometimes not matching the top results on the combo page.

Steps forward:
- Let us know how this solves this problem with examples (always good to add to PR desc)
- If this doesn't solve the issue in the ticket, test the live loans live current on dev + prod to see if the sort matches
- If the sort is currently matching, consider how the live loans caching might be changing how results compare between the live loans and the combo page.

tiarascoleman:
Closing this PR as this fix focused on the url rendering the correct/exact loans but the correct interpretation of the ticket is regarding rendering the correct loan images

PR #6835, opened 2026-04-15
feat: donation opt-in module ordering fix

tiarascoleman:
JIRA Ticket: https://kiva.atlassian.net/jira/software/projects/MP/boards/275?selectedIssue=MP-2670

This fix ensures that within the donation context, the opt-in module appears before the goals module


src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -22,6 +22,17 @@
 				@continue-clicked="handleContinue(NON_TIERED_BADGE)"
 				class="tw-mb-2.5"
 			/>
+			<!--  Donation opt-in module -->
+			<OptInModule
```
cristhianDt, line 26: @tiarascoleman could you please attach some evidences for both cases? thanks

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -59,12 +70,12 @@
...
 			<!-- End goal module variations -->
 			<OptInModule
-				v-if="showOptInModule"
+				v-if="showNonDonationOptInModule"
 				data-testid="opt-in-module"
 				:loans="loans"
 				:is-guest="isGuest"
 				:number-of-badges="numberOfBadges"
-				:only-donations="onlyDonations"
+				:only-donations="false"
```
christian14b: We can remove this since the default value is false

test/unit/specs/components/Thanks/ThanksPageSingleVersion.spec.js
```diff
@@ -114,7 +115,10 @@ function renderComponent(propsOverrides = {}) {
 			isGuest: false,
 			isOptedIn: true,
 			loans: [{ id: 1, unreservedAmount: '25.00' }],
-			receipt: { items: { values: [{ basketItemType: 'loan' }] } },
+			receipt: {
+				items: { values: [{ basketItemType: 'loan' }] },
+				totals: { itemTotal: '25.00', donationTotal: '0.00' },
```
dyersituations, line 120: Why was this change needed? It looks like the test also overrides this. Just want to make sure we aren't obscuring another test failure from the changes in this PR.

tiarascoleman: @dyersituations This change was needed because the original tests would fail due to the compiler not being able to distinct the two contexts (donationOptIn and nonDonationOptIn) due to them both using the showOptIn boolean.

dyersituations: Looks like the key here is the missing `totals` field, adding that allows the render versions to compare the donation total against it.

dyersituations (approved):

christian14b (approved):

PR #6843, opened 2026-04-16
fix: handle line breaks in team messages [CP-2515]

leoperry:
Not sure why I didn't notice this before...

Replace line breaks in team message text with `<br>`. If there is a preferred library method way of doing this, please let me know. Otherwise, classic regex replace it is!

### DIFF
--- a/src/components/LendingTeams/MyTeams/TeamMessageCard.vue
+++ b/src/components/LendingTeams/MyTeams/TeamMessageCard.vue
@@ -77,7 +77,7 @@ export default {
 			return `/team/${teamPublicId}/messages?msgID=${messageId}#msg_${messageId}`;
 		},
 		formatMessageBody(body, teamPublicId) {
-			const escaped = _escape(body);
+			const escaped = _escape(body).replace(/\r\n|\r|\n/g, '<br>');
 
 			// Replace embedded plain text message IDs (e.g. #123456) with direct links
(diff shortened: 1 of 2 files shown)


dyersituations (approved):
There's probably more complex deserialization of sorts that can be done, but this is a simple solution for this one off issue.

leoperry:
> There's probably more complex deserialization of sorts that can be done

Agreed. I'm happy to take any ideas or suggestions here. At least we are now at parity with the legacy page: https://github.com/kiva/kiva/blob/main/sites/www_kiva/client/src/views/components/LenderActivity/TeamMessageActivityView.js#L53-L59

(Plus anyone who managed to type line breaks on a very old Mac will now see those on the page!)

dyersituations:
> > There's probably more complex deserialization of sorts that can be done
> 
> Agreed. I'm happy to take any ideas or suggestions here. At least we are now at parity with the legacy page: https://github.com/kiva/kiva/blob/main/sites/www_kiva/client/src/views/components/LenderActivity/TeamMessageActivityView.js#L53-L59
> 
> (Plus anyone who managed to type line breaks on a very old Mac will now see those on the page!)

If this is the only formatting translation we need, I think this simple solution is just fine.

mcstover (approved):
Def good with the replace for now. I'm not familiar off hand how these are getting saved during input so they end up as part of the content we store and retrieve from the api...

PR #6845, opened 2026-04-17
feat: goal custom amount validation

roger-in-kiva:
- validation to goal custom amount input
- https://kiva.atlassian.net/browse/MP-2641

Invalid:
Screenshot 2026-04-17 at 9 19 27 a m
Screenshot 2026-04-17 at 9 20 01 a m

Valid:
Screenshot 2026-04-17 at 9 20 57 a m
Screenshot 2026-04-17 at 9 21 14 a m

### DIFF
--- a/src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
+++ b/src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
@@ -1,0 +1,65 @@
+<template>
+	<div class="tw-relative">
+		<input
+			type="number"
+			name="customGoalAmount"
+			class="tw-rounded-sm tw-border-2 tw-border-gray-400 tw-pl-1.5 tw-py-0.5 tw-pr-4 tw-w-full
+                tw-ring-inset focus:tw-outline-none focus:tw-ring-0 focus:tw-border-gray-400 custom-input"
+			:class="{
+				'focus:!tw-border-brand-400': validCustomAmount,
+				'!tw-border-desert-rose': validCustomAmount === false,
+				'tw-min-h-5': isMobile,
+			}"
+			placeholder="Add number"
+			@input="validateCustomAmount"
+			autofocus
+		>
+		<KvMaterialIcon
+			v-if="validCustomAmount !== null"
+			:icon="validCustomAmount ? mdiCheckCircle : mdiAlertCircle"
+			class="tw-w-2 tw-h-2 tw-absolute tw-top-1/2 tw-right-1.5 tw--translate-y-1/2"
+			:class="{
+				'!tw-text-brand-650': validCustomAmount,
+				'!tw-text-desert-rose': validCustomAmount === false,
+			}"
+		/>
+	</div>
+</template>
+
+<script setup>
+import { mdiCheckCircle, mdiAlertCircle } from '@mdi/js';
+import { KvMaterialIcon } from '@kiva/kv-components';
+
+const emit = defineEmits([
+	'validate-custom-amount',
+]);
+
+defineProps({
+	validCustomAmount: {
+		type: Boolean,
+		required: null
+	},
+	isMobile: {
+		type: Boolean,
+		required: false
+	}
+});
+
+const validateCustomAmount = event => {
+	emit('validate-custom-amount', event.target.value);
+};
+</script>
+
+<style scoped>
+/* Chrome, Safari, Edge, Opera */
+.custom-input::-webkit-outer-spin-button,
+.custom-input::-webkit-inner-spin-button {
+  appearance: none;
+  margin: 0;
+}
+
+/* Firefox */
+.custom-input[type=number] {
+  appearance: textfield;
+}
+</style>
--- a/src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
+++ b/src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
@@ -56,12 +56,11 @@ const validateCustomAmount = event => {
 };
 </script>
 
-<style scoped>
+<style lang="postcss" scoped>
 /* Chrome, Safari, Edge, Opera */
 .custom-input::-webkit-outer-spin-button,
 .custom-input::-webkit-inner-spin-button {
-  appearance: none;
-  margin: 0;
+    @apply tw-appearance-none tw-m-0
 }
 
 /* Firefox */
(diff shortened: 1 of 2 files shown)


src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -512,13 +522,21 @@ const localGoalProgressPercentage = computed(() => {
 	return Math.min(Math.round((loansThisYear.value / effectiveGoalLoans.value) * 100), 100);
 });
 
+const minCustomAmount = computed(() => {
+	return loansThisYear.value > 1 ? loansThisYear.value + 1 : 2;
+});
+
+const warningCustomAmountText = computed(() => `Your goal must be a valid number above ${minCustomAmount.value - 1} loans`); // eslint-disable-line max-len
```
mcstover, line 532: It seems odd that we don't allow a user to set a goal at the same level or lower than prior goals... suppose they can just cancel out instead and skip setting any...

roger-in-kiva: That's a product choice, this is just to avoid setting a completed goal.

mcstover: ok yea if it's just editing an existing goal that makes sense

roger-in-kiva: For sure, we are able to edit the goal at mykiva 👍

dyersituations: Yeah this min is based on `loansThisYear`, so the goal isn't autocompleted. Goals are still based on yearly lending, rather than a specific start date. Good question though!

src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
```diff
@@ -0,0 +1,71 @@
...
+	isMobile: {
+		type: Boolean,
+		required: false
+	}
+});
+
+const customAmount = ref('');
+
+const validateCustomAmount = event => {
+	customAmount.value = event.target.value.replace(/[^0-9]/g, '');
```
dyersituations, line 63: Do we need optional chaining here?

src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
```diff
@@ -0,0 +1,71 @@
...
+/* Chrome, Safari, Edge, Opera */
+.custom-input::-webkit-outer-spin-button,
+.custom-input::-webkit-inner-spin-button {
+  appearance: none;
+  margin: 0;
+}
+
+/* Firefox */
+.custom-input[type=number] {
+  appearance: textfield;
```
dyersituations, line 85: Can these 3 style entries be replaced with TW?

src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
roger-in-kiva: This is defined by html element by itself.

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -80,68 +80,76 @@
...
+									<span
+										class="tw-font-medium lg:tw-hidden"
+										:class="{ 'tw-text-small': isCustomIndex }"
+									>
+										Set your number
+									</span>
+									<GoalCustomAmountInput
+										v-if="isCustomIndex"
+										class="lg:tw-hidden"
+										is-mobile
```
dyersituations, line 114: Why is this always passed as `true`?

roger-in-kiva: This is mobile input version that needs a min height:
`'tw-min-h-5': isMobile,`

src/components/MyKiva/GoalSetting/GoalCustomAmountInput.vue
roger-in-kiva: Not this specific one, others are replaced.

src/components/MyKiva/GoalSetting/GoalSelector.vue
dyersituations: So this specific component is only rendered for mobile?

dyersituations:
After editing the category, the modal gets in a weird state:
- Message about lending in current year disappears
- Validation disappears
image

src/components/MyKiva/GoalSetting/GoalSelector.vue
roger-in-kiva: Yup

roger-in-kiva: Well, hidden on large screens
`class="lg:tw-hidden"`

roger-in-kiva:
> After editing the category, the modal gets in a weird state:
> 
> * Message about lending in current year disappears
> * Validation disappears
> 
> image

That's b/c you don't have a current year progress, so min is 2 loans per goal for custom amount.

dyersituations:
> That's b/c you don't have a current year progress, so min is 2 loans per goal for custom amount.

So the bug then is that the custom input isn't revalidated after category change, and you can't click set:

image

dyersituations:
Maybe we just reset the user selection on category change?

roger-in-kiva:
> Maybe we just reset the user selection on category change?

That's a good question, as current behavior is to keep selected amount option as category changes, I'll ask.

dyersituations:
> > Maybe we just reset the user selection on category change?
> 
> That's a good question, as current behavior is to keep selected amount option as category changes, I'll ask.

If we reset on category change, it would fix 2 bugs:
1. Number is revalidated so set can get incorrectly disabled
2. Number field doesn't get auto-focused like when you open the custom option

I think just reseting might be the most elegant solution here (only if they chose custom before changing category).

roger-in-kiva:
> > > Maybe we just reset the user selection on category change?
> > 
> > 
> > That's a good question, as current behavior is to keep selected amount option as category changes, I'll ask.
> 
> If we reset on category change, it would fix 2 bugs:
> 
> 1. Number is revalidated so set can get incorrectly disabled
> 2. Number field doesn't get auto-focused like when you open the custom option
> 
> I think just reseting might be the most elegant solution here (only if they chose custom before changing category).

Yeah, makes sense, also wondering if they want to reset custom selection, like default one.

dyersituations (approved):

PR #6846, opened 2026-04-17
fix: add completed goal logic to my kiva page

christian14b:
https://kiva.atlassian.net/browse/MP-2698

image


roger-in-kiva:
Wondering if we want to edit as completed.

christian14b:
> Wondering if we want to edit as completed.

I can do a follow-up, seems like there's a lot of code that can be simplified and reused

roger-in-kiva (approved):

cristhianDt (approved):

roger-in-kiva:
> > Wondering if we want to edit as completed.
> 
> I can do a follow-up, seems like there's a lot of code that can be simplified and reused

Pls ask Lauren if we want to allow editing completed goals.

src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -438,6 +439,7 @@ export default {
 			tieredAchievements: this.currentYearTieredAchievements,
 			transactions: this.transactions,
 		});
+		await this.checkCompletedGoal({ category: 'portfolio' });
```
dyersituations: Isn't this already called in lending stats on mount?

dyersituations (changes requested):

src/pages/MyKiva/MyKivaPage.vue
christian14b: Yes, `LendingStats` calls it on mount too, but that can run before goal data has finished loading on the parent. Running it on `MyKivaPage` right after `loadGoalData` makes sure we use the same data. Happy to drop the call in LendingStats so we only do this once

dyersituations: Ideally this only runs once, so definitely look into how this could be simplified, and if there are issues with the simplification.

src/composables/useGoalData.js
```diff
@@ -695,6 +695,7 @@ export default function useGoalData({ apollo } = {}) {
 		tieredAchievements = [], // Tiered achievements from achievement service to calculate fresh progress
 		transactions = [], // User transactions to get purchase dates for year filtering
 		fetchPolicy = 'cache-first', // Apollo fetch policy for loading preferences
+		checkMyKivaCompletedGoalAfterLoad = false, // Flag to check if completed goal should be checked after load
```
dyersituations, line 698: Why is this flag needed? Does this method get called from multiple code paths? Does it need to be called from multiple places?

christian14b: Yes, `loadGoalData` is used in different parts, so the flag keeps this behavior scoped to MyKiva only.

`loadGoalData` is shared by 
- MyKiva, 
- Thanks page
- Basket/ATB 
- Goal-setting modal

Only MyKiva needs the run completion check for portfolio behavior. Without the flag, we’d add side effects to the other components so this act as a guard.

I'll still be validating if everything looks correct in those different places, but TY page and my kiva are looking good so far

dyersituations (approved):

PR #6852, opened 2026-04-21
fix: next year goal card stories and progress ring values

christian14b:
https://kiva.atlassian.net/browse/MP-2700

FIxing:
- Stories were broken
- Ring values sizes for big numbers (Added new cases)

Images:
image
image

### DIFF
--- a/.storybook/stories/NextYearGoalCard.stories.js
+++ b/.storybook/stories/NextYearGoalCard.stories.js
@@ -1,3 +1,4 @@
+import { computed, provide } from 'vue';
 import MyKivaNextYearGoalCard from '#src/components/MyKiva/NextYearGoalCard.vue';
 
 export default {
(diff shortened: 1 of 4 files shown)


roger-in-kiva:
Sorry if I missed in the ticket, but looks like this card should be updated as well.
Screenshot 2026-04-21 at 2 22 54 p m

.storybook/stories/NextYearGoalCard.stories.js
```diff
@@ -6,75 +7,123 @@ export default {
...
 export const UserGoalWithHalfProgress = story({
-  prevYearLoans: 8,
-  userGoal: {
-    target: 10,
-    category: 'ID_WOMENS_EQUALITY',
-  },
-  goalProgress: 5,
+	prevYearLoans: 8,
+	userGoal: {
+		target: 10,
```
cristhianDt, line 80: nice fixing this indents 👍

.storybook/stories/NextYearGoalCard.stories.js
```diff
@@ -6,75 +7,123 @@ export default {
...
-  prevYearLoans: 8,
-  userGoal: {
-    target: 10,
-    category: 'ID_WOMENS_EQUALITY',
-  },
-  goalProgress: 10,
+	prevYearLoans: 8,
+	userGoal: {
+		target: 10,
+		category: 'ID_WOMENS_EQUALITY',
```
cristhianDt, line 99: (optional) we can define a global constant for this value

cristhianDt (approved):

roger-in-kiva:
Looks like we need more padding at top for this case to show it vertically centered.
Screenshot 2026-04-21 at 2 27 28 p m

christian14b:
> Sorry if I missed in the ticket, but looks like this card should be updated as well. Screenshot 2026-04-21 at 2 22 54 p m

Thanks for the heads-up! This is trickier since progress are a computed property. https://github.com/kiva/ui/blob/003833930955145113e8d46f16449f950e42a52e/src/components/MyKiva/MyKivaProgressCard.vue#L149, but I'll update the ticket and fix the stories here too

dyersituations (approved):

PR #6858, opened 2026-04-22
feat: change bundle size based on balance

dyersituations:
https://kiva.atlassian.net/browse/MP-2713

An optional `balance` parameter that allows live loan bundles to be up to 6, existing live loan usage untouched functionally (and cache-wise).

### DIFF
--- a/server/util/live-loan/bundle-size.js
+++ b/server/util/live-loan/bundle-size.js
@@ -1,0 +1,39 @@
+export const DEFAULT_BUNDLE_COUNT = 4;
+
+const REDIRECT_PATH = '/lend/filter';
+const MAX_BALANCE = 150;
+const TIER_STEP = 25;
+const MIN_LOANS = 1;
+const MAX_LOANS = 6;
+
+/**
+ * Resolves how many loans should be included in a live-loan email bundle, based on the
+ * recipient's account balance. Tiers: $25.01–$50 → 2 loans, $50.01–$75 → 3, $75.01–$100 → 4,
+ * $100.01–$125 → 5, $125.01–$150 → 6. Balance ≤ $25 → 1 loan. Balance > $150 → redirect to
+ * `/lend/filter` (combo page) so the user can curate a larger bundle themselves.
+ *
+ * Falsy, malformed, non-string/non-number, or negative input returns the 4-loan default so
+ * that pre-MP-2713 email links without a `?balance=` parameter keep their original behavior.
+ *
+ * @param {string|number} balanceInput Raw `?balance=` query value from the request.
+ * @returns {{ count: number } | { redirect: string }} Either a count in [1, 6] or a redirect path.
+ */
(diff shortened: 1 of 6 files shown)


server/util/live-loan/bundle-size.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function resolveBundleSize(balanceInput) {
+	if (typeof balanceInput !== 'string' && typeof balanceInput !== 'number') {
+		return { count: DEFAULT_BUNDLE_COUNT };
+	}
+	const normalized = typeof balanceInput === 'string' ? balanceInput.trim() : balanceInput;
+	if (normalized === '') {
+		return { count: DEFAULT_BUNDLE_COUNT };
+	}
+	const balance = Number(normalized);
```
roger-in-kiva, line 29: Do we want to check if integer at some point?

dyersituations: Does it matter if there are decimals? The comparisons should still be valid here in the method.

christian14b (approved):
Nice!

roger-in-kiva (approved):

cristhianDt (approved):

tiarascoleman (approved):

mcstover (approved):
Is there any concern with us not getting back the total of `count` during the primary loaf fetch?

dyersituations:
> Is there any concern with us not getting back the total of `count` during the primary loaf fetch?

@mcstover Where specifically? There should be a fallback `count` everywhere.

When loading the loans? Like the loans returned are less than `count`?

mcstover:
> > Is there any concern with us not getting back the total of `count` during the primary loaf fetch?
> 
> @mcstover Where specifically? There should be a fallback `count` everywhere.
> 
> When loading the loans? Like the loans returned are less than `count`?

Yea I was thinking if the query didn't result in the full `count`. Prob not an issue with the recommended loans query but I didn't know for sure if it always fulfilled the requested count.

dyersituations:
> > > Is there any concern with us not getting back the total of `count` during the primary loaf fetch?
> > 
> > 
> > @mcstover Where specifically? There should be a fallback `count` everywhere.
> > When loading the loans? Like the loans returned are less than `count`?
> 
> Yea I was thinking if the query didn't result in the full `count`. Prob not an issue with the recommended loans query but I didn't know for sure if it always fulfilled the requested count.

Good question. From a technical standpoint, all's good. We only add the loan IDs returned:
https://github.com/kiva/ui/blob/81dc9e18803f3135925f06126b44ca771d766d26/server/live-loan-router.js#L109

And we catch if no loan IDs returned and redirect:
https://github.com/kiva/ui/blob/81dc9e18803f3135925f06126b44ca771d766d26/server/live-loan-router.js#L114

From a product standpoint, maybe this isn't super, but we have a good fallback in place.

PR #6859, opened 2026-04-22
feat: removing lending routes as ported to cps combo page

roger-in-kiva:
- removing lending routes as ported to cps combo page
- https://kiva.atlassian.net/browse/MP-2683

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -291,24 +291,8 @@ export default [
 		redirect: '/lend-by-category',
 	},
 	{
-		path: '/lend-by-category',
-		component: () => import('#src/pages/LoanFinding/LoanFinding'),
-		meta: {
-			unbounceEmailCapture: true,
-		},
-	},
-	{
 		path: '/lend-by-category/recommended-by-lenders',
 		redirect: '/lend-by-category'
-	},
-	{
-		path: '/lend-by-category/:category',
-		component: () => import('#src/pages/Lend/LoanChannelCategoryPage'),
-		meta: {
-			excludeFromStaticSitemap: true,
-			unbounceEmailCapture: true,
-			unbouncePopUp: true,
-		},
 	},
 	{
 		path: '/lend-beta/:id',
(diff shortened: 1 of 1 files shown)


dyersituations:
@roger-in-kiva we don't want to do this yet. We can test the pages on dev on CPS via flux, leaving these in place.

roger-in-kiva:
> @roger-in-kiva we don't want to do this yet. We can test the pages on dev on CPS via flux, leaving these in place.

Oh right, let's keep this open so.

src/router/routes.js
```diff
@@ -290,26 +290,10 @@ export default [
 		path: '/lend-by-category/blackrock',
 		redirect: '/lend-by-category',
 	},
-	{
-		path: '/lend-by-category',
-		component: () => import('#src/pages/LoanFinding/LoanFinding'),
```
mcstover, line 295: I think both of these `/lend-by-cat` routes should become redirects. This one is easy, not sure exactly how the category specific endpoint below will translate to the new page.

src/router/routes.js
```diff
@@ -346,14 +330,6 @@ export default [
 			unbouncePopUp: true,
 		}
 	},
-	{
-		name: 'lend-filter',
-		path: '/lend/filter',
-		component: () => import('#src/pages/Lend/LoanSearchPage'),
```
mcstover, line 352: If it's possible to set this as a redirect and trigger a full page refresh, that would probably be good to set here. That way we'd be back out at the ingress and flux configs would handle pushing this to cps instead.

roger-in-kiva: As we are currently having same route on cps, not sure what you mean.
From `/lend/filter` to `/lend/filter`?

My guess is adding these routes to cps flux and removing from ui flux, like this:
https://github.com/kiva/flux/compare/MP-2683_combo_page_routes_update?expand=1

Open if you have other option in mind.

mcstover: For this one, I was just thinking about how to handle internal to `ui` links routing to `lend/filter`. We want to make sure it causes a full page navigation. Probably just removing it like you have here will do that though ;)

PR #6870, opened 2026-04-29
fix: load yearly loan count when setting a goal in ty page

christian14b:
https://kiva.atlassian.net/browse/MP-2810

For support all category, fixing issue where goal progress is zero, even though the user have just checkout a loan, causing 0/X loans when setting a goal in TY page.

### DIFF
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -606,6 +606,12 @@ export default function useGoalData({ apollo } = {}) {
 			// For ID_SUPPORT_ALL, use totalLoanCount since it tracks total loans, not category-specific progress
 			let loanTotalAtStart;
 			if (updates.category === ID_SUPPORT_ALL) {
+				// When setting a new goal in Thanks page, we don't have a userGoal object, so we have not loaded support-all loan count on loadGoalData.
+				// We need to get the total loan count in this step.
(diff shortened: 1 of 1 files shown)


tiarascoleman (approved):
Linting errors still need to be fixed! Approving for code

src/composables/useGoalData.js
```diff
@@ -606,6 +606,13 @@ export default function useGoalData({ apollo } = {}) {
 			// For ID_SUPPORT_ALL, use totalLoanCount since it tracks total loans, not category-specific progress
 			let loanTotalAtStart;
 			if (updates.category === ID_SUPPORT_ALL) {
+				// When setting a new goal in Thanks page, we don't have a userGoal object
+				// so we have not loaded support-all loan count on loadGoalData.
+				// We need to get the total loan count in this step.
+				if (!userGoal.value?.category) {
+					const stats = await getSupportAllLoanCountByYear(GOALS_CURRENT_YEAR, 'network-only');
```
roger-in-kiva: Wondering if `stats` makes sense with loan count.

christian14b: true, I don't think I neither need to use another variable. good catch

roger-in-kiva (approved):

dyersituations (approved):

PR #6874, opened 2026-04-29
fix: condition to show loan question prompt

christian14b:
image
image

### DIFF
--- a/src/components/MyKiva/GoalSetting/GoalSelector.vue
+++ b/src/components/MyKiva/GoalSetting/GoalSelector.vue
@@ -428,16 +428,16 @@ const customGoalAmount = ref(null);
 const validCustomAmount = ref(null);
 const customGoalAmountError = ref('');
 
-const showLoanQuestionPrompt = computed(() => {
-	return loansLastYear.value > 0 || props.selectedCategoryId === ID_WOMENS_EQUALITY;
-});
-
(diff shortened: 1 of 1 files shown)


src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -428,6 +428,10 @@ const customGoalAmount = ref(null);
 const validCustomAmount = ref(null);
 const customGoalAmountError = ref('');
 
+const showLoanQuestionPrompt = computed(() => {
+	return loansLastYear.value > 0 || props.selectedCategoryId === ID_WOMENS_EQUALITY;
```
dyersituations: Why is this special cases to just women? Usually all of the categories other than support all operate the same.

dyersituations: Also linting error here

christian14b: It's because of this condition in title: 
https://github.com/kiva/ui/blob/e2a91958434d4a07976caf6355a56a3cd8a470dd/src/components/MyKiva/GoalSetting/GoalSelector.vue#L483

dyersituations (approved):
Approved pending linting fix.

cristhianDt (approved):

PR #6876, opened 2026-04-30
feat: mp-2808-loaders-next-steps

infante-jaime:
Added loader coverage for the missing cards in both places:

Next Steps page: lending-focused cards now respect the same loading condition, so missing cards show skeletons until data is ready.
MyKiva dashboard: applied the same loader condition, so those missing lending-focused cards also render as placeholders during load.
So both views now use the same loading behavior instead of showing partial card sets.

### DIFF
--- a/src/pages/MyKiva/MyKivaNextStepsContent.vue
+++ b/src/pages/MyKiva/MyKivaNextStepsContent.vue
@@ -21,7 +21,7 @@
 				}"
 			>
 				<JourneyCardCarousel
-					v-if="showLendingNextStepsCards"
+					v-if="showLendingNextStepsCards && !goalProgressLoading"
 					class="carousel tw-w-full"
 					user-in-homepage
 					in-lending-stats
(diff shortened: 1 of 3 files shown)


src/pages/MyKiva/MyKivaNextStepsContent.vue
```diff
@@ -46,7 +46,7 @@
 					@open-goal-modal="openGoalModal($event)"
 					@open-impact-insight-modal="showImpactInsightsModal = true"
 				/>
-				<template v-else-if="showRegionExperienceInFirstRow">
+				<template v-else-if="showRegionExperienceInFirstRow && !showLendingNextStepsCards">
```
dyersituations, line 49: Why is `!showLendingNextStepsCards` needed here?

dyersituations: Oh I see, it's `false` here, nevermind.

dyersituations (approved):

PR #6877, opened 2026-04-30
feat: [MP-2759] - edge case where user dont see an event day badge

cristhianDt:
# MP-2759

## Summary

two possible edge cases here:

1. postCheckoutAchievementsQuery may not identify earth day as a newly achieved badge if the before/after progress state does not capture the timing correctly. in that scenario, the issue would be in how we determine that the achievement was newly completed during checkout. the UI change I’m making would not solve that by itself.
2. the UI timing/reactivity case I’m trying to cover right now: if postCheckoutAchievementsQuery eventually provides the achieved badge IDs, but BadgeMilestone.vue has already calculated its displayed badge using an earlier badgeData/contentful state, then the component may not recompute when badgeAchievedIds changes. in that case, it can keep showing equity instead of the event badge.

so this PR is only intended to cover the second case. the proposed change is to make BadgeMilestone.vue also watch badgeAchievedIds, so the badge display is recalculated whenever the achieved ids change, not only when badgeData or contentful data changes

https://kiva.atlassian.net/browse/MP-2759


src/components/Thanks/SingleVersion/BadgeMilestone.vue
```diff
@@ -213,7 +213,11 @@ watch(() => badgeContentfulData.value, () => {
 	}
 });
 
-watch([() => badgeData.value, () => badgeContentfulData.value], ([newBadgeData]) => {
+watch([
+	() => badgeData.value,
+	() => badgeContentfulData.value,
+	() => props.badgeAchievedIds.join('|'),
```
dyersituations, line 219: Where/how are these IDs loaded again after initial page load?

cristhianDt: yes, I think you’re right that there may be another edge case around the before/after progress calculation (following our conversation in slack)

I see two possible edge cases here:

1. the case you’re pointing out: `postCheckoutAchievementsQuery` may not identify earth day as a newly achieved badge if the before/after progress state does not capture the timing correctly. in that scenario, the issue would be in how we determine that the achievement was newly completed during checkout. the UI change I’m making would not solve that by itself.

2. the UI timing/reactivity case I’m trying to cover right now: if `postCheckoutAchievementsQuery` eventually provides the achieved badge IDs, but `BadgeMilestone.vue` has already calculated its displayed badge using an earlier badgeData/contentful state, then the component may not recompute when badgeAchievedIds changes (here where we're merging the fresh user progress with `badgeContentfulData` to refresh the badgesData https://github.com/kiva/ui/blob/039cfb575e83d0ac769fb1fda71fb7d5e640b5aa/src/composables/useBadgeData.js#L494 that could change the ids in BadgeMilestone). in that case, it can keep showing equity instead of the event badge.

right now my PR is only intended to cover the second case. the proposed change is to make BadgeMilestone.vue also watch badgeAchievedIds, so the badge display is recalculated whenever the achieved ids change, not only when badgeData or contentful data changes.

**I'm checking your slack comment more closely to see how MyKiva handles the before/after progress flow and whether we should apply a similar approach here**

cristhianDt: @dyersituations for this PR I’m limiting scope to the UI timing/reactivity issue (recompute when achieved IDs update), which is low-risk and test-covered.

I suggest we handle the before/after reconciliation path (the other possible edge case -point No.1-) in a follow-up PR, since that likely requires broader logic changes and separate validation

thoughts? or agree?

dyersituations: Or we just try this and see what happens with the next campaign? Or do you think the other fix is more critical?

cristhianDt: yes, it's a good idea, if we could test it in development beforehand, that would be fantastic, haha, but there's no problem waiting for the next campaign; and to answer your question, I don't think it would be more critical or we'd have more open issues

let me move the PR as ready

dyersituations: Right, you can enable an achievement on dev directly in dynamo for a few days to test. Please do that once it's on dev.

dyersituations (approved):

PR #6879, opened 2026-04-30
fix: show new copy updates only in ty page goal entrypoint

christian14b:
### My Kiva
image

image

### Thank you page
image
image

### My Kiva
image
image

### Thank you page
image
image

### DIFF
--- a/src/components/GoalSetting/GoalSettingContainer.vue
+++ b/src/components/GoalSetting/GoalSettingContainer.vue
@@ -67,6 +67,7 @@
 					:loading-current-year="loadingCurrentYear"
 					:goal-progress="goalProgress"
 					:is-goal-completed="isGoalCompleted"
+					show-original-copy-version
 					@set-goal-target="setTarget($event)"
 					@set-goal="setGoal($event)"
 					@update-goal="updateGoal($event)"
--- a/src/components/GoalSetting/GoalSettingContainer.vue
+++ b/src/components/GoalSetting/GoalSettingContainer.vue
@@ -67,7 +67,7 @@
 					:loading-current-year="loadingCurrentYear"
 					:goal-progress="goalProgress"
 					:is-goal-completed="isGoalCompleted"
-					show-original-copy-version
+					show-goal-value-props-copy
 					@set-goal-target="setTarget($event)"
 					@set-goal="setGoal($event)"
 					@update-goal="updateGoal($event)"
--- a/src/components/MyKiva/GoalSetting/GoalSelector.vue
+++ b/src/components/MyKiva/GoalSetting/GoalSelector.vue
@@ -57,7 +57,7 @@
 			</p>
 
 			<p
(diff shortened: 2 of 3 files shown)


src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -67,6 +67,7 @@
 					:loading-current-year="loadingCurrentYear"
 					:goal-progress="goalProgress"
 					:is-goal-completed="isGoalCompleted"
+					show-original-copy-version
```
dyersituations: Let's rename this to be related to the goal value props, in case we keep both versions longterm.

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -55,6 +55,12 @@
 				How many loans will you make this year?
 			</p>
 
+			<p
+				v-if="showOriginalCopyVersion"
+				v-html="subtitleText"
+				class="tw-text-base lg:tw-text-subhead tw-my-1.5 lg:tw-mb-1 lg:tw-mt-2 tw-text-center">
```
dyersituations: New space needed before angle bracket on these lines (linting).

dyersituations:
@christian14b see test failure for more linting details

src/components/MyKiva/GoalSetting/GoalSelector.vue
christian14b: I don't why my extension is not auto fixing this.. I'll need to take a look at it. Good catch

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -395,9 +395,9 @@ const props = defineProps({
 		default: false,
 	},
 	/**
-	 * Flag to indicate if the original copy version should be shown
+	 * Flag to indicate if the goal value props copy version should be shown
 	 */
-	showOriginalCopyVersion: {
+	showGoalValuePropsCopy: {
```
dyersituations, line 400: Is this naming accurate? What was original? Based on naming this seems like the opposite now?

christian14b: New copy was under the Custom goal amount test. Does the current copy in my kiva and goal setting page isn't for goal value props?

christian14b: I called "original" just to make a difference between these new updates and the copy in the other pages

dyersituations: So is calling this "goal value prop" correct then?

christian14b: Yes, copy in My Kiva and Goal Setting pages are linked to the Goal Value Props figma file

dyersituations (approved):

PR #6884, opened 2026-05-05
feat: goal category loans data

roger-in-kiva:
- function added to fetch recommended loans by goal category
- https://kiva.atlassian.net/browse/MP-2740


src/composables/useGoalData.js
```diff
@@ -49,6 +50,15 @@ const GOAL_1_DISPLAY_MAP = {
 	[ID_WOMENS_EQUALITY]: 'woman',
 };
 
+// Filters with loanSearchState format to be used directly in graphql queries depending on the category
+const FLSS_FILTERS_BY_GOAL = {
```
dyersituations, line 54: Could we use the filters defined in `useBadgeData`?

roger-in-kiva: Seems similar but not, these are the same we are using on combo page for goal categories.

dyersituations: Should be the same, since goals use the badge filters. But having them here isn't an issue either.

dyersituations (approved):

src/composables/useGoalData.js
roger-in-kiva: Mismatch are these cases:

```
 [ID_US_ECONOMIC_EQUALITY]: { country: ['PR', 'US'] },
 [ID_REFUGEE_EQUALITY]: { attribute: ['28'] },
```

So, we should fix one of them, and also, matching loans with our lending page.

PR #6888, opened 2026-05-08
fix: origin added to goal recommended loan data method

roger-in-kiva:
- origin added to goal recommended loan data method
- https://kiva.atlassian.net/browse/MP-2740


dyersituations (approved):
Good catch!

cristhianDt (approved):

PR #6893, opened 2026-05-11
feat: linkify My Teams message links using linkify-string [CP-2620]

leoperry:
https://kiva.atlassian.net/browse/CP-2620

Walked through this with Soroush at Kiva Connect, and feedback was generally very positive!

The only issue we noticed was that messages with long URL parameters failed to line break, and due to overflow effectively were visually truncated to the boundary box. We solve that here by following a suggestion by Eddie and adding `tw-break-words` to the wrapper paragraph.

We also noticed that the URLs on this page were not being rendered as links, which is consistent with the old stack My Teams page. However, the old stack team message board pages _do_ actually linkify URLs.

To sweeten the transition to the new page by introducing a minor nicety, I decided to utilize `linkify-string` to handle the escaping, link rendering, etc. This ultimately results in a simplification of the code overall since linkify handles this as a one-stop shop.

I noticed that CPS and rtp-nuxt both already utilize this dependency, although it is new to UI. So it seemed fine to me to add the new dependency, but please let me know if there is an issue with that! It's definitely not strictly necessary but I think it's a reasonable feature add for users of this page that brings it closer to parity with the full team message board page.

Fine detail on embedded message pattern: The general embedded message pattern of `#12345` can run afoul of HTML entities as well as certain URLs that may include this pattern. However the conventional usages of these embedded message IDs are always padded with whitespace (as in `re @LenderID #12345`). Adding the whitespace assumption simplifies our regex strategy, although admittedly it is more restrictive since usages like e.g. `(#12345)` would not longer render as a link. (I haven't seen that usage in practice although it's possible there are some power users who do stuff like that.) I erred toward the side of simplicity here as that was a general theme that emerged here.

## Before

Screenshot 2026-05-11 at 11 07 10 AM

## After

Screenshot 2026-05-11 at 11 07 17 AM


src/components/LendingTeams/MyTeams/TeamMessageCard.vue
```diff
@@ -77,13 +93,12 @@ export default {
...
-			const escaped = _escape(body).replace(/\r\n|\r|\n/g, '<br>');
+			const linkedBody = linkifyStr(body || '', LINKIFY_OPTIONS);
 
 			// Replace embedded plain text message IDs (e.g. #123456) with direct links
-			// Use negative lookbehind / lookahead to exclude HTML entities that may have been escaped (e.g., &#39;)
-			return escaped.replace(/(?<!&)#(\d+)(?!;)/g, (match, msgId) => {
+			return linkedBody.replace(EMBEDDED_MESSAGE_PATTERN, (match, msgId) => {
 				const url = this.getDirectMessageUrl(teamPublicId, msgId);
-				return `<a href="${url}" class="tw-text-link hover:tw-underline">${match}</a>`;
+				return ` <a href="${url}" class="${MESSAGE_BODY_LINK_CLASS}">#${msgId}</a> `;
```
dyersituations, line 99: Do we want `target="_blank"` and/or other configs on this links?

package.json
```diff
@@ -104,6 +104,8 @@
 		"js-cookie": "^3.0.1",
 		"jsonwebtoken": "^9.0.2",
 		"libphonenumber-js": "^1.11.4",
+		"linkify-string": "^4.3.2",
+		"linkifyjs": "^4.3.2",
```
dyersituations, line 108: I see this isn't imported but is it just needed as a dependency for `linkify-string`?

leoperry: That's correct. I followed the docs on https://linkify.js.org/docs/linkify-string.html which explicitly installs both.

src/components/LendingTeams/MyTeams/TeamMessageCard.vue
leoperry: Good question. I don't think we need to target blank with these deep links, as generally they will place the linked message in deeper context on the team message board -- so this makes the most sense to me.

My question would be whether we need the blank target & `rel="noopener noreferer"` on the more general linkify strategy? The old stack team message board behavior doesn't provide any such adornment, so if we want to favor consistency in link behavior I'd lean toward removing those from that context.

dyersituations: All links in the old board just linked/navigated in the current browser window? If so, I suppose we could indeed just keep it consistent.

leoperry: Yeah, to clarify: links rendered on the old stack team message board page link directly, whether it's an external or internal link. Because we're not modifying those pages any time soon, I agree it probably makes the most sense to keep the link behavior consistent.

Note though that (non-embedded-message-ID) links on the current / old My Teams page were never even rendered as links at all, so there's no direct comparison there. However being consistent with the single-team message board link behavior seems like the right way to go with this.

dyersituations (approved):

eddieferrer (approved):

leoperry:
@dyersituations @eddieferrer -- thanks for the reviews. I think I've run into the interminable but required coveralls check before, which prevents me from merging. Is there a way to resolve that, or would one of you mind merging if that's the best route forward?

dyersituations:
> @dyersituations @eddieferrer -- thanks for the reviews. I think I've run into the interminable but required coveralls check before, which prevents me from merging. Is there a way to resolve that, or would one of you mind merging if that's the best route forward?

We really need to kick coveralls, but for now, I have permissions...

PR #6898, opened 2026-05-13
fix: add logic to get loan amount from card

christian14b:
- Small padding tweak needed for lightbox
- Adding functionality to get loan amount from card

### DIFF
--- a/src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContent.vue
+++ b/src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContent.vue
@@ -19,7 +19,12 @@
 </template>
 
 <script setup>
-import { computed, useAttrs, ref, defineExpose } from 'vue';
+import {
+	computed,
+	useAttrs,
+	ref,
+	defineExpose,
+} from 'vue';
 import { KvCompactLoanCard } from '@kiva/kv-components';
 
 defineOptions({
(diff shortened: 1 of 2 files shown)


cristhianDt (approved):

src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContent.vue
```diff
@@ -50,4 +57,12 @@ const cardAttrs = computed(() => {
 	const { class: _c, style: _s, ...rest } = attrs;
 	return rest;
 });
+
+const getSelectedAmount = () => {
+	return kvCompactLoanCard.value?.selectedAmount;
+};
+
+defineExpose({
+	getSelectedAmount,
```
dyersituations, line 66: Should we expose this or is an emit needed?

christian14b: With this definition it will be exposed. I'm getting that value by using 

`const lendAmount = recommendLoanForGoalContentRef.value?.getSelectedAmount();`

so no need for an emit

dyersituations: I meant, should it be emitted instead?

christian14b: That was my first thought, but it only works when the amount changes. Ref handles that situation

dyersituations (approved):

PR #6900, opened 2026-05-13
fix: ensure goal card still appears once on mykiva after completing

dyersituations:
https://kiva.atlassian.net/browse/MP-2824

Encountered some more interesting edge cases for the goal card on MyKiva while testing my previous PR.

Here it's ensuring the goal card appears once on MyKiva when completed, gets hidden, and doesn't have a pop-in situation on the next steps page.

Moving the setting of the hidden status to the composable is also a good consolidation of goal changing code.


emuvente (approved):

christian14b (approved):

eddieferrer (approved):

cristhianDt (approved):
Cool!

PR #6901, opened 2026-05-13
feat: my kiva recommended loan experience (MP-2473)

christian14b:
## How to test

After you create a goal, you can go to `/goal-setting` page and remove the goal to start again

Screenshot 2026-05-13 at 3 43 40 PM
Screenshot 2026-05-13 at 3 43 50 PM

### DIFF
--- a/src/components/MyKiva/GoalSettingModal.vue
+++ b/src/components/MyKiva/GoalSettingModal.vue
@@ -2,15 +2,26 @@
 	<KvLightbox
 		class="goal-setting-lightbox"
 		:class="{
-			'goal-tile-modal': showGoalTile,
-			'goal-tile-modal-expanded': showGoalTile && showCategories
+			'goal-tile-modal': showGoalTile && !showRecommendLoanAfterGoalView,
+			'goal-tile-modal-expanded': showGoalTile && showCategories && !showRecommendLoanAfterGoalView,
+			'goal-tile-modal-recommend-loan': showRecommendLoanAfterGoalView,
 		}"
 		title=""
 		:visible="show"
 		@lightbox-closed="closeLightbox"
 	>
 		<template
-			v-if="!showGoalTile"
+			v-if="showRecommendLoanAfterGoalView"
+			#header
+		>
+			<RecommendLoanForGoalHeader
+				class="!tw-p-0"
+				:title="recommendLoanHeaderTitle"
+				:details="recommendLoanHeaderDetails"
+			/>
+		</template>
+		<template
+			v-else-if="!showGoalTile"
 			#header
 		>
 			<h2
@@ -21,12 +32,22 @@
 			</h2>
 		</template>
 		<h2
-			v-if="isMobile && (showCategories || isThanksPage)"
+			v-if="!showRecommendLoanAfterGoalView && isMobile && (showCategories || isThanksPage)"
 			class="tw-mb-3 tw-text-left md:tw-text-center"
 		>
 			Choose an impact area
 		</h2>
 		<section
+			v-if="showRecommendLoanAfterGoalView"
+		>
+			<RecommendLoanForGoalContent
+				ref="recommendLoanForGoalContentRef"
+				v-bind="recommendLoanCardProps"
+				:is-adding="isAdding"
+			/>
+		</section>
+		<section
+			v-else
 			:class="{ 'tw-flex tw-flex-col md:tw-flex-row tw-gap-0': showGoalTile }"
 		>
 			<div
@@ -171,8 +192,8 @@
 					:is-goal-tile-experiment-enabled="isGoalTileExperimentEnabled"
 					show-goal-value-props-copy
 					@set-goal-target="setGoalTarget"
-					@set-goal="$emit('set-goal', $event)"
-					@update-goal="$emit('set-goal', $event)"
+					@set-goal="onGoalSelectorSetGoal"
+					@update-goal="onGoalSelectorUpdateGoal"
 					@edit-goal="editGoalCategory"
 					@close-modal="closeLightbox"
 				/>
@@ -192,7 +213,7 @@
 				/>
 				<!-- second continue button for goal tile variant -->
 				<div
-					v-if="showGoalTile && (showCategories || isThanksPage)"
+					v-if="showGoalTile && !showRecommendLoanAfterGoalView && (showCategories || isThanksPage)"
 					class="tw-flex tw-justify-end tw-gap-2 goal-tile-categories-controls"
 				>
 					<KvButton
@@ -209,7 +230,18 @@
 			</div>
 		</section>
 		<template
-			v-if="!showGoalTile && (showCategories || isThanksPage)"
+			v-if="showRecommendLoanAfterGoalView"
+			#controls
+		>
+			<RecommendLoanForGoalFooter
+				class="tw-mx-auto tw-w-full"
+				:is-adding="isAdding"
+				:is-in-basket="recommendLoanIsInBasket"
+				@primary-cta-click="addToBasket"
+			/>
+		</template>
+		<template
+			v-else-if="!showGoalTile && (showCategories || isThanksPage)"
 			#controls
 		>
 			<div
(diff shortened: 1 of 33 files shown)


src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * State and props for the “recommended loan after goal” step in GoalSettingModal.
+ * @param {object} options
+ * @param {import('vue').Ref<object|null>} options.userGoal — annual goal from `useGoalData`.
+ * @param {Function} options.getCtaHref — from `useGoalData`; lend URL + header query.
+ */
+export default function useGoalSettingRecommendedLoan({
+	emit,
+	props,
```
dyersituations, line 36: Add comments for this input object in general. The `props` object is confusing without knowing where it's coming from.

src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -0,0 +1,154 @@
...
+			const p = goalProgress.value;
+			segments.push(`${p} ${p === 1 ? 'loan' : 'loans'} completed`);
+		}
+		return segments;
+	});
+
+	const recommendLoanCardProps = computed(() => {
+		const loan = recommendedLoan.value;
+		const photoPath = appConfig?.photoPath ?? '';
+		const kvTrackFunction = (...args) => {
```
dyersituations: Why is this needed instead of just passing the method directly?

src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -0,0 +1,154 @@
...
+			photoPath,
+			showTags: true,
+			externalLinks: true,
+			customLoanDetails: true,
+			showLightView: true,
+			basketItems: basketItems.value ?? [],
+			route: router.currentRoute.value,
+			kvTrackFunction,
+		};
+		if (!loan?.id) {
```
dyersituations, line 88: What does this return mean? No loan recommended? What happens in the UI?

src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -0,0 +1,154 @@
...
+			if (!visible || !categoryId) {
+				recommendedLoan.value = null;
+				return;
+			}
+			recommendedLoan.value = null;
+			try {
+				const loans = await getRecommendedLoans(categoryId);
+				recommendedLoan.value = loans[0] ?? null;
+			} catch {
+				recommendedLoan.value = null;
```
dyersituations, line 159: What happens in the case of `null` in these two branches?

src/composables/useGoalSettingRecommendedLoan.js
christian14b: Good catch, I'll send it directly

src/composables/useGoalSettingRecommendedLoan.js
christian14b: Yes, it means the recommended loan is still not loaded, UI will show the loading state

src/composables/useGoalSettingRecommendedLoan.js
christian14b: It will show the loading state, but I'll add a comment to double check in [MP-2746](https://kiva.atlassian.net/browse/MP-2746) what happens if the query fails...

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -209,7 +231,19 @@
...
 		<template
-			v-if="!showGoalTile && (showCategories || isThanksPage)"
+			v-if="showRecommendLoanAfterGoalView"
+			#controls
+		>
+			<RecommendLoanForGoalFooter
+				class="tw-mx-auto tw-w-full"
+				:is-adding="isAdding"
+				:is-in-basket="recommendLoanIsInBasket"
+				@primary-cta-click="addToBasket"
```
dyersituations, line 241: ATB worked but I encountered the below error in the browser console:

Image

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -21,12 +33,22 @@
...
-			v-if="isMobile && (showCategories || isThanksPage)"
+			v-if="!showRecommendLoanAfterGoalView && isMobile && (showCategories || isThanksPage)"
 			class="tw-mb-3 tw-text-left md:tw-text-center"
 		>
 			Choose an impact area
 		</h2>
 		<section
+			v-if="showRecommendLoanAfterGoalView"
+		>
+			<RecommendLoanForGoalContent
```
dyersituations, line 44: If I already had the loan in my basket, it shows this immediately. Seems like we should show the first loan that isn't in the basket currently?

Image

src/composables/useGoalSettingRecommendedLoan.js
dyersituations: So this gets called multiple times (before and after loan loaded)? Is there a simpler path if so?

src/composables/useGoalSettingRecommendedLoan.js
dyersituations: If the query fails, the loaders just never disappear? Should be instead collapse the loan card and show the button to the loan finding page?

src/components/MyKiva/GoalSettingModal.vue
christian14b: Yeah, I haven't looked at that method yet. But that makes sense, hopefully there's a way to filter all the loan ids in the current basket. I'll take a look

christian14b: Found it, we should be good with this, but let me know if you see anything odd

src/components/MyKiva/GoalSettingModal.vue
christian14b: This error still remains, I'll probably need to modify some code of that file + its tests

src/composables/useGoalSettingRecommendedLoan.js
christian14b: this will be only called when `recommendedLoan` changes. I can add some logs to see how many times this computed is actually being called

src/composables/useGoalSettingRecommendedLoan.js
christian14b: I can do it in this work, but It's not called out in the ticke. I haven't looked at the description of the Error handling ticket, but If it makes sense I can just hide the card for now

src/composables/useGoalSettingRecommendedLoan.js
dyersituations: Can `recommendedLoan` currently change?

src/composables/useGoalSettingRecommendedLoan.js
dyersituations: Could you update that ticket if it doesn't satisfy this error fallback? No need to do that with this PR.

test/unit/specs/plugins/borrower-profile-exp-mixin.spec.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-extraneous-dependencies -- @vue/test-utils devDependency */
```
dyersituations, line 1: A test in this file is currently failing in the pipeline.

src/composables/useGoalSettingRecommendedLoan.js
christian14b: I can confirm this is being called just twice (When recommended loan is being queried and when the value is loaded)

src/composables/useGoalSettingRecommendedLoan.js
christian14b: Sure! because it definitely looks weird without the card

src/components/MyKiva/GoalSettingModal.vue
dyersituations: Is this error resolved with recent commits?

src/plugins/borrower-profile-exp-mixin.js
```diff
@@ -32,19 +32,26 @@ export default {
 				query: borrowerProfileSideSheetQuery,
 				variables: { loanId },
 				fetchPolicy: fetchPolicy ?? 'cache-first',
-			}).then(({ data }) => {
-				this.selectedLoan = data?.lend?.loan;
+			}).then(result => {
+				const loan = result?.data?.lend?.loan;
+				if (loan) {
```
dyersituations, line 37: When is this not defined? Only in tests?

christian14b: Yeah, only for the tests.

src/components/MyKiva/GoalSettingModal.vue
christian14b: Yes, the changes in the mixin file fixed this problem

dyersituations (approved):

PR #6911, opened 2026-05-19
feat: [MP-2744] - integrate recommended loan experience after set a goal on goal setting page

cristhianDt:
# MP-2744

## Summary

updates the Goals page so that after a lender sets a goal, they see the new "start your goal" experience instead of the current confirmation page; adding a new storybook

https://kiva.atlassian.net/browse/MP-2744

## Evidences

https://github.com/user-attachments/assets/f58c308a-6e69-41e3-bd33-f3fc9ccbea2b

### DIFF
--- a/src/components/GoalSetting/GoalSettingContainer.vue
+++ b/src/components/GoalSetting/GoalSettingContainer.vue
@@ -431,8 +431,9 @@ const setGoal = async preferences => {
 	} catch (e) {
 		logFormatter('GoalSettingContainer: failed to setting up this goal', 'error', { error: e });
 		$showTipMsg('There was a problem setting up this goal', 'error');
+	} finally {
+		loading.value = false;
 	}
-	loading.value = false;
 };
 
 const handleCategorySelected = categoryId => {
--- a/src/components/GoalSetting/GoalSettingContainer.vue
+++ b/src/components/GoalSetting/GoalSettingContainer.vue
@@ -345,8 +345,9 @@ const categories = getCategories(props.categoriesLoanCount, props.totalLoans);
 
 const selectedCategory = ref(categories[0]);
 
-// Composable for recommended loan state (shared logic with GoalSettingModal)
-const showPage = ref(true);
+// This container doesn't have the same multi-step flow as the modal,
+// so we can always show the recommended loan section when enabled
+const showPage = true;
 const loadedSetData = ref(false);
 const {
 	showRecommendLoanAfterGoalView,
--- a/src/components/GoalSetting/GoalSettingContainer.vue
+++ b/src/components/GoalSetting/GoalSettingContainer.vue
@@ -45,7 +45,21 @@
 			style="max-width: 644px; min-height: 495px;"
 		/>
 		<template v-else>
+			<RecommendLoanForGoalContainer
+				v-if="showRecommendLoanAfterGoalView"
+				class="tw-mx-auto"
+				style="max-width: 700px;"
+				header-title="Goal set!"
+				:header-details="recommendLoanHeaderDetails"
+				:content-card-props="recommendLoanCardProps"
+				:is-adding="isAdding"
+				:is-in-basket="recommendLoanIsInBasket"
+				:loaded-set-data="loadedSetData"
+				@primary-cta-click="addToBasket"
+				@secondary-cta-click="handleExploreMoreLoans"
+			/>
 			<div
+				v-else
 				class="tw-mx-auto"
 				style="max-width: 644px;"
 			>
@@ -150,7 +164,11 @@ import {
 	KvUtilityMenu
 } from '@kiva/kv-components';
 import GoalSelector from '#src/components/MyKiva/GoalSetting/GoalSelector';
+import RecommendLoanForGoalContainer from
+	'#src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContainer';
 import useGoalData, { GOAL_STATUS } from '#src/composables/useGoalData';
+import useGoalSettingRecommendedLoan from
+	'#src/composables/useGoalSettingRecommendedLoan';
 import { buildEmailFlowGoalData } from '#src/util/goalEmailFlow';
 import logFormatter from '#src/util/logFormatter';
 import {
@@ -165,6 +183,7 @@ import useTipMessage from '#src/composables/useTipMessage';
 
 const apollo = inject('apollo');
 const $kvTrackEvent = inject('$kvTrackEvent');
+const $appConfig = inject('$appConfig', {});
 const router = useRouter();
 
 const { $showTipMsg } = useTipMessage(apollo);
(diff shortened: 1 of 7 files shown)


src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -306,6 +342,37 @@ const categories = getCategories(props.categoriesLoanCount, props.totalLoans);
...
+	userGoal,
+	kvTrackEvent: $kvTrackEvent,
+	appConfig: $appConfig,
+});
+
+const addToBasket = () => {
+	const { loanId } = recommendLoanCardProps.value;
+	if (!loanId) return;
+	// Delegate to parent (GoalSetting.vue) which uses borrower-profile-exp-mixin
+	emit('add-to-basket', { loanId, lendAmount: 25 });
```
dyersituations: This shouldn't be hardcoded to `25`, it should use the selection in the loan card.

christian14b: +1, take a look at how  `getSelectedAmount` is used to get the actual amount

src/components/GoalSetting/GoalSettingContainer.vue
infante-jaime: Delete your 2026 impact goal? is hardcoded. use GOALS_CURRENT_YEAR or the existing yearToDate, otherwise it becomes wrong in 2027.

src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -350,9 +417,13 @@ const updateGoal = async preferences => {
 };
 
 const setGoal = async preferences => {
+	loading.value = true;
```
christian14b, line 426: Seems like loading is not needed

src/pages/GoalSetting.vue
```diff
@@ -83,6 +88,7 @@ export default {
 
 		// Get feature flag for goal recommended loan
 		this.goalRecommendedLoanEnable = readBoolSetting(goalSettingPageResult, 'general.goal_recommended_loan_enable.value') ?? false; // eslint-disable-line max-len
+		await this.loadInitialBasketItems();
```
christian14b: We should not load basket items if `goalRecommendedLoanEnable` is false

src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -306,6 +342,37 @@ const categories = getCategories(props.categoriesLoanCount, props.totalLoans);
 
 const selectedCategory = ref(categories[0]);
 
+// Composable for recommended loan state (shared logic with GoalSettingModal)
+const showPage = ref(true);
```
christian14b: I believe this might be related to `show` in modal, but it's not used in the composable either so we can remove it entirely

cristhianDt: Wouldn't that break your watcher?

https://github.com/kiva/ui/blob/18e0db911b33f2dddf71192d49dcfdde2dc3a462/src/composables/useGoalSettingRecommendedLoan.js#L145-L153

src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -306,6 +342,37 @@ const categories = getCategories(props.categoriesLoanCount, props.totalLoans);
...
+const {
+	showRecommendLoanAfterGoalView,
+	recommendLoanHeaderDetails,
+	recommendLoanCardProps,
+	recommendLoanIsInBasket,
+	enterRecommendedLoanStepAfterGoalSave,
+	handleExploreMoreLoans,
+} = useGoalSettingRecommendedLoan({
+	emit: () => {},
+	props,
```
christian14b: Giving it a second thought, we only use `goalRecommendedLoanEnable` and `basketItems` from props. Can you replace props for those two instead? you might need to do a small change in GoalSettingModal too

src/components/GoalSetting/GoalSettingContainer.vue
christian14b: `show` is handled in `GoalSettingModal`, so removing it from the composable doesn't do any harm

christian14b: Oh wait, never mind. Yeah, leave at this, didn't look at that closely. Good catch!

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: in fact it is necessary, because the user watch the progress ring for a while then the recommended loan is shown, so yes, I need to show the loading of the parent for a moment while the composable is updated and the change is not abrupt more for aesthetic reasons 😅

christian14b: Got it! I think another way can be calling `enterRecommendedLoanStepAfterGoalSave` before `storeGoalPreferences` expecting an optimistic result

cristhianDt: true, but if something happen saving the preferences/goal is very likely that the user will see the recommended loan and then the error (BTW I moved `loading.value = false;` because I realized the Placeholder wont disappear if something happen)

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: done, you can clearly see the refactor here chris https://github.com/kiva/ui/pull/6911/changes/e0f69994ebbb2b1ec5a1e55cadf77812a90f5705

src/pages/GoalSetting.vue
cristhianDt: added, thank you!

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: good call!, fixed

src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -306,6 +345,39 @@ const categories = getCategories(props.categoriesLoanCount, props.totalLoans);
...
+	recommendLoanIsInBasket,
+	enterRecommendedLoanStepAfterGoalSave,
+	handleExploreMoreLoans,
+} = useGoalSettingRecommendedLoan({
+	emit: () => {},
+	goalRecommendedLoanEnable: toRef(props, 'goalRecommendedLoanEnable'),
+	basketItems: toRef(props, 'basketItems'),
+	selectedGoalNumber: goalSelectorLoanTarget,
+	selectedCategory,
+	show: showPage,
```
christian14b, line 365: If showPage does not chang. can we just use it as a const?

dyersituations: @cristhianDt looks like this is the last open question from yesterday, a small tweak still needed?

cristhianDt: changed

cristhianDt: @christian14b do you wanna check and close the thread?

dyersituations (approved):

christian14b (approved):

PR #6913, opened 2026-05-20
feat: mp 2757 goal copy update

infante-jaime:
### Summary
Updates the copy in the goal-setting flow to be more context-aware — showing different messaging depending on whether you're a new user, returning user, or editing an existing goal category.
https://kiva.atlassian.net/browse/MP-2757

### DIFF
--- a/src/components/MyKiva/GoalProgressRing.vue
+++ b/src/components/MyKiva/GoalProgressRing.vue
@@ -105,9 +105,6 @@ import { COMPLETED_GOAL_THRESHOLD, HALF_GOAL_THRESHOLD } from '#src/composables/
 import goalCopy from '#src/util/goalCopy';
 import {
 	ID_SUPPORT_ALL,
-	ID_CLIMATE_ACTION,
-	ID_REFUGEE_EQUALITY,
-	ID_BASIC_NEEDS,
 	ID_US_ECONOMIC_EQUALITY,
 } from '#src/composables/useBadgeData';
 import useBreakpoints from '#src/composables/useBreakpoints';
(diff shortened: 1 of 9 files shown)


src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -531,8 +532,9 @@ const titleText = computed(() => {
 const subtitleText = computed(() => {
 	if (loansThisYear.value > 0) {
 		const loanWord = loansThisYear.value === 1 ? 'loan' : 'loans';
+		const progressCopy = hasEditedCategory.value ? 'You’ve already made' : 'You already made';
```
dyersituations: I'm not seeing this difference in the mocks, where is this? I see "you've" for all MyKiva screens.

src/components/MyKiva/NextYearGoalCard.vue
```diff
@@ -124,12 +121,14 @@ const title = computed(() => {
 		return 'You haven’t set your goal yet!';
```
dyersituations: The copy here is currently conflicting between this ticket and this previous experiment. Please chat with Lauren + Laura on how we want to move forward on these conflicting copy variations.

Conflicting epic: https://kiva.atlassian.net/browse/MP-2605

Image

dyersituations: @infante-jaime let's just ensure the behavior of this ticket works as described if the experiment that shows "You haven't set..." is disabled. Let me know once that is verified locally.

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -266,6 +266,7 @@ import GoalProgressRing from '#src/components/MyKiva/GoalProgressRing';
 import GoalCustomAmountInput from '#src/components/MyKiva/GoalSetting/GoalCustomAmountInput';
 import useGoalData, { LAST_YEAR_KEY, GOAL_STATUS } from '#src/composables/useGoalData';
 import useBreakpoints from '#src/composables/useBreakpoints';
+import goalCopy from '#src/util/goalCopy';
```
roger-in-kiva, line 269: As we already have imported and centralized data in `useGoalData` composable.
Is there a way to make this changes in there?

infante-jaime: Yes, the idea is that if we need an extra copy variable, we can add it and reference it. And if the variable is used in multiple components, we can just change the copy in a single place. So that copy is consistent between components

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -603,7 +598,7 @@ const validateCustomAmount = value => {
 		const loanWord = loansThisYear.value === 1 ? 'loan' : 'loans';
 		validCustomAmount.value = false;
 		// eslint-disable-next-line max-len
-		customGoalAmountError.value = `Enter a number higher than the <strong>${loansThisYear.value} ${loanWord}</strong> you’ve already made this year`;
+		customGoalAmountError.value = `Enter a number higher than the <strong>${loansThisYear.value} ${loanWord}</strong> you've already made this year`;
```
roger-in-kiva: Do we need to add this one to the goalCopy JS file?

infante-jaime: makes sense

roger-in-kiva: double check if there are other missing cases pls.

src/components/MyKiva/NextYearGoalCard.vue
cristhianDt: @infante-jaime is this thread closed? or are you still waiting confirmation by product/design?

src/components/MyKiva/GoalSetting/GoalSelector.vue
cristhianDt: +1

src/components/MyKiva/GoalProgressRing.vue
```diff
@@ -201,7 +202,6 @@ const emit = defineEmits(['button-click', 'edit-button-click', 'edit-goal-from-s
 const router = useRouter();
 const { isLarge } = useBreakpoints();
 
-const yearToDate = new Date().getFullYear();
```
cristhianDt, line 204: thanks for removing it

src/components/MyKiva/NextYearGoalCard.vue
infante-jaime: we had this answer 

at the moment, because we are testing, we have different implementations for MyKiva and the Thank you Page

MyKiva is testing 50-50 the above copy, with the control being correct as written here https://kiva.atlassian.net/browse/MP-2757
The thank you page copy is testing the custom goal amount, otherwise the copy should be the same between the control and variant. I believe we also had a ticket to fix the thank you page copy, and it has already been completed.

still need to check how to keep the experiment going so please review the rest and I can keep the correct values between components

test/unit/specs/util/goalCopy.spec.js
```diff
@@ -0,0 +1,172 @@
+import { GOALS_CURRENT_YEAR } from '#src/composables/useGoalData';
```
cristhianDt, line 1: (thoughs) are we already covering the new cases in the component(s) where goalCopy is used? if it so, what do you think to use the value of goalCopy instead of hardcoded text to check the content?

src/components/MyKiva/NextYearGoalCard.vue
infante-jaime: just reviewed and we are good we just updated the copy but the logic that changes it is intact so we are good

test/unit/specs/util/goalCopy.spec.js
infante-jaime: good catch let me update that

src/components/MyKiva/NextYearGoalCard.vue
cristhianDt: oka, I will close the thread then, thanks !

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: updated it as much as I found

test/unit/specs/util/goalCopy.spec.js
infante-jaime: done

src/components/MyKiva/GoalProgressRing.vue
```diff
@@ -273,23 +273,23 @@ const modalVariantDescriptionText = computed(() => {
 
 	if (props.goalProgress > 0) {
 		const strongTag = `<strong ${brandClass}>${loans} loans</strong>`;
-		const prefix = "You're already on your way to making";
+		const prefix = 'You’re already on your way to making';
```
roger-in-kiva: Do we want these cases in `goalCopy` file?

infante-jaime: good call, moved it into the file

src/components/MyKiva/GoalSetting/GoalSelector.vue
roger-in-kiva: Port these cases as well?

cristhianDt (approved):
thanks @infante-jaime for applying my recommendations, there are still comments from @roger-in-kiva pending for you to validate

roger-in-kiva (approved):

PR #6915, opened 2026-05-20
feat: connect repayment data to loans-beta

dyersituations:
https://kiva.atlassian.net/browse/MP-2735

New API data available for repayment history for a user on a loan.

image

### DIFF
--- a/src/components/Portfolio/PaidAmountModal.vue
+++ b/src/components/Portfolio/PaidAmountModal.vue
@@ -12,14 +12,16 @@
 			@lightbox-closed="showModal = false"
 			title="Payment history"
 		>
-			<div class="tw-px-4">
-				<div class="tw-grid tw-grid-cols-2">
+			<div class="tw-text-left">
+				<p v-if="!paymentHistory?.length">
+					This loan has no repayments
+				</p>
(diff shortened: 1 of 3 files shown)


src/components/Portfolio/PaidAmountModal.vue
```diff
@@ -12,14 +12,16 @@
 			@lightbox-closed="showModal = false"
 			title="Payment history"
 		>
-			<div class="tw-px-4">
-				<div class="tw-grid tw-grid-cols-2">
+			<div class="tw-text-left">
+				<p v-if="!paymentHistory?.length">
+					This loan has no repayments
+				</p>
+				<div v-else class="tw-grid tw-grid-cols-2">
```
mcstover, line 19: Styles here could use some sprucing up or at least a little spacing. I thought we had a "SimpleTable" component and we do but it's only in CPS for now...

mcstover (approved):
Ready enough when the new field is

src/components/Portfolio/PaidAmountModal.vue
dyersituations: Yup agreed, I'm doing the bare minimum at the moment in these PRs, with the plan of doing a pass over the entire page when the data is complete.

PR #6925, opened 2026-05-26
feat: add tracking events for recommended goal loans

christian14b:
https://kiva.atlassian.net/browse/MP-2748

### DIFF
--- a/src/composables/useGoalSettingRecommendedLoan.js
+++ b/src/composables/useGoalSettingRecommendedLoan.js
@@ -28,9 +28,22 @@ import useTipMessage from '#src/composables/useTipMessage';
  * @param {Function} options.getCtaHref — From `useGoalData`; builds lend URL + optional header query.
  * @param {object} options.userGoal — Vue ref from `useGoalData`; `.value` is saved goal or null (`target`, `category`).
  * @param {Function} options.kvTrackEvent — Analytics helper from GoalSettingModal (`$kvTrackEvent`).
+ * @param {string} [options.entrypoint] — Where the composable is mounted. Use one of the exported
+ *   `GOAL_RECOMMENDED_LOAN_ENTRYPOINT_*` constants; selects the analytics category for
+ *   recommended-loan events. Omit to disable recommended-loan tracking.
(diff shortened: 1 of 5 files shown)


src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -114,6 +146,7 @@ export default function useGoalSettingRecommendedLoan({
 	const enterRecommendedLoanStepAfterGoalSave = () => {
```
cristhianDt, line 146: could we add an extra test for these method and confirm that kvTrackEvent is called?

mcstover (approved):
The extra test would prob be useful, otherwise looks good to me

src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -28,9 +28,22 @@ import useTipMessage from '#src/composables/useTipMessage';
...
+ *   `GOAL_RECOMMENDED_LOAN_ENTRYPOINT_*` constants; selects the analytics category for
+ *   recommended-loan events. Omit to disable recommended-loan tracking.
  * @param {object} [options.appConfig] — From GoalSettingModal (`$appConfig`), e.g. `photoPath`.
  * @param {object} options.apollo — Apollo client instance for tip message mutations.
  */
+export const GOAL_RECOMMENDED_LOAN_ENTRYPOINT_PORTFOLIO = 'portfolio';
+export const GOAL_RECOMMENDED_LOAN_ENTRYPOINT_POST_CHECKOUT = 'post-checkout';
+export const GOAL_RECOMMENDED_LOAN_ENTRYPOINT_GOALS_PAGE = 'goals-page';
+
+const ENTRYPOINT_TRACK_CATEGORY = {
```
roger-in-kiva, line 41: we can also add a test if this is returning correctly.

src/composables/useGoalSettingRecommendedLoan.js
roger-in-kiva: Yeah, worth adding tests to new functions.

roger-in-kiva (approved):

cristhianDt (approved):
thanks for adding more test cases!

PR #6926, opened 2026-05-26
feat: [MP-2861] show progress ring instead recommended row when not recommended loan

cristhianDt:
# MP-2861

## Summary

updates the Thank You page, Goal Setting Page and MyKiva Page to show progress ring instead recommended row when there is not a recommended loan after set a goal

https://kiva.atlassian.net/browse/MP-2861

### TY Page

https://github.com/user-attachments/assets/7ae9fc37-6639-4925-a979-a81e6ea8ad86

https://github.com/user-attachments/assets/303910ee-a736-4d50-a3c1-f2967ac59830

### Goals Page

https://github.com/user-attachments/assets/6218df3f-1ae3-4b51-9cf3-a01588fec76a

https://github.com/user-attachments/assets/2689086a-8951-4910-8b4e-a5ad0dc1e89f

### MyKiva Page

https://github.com/user-attachments/assets/895bb116-ee5a-4d02-a5c2-b89139d39010

https://github.com/user-attachments/assets/b94f80c4-9d7b-4898-837c-70156b7d5ef0


src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -462,6 +464,12 @@ const {
 	apollo,
 });
 
+// Render the recommended-loan UI only when the trigger fired AND the fetch
+// returned at least one loan; otherwise fall back to the goal-selector flow.
+const showRecommendLoanAfterGoalView = computed(() => (
```
mcstover: I think this is going to conflict due to being the same name with the property assigned/extracted from `useGoalSettingRecommendedLoan` above.

cristhianDt: yeah, but I renamed its name in the destructuring step

https://github.com/kiva/ui/blob/feature/mp-2861-show-progress-ring-instead-recommended-row-when-not-recommended-loan/src/components/MyKiva/GoalSettingModal.vue#L436C34-L436C58

roger-in-kiva: Probably better to rename this new function instead of destructuring with a different name.

mcstover: Ok yea I see now, I think it makes ok sense and see why you'd just rename the "root" property out of the composable and then keep using the original name to avoid changing the usages.

Maybe add a an additional comment that this is an "override" property...

src/components/GoalSetting/GoalSettingContainer.vue
```diff
@@ -40,13 +40,13 @@
...
 		<KvLoadingPlaceholder
-			v-if="loading || emailLoading"
+			v-if="loading || emailLoading || isLoadingRecommendedLoan"
 			class="!tw-rounded tw-mx-auto"
 			style="max-width: 644px; min-height: 495px;"
 		/>
 		<template v-else>
 			<RecommendLoanForGoalContainer
-				v-if="showRecommendLoanAfterGoalView"
+				v-if="showRecommendLoanAfterGoalView && hasRecommendedLoans"
```
christian14b, line 49: Can we add hasRecommendedLoan flag to showRecommendLoanAfterGoalViewn logic?

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -330,7 +333,7 @@ const {
 	basketItems: toRef(props, 'basketItems'),
 	selectedGoalNumber: goalTarget,
 	selectedCategory,
-	show: true,
+	show: ref(true),
```
christian14b, line 336: Why ref?

src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -196,16 +205,20 @@ export default function useGoalSettingRecommendedLoan({
 				recommendedLoans.value = [];
 				recommendedLoanIndex.value = 0;
 				recommendedLoan.value = null;
+				isLoadingRecommendedLoan.value = false;
 				return;
 			}
 			recommendedLoans.value = [];
```
christian14b: This four lines are repeated above. Can you create a new method to avoid duplication?

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: I did consider that, but it would require a major refactoring of the composable, the watcher, and the logic. I think this is probably the simplest approach, and it also abstracts the data from the composable, making it more granular and usable.

src/components/Thanks/ThanksPageSingleVersion.vue
cristhianDt: old version breaks my tests, let me try to adjust them to work without ref on this property

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: exactly, the idea is add an extra condition that it's not related with the original approach of the property

I will rename the new method, like roger suggested, to reduce some misunderstanding

src/components/GoalSetting/GoalSettingContainer.vue
christian14b: Ok I see to, this change is not for all pages that uses recommended loan?

src/components/Thanks/ThanksPageSingleVersion.vue
roger-in-kiva: Issue is we have a watcher for that var in the composable, watcher vars should be reactive.

christian14b: True. We can leave it for now if it's complicated to solve, not a huge deal
that should block this Pr

src/composables/useGoalSettingRecommendedLoan.js
cristhianDt: @christian14b I moved to `resetRecommendedLoanState` method, what do you think?

christian14b: I don't see the change, but sounds good! avoiding repeated code is the main goal. thanks!

cristhianDt: haha sorry for the delay (change is coming...)

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: it is but if you check the flow of `showRecommendLoanAfterGoalView` this prop trigger the query to fetch the recommended loans so adding the `hasRecommendedLoans` in its logic will create a type of circular dependency

my idea:

1. avoid that
2. make each little part re-usable for different situations

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: changed, thank you both!

mcstover (approved):
Looks good from a code perspective, I'll let you all decided if the logic and presentation is correct for your task

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -462,6 +477,19 @@ const {
 	apollo,
 });
 
+// Render the recommended-loan UI only when the trigger fired AND the fetch
+// returned at least one loan; otherwise fall back to the goal-selector flow.
+const showRecommendLoan = computed(() => (
```
roger-in-kiva, line 482: Recommended? for consistency with comment above and other similar instances

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -462,6 +477,19 @@ const {
...
+	showRecommendLoanAfterGoalView.value && hasRecommendedLoans.value
+));
+
+// True while the recommended-loan area occupies the modal — either fetching
+// the recommendation or already showing it. Used to gate header/controls/
+// layout classes so the pre-recommend UI doesn't flash mid-transition.
+const showRecommendLoanArea = computed(() => (
```
roger-in-kiva, line 489: same

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: so should I rename the components as well?

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: so should I rename the components as well?

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: the name is related with the components name, I will adjust the comments instead

src/components/MyKiva/GoalSettingModal.vue
roger-in-kiva: Up to you, I just think to keep some consistency.

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: updated comments

src/components/MyKiva/GoalSettingModal.vue
cristhianDt: updated comments

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: @christian14b can we close the thread?

infante-jaime (approved):

src/components/GoalSetting/GoalSettingContainer.vue
cristhianDt: we could follow this later if it's necessary

PR #6927, opened 2026-05-26
feat: add loan id and amount to add to basket event

christian14b:
Adding `loanId` and `loanAmount` to add to basket tracking event

### DIFF
--- a/src/composables/useGoalSettingRecommendedLoan.js
+++ b/src/composables/useGoalSettingRecommendedLoan.js
@@ -121,13 +121,9 @@ export default function useGoalSettingRecommendedLoan({
 	const trackCategory = ENTRYPOINT_TRACK_CATEGORY[entrypoint];
 	const isPostCheckout = entrypoint === GOAL_RECOMMENDED_LOAN_ENTRYPOINT_POST_CHECKOUT;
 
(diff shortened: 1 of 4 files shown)


src/composables/useGoalSettingRecommendedLoan.js
```diff
@@ -121,13 +121,22 @@ export default function useGoalSettingRecommendedLoan({
 	const trackCategory = ENTRYPOINT_TRACK_CATEGORY[entrypoint];
 	const isPostCheckout = entrypoint === GOAL_RECOMMENDED_LOAN_ENTRYPOINT_POST_CHECKOUT;
 
-	const trackRecommendedLoanEvent = (action, label) => {
+	const trackRecommendedLoanEvent = (action, label, property, value) => {
```
mcstover: I think you can just default the two new params to null and then not have to worry about forking the call below. The kvTrackEvent should handle things gracefully if those properties are null or undefined.

christian14b: Nice! I was not sure about it, thank you!

roger-in-kiva (approved):

eddieferrer (approved):

PR #6929, opened 2026-05-27
feat: filter loan ids from transaction in thank you page goal entrypoint

christian14b:
https://kiva.atlassian.net/browse/MP-2880

image


mcstover (approved):

src/components/Thanks/ThanksPageSingleVersion.vue
cristhianDt, line 235: @christian14b could we rename this property to make it more meaningful? something like supportedLoans or loansLent, idk..., what do you think?

cristhianDt:
@christian14b please merge this PR after mine https://github.com/kiva/ui/pull/6926, thank you

roger-in-kiva (approved):

src/components/Thanks/ThanksPageSingleVersion.vue
christian14b: Prob, I'll need to change more things in this file if doing that, so I don't think its worth the effort

PR #6932, opened 2026-05-27
feat: utilize new My Teams page at standard teams/my-teams route [CP-2628]

leoperry:
https://kiva.atlassian.net/browse/CP-2628

In coordination with CS, we sent out the teams/my-teams-beta page to team captains and gathered feedback. Feedback was generally positive, so we are now ready to roll out this in non-beta form in production!

Once this route update releases, we'll follow up with Flux ingress updates as well.

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -171,6 +171,13 @@ export default [
 		component: () => import('#src/pages/LendingTeams/LendingTeams')
 	},
 	{
+		path: '/teams/my-teams',
+		component: () => import('#src/pages/LendingTeams/MyTeamsPage'),
+		meta: {
+			authenticationRequired: true,
+		}
+	},
+	{
 		path: '/teams/my-teams-beta',
 		component: () => import('#src/pages/LendingTeams/MyTeamsPage'),
 		meta: {
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -179,11 +179,7 @@ export default [
 	},
 	{
 		path: '/teams/my-teams-beta',
-		component: () => import('#src/pages/LendingTeams/MyTeamsPage'),
-		meta: {
-			authenticationRequired: true,
-			excludeFromStaticSitemap: true,
-		}
+		redirect: '/teams/my-teams'
 	},
 	{
 		path: '/lend-by-category/loans-to-women',


src/router/routes.js
```diff
@@ -170,6 +170,13 @@ export default [
...
 	},
+	{
+		path: '/teams/my-teams',
+		component: () => import('#src/pages/LendingTeams/MyTeamsPage'),
+		meta: {
+			authenticationRequired: true,
+		}
+	},
 	{
 		path: '/teams/my-teams-beta',
```
eddieferrer, line 181: We should probably change the beta route  to a redirect.

leoperry: We certainly could, but we'd lose the ability to access the "new" beta page until we get the ingress etc straightened out in the given environment. Not a huge use case to be concerned about I suppose, because it's just the cohort of team captains who might happen to hit that route after UI release but before ingress update.

Because we don't plan to maintain the beta route in perpetuity, my plan was to keep it simple and just delete the beta route here once we had the UI release and ingress all buttoned up -- which I'd do whether it was a redirect or standalone route mapping in this file.

You've done this more than me though, so I'd certainly be happy to amend my approach based on your experience here.

mcstover (approved):
I think the captains will know where to look once this goes live and the redirect prob isn't strictly required. Could be useful temporarily though...

leoperry:
I put in the redirect based on the feedback here. Long term I don't think there will be any issues what happens with the beta route as it's an extremely small real world cohort, and the expectation is that was for special access to the beta page for testing purposes (i.e., no expectation it would be a permanent URL).

I have it in the ticket, but to clarify my overall plan for rolling this out:

1. Update routes in UI to support /teams/my-teams (this PR)
2. Update Tilt local ingress to support /teams/my-teams routing to UI
3. After UI is released, update Traefik ingress in all environments to route /teams/my-teams to UI (and drop beta page ingress)
4. Update monolith router config and usages to route to UI
5. Delete UI beta route (and local Tilt ingress if still exists)

eddieferrer (approved):

leoperry:
Wowza! Coveralls works so I can actually merge this 🎉

PR #6933, opened 2026-05-27
build(dependabot): group related dependencies AD-255

emuvente:
This groups related dependency updates to reduce dependabot PR noise and simplify ecosystem updates. This includes a catch-all group for minor and patch updates that aren't placed in another group first. My hope is that this will unclog the backlog of updates and make it easier to review updates going forward.

Related to https://github.com/kiva/cms-page-server/pull/3000

https://kiva.atlassian.net/browse/AD-255

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,19 +1,90 @@
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "13:00"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,19 +1,90 @@
 version: 2
 updates:
-- package-ecosystem: npm
```
emuvente, line 3: While I was here I also added the proper indentation

mcstover (approved):

PR #6944, opened 2026-05-29
feat: mp-2821-fixing the apply credit at checkout

infante-jaime:
[MP-2821](https://kiva.atlassian.net/browse/MP-2821 )

The checkout basket was showing incorrect credit badges on loan items because the component was reading the amount field (the total credit pool size) instead of applied (what's actually allocated to that specific loan), and wasn't filtering out credits with zero applied — so in a multi-loan basket, every loan could show badges for credits not funding it, with wrong amounts. The fix is two small changes: filter credits to only those where applied is non-zero, and display applied instead of amount, falling back to amount when applied is null (which the API returns in the single-loan case). Tests were added to cover both the zero-applied case and the applied-vs-amount display case.

[MP-2821]: https://kiva.atlassian.net/browse/MP-2821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Checkout/BasketItem.vue
+++ b/src/components/Checkout/BasketItem.vue
@@ -224,7 +224,8 @@ export default {
 		appliedPromoCredits() {
 			if (this.creditsUsed.length) {
 				const appliedCredits = this.creditsUsed.filter(credit => {
-					return credit.creditType !== 'kiva_credit';
+					return credit.creditType !== 'kiva_credit'
+						&& (credit.applied === null || parseFloat(credit.applied) > 0);
 				});
 				return appliedCredits.length ? appliedCredits : [];
 			}
--- a/src/components/Checkout/LoanPromoCredits.vue
+++ b/src/components/Checkout/LoanPromoCredits.vue
@@ -35,7 +35,7 @@ export default {
 		formattedPromoCredits() {
 			return this.appliedPromoCredits.map(credit => {
 				return {
-					amount: credit?.amount ?? 0,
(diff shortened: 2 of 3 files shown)


src/components/Checkout/BasketItem.vue
```diff
@@ -224,7 +224,8 @@ export default {
 		appliedPromoCredits() {
 			if (this.creditsUsed.length) {
 				const appliedCredits = this.creditsUsed.filter(credit => {
-					return credit.creditType !== 'kiva_credit';
+					return credit.creditType !== 'kiva_credit'
+						&& (credit.applied === null || parseFloat(credit.applied) > 0);
```
mcstover, line 228: I think this could be `(credit?.applied && parseFloat(credit.applied) > 0)` so we are chekcing that credit.applied is actionable and that it's greater than 0....

infante-jaime: Hey Matt 

in this case the differenciation of null and 0 is actually intentional ,  the API returns applied: null in the single-loan case, so telling us it hasn't broken down the value per item, so we fall back to amount. 

Treating null the same as 0 would silently hide credits in that scenario. The explicit applied === null check keeps those two states separate.

mcstover (approved):

roger-in-kiva (approved):

christian14b (approved):

PR #6955, opened 2026-06-03
feat: [MP-2747] - add express checkout modal in thankyou page after set goal

cristhianDt:
# MP-2747

## Summary

add express checkout modal in thankyou page after set goal

1. new express checkout modal modal
2. new unit tests
3. fix TYPageSingle tests
4. new storybook Recommended Loan + Express Checkout after set goal on TYPageSingle story
5. new composable useExpressCheckoutModal + components

https://kiva.atlassian.net/browse/MP-2747

### Happy Path without kiva credit or bonus

https://github.com/user-attachments/assets/aa46e642-c393-450b-8b2d-6f11cd2ae8c9

### Loan with matchingText

Captura de pantalla


src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -212,7 +209,7 @@ const onSubmit = async () => {
 		// TODO(MP-2747): track 'fail' event for express-checkout
 
 		if (e?.code === 'shop.failedCheckoutValidation') {
-			window.location.href = '/checkout';
+			router.push('/checkout');
```
roger-in-kiva: We usually go to `/basket` for some validations instead of `/checkout` directly.

src/composables/useExpressCheckoutModal.js
```diff
@@ -48,7 +48,7 @@ export default function useExpressCheckoutModal({
 		// current basket state: open the modal, re-open it (Checkout now
 		// re-entry), or redirect to /basket for the full checkout.
 		await loadInitialBasketItems();
-		let items = basketItems.value ?? [];
+		let items = unref(basketItems) ?? [];
```
roger-in-kiva, line 78: what is unref?

src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContainer.vue
```diff
@@ -86,8 +87,21 @@ defineProps({
 		type: Boolean,
 		default: true,
 	},
+	/**
+	 * When we're redirecting to /basket instead of opening the express checkout modal.
+	 */
+	isRedirecting: {
```
christian14b, line 93: This flag is for the scenario when the user has something in their basket, right? Can we add it in the comment for more context?

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,239 @@
...
+	<KvLightbox
+		title="Confirm Payment"
+		:visible="lightboxOpen"
+		@lightbox-closed="closeLightbox"
+	>
+		<form
+			v-if="lightboxOpen"
+			style="max-width: 25rem;"
+			class="tw-mx-auto"
+			action="."
```
christian14b: Is this action definition needed?

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: okey, so should we change this in CMS ?, I used part of this modal from the existing one on CMS

src/composables/useExpressCheckoutModal.js
cristhianDt: a safety method that check if the argument is a ref if it so it handle as it is if not just return the argument like final value

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,239 @@
...
+
+		const checkoutId = transactionResult?.data?.checkoutStatus?.receipt?.checkoutId ?? 0;
+		await trackTransactionEvent({
+			apollo,
+			transactionId: Number(checkoutId),
+		});
+
+		await createBasket(apollo);
+
+		if (transactionResult?.data?.checkoutStatus?.status !== 'COMPLETED') {
```
christian14b: Do you see an scenario where the methods above failed but didn't throw an error to add this guard?

src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContainer.vue
cristhianDt: sure, actually I already added in the previous line "When we're redirecting..." this property is generic in case we add another flow to redirecting

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
roger-in-kiva: Probably, but not part of this changeset.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,239 @@
...
+		closeLightbox();
+	} catch (e) {
+		// TODO(MP-2747): track 'fail' event for express-checkout
+
+		if (e?.code === 'shop.failedCheckoutValidation') {
+			router.push('/checkout');
+			return;
+		}
+
+		if (e?.code === 'shop.dropinNoPaymentMethod') {
```
christian14b, line 218: All these 3 lines bellow do the same $showTipMsg(customMsg, 'error'), I think you can add these conditions for a customMsg value and execute it once

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: yes

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: yeah it's a good improvement for the initial version (cms version) thanks

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -170,6 +171,14 @@
 			@close-goal-modal="closeGoalModal"
 			@set-goal="setGoal"
 		/>
+		<ExpressCheckoutModal
+			v-if="!isGuest && goalRecommendedLoanEnable"
+			ref="expressCheckoutModalRef"
+			:loan="expressCheckoutLoan"
+			:is-logged-in="!isGuest"
```
christian14b: If there's a v-if with `isGuest`, Is `isLoggedIn` not always true?

roger-in-kiva: Nice call.
Ensure this work for logged out users as well.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: removed

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: changed

src/components/Thanks/ThanksPageSingleVersion.vue
cristhianDt: @roger-in-kiva that is not possible, I mean, the approach of this version is work with the GoalEntryPoint https://github.com/kiva/ui/blob/main/src/components/Thanks/ThanksPageSingleVersion.vue#L52-L53, at this point GoalEntryPoint is shown to logged-in users

if the componente (ExpressCheckoutModal) will be used for guest that means new approach followed in a new ticket, for the moment we enable it only for logged-in users

roger-in-kiva: Oh yeah, that makes sense, thanks checking!

roger-in-kiva:
Looks like kiva credit selector is wrong:
Screenshot 2026-06-03 at 11 16 12 a m

Check if loaders should be centered, instead that probably in CMS are like this, there is always an opportunity to improve new components:
Screenshot 2026-06-03 at 11 17 01 a m
Screenshot 2026-06-03 at 11 17 05 a m

### Also check with Lauren if we want the kiva credit selector enabled as default.

roger-in-kiva:
Here some issues Claude found should be nice to check:

1. #1 ExpressCheckoutModal.vue:184 — failed checkouts wipe the user's basket and emit phantom transactions. The status check must be the first thing after executeOneTimeCheckout. Fix is a 3-line reorder.

2. #2 ExpressCheckoutModal.vue:183 — user gets charged but never reaches the thanks page on a missing-receipt path. Replace ?? 0 with an explicit check that throws or skips the emit when checkoutId is missing.

3. #3 ExpressCheckoutModal.vue:205 — call closeLightbox() before the redirect on shop.failedCheckoutValidation.

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
```diff
@@ -0,0 +1,134 @@
...
+const toggleCredit = async apply => {
+	changingCredit.value = true;
+	try {
+		if (apply) {
+			await applyKivaCredit(apollo);
+		} else {
+			await removeKivaCredit(apollo);
+		}
+		// TODO: track success event for apply/remove kiva credit
+	} catch (e) {
```
infante-jaime, line 116: One thing I noticed is that if toggleCredit throws, changingCredit gets reset but creditAppliedModel stays in the state the user selected. That leaves the toggle showing the wrong value until the next watchBasketTotals update comes through. We could probably set creditAppliedModel.value = isCreditApplied.value in the catch block to keep it in sync right away when the request fails.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,233 @@
...
+	if (depositRequired.value && !transactionsEnabled.value) {
+		return 'disabled';
+	}
+	if (paying.value) {
+		return 'loading';
+	}
+	return '';
+});
+
+const subscribeTotals = () => {
```
infante-jaime, line 120: both ExpressCheckoutModal and ExpressCheckoutTotals subscribe to watchBasketTotals separately. The modal only uses it to read totalDue for depositRequired, while the totals component already uses totalDue along with the rest of the basket totals.

We could pass totalDue down as a prop from the totals component and remove the modal’s subscription, so we keep a single source of truth.

cristhianDt:
moving to draft to handle new requirements, in the meantime

dyersituations:
@cristhianDt is this PR good to go for review?

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: @infante-jaime this comment requires a refactor so I'll do it in the next PR for MP-2900, agree?

infante-jaime: sure, add a comment in the ticket to track

cristhianDt:
> @cristhianDt is this PR good to go for review?

its ready now I missed commit a stash, also I added more cases in our suites

src/pages/Thanks/ThanksPage.vue
```diff
@@ -341,6 +347,7 @@ export default {
 		};
 
 		this.goalRecommendedLoanEnable = readBoolSetting(data, 'general.goal_recommended_loan_enable.value') ?? false;
+		this.expressCheckoutEnabled = readBoolSetting(data, 'general.ty_page_express_checkout_enabled.value') ?? false;
```
dyersituations, line 350: Should we also be checking `goalRecommendedLoanEnable`?

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,246 @@
...
+import logFormatter from '#src/util/logFormatter';
+
+defineProps({
+	loan: {
+		type: Object,
+		default: () => ({}),
+	},
+	analyticsCategory: {
+		type: String,
+		default: 'post-checkout-express',
```
dyersituations: We need to use existing analytics categories unless there is a important reason to add a new category, like for a major new flow in the UI. Here we can just use `post-checkout`. Make sure it's used everywhere as needed in the new changes.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,246 @@
...
+
+		// TODO(MP-2747): track 'success' event for express-checkout
+		emit('checkout-complete', {
+			transactionId: checkoutId,
+			amount: numeral(totalDue.value).format('0.00'),
+			transactionInfo: transactionResult?.data?.checkoutStatus,
+		});
+		closeLightbox();
+	} catch (e) {
+		// TODO(MP-2747): track 'fail' event for express-checkout
```
dyersituations: There are a bunch of TODOs here about events being needed. Can we just add them now so we don't miss doing it? Or add a ticket that covers these?

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
```diff
@@ -0,0 +1,184 @@
...
+			{{ formattedItemTotal }}
+		</p>
+
+		<template v-if="hasRedemptionCode">
+			<p>
+				{{ formattedRedemptionApplied }} Promotion
+			</p>
+			<div class="tw-flex tw-items-center tw-justify-end">
+				<span>-{{ formattedRedemptionApplied }}</span>
+				<!-- <button
```
dyersituations: What is the commented out code for specifically?

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
```diff
@@ -0,0 +1,184 @@
...
+			</div>
+		</template>
+
+		<template v-if="hasBonusCredit">
+			<p>
+				{{ formattedBonusApplied }} Kiva Promotion
+			</p>
+			<div class="tw-flex tw-items-center tw-justify-end">
+				<span>-{{ formattedBonusApplied }}</span>
+				<!-- <button
```
dyersituations: What is the commented out code for specifically?

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
```diff
@@ -0,0 +1,184 @@
...
+		const message = e?.message
+			|| `Something went wrong ${verb} your Kiva Credit. Please, refresh the page and try again.`;
+		$showTipMsg(message, 'error');
+		// Revert the switch to the actual basket state. Without this the
+		// switch stays stuck on the user's selected position while the
+		// credit was never applied/removed on the server — and the
+		// `watchBasketTotals` subscription won't re-fire (the cache didn't
+		// change because the mutation failed).
+		creditAppliedModel.value = isCreditApplied.value;
+		// TODO: track fail event for apply/remove kiva credit
```
dyersituations: 2 TODOs for events here, can we add them now?

src/composables/useExpressCheckoutModal.js
```diff
@@ -0,0 +1,152 @@
...
+
+	const expressCheckoutModalRef = ref(null);
+	const expressCheckoutLoan = ref(null);
+	const isRedirecting = ref(false);
+
+	async function handleAddRecommendedLoanToBasket(payload) {
+		// Feature flag off → skip the modal flow entirely. Add the recommended
+		// loan to the basket and redirect to /basket, matching the pre-express
+		// behaviour. No donation cleanup, no re-entry detection, no modal.
+		if (!unref(isExpressCheckoutEnabled)) {
```
dyersituations, line 60: Just curious, why use `unref` here?

src/pages/Thanks/ThanksPage.vue
cristhianDt: we are checking that flag in goal entry point, if it is false the user won't never see the action to open the modal so I think we are fine

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: those are mine, I confirmed with Lauren that we don't need it, let me remove the comments, thanks for reminding me

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
cristhianDt: related with the new ticket that we talked recently in slack, allow to remove the promotion rows, laura wanted to change the final design I just commented my initial version, I will enable and adjust according with the new mocks (talking with Laura if there are ready)

cristhianDt: removing in the meantime just to not confuse

src/composables/useExpressCheckoutModal.js
cristhianDt: as it is a composable, I just add this pretty "unpacking" method, this allows you to check if the argument is a ref and handle at it is otherwise return it as final value

dyersituations: So in this situation it would basically be equivalent to using `isExpressCheckoutEnabled.value`?

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
cristhianDt: not we don't need let me remove them

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: fixed

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
cristhianDt: same answer, just removed to reduce noise

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
cristhianDt: removed

src/composables/useExpressCheckoutModal.js
cristhianDt: if it's a ref type, check this example to know what unref is, it's pretty cool:

return > `val = isRef(val) ? val.value : val`

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -0,0 +1,239 @@
+<template>
```
dyersituations, line 1: You can currently close this modal, click on payment options, interactions generally while the checkout is happening. We shouldn't allow the user to close the modal during checkout, and we shouldn't allow any payment buttons/elements to be changed.

src/composables/useExpressCheckoutModal.js
```diff
@@ -0,0 +1,153 @@
...
+				isRedirecting.value = false;
+				previousOnError?.();
+			},
+		});
+	}
+
+	function handleExpressCheckoutComplete({ transactionId }) {
+		if (!transactionId) return;
+		kvTrackEvent?.(EVENT_CATEGORY, 'click', 'complete-order');
+		router.push(`/checkout/thanks?kiva_transaction_id=${transactionId}`);
```
dyersituations: This push isn't working like expected, the page isn't redirecting/refreshing and stays on previous transaction but with the updated URL. Let's change this to `window.location` to force full page reload.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
dyersituations: @cristhianDt this can be a follow-up ticket, but please log it if that's the path you want to take.

src/composables/useExpressCheckoutModal.js
dyersituations: @cristhianDt looks like this is the only critical fix to get in place before we approve this PR.

cristhianDt: haha yes I just asked about it in slack https://kivahq.slack.com/archives/C0AA9001B4H/p1781117921043809, thanks!

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: @dyersituations agree, here the [new ticket](https://kiva.atlassian.net/browse/MP-2905) for that, thanks!

src/composables/useExpressCheckoutModal.js
cristhianDt: changed

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+	<div
+		class="tw-grid tw-grid-cols-2 tw-gap-2 tw-place-content-between tw-mb-2"
+	>
+		<div>
+			<p class="tw-font-medium">
+				{{ loanLabel }}
+			</p>
+			<p v-if="loan && loan.matchingText" class="tw-text-small tw-text-secondary matching-text">
+				Matched by {{ loan.matchingText }}
```
eddieferrer, line 10: I think we'll need to add simultaneous matching information here when that feature rolls out.

cristhianDt: is a new property eddie?, I mean, will be

dyersituations (approved):

dyersituations:
@cristhianDt looks good pending conflicts being resolved with test file.

src/components/Thanks/ExpressCheckout/ExpressCheckoutTotals.vue
eddieferrer: I can implement it here when we're closer to rolling simultaneous matching out. Heres the epic: https://kiva.atlassian.net/browse/CIT-4067 and heres the ticket I created for this: https://kiva.atlassian.net/browse/CIT-4398

dyersituations: Thanks @eddieferrer for tracking this! This is an experiment, so it may or may not roll out permanently.

cristhianDt: thanks @eddieferrer we will keep this in mind in the future if we persist the experiment

PR #6963, opened 2026-06-05
feat: mp-2868-goal setting copy change + thank you cap

infante-jaime:
[MP-2868](https://kiva.atlassian.net/browse/MP-2868)
Also includes 
[MP-2870](https://kiva.atlassian.net/jira/software/projects/MP/boards/275?selectedIssue=MP-2870
)Adds calendar-based goal signup copy for annual goal entrypoints, switching from prior-year impact copy in Q1 to “You haven’t set your goal yet!” starting April 1.

Also enables custom goal amounts by default, adds a 3-view Thanks-page cap for the goal signup ask, and includes a low-environment-only goalSignupDate=YYYY-MM-DD query param for QA date testing.

[MP-2868]: https://kiva.atlassian.net/browse/MP-2868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

goalsetting in march
image

goalsetting may
image

portfolio march
image

portfolio in may 
image

mykiva march
image
image

mykiva may
image

image

thankyou april

thank you may 

[MP-2870]: https://kiva.atlassian.net/browse/MP-2870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -439,6 +439,7 @@ const goalTileAccordion = ref(null);
 const customGoalAmount = ref(null);
 const validCustomAmount = ref(null);
 const customGoalAmountError = ref('');
+const goalSignupCopyVariant = computed(() => goalCopy.getGoalSignupCopyVariant());
```
roger-in-kiva: why do we need computed?

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -381,11 +381,11 @@ const props = defineProps({
 		default: false,
 	},
 	/**
-	 * Whether the custom goal amount feature is enabled (from experiment)
+	 * Whether the custom goal amount feature is enabled
 	 */
 	customGoalAmountEnable: {
 		type: Boolean,
-		default: false,
+		default: true,
```
christian14b: Is this no longer an experiment? should we remove `customGoalAmountEnable` flag?

src/components/MyKiva/GoalSetting/GoalSelector.vue
christian14b: Same, if the only variable is date I don't think this should be a computed

src/util/goalSignupCopyVariant.js
```diff
@@ -13,9 +13,57 @@ export const GOAL_SIGNUP_COPY_VARIANT = {
...
+	/(^|\.)development\.kiva\.org$/,
+	/(^|\.)dev\.kiva\.org$/,
+	/(^|\.)stage\.kiva\.org$/,
+	/(^|\.)test\.kiva\.org$/,
+	/(^|\.)qa\.kiva\.org$/,
+	/(^|\.)audit\.kiva\.org$/,
+	/^ui-(dev|stage|test|qa|audit)\.dk1\.kiva\.org$/,
+];
+
+function isLowEnvironment(hostname = '') {
```
roger-in-kiva: Why we need to check if is a low env?

infante-jaime: That is for the date setter, which is only used for testing purposes.

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: you are correct, removed all references

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: right removed unecessary computed values

src/components/MyKiva/GoalSetting/LoanNumberSelector.vue
```diff
@@ -62,17 +58,5 @@ const props = defineProps({
 		type: String,
 		default: '',
 	},
-	/**
-	 * Whether the custom goal amount feature is enabled (from experiment)
-	 */
-	customGoalAmountEnable: {
```
dyersituations, line 68: Is there any remaining experiment removal/roll-out work needed for custom goal amount?

src/pages/Portfolio/ImpactDashboard/GoalEntrypoint.vue
```diff
@@ -62,6 +62,9 @@ const router = useRouter();
 const womenLoansLastYear = ref(0);
 
 const title = computed(() => {
+	if (goalCopy.getGoalSignupCopyVariant() === GOAL_SIGNUP_COPY_NO_GOAL_YET) {
```
dyersituations: It seems like this logic is repeated and could instead be a new method on `goalCopy`.

src/util/goalSignupCopyVariant.js
```diff
@@ -13,9 +13,57 @@ export const GOAL_SIGNUP_COPY_VARIANT = {
...
+	return LOW_ENV_HOST_PATTERNS.some(pattern => pattern.test(hostname.toLowerCase()));
```
dyersituations: What specifically is this check needed for with testing and how is it used? It seems like some decently complex regex added to the built app just for a test situation.

src/util/goalSignupCopyVariant.js
```diff
@@ -39,8 +87,10 @@ function resolveGoalSignupCopyVariant(date = new Date()) {
  *
  * @param {object} [options]
  * @param {Date}   [options.now] - Override date (used by tests to pin a boundary).
+ * Query param override: ?goalSignupDate=YYYY-MM-DD works only in low environments
```
dyersituations: It looks like this is the use case. I think we can keep this without the lower env check, since normal users won't be able to find this query string. Also all it does is change the copy, right? Thoughts?

dyersituations:
@infante-jaime please add some screenshots of the UI locations affected by this PR with the copy as expected.

dyersituations:
@infante-jaime it looks like work remains on this PR?

infante-jaime:
@dyersituations yes and no as per the initial  ticket is done yet I guess it will be better to pass the other two tickets in a single PR so marking this as draft in the meantime.

dyersituations:
> @dyersituations yes and no as per the initial ticket is done yet I guess it will be better to pass the other two tickets in a single PR so marking this as draft in the meantime.

See open questions in the PR comments as well.

src/components/MyKiva/GoalSetting/LoanNumberSelector.vue
infante-jaime: No, The customGoalAmountEnable prop and custom_goal_amount gating have been removed, and custom amount is now always available in GoalSelector.

src/pages/Portfolio/ImpactDashboard/GoalEntrypoint.vue
infante-jaime: fixed

src/util/goalSignupCopyVariant.js
infante-jaime: Yes it was a bit messy and unnecessary, yet we should keep the date change  parameter since is hard to get, and it only changes copy

src/util/goalSignupCopyVariant.js
infante-jaime: same as above

src/util/goalSignupCopyVariant.js
```diff
@@ -39,8 +65,9 @@ function resolveGoalSignupCopyVariant(date = new Date()) {
  *
  * @param {object} [options]
  * @param {Date}   [options.now] - Override date (used by tests to pin a boundary).
+ * Query param override: ?goalSignupDate=YYYY-MM-DD only affects this copy decision.
  * @returns {'last-year' | 'no-goal-yet'}
  */
 export default function useGoalSignupCopyVariant({ now } = {}) {
-	return resolveGoalSignupCopyVariant(now ?? new Date());
+	return resolveGoalSignupCopyVariant(now ?? getGoalSignupDateOverride() ?? new Date());
```
dyersituations, line 72: It looks like `now` is never passed into this, correct? I was confused at first assuming that `now` would always be the default here. Do you think it would make sense to remove `now` from this to prevent future confusion?

src/util/goalSignupCopyVariant.js
```diff
@@ -13,9 +13,35 @@ export const GOAL_SIGNUP_COPY_VARIANT = {
...
 };
 
+export const GOAL_SIGNUP_DATE_QUERY_PARAM = 'goalSignupDate';
+
 // April (Date#getMonth is 0-indexed) is the earliest 'no-goal-yet' month.
 const NO_GOAL_YET_START_MONTH = 3;
 const NO_GOAL_YET_START_DAY = 1;
+const GOAL_SIGNUP_DATE_PATTERN = /^(\d{4})-(\d{2})-(\d{2})$/;
+
+function parseGoalSignupDateParam(value) {
```
dyersituations, line 23: Please add a comment to the ticket tagging Lauren for how to use this query param for testing the feature.

dyersituations (approved):
@infante-jaime looks good. The minor refactor comment can be ignored, but do please add comment to the ticket for manually testing the different copies throughout the year. Thanks!

src/util/goalSignupCopyVariant.js
infante-jaime: Correct, production callers don't pass now. It's only there to make the date-boundary tests easier without relying on fake timers or changing global time. Runtime behavior is unchanged.

src/util/goalSignupCopyVariant.js
infante-jaime: done

PR #6965, opened 2026-06-08
feat: integrate goal row in my kiva while adjusting next steps logic

christian14b:
https://kiva.atlassian.net/browse/MP-2869

image

- Add Goal Row to My Kiva
- Handles scenarios where Goal Row is active so Goal Card should not be visible in next steps, but the logic in the card sort should remain.

### DIFF
--- a/src/components/MyKiva/FeaturedGoalCard.vue
+++ b/src/components/MyKiva/FeaturedGoalCard.vue
@@ -1,7 +1,7 @@
 <template>
 	<KvLoadingPlaceholder
 		v-if="loading"
-		class="featured-goal-card__loading tw-rounded-lg tw-w-full"
+		class="featured-goal-card__loading !tw-rounded-lg tw-w-full"
 		aria-busy="true"
 	/>
 	<div
@@ -149,13 +149,14 @@
 </template>
 
 <script setup>
-import { computed } from 'vue';
+import { computed, ref, watch } from 'vue';
 import {
 	KvButton, KvLoadingPlaceholder, KvProgressCircle, KvUtilityMenu,
 } from '@kiva/kv-components';
 import KvIcon from '#src/components/Kv/KvIcon';
 import goalCopy from '#src/util/goalCopy';
 import { GOALS_CURRENT_YEAR } from '#src/composables/useGoalData';
+import { showConfetti } from '#src/util/animation/confettiUtils';
 
 const HALF_GOAL_THRESHOLD = 50;
 const COMPLETED_GOAL_THRESHOLD = 100;
@@ -193,6 +194,10 @@ const props = defineProps({
 	userName: {
 		type: String,
 		default: '',
+	},
+	suppressCompletionConfetti: {
+		type: Boolean,
+		default: false,
 	},
 });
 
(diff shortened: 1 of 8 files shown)


src/components/MyKiva/FeaturedGoalCard.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<KvLoadingPlaceholder
 		v-if="loading"
```
dyersituations, line 3: I can't comment on the line, but the button is too short, it should the the usual size. For example, in desktop it's currently 40px tall instead of 48px like other buttons on the page.

dyersituations: Same for "work towards your goal" button for in-progress.

dyersituations:
Could be a follow-up, but we shouldn't be showing edit on the completed goal version:

image

src/components/MyKiva/FeaturedGoalCard.vue
dyersituations: Also applies to "View your achievements" variant.

mcstover: Agree we should not be changing button dimensions at this time.

christian14b: Good catch! I though 40 was the expected on figma but that could have changed..

christian14b:
> Could be a follow-up, but we shouldn't be showing edit on the completed goal version:
> 
> image

image

dyersituations:
Do we currently prefetch the goal data? It would be nice to not show these loaders if the component won't be rendered on pageload. This could be a follow-up. Thoughts?

image

christian14b:
> Do we currently prefetch the goal data? It would be nice to not show these loaders if the component won't be rendered on pageload. This could be a follow-up. Thoughts?
> 
> image

yeah, it might be good to do that. I would like to do it as a follow-up in case there's an impact in the control variant.

test/unit/specs/components/MyKiva/MyKivaFeaturedSlot.spec.js
```diff
@@ -0,0 +1,219 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
dyersituations, line 1: Do we have tests somewhere that cover all of the copy use cases in Figma? If not, could you create a follow-up ticket to handle that?

https://www.figma.com/design/ARJWLwdnFJ84FNSvPX4wYz/Goals-V2?node-id=6350-40319&t=azIwWQGjduiieEwx-4

dyersituations:
> , it might be good to do that. I would like to do it as a follow-up in case there's an impact in the co

Sounds good, could you add a follow-up ticket for that with the details?

dyersituations (approved):
Two follow-up tickets to create @christian14b, but otherwise looks great!

test/unit/specs/components/MyKiva/MyKivaFeaturedSlot.spec.js
christian14b: There should be a ticket for the copy rotation depending on the date we talked about.. I'll add this work to that ticket

dyersituations: Sounds good if that's related to the same epic.

dyersituations (approved):
Thanks for adding the test cases!

PR #6968, opened 2026-06-09
feat: add remaining stats to loans beta page and missing graph bg on rows

dyersituations:
https://kiva.atlassian.net/browse/MP-2731

Adding missing stats that weren't being displayed yet in the loans beta page. Now all fields have data being displayed in the beta page (yay!). Also added some gray background on alternating rows to match legacy (more formatting/style improvements to come in the future).

image

image

### DIFF
--- a/src/components/Portfolio/LoanStatsTable.vue
+++ b/src/components/Portfolio/LoanStatsTable.vue
@@ -96,19 +96,19 @@ export default {
 			avgStats: {},
 			loading: true,
 			statsRows: [
-				{ label: 'Amount lent', key: 'amount_of_loans', type: 'currency' },
+				{ label: 'Amount lent', key: 'amount_lent', type: 'currency' },
 				{ label: 'Amount repaid', key: 'amount_repaid', type: 'currency' },
-				{ label: 'Amount lost', key: 'amount_defaulted', type: 'currency' },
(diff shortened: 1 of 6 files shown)


src/components/Portfolio/LoanStatsTable.vue
```diff
@@ -36,8 +36,12 @@
 		</tbody>
 
 		<tbody v-else>
-			<tr v-for="row in statsRows" :key="row.key">
-				<td class="tw-p-1">
+			<tr
+				v-for="row in statsRows"
+				:key="row.key"
+				:class="{ 'tw-bg-white': row.showInWhite, 'tw-bg-gray-50': row.showInGray }"
```
roger-in-kiva, line 42: Is index not working here as well? Probably a way to avoid a hardcoded data?

cristhianDt: +1

mcstover (approved):

src/components/Portfolio/LoanStatsTable.vue
dyersituations: Unfortunately the rows aren't just alternating, so using index doesn't quite work, but good question!

Image

roger-in-kiva (approved):

christian14b (approved):

PR #6971, opened 2026-06-09
fix: add goal row section if it is going to be visible and tracking a…

christian14b:
…djustments

- Loading section will only be visible if goal is not complete (section will be hidden)
- Few styles adjustment in placeholder to avoid jumps on load
- Tracking information

### DIFF
--- a/src/pages/MyKiva/MyKivaPageContent.vue
+++ b/src/pages/MyKiva/MyKivaPageContent.vue
@@ -343,8 +343,7 @@ export default {
 			type: Boolean,
 			default: false,
 		},
-		// When false, the featured slot would only render to hide itself,
-		// so we skip mounting it
+		// Flag to indicate if the goal row should be rendered (loading and loaded state)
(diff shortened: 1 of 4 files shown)


cristhianDt (approved):

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -343,6 +343,12 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		// When false, the featured slot would only render to hide itself,
```
dyersituations: This comment is confusing, what does it mean specifically in real use cases?

christian14b: Fair enough, in simple words is: 'Make the whole goal row visible (loading and loaded state)'

christian14b: I would update this

dyersituations (approved):

PR #6967, opened 2026-06-09
feat: admin reward tip helper

roger-in-kiva:
## Summary

Foundational helper for epic [MP-2829](https://kiva.atlassian.net/browse/MP-2829), unblocks UI ticket [MP-2840](https://kiva.atlassian.net/browse/MP-2840). Adds a feature-flagged `isAdminRewardTipEligible` helper so the optional tip section can later be shown on Admin Reward lending-credit checkouts.

Ticket: [MP-2839](https://kiva.atlassian.net/browse/MP-2839)

## What changed

- **Flag plumbing** — adds `uiConfigSetting(key: "enable_tip_for_admin_reward_lc")` to `initializeCheckout.graphql` and surfaces it on `CheckoutPage` as `enableAdminRewardTipFlag` via `readBoolSetting`.
- **Helper** — new `isAdminRewardTipEligible(promoData, flagEnabled)` in `src/util/promoCredit.js` next to `hasPromoSession`. Returns `true` only when the flag is `=== true` **and** `promoData.managedAccount.managementType === 'Admin Reward'`.
- **Computed** — `CheckoutPage.vue` exposes `adminRewardTipEligible` so MP-2840 reads a single boolean to gate the UI.
- **No schema change** — `managementType` is already on `ManagedAccount` (already requested by `promoCampaign.graphql`).

[MP-2829]: https://kiva.atlassian.net/browse/MP-2829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2840]: https://kiva.atlassian.net/browse/MP-2840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2839]: https://kiva.atlassian.net/browse/MP-2839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6977, opened 2026-06-10
feat: render optional tip section in admin reward LC checkout variant b (MP-2840)

roger-in-kiva:
## Update

- Moves donations array sync into refreshTotals (gated by
  isStopHidingTipExpEnabled), so the in-memory donations match the
  latest basket fetch when the experiment is active.
- Simplifies ensureTipDonationExists to delegate donation updates to
  refreshTotals instead of patching the array from the mutation result.
- Drops isUserEdited from the updateDonation mutation selection (no
  longer consumed; basket-side reads come from shopBasketUpdate).
- Updates unit tests to assert refreshTotals is invoked on success
  rather than asserting the donations array was patched directly.

## Summary

**This is a first iteration to test on dev and get some feedback.**

Wires the `stop_hiding_tip_campaign` experiment (MP-2852) into the LC checkout UI. When an eligible Admin Reward lender is bucketed into variant `b`, the frontend injects an `isTip: true` Donation into the basket so the existing `DonationItem` renders the standard tip section. For variant `a` (control), `reward_credit` LC checkouts continue to suppress the tip exactly as today.

- Ticket: [MP-2840](https://kiva.atlassian.net/browse/MP-2840)
- Parent epic: [MP-2829](https://kiva.atlassian.net/browse/MP-2829) — Stop hiding tip - LC Campaigns
- Depends on: [MP-2839](https://kiva.atlassian.net/browse/MP-2839) (helper, #6967) · [MP-2852](https://kiva.atlassian.net/browse/MP-2852) (bucketing, #6970)
- Followed by: [MP-2841](https://kiva.atlassian.net/browse/MP-2841) (Playwright)

## Why a frontend mutation rather than a `v-if`

Tip suppression for `reward_credit` LC checkouts is structural — the backend omits the `isTip: true` Donation from `shop.basket.items.values`. `BasketItemsList` renders whatever donations are in the basket; there is no UI-side gate to flip. To show the tip, the frontend must inject a tip Donation. We use `setDonationAmount` from `#src/util/basketUtils`, the same `updateDonation` wrapper `DonationItem` itself uses, so backend upsert semantics on `isTip` are consistent.

## Behavior

- **Gate:** double-gated by the backend setting `enable_tip_for_admin_reward_lc` (already wired in MP-2839) AND the variant `b` assignment from `stop_hiding_tip_campaign` (MP-2852). Flipping the backend flag off short-circuits in `getPromoInformationFromBasket` before the experiment is initialized — kill-switch behavior is clean.
- **Default amount:** `TIP_PERCENTAGE` × (`loanReservationTotal` + `kivaCardTotal`), quantized to 2 decimals.
- **Idempotency:** `ensureTipDonationExists` skips the mutation when an existing tip already has a non-zero price (preserves user-entered tips across callback re-fires from `?setuiab`) OR already equals the computed amount.
- **Empty-basket race:** the method short-circuits when `emptyBasket` is true. An `emptyBasket` watcher fires `ensureTipDonationExists` on the `true → false` transition so users who land empty get the tip on their first add.
- **Local cache hydration:** on success we splice/push the returned Donation into `this.donations` because `refreshTotals()` only refetches the totals fragment and `setDonationAmount` doesn't run `refetchQueries`.
- **Loading state:** `setUpdatingTotals(false)` is reset on the no-data and `.catch` paths so a failed/empty mutation doesn't leave the basket UI stuck.
- **No backend, GraphQL schema, `DonationItem`, or `BasketItemsList` changes.**

## Test plan

- [ ] **Path A — non-promo checkout** renders the tip section as today (no regression).
- [ ] **Path B — `reward_credit` LC, backend flag OFF** suppresses the tip (no regression).
- [ ] **Path C — `reward_credit` LC, flag ON, variant `a`** suppresses the tip.
- [ ] **Path D — `reward_credit` LC, flag ON, variant `b`** shows the tip with the default; editing updates totals; checkout completes with the tip; checkout also completes with $0 tip.
- [ ] **Force the variant** with `?setuiab=stop_hiding_tip_campaign.b` and confirm the callback double-fire does not stack tip donations or reset `isUserEdited`.
- [ ] **Empty-basket flow:** land with empty basket in variant `b`, add a loan via the upsell module, confirm one tip injection with the proper amount (not $0).
- [ ] **Breakpoints:** path D at mobile / tablet / desktop / wide — layout matches the standard non-promo tip.
- [ ] Local: `npm run lint` (0 errors) and `npm run unit` (3806 tests pass, 0 failures) ✅

[MP-2840]: https://kiva.atlassian.net/browse/MP-2840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2829]: https://kiva.atlassian.net/browse/MP-2829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2839]: https://kiva.atlassian.net/browse/MP-2839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2852]: https://kiva.atlassian.net/browse/MP-2852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MP-2841]: https://kiva.atlassian.net/browse/MP-2841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1373,6 +1378,51 @@ export default {
 				'basket',
 			);
 		},
+		ensureTipDonationExists() {
```
dyersituations, line 1387: Can/should we add some testing around this new method?

roger-in-kiva: I'll try!

roger-in-kiva: Done.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -705,6 +707,9 @@ export default {
 				]);
 				this.getUpsellModuleData();
 			}
+			if (!newValue && this.isStopHidingTipExpEnabled) {
```
dyersituations, line 710: When does `emptyBasket` trigger this `watch`?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -379,6 +380,7 @@ const BANDIT_UPSELL_EXP_KEY = 'checkout_bandit_upsell_enable';
 const EXPIRING_SOON_EXP_KEY = 'checkout_expiring_soon_upsell';
 const KIVA_CREDIT_REPLACEMENT_EXP_KEY = 'checkout_kiva_credit_copy_replacement';
 const STOP_HIDING_TIP_EXP_KEY = 'stop_hiding_tip_campaign';
+const TIP_PERCENTAGE = 0.2;
```
dyersituations, line 383: Do we have a tip percentage anywhere else in the FE? Just making sure we have a single point of truth.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1373,6 +1378,51 @@ export default {
...
+			const loanTotal = numeral(this.totals?.loanReservationTotal).value() ?? 0;
+			const kivaCardTotal = numeral(this.totals?.kivaCardTotal).value() ?? 0;
+			const donationAmount = Math.round((loanTotal + kivaCardTotal) * TIP_PERCENTAGE * 100) / 100;
+			const existingTip = this.donations.find(d => d.isTip);
+			if (existingTip) {
+				const existingPrice = numeral(existingTip.price).value() ?? 0;
+				if (existingPrice !== 0 || existingPrice === donationAmount) return;
+			}
+			this.setUpdatingTotals(true);
+			setDonationAmount({ apollo: this.apollo, donationAmount })
```
dyersituations, line 1398: I'm not seeing a check for admin reward specifically in this method, does this donation get always added now when the experiment is enabled?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1373,6 +1378,51 @@ export default {
...
 		},
+		ensureTipDonationExists() {
+			if (this.emptyBasket) return;
+			const loanTotal = numeral(this.totals?.loanReservationTotal).value() ?? 0;
+			const kivaCardTotal = numeral(this.totals?.kivaCardTotal).value() ?? 0;
+			const donationAmount = Math.round((loanTotal + kivaCardTotal) * TIP_PERCENTAGE * 100) / 100;
+			const existingTip = this.donations.find(d => d.isTip);
+			if (existingTip) {
+				const existingPrice = numeral(existingTip.price).value() ?? 0;
+				if (existingPrice !== 0 || existingPrice === donationAmount) return;
```
infante-jaime, line 1395: Do we need to consider the user-edited $0 tip case.

Right now the code can't tell that case from the initial $0 — if a lender sets the tip to $0 and reloads the page, the backend returns isUserEdited: true on that donation, and the previous chioise gets reseted  on every reload.

We do have  _isUserEdited_ in this.donations , so we can just add the condition:

if (existingTip.isUserEdited || existingPrice !== 0 || existingPrice === donationAmount) return;

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: We don't have it elsewhere in FE. This is driven by BE side.
In fact, if this test is a win, this should be fixed in BE to avoid 0 as default tip for promo cases.

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: If you reach checkout page with an empty basket, then add a loan to the basket from there, we are triggering this to ensure tip applies for the test.

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: this method is triggered in line 1435, for test group only.

src/pages/Checkout/CheckoutPage.vue
roger-in-kiva: Unfortunately not for this case.
Default tip is 20% and `isUserEdited = false`.
But for promo case, BE is updating to `price = 0` and `isUserEdited = true`, so if we check for that this will become a dead code for the test.

src/pages/Checkout/CheckoutPage.vue
dyersituations: Right, but does it only target this specific reward type?

roger-in-kiva: `initializeStopHidingTipExperiment` function is called in line 1074 for reward type users only.

src/pages/Checkout/CheckoutPage.vue
dyersituations: So the tip could be edited to $0? Will that be an issue like @infante-jaime described?

roger-in-kiva: Yeah, user could edit to $0 in the session, but if refresh page will be re-calculated based on 20%.
`updateDonation` mutation don't have a way to set `isUserEdited = false` at this point, should be ideal, but I think for the test is enough.
If a win, this feature will be ported to BE side as current.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1373,6 +1378,51 @@ export default {
...
+			this.setUpdatingTotals(true);
+			setDonationAmount({ apollo: this.apollo, donationAmount })
+				.then(result => {
+					const updated = result?.data?.shop?.updateDonation;
+					if (!updated) {
+						this.setUpdatingTotals(false);
+						return;
+					}
+					const existingIdx = this.donations.findIndex(d => d.isTip);
+					if (existingIdx >= 0) {
```
mcstover: Seems like this whole _if/else_ block should just be a basket refresh call to get the actual server state of the donation...

roger-in-kiva: We are not currently refreshing tip for edited ones, do we want that?

mcstover: It seems risky to be manipulating basket state only in the client without persisting and reflecting that state from the server. I think we are persisting it to the server if we're calling setDonatingAmount which would call a real mutation...

roger-in-kiva: Oh I see.
We are calling a mutation in line 1392 via `setDonationAmount` function.
Based on that result we are updating client side.

dyersituations: @mcstover any further concerns on this thread related to updating the basket?

mcstover: I don't think we should be updating `this.donations` with anything but server state as it's used by the overall basket state. What is the need here for manually re-building an apollo response type object instead of fetching latest basket state of donations and updating with it?

mcstover: I suspect the manual changes to just the local property would mean we're out of sync with the apollo state/cache for the page.

roger-in-kiva: Nice call, thanks!

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1373,6 +1384,30 @@ export default {
...
+				if (existingPrice !== 0 || existingPrice === donationAmount) return;
+			}
+			this.setUpdatingTotals(true);
+			setDonationAmount({ apollo: this.apollo, donationAmount })
+				.then(result => {
+					if (!result?.data?.shop?.updateDonation) {
+						this.setUpdatingTotals(false);
+						return;
+					}
+					this.refreshTotals();
```
mcstover, line 1404: This is how we update most other aspects of the basket as well. Please verify your donation updates are occurring properly.

roger-in-kiva: Ok.
I tested from UI and programmatically then donation updates work fine.

mcstover: awesome, thanks for moving back to the server/apollo synced data state!

mcstover (approved):

dyersituations (approved):

PR #6980, opened 2026-06-11
feat: [MP-2900-2905] - allow removal promotion rows in express checkout and prevent close modal

cristhianDt:
# MP-2900-2905

## Summary

1. allows removal promotion rows in express checkout modal
2. initialize checkout to apply promotion balance
3. prevent closing modal when user is paying

https://www.figma.com/design/ARJWLwdnFJ84FNSvPX4wYz/Goals-V2?node-id=5831-98211&t=QSdzKl5WSyTatxBr-0

- https://kiva.atlassian.net/browse/MP-2900
- https://kiva.atlassian.net/browse/MP-2905

## How test it

- 'kiva credit', add some credit to the user.
- - login in our admin site and go to the user page: https://admin.development.kiva.org/admin/viewUser?user_id=XXXXX
- - click on '[Change Credit](https://admin.development.kiva.org/admin/changeCredit?user_id=XXXXX)' 
- - add new credit amount using the type **'m_credit'**, any value works

- 'promotion', there are two types 'bonu' and 'kiva promotion', in other words free credit and kiva card redemption 
- - for free credit: login in our admin site and go to the user page > click on [Add Free Credit](https://admin.development.kiva.org/admin/viewUserLenderPromoCredits?user_id=6356035) > add new free credit amount using the type **'147 - kiva'**, any value works
- - buy a new kiva card then redeem the card

## Evidences

Captura de pantalla 2026-06-11 a la(s) 13 50 33


src/util/thanksPage/expressCheckoutUtils.js
```diff
@@ -55,3 +57,30 @@ export async function clearBasketDonation({ apollo, basketId, donation }) {
...
+ *
+ * Mirrors the `OrderTotals` → `CheckoutPage.refreshTotals` pattern: the
+ * mutation alone does not return totals, so the cache must be invalidated by
+ * a `network-only` re-fetch of the watched query.
+ *
+ * Throws if either the mutation or the refetch returns GraphQL errors so the
+ * caller can surface a tip message.
+ */
+export async function removeBasketCredit({ apollo, basketId, creditType }) {
+	const { errors } = await apollo.mutate({
```
mcstover, line 78: We also have this available from kv-shop https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/basketCredits.ts#L32

mcstover: https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/validatePreCheckout.ts#L7 may also be useful as you get closer to the actual checkout.

cristhianDt: good point, yeah I'm trying to get the same or similar flow in checkout, I will check how we are using the validate and what point there, thanks matt

cristhianDt: hey @mcstover , I can't use your recomendation https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/basketCredits.ts#L32, because that method is not using the `creditType` and I need to specify the promo to remove, we could add a new ticket to update it later, what do you think?

dyersituations: Thoughts on adding the parameter now? It would only take a few mins, and the package deploys quickly.

cristhianDt: @dyersituations @mcstover  in that case I prefer to use and update this method https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/basketCredits.ts#L94 instead of `removeKivaCredit`, I just need to remove the variable 'redemptionCode' we don't need it anymore in our mutation, agree?

sandbox

dyersituations: Those two methods are pretty similar, and it looks like `removeKivaCredit` would be simpler to change for the purpose you need, correct?

cristhianDt: yeah, but based on the name, it makes more sense when you only want to remove the 'kiva credit' `removeKivaCredit`; the other `removePromoCredit` makes more sense for any kind of promotion you want to remove. so I would leave 'removeKivaCredit' as it's and make a quick update for the other

dyersituations: The problem though is the other method is designed around the redemption, and we don't have to break other use cases. I'm starting to wonder if we just leave the code in this PR as is for now, since this is still an experiment of sorts anyways...

cristhianDt: well correct me if I'm wrong, but that mutation doesn't work for me. as I shared t in the previous screenshot, the documentation for our sandbox no longer expects or uses that argument, and therefore the query is failing; the method is failing for me

cristhianDt: sure I agree but if we could kill two birds with one stone, I would like to fix that method in once as well XD

dyersituations: I think the problem is that I don't know if that method is technically wrong, it was built for a different use case.

mcstover: Sorry for that mis-link above. There is a method to remove promo credit as well: https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/basketCredits.ts#L94 Not sure if this one supports your new `admin_credit` type... if it doesn't that would be something to tweak in the php to ensure it allows ALL promo types...

cristhianDt: dont worries and that one was that I shared in my comment https://github.com/kiva/ui/pull/6980#discussion_r3421853548 and yeah, looks like the php support all of them

cristhianDt: in the end both methods calls the same mutation, the first one without 'type' argument and the second one with an expected argument 'redeemptionCode'

image

cristhianDt: BTW - @mcstover a new admin_credit type... maybe I missing something haha but I don't know about new type, sorry

mcstover: Yep, have you found this method to work when passing that type?

cristhianDt: no, in fact I'm not including a new type 🤔... but when I try the other types works correctly **if I remove the argument 'redeemptionCode'** (that's the update that needs to be made according to my suggestion)

https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-shop/src/basketCredits.ts#L109-L112

```javascript
	shop (basketId: $basketId) {
		id
		removeCreditByType(creditType: $creditType, redemptionCode: $redemptionCode)
	}
```

mcstover: did you try to pass `null` there?

It's true that UPC and redemption codes do have a value we can pass there... wonder what the php is doing within this mutation when it see's the new "admin_credit" type...

mcstover: It's ok to just skip using the global method if it's not working. We also use that mutation without any value in the ui version of checkoutUtils... so if that's working cool

cristhianDt: yes I did, I get the error: the mutation does not expect to receive `redemptionCode` argument

dyersituations: I'm thinking we could use the mutation as already defined in this PR since it's relatively simple, and this feature is part of an experiment. It would definitely be good to move this entire feature (express checkout) into the component library if/when the experiment does well.

cristhianDt: agree, updating the global or using my new one, I'll just drop a comment in case we fix the global method and we could start using it, 

I'll let you know @mcstover when I add `validatePreCheckout` query in this flow 

thanks team!

dyersituations (commented):
@cristhianDt changes are looking good. I'll functionally test once the `validatePreCheckout` method is in place and the dev DB sync is complete.

src/plugins/checkout-utils-mixin.js
```diff
@@ -22,28 +22,13 @@ export default {
 		validateBasket() {
 			checkInjections(this, injections);
 
-			return new Promise((resolve, reject) => {
```
cristhianDt, line 25: @mcstover @dyersituations I moved this logic to this util file > `src/util/checkout/checkoutValidationUtils.js`, shared by this mixin and the new modal, lmk any doubt or comment, thanks

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
```diff
@@ -170,6 +174,16 @@ const onSubmit = async () => {
 	paying.value = true;
 
 	try {
+		const validationStatus = await validatePreCheckoutBasket({ apollo });
+		if (validationStatus !== true) {
+			const validationMessage = formatPreCheckoutValidationErrors(validationStatus);
+			logFormatter(`ExpressCheckoutModal validation failed: ${validationMessage}`, 'error');
+			paying.value = false;
+			closeLightbox();
+			router.push('/basket');
```
dyersituations, line 183: I'm now seeing an error initially or a redirect to `/basket` when I test the latest changes locally.

cristhianDt: do you see any message in the console?, could you share a screenshot please

the idea is: if something happen during the verification step we redirect so I would like to check whether the reservation service or checkout service is failing again

did you try in the checkout page too?

dyersituations: Yes the checkout page, testing express checkout. No specific error, but now I just get redirected to basket each time, meaning errors are happening. Please see if you can reproduce locally.

cristhianDt: on it

cristhianDt: @dyersituations I tested with 2 users and both pages are working correctly for those users, what is your user id?, let me try with your user

dyersituations: Looks like transient dev errors, checkouts worked when I tested again.

dyersituations (approved):
Looks good. Make sure to validate the `checkout-utils-mixin.js` extra once these changes are on dev. We don't want to risk affecting checkout in general.

src/components/Thanks/ExpressCheckout/ExpressCheckoutModal.vue
cristhianDt: great!, thank you

cristhianDt:
> Looks good. Make sure to validate the `checkout-utils-mixin.js` extra once these changes are on dev. We don't want to risk affecting checkout in general.

absolutely, I'll do it

I'll wait for @mcstover review to get an extra approval, and then I'll merge the PR, thanks!

mcstover (approved):

PR #6981, opened 2026-06-11
feat: route guard added to test

roger-in-kiva:
- route guard added to test on dev under flux
- https://kiva.atlassian.net/browse/MP-2906

### DIFF
--- a/src/router/routes.js
+++ b/src/router/routes.js
@@ -347,7 +347,18 @@ export default [
 	},
 	{
 		path: '/lend/:category',
-		redirect: '/lend-by-category/:category'
+		beforeEnter(to) {
+			const url = to.params.category
+				? `/lend-by-category/${to.params.category}`
+				: '/lend-by-category';
+			if (typeof window !== 'undefined') {
+				window.location.href = url;
+			} else {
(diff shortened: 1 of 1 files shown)


src/router/routes.js
```diff
@@ -347,7 +347,18 @@ export default [
...
+		beforeEnter(to) {
+			const url = to.params.category
+				? `/lend-by-category/${to.params.category}`
+				: '/lend-by-category';
+			if (typeof window !== 'undefined') {
+				window.location.href = url;
+			} else {
+				// SSR: triggers res.redirect in server/vue-middleware.js
+				// eslint-disable-next-line no-throw-literal
+				throw { url, code: 301 };
```
dyersituations, line 359: What happens in this situation? This URL won't exist in UI, so if it can't redirect on the window object, I'm not sure what happens.

roger-in-kiva: As throwing an error, redirect under this function should be triggered:
https://github.com/kiva/ui/blob/d827e9f3c7c5de3eaa69dee8f3a28e5a831bd874/server/vue-middleware.js#L16

dyersituations (approved):

PR #6984, opened 2026-06-12
feat: add completed goal to impact progress row

christian14b:
https://kiva.atlassian.net/browse/MP-2876
- Following same design for goal card just changing the year.

image


dyersituations (approved):
Looks good! @christian14b could you find/create an account for testing this on dev with fake auth and share it in the ticket comments for VQA?

PR #7003, opened 2026-06-17
feat: suppress achievement messages when goal is active

christian14b:
https://kiva.atlassian.net/browse/MP-2875

- If no goal set, achievements are visible in ATB and checkout pill.
- When user has an active goal, all achievements messages from ATB and Checkout will be hidden until the experience is over. 

### How to test
- Set a goal, all atb messages should be related to the goal
- Do not set a goal or remove your active goal, atb messages related to achievements will appear

### DIFF
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -66,6 +66,7 @@ const {
 	loadGoalData,
 	getPostCheckoutProgressByLoans,
 	isProgressCompletingGoal,
+	suppressAchievementNudges,
 } = useGoalData({ apollo });
 
 const userData = ref({});
@@ -81,6 +82,12 @@ const milestonesProgress = ref({});
 const hasEverLoggedIn = ref(false);
 const basketTotal = ref(0);
 const loanGoalProgress = ref(0);
+// Tracks whether the just-added loan actually contributes to the goal's
+// category (e.g. a male loan does not contribute to a women-equality goal).
+// loanGoalProgress alone is the cumulative yearly total and stays > 0 across
+// non-contributing adds, so we need this flag to correctly identify a goal
+// add vs. an unrelated achievement add.
+const loanContributesToGoal = ref(false);
 
 const basketCount = computed(() => {
 	return addedLoan.value?.basketSize ?? 0;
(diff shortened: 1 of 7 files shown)


src/components/WwwFrame/Header/KvAtbModalContainer.vue
```diff
@@ -204,7 +209,9 @@ const fetchPostCheckoutAchievements = async loanIds => {
 
 	// If added loan is not related to user goal, proceed with achievements logic.
 	// This condition will prevent any conflict between goal and achievement messages.
-	if (!isLoanGoal.value) {
+	// While the lender has an active goal experience we also suppress achievement
+	// nudges entirely so they don't compete with the goal flow.
+	if (!isLoanGoal.value && !suppressAchievementNudges.value) {
```
dyersituations: How does `isLoanGoal` differ from `suppressAchievementNudges`?

src/composables/useGoalData.js
```diff
@@ -727,6 +727,17 @@ export default function useGoalData({ apollo } = {}) {
...
 
+	/**
+	 * Drives the basket / ATB-modal achievement
+	 * nudges so they stay out of the way during the goal experience and
+	 * resume only after the viewed-goal-complete flag is set (or the lender
+	 * has no goal at all).
+	 */
+	const suppressAchievementNudges = computed(() => {
+		const hasGoal = !!userGoal.value && Object.keys(userGoal.value).length > 0;
+		return hasGoal && !hasViewedCompletedGoalForYear(GOALS_CURRENT_YEAR);
```
dyersituations: Shouldn't we check if goal is completed instead of `hasViewedCompletedGoalForYear`?

src/components/WwwFrame/Header/KvAtbModalContainer.vue
christian14b: True, this can be simplified

src/composables/useGoalData.js
christian14b: This one is because of the ticket requirement: 

`After the "viewed goal-complete" session flag is set (the flag introduced in MP-2869 — lender has seen goal-complete on both TY and MyKiva), allow the achievement ATB modal and checkout banners to resume normally.`

dyersituations: Won't this mean though that achievement pills won't be displayed after a user completes their goal until the user returns to MyKiva?

christian14b: Yeah, it's true they will remain hidden until then. I can remove the requirement from the ticket and check if goal is completed instead, if that's fine

dyersituations: Yeah I think it makes sense to check goal completed instead, thanks!

dyersituations (approved):

PR #7004, opened 2026-06-17
feat: restore team reassignment parity to loans-beta

dyersituations:
https://kiva.atlassian.net/browse/MP-2755

Apparently you can reassign team attribution on loans page. That functionality is now migrated in the API, and here's the associated FE work to wire it up in the loans table.

image

image


src/components/Portfolio/LoanList.vue
```diff
@@ -277,6 +315,43 @@ export default {
...
 		},
+		canReassignTeam(loan) {
+			// Legacy parity: the dropdown only appears when the loan is eligible AND the
+			// user actually belongs to at least one team to reassign to. With no teams there
+			// is nothing to pick (and no "None" detach), so the cell stays read-only.
+			return Boolean(loan.userProperties?.canChangeTeamAssignment) && this.lendingTeams.length > 0;
+		},
+		currentTeamId(loan) {
+			const team = loan.userProperties?.userAttributedTeam;
+			return team ? String(team.id) : '';
```
roger-in-kiva, line 326: Instead that team is defined above, is any case id is not?

eddieferrer (approved):

src/components/Portfolio/LoanList.vue
dyersituations: The `id` is guaranteed not null by the graphql type/return.

PR #7011, opened 2026-06-19
feat: make visible previous years goals with progress in impact row

christian14b:
https://kiva.atlassian.net/browse/MP-2888

image

### DIFF
--- a/src/components/MyKiva/BadgesSectionV2.vue
+++ b/src/components/MyKiva/BadgesSectionV2.vue
@@ -50,7 +50,7 @@ import useBadgeData from '#src/composables/useBadgeData';
 import { KvCarousel, KvLoadingPlaceholder } from '@kiva/kv-components';
 import MyKivaProgressCard from '#src/components/MyKiva/MyKivaProgressCard';
 import { useRouter } from 'vue-router';
-import { COMPLETED_GOAL_THRESHOLD, GOALS_CURRENT_YEAR } from '#src/composables/useGoalData';
+import { COMPLETED_GOAL_THRESHOLD, GOALS_CURRENT_YEAR, GOAL_STATUS } from '#src/composables/useGoalData';
 
 const CARD_MIN_HEIGHT = '111px';
 const SINGLE_SLIDE_WIDTH = '336px';
@@ -99,6 +99,9 @@ const userHasGoal = computed(() => !!userGoal.value && Object.keys(userGoal.valu
 
 const formatHistoricalGoal = goal => {
 	const year = goal.dateStarted ? new Date(goal.dateStarted).getFullYear() : null;
+	const historicalProgress = goal.status === GOAL_STATUS.COMPLETED
+		? (goal.target || 0)
+		: (goal.loansTowardGoal || 0);
 	return {
 		id: `annual-goal-${year}`,
 		goal: {
@@ -108,7 +111,7 @@ const formatHistoricalGoal = goal => {
 			nextAchievementAt: null,
 			totalLoans: null,
 		},
-		goalProgress: goal.target || 0,
+		goalProgress: historicalProgress,
 		isAnnualGoal: true,
 		isHistoricalGoal: true,
 		year,
--- a/src/components/MyKiva/BadgesSectionV2.vue
+++ b/src/components/MyKiva/BadgesSectionV2.vue
@@ -27,6 +27,7 @@
 					v-if="badge?.goal"
 					:goal="badge.goal"
 					:is-annual-goal="badge.isAnnualGoal"
+					:is-historical-goal="badge.isHistoricalGoal"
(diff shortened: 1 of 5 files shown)


dyersituations (commented):
@christian14b please post a comment on this PR when the latest tweaks from design feedback are ready.

christian14b:
@dyersituations this is ready for review

src/composables/useGoalData.js
```diff
@@ -844,6 +849,47 @@ export default function useGoalData({ apollo } = {}) {
 		}
 	}
 
+	// Fills loansTowardGoal for expired prior-year goals that predate the field.
+	// Only persist when there's actual progress to surface — zero-progress goals
+	// are already hidden by the completedGoalsHistory filter.
+	async function backfillExpiredGoalProgress(parsedPrefs) {
```
dyersituations, line 855: Why is this backfill needed? What specific situations?

src/composables/useGoalData.js
```diff
@@ -915,35 +964,49 @@ export default function useGoalData({ apollo } = {}) {
...
 		}
 
-		// Renew goals every following year
 		const hadCompletedGoal = goals.some(g => {
 			const goalYear = g.dateStarted ? new Date(g.dateStarted).getFullYear() : null;
 			return goalYear < currentYear && g.status === GOAL_STATUS.COMPLETED;
 		});
-		const expiredGoals = goals.map(goal => {
+
+		const renewedResults = await Promise.all(goals.map(async goal => {
```
dyersituations: What are the changes to this method fixing?

src/components/MyKiva/BadgesSectionV2.vue
```diff
@@ -27,6 +27,7 @@
 					v-if="badge?.goal"
 					:goal="badge.goal"
 					:is-annual-goal="badge.isAnnualGoal"
+					:is-historical-goal="badge.isHistoricalGoal"
```
dyersituations, line 30: Related to the conversation in Jira, are the two different titles expected as we have them now:

Image

src/composables/useGoalData.js
christian14b: Currently, there's no loans made in past goals within user preferences, that's why I needed to fetch this information. It wasn't necessary for completed goals since loans made >= target was always happening

src/components/MyKiva/BadgesSectionV2.vue
christian14b: Let me check again. I thought the new copy only target was previous goals. Current goal will remain the same

src/composables/useGoalData.js
christian14b: It's getting loans made for previous years so userPreferences can have this information. Looking at it twice, i think it's confusing, I'll rethink this logic to make it simpler

src/composables/useGoalData.js
dyersituations: Ah for the circle progress, got it, thanks for the info!

src/components/MyKiva/BadgesSectionV2.vue
christian14b: Now, it's standarized, thanks for checking
image

src/composables/useGoalData.js
```diff
@@ -915,35 +964,47 @@ export default function useGoalData({ apollo } = {}) {
...
+			goals
+				.filter(g => {
+					const goalYear = g.dateStarted ? new Date(g.dateStarted).getFullYear() : null;
+					return goalYear !== null
+						&& goalYear < currentYear
+						&& g.status !== GOAL_STATUS.COMPLETED;
+				})
+				.map(async goal => {
+					const goalYear = new Date(goal.dateStarted).getFullYear();
+					const loansTowardGoal = await getCategoryLoanCountByYear(goal.category, goalYear);
```
dyersituations, line 982: Is this solving the same this as `backfillExpiredGoalProgress`?

christian14b: Yes, `backfillExpiredGoalProgress` uses `getCategoryLoanCountByYear` to populate the loans made values for expired goals, this line just adds the loans made in the `renewAnnualGoal` process (only for that moment)

dyersituations: Both paths aren't hit regularly at the same time? Just curious if these similar fixes could be consolidated.

christian14b: Good call, yes, they do overlap. I'm tend to leave them split in this PR because consolidating now would also pull in the renewal-toast logic (`hadCompletedGoal`, `goalsRenewedDate`, the `setMyKivaGoal`). I can write a follow-up work if that makes sense, probably not too complex but worth having dedicated work

dyersituations (approved):

PR #7012, opened 2026-06-22
feat: mp-2912-hidden goal progress

infante-jaime:
## Summary

Adds regression coverage for the TY goal module behavior.

The existing display conditions were already solid: goal progress only shows when the current checkout includes a loan that advanced the user’s active goal. The TY hide cookie only hides the goal creation ask, not valid progress or completion messaging.

## Use Cases

| Scenario | Expected |
|---|---|
| No goal + cookie not capped | Show goal creation |
| No goal + cookie capped | Hide goal creation |
| Active goal + checkout loan advanced goal | Show goal progress |
| Active goal + checkout loan did not advance goal | Do not show goal progress |
| Goal completed by checkout | Show goal completion |

This protects the reported flow: after setting a goal on TY and funding a recommended loan, the next confirmation page should show progress if that loan advanced the goal, even when the TY goal ask cookie is capped.

### DIFF
--- a/src/components/Thanks/ThanksPageSingleVersion.vue
+++ b/src/components/Thanks/ThanksPageSingleVersion.vue
@@ -291,7 +291,7 @@ const showGuestAccountModal = ref(false);
 const showReceipt = ref(false);
 const router = useRouter();
 const showGoalModal = ref(false);
-const showGoalInProgressModule = ref(false);
+const hasContributingLoans = ref(false);
 const isGoalSet = ref(false);
 const isEmptyGoal = ref(true);
 const goalTarget = ref(0);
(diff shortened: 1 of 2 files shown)


dyersituations:
@infante-jaime this PR just adds tests, is there any code safety we could move into the app itself? This won't help potentially address the issue that Lauren encountered.

infante-jaime:
I agree tests won't fix it, though they do help avoid future mismatches.

Just to be safe, I also made a small change so the condition can react if some of the values take a little longer to load.

This was pretty hard to repro because the logic was already right. The issue was that showGoalInProgressModule was only checked once during the initial load. If the goal data finished loading later, the page could end up showing the wrong state.

The logic itself is unchanged; it now stays in sync with the current state instead of potentially getting stuck with an outdated value.

dyersituations (approved):

PR #7014, opened 2026-06-23
feat: mp-2913-build-impact-contentful-issue

infante-jaime:
##Issue

The MyKiva next steps page could show the Build impact beyond your loan section title even when there was no content to display. This could happen when Contentful returned no eligible content or when the cache read failed.

##Fix

Added a showBuildImpactSection check so the section only appears when there is at least one card or Contentful slide to show.

Also split the Contentful cache reads into separate helpers for slides and badge data. If a cache read fails, the error is logged and an empty array is returned so the page can continue loading.

##Tests

Added coverage for:

Hiding the section when there is no content
Showing the section when eligible content exists
Handling missing or failed Contentful cache reads gracefully

### DIFF
--- a/src/pages/MyKiva/MyKivaPage.vue
+++ b/src/pages/MyKiva/MyKivaPage.vue
@@ -240,6 +240,34 @@ export default {
 				return [];
 			}
 		},
+		readContentfulSlides() {
+			try {
+				const slidesResult = this.apollo.readQuery({
+					query: contentfulEntriesQuery,
+					variables: {
+						contentType: 'carousel',
+						contentKey: CONTENTFUL_CAROUSEL_KEY,
+					}
+				});
+				return slidesResult.contentful?.entries?.items?.[0]?.fields?.slides ?? [];
+			} catch (e) {
+				logReadQueryError(e, 'MyKivaPage readContentfulSlides');
+				return [];
+			}
+		},
+		readContentfulBadgeData() {
+			try {
+				const contentfulChallengeResult = this.apollo.readQuery({
+					query: contentfulEntriesQuery,
+					variables: { contentType: 'challenge', limit: 200 }
+				});
+				return (contentfulChallengeResult.contentful?.entries?.items ?? [])
+					.map(entry => getContentfulLevelData(entry));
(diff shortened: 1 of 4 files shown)


src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -351,20 +379,8 @@ export default {
...
-				query: contentfulEntriesQuery,
-				variables: {
-					contentType: 'carousel',
-					contentKey: CONTENTFUL_CAROUSEL_KEY,
-				}
-			});
-			this.heroSlides = slidesResult.contentful?.entries?.items?.[0]?.fields?.slides ?? [];
-			this.heroBadgeContentfulData = (contentfulChallengeResult.contentful?.entries?.items ?? [])
-				.map(entry => getContentfulLevelData(entry));
+			this.heroSlides = this.readContentfulSlides();
```
dyersituations, line 382: What is this change doing? It looks like it's just moving code to new methods?

dyersituations:
@infante-jaime looks like we now just hiding the section if no Contentful data is loaded, which is a good fallback, but did you identify any potential reasons for why the Contentful read failed sometimes on dev (or the data returned was handled wrong)?

src/pages/MyKiva/MyKivaPage.vue
infante-jaime: This extracts the same Contentful reads into sepatarted methods so each one can fail independently, log the specific error, and fall back to [].
That fallback lets the section logic treat Contentful failures the same as no eligible content, preventing an empty “Build impact beyond your loan” title.

infante-jaime:
> @infante-jaime looks like we now just hiding the section if no Contentful data is loaded, which is a good fallback, but did you identify any potential reasons for why the Contentful read failed sometimes on dev (or the data returned was handled wrong)?

Yes, I looked into it and didn’t find an obvious systemic failure. With the new scoped logging, we can now tell when the Contentful read actually fails versus when it succeeds but returns no eligible content, which should give us enough signal to dig deeper if this keeps happening.

dyersituations (approved):

PR #7019, opened 2026-06-23
fix: avoid progress-ring or form flash when closing goal modal

christian14b:
https://kiva.atlassian.net/browse/MP-2898

https://github.com/user-attachments/assets/99109cf5-7077-40a1-b800-c0f9c15501f0

### DIFF
--- a/src/components/MyKiva/GoalSettingModal.vue
+++ b/src/components/MyKiva/GoalSettingModal.vue
@@ -646,10 +646,6 @@ const resetForm = () => {
 
 const closeLightbox = () => {
 	emit('close-goal-modal');
-	// Avoid showing category choice step when closing the modal
-	setTimeout(() => {
-		resetForm();
-	}, 300);
 };
 
 const editGoalCategory = () => {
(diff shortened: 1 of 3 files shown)


src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -689,6 +685,7 @@ watch(show, async newVal => {
 	if (!newVal) {
 		return;
 	}
+	resetForm();
```
roger-in-kiva, line 688: We already have a test file for this component, should we add this change there?

dyersituations (approved):
Looks good pending test question.

src/components/MyKiva/GoalSettingModal.vue
christian14b: I thought about it too, but the issue was not related to the modal, recommended loan was hiding too quickly that's why the updated test ensures that the loan remains visible. 

I validated this change with control and seems that the timeout logic was no longer needed

PR #7023, opened 2026-06-24
feat: adding guard to create goal to avoid duplicate creation

christian14b:
https://kiva.atlassian.net/browse/MP-2929

Adding a guard to avoid any possibilities to create a goal twice due to fast clicks.

### DIFF
--- a/src/components/MyKiva/GoalSettingModal.vue
+++ b/src/components/MyKiva/GoalSettingModal.vue
@@ -277,7 +277,7 @@
 				>
 					Back
 				</KvButton>
-				<KvButton @click="handleClick">
+				<KvButton :disabled="isSubmitting" @click="handleClick">
 					{{ ctaCopy }}
 				</KvButton>
 			</div>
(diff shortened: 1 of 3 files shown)


cristhianDt (approved):

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -611,6 +612,11 @@ const handleClick = () => {
 		formStep.value = 1;
 		emit('update-goal-choices', selectedCategory.value);
 	} else {
+		// Guard against the same click emitting twice
+		// since each emit triggers a goal write.
+		if (isSubmitting.value) return;
```
dyersituations, line 617: Do we need the same fix for other entry points like goal-setting page?

christian14b: Good catch, probably in GoalContainer. I'll take a look

dyersituations (approved):

PR #7025, opened 2026-06-24
fix: more loans beta parity

dyersituations:
https://kiva.atlassian.net/browse/MP-2940

Fixed some issues identified while comparing legacy vs beta further, both manually and automated.

### DIFF
--- a/src/components/Portfolio/LoanFilterBar.vue
+++ b/src/components/Portfolio/LoanFilterBar.vue
@@ -84,6 +84,7 @@
 			<div :class="[filtersExpanded ? 'tw-mt-2' : 'tw-mt-1.5', 'lg:tw-mt-2 tw-flex tw-items-center tw-gap-2']">
 				<kv-button
 					variant="primary"
+					:state="totalLoans ? '' : 'disabled'"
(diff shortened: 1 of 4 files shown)


src/components/Portfolio/LoanFilterBar.vue
```diff
@@ -253,6 +253,11 @@ const loanCountLabel = computed(() => {
 	return `${count.toLocaleString('en-US')} ${count === 1 ? 'loan' : 'loans'}`;
 });
 
+const exportLabel = computed(() => {
+	const count = props.totalLoans || 0;
```
roger-in-kiva, line 258: As 0 is fallback, do we want to show that case as a button copy?

mcstover (approved):

src/components/Portfolio/LoanFilterBar.vue
cristhianDt: (thoughts) or maybe disabling the button if it's 0

dyersituations: Good call, I went with disabling the button to reduce pop-in/out.

PR #7026, opened 2026-06-25
feat: mp-2945-full-remove-green-props

infante-jaime:
removeing the green props experiment 

https://kiva.atlassian.net/browse/MP-2945

### DIFF
--- a/src/components/MyKiva/GoalSetting/GoalSelector.vue
+++ b/src/components/MyKiva/GoalSetting/GoalSelector.vue
@@ -18,14 +18,12 @@
 			:go-to-url="goToUrl"
 			:is-updating-goal="isUpdatingGoal"
 			:is-goal-completed="isGoalCompleted"
-			:is-goal-tile-experiment-enabled="isGoalTileExperimentEnabled"
 			@edit-goal-from-settings="handleEditGoalFromSettings"
 			@button-click="handleSuccessContinue"
 		/>
 		<!-- Goal Selection Form (shown before goal is set) -->
 		<template v-else>
 			<img
-				v-if="!isGoalTileExperimentEnabled"
 				:src="HandsPlant"
 				class="goal-selector__image lg:tw-mb-1 tw-w-10 lg:tw-w-12.5"
 			>
@@ -36,11 +34,6 @@
 			<template
 				v-else
 			>
-				<img
-					v-if="isGoalTileExperimentEnabled && !isGoalSet && !isLarge"
-					:src="HandsPlant"
-					class="goal-selector__image lg:tw-mb-1 tw-w-10 lg:tw-w-12.5 tw-mx-auto"
-				>
 				<h2
 					class="goal-selector__title tw-text-headline tw-px-4 lg:tw-px-7 tw-text-center"
 					v-html="titleText"
@@ -62,15 +55,7 @@
 			</p>
 
 			<p
-				v-if="showGoalValuePropsCopy"
-				v-html="subtitleText"
-				class="goal-selector__prompt tw-text-base lg:tw-text-subhead tw-my-1.5 lg:tw-mb-1 lg:tw-mt-2
-					tw-text-center"
-			>
-			</p>
-
-			<p
-				v-if="subtitleText && !showGoalValuePropsCopy && progressSubtitleBeforeOptions"
+				v-if="subtitleText && progressSubtitleBeforeOptions"
 				v-html="subtitleText"
 				class="goal-selector__prompt tw-text-base lg:tw-text-subhead tw-my-1.5 lg:tw-mb-1 lg:tw-mt-2
 					tw-text-center"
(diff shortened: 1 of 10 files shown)


src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -387,20 +317,6 @@ const props = defineProps({
...
-	 * Flag to indicate if the goal tile experiment is enabled
-	 */
-	isGoalTileExperimentEnabled: {
-		type: Boolean,
-		default: false,
-	},
-	/**
-	 * Flag to indicate if the goal value props copy version should be shown
-	 */
-	showGoalValuePropsCopy: {
```
dyersituations, line 400: We want to remove the green section but keep the copy changes, so do we want to remove the copy that was being enabled by `showGoalValuePropsCopy`?

infante-jaime: Good callout. showGoalValuePropsCopy was doing two things: 

- controlled the green value-props section 
- some of the selector copy.

Since we're removing the green section, I removed the prop and the value-props UI, So that we have a single version of the copy layout

dyersituations: Changes to the copy outside the green section was kept? It looked like it was removed in a couple places.

infante-jaime: The prop removal is tricky yet the intended idea was to keep everything in place and as far as it goes looks ok, to do that removal that way simplifies code for future aditions, yet if you preffere I can do a full regression test to see that everything matches exactly the designs before merging just to be sure.

dyersituations: What I meant was that we do want to remove the green section, but I noticed for example `GoalSelector.vue` that the content outside of the green section is also being changed. Are we keeping the experiment copy versions in those sections?

infante-jaime: Yes, we should keep the experiment copy outside the green section.

The confusing part is that showGoalValuePropsCopy was also controlling some non-green selector copy, so removing it made some of those changes look like copy removals. The intent is to keep those experiment copy versions, just without tying them to the removed showGoalValuePropsCopy prop.

So the green value-props section goes away, but the retained selector copy becomes part of the normal flow.

dyersituations: Nice, thanks for confirming!

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -524,7 +436,7 @@ const isCustomIndex = computed(() => selectedIdx.value === CUSTOM_LOAN_NUMBER_IN
 
 const isContinueDisabled = computed(() => (
 	props.isLoadingData
-	|| props.loadingCurrentYear
+	|| loadingCurrentYear.value
```
dyersituations, line 439: Why was this changed from being the prop?

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -309,8 +159,6 @@ import GoalSelector from '#src/components/MyKiva/GoalSetting/GoalSelector';
 import RecommendLoanForGoalContent from '#src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalContent';
 import RecommendLoanForGoalFooter from '#src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalFooter';
 import RecommendLoanForGoalHeader from '#src/components/LoanCards/RecommendLoanForGoal/RecommendLoanForGoalHeader';
-import HandsPlant from '#src/assets/images/thanks-page/hands-plant-v3.png';
```
dyersituations, line 312: Is this still used elsewhere or can it be removed?

src/components/MyKiva/GoalSettingModal.vue
```diff
@@ -392,20 +240,6 @@ const props = defineProps({
 		type: Boolean,
 		default: false,
 	},
-	/**
-	 * Flag to indicate if the goal tile experiment is enabled
-	 */
-	isGoalTileExperimentEnabled: {
```
dyersituations, line 398: Looks like we still need to remove the experiment assignment for this?

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: Good catch. This wasn’t part of the green-section removal.

I changed it because loadingCurrentYear is a local ref in GoalSelector, and it isn’t declared as a prop there, so props.loadingCurrentYear was always undefined. The template already uses the local loadingCurrentYear, so isContinueDisabled was the inconsistent spot.

So this was a non related issue yet an issue I can remove it from here and flag it as a separate issue or keep it in this PR

dyersituations: What does it fix by including it?

src/components/MyKiva/GoalSettingModal.vue
infante-jaime: yes is  still used  in a couple of places

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: It fixes the continue button so it stays disabled while GoalSelector is fetching current-year loan progress.

src/components/MyKiva/GoalSettingModal.vue
dyersituations: @infante-jaime see this remaining comment.

dyersituations: @infante-jaime see comment here, can we remove the experiment assignment code?

christian14b: +1, I think we can remove the assignment code in `MyKivaPage.vue`

src/components/MyKiva/GoalSetting/GoalSelector.vue
```diff
@@ -36,11 +34,6 @@
 			<template
 				v-else
 			>
-				<img
-					v-if="isGoalTileExperimentEnabled && !isGoalSet && !isLarge"
-					:src="HandsPlant"
```
roger-in-kiva, line 41: Looks like this image could be removed.

src/components/MyKiva/GoalSettingModal.vue
infante-jaime: Removed the stale GoalSettingModal usage. The remaining mykiva_goal_tile assignment is intentional because it still drives the carousel goal-card copy via JourneyCardCarousel -> NextYearGoalCard. This cleanup removes the retired modal/value-props behavior without fully sunsetting the experiment.

I also did a full mobile regression pass and confirmed the carousel/modal flows still look correct.

dyersituations (approved):
Approved pending merge conflict resolution.

src/components/MyKiva/GoalSetting/GoalSelector.vue
infante-jaime: Yep, the duplicate experiment-specific image path was removed. I kept the base HandsPlant image because it’s still used by the normal pre-goal selector, and the same asset is also used in NextYearGoalCard, so the image asset itself should stay.

roger-in-kiva: True, I was confused as we have V2 and V3 for this asset.
Thanks checking!

roger-in-kiva (approved):

PR #7029, opened 2026-06-25
fix: don't show atb modal on mykiva when loan added to basket after goal set

dyersituations:
https://kiva.atlassian.net/browse/MP-2946

Final tweak for the "lend after goal set" feature.


christian14b (approved):
👍

cristhianDt (approved):

roger-in-kiva (approved):

PR #7033, opened 2026-06-29
feat: update borrower profile typography heading to new ecosystem

christian14b:
https://kiva.atlassian.net/browse/MP-2926

image
image
image
image


mcstover (approved):
These are all fine with regard to rolling out the new baseline type styles via html element. What I don't see here is any attention to the spacing that would make the new scale relate better to it's content. That will be something to visually work on with design to improve the page layout overall.

dyersituations (approved):

PR #7036, opened 2026-07-01
fix(graphql): add id field for Video queries

emuvente:
An id field has been added to the Video type in development, and will soon be released to production. This PR will be needed to unblock linting once that's in production.


eddieferrer (approved):

dyersituations (approved):

mcstover (approved):
guess we need to get the api in prod updated too

PR #7041, opened 2026-07-07
fix: resolve graphql name collision on id

dyersituations:
BP page currently not loading on dev.

### DIFF
--- a/src/components/BorrowerProfile/LoanStory.vue
+++ b/src/components/BorrowerProfile/LoanStory.vue
@@ -77,7 +77,7 @@ export default {
 							hash
 						}
 						... on Video {
-							id
+							videoId: id
 							youtubeId
 						}
 					}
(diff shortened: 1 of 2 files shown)


eddieferrer (approved):

roger-in-kiva (approved):

src/components/BorrowerProfile/LoanStory.vue
```diff
@@ -77,7 +77,7 @@ export default {
 							hash
 						}
 						... on Video {
-							id
+							videoId: id
```
emuvente, line 80: we should probably move the id field instead

emuvente (commented):
Ok, I understand the issue now. We should fix this on the api, not here.

dyersituations:
> Ok, I understand the issue now. We should fix this on the api, not here.

Yeah it can be fixed at API or type policy, but API is "true" fix. I'll put something together.

PR #7043, opened 2026-07-07
feat: implement migrated withdraw-beta pages

dyersituations:
https://kiva.atlassian.net/browse/MP-2960
https://kiva.atlassian.net/browse/MP-2961
https://kiva.atlassian.net/browse/MP-2962

Implemented the FE portion of the `/withdraw-beta` page migration. I was able to test all 3 new pages using our sandbox PayPal email and manually adding then transferring credit in dev Kiva Admin.

image

image

image


src/util/withdraw/withdrawConstants.js
```diff
@@ -0,0 +1,32 @@
+// Mirrors the WithdrawalSubmitStatus GraphQL enum returned by the
```
cristhianDt, line 1: (thoughts) I'm starting to see that maybe we need a generic folder to store these constants, something like `/src/common/constants/**` or `/src/config/constants/**` idk

src/pages/Withdraw/WithdrawConfirmPage.vue
```diff
@@ -0,0 +1,336 @@
...
+	},
+	methods: {
+		onDonationInput(value) {
+			this.donationAmount = value;
+			this.donationTouched = true;
+		},
+		onToggleDonation(value) {
+			// Reset the touched flag so re-opening the field starts without an error.
+			this.donationTouched = false;
+			this.$kvTrackEvent('portfolio', 'click', 'Withdraw donation toggle', value ? 1 : 0);
```
cristhianDt, line 261: (doubt) I just see it this event, are there any other events for the other pages that are also pending migration? or just this one already exist in the old versions

cristhianDt:
@dyersituations even if my user hasn't credit, I see the form in development, but in the new page I get the message "You do not have enough Kiva credit to request a withdrawal.", is that correct?

Good question, I do think the new version makes more sense UX-wise.

dyersituations:
@cristhianDt oops edited your post... I think the new version makes more sense UX-wise, but I could be convinced otherwise. Thoughts?

cristhianDt:
> @cristhianDt oops edited your post... I think the new version makes more sense UX-wise, but I could be convinced otherwise. Thoughts?

well, I think the change is fine, It makes more sense than the initial version, thanks for confirm!

src/util/withdraw/withdrawConstants.js
dyersituations: Good point, we typically put constants in the util folder at least, and I missed that on the first pass. I moved this new file to a new folder in utils. The one potential problem with a shared folder is there is less context if the files don't exist in a named/specific area, and our system doesn't consume constants globally like something like a state management library would.

src/pages/Withdraw/WithdrawConfirmPage.vue
dyersituations: The legacy version of this page had no analytics other than pageview. But I added in click/interaction events with the migration. Good call out though, and I did another analysis of the events and found a few that were missing, so thanks for the comment!

christian14b (approved):
Nice 🚀

mcstover (approved):

roger-in-kiva (approved):

src/util/withdraw/withdrawConstants.js
cristhianDt: thanks for sharing this!

cristhianDt (commented):
👍

PR #7049, opened 2026-07-08
feat: mp 2933 remove deposit toggle experiment

infante-jaime:
Updates the frontend to use the backend-owned donation Kiva Credit default behavior for the /donate/supportus in-context checkout flow.
Changes
Removed old EXP-MP-2827 variant tracking from InContextCheckout.
Added a supportus-modal-specific setup mutation that passes applyDonationCreditDefault: true.
Applies that mutation only for logged-in, donation-only /donate/supportus in-context checkout baskets.
Leaves normal /checkout and /donate/supportusprocess behavior unchanged.
Added unit coverage for the supportus modal opt-in and excluded paths.


mcstover:
The `/donate/supportus` route is in cms-page-server so the in-context checkout would only load from that server as well... If these changes are needed in ui they will be against other contexts...

`'/donate/supportusprocess'` is in the monolith and has this route: `$router->connect('/donate/supportusprocess', array('page'=>'basket', 'action'=>'supportUsSaveDonation'));`

infante-jaime:
you are right sorry for that closing now

PR #7050, opened 2026-07-09
feat: create helper to retrieve needed information for goal in review

christian14b:
https://kiva.atlassian.net/browse/MP-2948

Use `?goalSummaryTiming` to calculate query timing build in the monolith. This is a temporary code and will be removed after testing with superlenders.

This is not a final look of the helper since it's only focused on the first view of the goal in review


src/composables/useGoalData.js
```diff
@@ -375,6 +382,122 @@ export default function useGoalData({ apollo } = {}) {
...
+				return null;
+			}
+		}
+
+		const year = new Date(goal.dateStarted).getFullYear();
+		let achievement = null;
+		try {
+			const response = await apolloClient.query({
+				query: goalSummaryAchievementQuery,
+				variables: { year, loanPurchasesLimit: GOAL_SUMMARY_LOAN_PURCHASES_LIMIT },
```
dyersituations: For this particular feature, we don't need to know about specific loans, correct? So why do we need to lean on the loan purchase limit of 1000 in the achievement service? The achievement service also stores a count per year per achievement. The loan purchase rolling window just means we only have specific data about the last 1000 loans purchased towards that achievement.

src/composables/useGoalData.js
```diff
@@ -375,6 +382,122 @@ export default function useGoalData({ apollo } = {}) {
...
+	 * (super lenders have too many transactions to compute on the FE).
+	 *
+	 * For badge-journey goals, `count`, `percent`, and `borrowerCount` come from
+	 * achievements-service. `amount` stays null: LoanPurchase doesn't expose a
+	 * per-share money field, so we can't compute what the user actually lent
+	 * without an achievements-service extension.
+	 *
+	 * `borrowerCount` is summed from the federated `loanPurchases.loan.borrowerCount`.
+	 * achievements-service retains `loanPurchases` in a rolling window of
+	 * MAX_ROLLING_WINDOW_SIZE (1000) loans, so we request that many
```
dyersituations: Similar to the other comment, I won't think the rolling window is important to this feature. In general this comment should be simplified. We need to be careful with the larger AI-generated comments. Sometimes they are super useful, sometimes they overload the file.

src/composables/useGoalData.js
christian14b: Oh, you're right the count per year might be neat. I didn't think this through since my main goal was to test the support all timing first and the remove some code + comments. But thank you for the feedback, I'll apply your suggestion

src/composables/useGoalData.js
christian14b: Agreed, thought I trimmed most of the comments correctly, but this is not good enough. This will be gone with your above suggestion, but I'll take a closer look at the comments again

dyersituations (approved):

roger-in-kiva (approved):

PR #7051, opened 2026-07-10
feat: roll out AI loan pills across My Kiva, Checkout, and borrower profile

roger-in-kiva:
## Summary

Rolls out **AI loan pills** across My Kiva, Checkout, and the borrower profile side sheet by removing the experiment gate that kept the feature dark in the `ui` repo. Pills now render whenever the backend returns them for a loan, falling back to the default loan callouts otherwise.

Previously the feature was hard-disabled here via `ai-loan-pills-mixin` (which forced `enableAILoanPills = false`) while the experiment ran only on the combo page in `cms-page-server`. This PR removes that gate and the now-unused plumbing.

## What changed

**Removed the experiment gate**
- Deleted `src/plugins/ai-loan-pills-mixin.js` (and its spec). It hardcoded the feature off and owned the experiment enrollment/tracking. Enrollment continues on the combo page in `cms-page-server`, so no analytics is lost by removing it here.
- Dropped the `enableAiLoanPills` / `enableAILoanPills` prop threaded through the component chain:
  - `CheckoutPage → EmptyBasketCarousel → KvClassicLoanCardContainer`
  - `MyKivaPage → MyKivaPageContent → LendingCategorySection` and `BorrowerSideSheetWrapper → BorrowerSideSheetContent → SideSheetLoanTags`
- Pills render unconditionally when present; the default loan callouts remain the fallback.

**New shared helper**
- Added `withAiPills(apollo, loans)` in `src/util/aiLoanPIillsUtils.js`, consolidating `getLoansIds` + `fetchAiLoanPills` + `addAiPillsToLoans`. It skips the network request when there are no loans and degrades gracefully to empty pills when the query fails.

**Non-blocking load ordering**
- `EmptyBasketCarousel` and `MyKivaPageContent` now show loans first and enrich with AI pills afterward, so the pills query no longer blocks the grid/carousel.

**Bug fixes (pre-existing, fixed alongside the rollout)**
- `BorrowerSideSheetContent`: `aiPills` is now a `ref`, so the async pills result actually triggers a re-render (the previous non-reactive array `.push()` did not).
- Guarded the first result (`result?.[0]?.pills`) to avoid a crash on an empty response.

## Related links

- Jira: [MP-2964 — AI loan pills rollout](https://kiva.atlassian.net/browse/MP-2964)
- Original experiment: [MP-2914 — AI loan pills](https://kiva.atlassian.net/browse/MP-2914)

## Related tests

- [`test/unit/specs/util/aiLoanPIillsUtils.spec.js`](https://github.com/kiva/ui/blob/MP-2964_ai_pills_rollout/test/unit/specs/util/aiLoanPIillsUtils.spec.js) — new `withAiPills` suite: happy path, empty/null loans (skips the query), and query-failure (returns loans with empty pills).
- [`test/unit/specs/components/BorrowerProfile/SideSheetLoanTags.spec.js`](https://github.com/kiva/ui/blob/MP-2964_ai_pills_rollout/test/unit/specs/components/BorrowerProfile/SideSheetLoanTags.spec.js) — **new**: renders AI pills when provided; falls back to default callouts when none.
- [`test/unit/specs/components/LoanCards/KvClassicLoanCardContainer.spec.js`](https://github.com/kiva/ui/blob/MP-2964_ai_pills_rollout/test/unit/specs/components/LoanCards/KvClassicLoanCardContainer.spec.js) — **new**: AI pills surface as custom callouts when present; none render otherwise.
- Removed `test/unit/specs/plugins/ai-loan-pills-mixin.spec.js` and the `ai-loan-pills-mixin` mock from the two Checkout specs (`CheckoutPage.spec.js`, `CheckoutPageUpsell.spec.js`).

**Test run:** `npm run unit` → 267 files, **4078 passed / 6 skipped / 0 failing**. `npm run lint` → **0 errors** (pre-existing GraphQL warnings only, in files untouched here).

## QA / how to test

1. **My Kiva** — recommended loans grid shows AI pills when the backend returns them; loans render immediately and pills fill in without blocking.
2. **Checkout (empty basket)** — the suggested-loans carousel renders, then pills appear.
3. **Borrower profile side sheet** (My Kiva and Portfolio Impact Dashboard) — open a loan; pills show on the tags row, with default callouts as the fallback.
4. Confirm no console errors and that loans still render when the pills query returns nothing or errors.


src/util/aiLoanPIillsUtils.js
```diff
@@ -47,3 +47,20 @@ export const addAiPillsToLoans = (loans, pillsLoans) => {
...
+/**
+ * Fetches AI pills for the provided loans and returns them with the pills attached.
+ * Skips the network request when there are no loans to enrich.
+ *
+ * @param apollo The current Apollo client
+ * @param {Array} loans The loans to enrich with AI pills
+ * @returns The loans with their AI pills attached
+ */
+export const withAiPills = async (apollo, loans) => {
+	const loanIds = getLoansIds(loans ?? []);
```
christian14b: Minor  but seems like `loans ?? []` appears twice and `addAiPillsToLoans` also does its own `loans ?? []` internally. You could normalize once at the top something like:

```
const safeLoans = loans ?? [];
const loanIds = getLoansIds(safeLoans);
if (!loanIds.length) return safeLoans;
const pillsLoans = await fetchAiLoanPills(apollo, loanIds);
return addAiPillsToLoans(safeLoans, pillsLoans ?? []);
```

src/components/Checkout/EmptyBasketCarousel.vue
```diff
@@ -13,25 +13,24 @@
 		>
 			<div class="section-container tw-mx-auto tw-my-0">
 				<kv-carousel
-					v-if="randomLoans.length > 0 && !loading"
+					v-if="suggestedLoans.length > 0 && !loading"
 					class="tw-w-full tw-overflow-visible md:tw-overflow-hidden"
 					data-testid="random-loan-card-carousel"
```
christian14b, line 18: Should this also change to `suggested-loan` to keep the naming change convention?

src/util/aiLoanPIillsUtils.js
christian14b: Isn't this always `pills` since pills is already guaranteed above to be an array (?? [])

christian14b (commented):
Looks good! just a couple of comments/questions about some points.

src/components/Checkout/EmptyBasketCarousel.vue
roger-in-kiva: Leave it if we already have some test based on that, but I'll revert name so, thanks!

christian14b: Yeah, it makes sense to leave it as `random-loan` for that, I made a quick search in https://github.com/kiva/kiva-tests and seems this is not used so a change won't break anything as far as I know

src/components/BorrowerProfile/BorrowerSideSheetContent.vue
```diff
@@ -151,16 +147,13 @@ export default {
...
-		if (props.enableAiLoanPills) {
-			fetchAiLoanPills(apollo, [props.loanId]).then(result => {
-				const pills = result?.[0].pills ?? [];
-				const labeledPills = pills.map(pill => ({
-					label: pill,
-				}));
-				aiPills.push(...labeledPills);
-			});
-		}
+		const aiPills = ref([]);
```
dyersituations, line 150: Why switch to `ref` here?

roger-in-kiva: As mentioned in description:

BorrowerSideSheetContent: aiPills is now a ref, so the async pills result actually triggers a re-render (the previous non-reactive array .push() did not).
Guarded the first result (result?.[0]?.pills) to avoid a crash on an empty response.

dyersituations (approved):

christian14b (approved):

PR #7052, opened 2026-07-10
feat: show AI loan pill in borrower profile summary card (MP-2994)

roger-in-kiva:
## Summary

Renders an AI loan "pill" on the borrower profile page (`/lend/:id`) for the
single loan shown, completing the AI Loan Pill Emphasis rollout that already
shipped on My Kiva, Checkout, and the borrower profile side sheet.

When the backend returns pills for the loan, the summary card's tag row shows
**one** AI pill in place of the activity tag. The country tag always stays.
When there are no pills (or the query errors), the row falls back to the
existing activity tag — matching the side sheet's fallback behavior.

## What changed

- **`src/components/BorrowerProfile/SummaryCard.vue`**
  - Fetches pills for the single loan via `fetchAiLoanPills(apollo, [loanId])`
    from `#src/util/aiLoanPillsUtils`, mirroring the side sheet reference.
  - New `aiPill` string; `fetchAiPills()` sets it to `result?.[0]?.pills?.[0] ?? ''`.
  - Pills are fetched **client-side, in parallel** with the summary mount query
    (`Promise.all`), so there is no added serial latency and SSR/CDN render is
    unaffected.
  - The card's loader is held until both the mount query and pills resolve, so
    the tag renders once instead of the activity tag popping to a pill.
  - Tag row: `country` (always) → AI pill if present, else activity tag.
- **`test/unit/specs/components/BorrowerProfile/SummaryCard.spec.js`** — new
  `SummaryCard AI pills` suite (details below).

## Behavior

| Backend response | Tag row (slot 2) |
| --- | --- |
| Pills present | First AI pill |
| Entry present, empty `pills` | Activity tag (fallback) |
| Empty `values` | Activity tag (fallback) |
| Query error | Activity tag (fallback), no card stall |

## Related links

- Jira: [MP-2994 – Implement AI pills in borrower profile page](https://kiva.atlassian.net/browse/MP-2994)
- Epic: [MP-2048 – AI Loan Pill Emphasis](https://kiva.atlassian.net/browse/MP-2048)

## Related tests

`test/unit/specs/components/BorrowerProfile/SummaryCard.spec.js` — run with `npm run unit -- SummaryCard` (8/8 passing):

- renders a single AI pill in place of the activity tag when pills are returned
  (asserts only the first of several pills renders, country tag stays, and the
  pills query fires with `{ loanIds: [12345] }`)
- falls back to the activity tag when no pills are returned
- falls back to the activity tag when the loan entry has an empty pills list
- falls back to the activity tag when the pills query rejects
- keeps the loader until the pills fetch resolves (no activity-then-pill pop)

## Notes / trade-off

Because the change reuses the single `isLoading` flag, the summary card's loader
now also waits on the pills fetch (not just the tag row). Pills run concurrently
with the existing mount query, so there's no added serial latency, and a pills
**error** is swallowed by the util and falls back to the activity tag without
stalling the card.

### DIFF
--- a/src/components/BorrowerProfile/SummaryCard.vue
+++ b/src/components/BorrowerProfile/SummaryCard.vue
@@ -121,7 +121,10 @@
 					</span>
 				</summary-tag>
 
-				<summary-tag data-testid="bp-summary-activity-tag" v-if="activityName">
+				<summary-tag v-if="aiPill" data-testid="bp-summary-ai-pill">
+					{{ aiPill }}
+				</summary-tag>
+				<summary-tag v-else-if="activityName" data-testid="bp-summary-activity-tag">
 					{{ activityName }}
 				</summary-tag>
 			</template>
@@ -147,6 +150,7 @@ import {
 	KvIconButton, KvMaterialIcon, KvLoadingPlaceholder, KvTextLink,
 } from '@kiva/kv-components';
 import useBorrowerProfileDefinitions from '#src/composables/useBorrowerProfileDefinitions';
+import { fetchAiLoanPills } from '#src/util/aiLoanPillsUtils';
 import BorrowerImage from './BorrowerImage';
 import BorrowerName from './BorrowerName';
 import ContentLightbox from './ContentLightbox';
@@ -266,6 +270,7 @@ export default {
 			pfpMinLenders: 0,
 			numLenders: 0,
 			totalComments: 0,
+			aiPill: '',
 		};
 	},
 	computed: {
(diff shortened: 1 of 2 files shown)


dyersituations:
@roger-in-kiva the changes look good, and I tested locally, but I think we should confirm the specifics of the change with Lauren. Here on the BP we didn't have a list of possible pills like we do on loan cards, just activity, so changing to the AI loan pills is a larger jump.

roger-in-kiva:
> @roger-in-kiva the changes look good, and I tested locally, but I think we should confirm the specifics of the change with Lauren. Here on the BP we didn't have a list of possible pills like we do on loan cards, just activity, so changing to the AI loan pills is a larger jump.

Just followed conversation in ticket comments, asking her now, thanks!

roger-in-kiva:
Closing after product feedback.

PR #7060, opened 2026-07-14
feat: implement portfolio credit deposit migration

dyersituations:
https://kiva.atlassian.net/browse/MP-2985

Implemented the migrated `/portfolio/credit/deposit-beta` using the new backend query and mutation. All functionality migrated and matching newstack.

Also a minor experiment cleanup that I noticed while looking through tech debt. Unrelated but super minor so I left it in the branch.

<br />

image

<br />

image

<br />

image

### DIFF
--- a/src/components/Deposit/DepositDropInPaymentWrapper.vue
+++ b/src/components/Deposit/DepositDropInPaymentWrapper.vue
@@ -1,0 +1,171 @@
+<template>
+	<div class="dropin-payment-holder tw-px-0 tw-mt-2">
+		<braintree-drop-in-interface
+			v-if="isClientReady"
+			ref="braintreeDropInInterface"
+			:amount="formattedAmount"
+			flow="checkout"
+			:payment-types="['paypal', 'card']"
+			@transactions-enabled="enableConfirmButton = $event"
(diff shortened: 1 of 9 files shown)


roger-in-kiva (approved):

christian14b (approved):

cristhianDt (approved):

src/components/Deposit/DepositDropInPaymentWrapper.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+	<div class="dropin-payment-holder tw-px-0 tw-mt-2">
+		<braintree-drop-in-interface
```
eddieferrer: Since this is a new page, we should use the KvPaymentSelect component from kv-ui-elements. ~~We should probably also allow flow="vault" to allow users to vault their cards~~

Actually I think flow checkout is fine

dyersituations: Good call, I forgot about the newer shared component. Updated with some adjustments to support the new component.

eddieferrer (approved):

PR #7067, opened 2026-07-17
feat: add goal in review flag and date range validation

christian14b:
https://kiva.atlassian.net/browse/MP-3053

Flag for goal in review

### DIFF
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -71,6 +71,15 @@ export const GOALS_CURRENT_YEAR = new Date().getFullYear();
 export const LAST_YEAR_KEY = GOALS_CURRENT_YEAR - 1;
 export const COMPLETED_GOAL_THRESHOLD = 100;
 export const HALF_GOAL_THRESHOLD = 50;
+
+// TODO(MP-3053): update these placeholder dates once the exact release window is confirmed.
+export const GOAL_IN_REVIEW_START_DATE = new Date('2026-11-15T00:00:00-08:00');
+export const GOAL_IN_REVIEW_END_DATE = new Date('2027-01-15T23:59:59-08:00');
+
+export function isWithinGoalInReviewWindow() {
+	const now = new Date();
+	return now >= GOAL_IN_REVIEW_START_DATE && now <= GOAL_IN_REVIEW_END_DATE;
+}
 const MIN_CATEGORY_LOANS_AMOUNT = 100;
 const RECOMMENDED_LOANS_LIMIT = 4;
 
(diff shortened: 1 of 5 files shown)


src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -177,6 +178,10 @@ export default {
 		});
 
 		this.showTeamChallenge = teamsChallengeEnable && this.allowedTeams.length > 0;
+
+		this.goalInReviewEnable = (readBoolSetting(portfolioQueryData, 'general.goal_in_review_enable.value') ?? false)
```
dyersituations, line 182: This is appearing on portfolio page as well?

src/composables/useGoalData.js
```diff
@@ -71,6 +71,15 @@ export const GOALS_CURRENT_YEAR = new Date().getFullYear();
 export const LAST_YEAR_KEY = GOALS_CURRENT_YEAR - 1;
 export const COMPLETED_GOAL_THRESHOLD = 100;
 export const HALF_GOAL_THRESHOLD = 50;
+
+// TODO(MP-3053): update these placeholder dates once the exact release window is confirmed.
+export const GOAL_IN_REVIEW_START_DATE = new Date('2026-11-15T00:00:00-08:00');
+export const GOAL_IN_REVIEW_END_DATE = new Date('2027-01-15T23:59:59-08:00');
```
dyersituations: Are we turning if off? I thought it would remain on, or specifically remain on in the goal card on MyKiva. Is this flag for the automatically shown version?

src/composables/useGoalData.js
```diff
@@ -71,6 +71,15 @@ export const GOALS_CURRENT_YEAR = new Date().getFullYear();
...
+
+export function isWithinGoalInReviewWindow() {
```
dyersituations: We'll want a query string override to enable this.

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
christian14b: Yes, in the epic there's a mention for this:

Where users see it: 

Lenders who completed their goal - the next session after they complete their goal, show as a pop up when they visit MyKiva or the Portfolio

src/composables/useGoalData.js
christian14b: Good catch, yeah it only says starts at ... I'll remove and probably simplify this logic with your comment bellow

dyersituations: Does the epic mention anything about when we stop automatically showing the modal to users? It can be a follow-up ticket.

dyersituations (approved):

src/composables/useGoalData.js
christian14b: Yes, it's not explicit with the wording but we can assume the content should happen only after completed goal or mid-november (1 loan away): `After a user completes their goal, this will show/pop up in the next session after goal complete. `

And yes, this work is described in the ticket that gathers the modal logic

PR #7079, opened 2026-07-21
feat: implement fe for /portfolio/transactions-beta

dyersituations:
https://kiva.atlassian.net/browse/MP-2984

FE implemented for `/portfolio/transactions-beta` migration. Functional parity with legacy.

image

image

### DIFF
--- a/src/components/Portfolio/TransactionFilterBar.vue
+++ b/src/components/Portfolio/TransactionFilterBar.vue
@@ -1,0 +1,315 @@
+<template>
+	<div>
+		<div class="tw-flex tw-flex-col">
+			<kv-text-input
+				id="transaction-filter-text-input"
+				v-model="searchText"
+				:can-clear="true"
+				placeholder="Search by loan ID or payment transaction ID"
+				@keyup.enter.prevent="emitDraftKeywordSearchIfChanged"
+				@blur="emitDraftKeywordSearchIfChanged"
+			/>
+			<div class="tw-flex tw-flex-col tw-gap-2 tw-mt-2">
+				<!-- Category + Sort share one row, each taking half the width from md up. -->
+				<div class="tw-flex tw-flex-col md:tw-flex-row tw-gap-2">
+					<div class="tw-flex tw-items-center tw-gap-2 md:tw-flex-1 md:tw-min-w-0">
+						<span class="tw-whitespace-nowrap">Category:</span>
+						<kv-select
+							id="transaction-category-select"
+							class="transaction-filter-control"
+							v-model="selectedCategory"
+							@update:model-value="handleCategoryChange"
+						>
+							<option
+								v-for="option in categoryOptions"
+								:key="option.value"
+								:value="option.value"
+							>
+								{{ option.label }}
+							</option>
+						</kv-select>
+					</div>
+					<div class="tw-flex tw-items-center tw-gap-2 md:tw-flex-1 md:tw-min-w-0">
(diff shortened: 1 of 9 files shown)


src/components/Portfolio/TransactionFilterBar.vue
```diff
@@ -0,0 +1,315 @@
...
+								v-for="option in sortOptions"
+								:key="option.value"
+								:value="option.value"
+							>
+								{{ option.label }}
+							</option>
+						</kv-select>
+					</div>
+				</div>
+				<!-- Date range on its own row, From/To each taking half the width from md up. -->
```
mcstover, line 50: Would our global KvDatepicker component be useful here?

dyersituations: We could use it, the KvTextInput looks OK. Any preference?

src/components/Portfolio/TransactionFilterBar.vue
```diff
@@ -0,0 +1,315 @@
...
+			const count = this.totalTransactions || 0;
+			return `${count.toLocaleString('en-US')} ${count === 1 ? 'transaction' : 'transactions'}`;
+		},
+		// The export respects the applied filters (see exportHref), so the count matches the
+		// filtered total shown on the page — not every transaction on the account.
+		exportLabel() {
+			const count = this.totalTransactions || 0;
+			return `Export ${count.toLocaleString('en-US')} ${count === 1 ? 'transaction' : 'transactions'}`;
+		},
+		// Mirror the legacy transactions export: a GET to /portfolio/transactions with
```
mcstover, line 212: Looks like this is a POST request to the same url so we will probably want to create a new route (maybe `/portfolio/transactions/export`) and point it to this same method so we can point `/portfolio/transactions` to ui...

dyersituations: Yeah we'll need a similar route to the one I created for exporting on another page. I'll get that in flux for dev once this is deployed.

src/components/Portfolio/TransactionFilterBar.vue
dyersituations: Tried changing to `KvDatePicker`. The component does look cooler in general. Changes pushed up.

image

src/components/Portfolio/TransactionFilterBar.vue
dyersituations: https://github.com/kiva/flux/pull/2979

mcstover: Ok yea that works. Maybe in the future we could get rid of the odd exclusion definitions and introduce the `/export` path for those contexts that need to hit the monolith.

dyersituations: Yeah there are a couple paths we could introduce on the monolith during code deletion work.

src/components/Portfolio/TransactionFilterBar.vue
mcstover: Cool, just thought it was work exploring. It also supports a range option so you wouldn't necessarily need 2 fields.

mcstover (approved):

src/components/Portfolio/TransactionFilterBar.vue
dyersituations: I definitely considered the range option but decided to keep parity for now.

PR #7080, opened 2026-07-21
feat: [CP-2930] gate borrower profile comment section by anonymizatio…

kathrynlovett:
…n and add-permission

Hide the LoanComments section entirely on full/pii anonymized loans for non-admin users, and only show the add-comment form when the loan is fundraising, or the user is an admin/trustee/trustee-to-the-loan, or is a privileged user to the loan who has lent to >=1 loan (or is the borrower of this loan).

Claude-Session: https://claude.ai/code/session_014sgTcV3cp97cKkp5gp5UA2

During PII anonymization validation, we realized the comment creation box was still displaying for PII-anonymized loans. Reviewing the logic in the monolith, it looks like the entire comment section wasn't visible if [$borrower_privacy->showSocialInfo](https://github.com/kiva/kiva/blob/dbcff827bf51e4d8d5874838bf8a5da95da3b15a/sites/www_kiva/server/views/Xb/Loan/LoanView.php#L198C8-L200) returns `false`. That method returns false when the ACL is `user` and the loan has an `anonymization_level` of `pii` or `full` OR if the ACL is public and the loan is not fundraising or the loan doesn't have the `anonymization_level` of `none`. Said a bit more succinctly:
> 	 * Rules for showing social information:
	 *  1) Admins and partners can see everything
	 *  2) Users can't see social info if the loan's been fully anonymized
	 *  3) Public can't see social info for non-fundraising or anonymized loans (both full and public levels)

That simply determines if the comment section is displayed at all. Adding comments is enabled in [these cases](https://github.com/kiva/kiva/blob/dbcff827bf51e4d8d5874838bf8a5da95da3b15a/sites/www_kiva/server/views/Xb/Loan/LoanView.php#L1090-L1111):

1. The logged-in user is a trustee
2. The logged-in user is an admin
3. The logged-in user is a trustee to the loan
4. They're a privileged user (trustee to the loan, partner to the loan, the borrower on the loan, or a lender on the loan) AND they've lent to at least one loan or it's a direct loan and they're the borrower for that loan
5. The loan is fundraising and they're a lender on the loan

This PR is an attempt to update the logic on the new UI to also hide/display the add comment functionality for the same reasons. I didn't thoroughly review the other comment-related logic from the legacy profile, though I see we set other booleans in the same section that we may want to review.

### DIFF
--- a/src/components/BorrowerProfile/FullBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FullBorrowerProfile.vue
@@ -379,7 +379,8 @@ export default {
 		showComments() {
 			// Comments remain privileged-only; additionally hide the section entirely on
 			// anonymized (full/pii) loans for anyone who is not an admin.
-			return this.isPrivileged && !(this.isAnonymized && !this.isAdmin);
+			const nonAdminViewingAnonymousLoan = this.isAnonymized && !this.isAdmin;
+			return this.isPrivileged && !nonAdminViewingAnonymousLoan;
 		},
 		shareCampaign() {
 			return this.inPfp ? 'social_share_bp_pfp' : 'social_share_bp';
--- a/src/components/BorrowerProfile/FullBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FullBorrowerProfile.vue
@@ -157,7 +157,7 @@
 				@hide-section="showTags = false"
 			/>
 			<loan-comments
-				v-if="isPrivileged"
+				v-if="showComments"
 				class="tw-mb-5 md:tw-mb-6 lg:tw-mb-8"
 				:loan-id="loanId"
 			/>
@@ -232,6 +232,7 @@ export const fullProfileQuery = gql`
 		}
 		my {
 			id
+			isAdmin
 			userPreferences {
 				id
 				preferences
@@ -285,6 +286,7 @@ export default {
 		},
 		result({ data }) {
 			this.loanData = data?.lend?.loan ?? {};
+			this.isAdmin = data?.my?.isAdmin ?? false;
 			this.userPreferences = data?.my?.userPreferences ?? null;
 			// Reconcile with localStorage (client-only); account preference still wins.
 			this.showDetailsInRail = resolveRailPreference({
@@ -343,6 +345,7 @@ export default {
 			// Initialize from the SSR-resolved prop so logged-in opted-in renders without flash.
 			showDetailsInRail: this.initialShowDetailsInRail,
 			userPreferences: null,
+			isAdmin: false,
 		};
 	},
 	computed: {
@@ -368,6 +371,15 @@ export default {
 		},
 		isPrivileged() {
 			return this.loanData?.userProperties?.isPrivileged ?? false;
+		},
+		isAnonymized() {
+			const level = this.loanData?.anonymizationLevel;
+			return level === 'full' || level === 'pii';
+		},
+		showComments() {
+			// Comments remain privileged-only; additionally hide the section entirely on
+			// anonymized (full/pii) loans for anyone who is not an admin.
+			return this.isPrivileged && !(this.isAnonymized && !this.isAdmin);
 		},
 		shareCampaign() {
 			return this.inPfp ? 'social_share_bp_pfp' : 'social_share_bp';
--- a/src/components/BorrowerProfile/LoanComments.vue
+++ b/src/components/BorrowerProfile/LoanComments.vue
@@ -184,10 +184,6 @@ const commentsQuery = gql`query loanCommentsFullList($loanId: Int!) {
 		loan(id: $loanId) {
 			id
 			status
-			borrowers {
-				id
-				firstName
-			}
 			comments {
 				values {
 					id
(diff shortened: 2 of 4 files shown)


src/components/BorrowerProfile/LoanComments.vue
```diff
@@ -285,6 +314,32 @@ export default {
...
+		isTrustee() {
+			return !!this.myTrusteeId;
+		},
+		// The logged-in user is the trustee for this specific loan
+		isTrusteeToLoan() {
+			return !!this.myTrusteeId && this.myTrusteeId === this.loanTrusteeId;
+		},
+		// The logged-in user is one of the borrowers/entrepreneurs on this loan
+		isBorrowerOfLoan() {
+			return this.myId != null && this.loanBorrowerIds.includes(this.myId);
```
kathrynlovett: I'm not sure about this - need to double check if the `borrower->id` is the same as the login/user ID 👀

kathrynlovett: Okay, I think `borrower { id }` is the person ID so we'll have to do this check some other way.

kathrynlovett: Updated it to check the last loan ID for a borrower. Seems not ideal because if they're a repeat borrower then this could return incorrectly for an older loan. Not sure how much we want to invest in enabling them to comment on their older loans, though?

src/components/BorrowerProfile/FullBorrowerProfile.vue
```diff
@@ -369,6 +372,15 @@ export default {
...
 			return this.loanData?.userProperties?.isPrivileged ?? false;
 		},
+		isAnonymized() {
+			const level = this.loanData?.anonymizationLevel;
+			return level === 'full' || level === 'pii';
+		},
+		showComments() {
+			// Comments remain privileged-only; additionally hide the section entirely on
+			// anonymized (full/pii) loans for anyone who is not an admin.
+			return this.isPrivileged && !(this.isAnonymized && !this.isAdmin);
```
dyersituations: This looks right, but the !() makes it tougher to read.

src/components/BorrowerProfile/LoanComments.vue
```diff
@@ -7,18 +7,21 @@
...
-				Comment
-			</label>
-			<textarea
-				id="bp-comment-field"
-				v-model="newCommentText"
-				class="tw-w-full tw-border tw-border-tertiary tw-rounded tw-p-1 tw-mb-1"
-				rows="4"
-				data-testid="bp-comment-form-textarea"
-			></textarea>
+			<template v-if="canAddComment">
```
dyersituations, line 10: Is this check needed, since it's happening on the parent?

src/components/BorrowerProfile/LoanComments.vue
```diff
@@ -285,6 +312,34 @@ export default {
...
+		},
+		// The logged-in user is the trustee for this specific loan
+		isTrusteeToLoan() {
+			return !!this.myTrusteeId && this.myTrusteeId === this.loanTrusteeId;
+		},
+		// The logged-in user is the borrower of this loan. The API has no per-loan borrower
+		// flag, so we approximate with the viewer's most recent borrowed loan id (the same
+		// signal ShareButton uses). This misses borrowers whose most recent loan isn't this one.
+		isBorrowerOfLoan() {
+			return this.myBorrowedLoanId != null && Number(this.myBorrowedLoanId) === this.loanId;
```
eddieferrer: Nit -- but this seems verbose, I'm pretty confident `return this.myBorrowedLoanId === this.loanId;` would work

src/components/BorrowerProfile/LoanComments.vue
eddieferrer: @kathrynlovett You can do my > borrowedLoans to get a full collection of all of a users loans.

src/components/BorrowerProfile/LoanComments.vue
kathrynlovett: Do you mean the `showComments` check? I think the `canAddComment` permissions are supposed to be different from `showComments` - we show comments in more cases than we allow users to add comments.

src/components/BorrowerProfile/LoanComments.vue
kathrynlovett: Adjusted this to check for all loans. Let me know if the updated version looks good or if you think it could be improved!

src/components/BorrowerProfile/FullBorrowerProfile.vue
kathrynlovett: Added another var to try and improve readability. Let me know what you think!

src/components/BorrowerProfile/LoanComments.vue
kathrynlovett: Thank you! Updated to use that 🙏🏻

src/components/BorrowerProfile/LoanComments.vue
dyersituations: Show there's a case where we show comments but they can't be added by the current user?

kathrynlovett: Right - you can view existing comments but you cannot add new comments.

kathrynlovett: There's some additional detail in the PR description about the original logic in the monolith and the difference between displaying comments and allowing users to add comments.

dyersituations (approved):

eddieferrer (approved):

PR #7082, opened 2026-07-21
fix: misc fixes and improvements to fb/meta tracking

dyersituations:
https://kiva.atlassian.net/browse/MP-3010

- Safety built into potential duplicate events
- Added amounts to transactions which should improve tracking on Meta side
- Added missing ATB event on loan card and express checkout
- Added `transactor` field to page view event
- Migrated to using shared method (didn't migrate entire analytics mixin)

### DIFF
--- a/src/components/LoanCards/Buttons/LendButton.vue
+++ b/src/components/LoanCards/Buttons/LendButton.vue
@@ -16,6 +16,7 @@ import updateLoanReservation from '#src/graphql/mutation/updateLoanReservation.g
 import loanCardBasketed from '#src/graphql/query/loanCardBasketed.graphql';
 import { handleInvalidBasket, hasBasketExpired } from '#src/util/basketUtils';
 import { KvButton } from '@kiva/kv-components';
+import { trackAddToCart } from '@kiva/kv-analytics';
 
 export default {
 	name: 'LendButton',
@@ -89,10 +90,8 @@ export default {
 					});
 				} else {
 					try {
-						// track facebook add to basket
-						if (typeof window !== 'undefined' && typeof fbq === 'function') {
-							window.fbq('track', 'AddToCart', { content_category: 'Loan' });
-						}
+						// Track facebook add to basket
+						trackAddToCart('Loan');
 					} catch (e) {
 						console.error(e);
 					}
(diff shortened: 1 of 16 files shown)


src/components/LoanCards/Buttons/LendButton.vue
```diff
@@ -89,10 +90,8 @@ export default {
 					});
 				} else {
 					try {
-						// track facebook add to basket
-						if (typeof window !== 'undefined' && typeof fbq === 'function') {
-							window.fbq('track', 'AddToCart', { content_category: 'Loan' });
-						}
+						// Track facebook add to basket
+						trackAddToCart('Loan', this.price);
```
mcstover: We should probably rename this method to be `fb` specific. Sorry, I didn't catch that in the utiltity pr :(

dyersituations: I did think about this, figured the import would be clear, but now realizing all the other methods have `Fb` in them... Here it is:

https://github.com/kiva/kv-ui-elements/pull/853

dyersituations: Rename changes pushed up.

mcstover: Thank you for that update!

src/plugins/kv-analytics-plugin.js
```diff
@@ -1,6 +1,20 @@
 /* eslint-disable no-underscore-dangle */
 import logFormatter from '#src/util/logFormatter';
 import SimpleQueue from '#src/util/simpleQueue';
+import { getUserType, trackFBCustomEvent } from '@kiva/kv-analytics';
+import { hasLentBeforeCookie, hasDepositBeforeCookie } from '#src/util/optimizelyUserMetrics';
```
mcstover, line 5: Can we take a future task to globalize these methods into kv-analytics?

src/plugins/kv-analytics-plugin.js
```diff
@@ -63,9 +80,18 @@ export default {
...
 				// facebook pixel pageview
 				if (fbLoaded) {
-					// we used to pass a user_type but it's always empty across the site
-					// { user_type: '???'}
-					window.fbq('track', 'PageView');
+					// Segment by transactor status (has ever lent or deposited) via the kvu_lb / kvu_db
+					// cookies. Uses the shared cookie-name constants and matches the exact 'true' value
+					// the cookies are written with (see optimizelyUserMetrics.setUserDataCookies), so the
+					// read stays in step with how cms derives the same flag.
+					const readCookie = name => {
```
mcstover, line 87: Can we use our cookieStore import here? It has a get method that also takes the name...

PR #7088, opened 2026-07-23
fix: [CP-2930] fix comment visibility

kathrynlovett:
Follow up to #7080

Updates logic around comments to address the following inconsistencies with the legacy profile:

* Comments should be visible unless the loan is anonymized, or the user is logged out and the loan is not in fundraising or doesn't have an anonymization level of `none`
* If comments are visible but the user cannot add comments, we show a message indicating that they cannot comment
* You can only comment on a fundraising loan if you have lent to that loan (tightens up this check), unless you meet one of the other cases

This matches our logic on the [legacy borrower profile](https://github.com/kiva/kiva/blob/dbcff827bf51e4d8d5874838bf8a5da95da3b15a/sites/www_kiva/server/views/Xb/Loan/LoanView.php#L1090-L1118). The logged-out/non-lender view on the legacy profile looks like this:

Screenshot 2026-07-22 at 4 08 22 PM

### DIFF
--- a/src/components/BorrowerProfile/LoanComments.vue
+++ b/src/components/BorrowerProfile/LoanComments.vue
@@ -19,6 +19,13 @@
 					data-testid="bp-comment-form-textarea"
 				></textarea>
 			</template>
+			<p
+				v-else-if="showMustBeLenderMessage"
+				class="tw-text-secondary tw-mb-1"
+				data-testid="bp-comment-must-be-lender"
+			>
+				You must be a lender to comment on this loan.
+			</p>
 			<div class="tw-flex tw-items-center tw-gap-2">
 				<kv-button
 					v-if="canAddComment"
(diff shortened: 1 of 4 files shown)


src/components/BorrowerProfile/LoanComments.vue
```diff
@@ -19,6 +19,13 @@
 					data-testid="bp-comment-form-textarea"
 				></textarea>
 			</template>
+			<p
+				v-else-if="showMustBeLenderMessage"
+				class="tw-text-secondary tw-mb-1"
+				data-testid="bp-comment-must-be-lender"
+			>
+				You must be a lender to comment on this loan.
```
mcstover, line 27: @kathrynlovett is this the same copy as our legacy BP comments notice? If so, cool. If not, maybe we could improve a little to make it clear that the lender "must have lent to this loan in order to comment..." or something along those lines.

kathrynlovett: @mcstover it's the same as our [legacy BP notice](https://github.com/kiva/kiva/blob/384c4c8872a2d8842dd1e50893d64b3e5d2311c7/sites/www_kiva/styleguide/source/_patterns/01-molecules/15-loan-sections/31-comments-and-updates.mustache#L16). I didn't want to change it without consulting with product/design/etc.

mcstover (approved):

dyersituations (approved):

PR #7098, opened 2026-07-23
fix(borrower-profile): show comments only to privileged users

kathrynlovett:
The comment section was visible to any logged-in user on a non-anonymized loan, and to logged-out visitors on fundraising loans with no anonymization. Gate it on `isPrivileged` instead, still hiding it on full/pii anonymized loans. Drops the now-unused `isFundraising` computed.

Claude-Session: https://claude.ai/code/session_01Tnp7Uba43ZJtwmiNVYTPCq

During validation, I realized that the earlier changes likely broke this functionality in some cases. The new profile has a "comment and update" carousel near the top of the loan. For non-privileged users that should be able to see comments (e.g., logged out users viewing a fundraising loan) that is where the comment displays. The section lower down that also displays comments and allows users to add, subscribe, or flag comments should not be displayed unless it is a privileged user.

### DIFF
--- a/src/components/BorrowerProfile/FullBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FullBorrowerProfile.vue
@@ -373,15 +373,10 @@ export default {
 			const level = this.loanData?.anonymizationLevel;
 			return level === 'full' || level === 'pii';
 		},
-		isFundraising() {
-			return this.loanData?.status === 'fundraising';
-		},
 		showComments() {
-			// Mirrors legacy showSocialInfo: logged-in users see the section unless
-			// the loan is anonymized (full/pii); logged-out visitors see it only on
-			// fundraising loans with no anonymization at all.
-			if (this.isLoggedIn) return !this.isAnonymized;
-			return this.isFundraising && (this.loanData?.anonymizationLevel ?? 'none') === 'none';
+			// Only privileged users (e.g. lenders on this loan) can see the comment
+			// section, and only while the loan is not anonymized (full/pii).
+			return this.isPrivileged && !this.isAnonymized;
 		},
 		shareCampaign() {
 			return this.inPfp ? 'social_share_bp_pfp' : 'social_share_bp';
--- a/src/components/BorrowerProfile/FullBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FullBorrowerProfile.vue
@@ -157,7 +157,7 @@
 				@hide-section="showTags = false"
 			/>
 			<loan-comments
-				v-if="showComments"
+				v-if="showAddCommentsSection"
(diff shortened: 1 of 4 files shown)


src/components/BorrowerProfile/FullBorrowerProfile.vue
```diff
@@ -373,15 +373,10 @@ export default {
...
-		isFundraising() {
-			return this.loanData?.status === 'fundraising';
-		},
 		showComments() {
-			// Mirrors legacy showSocialInfo: logged-in users see the section unless
-			// the loan is anonymized (full/pii); logged-out visitors see it only on
-			// fundraising loans with no anonymization at all.
-			if (this.isLoggedIn) return !this.isAnonymized;
-			return this.isFundraising && (this.loanData?.anonymizationLevel ?? 'none') === 'none';
+			// Only privileged users (e.g. lenders on this loan) can see the comment
```
mcstover, line 377: Maybe we should specify these controls are only for the "commenting" functionality, not the general display of the comments against a loan which I beleive should remain visible always...

mcstover (approved):

src/components/BorrowerProfile/FullBorrowerProfile.vue
kathrynlovett: Good call!

kathrynlovett:
@mcstover I updated the variable name to be a bit clearer. I also removed the logical gate I'd put in `LoanComments` before when I thought it was supposed to be displayed for more users. That check is now redundant with these changes.

src/components/BorrowerProfile/FullBorrowerProfile.vue
```diff
@@ -373,15 +373,10 @@ export default {
...
-		showComments() {
-			// Mirrors legacy showSocialInfo: logged-in users see the section unless
-			// the loan is anonymized (full/pii); logged-out visitors see it only on
-			// fundraising loans with no anonymization at all.
-			if (this.isLoggedIn) return !this.isAnonymized;
-			return this.isFundraising && (this.loanData?.anonymizationLevel ?? 'none') === 'none';
+		showAddCommentsSection() {
+			// Only privileged users (e.g. lenders on this loan) can see the comment
+			// section, and only while the loan is not anonymized (full/pii).
+			return this.isPrivileged && !this.isAnonymized;
```
mcstover, line 379: Do these conditions already cover the admin, trustee and borrower cases?

kathrynlovett: I think so - based on the documentation for the field we use to populate `isPrivileged`, it should cover those cases: 
Screenshot 2026-07-23 at 4 16 31 PM

mcstover (approved):

dyersituations (commented):
👍

PR #7113, opened 2026-07-28
feat: mp-3012-meta-tracking-new-events-pt-1

infante-jaime:
<html><head></head><body><h2>Summary</h2><p>Adds two Meta custom events for logged-in lenders who return and complete a loan purchase:</p><ul><li><code>lapsedLenderReEngaged</code></li><li><code>idleLenderReEngaged</code></li></ul><h2>Behavior</h2><ul><li>Captures the lender’s lifecycle data before transaction completion.</li><li>Derives lifecycle stage from registration date and loan purchase history.</li><li>Fires only when the transaction contains at least one loan.</li><li>Does not fire for deposit-, donation-, or Kiva Card-only transactions.</li><li>Applies only to the active checkout flow.</li></ul><h2>Lifecycle Stages</h2><div><div>
Event | Stages
-- | --
idleLenderReEngaged | idle90, idle180, idle365
lapsedLenderReEngaged | lapsedChurned

</div></div><h2>Event Parameters</h2><ul><li><code>lifecycleStage</code></li><li><code>daysSinceLastLoan</code></li><li><code>loanTotal</code></li><li><code>itemTotal</code></li></ul><p>Standard Purchase tracking continues through <code>@kiva/kv-analytics</code>. Re-engagement is tracked separately using its shared custom-event utility.</p><h2>Documentation</h2><p>Updated the Meta event inventory with both events, their parameters, and loan-purchase-only behavior.</p><h2>Validation</h2><ul><li><div></div>Full lint passes with no errors</li><li><div></div>Full unit test suite passes</li><li><div></div>Lifecycle boundary coverage</li><li><div></div>Loan-only event coverage</li><li><div></div>Analytics failure handling coverage</li></ul></body></html>


src/components/Checkout/InContext/InContextCheckout.vue
```diff
@@ -151,8 +151,10 @@
 </template>
 
 <script>
+import { inject } from 'vue';
```
dyersituations: This component is just used in the old corporate pages, which are now managed by the CPS impact dashboard, so I think we can revert this file.

src/plugins/kv-analytics-plugin.js
```diff
@@ -234,11 +234,16 @@ export default {
...
 						currency: 'USD',
 						value: itemTotal,
 						content_type: transactionData.isFTD ? 'FirstTimeDepositor' : 'ReturningLender'
-					});
+					};
+					// guests have no lender record, so no stage to report
+					if (transactionData.lifecycleStage) {
+						purchaseParams.lifecycleStage = transactionData.lifecycleStage;
+					}
+					window.fbq('track', 'Purchase', purchaseParams);
```
dyersituations: This will need to update to use the latest util from my recent PRs related to FB tracking.

src/util/lifecycleStage.js
```diff
@@ -0,0 +1,214 @@
...
+			stage: deriveLifecycleStage({ memberSince, loanPurchaseCount, lastLoanPurchase }, now),
+			daysSinceLastLoan,
+			alreadyReEngaged: hasReEngagedSinceIdle(
+				daysSinceLastLoan,
+				// whichever of the two happened more recently
+				Math.min(daysSinceNewest(data?.my?.lastDeposit), daysSinceNewest(data?.my?.lastDonation)),
+			),
+		};
+	} catch (e) {
+		console.error(e);
```
dyersituations: `logFormatter`

src/plugins/kv-analytics-plugin.js
```diff
@@ -234,11 +234,16 @@ export default {
...
-					window.fbq('track', 'Purchase', {
+					const purchaseParams = {
 						currency: 'USD',
 						value: itemTotal,
 						content_type: transactionData.isFTD ? 'FirstTimeDepositor' : 'ReturningLender'
-					});
+					};
+					// guests have no lender record, so no stage to report
+					if (transactionData.lifecycleStage) {
+						purchaseParams.lifecycleStage = transactionData.lifecycleStage;
```
dyersituations: Why add `lifecycleStage` here if we also have a separate custom event?

src/util/checkoutUtils.js
```diff
@@ -44,9 +45,28 @@ export function formatTransactionData(
...
+	// Which qualifying actions this transaction contains. A basket can hold more than
+	// one, so these are reported as a list rather than a single label — marketing needs
+	// to tell a deposit-only return apart from one that also included a loan.
+	// Kiva Cards do not qualify on their own, though buying one usually needs a deposit.
+	const reEngagementTriggers = [];
+	if (loans.length) {
+		reEngagementTriggers.push('loan');
+	}
+	// parsed rather than compared to '0.00', which reads a missing amount as a deposit
+	if (numeral(depositTotal).value() > 0) {
```
dyersituations: I thought we were just checking for loan transactions, not deposit/donation?

src/util/checkoutUtils.js
```diff
@@ -104,6 +128,63 @@ export function myFTDQuery(apollo) {
...
+ * already paid.
+ *
+ * @param {Object} apollo Apollo Client instance
+ * @param {Promise<Object|null>|null} lifecycleDataPromise Pre-transaction lifecycle request
+ * @returns {Promise<Object>}
+ */
+export async function getTransactionAnalyticsData(apollo, lifecycleDataPromise) {
+	let timer;
+	const timeout = new Promise(resolve => {
+		timer = setTimeout(() => resolve([null, null]), ANALYTICS_TIMEOUT_MS);
```
dyersituations: Why is this 10s exit needed? We ran `myFTDQuery` previously before redirect before? Is there that much increased risk with the new query?

src/util/checkoutUtils.js
```diff
@@ -104,6 +128,63 @@ export function myFTDQuery(apollo) {
...
+			isFTD: ftdResponse?.data?.my?.userAccount?.isFirstTimeDepositor,
+			lifecycleStage: lifecycleData?.stage ?? null,
+			daysSinceLastLoan: lifecycleData?.daysSinceLastLoan ?? null,
+			// "re-engaged" marks the return itself, so it fires once per inactive period
+			reEngagementEvent: lifecycleData?.alreadyReEngaged
+				? null
+				: getReEngagementEvent(lifecycleData?.stage),
+		};
+	} catch (e) {
+		console.error(e);
```
dyersituations: `logFormatter`

src/util/checkoutUtils.js
```diff
@@ -104,6 +128,63 @@ export function myFTDQuery(apollo) {
 	});
 }
 
+/**
+ * Checkout blocks the thanks-page redirect on these lookups, so they must always
+ * settle. Past this point we report what we have and let the lender through.
+ */
+const ANALYTICS_TIMEOUT_MS = 10000;
+
+function emptyTransactionAnalyticsData() {
```
dyersituations: Why is this method needed?

src/components/Checkout/InContext/InContextCheckout.vue
infante-jaime: reverted

src/plugins/kv-analytics-plugin.js
infante-jaime: after merging is ok I was just comparing to an previous version.

src/util/lifecycleStage.js
infante-jaime: updated

src/plugins/kv-analytics-plugin.js
infante-jaime: Agreed—it was redundant. After merging the latest Meta tracking changes, lifecycleStage is no longer added to Purchase and is only sent with the custom re-engagement event.

src/util/checkoutUtils.js
infante-jaime: Updated to loan transactions only. Deposit- and donation-only transactions no longer trigger re-engagement events, and the related query and suppression logic were removed.

src/util/checkoutUtils.js
infante-jaime: not needed really just an extra guard but I guess we can live without it

src/util/checkoutUtils.js
infante-jaime: updated

src/util/checkoutUtils.js
infante-jaime: Agreed—it was only used once and no longer provided meaningful reuse. I removed the helper and inlined the fallback response.

src/util/checkoutUtils.js
```diff
@@ -104,6 +112,42 @@ export function myFTDQuery(apollo) {
...
+ * Never throws. Analytics failures must not prevent transaction tracking or redirect.
+ *
+ * @param {Object} apollo Apollo Client instance
+ * @param {Promise<Object|null>|null} lifecycleDataPromise Pre-transaction lifecycle request
+ * @returns {Promise<Object>}
+ */
+export async function getTransactionAnalyticsData(apollo, lifecycleDataPromise) {
+	try {
+		const [ftdResponse, lifecycleData] = await Promise.all([
+			myFTDQuery(apollo),
```
dyersituations: Since these are both `my` queries, is there a reason to not combine into a single new query?

src/util/lifecycleStage.js
```diff
@@ -0,0 +1,178 @@
...
+};
+
+/**
+ * Whole days elapsed between a date and now
+ *
+ * @param {String|Number|Date} date
+ * @param {Date} now
+ * @returns {Number|null} null when the date is missing or unparseable
+ */
+function daysSince(date, now) {
```
dyersituations: Do we have a date util file this should live in?

dyersituations: Or does `date-fns` have this functionality? We already use that dependency.

infante-jaime: Moved it into dateUtils.js next to toValidDate.

src/util/checkoutUtils.js
infante-jaime: Done. I also replaced totalCount with limit: 2 + values.length, since we only need to know none, one, or many.

Tradeoff: Express Checkout and DonateRepaymentsToggle fetch two extra rows, but Checkout now uses one request instead of two.

infante-jaime: "Isn't there added risk here to the checkout page initial loading with adding this extra somewhat slower query? Thoughts on keeping it separate in a new query?" 

Yeah, there's real risk keeping it together. Splitting it back out costs an extra query with its own overhead, but it'd run fewer times and only where we actually need it. Your original point was about combining it with myFTDQuery, but that gets messy on timing too — lifecycle is read at entry, FTD at completion. So I got carried away folding it into initializeCheckout instead.

I think keeping it separate is better. What do you think?

dyersituations: Yeah not adding extra query time to checkout page load would probably be a good path.

infante-jaime: yes let me get that back as it was

dyersituations: Do consider combining the two queries that were run in parallel in the last version.

infante-jaime: yes reverted and queries run in parallel

src/util/checkoutUtils.js
```diff
@@ -104,6 +112,41 @@ export function myFTDQuery(apollo) {
...
+ */
+export async function getTransactionAnalyticsData(apollo, lifecycleDataPromise) {
+	// Only the FTD query starts here; the lifecycle request began at checkout entry and
+	// is usually settled already. They are awaited together so neither blocks the other.
+	const [ftdResult, lifecycleResult] = await Promise.allSettled([
+		myFTDQuery(apollo),
+		lifecycleDataPromise,
+	]);
+
+	if (ftdResult.status === 'rejected') {
```
dyersituations: It looks like this is a new check in this changeset that wasn't migrated from where this query was called previously, why is this check and log needed?

src/util/checkoutUtils.js
```diff
@@ -104,6 +112,41 @@ export function myFTDQuery(apollo) {
...
+		logFormatter('Failed to resolve FTD status for transaction analytics', 'error', {
+			error: ftdResult.reason?.message,
+		});
+	}
+
+	// Unrelated signals, so losing one must not discard the other
+	const lifecycleData = lifecycleResult.status === 'fulfilled' ? lifecycleResult.value : null;
```
dyersituations: What is this check?

src/util/checkoutUtils.js
```diff
@@ -104,6 +112,41 @@ export function myFTDQuery(apollo) {
...
+ * @param {Promise<Object|null>|null} lifecycleDataPromise The lookup started on checkout
+ *   entry, since the purchase being tracked is what moves a lender out of idle or lapsed
+ * @returns {Promise<Object>}
+ */
+export async function getTransactionAnalyticsData(apollo, lifecycleDataPromise) {
+	// Only the FTD query starts here; the lifecycle request began at checkout entry and
+	// is usually settled already. They are awaited together so neither blocks the other.
+	const [ftdResult, lifecycleResult] = await Promise.allSettled([
+		myFTDQuery(apollo),
+		lifecycleDataPromise,
```
dyersituations, line 130: Still wondering why we don't combine these two queries into a new query instead of running two `my` queries in parallel here. Is `myFTDQuery` possibly cached from another use?

infante-jaime: It's not cached. `isFirstTimeDepositor` is only used in `myFTD.graphql`, so nothing else is querying it.

Only the FTD query runs here. The lifecycle request already starts when the page loads in the `initializeCheckout` handler, so this line is just waiting for it to finish.

We can't combine them into a single query because of timing. Either the FTD value is read before the deposit, or the lifecycle stage is read after the purchase has already changed it.

src/util/checkoutUtils.js
infante-jaime: removed

src/util/checkoutUtils.js
infante-jaime: removed

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -427,6 +428,9 @@ const getLoanIds = loans => (loans ?? []).map(l => l.id).filter(id => !!id);
 
 export default {
 	name: 'CheckoutPage',
+	setup() {
+		return useLifecycleCapture(inject('apollo'));
```
roger-in-kiva: weird injecting apollo twice in this page, any way to simplify?

infante-jaime: Good catch. The composable only needs apollo at the moment it starts the lookup, not when it's constructed — so it now takes it as an argument and setup() doesn't inject at all. The call site in the result handler already has this.apollo from the Options inject.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -568,6 +572,12 @@ export default {
 			// user data
 			this.myBalance = _get(data, 'my.userAccount.balance');
 			this.myId = _get(data, 'my.userAccount.id');
+			// Start the lifecycle lookup now, while it still reflects the lender as they
+			// arrived: the purchase being tracked is what moves them out of an idle or
+			// lapsed stage. Starts once, so re-running this handler is harmless.
+			if (this.myId) {
+				this.startLifecycleCapture(this.apollo);
+			}
```
mcstover: Why can't this run on mounted? We don't appear to use it until the transaction is completed which requires manual steps so there should be ample time...

infante-jaime: You're right, timing isn't the issue. myId just isn't always available in mounted(). On a hydrated load it's already there, but on a cold SPA navigation it arrives later, so mounted() would miss it.

I moved it to a watch on myId instead, which covers both cases and keeps the side effect out of the data handler.

dyersituations: When you mention "it arrives later" when is it arriving?

infante-jaime: I was wrong about SPA nav, preFetchAll runs on client navigation too so myId is there by mounted() either way.

The watcher is really for mid-checkout login: a guest logs in, the prefetch re-runs, and myId arrives after mounted. Updated the comment.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -717,6 +721,16 @@ export default {
...
+		// Start the lifecycle lookup as soon as we know there is a lender. Usually that
+		// is during created, but a guest who logs in partway through checkout only gets
+		// an id once doPopupLogin re-runs the prefetch, which is well after mounted.
+		// Either way it must happen before the transaction: the purchase being tracked
+		// is what moves a lender out of an idle or lapsed stage.
+		myId(myId) {
+			if (myId) {
+				this.startLifecycleCapture(this.apollo);
+			}
+		},
```
mcstover: Login always requires a full page refresh. The general query that will have the user id will run on the server after that so the user id will be present when mounted runs.

mcstover: Having this in a watch risks it running on the server and potentially not completing as it's not an async await. I just don't see any need for it to run on the server.

infante-jaime: Agreed, and already moved it to mounted(). Your point about the server is correct — a watcher would just fire the request and forget it, with nothing awaiting it.

Also dropped the typeof window === 'undefined' guard I had in the composable; mounted() makes it unreachable.

mcstover (approved):
Thanks @infante-jaime for the additional updates!

roger-in-kiva (approved):

dyersituations:
👍

PR #7118, opened 2026-07-29
fix(logFormatter): stop losing caught errors, and fix 17 wrong-level calls

eddieferrer:
## The problem

`logFormatter(message, type, meta)` serializes with `JSON.stringify`, and a plain
`Error`'s `message` and `stack` are non-enumerable. So a caught Error passed as the
message logged this:

```
{"meta":{},"level":"error","message":{}}
```

The failure itself was gone. 28 log lines in `npm run unit` read `"message":{}`,
which is how this was found.

Separately, **17 call sites passed a failure description where the level belongs**,
so it matched no case in the switch and fell through to `console.log` — those
errors went to **stdout at log level and never reached error alerting**:

```js
logFormatter(error, 'Failed to fetch categories progress by year');
// → stdout: {"meta":{},"level":"Failed to fetch categories progress by year","message":{}}
```

Two of them passed `'warning'`, which is not a level either (`'warn'` is).

## The fix — two commits

**1. The utility unwraps Errors**, in the message slot and inside `meta`.

This deliberately makes kiva/ui match **cms-page-server**, whose `logFormatter`
already does the same thing (`message.message ?? message`). kiva/ui was the
outlier. After this, `logFormatter(e, 'error')` behaves the same in both repos.

Level handling is untouched: an unrecognised type still logs as a plain log. That's
the documented contract, three existing tests pin it, and CPS behaves identically —
not something a call-site bug should override.

**2. The 17 level fixes** — description moved to the message slot, `'error'` (or
`'warn'`) into the level slot, with context that was already in scope:

```js
- logFormatter(error, 'Failed to fetch loan stats by year');
+ logFormatter('Failed to fetch loan stats by year', 'error', { error, year });
```

This is a bug rather than a format choice. CPS can't have it because its `LogLevel`
is a typed union, so nothing here diverges from that repo.

## Why the call sites are otherwise untouched

An earlier revision of this PR rewrote all 62 call sites to
`logFormatter('what failed', 'error', { error })`. **That was reverted.**

`logFormatter(e, 'error')` is the dominant shape in cms-page-server — 115 of its 169
call sites — and that repo's utility already handles it correctly. Rewriting
kiva/ui's call sites would have made the two repos diverge on the common shape for
no functional gain, since the utility change alone stops the data loss.

If we later want richer logs (a description of *what* broke plus context, with the
error in meta), that's worth doing — but as a deliberate convention change across
both repos, not unilaterally inside a bug fix.

## Eight message strings changed text

In `KvAtbModalContainer`, `BailoutChips` and `useMyKivaHome`, the strings named a
query (`'MyKiva LendMenuQuery'`) or a location (`'useMyKivaHome composable'`) rather
than a failure; one had a trailing space. Since these were logging at `log` level to
stdout, nothing could have been alerting on the old strings — but shout if a
dashboard greps them. The seven in `useGoalData` were already prose and are
unchanged.

## Result

| | before | after |
|---|---|---|
| Lines logging `"message":{}` | 28 | **0** |
| Sites logging at `log` instead of `error` | 17 | **0** |
| `logFormatter.spec.js` | 15 tests | 21 tests |

The utility change alone fixed all 28 — it does not depend on call sites being
updated, which is what makes the revert above safe.

## Testing

- `npm run unit` — 302 files / 4453 tests pass
- `npm run lint` — clean
- All 15 pre-existing `logFormatter` tests pass **untouched**; 6 added, each verified
  to fail against the previous implementation. One is a regression guard proving a
  plain object in `meta` still passes through whole rather than being flattened —
  the case that separates this from CPS's looser `message.message ?? message`
- 2 assertions in `useGoalData.spec.js` pinned the old call shape and are updated

## Follow-up

#7119 fixes 10 unrelated error-handling defects the audit surfaced in the same catch
blocks — including `basketUtils`, where `this.loanId` in a plain module function
throws and stops `Sentry.captureException` from ever running. It now targets `main`
directly and can be reviewed independently of this PR.

### DIFF
--- a/src/util/logFormatter.js
+++ b/src/util/logFormatter.js
@@ -1,13 +1,33 @@
 // generate various console types using a stringified message
 // and a format that matcher our server request logs
 
+// An Error's `message` and `stack` are non-enumerable, so JSON.stringify reduces
+// a plain Error to `{}` and the failure is lost. Callers should pass text as the
+// message and hand the error to `meta`, but unwrap Errors here as well so that a
+// stray one still records something, and so `{ error }` is always safe to write
+// at a call site regardless of what kind of error was caught.
+const errorMessage = value => {
+	if (!(value instanceof Error)) return null;
+	// An Error carrying an empty message would be dropped by the falsy guard
+	// below, turning a real failure back into no log at all, so fall back to its
+	// string form ('Error', or 'TypeError: ...' for a subclass).
+	return value.message || String(value);
+};
+
 export default (message, type, meta = {}) => {
-	if (!message || message === '') return false;
+	const text = errorMessage(message) ?? message;
+
+	if (!text || text === '') return false;
+
+	const detail = { ...meta };
+	Object.keys(detail).forEach(key => {
+		detail[key] = errorMessage(detail[key]) ?? detail[key];
+	});
 
 	const stringifiedMessage = JSON.stringify({
-		meta,
+		meta: detail,
 		level: type || 'log',
-		message,
+		message: text,
 	});
 
 	switch (type) {
--- a/src/util/logFormatter.js
+++ b/src/util/logFormatter.js
@@ -15,14 +15,19 @@ const errorMessage = value => {
 };
 
 export default (message, type, meta = {}) => {
-	const text = errorMessage(message) ?? message;
-
-	if (!text || text === '') return false;
-
 	const detail = { ...meta };
 	Object.keys(detail).forEach(key => {
 		detail[key] = errorMessage(detail[key]) ?? detail[key];
 	});
+
+	let text = errorMessage(message) ?? message;
+
+	// A call with no message but real context still describes something that
+	// happened, so drop it only when there would be nothing at all to record.
+	if (!text || text === '') {
+		if (!Object.keys(detail).length) return false;
+		text = '(no message)';
+	}
 
 	const stringifiedMessage = JSON.stringify({
 		meta: detail,
--- a/src/components/MyKiva/BailoutChips.vue
+++ b/src/components/MyKiva/BailoutChips.vue
@@ -90,7 +90,7 @@ const fetchCategories = () => {
 		});
 		categories.value = categories.value.sort((a, b) => a.name.localeCompare(b.name));
 	}).catch(error => {
-		logFormatter(error, 'MyKiva LendMenuQuery');
+		logFormatter('Failed to load MyKiva lend menu categories', 'error', { error });
 	});
 };
 
@@ -127,7 +127,7 @@ const fetchCountries = () => {
 			};
 		});
 	}).catch(error => {
-		logFormatter(error, 'MyKiva CountryListQuery');
+		logFormatter('Failed to load MyKiva country list', 'error', { error });
 	});
 };
 
--- a/src/components/WwwFrame/Header/KvAtbModalContainer.vue
+++ b/src/components/WwwFrame/Header/KvAtbModalContainer.vue
@@ -112,7 +112,7 @@ const fetchUserData = async () => {
 		userData.value = data ?? null;
 		hasEverLoggedIn.value = data?.hasEverLoggedIn ?? false;
 	}).catch(e => {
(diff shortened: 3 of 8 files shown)


src/util/logFormatter.js
```diff
@@ -1,13 +1,27 @@
...
+// An Error's `message` and `stack` are non-enumerable, so JSON.stringify reduces
+// a plain Error to `{}` and the failure is lost. Callers should pass text as the
+// message and hand the error to `meta`, but unwrap Errors here as well so that a
+// stray one still records something, and so `{ error }` is always safe to write
+// at a call site regardless of what kind of error was caught.
+const errorMessage = value => (value instanceof Error ? value.message : null);
+
 export default (message, type, meta = {}) => {
-	if (!message || message === '') return false;
+	const text = errorMessage(message) ?? message;
```
mcstover: This seems odd after you just added messages to so many locations...I guess if most instances are still passing the error object in the message param it works but if we're mostly passing strings as messages now, maybe we should swap the order...

dyersituations: Yeah curious why we don't handle the message being passed in here instead of adding new custom messages and passing in the error in meta.

eddieferrer: > This seems odd after you just added messages to so many locations...I guess if most instances are still passing the error object in the message param it works but if we're mostly passing strings as messages now, maybe we should swap the order...

Yeah the message unwrapping is not necessary now and we can remove it, just extra defensive code.

eddieferrer: > Yeah curious why we don't handle the message being passed in here instead of adding new custom messages and passing in the error in meta.

error.message is the cause, not what broke. "Network error: 503" is identical for a failed comment post, profile save, or checkout, and since aggregators group by message you can't alert on "profile saves are failing." The authored message says which operation failed, the error text stays in meta, and loanId/basketId/etc. weren't being logged at all before. This new format gives us all the information. 

That said, the 31 ApolloError sites already logged their message, so that commit buys grouping and context rather than recovering lost data. We could drop it and keep the utility + the 17 level fixes (~12 files instead of 40) to minimize churn

dyersituations: What is your recommendation?

eddieferrer: That being said, the `logFormatter(e, 'error');` format is still pervasive in CPS, so I might rework this so both repos follow the same pattern

mcstover: Yea I think we were generally trying to handle passing in either a message string or the error object and CPS already has some very basic handling for that in a way this repo did not.

I'm generally ok with the improvements here...

src/util/logFormatter.js
```diff
@@ -1,13 +1,33 @@
...
+// a plain Error to `{}` and the failure is lost. Callers should pass text as the
+// message and hand the error to `meta`, but unwrap Errors here as well so that a
+// stray one still records something, and so `{ error }` is always safe to write
+// at a call site regardless of what kind of error was caught.
+const errorMessage = value => {
+	if (!(value instanceof Error)) return null;
+	// An Error carrying an empty message would be dropped by the falsy guard
+	// below, turning a real failure back into no log at all, so fall back to its
+	// string form ('Error', or 'TypeError: ...' for a subclass).
+	return value.message || String(value);
```
dyersituations, line 14: Will the fallback here ever be hit due to the `!(value instanceof Error)` check above? I assume all `Error` have `.message`.

src/util/logFormatter.js
```diff
@@ -1,13 +1,33 @@
...
+};
+
 export default (message, type, meta = {}) => {
-	if (!message || message === '') return false;
+	const text = errorMessage(message) ?? message;
+
+	if (!text || text === '') return false;
```
dyersituations: Do we want to return here or use a fallback message? For example if there was no `message`, but we still have `type` + `meta`?

eddieferrer: good idea, added fallback.

dyersituations (approved):

mcstover (approved):

PR #7119, opened 2026-07-29
fix: repair error handling defects found while auditing logFormatter

eddieferrer:
## What

Nine error-handling defects found while auditing `logFormatter` call sites across
the repo. None of them are logFormatter misuse — they're bugs that happened to live
in the same catch blocks, so it was cheaper to find them than to find them later.

Independent of #7118 and reviewable on its own.

## The significant one

**`basketUtils.js:17` — set-lend-amount errors have never reached Sentry.**

```js
function logSetLendAmountError(loanId, err) {
    logFormatter(err, 'error');
    try {
        Sentry.withScope(scope => {
            scope.setTag('loan_id', this.loanId);   // ← `this` is undefined here
            scope.setTag('mutation', 'addToBasket');
            Sentry.captureException(err);            // ← never runs
        });
    } catch (e) {
        // no-op                                     ← hides all of it
    }
}
```

This is an ES module, so strict mode, and the function is called as a plain function
(`:69`, `:77`) — `this` is `undefined`. It throws a `TypeError` on the first
`setTag`, so `captureException` on the next line never executes, and the no-op catch
swallows the evidence. Verified by reproduction. Fix is `loanId`.

## Unguarded property access that masks the real error

- **`useGoalData.js:303`** — `response.data.userAchievementProgress` had no optional
  chaining, so a response without `data` threw a `TypeError` and the catch logged
  *that* rather than the actual query failure.

  Worth merging alongside #7118: once Errors stop serializing to `{}`, this site
  starts reporting `"Cannot read properties of undefined (reading
  'userAchievementProgress')"` instead of the real cause.

- **`EmailSettings.vue`** ×3 — `errors[0].extensions.code` could throw *inside* the
  `throw`, masking the real error. `errors[0]` can itself be undefined, since the
  guard is `if (response.errors)` and `[]` is truthy.

- **`borrower-profile-exp-mixin.js`** ×3 and **`ProcessInstantLending.vue`** ×2 —
  unguarded `error.message` inside no-op catch blocks. Same shape as the
  `basketUtils` bug: it throws before the `Sentry.captureMessage` on the next line,
  and the empty catch hides it.

## The rest

- **`useBorrowerProfileData.js:109`** — a `.then()` with no `.catch`, where the
  enclosing `try/catch` cannot catch the rejection. Silent unhandled rejection on a
  postCheckoutAchievements failure.
- **`CCLandingPage.vue:1070`** — an `Error` object was assigned to
  `promoErrorMessage`, which is passed to child components and rendered, so it
  leaked `"Error: "` into user-facing copy. Every other assignment in that file
  stores `errors[0].message`.
- **`OrderTotals.vue:424`** — a raw `console.error` bypassing the structured log
  format entirely.
- **`ProcessInstantLending.vue`** — removed a stray `console.log(error)` left in
  production directly above the logFormatter call.

## Deliberately not changed

The remaining `catch { // no-op }` blocks stay. They're intentional guards around
third-party and analytics calls, and now that the unguarded `error.message` inside
them is fixed, they no longer mask real failures.

Log calls use the existing `logFormatter(e, 'error')` shape, matching the rest of
this repo and cms-page-server. This PR introduces no new logging convention.

## Testing

- `npm run unit` — 302 files / 4447 tests pass, exit 0
- `npm run lint` — clean
- No test assertions needed updating

### DIFF
--- a/src/components/Checkout/OrderTotals.vue
+++ b/src/components/Checkout/OrderTotals.vue
@@ -421,7 +421,7 @@ export default {
 				this.$kvTrackEvent('basket', 'Kiva Credit', 'Apply Credit Success');
 				this.$emit('refreshtotals');
 			}).catch(error => {
-				console.error(error);
+				logFormatter(error, 'error');
 				this.setUpdating(false);
 			});
 		},
--- a/src/composables/useBorrowerProfileData.js
+++ b/src/composables/useBorrowerProfileData.js
@@ -108,6 +108,10 @@ export default function useBorrowerProfileData(apolloClient, cookieStore) {
 				variables: { loanIds: [loanDataId] },
 			}).then(result => {
 				achievementsData.value = result;
+			}).catch(e => {
+				// The enclosing try/catch cannot catch this rejection, so without a
+				// handler here a failure is an unhandled rejection.
+				logFormatter(e, 'error');
 			});
 		} catch (e) {
 			subscription.value?.unsubscribe();
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -300,7 +300,7 @@ export default function useGoalData({ apollo } = {}) {
 				variables: { year },
 				fetchPolicy: 'no-cache',
 			});
-			const progress = response.data.userAchievementProgress.tieredLendingAchievements;
+			const progress = response.data?.userAchievementProgress?.tieredLendingAchievements ?? null;
 			return progress;
 		} catch (error) {
 			logFormatter(error, 'Failed to fetch categories progress by year');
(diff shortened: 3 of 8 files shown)


mcstover (approved):

dyersituations (approved):

src/components/Checkout/OrderTotals.vue
```diff
@@ -421,7 +421,7 @@ export default {
 				this.$kvTrackEvent('basket', 'Kiva Credit', 'Apply Credit Success');
 				this.$emit('refreshtotals');
 			}).catch(error => {
-				console.error(error);
+				logFormatter('Failed to apply Kiva credit', 'error', { error, creditType: type });
```
dyersituations: Similar to my recent comment on the other PR: why pass in meta data instead of having `logFormatter` handle different error message formats in the first parameter.

mcstover (approved):

dyersituations (approved):

PR #7126, opened 2026-07-30
fix: loans page migration design feedback

dyersituations:
https://kiva.atlassian.net/browse/MP-2940
https://kiva.atlassian.net/browse/MP-2967

Design feedback for loans beta table (sizing, shadows, sticky scrolling, etc.)

Missed another helper text that needed to be darker in the last PR for withdraw beta

image

### DIFF
--- a/src/components/Portfolio/LoanList.vue
+++ b/src/components/Portfolio/LoanList.vue
@@ -1,12 +1,15 @@
 <template>
 	<div class="tw-mt-2">
 		<div class="tw-relative">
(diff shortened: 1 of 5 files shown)


mcstover:
Why would we have vertical scrolling here if we also have pagination? How does this work on mobile?

dyersituations:
> Why would we have vertical scrolling here if we also have pagination? How does this work on mobile?

It was a request from design for something to try to make the table more easily consumable. I could see it being tweaked further after another round of design testing once deployed. It works fairly well in mobile:
image

mcstover (approved):

PR #7128, opened 2026-07-31
feat(apollo): prefetch for composables AD-360

emuvente:
Since the Vue 3 migration, component `apollo` options blocks have been the only way into SSR prefetch. Composables, the reuse primitive the migration was supposed to unlock, had no legitimate way to fetch render state: their only option was a raw `apollo.query()` inside `setup()`, which the prefetch machinery cannot see. The data always arrived after first render, so the server could never accurately display the value. So data-driven logic stayed trapped in components and mixins, and every composable that tried anyway shipped the same class of bug.

**This PR removes that limitation**. Composables can now prefetch GraphQL data. A composable registers its operations that need to be prefetched by exporting them as `preFetchOperations`:

```javascript
import useApolloQuery from '#src/composables/useApolloQuery';
import myQuery from '#src/graphql/query/myQuery.graphql';

const operation = { query: myQuery };

export const preFetchOperations = [operation];

export default function useMyThing() {
	const { result, loading, error } = useApolloQuery(operation);
	// derive state from result; loading distinguishes "not loaded yet" from real values
}
```

Component authors change nothing: they import the composable and call it in `setup()` as usual. The magic happens at build time: a vite transform (`build/composable-operations-plugin.js`) attaches the `preFetchOperations` of imported composables to each compiled component definition (composables that use other composables get the same treatment, so composition needs no authoring), and `preFetchAll` executes attached operations in the same pass as component `apollo` blocks. How it works, every option, and what the warnings mean are documented in `src/graphql/README.md` ("Composable Data Fetching") and AGENTS.md.

## Behavior changes

- A failed composable-operation prefetch fails the render, exactly like component operations.
- The "powered by N lenders" pill now renders in SSR output; it previously appeared at hydration.
- The shared variable builder (`operationVariables.js`) ships two fixes: falsy basket ids are omitted instead of sent as `undefined`, and contentful prefetches gain the preview flag in preview mode.
- Attached operations refetch on client navigations that activate their components, matching component-block semantics.
- `matching_highlight` (EXP-ACK-538-Mar2023) stops being tracked as shown on multi-matched loans, where its UI is suppressed anyway.

## The bug that forced the issue, AD-360

`useMultiMatching` was one of the raw-fetching composables. LendCta rendered from the unknown setting value and snapshotted it into one-shot data properties, so on a matched loan with no lenders the message area went permanently blank once the real value arrived.

`useMultiMatching` is the first composable migrated onto the mechanism (`multiMatchingResolved` is true once the read settles; a failed read resolves to the disabled default), and LendCta now derives its stats slot from pure helpers (`lendCtaStats.js`) instead of one-shot data properties: the pill hides when the cycle has nothing to show. The remaining raw-fetching composables will be migrated in sub-tasks of AD-385.

### Before
ad360-before

### After
ad360-after

## Verification

- On a fresh module graph the first dev render produces zero cache-miss warnings, and disabling the transform as a control produces exactly the two expected unattached-operation warnings, proving the detector works.
- HMR: removing a composable's `preFetchOperations` export produces warnings on the next render and restoring it heals, no restart.
- On the production build: a matched zero-lender loan renders no pill and no blank bar, a loan with lenders renders "powered by N lenders" in the SSR HTML, and the setting is present in the serialized apollo state on both.
- Stories cover both flag states (`MatchedNoLenders`, `MatchedNoLendersMultiMatchingOff`).


eddieferrer (approved):
Nice, thanks for catching that bug!

mcstover (approved):

dyersituations (approved):

PR #7133, opened 2026-08-04
feat: mp-3058-meta-tracking-new-events-pt-2

infante-jaime:
## MP-3012 — Meta tracking: new lender events (pt. 2, ui)

Adds four Meta custom events, following the pattern from pt. 1
(`idleLenderReEngaged` / `lapsedLenderReEngaged`). The ticket's two giving-fund
events live in cms-page-server and are handled separately.

| Event | Fires on |
|---|---|
| `donation` | Stand-alone donation — total > 0 with no loans and no Kiva Cards, so a tip on an order isn't counted. Covers checkout and instant donations. |
| `kivaCardRedemption` | A Kiva Card code applied successfully in checkout. |
| `emailSignUp` | Lender or guest opting in to marketing email. Gated on `lenderNews === true && !globalUnsubscribed`, since the same mutations serve opt-outs. |
| `accountCreated` | Newly registered lender landing on the site. |

All fire on confirmed success — after the mutation or transaction resolves, never at
form submit.

**New:** `src/util/metaEvents.js` (event names + `trackMetaEvent`, so delivery can
change in one place), `src/util/registrationTracking.js`.

**Modified:** `kv-analytics-plugin.js`, `InstantDonationThanks.vue`,
`KivaCardRedemption.vue`, `useOptIn.js`, `CheckoutPage.vue`, `client-entry.js`.


src/util/registrationTracking.js
```diff
@@ -0,0 +1,59 @@
+import { META_EVENTS, trackMetaEvent } from '#src/util/metaEvents';
```
mcstover: This file could also move to kv-analytics

infante-jaime: Makes sense—kv-analytics for some reason I was in denial that that made sense but let me refactor both PRs so that changes and definitions come from kv-analytics`

src/util/metaEvents.js
```diff
@@ -0,0 +1,30 @@
+import { trackFBCustomEvent } from '@kiva/kv-analytics';
+
+export const META_EVENTS = {
+	ACCOUNT_CREATED: 'accountCreated',
+	DONATION: 'donation',
+	EMAIL_SIGN_UP: 'emailSignUp',
+	KIVA_CARD_REDEMPTION: 'kivaCardRedemption',
+};
+
+export function trackMetaEvent(eventName, eventData = null) {
```
cristhianDt: I think this functionality already exists in other file or looks very similar 🤔, (thoughts) is possible centralize at least the constants in kv-elements? you know to share all the types for other FEs projects

src/util/registrationTracking.js
```diff
@@ -0,0 +1,59 @@
...
+ * `registration=new` is added by the monolith on the post-authentication redirect, in the same
+ * place that raises the "Welcome to Kiva!" toast, and covers email and social sign-ups.
+ * `claimed=1` is added by GuestAccountRedirect and covers guest account claims.
+ *
+ * @param {String} search location.search
+ * @returns {String|null} The param name that matched, so the caller can remove it
+ */
+export function getRegistrationMarker(search) {
+	const params = new URLSearchParams(search);
+	if (params.get('registration') === 'new') {
```
christian14b: I believe registration and claimed can be constants here

src/util/registrationTracking.js
```diff
@@ -0,0 +1,59 @@
...
+export function trackAccountCreated(userId) {
+	const marker = getRegistrationMarker(window.location.search);
+	if (!marker) {
+		return;
+	}
+
+	const sessionKey = `${SESSION_KEY}:${userId ?? ''}`;
+	let alreadyReported = false;
+	try {
+		alreadyReported = window.sessionStorage.getItem(sessionKey) === '1';
```
christian14b: Should we check if window is defined?

src/composables/useOptIn.js
```diff
@@ -29,7 +55,11 @@ export default (apollo, _cookieStore) => {
 					globalUnsubscribed,
 				},
 			});
-			return true;
+			const updated = settingsApplied(data?.my?.updateCommunicationSettings, errors);
+			if (updated) {
```
christian14b: Not a blocker but these 3 lines of code are repeated bellow. Does it make sense to add an extra procedure for this logic?

dyersituations:
@infante-jaime since some of the comments aren't auto-marked as "outdated", please go through the comments on this PR and respond + resolve.

src/util/metaEvents.js
infante-jaime: Done

src/util/registrationTracking.js
infante-jaime: added

src/util/registrationTracking.js
infante-jaime: done

src/composables/useOptIn.js
infante-jaime: fixed

src/composables/useOptIn.js
```diff
@@ -1,11 +1,35 @@
...
+	 * Opt-outs and unsubscribes run through the same mutations, so the values are checked
+	 * rather than assuming any successful call here is a sign-up.
+	 *
+	 * @param {Boolean|null|undefined} result The mutation's payload
+	 * @param {Array} errors
+	 * @param {Boolean} lenderNews
+	 * @param {Boolean} globalUnsubscribed
+	 * @returns {Boolean} Whether the settings applied
+	 */
+	const reportOutcome = (result, errors, lenderNews, globalUnsubscribed) => {
```
dyersituations, line 22: Is the `result` here only `true` if the user enabled?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1041,6 +1046,12 @@ export default {
 			const finalizeTransaction = () => {
 				this.$kvTrackTransaction(transactionData);
 
+				// Guest checkout collects the email preference up front, but the sign-up only
+				// exists once the transaction that carries it has gone through.
+				if (this.checkingOutAsGuest && this.userOptedIn) {
+					trackMetaEvent(META_EVENTS.EMAIL_SIGN_UP);
```
dyersituations, line 1052: Do we send emails to guests?

src/util/registrationTracking.js
```diff
@@ -0,0 +1,67 @@
...
+		window.sessionStorage.setItem(sessionKey, '1');
+	} catch (e) {
+		// Storage is unavailable in some private browsing modes. Removing the marker below
```
dyersituations: In this situation should we set `alreadyReported` to `true` here so we don't log constantly for users with private browsing. It would mean we don't log for these users, but maybe more accurate in the big picture?

src/util/registrationTracking.js
```diff
@@ -0,0 +1,67 @@
...
+		// still prevents the common repeat, a refresh.
+	}
+
+	if (!alreadyReported) {
+		trackMetaEvent(META_EVENTS.ACCOUNT_CREATED);
+	}
+
+	// Drop the marker so a reload cannot replay it
```
dyersituations: Why do we need to drop the marker if we are already setting whether we logged in session storage?

src/util/registrationTracking.js
dyersituations: An alternative here might be a session cookie since the user already needs cookies enabled to login?

src/util/registrationTracking.js
infante-jaime: The two checks cover different cases. Removing the URL parameter handles reloads and new tabs, while the per-tab flag handles cases where the monolith adds the parameter again after a redirect.

Also, since it's a one-time parameter, it shouldn't stay in the URL after it's been used.

src/util/registrationTracking.js
infante-jaime: I'd keep it firing. The URL marker is removed after it's used, so users should only generate one event. The main duplicate case is the rare 30-second re-auth flow.

A small amount of over-counting can be reconciled against our server-side registration counts. Missing the event is silent and permanent, and this feature already under-reports when the pixel hasn't loaded.

src/pages/Checkout/CheckoutPage.vue
infante-jaime: Yes — guest checkout shows a marketing-email checkbox and sends it to the API as emailOptIn on validateGuestPreCheckout, tied to a visitorId so we can email someone with no account yet.

There's a whole guest branch for it (visitorEmailOptIn, feature-flagged and activity-checked in the monolith), so an opted-in guest is a real sign-up.

src/composables/useOptIn.js
infante-jaime: result is the mutation's "did it save" flag — it returns true for a successful opt-out too.

So line 23 checks it saved, line 24 checks they actually opted in; the event needs both.

Without line 24 we'd fire emailSignUp every time someone unsubscribes.

src/util/registrationTracking.js
dyersituations: My concern is if other flows on the website are still reading the value for some reason after the initial page load.

src/util/registrationTracking.js
dyersituations: What about using the session cookie instead?

src/util/registrationTracking.js
infante-jaime: Checked this in depth, and currently no flow iis reading these parameters from window.location. The only consumer is claimed=1, and both places read it from the router's route object, so stripping it from the URL doesn't affect them.

The timing also works out. The strip happens after the analytics setup completes, so all created() hooks have already run. registration=new isn't used anywhere else.

The only caveat is if someone later adds code that reads these values from window.location after page load. If that's a concern, we could add a comment where the URL is stripped.

dyersituations: What happens if we don't clear this from the URL here? Does it resubmit the event if the user refreshes the page? Any other use cases where there would be issue?

src/util/registrationTracking.js
infante-jaime: yes cookie makes it neater, implemenmted

src/util/registrationTracking.js
infante-jaime: Good question—the cookie changes the reasoning here.

A refresh is already covered by the cookie. If we see the marker again, we check the cookie, see we've already reported it, and skip the event. Removing the parameter isn't protecting us from refreshes anymore.

The bigger issue is shared or stale URLs. The cookie only exists in the current browser. If someone copies a URL with ?registration=new, another user won't have that cookie and could incorrectly fire an account creation event just by opening the link.

Removing the parameter prevents that. It also avoids the case where someone reloads an old tab after the cookie has expired.

So I'd still keep it, but mainly to avoid fabricated events from shared or stale URLs rather than to prevent duplicates.

dyersituations: I'd recommend not stripping the URL for now, if the cookies solves the concerns here.

infante-jaime: Makes sense — dropped the URL stripping, the cookie covers it.

dyersituations (approved):

cristhianDt (commented):
👍 thanks for use my recommendations 🥳

PR #7137, opened 2026-08-04
fix: hide region card if no regions remain on next steps page

dyersituations:
https://kiva.atlassian.net/browse/MP-3094

Minor bug I encountered today while testing other features: hide the region card on MyKiva next steps if the user has lent to all of the regions, since we don't have an "all regions done" state of the card.


christian14b (approved):

cristhianDt (approved):

roger-in-kiva (approved):
👍

mcstover (approved):

PR #7138, opened 2026-08-05
fix: hide journal section for anonymized loans

dyersituations:
https://kiva.atlassian.net/browse/MP-3083

Haven't been able to track down a loan yet to repro the "empty journal body" issue for an anonymized loan, but this should be the simple fix regardless.

### DIFF
--- a/src/components/BorrowerProfile/FullBorrowerProfile.vue
+++ b/src/components/BorrowerProfile/FullBorrowerProfile.vue
@@ -86,7 +86,7 @@
 		</content-container>
 		<content-container>
 			<journal-updates
-				v-if="showUpdates"
+				v-if="showUpdatesSection"
 				data-testid="bp-updates"
 				class="tw-mb-5 md:tw-mb-6 lg:tw-mb-8"
 				:loan-id="loanId"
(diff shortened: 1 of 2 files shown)


roger-in-kiva (approved):

mcstover (approved):

kathrynlovett:
@dyersituations I'm not actually sure that we want to hide journal entries for anonymized loans 🤔 We may want to check with Kate on that before putting this in place.

dyersituations:
> @dyersituations I'm not actually sure that we want to hide journal entries for anonymized loans 🤔 We may want to check with Kate on that before putting this in place.

@kathrynlovett I went this route based on the comment here:
https://kiva.atlassian.net/browse/MP-3083?focusedCommentId=404548

Based on your reply in Jira, it sounds like there might be a second issue. If we're good with this fix for anonymous, I can investigate the other issues in a follow-up PR.

src/components/BorrowerProfile/FullBorrowerProfile.vue
```diff
@@ -368,6 +368,13 @@ export default {
 			// section, and only while the loan is not anonymized (full/pii).
 			return this.isPrivileged && !this.isAnonymized;
 		},
+		showUpdatesSection() {
+			// Hide the journal updates section entirely for anonymized (full/pii) loans:
```
kathrynlovett: [nit] I'm not sure how this did/should work for `full` anonymized loans. That was existing borrower privacy functionality, so may be worthwhile just to double check that this is the desired functionality. It would make sense to me, but just wanted to call it out as I'm less familiar with that functionality.

kathrynlovett (approved):
LGTM!

kathrynlovett:
> > @dyersituations I'm not actually sure that we want to hide journal entries for anonymized loans 🤔 We may want to check with Kate on that before putting this in place.
> 
> @kathrynlovett I went this route based on the comment here: https://kiva.atlassian.net/browse/MP-3083?focusedCommentId=404548
> 
> Based on your reply in Jira, it sounds like there might be a second issue. If we're good with this fix for anonymous, I can investigate the other issues in a follow-up PR.

Thank you! I completely forgot about that thread. This looks good to me!

src/components/BorrowerProfile/FullBorrowerProfile.vue
dyersituations: Good call, I didn't realize there was a difference between anonymization types for journal entries. Looks like we only clear for `pii`, so I pushed an update to check `this.loanData?.anonymizationLevel !== 'pii'` instead.

PR #7139, opened 2026-08-05
feat: [MP-3065] - goal in review analytics events

cristhianDt:
# MP-3065

## Summary

Add events for goal in review modal:

1. View the recap — view / goal-in-review (property screen-1…screen-7)
2. Back to Kiva — click / goal-recap-back-to-kiva
3. Finish my goal — click / goal-recap-finish-my-${year}-goal
4. Submit feedback survey — submit / goal-recap-submit-feedback
5. Set my next-year goal — click / set-a-goal (property from-goal-recap)

https://kiva.atlassian.net/browse/MP-3065

## Evidences

https://github.com/user-attachments/assets/f2e14088-43a0-48b4-8fb7-19ed9a869647


cristhianDt:
@dyersituations @christian14b I suggest leaving event #1 (view-goal-recap) for MP-3064, since the "View goal recap" button is implemented there — right now the recap only opens via the `?goTo=goal-recap` deep link, so there's no button here to attach the click to. adding it in that ticket keeps the event next to its button (a one-line v-kv-track-event directive) and lets us unblock/merge this ticket now with events #2–#6, which all have existing elements to hook into, 

they'll QA together once MP-3064 lands

agree?

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
```diff
@@ -4,7 +4,7 @@
 			class="-tw-mt-2"
 			:form-assembly-id="FEEDBACK_FORM_ID"
 			title="Goals feedback survey"
-			@fa-form-submitted="emit('submitted')"
+			@fa-form-submitted="handleSubmitted"
```
roger-in-kiva, line 7: Do we need to add new shared component?

dyersituations:
> @dyersituations @christian14b I suggest leaving event #1 (view-goal-recap) for MP-3064, since the "View goal recap" button is implemented there — right now the recap only opens via the `?goTo=goal-recap` deep link, so there's no button here to attach the click to. adding it in that ticket keeps the event next to its button (a one-line v-kv-track-event directive) and lets us unblock/merge this ticket now with events #2–#6, which all have existing elements to hook into,
> 
> they'll QA together once MP-3064 lands
> 
> agree?

@cristhianDt sure, please update those tickets as needed with the event info so we make sure the events get added and tested with the other ticket(s).

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
cristhianDt: at the moment yes, in case Lauren decides to add a new event to persist the values, otherwise we can remove it

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
```diff
@@ -15,4 +15,10 @@ import { KvFormAssemblyForm } from '@kiva/kv-components';
 const FEEDBACK_FORM_ID = 659;
 
 const emit = defineEmits(['submitted']);
+
+const handleSubmitted = (event = {}) => {
+	if (event.valid) {
```
dyersituations, line 20: Do we need to do anything if the event is not valid? It looks like a silent failure currently.

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
roger-in-kiva: So you must to update shared components package in this PR?

cristhianDt: yeah, I'm waiting for the new version there

src/composables/useGoalData.js
```diff
@@ -870,6 +870,9 @@ export default function useGoalData({ apollo } = {}) {
 
 	async function setGoalFeedbackSubmittedPreference(year = GOALS_CURRENT_YEAR) {
 		if (!year) return;
+		if (!userPreferences.value?.id) {
```
dyersituations: Are we guaranteed that we've tried to load the user preferences at this point? We don't want to create them if they were just not loaded yet.

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
cristhianDt: if we wish yes but currently, there's no functionality to include for this case, perhaps just a log indicating that an error occurred?

dyersituations: We can leave as is for now then and see what happens live.

src/composables/useGoalData.js
roger-in-kiva: If this is for a goal in review only, my guess is that we already have user preferences set.
Wondering why we want to create them here.

cristhianDt: yes, there are, in some point but since I forced the modal to display without any preferences or goals, the process fails; this patch is more for an edge case

dyersituations: Is it a common edge case? How do you repro? I'm just concerned around forcing a creation, it will probably fail if a user already has preferences, right?

cristhianDt: opening the modal by the param in the url "goTo...", I erased my preferences first. well, if we load the preferences in some point (I think we already do that) calling "loadPrefences" method in this file, we are fine, because the ref property is filled out first. whoever integrates the modal must ensure that we are loading the preferences by the composable before show the modal

cristhianDt: cc @christian14b

dyersituations: Sounds like we can revert the create here then?

cristhianDt: I think so

cristhianDt: before that, do you want to check locally?, otherwise the submit will fail in our end (I mean, updating the preferences)

dyersituations: Verify what specifically? If a user has user preferences (an existing goal) we should never need to create.

roger-in-kiva: Probably part of functional testing for this is to ensure your account already have user preferences set.

cristhianDt: the changes with form and handling the result in our end, but I already did it so ... removing

cristhianDt: deleted

dyersituations (approved):

roger-in-kiva (approved):

src/components/MyKiva/GoalInReview/GoalInReviewFeedbackForm.vue
cristhianDt: finally done @roger-in-kiva

dyersituations (approved):

PR #7140, opened 2026-08-05
fix: hide why special when null

dyersituations:
https://kiva.atlassian.net/browse/MP-3065

Digging into the empty journal entry issue identified that we're actually also seeing an empty "why special" situation, as shown below. This hides the why special when it's empty.

image


eddieferrer (approved):

emuvente (approved):
Nice!

roger-in-kiva (approved):

christian14b (approved):

cristhianDt (approved):

PR #7141, opened 2026-08-06
feat: integrate goal in review modal with data

christian14b:
Goal in modal review integration 
https://kiva.atlassian.net/browse/MP-3066

https://github.com/user-attachments/assets/1af1bf4a-cf00-4057-8c2f-db33f22994c9

## Test instructions

1. Complete a goal (Support all and any other goal category)
2. Locally, go to `https://kiva-ui.local/mykiva?goTo=goal-recap&goal_in_review=true` (`goal-recap` is the trigger that shows the modal and `goal_in_review` enables the feature to get data. These are only for testing)
3. Validate data regarding your goal


src/composables/useGoalInReview.js
```diff
@@ -71,8 +109,34 @@ export default function useGoalInReview() {
...
+			const [summary, achievementsData, lenderData, contentfulData] = await Promise.all([
+				loadGoalSummary(),
+				query(goalInReviewAchievementsQuery, { year }),
+				query(goalInReviewLenderQuery),
+				query(contentfulEntriesQuery, { contentType: 'challenge', limit: 200 }, 'cache-first'),
+			]);
+
+			// The monolith only computes my.goalSummary for support-all.
+			const extras = summary?.category === ID_SUPPORT_ALL
+				? await query(goalInReviewSummaryQuery, { goalName: summary.goalName })
```
infante-jaime: getGoalSummary() already ran this same my.goalSummary resolver inside useGoalData. Should it be enough just to have a single fetch.

src/graphql/query/goalInReviewAchievements.graphql
```diff
@@ -0,0 +1,42 @@
+query goalInReviewAchievements($year: Int, $loanPurchasesLimit: Int = 1000) {
```
dyersituations: I don't know if loading `1000` entries here is the most efficient option. How long is this query currently? Would it be worth instead building a new query in the achievements service?

src/graphql/query/goalInReviewSummary.graphql
```diff
@@ -0,0 +1,42 @@
...
+        id
+        name
+        isoCode
+        fundsLentInCountry
+        geocode {
+          latitude
+          longitude
+        }
+      }
+      loans(limit: $loansLimit) {
```
dyersituations: Do we need to load the `loans` here if we're already loading them in `goalInReviewAchievements` or visa versa?

src/util/goalInReviewMap.js
```diff
@@ -0,0 +1,38 @@
+/** Centre used when no country has usable coordinates. */
```
dyersituations, line 1: Just out of curiosity, what happens with the map without the coordinates adjustments in these utils?

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -188,7 +188,7 @@
 		@sidesheet-closed="handleComponentClosed"
 	/>
 	<GoalInReviewModal
-		v-if="showGoalInReviewModal"
+		v-if="goalInReviewEnable && showGoalInReviewModal"
```
dyersituations, line 191: The PR description mentions the `goal_in_review` query string, but I'm not seeing it in the code. Do we need that query string currently? It seems like the `goTo=goal-recap` might be sufficient for testing, since users won't be able to guess that query string.

christian14b: Thanks! makes sense, I'll double check

src/util/goalInReviewMap.js
christian14b: It only shows Africa (20, 10 base points), so if you have 1 loan in USA, it won't be visible.. I realized that with one loan in Peru

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Yeah, this was merely for testing, let me try another thing and go back to you. This might be no necessary, and rather we can use other approaches (a new query might be good though)

src/graphql/query/goalInReviewSummary.graphql
christian14b: Nice catch! there were some unnecessary duplications here... currently, both queries are being called and that's wrong. The idea is that `goalSummary` is called for support-all and we call `goalInReviewAchievements` for the other categories because goalSummary will skip that logic in the monolith. That's why they need to be separate

dyersituations: So we keep the queries but ensure we only call one depending on which category?

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Ok so the situation is that goal in review selects the loan purchases *oldest to first*, so that later lending cannot displace the loans that actually met the goal. `loanPurchases` retains the *most recent* loans, so we have to pull the year and sort client-side.

### 1. New query

A field on the achievement service returning the loans that count toward the goal for a given year, sorted oldest-first server-side, with a limit (goal target). It can be something like:

```graphql
goalLoans(year: Int!, limit: Int = 25)
```

That would make this safer instead of depending on client-side sorting. I believe this will be sufficient.. What do you think?

dyersituations: Could we instead show the most recent loans on the goal summary? Would that allow us to have a lower limit? Or is there specific functionality that this would break?

src/graphql/query/goalInReviewSummary.graphql
christian14b: Yes, exactly! Although, `goalInReviewSummary` will be gone since `getGoalSummary` already have the query, I just need to add the necessary information for the map and chart

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Yes, we can show the most recent loans and reduce the lower limit, but the tradeoff is that the information will change depending on how many loans you "contributed" to your goal (achievement) since the first time you saw the goal recap.

Since the user can actually see the recap again (`If in 2026 (current goal year) and the goal is complete, user can view it from the Goal card in the Impact Progress row for the 2026 /current year goal`). It might be weird, but it won't break anything related to the content

dyersituations: I think that's probably fine, as long as the stats are consistent. It would be surprising is users notice that the featured borrowers change, if that's the change you're referring to is.

src/graphql/query/goalInReviewAchievements.graphql
```diff
@@ -1,4 +1,6 @@
-query goalInReviewAchievements($year: Int, $loanPurchasesLimit: Int = 1000) {
+# Only the first `count` loans are used, but they are selected oldest-first, so the
+# year has to be pulled and sorted locally. 200 covers all but the heaviest lenders.
```
dyersituations: What happens for users who had goals over `200`?

src/composables/useGoalInReview.js
dyersituations: @christian14b is this comment covered in recent changes?

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Problem is that it isn't only the borrower faces that will change. `Total Lent`, the map and the chart are all derived from:

```
amount:    getAmountLent(countedLoans),
countries: getGoalCountries(countedLoans),
sectors:   getGoalSectors(countedLoans),
```

So re-selecting on each open moves the dollar figure and the country list too. Stats stay internally consistent because they always describe exactly the loans shown, but they aren't stable across different openings. Only count and percent are, since those come from `progressForYear`.

Just to make sure the situation can be the following: user sets a goal: 2 loans toward women's equality.

**March** you lend to A (Peru, $25) and B (USA, $25). Goal complete. The recap
shows:

`2 borrowers · $50 total lent · Peru, USA`

**May**: you lend to two more women, Rosa (Kenya, $100) and Mei (Mexico, $75). You reopen the same
recap:

`2 borrowers · $175 total lent · Kenya, Mexico `

Same goal, same "100% complete", but every sections has changed

dyersituations: I feel like the main difference users might notice is the amount, but I'm think this might be something we release the first version with. Loading 1000 loans is a bit heavy handed I think, and the only other approach might be a new query in achievements service that returns the aggregated data.

What are your thoughts on further options?

src/composables/useGoalInReview.js
christian14b: Yes, we no longer call this query, we're using `getGoalSummary` now

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Yeah, this is wrong.. and won't be necessary with the conversation above

src/graphql/query/goalInReviewAchievements.graphql
christian14b: Yes, we can try the most recent ones in the first run and then change it with a new query since the only change will be in the BE side, client will remain the same.

I think using the most recent loans makes sense for now too, it's hard to know the real number of the loan purchases limit, but I think it's easier to wait or work in parallel for a new query.

In the long term, we could use a "snapshot"... so when the goal is completed, then we save all this information and stored per year, so when a user wants to see the recap again, we only get this specific record to populate the modal.

Should we write some of these ideas somewhere?

dyersituations: Yes please create a follow-up ticket with why this is an issue, what type of users/goals will see it, and possible solutions (that we can review with Lauren).

dyersituations (approved):

src/components/MyKiva/GoalInReview/GoalInReviewSlide1.vue
```diff
@@ -26,7 +26,7 @@
 			>
 		</div>
 
-		<div class="tw-px-2 tw-pt-3 tw-mx-auto tw-max-w-3xl tw-text-center">
```
cristhianDt, line 29: dont we need the padding-top anymore?

src/composables/useGoalInReview.js
```diff
@@ -51,14 +65,37 @@ export function getGoalInReviewCurrentYear() {
...
-export default function useGoalInReview() {
+export default function useGoalInReview({ apollo } = {}) {
+	const apolloClient = apollo || inject('apollo');
+	const { getGoalSummary } = useGoalData({ apollo: apolloClient });
 	const loading = ref(false);
 	const goalInReviewData = ref(null);
 
 	const isEligible = computed(() => Boolean(goalInReviewData.value?.isEligible));
 
+	async function query(recapQuery, variables, fetchPolicy = 'network-only') {
```
cristhianDt, line 80: I like this wrapper but maybe another name would be better what do you think?

src/util/goalInReviewMap.js
cristhianDt: interesting nice catch chris, maybe we could add a new story for this case 🤔

src/util/goalInReviewPayload.js
```diff
@@ -0,0 +1,203 @@
+import { getContentfulLevelData, ID_SUPPORT_ALL } from '#src/composables/useBadgeData';
```
cristhianDt, line 1: I think we should or merge all the utils related with goalInReview or create a new folder in utils/ and move our files there

src/components/MyKiva/GoalInReview/GoalInReviewSlide1.vue
christian14b: Needed so the whole content fits in the modal height, Also there's a whitespace that helps

src/util/goalInReviewPayload.js
christian14b: What do you mean? This all methods exist in a new file

cristhianDt: I mean, there multiple utils files related with goal in review (sectors, map, others..) so my idea is merge/join all of them or move them to a new folder "goalInReview" for example into "util"

cristhianDt (approved):
LGTM just some minor suggestions

infante-jaime (approved):
👍

src/util/goalInReviewMap.js
```diff
@@ -0,0 +1,38 @@
...
+export const DEFAULT_MAP_CENTER = { lat: 20, long: 10 };
+
+const toRadians = degrees => (degrees * Math.PI) / 180;
+const toDegrees = radians => (radians * 180) / Math.PI;
+
+// 0 is a real latitude and longitude, so coordinates need a finite check rather
+// than a truthiness one.
+const getPoints = countries => (countries ?? [])
+	.map(country => ({ lat: country?.geocode?.latitude, long: country?.geocode?.longitude }))
```
roger-in-kiva, line 10: should we use a fallback if lat and log are not defined for country?

roger-in-kiva: Just wondering based on this approach:
https://github.com/kiva/ui/blob/2dedfd55b27d9627037aa40c1befcfda410c1987/src/components/BorrowerProfile/BorrowerCountry.vue#L83

christian14b: Thanks for the suggestion! There's already a fallback but at map suggestion level, we don't really need to be that specific

roger-in-kiva (approved):

src/util/goalInReviewPayload.js
christian14b: I'll do it as a follow-up in the next ticket since it's possible these files are going to change

PR #7142, opened 2026-08-06
fix: prevent infinite UserPreferences refetch loop on MyKiva page

jmallatt-kiva:
The UserPreferences watchQuery in mounted() had no error handling and
refetched on every cache broadcast. With errorPolicy 'all' (the client
default), a resolver returning partial data (200 + errors) writes an
incomplete result to the cache, so the watched query can never be
satisfied and Apollo reobserves/refetches in a tight loop (~14 POST/s
per tab against Stellate, observed during the 2026-08-06 maintenance
window).
Watch the cache only (cache-only watchers can never trigger network
requests), populate it with a single one-shot query, log watcher
errors, and unsubscribe on unmount.

### DIFF
--- a/src/pages/MyKiva/MyKivaPage.vue
+++ b/src/pages/MyKiva/MyKivaPage.vue
@@ -391,23 +391,36 @@ export default {
 				this.getGoalSummary();
 			}
 
-			this.apollo.watchQuery({
-				query: gql`
-					query UserPreferences {
-						my {
+			const userPreferencesQuery = gql`
+				query UserPreferences {
+					my {
+						id
+						userPreferences {
+							preferences
 							id
-							userPreferences {
-								preferences
-								id
-							}
 						}
 					}
-				`,
+				}
+			`;
+
+			// Watch the cache only; a watcher with a network fetch policy refetches on
+			// every cache broadcast, which loops indefinitely (hammering the gateway)
+			// if the resolver keeps returning partial data (200 + errors).
+			this.userPreferencesSubscription = this.apollo.watchQuery({
+				query: userPreferencesQuery,
+				fetchPolicy: 'cache-only',
 			}).subscribe({
 				next: ({ data }) => {
 					this.userInfo = { ...this.userInfo, userPreferences: data?.my?.userPreferences };
 				},
+				error: error => {
+					logReadQueryError(error, 'MyKivaPage userPreferences cache watcher');
+				},
 			});
+
+			// Single network fetch to populate the cache; the watcher above receives
+			// this result and any later userPreferences mutation writes.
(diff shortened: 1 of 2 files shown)


src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -391,24 +391,37 @@ export default {
 				this.getGoalSummary();
 			}
 
-			this.apollo.watchQuery({
-				query: gql`
-					query UserPreferences {
-						my {
+			const userPreferencesQuery = gql`
```
dyersituations, line 394: Is this resolved by @mcstover's open PR? https://github.com/kiva/ui/pull/7143

src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -391,24 +391,37 @@ export default {
...
 						}
 					}
-				`,
+				}
+			`;
+
+			// Watch the cache only; a watcher with a network fetch policy refetches on
+			// every cache broadcast, which loops indefinitely (hammering the gateway)
+			// if the resolver keeps returning partial data (200 + errors).
+			this.userPreferencesSubscription = this.apollo.watchQuery({
```
dyersituations, line 409: `userPreferencesSubscription` needs to be added to `data()`

src/pages/MyKiva/MyKivaPage.vue
```diff
@@ -391,24 +391,37 @@ export default {
...
 					this.userInfo = { ...this.userInfo, userPreferences: data?.my?.userPreferences };
 				},
+				error: error => {
+					logReadQueryError(error, 'MyKivaPage userPreferences cache watcher');
+				},
 			});
 
+			// Single network fetch to populate the cache; the watcher above receives
+			// this result and any later userPreferences mutation writes.
+			await this.apollo.query({ query: userPreferencesQuery });
```
dyersituations, line 423: Doesn't the `watchQuery` above also trigger an initial query?

mcstover: Yea it should. In general this was part of several of us debugging with claude. I think in this case, it really depends on what and when the cache might get updated. I though I recalled you had some other in-page actions that could update user preferences. If so, those would need to force a network call or otherwise deliberately update the cache with the new data for this approach to work.

src/pages/MyKiva/MyKivaPage.vue
mcstover: Both my and userPreferences have an id field so they were already properly mergable. My pr is a guard against a list of fields that don't have ids that need to be intentionally set to mergable.

mcstover: I think this just makes it so the query defintion can be reused below

PR #7143, opened 2026-08-06
fix(apollo): guard against caching an empty GraphQL possibleTypes payload

mcstover:
## Problem

My Kiva was firing thousands of GraphQL requests per page load with no user interaction — `kcBasicLoanCard`, `bpHeroBackgroundCountry`, and `bpHeroBackgroundImage`, cycling through every loan on the page and repeating.

Root cause is in `server/util/getGqlPossibleTypes.js`. When schema introspection failed, it built `{ Mergable: [] }` and treated that as a success — writing it to memcache for 24 hours and pinning it to `process.env.FETCHED_GQL_TYPES` for the process lifetime. Every browser served by those pods then ran an Apollo cache with no type information. Confirmed in production: `window.__KV_CONFIG__.graphqlPossibleTypes.Mergable.length` returned `0`.

An empty payload costs the cache two things at once:

1. **No `Mergable` list** → the `Mergable: { merge: true }` policy matches nothing, so types with no `id` are *replaced* rather than merged on write. `Lend` has no `id`, so writing one loan evicts every other loan's entry from `ROOT_QUERY.lend`.
2. **No interface map** → `fragment loanCardFields on LoanBasic` can't be matched against a cached `LoanPartner`.

Either one leaves the cache read permanently incomplete. `cache-first` then goes to the network, that response evicts the next card's data, and you get an unbounded round robin.

## Fix

- **Validate before caching or returning.** A schema with hundreds of id-less types can never legitimately produce an empty `Mergable` list, so that's a reliable failure signal.
- **Never persist an invalid payload** to memcache or `process.env` — the next request retries instead of pinning every pod to a broken cache.
- **Validate on read too**, so entries poisoned before this change heal on the next request rather than persisting for the full TTL. **No manual memcache flush needed once this ships.**
- **Log introspection errors.** `fetchSchema` was discarding the `errors` array before `?? {}` swallowed it.

Deliberately keeps serving a degraded payload rather than throwing, so a failed introspection doesn't take the site down with a 500.

## Tests

`test/unit/specs/server/util/getGqlPossibleTypes.spec.js` (35 tests) — `isValidPossibleTypes` directly, plus five failure shapes (missing schema, GraphQL error response, empty body, no types, all types have an `id`), each asserting the payload isn't cached, isn't pinned to the process, is logged, and is re-fetched next request. Plus cache read paths: valid value reused, invalid/unparseable discarded and re-fetched, process-level value honoured or cleared.

Verified these fail against broken code — with only the two guard blocks removed, 20 of them fail.

`test/unit/specs/api/possibleTypesCacheIntegrity.spec.js` (3 tests) — replicates the meltdown against a real `InMemoryCache` with six loan cards and the `LoanBasic` interface fragment:

| `possibleTypes` | requests in 250ms |
|---|---|
| populated | 6 — one per card, flat |
| `{ Mergable: [] }` | hundreds, unbounded |

Full suite: 316 files, 4701 passed, 6 skipped. ESLint clean.

## Deploy note

This heals the cache on deploy. Until then the poisoned memcache entry plus `process.env.FETCHED_GQL_TYPES` keeps serving the empty payload, so **flush `ui-gql-possible-types` and restart the pods** for immediate relief.

## Follow-up (not in this PR)

These guards stop the failure being *sticky*, but a request served during an introspection failure still ships an empty payload to that browser and that session loops. Generating `possibleTypes` at build time from the checked-in `build/schema.graphql` would close that window entirely — worth doing, especially if introspection is intentionally restricted in prod.

### DIFF
--- a/server/util/getGqlPossibleTypes.js
+++ b/server/util/getGqlPossibleTypes.js
@@ -51,6 +51,16 @@ async function fetchSchema(url) {
 	if (data?.errors?.length) {
 		error('GraphQL errors while introspecting possible types', { errors: data.errors });
 	}
+	// The gateway answers an outage with a plain `{ code, message }` body rather than a GraphQL
+	// envelope, e.g. `{ "code": "503", "message": "Service Unavailable" }`. That parses as JSON
+	// too, so name the failure instead of reporting it as an empty schema further down.
+	if (!data?.data && (data?.code || data?.message)) {
+		error('Non-GraphQL error response while introspecting possible types', {
+			status: result.status,
+			code: data.code,
+			message: data.message,
+		});
+	}
 	// eslint-disable-next-line no-underscore-dangle
 	return data?.data?.__schema ?? {};
 }
--- a/server/util/getGqlPossibleTypes.js
+++ b/server/util/getGqlPossibleTypes.js
@@ -1,4 +1,5 @@
 import fetch from './fetch.js';
+import { error } from './log.js';
 import { getFromCache, setToCache } from './memJsUtils.js';
 
 const GQL_BUILT_IN_TYPES = [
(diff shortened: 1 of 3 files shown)


dyersituations (approved):
The fix makes sense, but what specifically caused the issue on MyKiva?

emuvente (approved):

server/util/getGqlPossibleTypes.js
```diff
@@ -21,17 +46,18 @@ async function fetchSchema(url) {
 		}),
 	});
 	const data = await result.json();
+	// A GraphQL error response still parses as JSON, so it would otherwise be swallowed by the
+	// `?? {}` below and cached as an empty result.
+	if (data?.errors?.length) {
```
JoeTravisKiva, line 51: We may also want to add a check for a 2-300 level status code. I suspect the request hit a 503 with a body of 
```json
{
	"code": "503",
	"message": "Service Unavailable"
}
```
Which will parse as json just fine, but won't trigger the error condition here.

dyersituations (approved):

JoeTravisKiva (approved):

PR #7148, opened 2026-08-07
feat: deploy seen-once logic for goal in review modal in my kiva and …

christian14b:
…portfolio pages

Opens the Goal in Review modal, once per user in total across both My Kiva and Portfolio pages.

## What changed

1. **A seen-once flag the feature owns.** `goalRecapViewed[year]` is written to user preferences when the recap opens, so opening and dismissing both count as seen. It is server side, so it holds across pages, sessions and devices (a cookie would not).

2. **A pending marker for the "session after completion" rule.** `goalRecapPending[year]` is set the first time a completed goal is seen, and the recap opens on the visit after that.

3. ** Both pages ask the same question.** Every rule about when the recap should appear lives in one place, `shouldAutoOpenRecap`. It takes the things it needs as arguments and answers yes or no, so each rule can be tested on its own without standing up a page.

4. **Portfolio now renders the modal.** It computed `goalInReviewEnable` but never had `GoalInReviewModal` mounted.

## Trigger rules

| Goal state | Behaviour |
|---|---|
| Completed, first sighting | nothing shown, arms `goalRecapPending[year]` |
| Completed, already pending | opens, marks `goalRecapViewed[year]` |
| In progress | opens on the first visit once the flag is on |
| Already seen | never again, on either page |
| Previous year or expired | never |
| No goal, or no loans toward it | never |
| Feature flag off | never |

On MyKiva a `?goTo=goal-recap` deep link returns early and skips the auto-open, so an explicit link does not silently consume the once-only flag. The deep link also does not set the flag, so a link in an email keeps working however many times it is clicked.

---

# Test instructions - VQA

You need:

* `goal_in_review_enable` switched on.
* A logged in lender with a goal for the current year and at least one loan toward it.

Use `/mykiva` or `/portfolio` with **no query parameters**. A `?goTo=goal-recap` parameter skips the auto-open path deliberately.

## Check your starting state

Run this in the gateway while logged in as the test account:

```graphql
query { my { id userPreferences { id preferences } } }
```

Look inside the `preferences` JSON for `goalRecapViewed` and `goalRecapPending`. They determine what happens next, and they are stored on the account (clearing cookies or changing browser will not reset them).

### In progress goal

Needs `goal_in_review_in_progress_start` in the past, or use `?recapDate` to jump past it.

1. No flags set. Visit `/mykiva`. The recap opens.
2. Reload. Nothing. `goalRecapViewed[year]` is now set.
3. Visit `/portfolio`. Nothing. This is the "once in total, not once per page" criterion.
4. Move the setting into the future instead. Nothing, on either page. That is the mid-November gate.

### Completed goal, the session rule

This is the one worth walking end to end, because it spans checkout.

1. With a goal one loan short, check out the loan that completes it. The thanks page shows the completion.
2. Go straight to `/mykiva`. **Nothing.** You are still in the session you completed in. Check that the `kvgrcs` cookie is set to the goal year.
3. Reload, then visit `/portfolio`. Still nothing, for the same reason.
4. End the session, using one of the methods below, and return to `/mykiva`. The recap opens.
5. Dismiss it and reload. Nothing. `goalRecapViewed[year]` is now set.

### How to end the session properly

Closing the tab does **not** work. `kvgrcs` is a session cookie, and a browser session means the whole browser, not one tab. It survives as long as any window of that browser is open.

Quitting the browser usually works, but not always. Chrome's "Continue where you left off" startup setting and Firefox's session restore both bring session cookies back across a full quit. If you have either one enabled, quitting will make the feature look broken.

Two reliable options:

1. **Delete the cookie by hand.** DevTools, then Application, then Cookies, then delete `kvgrcs`. Reload the page. This is the fastest and the one to use for most passes.
2. **Use a private window.** Close every private window to end that session for real.

Worth knowing beyond testing: a lender who never quits their browser stays in the completion session indefinitely and will not get the pop-up. If that is a concern, the alternative is a cookie with a short expiry of a few hours instead of a session cookie. That is looser than "next session", but it always clears itself. Only the way `completedThisSession` is calculated would change.

### Completed goal, already completed in an earlier session

A lender who completed before the feature shipped has no `kvgrcs` cookie, so their first visit opens the recap straight away. Test this by clearing the cookie with a completed goal and no `goalRecapViewed` set.

### Portfolio on its own, without visiting MyKiva

Worth running as its own pass, since a lender who never opens MyKiva must still get the recap.

1. Completed goal, no `goalRecapViewed`, no `kvgrcs` cookie. Visit `/portfolio`. The recap opens.
2. Visit `/mykiva`. Nothing.

### Dismissal counts as seen

Open the recap, then close it with the X or by clicking outside it. Reload. It does not come back. The flag is written when it opens, not when it is read.

### Cases that must never pop

* A goal from a previous year, or an expired goal.
* A lender with no goal, or a goal with no loans toward it.
* Anyone, with `goal_in_review_enable` off.

## Resetting between runs

The flags are written to the account, so each scenario is one-shot. To run them again, clear `goalRecapViewed` and `goalRecapPending` from that user's preferences.

### DIFF
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -852,6 +852,55 @@ export default function useGoalData({ apollo } = {}) {
 		);
 	}
 
+	const goalRecapViewedByYear = computed(() => {
+		const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
+		return parsedPrefs.goalRecapViewed || {};
+	});
+
+	function hasViewedGoalRecapForYear(year) {
+		return Boolean(goalRecapViewedByYear.value?.[year]);
+	}
+
+	async function setGoalRecapViewedPreference(year = GOALS_CURRENT_YEAR) {
+		if (!year) return;
+		const parsedPrefs = await loadPreferences('network-only');
+		const prev = parsedPrefs?.goalRecapViewed || {};
+		// Year-keyed flag so seeing this year's recap does not suppress next year's.
+		if (prev[year]) return;
+		const updatedPreference = { goalRecapViewed: { ...prev, [year]: true } };
+		await updateUserPreferences(
+			apolloClient,
+			userPreferences.value,
+			parsedPrefs,
+			updatedPreference
+		);
+	}
+
+	const goalRecapPendingByYear = computed(() => {
+		const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
+		return parsedPrefs.goalRecapPending || {};
+	});
+
+	function hasGoalRecapPendingForYear(year) {
+		return Boolean(goalRecapPendingByYear.value?.[year]);
+	}
+
+	// Set the first time a completed goal is seen, so the recap can open on the session
+	// after completion without depending on the goal card's own celebration flag.
+	async function setGoalRecapPendingPreference(year = GOALS_CURRENT_YEAR) {
+		if (!year) return;
+		const parsedPrefs = await loadPreferences('network-only');
+		const prev = parsedPrefs?.goalRecapPending || {};
+		if (prev[year]) return;
+		const updatedPreference = { goalRecapPending: { ...prev, [year]: true } };
+		await updateUserPreferences(
+			apolloClient,
+			userPreferences.value,
+			parsedPrefs,
+			updatedPreference
+		);
+	}
+
 	const goalFeedbackSubmittedByYear = computed(() => {
 		const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
 		return parsedPrefs.goalFeedbackSubmitted || {};
@@ -1291,6 +1340,12 @@ export default function useGoalData({ apollo } = {}) {
 		setViewedGoalCompletePreference,
 		goalFeedbackSubmittedByYear,
 		hasSubmittedGoalFeedbackForYear,
+		goalRecapViewedByYear,
+		hasViewedGoalRecapForYear,
+		setGoalRecapViewedPreference,
+		goalRecapPendingByYear,
+		hasGoalRecapPendingForYear,
+		setGoalRecapPendingPreference,
 		setGoalFeedbackSubmittedPreference,
 		getSupportAllLoanCountByYear,
 		setGoalState,
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -819,14 +819,44 @@ export default function useGoalData({ apollo } = {}) {
 		// MyKiva renders the completed card once, then hides it on the next page load.
 	}
 
-	const viewedGoalCompleteByYear = computed(() => {
-		const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
-		return parsedPrefs.viewedGoalComplete || {};
-	});
+	/**
+	 * Builds the reader/writer pair for a preference stored as `{ [year]: true }`,
+	 * so a flag set for one year never applies to another.
+	 *
+	 * @param {string} preferenceKey Key this flag is stored under in user preferences
+	 * @returns {object} The by-year map, a per-year reader, and a per-year setter
+	 */
+	function yearKeyedPreference(preferenceKey) {
+		const byYear = computed(() => {
+			const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
+			return parsedPrefs[preferenceKey] || {};
+		});
 
-	function hasViewedCompletedGoalForYear(year) {
-		return Boolean(viewedGoalCompleteByYear.value?.[year]);
+		function hasForYear(year) {
+			return Boolean(byYear.value?.[year]);
+		}
+
+		async function setForYear(year = GOALS_CURRENT_YEAR) {
+			if (!year) return;
+			const parsedPrefs = await loadPreferences('network-only');
+			const prev = parsedPrefs?.[preferenceKey] || {};
+			if (prev[year]) return;
+			const updatedPreference = { [preferenceKey]: { ...prev, [year]: true } };
+			await updateUserPreferences(
+				apolloClient,
+				userPreferences.value,
+				parsedPrefs,
+				updatedPreference
+			);
+		}
+
+		return { byYear, hasForYear, setForYear };
 	}
+
+	const viewedGoalComplete = yearKeyedPreference('viewedGoalComplete');
+	const goalRecapViewed = yearKeyedPreference('goalRecapViewed');
+	const goalRecapPending = yearKeyedPreference('goalRecapPending');
+	const goalFeedbackSubmitted = yearKeyedPreference('goalFeedbackSubmitted');
 
 	/**
 	 * Drives the basket / ATB-modal achievement nudges so they stay out of the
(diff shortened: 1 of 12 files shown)


src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -103,6 +115,19 @@ export default {
 		LoanCards,
 		GoalEntrypoint
 	},
+	setup() {
+		const apollo = inject('apollo');
```
roger-in-kiva: Is there a way to avoid injecting apollo twice?

christian14b: Interesting question, I'll take a look

roger-in-kiva: Also, `useGoalInReview({ apollo })` spins up a `useGoalData` internally, and setup creates a second one directly. Since `userPreferences` is a per-call ref(null) rather than module-scoped, the instance `loadAutoOpenRecap` writes `goalRecapViewed/goalRecapPending` into is a different reactive store from the one the page reads feedback state from. It works today only because everything reloads network-only, but it's a duplicate round-trip and a divergence trap if anything later leans on shared reactive prefs. Could we thread one `useGoalData` through, or have `useGoalInReview` re-expose the feedback helpers the page needs? (MyKiva has the same shape.)

src/composables/useGoalData.js
```diff
@@ -852,6 +852,55 @@ export default function useGoalData({ apollo } = {}) {
...
+
+	function hasViewedGoalRecapForYear(year) {
+		return Boolean(goalRecapViewedByYear.value?.[year]);
+	}
+
+	async function setGoalRecapViewedPreference(year = GOALS_CURRENT_YEAR) {
+		if (!year) return;
+		const parsedPrefs = await loadPreferences('network-only');
+		const prev = parsedPrefs?.goalRecapViewed || {};
+		// Year-keyed flag so seeing this year's recap does not suppress next year's.
```
dyersituations: I don't know if this comment is accurate.

src/composables/useGoalData.js
```diff
@@ -852,6 +852,55 @@ export default function useGoalData({ apollo } = {}) {
...
+			apolloClient,
+			userPreferences.value,
+			parsedPrefs,
+			updatedPreference
+		);
+	}
+
+	const goalRecapPendingByYear = computed(() => {
+		const parsedPrefs = JSON.parse(userPreferences.value?.preferences || '{}');
+		return parsedPrefs.goalRecapPending || {};
```
dyersituations: There are two methods here that are really similar and could be combined and passed a key.

src/composables/useGoalData.js
```diff
@@ -852,6 +852,55 @@ export default function useGoalData({ apollo } = {}) {
...
+	});
+
+	function hasGoalRecapPendingForYear(year) {
+		return Boolean(goalRecapPendingByYear.value?.[year]);
```
dyersituations: Similarly this method could be combined/abstracted.

src/composables/useGoalData.js
```diff
@@ -852,6 +852,55 @@ export default function useGoalData({ apollo } = {}) {
...
+	}
+
+	// Set the first time a completed goal is seen, so the recap can open on the session
+	// after completion without depending on the goal card's own celebration flag.
+	async function setGoalRecapPendingPreference(year = GOALS_CURRENT_YEAR) {
```
dyersituations: Another method that is almost duplicated above.

src/util/goalInReviewTrigger.js
```diff
@@ -0,0 +1,50 @@
...
+
+	if (goalStatus === GOAL_STATUS.COMPLETED) {
+		// The session after completion: the completed goal was already seen on an earlier visit.
+		return hasCompletionPending;
+	}
+
+	// In progress goal setters are reached when the feature itself goes live, so the
+	// flag is the only gate.
+	if (goalStatus === GOAL_STATUS.IN_PROGRESS) {
+		return true;
```
dyersituations: This means any in-progress goal is shown in the new modal after the feature is enabled?

src/pages/MyKiva/MyKivaPageContent.vue
```diff
@@ -720,6 +722,17 @@ export default {
...
 		},
+		// Called from mounted: the decision reads user preferences and writes one back,
+		// so it must not run during server render.
+		async openGoalRecapIfDue() {
+			const goalInReview = await this.loadAutoOpenRecap({ enabled: this.goalInReviewEnable });
+			if (!goalInReview) {
+				return;
+			}
+			await this.loadGoalPreferences('network-only');
+			this.goalInReviewFeedbackSubmitted = this.hasSubmittedGoalFeedbackForYear(goalInReview.year);
```
dyersituations, line 738: Why is `goalInReviewFeedbackSubmitted` set here?

src/composables/useGoalInReview.js
```diff
@@ -143,10 +151,59 @@ export default function useGoalInReview({ apollo } = {}) {
...
+			enabled,
+			isEligible: Boolean(data?.isEligible),
+			goalStatus: data?.goalSummary?.status,
+			goalYear,
+			currentGoalYear: GOALS_CURRENT_YEAR,
+			hasViewedRecap,
+			hasCompletionPending,
+		});
+
+		if (!shouldOpen) {
```
dyersituations, line 199: Why are we exiting if the user just completed a goal? Wouldn't this be true when returning to MyKiva after checking out the last loan? And I assume we'd want to show the new modal then?

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
christian14b: Good catch! and it was worse than two. `MyKiva` had three `useGoalData` instances: the page-wide one `MyKivaPage` provides, the one `MyKivaPageContent` was injecting, and the one `useGoalInReview` creates internally. Portfolio had two. 

Indeed, `userPreferences` is `ref(null)` inside the composable, so each call gets its own store, and it only worked because every read forced network-only.

I will fix it by letting `useGoalInReview` accept an existing instance

src/util/goalInReviewTrigger.js
christian14b: Yes, called out in the epic:

```
Lenders who still have a goal in progress - show on log in as a pop up when they visit MyKiva or the Portfolio mid November just pop up once (total, not once per page)
```

dyersituations: That mentions "mid November" instead of when the feature is enabled. Is that "November" check already in place for in-progress goals?

src/composables/useGoalInReview.js
christian14b: It's called out in the epic too: `Lenders who completed their goal - the next session after they complete their goal, show as a pop up when they visit MyKiva or the Portfolio` , but I'll reconfirm it asking in the channel

dyersituations: Ah, so first load on MyKiva sets the pending, second load shows?

src/util/goalInReviewTrigger.js
christian14b: Yes, I added the date settings for this feature after the convo in slack

src/pages/MyKiva/MyKivaPageContent.vue
christian14b: Because `handleGoToDeepLink` only runs when there's a `goTo` param in the URL:

On  `/mykiva` visit with no query string the deeplink handler never fires. The two are mutually exclusive, so neither can cover for the other.

src/composables/useGoalInReview.js
christian14b: I might be missing the next `login / session` logic to show the goal in review. I'll take a look

src/util/goalInReviewTrigger.js
```diff
@@ -0,0 +1,72 @@
...
+ * @param {number|string} options.currentGoalYear The goal year in progress now.
+ * @param {boolean} options.hasViewedRecap Whether the recap has already been seen.
+ * @param {boolean} options.hasCompletionPending Whether a completed goal was already
+ *   seen on an earlier visit, which makes this the session after completion.
+ * @param {Date|string|null} options.inProgressStartDate The goal_in_review_in_progress_start
+ *   setting, the date in-progress goal setters become eligible.
+ * @param {Date} options.now The effective current date.
+ * @returns {boolean} Whether to open the recap automatically.
+ */
+export function shouldAutoOpenRecap({
```
dyersituations, line 39: Do we still have a way to set the date in query string for testing the new changes around starting in November for in-progress goals?

src/graphql/query/myKiva.graphql
```diff
@@ -184,5 +184,9 @@ query myKivaQuery {
       key
       value
     }
+    goal_in_review_in_progress_start: uiConfigSetting(key: "goal_in_review_in_progress_start") {
```
dyersituations, line 187: Please add this flag to the active experiments doc so we have the key and whatnot tracked there.

src/util/goalInReviewTrigger.js
christian14b: Good call, and yes we still have `?recapDate=YYYY-MM-DD`

`getGoalInReviewNow()` reads the query param, loadAutoOpenRecap passes it as now, and `hasInProgressReleaseStarted(inProgressStartDate, now)` compares against it. That was deliberate, it's why I used `readDateSetting` plus a manual comparison instead of settingWithinDateRange, which calls `new Date()` internally and would have ignored the override.

So with `goal_in_review_in_progress_start` set to 2026-11-15:
```
/mykiva?recapDate=2026-10-01 nothing, still before the gate
/mykiva?recapDate=2026-11-15 opens, on the gate
/mykiva?recapDate=2026-12-01 opens, after the gate
```

dyersituations (approved):

src/util/goalInReviewTrigger.js
christian14b: I'll make sure to write down a confluence doc for this all use cases + testing information

dyersituations: Either that or add to the VQA ticket. Thanks!

PR #7150, opened 2026-08-10
feat(live-loan-ads): serve the live-loan google ads feed from the kiva-ui server

roger-in-kiva:
### Context

Kiva wants to advertise **live, still-fundraising** loans on Google Ads so a click always lands on a loan a lender can actually fund (no "already funded" dead ends). This PR serves that as a **Google Ads business-data (custom) feed** — a TSV at a public URL Google fetches on its own schedule — from the **kiva-ui Express server** (`server/`).

Two problems the design solves:
1. **Freshness / correctness** — candidates come from FLSS (updated via kafka events), the freshest source of fundraising loans: funded/refunded/expired loans drop out of it, and a code-side gate removes anonymized / nameless / imageless loans before rendering. The feed is regenerated every ~5 min to track FLSS.
2. **Image compliance** — Google dynamic display **requires RGB images with an embedded ICC profile**; Kiva's stored JPEGs are RGB but carry *no* profile. The feed's `Image URL` points at **Cloudinary** (already Kiva's image CDN), whose transform converts to sRGB and force-embeds an sRGB ICC profile on the fly — no self-hosted endpoint and no image dependency in the Node server.

**Scope:** the feed service, eligibility, TSV rendering, caching, and the public feed route. Analytics, manual loan exclusion, and partner/sector exclusion are **separate tickets** (see below).

### Related links

- **Spike (design source):** [Kiva's Exploration of Live Fundraising Loans as Google Ads](https://kiva.atlassian.net/wiki/spaces/B2C/pages/4721672200) (Confluence, B2C) — the locked decisions this implements (eligibility rule, feed constraints, image compliance).
- **Feed doc:** [Live Loan Ads — Architecture & Usage](https://kiva.atlassian.net/wiki/spaces/B2C/pages/4812996631) (Confluence, B2C) — how to use the feed and what the data looks like.
- **Jira:** [MP-3013 — Live loan ads – build (google)](https://kiva.atlassian.net/browse/MP-3013). Siblings (out of scope here): [MP-3095 (analytics)](https://kiva.atlassian.net/browse/MP-3095) · [MP-3098 (manual loan exclusion)](https://kiva.atlassian.net/browse/MP-3098) · [MP-3099 (partner + sector exclusion)](https://kiva.atlassian.net/browse/MP-3099).
- **Google Ads docs:** [Create a feed for responsive ads](https://support.google.com/google-ads/answer/6053288) (required ID/Item title/Final URL attributes, 25-char caps, same-domain Final URL, 11.4 MB / 6 MP image ceiling) · [Dynamic ad requirements](https://support.google.com/adspolicy/answer/7181012) (RGB + embedded ICC profile) · [About business data feeds](https://support.google.com/google-ads/answer/6072708).

### Architecture

```mermaid
flowchart TD
    subgraph Feed["Feed path — GET /live-loan/ads/google-feed.tsv"]
        direction TB
        G["Google Ads<br/>(scheduled fetch)"] --> R["liveLoanRouter"]
        R --> CACHE[("memjs<br/>fresh 5m · last-good 3d")]
        CACHE -->|"hit"| OUT["200 text/tab-separated-values<br/>Cache-Control: no-store"]
        CACHE -->|"miss"| GEN["generateGoogleFeed"]
        GEN --> ELIG["ads-eligibility<br/>FLSS fundraisingLoans · cap 100"]
        ELIG --> ROW["feed-row + copy-safety<br/>7-column TSV"]
        ROW --> OUT
        GEN -.->|"generation fails"| LG["serve last-good (3d)<br/>or 503 + Retry-After"]
    end

    subgraph Image["Image path — the row's Image URL (no kiva-ui endpoint)"]
        direction TB
        GI["Google / browser"] --> CLD["res.cloudinary.com/kiva<br/>c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc<br/>/remote/&lt;hash&gt;.jpg"]
        CLD --> SRC[("Cloudinary fetches Kiva's stored<br/>original via the remote/ mapping")]
        CLD -->|"sRGB + embedded ICC JPEG"| JOUT["200 image/jpeg (served by Cloudinary)"]
    end
```

**Components (`server/util/live-loan/ads/`)**

| Module | Responsibility |
|---|---|
| `ads-eligibility.js` | FLSS `fundraisingLoans` query with the locked filters (risk ≥3.5, repayment ≤12mo, default ≤0.05, amount-left ≥$200, expiry ≥3 days); caps at 100/generation; code-side gate (non-anonymized, has first name, has image hash) |
| `google-display/feed-row.js` | Maps a loan to the 7 Google columns; builds the Cloudinary `Image URL`; hard-caps short text to 25 chars (code-point-safe); drops ad-unsafe copy (banned words / ALL-CAPS) |
| `google-display/google-feed.js` | Orchestrates eligibility → rows → TSV; owns the feed cache keys/TTLs (fresh 5m, last-good 3d, 60s failure backoff) |
| `constants.js` | `KIVA_PROD_HOST` for the ad Final URL + `CLOUDINARY_AD_IMAGE_BASE`/`_TRANSFORM` for the image URL — hardcoded prod (never `BASE_URL`-derived), since Google serves these to real lenders even when the feed is generated in a non-prod env |

The feed route lives in the existing `server/live-loan-router.js` (mounted at `/live-loan`).

### Endpoints

| Method | Path | Response |
|---|---|---|
| GET | `/live-loan/ads/google-feed.tsv` | `200` TSV (`Cache-Control: no-store`); serves the cached copy between regenerations; last-good on a generation failure; `503 + Retry-After` when there's no last-good |

The `Image URL` column is a **Cloudinary** URL (`https://res.cloudinary.com/kiva/c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc/remote/<hash>.jpg`) — Google fetches it directly, so there is no kiva-ui image route.

### Key design decisions

- **Feed shape = the 7 Google Ads Custom-feed attributes** — the header row is Google's exact attribute names (capitalization/spacing must match Google's spec): `ID, Item title, Item description, Item subtitle, Item category, Image URL, Final URL`. `ID` and `Item title` are required, `Final URL` is required (unless a Destination URL is used), the rest are recommended. Row values are `Item title=Help {name}`, `Item description=Support {name}`, `Item subtitle=country`, `Item category=sector`; Google's 25-char caps on title/subtitle/description are enforced by `truncate()`. The copy-safety drop (banned words + ALL-CAPS) removes policy-risky rows. The "$25" LSP amount and fuller loan-use/story copy live at Google's **account level**, not in the feed rows.
- **FLSS is the source of freshness** — no re-check against a slower resolver. FLSS `fundraisingLoans` (kafka-updated) already excludes funded/refunded/expired, and the eligibility gate removes anonymized / nameless / imageless loans on the FLSS record.
- **ICC images via Cloudinary** — `Image URL` points at a Cloudinary transform (`cs_srgb` converts to sRGB, `fl_force_icc` embeds the sRGB ICC profile Google requires; `f_jpg` pins JPEG — never `f_auto`, which serves webp Google rejects; `c_limit,w_1200,h_1200` bounds size without upscaling). Cloudinary fetches Kiva's stored original through the pre-configured `remote/` mapping, so no self-hosted endpoint, no `sharp`, no monolith work, and no uploads. Verified against a real loan image: JPEG, sRGB, embedded 3144-byte ICC, ≤1.1 MP, ~128 KB, HTTPS. (An off-domain image host is allowed — Google's same-domain rule applies only to the Final URL.)
- **Caching is server-side only** — the feed is served `Cache-Control: no-store` so a browser/CDN/Google copy can't linger and keep advertising loans that have since funded or expired; the rows must stay FLSS-fresh. Load/scraper protection is the server-side memjs cache (fresh 5m): a fresh copy is served between regenerations so scrapers can't drive FLSS + hydration on every hit. Last-good is served on a generation failure and re-primed into the fresh key for 60s so an outage re-runs the pipeline at most once/window.
- **No server-side kill switch** — pausing is handled on the Google Ads side (pause or unregister the campaign in Business data), so a separate server flag would be redundant. Consequently the launch gate is *not* wiring the feed URL into Google Ads until Legal clears (see below).
- **Ad URLs hardcode prod** — the Final URL host and the Cloudinary image URL are deliberately not derived from `BASE_URL`; an env-aware host would leak a non-prod URL into a live ad.

### Tests

`npm run unit` — the full unit suite is green (**4699 passed / 6 skipped**). Ads coverage:

| Spec | Covers |
|---|---|
| `ads-eligibility.spec.js` (15) | filter thresholds, cap 100, no sort, no pageNumber, image-hash gate, anonymization/name drops |
| `feed-row.spec.js` (16) | 7-column mapping, `Help`/`Support` copy, country subtitle, code-point-safe truncation, copy-safety (banned words / ALL-CAPS), UTM Final URL, Cloudinary Image URL |
| `google-feed.spec.js` (11) | TSV serialization, eligibility→rows orchestration, ad-safety drop, header-only feed when no loans are eligible |
| `live-loan-router.spec.js` (ads cases) | `no-store` on the generated *and* cache-hit paths, cache hit/miss, generate + cache fresh & last-good, outage backoff, 503+Retry-After |

Live-verified end-to-end: the full HTTP path over `https://kiva-ui.local/` returns a 100-row 7-column feed, and the Cloudinary `Image URL` returns a Google-compliant sRGB+ICC JPEG.

### How to test locally (Caddy / `kiva-ui.local`)

Prereqs: `/etc/hosts` maps `kiva-ui.local` → `127.0.0.1`, and Caddy is running (it fronts `kiva-ui.local` → the server on `:8888` with a local TLS cert).

```bash
# Build, then start the server behind Caddy. dev-custom-host points at the dev GraphQL gateway,
# which exposes fundraisingLoans — the default k8s-local config routes to the monolith graph and
# will NOT resolve the feed's FLSS query.
npm run build
npm run start -- --config=dev-custom-host
```

**Feed** — no flag to set; it serves rows as soon as FLSS returns eligible loans:
```bash
curl -sSk -D - https://kiva-ui.local/live-loan/ads/google-feed.tsv | head
# → HTTP/2 200, text/tab-separated-values, Cache-Control: no-store, header + up to 100 rows
```
(`-k` skips the local Caddy cert.)

**Image** — the `Image URL` column is a Cloudinary URL; verify it's Google-compliant:
```bash
HASH=<a hash from a feed row's Image URL>
curl -sS "https://res.cloudinary.com/kiva/c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc/remote/$HASH.jpg" -o /tmp/ad.jpg
sips -g space -g profile /tmp/ad.jpg     # → RGB / sRGB IEC61966-2.1
magick identify -verbose /tmp/ad.jpg | grep -i "Colorspace:\|Profile-icc"
# → Colorspace: sRGB   /   Profile-icc: 3144 bytes
```

### Rollout & rollback

- No server-side flag. The feed becomes live when its URL is **registered in Google Ads Business data**; it stops serving ads when the campaign is **paused or unregistered** there.
- Rollback = pause/unregister the campaign in Google Ads — no deploy.

### Out of scope / follow-ups

- **MP-3095** analytics (incl. the funded-click tracking event) · **MP-3098** manual loan exclusion · **MP-3099** partner + sector exclusion.
- **Launch gate:** consent / borrower-terms clearance is a Legal go/no-go, and sector curation lands in MP-3099 — **do not register the feed URL in Google Ads until Legal confirms borrower terms cover paid third-party ads.**
- **Fastly proxy (usage-gated):** if Cloudinary usage turns out heavy, front the image URL with Fastly (also puts the image on-domain). Google re-hosts feed images on its own crawl schedule, so this is expected to be unnecessary — revisit on metrics.

---
*Refs: MP-3013*


server/util/live-loan/ads/ads-freshness.js
```diff
@@ -0,0 +1,57 @@
...
+
+// Re-fetch a chunk of candidates from the authoritative lend.loan resolver in one round trip, using
+// GraphQL field aliases (l0, l1, ...). lend.loan(id) is the source-of-truth loan record, distinct
+// from the FLSS search index which can lag. A failed chunk drops only its own loans (they stay
+// unverified and are excluded downstream), so one bad chunk never poisons the rest.
+async function fetchLiveLoans(ids) {
+	const selections = ids
+		.map((id, i) => `l${i}: loan(id: ${id}) { id status anonymizationLevel }`)
+		.join('\n\t\t\t');
+	const query = `query adFreshness {\n\t\tlend {\n\t\t\t${selections}\n\t\t}\n\t}`;
```
mcstover: This endpoint is not more "fresh" than flss and lags by 5-10 minutes as it's driven by an index that's updated by the every 5 minute cron which sometimes takes more than 5 mins to complete.

I don't think it's useful to try and check against monolith data as a backup for flss.

mcstover: FLSS gets updated via kafka events so will be more up to date.

mcstover:
@roger-in-kiva @dyersituations Have we considered designing this feed to be more like our live-loans endpoints that do not try to rely on or cache a "set" list of loans?

I'm also generally concered by the 1hr cache on the feed result. I think what we need here is a feed that lives for about 5-10 minutes max...

roger-in-kiva:
> @roger-in-kiva @dyersituations Have we considered designing this feed to be more like our live-loans endpoints that do not try to rely on or cache a "set" list of loans?
> 
> I'm also generally concered by the 1hr cache on the feed result. I think what we need here is a feed that lives for about 5-10 minutes max...

All feedback to improve this is welcome, thanks!

server/util/live-loan/ads/ad-image.js
```diff
@@ -0,0 +1,48 @@
...
+			return null;
+		}
+		source = Buffer.from(await response.arrayBuffer());
+	} catch (err) {
+		warn(`Ad image: source fetch failed for ${hash}, ${err}`);
+		return null;
+	}
+
+	try {
+		return await sharp(source)
```
dyersituations: Does Cloudinary have an image processing option we could use?

server/live-loan-router.js
```diff
@@ -5,6 +5,21 @@ import drawLoanCard from './util/live-loan/live-loan-draw.js';
...
+// can't drive the FLSS/hydrate pipeline on every hit; the last-good copy is served if a regeneration
+// fails so a transient outage never empties the feed.
+const ADS_FEED_FRESH_KEY = 'google-ads-feed';
+const ADS_FEED_LAST_GOOD_KEY = 'google-ads-feed-last-good';
+const ADS_FEED_FRESH_TTL = 5 * 60; // 5 minutes — keep the feed close to FLSS's ~5-min refresh cadence
+const ADS_FEED_LAST_GOOD_TTL = 3 * 24 * 60 * 60; // 3 days
+// On a generation failure the last-good feed is re-primed into the fresh key for this short window,
+// so an FLSS/gateway outage re-runs the full pipeline at most once per window instead of every request.
+const ADS_FEED_FAILURE_BACKOFF_TTL = 60; // 1 minute
+const ADS_IMAGE_TTL = 24 * 60 * 60; // 1 day
```
dyersituations: Might make sense to put these Google-ad-specific constants into one of the new google util files.

server/live-loan-router.js
```diff
@@ -414,6 +429,101 @@ export default function liveLoanRouter(cache) {
 		});
 	});
 
+	// Google Ads live-loan feed (public TSV, fetched on Google's own schedule)
+	router.get('/ads/google-feed.tsv', async (req, res) => {
+		await trace('live-loan.ads.googleFeed', { resource: req.path }, async () => {
+			const sendFeed = feed => {
+				res.contentType('text/tab-separated-values');
+				// The feed is kill-switchable, so it must never be held by a downstream cache
```
dyersituations: I'm wondering if the kill switch is needed now that we're looking to have this in `ui`. If marketing can turn this off directly in Google ad settings, what would a kill switch give us beyond that functionality?

roger-in-kiva: Good call.
This is basically an spike requirement, only thing I see beyond that functionality is the access to kill it.
Google ads is probably marketing only against kill switch based on admin allows others to do it.

server/util/live-loan/ads/ad-image.js
roger-in-kiva: You were right to push on this — turns out Cloudinary *can* give us a Google-ready image.

The trick is `cs_srgb` + `fl_force_icc` together. On their own neither embeds a profile, but combined they output a JPEG that's converted to sRGB *and* carries an embedded sRGB ICC profile — which is exactly what Google wants. Tested it against a real loan image:

`https://res.cloudinary.com/kiva/w_1200,f_jpg,cs_srgb,fl_force_icc/remote/<hash>.jpg`
→ JPEG, sRGB, embedded ICC (3144 bytes), 1200×891, ~125 KB ✅

No custom profile upload needed, and no monolith work — the `remote/` fetch mapping into our image store is already set up on our Cloudinary account (it's how the monolith serves loan images today). The feed would just build this URL string.

**What we'd get:** drop `ad-image.js`, the `/ads/image/:hash` route, the image cache, and the `sharp` dependency entirely — the whole thing becomes a URL.

**What to watch:**
- Must use `f_jpg`, never `f_auto` — `f_auto` serves webp to browsers/Google, which they reject.
- The URL is unsigned, which works because strict transformations is off on our account. If that ever gets turned on, unsigned URLs like this would break — so it's a dependency on account config.
- Google's fine with the image being on `res.cloudinary.com` (off-domain images are allowed; only the Final URL has to be same-domain).

Leaning toward making the switch since it removes a fair bit of code and a native dep. Happy to go either way — let me know what you think.

emuvente: I think using Cloudinary that way is a good solution. My only concern would be the bandwidth usage, so we should consider proxying that through Fastly as well if there's a lot of usage.

roger-in-kiva: Agreed it's a good approach. On the bandwidth point — I dug into how Google handles feed images, and it should be light:

Google **downloads the image and stores its own copy for delivery at impression** rather than hot-linking our URL on every view — it only re-crawls the URL periodically (every few days). Straight from Google's docs:

- *"Google pulls the image down when the feed is uploaded … and stores a copy of it on Google servers for delivery at impression."* — [Dynamic creative feed guidelines](https://support.google.com/richmedia/answer/3123027)
- Feed images are cached and re-crawled on Google's own schedule, not per request — [Dynamic remarketing: create a feed for responsive ads](https://support.google.com/google-ads/answer/6053288)

So Cloudinary would see ~100 images fetched at crawl cadence, not impression-level traffic. I'd ship pointing at Cloudinary, watch the actual usage numbers, and only add a Fastly proxy if it turns out heavy — which would also get us on-domain image URLs. I'll capture the Fastly proxy as a usage-gated follow-up so it's not lost.

dyersituations: @emuvente based on the Google doc references, do you think it's worth investing in the Fastly layer now?

server/live-loan-router.js
dyersituations: Thoughts on removing the kill switch to simplify the implementation slightly? If Google requests and caches like you mentioned here in the other thread, I'm not too concerned about needing to turn this feature off quickly.

test/unit/specs/server/util/live-loan/ads/google-display/feed-row.spec.js
```diff
@@ -0,0 +1,133 @@
...
+			// 24 ASCII + an astral emoji lands exactly at the 25th code point; a UTF-16 slice would
+			// cut the emoji mid-surrogate and emit invalid UTF-8.
+			const result = truncate(`${'a'.repeat(24)}\u{1F600}bcd`);
+			expect(result).toEqual(`${'a'.repeat(24)}\u{1F600}`);
+			expect([...result]).toHaveLength(25);
+		});
+	});
+
+	describe('buildTitle', () => {
+		it('builds "Help {firstName}" from the primary borrower', () => {
```
dyersituations, line 62: @roger-in-kiva please create a follow-up ticket to review the `Help ...` and `Support ...` copy with marketing.

server/util/live-loan/ads/ad-image.js
dyersituations: @roger-in-kiva please create a follow-up ticket for if/when we want to add a Fastly layer to this feature. We can test on dev without this layer for the time being.

@emuvente let us know your thoughts on needing the Fastly layer in general.

Thanks!

server/util/live-loan/ads/google-display/feed-row.js
```diff
@@ -0,0 +1,73 @@
...
+const UTM_QUERY = 'utm_medium=paid&utm_source=google&utm_campaign=liveloans';
+const KIVA_LEND_BASE = `${KIVA_PROD_HOST}/lend`;
+// Cloudinary's `remote/` mapping fetches the loan's stored original at <hash>.jpg to transform on the fly.
+const AD_IMAGE_BASE = `${CLOUDINARY_AD_IMAGE_BASE}/${CLOUDINARY_AD_IMAGE_TRANSFORM}/remote`;
+// Google caps Item title / subtitle / description at 25 characters.
+const CAP = 25;
+
+// Words barred from ad copy: financial-services / ecommerce framing that risks Google reclassifying
+// the ads out of the nonprofit lane.
+const BANNED_WORDS = ['loan', 'buy', 'invest', 'interest', 'guaranteed', 'earn'];
```
dyersituations, line 13: Do we have an idea of how frequent these words appear in the loans? Since we filter out these loans, I'm wondering if we end up filtering a large number of loans. If so, we might need to instead clean copy.

package-lock.json
```diff
@@ -15036,9 +15036,9 @@
 			}
 		},
 		"node_modules/detect-libc": {
-			"version": "2.0.4",
-			"resolved": "https://registry.npmjs.org/detect-libc/-/detect-libc-2.0.4.tgz",
-			"integrity": "sha512-3UDv+G9CsCKO1WKMGw9fwq/SWJYbI0c5Y7LU1AXYoDdbhE2AHQ6N6Nb34sG8Fj7T5APy8qXDCKuuIHd1BR0tVA==",
+			"version": "2.1.2",
```
dyersituations: Looks like the lock file can be reverted?

server/util/live-loan/ads/google-display/google-feed.js
```diff
@@ -0,0 +1,43 @@
+import { fetchAdEligibleLoans } from '../ads-eligibility.js';
+import { loanToFeedRow, isRowAdSafe, FEED_COLUMNS } from './feed-row.js';
+import { warn } from '../../../log.js';
+
+// Cache keys + TTLs for serving the feed. A fresh copy is served from cache between regenerations so
+// scrapers can't drive the FLSS/hydrate pipeline on every hit; the last-good copy is served if a
+// regeneration fails so a transient outage never empties the feed.
+export const ADS_FEED_FRESH_KEY = 'google-ads-feed';
+export const ADS_FEED_LAST_GOOD_KEY = 'google-ads-feed-last-good';
+export const ADS_FEED_FRESH_TTL = 5 * 60; // 5 minutes — keep the feed close to FLSS's ~5-min refresh cadence
```
dyersituations, line 10: If Google only queries daily, can we make this window larger?

server/util/live-loan/ads/ads-eligibility.js
```diff
@@ -0,0 +1,82 @@
...
+
+// Fetch fundraising loans that pass the FLSS ad-eligibility filters, then apply the code-side gate.
+export async function fetchAdEligibleLoans(count = AD_FEED_LOAN_COUNT) {
+	const data = await fetchGraphQL(
+		{
+			query: `query adEligibleLoans($filters: [FundraisingLoanSearchFilterInput!]) {
+				fundraisingLoans(
+					limit: ${count},
+					filters: $filters,
+					origin: "google:live-loan-ads"
```
dyersituations, line 68: Do we need to tell the data team about this new `origin`?

roger-in-kiva: For sure, I'll tag them in the ticket when merged.

server/util/live-loan/ads/constants.js
```diff
@@ -0,0 +1,12 @@
...
+// it must always point at the live prod site even when the feed is generated in a non-prod environment.
+export const KIVA_PROD_HOST = 'https://www.kiva.org';
+
+// Ad images are delivered by Cloudinary, which fetches the loan's stored original through Kiva's
+// pre-configured `remote/` mapping and transforms it on the fly. The transform pins JPEG (f_jpg — never
+// f_auto, which serves webp that Google rejects), converts to sRGB, and force-embeds an sRGB ICC profile
+// (fl_force_icc) because Google Ads dynamic display rejects profile-less JPEGs, which Kiva's originals
+// are; c_limit,w_1200,h_1200 bounds the longest edge to 1200px without upscaling smaller sources.
+export const CLOUDINARY_AD_IMAGE_BASE = 'https://res.cloudinary.com/kiva';
+export const CLOUDINARY_AD_IMAGE_TRANSFORM = 'c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc';
```
dyersituations, line 14: Is 1200 x 1200 in the list of image sizes we generate in Cloudinary already?

server/util/live-loan/ads/google-display/google-feed.js
roger-in-kiva: This is @mcstover suggestion above.

dyersituations: I suppose a new generation every 5 mins isn't a huge amount... Sounds good.

server/live-loan-router.js
roger-in-kiva: Already done — the kill switch is removed as of `587121373`. The feed now always generates, and pausing is handled Google-Ads-side (pause/unregister the campaign in Business data), which is exactly what you landed on here. `Cache-Control: no-store` + the 5-min cache stay, now motivated by keeping rows FLSS-fresh rather than defeating a flag.

server/util/live-loan/ads/google-display/feed-row.js
roger-in-kiva: Measured it against live FLSS data: of **323 eligible loans, 0 were dropped** by copy-safety (0% — 0 banned-word, 0 ALL-CAPS). It's structural: the rendered copy is only `Help {first name}` / `Support {first name}`, so the banned words (`loan|buy|invest|interest|guaranteed|earn`) can't match a first name, and borrower names come title-cased so the ALL-CAPS check doesn't fire either. So it's a near-inert safety net today — it only starts dropping rows if we introduce richer loan-use/story copy. Nothing's being over-filtered, so no need to "clean copy instead" for now. Captured this (and the wider copy review) in MP-3115: https://kiva.atlassian.net/browse/MP-3115

test/unit/specs/server/util/live-loan/ads/google-display/feed-row.spec.js
roger-in-kiva: Created MP-3115 for this: https://kiva.atlassian.net/browse/MP-3115

I expanded it beyond just the `Help` / `Support` copy to track the full set of Marketing/Legal pre-launch sign-offs the spike calls for — borrower consent for paid third-party ads, nonprofit-vs-financial-services framing/classification, the ad-copy + ad-safety wordlist, campaign setup, and the landing-page nonprofit disclosure — so it's one tracker for the launch prerequisites.

server/util/live-loan/ads/ad-image.js
roger-in-kiva: Created MP-3116 for the Fastly layer: https://kiva.atlassian.net/browse/MP-3116

Usage-gated — we ship pointing at Cloudinary and only add the Fastly proxy if usage proves heavy (which also gets us on-domain image URLs). Agreed we can run on dev without it for now.

server/util/live-loan/ads/constants.js
roger-in-kiva: Not a pre-generated / named size — it's an on-the-fly transform the feed builds (`c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc`), which Cloudinary applies live via the `remote/` fetch mapping. The `cs_srgb` + `fl_force_icc` part (the embedded sRGB ICC profile Google requires) is specific to this feed, so it's a new transform rather than reusing a loan-card size. It works unsigned because strict transformations is off on our account (arbitrary transforms return 200); if that's ever enabled we'd register this as an allowed named transformation — the account-config caveat noted in the Cloudinary thread above.

dyersituations (approved):
Tested it locally and the new route generated the tsv file quickly!

PR #7151, opened 2026-08-10
feat: add button to impact progress row to see goal recap

christian14b:
https://kiva.atlassian.net/browse/MP-3064

image

### DIFF
--- a/src/components/MyKiva/BadgesSectionV2.vue
+++ b/src/components/MyKiva/BadgesSectionV2.vue
@@ -29,6 +29,7 @@
 					:is-annual-goal="badge.isAnnualGoal"
 					:is-historical-goal="badge.isHistoricalGoal"
 					:goal-progress="badge.goalProgress"
+					:show-recap-cta="Boolean(badge.showRecapCta)"
 					:year="badge.year ?? GOALS_CURRENT_YEAR"
 					class="tw-self-start tw-mx-auto"
 				/>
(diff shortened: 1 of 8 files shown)


src/util/goalRecapEntryPoint.js
```diff
@@ -0,0 +1,62 @@
+import { GOAL_STATUS } from '#src/composables/useGoalData';
```
dyersituations, line 1: We are getting a lot of utils for goal recap. Can we either: a) move all to a folder under `util` or b) combine files as makes sense.

src/components/MyKiva/MyKivaFeaturedSlot.vue
```diff
@@ -192,4 +214,8 @@ const handleCtaClick = () => {
 const handleEditClick = () => {
 	emit('edit-click');
 };
+
+const handleViewGoalRecap = () => {
+	emit('view-goal-recap', goalYear.value);
```
dyersituations, line 217: This one is missing analytics, should we move the analytics event to the parent handler?

src/components/MyKiva/BadgesSectionV2.vue
```diff
@@ -98,6 +105,22 @@ const {
 
 const userHasGoal = computed(() => !!userGoal.value && Object.keys(userGoal.value).length > 0);
 
+const userGoalYear = computed(() => (
+	userGoal.value?.dateStarted ? new Date(userGoal.value.dateStarted).getFullYear() : null
```
dyersituations: This logic is duplicated and could exist in `shouldShowRecapEntryPoint`.

src/util/goalRecapEntryPoint.js
christian14b: Yes, you're right. I created a ticket specifically for refactoring: https://kiva.atlassian.net/browse/MP-3104

Once this is done I'll start that to clean up the new code

dyersituations (approved):

PR #7152, opened 2026-08-10
feat: [MP-3082] - goal in review content animations

cristhianDt:
# MP-3082

## Summary

add some animations to goal in review content: https://www.figma.com/make/dnnmbQAGDESwHSAhvrWfk7/Goal-review-animations?p=f&t=mi69LQS8Mznk9wqg-0

https://kiva.atlassian.net/browse/MP-3082

## Evidences

https://github.com/user-attachments/assets/94d3f694-a0a9-4e4b-828d-f053f194ae81


src/assets/css/animations.css
```diff
@@ -0,0 +1,86 @@
+/*
```
cristhianDt, line 1: (thoughts) maybe we could include a ‘technical improvement’ ticket and tackle it later. the idea would be to create a Storybook for animations that can be reused, and maybe include animation ‘tokens’? Haha, @dyersituations  @mcstover  you’re more experienced with this than I am. what do you think?

dyersituations: Is there a path here to use postcss in a component instead of scss? We're generally moving away from using scss in our apps.

dyersituations:
@cristhianDt the animates are looking awesome! The only aspect the looked off for me when testing locally was that the pie chart is shown "loading" briefly, then disappears, then animates in. Can we hide the "loading" state so we just get the pie chart animating in smoothly?

image

cristhianDt:
> @cristhianDt the animates are looking awesome! The only aspect the looked off for me when testing locally was that the pie chart is shown "loading" briefly, then disappears, then animates in. Can we hide the "loading" state so we just get the pie chart animating in smoothly?
> 
> 
> 
> image

I could yes, actually in the beginning I did it but the section load a first blank space with a small height then the space expand it height abruptly, I can't find a way to make it smoothly, any idea is welcome

src/assets/css/animations.css
cristhianDt: Yes but I am not sure if we could reuse the animations, otherwise I have to duplicate the code and that's what I don't want, let me search about a bit more

dyersituations:
> I could yes, actually in the beginning I did it but the section load a first blank space with a small height then the space expand it height abruptly, I can't find a way to make it smoothly, any idea is welcome

Could we just use a default height for that section that would be the ending space needed for the graph?

src/assets/css/animations.css
dyersituations: Not the end of the world to use scss since it exists in this app, but worth a look, thanks for investigating!

cristhianDt:
> > I could yes, actually in the beginning I did it but the section load a first blank space with a small height then the space expand it height abruptly, I can't find a way to make it smoothly, any idea is welcome
> 
> 
> 
> Could we just use a default height for that section that would be the ending space needed for the graph?

Let me try it

cristhianDt:
> > I could yes, actually in the beginning I did it but the section load a first blank space with a small height then the space expand it height abruptly, I can't find a way to make it smoothly, any idea is welcome
> 
> Could we just use a default height for that section that would be the ending space needed for the graph?

done, looks good! thank you @dyersituations

src/assets/css/animations.css
cristhianDt: okey, I changed the approach and register the new file in `client-entry.js` and works, let me know if you are agree

src/components/MyKiva/GoalInReview/GoalInReviewHeadline.vue
```diff
@@ -190,25 +196,75 @@ const stats = computed(() => [
 	}
 }
 
+/* Title stack entrance — shares the global .kv-fade-up effect; this slide sets
+   each line's travel distance (--kv-fade-up-distance, the value baked inside the
+   keyframe) and its delay, so eyebrow -> headline -> subtitle rise in sequence. */
+.slide1-eyebrow {
```
dyersituations: After merging with file rename, thoughts on renaming these classes to remove "1" and other numbers.

src/components/MyKiva/GoalInReview/GoalInReviewSlide2.vue
```diff
@@ -76,13 +86,38 @@ const totalBorrowers = computed(() => Number(props.borrowerCount) || cards.value
...
+const cardAnimationDelay = index => ({ animationDelay: `${0.2 + index * 0.175}s` });
 </script>
 
 <style lang="postcss" scoped>
 .borrower-grid {
 	grid-template-columns: repeat(3, minmax(0, 1fr));
 }
 
+/* Entrance choreography — the shared .kv-fade-up effect lives in
+   global/animations.scss. */
```
dyersituations: Old file name here.

dyersituations: I see other instances in the changeset as well.

src/components/MyKiva/GoalInReview/GoalInReviewCollectiveImpact.vue
```diff
@@ -55,24 +59,107 @@
...
+const COUNT_UP_DURATION = 2200;
+const easeOutCubic = t => 1 - (1 - t) ** 3;
+
+const section = ref(null);
+const displayValues = ref(stats.map(stat => stat.target));
+
+let observer = null;
+let rafId = null;
+
+const prefersReducedMotion = () => typeof window !== 'undefined'
```
dyersituations, line 102: Why is reduced motion needed?

src/components/MyKiva/GoalInReview/GoalInReviewHeadline.vue
cristhianDt: 😔 It would have been great to have waited for that PR until mine was merged

src/assets/css/animations.css
```diff
@@ -0,0 +1,86 @@
...
+	Global animations.
```
dyersituations: Was both this and the scss version global? I'm wondering if there's a route here that is more localized and not served with each page. Thoughts?

src/components/MyKiva/GoalInReview/GoalInReviewHeadline.vue
dyersituations: Luckily AI can do these types of merge conflict resolutions (and renames) for us these days.

src/assets/css/animations.css
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+/* Rebound "pop": the element rises and scales slightly past full size, then
+   settles back to 1 — a celebratory entrance for cards, stats, callouts, etc.
+   The overshoot lives in the keyframe (scale 1.09 at 60%) so the timing
+   function can stay a plain ease-out; a springy cubic-bezier on top would
+   double up on the bounce and read as jittery. */
+.kv-rebound-in {
```
christian14b, line 10: Can this be moved under `GoalInReviewModal.vue` container as a `css` scoped section?

src/util/animation/confettiUtils.js
```diff
@@ -9,5 +9,8 @@ export const showConfetti = () => {
 		spread: 200,
 		colors: ['#6AC395', '#223829', '#95D4B3'],
 		disableForReducedMotion: true,
+		// Above the design system's modal layer (z-modal = 1400) so bursts render
+		// over modal overlays; kept below toast (1700) and tooltip (1800).
+		zIndex: 1600,
```
christian14b, line 14: My guess is that we can use 1800 since might be weird that some elements can be above the confetti, but good callout

src/assets/css/animations.css
christian14b: Worth flagging one gotcha if we do move it: scoped on its own won't apply. The `kv-fade-up` / `kv-scale-in` classes are on elements inside the slide components, and scoped styles only match the component's own template plus a child's root element, so the selectors would compile to `.kv-fade-up[data-v-modal]` and match nothing. 

Two ways that work:
- `:deep(.kv-fade-up)` etc., keeping the @keyframes at the top level of the block
- Or drop them into the modal's existing `unscoped` block under `.goal-in-review-modal`, which is what the [data-test=kv-lightbox] and #kvLightboxBody rules right below already do

src/components/MyKiva/GoalInReview/GoalInReviewSlide2.vue
cristhianDt: updated

src/assets/css/animations.css
cristhianDt: those are good suggestions, and my agent actually recommended similar approaches as well. the main reason I decided not to go that route is that it started to move away from the original goal of keeping this reusable for other components outside of this project. in another thread, we’re currently looking at ways to avoid loading the classes on every page, so ideally I’d like to preserve the shared-file approach if possible, if not we could use those recommendations, for sure!

src/util/animation/confettiUtils.js
cristhianDt: you mean like a tooltip index https://github.com/kiva/kv-ui-elements/blob/main/@kiva/kv-tokens/tokens/core/misc.json#L19 and make sense, I also agree I don't think there will be any elements above the confetti, I was wondering whether to use `stratosphere = 3000`, what do you think?

src/components/MyKiva/GoalInReview/GoalInReviewCollectiveImpact.vue
cristhianDt: that guard is on the count-up (runCountUp returns early → shows final numbers instead of ticking 0→target). it's JS-driven motion, so prefers-reduced-motion users get the final values immediately; it's not mandatory to handle it this way, but I saw it as an AI improvement that could be worthwhile, do you want me to remove it?

src/assets/css/animations.css
cristhianDt: I tried another approach and works

src/assets/css/animations.css
cristhianDt: I removed from client-entry so it's not loaded in each page, looks like importing in the parent `src/components/MyKiva/GoalInReview/GoalInReviewModal.vue#L84` the children inherit the classes, I tested locally and each animation persists; I'd appreciate it if you could test it on your laptop as well, just to get a double check 😂

src/util/animation/confettiUtils.js
christian14b: 1800 should be good enough with the all the elements we hav... just thinking out load either value seems ok,

src/assets/css/animations.css
christian14b: Should this still be saying "Global animations" if it's narrowed to goal in review for now?

dyersituations: Since this is now scoped to just the goal in review modal, we could indeed rename the file to be specific to that project and rename this comment.

dyersituations (approved):
Looks great! I would consider the last rename of the animations file, but I don't think it's critical.

christian14b (approved):
Tested locally and looks great! only missing thing is the animation file renaming, but not a blocker

src/assets/css/animations.css
cristhianDt: I think we'll use this animation more often in the future, but if there's no problem, I'd like to leave it as is for now and add a comment like **'to use it, import the CSS in the parent component similar to GoalInReviewModal'**, agree?

cristhianDt:
```suggestion
	Animations for the GoalInReviewModal component (in the beginning), but reusable for any other page or project. To use it, import the CSS in the parent component similar to '‎src/components/MyKiva/GoalInReview/GoalInReviewModal.vue' does
```

dyersituations (approved):

PR #7153, opened 2026-08-11
refactor: move all goal in review files to one util, instead of multi…

christian14b:
…ples per component

https://kiva.atlassian.net/browse/MP-3104


roger-in-kiva:
As part of the refactor, worth renaming slide components from numbers to something more semantic related with each section.

cristhianDt:
nice!, yeah this is what I wanted and commented in a previous PR, thanks!

christian14b:
> As part of the refactor, worth renaming slide components from numbers to something more semantic related with each section.

Good callout. I'll do that too here

dyersituations (approved):

PR #7155, opened 2026-08-12
feat: [MP-3098] exclude admin-listed loan ids from the ads feed

roger-in-kiva:
## Summary
Adds an admin-editable denylist that keeps specific loans out of the live-loan Google Ads feed. The excluded ids are read from a `uiConfigSetting` (`live_loan_ads_excluded_loan_ids`, edited in KivaAdmin → Settings Manager) each time the feed regenerates, and pushed **into the FLSS `fundraisingLoans` query** as `loanIds: { none: [...] }` — so excluded loans never enter the candidate set (no post-fetch filtering, no wasted slots against the 100-candidate cap).

- **In-query exclusion**, not a post-fetch gate — an excluded loan never returns from FLSS, so it's dropped even if it passes every other eligibility criterion.
- **Fail-open:** an empty, unset, or malformed setting → no exclusions; a failed read returns `[]` + a `warn` so the feed never errors or empties.
- **Parsing:** comma-delimited integers, whitespace-tolerant, non-numeric ignored, de-duplicated; strict `/^\d+$/` so `Number()`-coercible junk (`0x1A`, `1e3`, `-7`, `5.5`) can't slip a bad id into the filter.
- **No separate setting cache** — the feed's existing ~5-min fresh cache already bounds reads, and keeps "flush the one feed key = immediate" true.
- **Reusable seam for MP-3099:** `fetchExcludedIds(key)` + `excludeIds(field, ids)` + the options-object signature make partner/sector exclusion purely additive.
- **Setting key string** is bare `live_loan_ads_excluded_loan_ids` (the `ui.` prefix is the Settings Manager storage namespace), matching the `excluded_partners`/`excluded_sectors` precedent — ⚠️ pending an ops confirmation before launch.
- **Urgent "pull now"** is an ops-level flush of memjs key `google-ads-feed` (runbook item); no flush endpoint in this PR.

## Related
- Story: [MP-3098](https://kiva.atlassian.net/browse/MP-3098) · Epic: [MP-2908](https://kiva.atlassian.net/browse/MP-2908)
- Related pattern: [MP-3099](https://kiva.atlassian.net/browse/MP-3099) — partner/sector exclusion (reuses this seam)

## Related tests (54 pass)
- `test/unit/specs/server/util/live-loan/ads/excluded-ids.spec.js` — `parseIdList` edge cases; `fetchExcludedIds` key wiring, absent setting, read failure
- `.../ads/ads-eligibility.spec.js` — `buildAdFeedFilters` adds/omits `loanIds:{none}`; `fetchAdEligibleLoans` threads exclusions into the query
- `.../ads/google-display/google-feed.spec.js` — `generateGoogleFeed` reads the setting, logs the applied count, passes ids through

## Dev verification
1. `npx vitest run test/unit/specs/server/util/live-loan/ads/` → 54 pass.
2. Set `ui.live_loan_ads_excluded_loan_ids` to a loan id currently in the feed; regenerate `/live-loan/ads/google-feed.tsv` (wait ≤5 min or flush `google-ads-feed`); confirm that id is gone from the TSV.
3. Blank the setting; confirm the feed returns to full (no `loanIds` filter sent).

[MP-3098]: https://kiva.atlassian.net/browse/MP-3098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/util/live-loan/ads/ads-eligibility.js
+++ b/server/util/live-loan/ads/ads-eligibility.js
@@ -17,17 +17,24 @@ const THRESHOLDS = {
 	minDaysToExpiry: 3,
 };
 
+// Exclude-by-id filter fragment for the FLSS `none` operator. Returns an empty object (key omitted)
+// for an empty/absent list so we never send `{ none: [] }`. Generic by field so partner/sector
+// exclusion reuses it unchanged.
+const excludeIds = (field, ids) => (ids?.length ? { [field]: { none: ids } } : {});
+
 // FLSS filters that narrow fundraising loans down to the set eligible for the ad feed.
 // Attributes within a single FundraisingLoanSearchFilterInput object are AND-ed together,
 // while separate objects in the array are OR-ed -- so all five thresholds must live in one
-// merged object for a loan to be required to satisfy every criterion.
-export function buildAdFeedFilters() {
+// merged object for a loan to be required to satisfy every criterion. Admin-managed exclusions
+// (excludedLoanIds) are merged into the same object so excluded loans never enter the candidate set.
+export function buildAdFeedFilters({ excludedLoanIds = [] } = {}) {
 	return [{
 		partnerRiskRating: { range: { gte: THRESHOLDS.minRiskRating } },
 		lenderRepaymentTerm: { range: { lte: THRESHOLDS.maxRepaymentMonths } },
 		partnerDefaultRate: { range: { lte: THRESHOLDS.maxDefaultRate } },
 		amountLeft: { range: { gte: THRESHOLDS.minAmountLeft } },
 		daysUntilExpiration: { range: { gte: THRESHOLDS.minDaysToExpiry } },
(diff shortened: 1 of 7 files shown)


mcstover (approved):
Seems fine but I wonder how large this list might grow to be...

dyersituations (approved):

roger-in-kiva:
> Seems fine but I wonder how large this list might grow to be...

I think this depends about how the list is managed, if an already funded loan stills there should be removed as won't be fetched again... Based on [Nearly a third of loans get funded within a day of activation](https://kiva.looker.com/explore/partner_activity/vertex_fact_loan_activated?toggle=fil&qid=pxRTKLjqurcVKC1xFUuSZ2) this option probably won't be used, but worth adding for any edge case as required from legal.

PR #7158, opened 2026-08-12
feat: stop asking for goal sign ups days before goal recap start date

christian14b:
https://kiva.atlassian.net/browse/MP-2993

### DIFF
--- a/src/components/Thanks/ThanksPageSingleVersion.vue
+++ b/src/components/Thanks/ThanksPageSingleVersion.vue
@@ -208,6 +208,7 @@ import GoalSettingModal from '#src/components/MyKiva/GoalSettingModal';
 import GoalInProgress from '#src/components/Thanks/SingleVersion/GoalInProgress';
 import ExpressCheckoutModal from '#src/components/Thanks/ExpressCheckout/ExpressCheckoutModal';
 import useGoalData, { GOAL_STATUS } from '#src/composables/useGoalData';
+import { shouldHideGoalSignup } from '#src/util/goalInReview';
 import useGoalSettingRecommendedLoan, {
 	GOAL_RECOMMENDED_LOAN_ENTRYPOINT_POST_CHECKOUT,
 } from '#src/composables/useGoalSettingRecommendedLoan';
@@ -279,6 +280,10 @@ const props = defineProps({
 	goalRecommendedLoanEnable: {
 		type: Boolean,
 		default: false,
+	},
+	goalInReviewInProgressStart: {
+		type: Date,
+		default: null,
 	},
 	isExpressCheckoutModalEnabled: {
 		type: Boolean,
@@ -454,10 +459,15 @@ const showJourneyModule = computed(() => {
 	return !userGoalAchievedNow.value;
 });
 const showLoanComment = computed(() => hasPfpLoan.value || hasTeamAttributedPartnerLoan.value);
+const hideGoalSignup = computed(() => shouldHideGoalSignup({
+	recapStartDate: props.goalInReviewInProgressStart,
+}));
+
 const showGoalEntrypoint = computed(() => {
 	return !props.isGuest
 		&& goalDataInitialized.value
 		&& isEmptyGoal.value
+		&& !hideGoalSignup.value
 		&& !goalSignupThanksViewCapped.value;
 });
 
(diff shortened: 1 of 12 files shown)


src/pages/Thanks/ThanksPage.vue
```diff
@@ -290,6 +292,7 @@ export default {
 
 		this.goalRecommendedLoanEnable = readBoolSetting(data, 'general.goal_recommended_loan_enable.value') ?? false;
 		this.expressCheckoutEnabled = readBoolSetting(data, 'general.ty_page_express_checkout_enabled.value') ?? false;
+		this.goalInReviewInProgressStart = readDateSetting(data, 'general.goal_in_review_in_progress_start.value');
```
dyersituations, line 295: Make sure to add this to experiment doc.

christian14b: Yes, this is already there. It was added in the previous work

src/components/Thanks/ThanksPageSingleVersion.vue
```diff
@@ -593,12 +603,10 @@ onMounted(async () => {
...
-	goalSignupThanksViewCapped.value = !props.isGuest
-		&& isEmptyGoal.value
-		&& isGoalSignupThanksViewCapped(cookieStore);
-	if (!props.isGuest
-		&& isEmptyGoal.value
-		&& !goalSignupThanksViewCapped.value) {
+	// Gated together so the view cap only counts asks the lender actually saw.
+	const askIsAvailable = !props.isGuest && isEmptyGoal.value && !hideGoalSignup.value;
+	goalSignupThanksViewCapped.value = askIsAvailable && isGoalSignupThanksViewCapped(cookieStore);
+	if (askIsAvailable && !goalSignupThanksViewCapped.value) {
```
dyersituations: `askIsAvailable` is checked twice here, can it be simplified?

src/util/goalInReview.js
```diff
@@ -85,6 +85,33 @@ export function shouldAutoOpenRecap({
...
+ * Whether to hide the goal sign up ask, from the lead time before the recap release
+ * until the end of that goal year.
+ *
+ * @param {object} options Inputs.
+ * @param {Date|string|null} [options.recapStartDate] The goal_in_review_in_progress_start setting.
+ * @param {Date} [options.now] The effective current date.
+ * @returns {boolean} Whether to hide the sign up ask.
+ */
+export function shouldHideGoalSignup({ recapStartDate = null, now = new Date() } = {}) {
+	const recapStart = recapStartDate instanceof Date ? recapStartDate : new Date(recapStartDate ?? NaN);
```
dyersituations: new Date(recapStartDate ?? NaN) -> fallback here creates an "Invalid Date", should there be a working fallback instead?

src/pages/Thanks/ThanksPage.vue
roger-in-kiva, line 295: Is goal_in_review_in_progress_start meant to be re-set each year by ops?

src/util/goalInReview.js
```diff
@@ -85,6 +85,33 @@ export function shouldAutoOpenRecap({
...
+	if (Number.isNaN(recapStart.getTime())) {
+		return false;
+	}
+
+	const hideFrom = new Date(recapStart);
+	hideFrom.setDate(hideFrom.getDate() - SIGNUP_HIDE_LEAD_DAYS);
+	const resumeAt = new Date(recapStart.getFullYear() + 1, 0, 1);
+
+	return now.getTime() >= hideFrom.getTime() && now.getTime() < resumeAt.getTime();
```
dyersituations, line 118: Should this be `now.getTime() <= resumeAt.getTime()`?

christian14b: Good catch!, but `<` was intentional. `resumeAt` is `Jan 1 00:00:00.000 `, the moment sign ups come back, not the last moment they're hidden. With <= the ask would still be hidden at midnight on January 1 (which I guess is safe too)

src/pages/Thanks/ThanksPage.vue
christian14b: Yes, seems like a trade off for now `goal_in_review_in_progress_start` , but it updates each year.
 
Deriving the hide date from it means one date to maintain instead of two that can drift out of the intended two-week gap. And if a year gets missed it fails open: the window closes on January 1 and sign ups resume as normal, rather than staying hidden.

The alternative was a recurring month/day needing no upkeep, but then moving the date needs a deploy. Happy to switch if product would rather not depend on the annual update, but for now I think this is safe.

src/util/goalInReview.js
christian14b: The `Invalid Date` was doing the "setting not configured" job, and the early return keeps sign ups showing, which is the safe direction. But you're right! that it reads as accidental, so I've made it an explicit `if (!recapStartDate) return false`.

Applied the same to `hasInProgressReleaseStarted` above, which used the same idiom.

dyersituations (approved):

roger-in-kiva (approved):

PR #7159, opened 2026-08-12
feat: [MP-3099] exclude admin-listed partner and sector ids from the ads feed

roger-in-kiva:
## Summary
Extends the live-loan Google Ads feed with two admin-editable denylists — **partner ids** and
**sector ids** — alongside the existing loan-id exclusion (MP-3098). Both are read from
`uiConfigSetting` (`live_loan_ads_excluded_partner_ids` / `live_loan_ads_excluded_sector_ids`, edited
in KivaAdmin → Settings Manager) each time the feed regenerates, and pushed **into the FLSS
`fundraisingLoans` query** as `partnerId: { none: [...] }` / `sectorId: { none: [...] }` — so excluded
loans never enter the candidate set (no post-fetch filtering, no wasted slots against the
100-candidate cap).

- **Purely additive on the MP-3098 seam** — reuses the generic `excludeIds(field, ids)` helper and
  `fetchExcludedIds(key)` reader; `buildAdFeedFilters` gains two params, `generateGoogleFeed` reads
  the three denylists in parallel.
- **In-query exclusion**, not a post-fetch gate — `partnerId`/`sectorId` are `IntKeyFilterModifierInput`
  (same type as `loanIds`), which supports `none: [Int!]`; verified against the live schema.
- **Exclude by id, not name** — ids are stable across partner/sector renames and match how the rest of
  the codebase filters; no case/whitespace normalization.
- **Fail-open per dimension:** empty / unset / malformed setting → that filter key is omitted; a failed
  read returns `[]` + a `warn`, so the feed never errors.
- **Parsing (also hardens MP-3098):** `parseIdList` strips JSON wrapper chars before the strict
  `/^\d+$/` digit filter, because Settings Manager returns string-typed values JSON-encoded (e.g.
  `"6,16"`). Per-id strictness is unchanged — `Number()`-coercible junk (`0x1A`/`1e3`/`-7`/`5.5`)
  still can't slip in.
- **Logging:** applied id counts per lever logged each generation (excluded loans never return, so
  there's no per-loan dropped count).
- **v1 sector denylist** seeded to `6,16` (Health, Personal Use) per the spike — config not code; the
  final list is a Product/Legal/Risk call, adjustable in Settings Manager without a deploy.

## Related
- Story: [MP-3099](https://kiva.atlassian.net/browse/MP-3099) · Epic: [MP-2908](https://kiva.atlassian.net/browse/MP-2908)
- Extends: [MP-3098](https://kiva.atlassian.net/browse/MP-3098) — loan-level exclusion (#7155); this PR
  also folds in a fix to that ticket's `parseIdList` (JSON-encoded / quoted value handling)

## Related tests (64 pass)
- `.../ads/excluded-ids.spec.js` — `parseIdList` incl. JSON-encoded / quoted / array values,
  wrapper-char recovery, garbage → `[]`; `fetchExcludedIds` key wiring, absent setting, read failure
- `.../ads/ads-eligibility.spec.js` — `buildAdFeedFilters` adds/omits `partnerId`/`sectorId` `{none}`
  and merges all three; `fetchAdEligibleLoans` threads partner/sector into the query
- `.../ads/google-display/google-feed.spec.js` — `generateGoogleFeed` reads all three settings, routes
  each to its own field, logs the counts

## Dev verification
1. `npx vitest run test/unit/specs/server/util/live-loan/ads/` → 64 pass.
2. Verified live on dev FLSS: with `sectorId: { none: [6,16] }` (the configured value), the Health(6)
   loan present in the baseline candidate pool was removed and the slot backfilled; also spot-checked
   `none: [12,7]`, which erased all Food + Retail loans. `partnerId` rides the identical
   `IntKeyFilterModifierInput`/`none` path.
3. Set `ui.live_loan_ads_excluded_sector_ids` / `..._partner_ids`; regenerate
   `/live-loan/ads/google-feed.tsv` (wait ≤5 min or flush `google-ads-feed`); confirm listed
   sectors/partners are gone. Blank the settings → feed returns to full (no `sectorId`/`partnerId`
   filter sent).

[MP-3099]: https://kiva.atlassian.net/browse/MP-3099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/util/live-loan/ads/ads-eligibility.js
+++ b/server/util/live-loan/ads/ads-eligibility.js
@@ -26,8 +26,9 @@ const excludeIds = (field, ids) => (ids?.length ? { [field]: { none: ids } } : {
 // Attributes within a single FundraisingLoanSearchFilterInput object are AND-ed together,
 // while separate objects in the array are OR-ed -- so all five thresholds must live in one
 // merged object for a loan to be required to satisfy every criterion. Admin-managed exclusions
-// (excludedLoanIds) are merged into the same object so excluded loans never enter the candidate set.
-export function buildAdFeedFilters({ excludedLoanIds = [] } = {}) {
+// (loan, partner, and sector ids) are merged into the same object so excluded loans never enter
+// the candidate set.
+export function buildAdFeedFilters({ excludedLoanIds = [], excludedPartnerIds = [], excludedSectorIds = [] } = {}) {
 	return [{
 		partnerRiskRating: { range: { gte: THRESHOLDS.minRiskRating } },
 		lenderRepaymentTerm: { range: { lte: THRESHOLDS.maxRepaymentMonths } },
@@ -35,6 +36,8 @@ export function buildAdFeedFilters({ excludedLoanIds = [] } = {}) {
 		amountLeft: { range: { gte: THRESHOLDS.minAmountLeft } },
 		daysUntilExpiration: { range: { gte: THRESHOLDS.minDaysToExpiry } },
 		...excludeIds('loanIds', excludedLoanIds),
+		...excludeIds('partnerId', excludedPartnerIds),
+		...excludeIds('sectorId', excludedSectorIds),
 	}];
 }
 
--- a/server/util/live-loan/ads/ads-eligibility.js
+++ b/server/util/live-loan/ads/ads-eligibility.js
@@ -18,26 +18,27 @@ const THRESHOLDS = {
 };
 
 // Exclude-by-id filter fragment for the FLSS `none` operator. Returns an empty object (key omitted)
-// for an empty/absent list so we never send `{ none: [] }`. Generic by field so partner/sector
-// exclusion reuses it unchanged.
+// for an empty/absent list so we never send `{ none: [] }`.
 const excludeIds = (field, ids) => (ids?.length ? { [field]: { none: ids } } : {});
 
 // FLSS filters that narrow fundraising loans down to the set eligible for the ad feed.
 // Attributes within a single FundraisingLoanSearchFilterInput object are AND-ed together,
 // while separate objects in the array are OR-ed -- so all five thresholds must live in one
 // merged object for a loan to be required to satisfy every criterion. Admin-managed exclusions
-// (loan, partner, and sector ids) are merged into the same object so excluded loans never enter
-// the candidate set.
-export function buildAdFeedFilters({ excludedLoanIds = [], excludedPartnerIds = [], excludedSectorIds = [] } = {}) {
+// arrive as a map of FLSS filter field -> ids (e.g. { loanIds, partnerId, sectorId }) and are merged
+// into the same object as `none` fragments -- so a new exclusion dimension is one more map entry at
+// the call site, not a change to this signature.
+export function buildAdFeedFilters(exclusions = {}) {
 	return [{
 		partnerRiskRating: { range: { gte: THRESHOLDS.minRiskRating } },
 		lenderRepaymentTerm: { range: { lte: THRESHOLDS.maxRepaymentMonths } },
 		partnerDefaultRate: { range: { lte: THRESHOLDS.maxDefaultRate } },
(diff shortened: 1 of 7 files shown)


server/util/live-loan/ads/google-display/google-feed.js
```diff
@@ -26,15 +30,24 @@ export function toTsv(rows, columns = FEED_COLUMNS) {
...
 export async function generateGoogleFeed(count) {
-	// Admin-managed denylist (Settings Manager), read fresh each generation; excluded loans are pushed
-	// into the FLSS query so they never enter the candidate set. Absent/failed read => no exclusions.
-	const excludedLoanIds = await fetchExcludedIds(EXCLUDED_LOAN_IDS_SETTING_KEY);
-	info(`Ad feed: applying ${excludedLoanIds.length} excluded loan id(s)`, { excludedLoanIds });
+	// Admin-managed denylists (Settings Manager), read fresh each generation; excluded loans, partners,
+	// and sectors are pushed into the FLSS query so they never enter the candidate set. The three reads
+	// are independent, so they run in parallel. Absent/failed read => no exclusions for that dimension.
+	const [excludedLoanIds, excludedPartnerIds, excludedSectorIds] = await Promise.all([
+		fetchExcludedIds(EXCLUDED_LOAN_IDS_SETTING_KEY),
```
dyersituations, line 37: How much would performance improve if these were a single graphql query?

dyersituations (approved):

christian14b (approved):

server/util/live-loan/ads/ads-eligibility.js
```diff
@@ -26,15 +26,18 @@ const excludeIds = (field, ids) => (ids?.length ? { [field]: { none: ids } } : {
...
+// (loan, partner, and sector ids) are merged into the same object so excluded loans never enter
+// the candidate set.
+export function buildAdFeedFilters({ excludedLoanIds = [], excludedPartnerIds = [], excludedSectorIds = [] } = {}) {
 	return [{
 		partnerRiskRating: { range: { gte: THRESHOLDS.minRiskRating } },
 		lenderRepaymentTerm: { range: { lte: THRESHOLDS.maxRepaymentMonths } },
 		partnerDefaultRate: { range: { lte: THRESHOLDS.maxDefaultRate } },
 		amountLeft: { range: { gte: THRESHOLDS.minAmountLeft } },
 		daysUntilExpiration: { range: { gte: THRESHOLDS.minDaysToExpiry } },
 		...excludeIds('loanIds', excludedLoanIds),
```
christian14b: Could be good to think later how this can be improved to only consider a single variable to avoid extending the method's arguments

server/util/live-loan/ads/google-display/google-feed.js
roger-in-kiva: Barely any, honestly. They already run in parallel (`Promise.all`), so we're paying about one round-trip, not three. Combining them into a single aliased query would drop 2 of the 3 requests — but since they already overlap, it wouldn't actually change the latency.

It also runs rarely: the feed sits behind a 5-min cache, so this only fires ~once every 5 minutes per worker, not per request. So we're talking a couple of extra background calls every few minutes, none of it on a user's path.

I split them on purpose, though — each read fails open on its own, so if one setting is missing or malformed we just skip that one exclusion and still build the feed. A single combined query ties their fate together. Happy to batch it if you'd rather cut the round-trips, but it won't buy us any real speed.

server/util/live-loan/ads/ads-eligibility.js
roger-in-kiva: Good call — done. Made it a single `{ field: ids }` map that `buildAdFeedFilters` just iterates over, so adding a new exclusion dimension is one more entry in the map `generateGoogleFeed` builds, with no change to the signature. Pushed in `381c4df98`.

dyersituations (approved):

PR #7165, opened 2026-08-13
fix: user has goal validation was thought to check null value instead…

christian14b:
… of empty object

Goal Card was still present days before releasing goal recap because the check was failing to recognize an empty goal (`{}`). 

## Testing

In a no goal account go to `https://kiva-ui.local/mykiva?recapDate=2027-01-01`, you should not see the goal card in next steps row since `goal_in_review_in_progress_start` is set to `2026-11-15` in Settings manager

### DIFF
--- a/src/components/MyKiva/JourneyCardCarousel.vue
+++ b/src/components/MyKiva/JourneyCardCarousel.vue
@@ -267,6 +267,8 @@ const showSurveyCard = computed(() => props.showSurveySlide && checkShowSurveyCa
 
 const nonBadgesSlides = computed(() => filterNonBadgesSlides(props.slides));
 
+const userHasGoal = computed(() => !!props.userGoal && Object.keys(props.userGoal).length > 0);
+
 const hideGoalSignup = computed(() => shouldHideGoalSignup({
 	recapStartDate: props.goalInReviewInProgressStart,
 	now: getGoalInReviewNow(),
@@ -275,7 +277,7 @@ const hideGoalSignup = computed(() => shouldHideGoalSignup({
 const shouldShowGoalCard = computed(() => {
 	if (!props.inLendingStats) return false;
 	// With no goal, this card is only the sign up ask.
-	if (!props.userGoal && hideGoalSignup.value) return false;
+	if (!userHasGoal.value && hideGoalSignup.value) return false;
 
 	return (!props.userGoal || !props.userGoalAchieved || props.userGoalAchieved) && !props.hideGoalCard;
 });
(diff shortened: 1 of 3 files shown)


roger-in-kiva (approved):

src/components/MyKiva/JourneyCardCarousel.vue
```diff
@@ -267,6 +267,8 @@ const showSurveyCard = computed(() => props.showSurveySlide && checkShowSurveyCa
 
 const nonBadgesSlides = computed(() => filterNonBadgesSlides(props.slides));
 
+const userHasGoal = computed(() => !!props.userGoal && Object.keys(props.userGoal).length > 0);
```
dyersituations: Why would there be a goal with `{}`?

cristhianDt (approved):

src/components/MyKiva/JourneyCardCarousel.vue
christian14b: It's happening in useGoalData composable https://github.com/kiva/ui/blob/76d04d750bd679002149b539921088d916ca5157/src/composables/useGoalData.js#L154

When a lender has no goals, `activeGoals` is empty, making `activeGoals[0]` `undefined`, then spreading `undefined` yields `{}` rather than throwing.

dyersituations: Will this cause issues anywhere else? Seems like `userHasGoal` might need to be centralized in the composable for reuse?

christian14b: It's not causing issues on the different entrypoints because all of them handles that use case, so yeah, worth making a centralized computed for this. I'll add it in this work

src/composables/useGoalData.js
```diff
@@ -76,6 +76,17 @@ export const HALF_GOAL_THRESHOLD = 50;
...
 const RECOMMENDED_LOANS_LIMIT = 4;
 
+/**
+ * A lender with no goal gets `{}` rather than null, since setGoalState spreads an
+ * absent goal. Truthiness is therefore never enough to tell whether a goal exists.
+ *
+ * @param {object} goal The lender's goal.
+ * @returns {boolean} Whether the lender has a goal.
+ */
+export function hasGoal(goal) {
```
dyersituations: In other places we use a method exported from the composable instead of the file. Why the difference in this composable?

christian14b: Good catch! moved it!

dyersituations (approved):

PR #7168, opened 2026-08-13
fix: temp view changes for disaster relief giving fund on mykiva and portfolio

dyersituations:
https://kiva.atlassian.net/browse/MP-3121

- Hide Giving Funds MyKiva card if user only participated in relief fund
- Update Portfolio card to be custom for relief fund if user only participated in that fund

General idea is to reduce confusion for users who only contributed to the Colombia relief fund.

### DIFF
--- a/src/components/GivingFunds/MyGivingFundsCard.vue
+++ b/src/components/GivingFunds/MyGivingFundsCard.vue
@@ -1,6 +1,5 @@
 <template>
 	<div
-		v-if="showCard"
 		class="giving-fund-card tw-rounded tw-bg-white tw-p-2 tw-w-full tw-flex tw-flex-col tw-gap-2 tw-items-stretch
 			md:tw-flex-row md:tw-items-center md:tw-justify-between md:tw-gap-3 md:tw-p-2.5"
 	>
@@ -15,7 +14,6 @@
 			</transition>
 		</div>
 		<KvButton
-			:key="`giving-funds-cta-${isDisasterReliefOnly}`"
 			class="tw-w-full md:tw-w-auto md:tw-ml-auto"
 			variant="primary"
 			:to="ctaTo"
@@ -43,7 +41,6 @@ import {
 	computed,
 	inject,
 	ref,
-	watch,
 	onMounted,
 } from 'vue';
 
@@ -57,10 +54,10 @@ const props = defineProps({
 		default: null,
 	},
 	/**
-	 * When true (MyKiva), the card is not rendered at all for lenders whose only
-	 * giving fund activity is supporting the disaster relief fund
+	 * Whether the lender's only giving fund activity is supporting the disaster
+	 * relief fund; renders the fund-specific variant of the card
 	 */
-	hideDisasterReliefOnly: {
+	isDisasterReliefOnly: {
 		type: Boolean,
 		default: false,
 	},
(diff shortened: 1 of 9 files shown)


src/components/GivingFunds/MyGivingFundsCard.vue
```diff
@@ -81,16 +131,32 @@ const textCopy = computed(() => {
 	return copy;
 });
 
+let viewEventTracked = false;
+
+watch(isDisasterReliefOnly, value => {
+	// Track views of the disaster relief variant, which only renders when not hidden (MP-3121)
```
roger-in-kiva: do we want ticket id's?

src/components/GivingFunds/MyGivingFundsCard.vue
```diff
@@ -59,10 +72,47 @@ const {
...
+
+const dataLoaded = computed(() => fundsCountLoaded.value && contributedFundsLoaded.value);
+
+// Lender supports the disaster relief fund and no other giving funds
+const isDisasterReliefOnly = computed(() => dataLoaded.value
+	&& myFundsCount.value === 0
+	&& contributedFundIds.value.length === 1
+	&& contributedFundIds.value[0] === givingFundIds.COLOMBIA_DISASTER_RELIEF);
+
+const showCard = computed(() => {
```
mcstover: We shouldn't need this as the visibility of this component is based on server data...

mcstover: We use any giving fund participation which would include donations to a fund they don't own to determine when to `// show giving funds card if user has any giving fund participation`

src/components/GivingFunds/MyGivingFundsCard.vue
```diff
@@ -35,22 +37,33 @@
...
 	userId: {
 		type: [String, Number],
 		required: false,
 		default: null,
 	},
+	/**
+	 * When true (MyKiva), the card is not rendered at all for lenders whose only
+	 * giving fund activity is supporting the disaster relief fund
+	 */
+	hideDisasterReliefOnly: {
```
mcstover: I think this should just be used in the parent check in combination with the participation driven v-if. Of course, I don't see anything that would set this prop so am not sure why it's needed at all unless there are future requirements...

dyersituations: This is being passed in from MyKiva, where we don't want to show this card if the user only lent to the disaster relief fund: https://github.com/kiva/ui/pull/7168/changes/BASE..a8d5cf9bfd7e832cbea422cffb4dc515d9ce7f96#diff-d1e1c50ddac2c868077f1e04eb94477caf0015ec64df8320519c7cb6b74c2734R19

src/components/GivingFunds/MyGivingFundsCard.vue
dyersituations: Fair, moving some of this logic into the parent for simplicity.

src/components/GivingFunds/MyGivingFundsCard.vue
dyersituations: Oops, missed a couple, commit incoming.

roger-in-kiva (approved):

src/components/GivingFunds/MyGivingFundsCard.vue
dyersituations: Looking into alternatives means fairly similar complexity. The situation this covers is when a user has only participated towards the relief fund -> we don't want to display the card on MyKiva. Alternatives would include pulling participation data in both places to toggle on MyKiva or passing that data down from both MyKiva and Portfolio pages. The solution here already implemented seems like a fair alternative, thoughts?

dyersituations: There is a way to move this visibility logic to the parent with minimal extra data query (adding to `myKiva.graphql`), but it still does involve adding extra fields to that query. The solution here has the minimal extra data pulling for the logic needed.

mcstover: Ok I think this will have to do. Does it still server renders for standard giving fund users and isn't delayed until for them? ie the component doesn't pop in?

src/components/GivingFunds/MyGivingFundsCard.vue
mcstover: Why would we hide the direct link on mykiva if they donated to this fund?

dyersituations: Request from product, since the giving funds card on MyKiva doesn't give much context as currently built. They thought it would be confusing if that user only participated in the relief fund.

src/components/GivingFunds/MyGivingFundsCard.vue
dyersituations: There would be pop-in due to the participation details loading on mount. So I suppose less ideal. I'll move this server side in the morning, it will just require more data being pulled by the preFetch query. My focus was on reducing extra data querying on a busy page, but pop-in is probably more user noticed, so good call...

mcstover: Ok yea we should either fix the server render or ditch all the server fetching and checking id this defeats it.

dyersituations: @mcstover added the needed fields to support SSR, tested locally with a couple account situations, all should be good to go.

mcstover (approved):
Thank you for shoring this up and adding the new variant!

roger-in-kiva (approved):

PR #7170, opened 2026-08-14
fix: mp-3058-meta-mykiva-event-removal

infante-jaime:
### Description

MyKiva, portfolio and admin pages are excluded from marketing tags in GTM by policy — we don't load PII-scraping tags in those contexts. That means there's no Meta pixel on them, and any event we fire there is dropped silently before it leaves the browser.

Two of our events were in that position, so they're removed rather than left looking functional:

Account Created. Registrations redirect to MyKiva or portfolio, which is exactly where the pixel isn't. Gone, along with the cookie that deduped it.

Email Sign Up on MyKiva. The card's opt-in worked, but the tracking never arrived. The call came out of the shared useOptIn composable, so the composable now just reports whether the settings applied and leaves reporting to callers.

The thanks-page opt-in keeps its event — the pixel is loaded there — so it moved to the call site in OptInModule, gated on the mutation actually applying. Guest checkout is untouched.

Net: nothing that could fire still fires, and nothing that couldn't is left in the code pretending otherwise.


infante-jaime:
Worth noting this fixes the cookie bug but won't make the event fire in practice — MyKiva and portfolio have no marketing tags by design, and that's where registrations land. Follow-up idea: read memberSince at boot and report on the first page where the pixel is available, deduped by the same cookie.

Thoughts @dyersituations @mcstover

mcstover:
> Worth noting this fixes the cookie bug but won't make the event fire in practice — MyKiva and portfolio have no marketing tags by design, and that's where registrations land. Follow-up idea: read memberSince at boot and report on the first page where the pixel is available, deduped by the same cookie.
> 
> Thoughts @dyersituations @mcstover

When was this cookie getting exercised before? We should find a way to check and resolve this event outside of protected areas fo the site. Basically something like your follow up idea and remove the cookie after firing the event.

infante-jaime:
> > Worth noting this fixes the cookie bug but won't make the event fire in practice — MyKiva and portfolio have no marketing tags by design, and that's where registrations land. Follow-up idea: read memberSince at boot and report on the first page where the pixel is available, deduped by the same cookie.
> > Thoughts @dyersituations @mcstover
> 
> When was this cookie getting exercised before? We should find a way to check and resolve this event outside of protected areas fo the site. Basically something like your follow up idea and remove the cookie after firing the event.

The cookie's only job is deduping the pixel event — it came in with the previous PR, and nothing else reads it. The follow-up keeps that logic: pull memberSince at boot, use it to recognise the registration, fire the event on the first page where the pixel is actually available, then set the cookie for an hour so it can't report twice.

This PR should also roll back the MyKiva new-registered-user tracking and reshape the cookie to match that approach, so we're not carrying code that can't fire.

src/client-entry.js
```diff
@@ -129,9 +129,13 @@ async function setupAnalytics(app, apolloClient, cookieStore) {
 	const { trackAccountCreated } = await import('#src/util/registrationTracking');
 	trackAccountCreated(userId, {
 		get: name => cookieStore.get(name),
-		// An hour outlives the monolith's re-authentication redirect, which is all the
-		// dedup window needs to cover
-		set: (name, value) => cookieStore.set(name, value, { path: '/', expires: 1 / 24 }),
+		// An hour outlives the monolith's re-authentication redirect, which is all the dedup
+		// window needs to cover. Must be a Date — cookieStore serializes through `cookie`,
+		// which rejects a number outright.
+		set: (name, value) => cookieStore.set(name, value, {
```
dyersituations: We might need to push back and say we cannot support this event for now. Let's talk about it next week in our team meeting. This PR can remain open for now as a draft. We might need to rollback this event in general. I forgot about MyKiva not loading the pixel when I reviewed this original PR...

dyersituations (approved):

PR #7172, opened 2026-08-14
fix: category name displayed and small design color in stats for goal…

christian14b:
… in review

Categories related to goal are now consistent (Goal in review and Choose Category on goal creation)

image

### DIFF
--- a/src/composables/useGoalData.js
+++ b/src/composables/useGoalData.js
@@ -35,15 +35,6 @@ import usEntrepreneursImg from '#src/assets/images/my-kiva/goal-setting/us-entre
 import basicNeedsImg from '#src/assets/images/my-kiva/goal-setting/basic-needs.svg?url';
 import supportAllImg from '#src/assets/images/my-kiva/goal-setting/support-all.svg?url';
 
-export const CATEGORY_NAMES = {
-	[ID_WOMENS_EQUALITY]: 'Women',
-	[ID_SUPPORT_ALL]: 'Choose as I go',
-	[ID_CLIMATE_ACTION]: 'Climate Action',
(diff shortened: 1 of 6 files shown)


src/util/goalInReview.js
```diff
@@ -208,6 +209,10 @@ export function getCategoryName(category, contentfulEntries = []) {
 		return '';
 	}
 
+	if (CATEGORY_NAMES[category]) {
```
roger-in-kiva: Is there some case this could be `false`?

dyersituations: Also lots of errors on this page here in GH.

src/composables/useGoalData.js
```diff
@@ -35,6 +35,15 @@ import usEntrepreneursImg from '#src/assets/images/my-kiva/goal-setting/us-entre
 import basicNeedsImg from '#src/assets/images/my-kiva/goal-setting/basic-needs.svg?url';
 import supportAllImg from '#src/assets/images/my-kiva/goal-setting/support-all.svg?url';
 
+export const CATEGORY_NAMES = {
```
dyersituations: Why was this new object needed? It looks like we're returning the same strings below.

christian14b: You're right, it's unnecessary.. I'm cleaning up this new code

roger-in-kiva (approved):

cristhianDt (approved):

src/util/goalInReview.js
christian14b: it's only used under goalInReview context, so unlikely to be false, either way will fallback to an empty string if

dyersituations (approved):

PR #7175, opened 2026-08-18
fix: scope recap data to the goal's year

christian14b:
https://kiva.atlassian.net/browse/MP-3135

Filter loan data from different year than the user goal

### DIFF
--- a/src/util/goalInReview.js
+++ b/src/util/goalInReview.js
@@ -269,8 +269,8 @@ function getYearPurchases(goalSummary, tieredLendingAchievements, year) {
 	// mean pulling the whole year, and is not even reachable for lenders past the
 	// rolling window, which retains only the most recent loans.
 	return (getGoalAchievement(goalSummary, tieredLendingAchievements)?.loanPurchases ?? [])
-		.filter(purchase => purchase?.loan)
-		.filter(purchase => !Number.isFinite(year) || getPurchaseYear(purchase) === year);
+		.filter(purchase => purchase?.loan
+			&& (!Number.isFinite(year) || getPurchaseYear(purchase) === year));
 }
 
 function getYearLoans(goalSummary, tieredLendingAchievements, year) {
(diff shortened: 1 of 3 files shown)


src/util/goalInReview.js
```diff
@@ -256,16 +256,25 @@ function getGoalAchievement(goalSummary, tieredLendingAchievements) {
...
+function getYearPurchases(goalSummary, tieredLendingAchievements, year) {
+	// The year the recap asks the service for scopes progressForYear, not this list, so it
+	// arrives carrying other years.
 	// Most recent first, as the service returns them. Selecting the oldest instead would
 	// mean pulling the whole year, and is not even reachable for lenders past the
 	// rolling window, which retains only the most recent loans.
 	return (getGoalAchievement(goalSummary, tieredLendingAchievements)?.loanPurchases ?? [])
-		.filter(purchase => purchase?.loan);
+		.filter(purchase => purchase?.loan)
+		.filter(purchase => !Number.isFinite(year) || getPurchaseYear(purchase) === year);
```
dyersituations: These filters can be combined?

src/util/goalInReview.js
```diff
@@ -327,17 +336,20 @@ function getGoalSectors(loans) {
...
 	}
 
-	const yearPurchases = getYearPurchases(goalSummary, tieredLendingAchievements);
+	const yearPurchases = getYearPurchases(goalSummary, tieredLendingAchievements, year);
 	const achievement = getGoalAchievement(goalSummary, tieredLendingAchievements);
 	// progressForYear is authoritative: the rolling window can trim loanPurchases.
-	const lent = Number(achievement?.progressForYear) || yearPurchases.length;
+	// Finite rather than truthy, so a real zero is not read as missing.
+	const progress = achievement?.progressForYear;
+	const lent = Number.isFinite(progress) ? progress : yearPurchases.length;
```
dyersituations, line 354: Isn't yearPurchases what we always want here?

christian14b: Good question, and you're right that they're equal in practice. I checked the super lender across all five categories: `progressForYear` matches the in-year purchase count exactly every time.

They can only come apart on a past-year recap, and it's the newest-first ordering that decides it. `loanPurchasesLimit` is the goal target, and the service fills that request newest first. On a current-year recap this year's loans are the newest, so they're served first and filtering can never lose any of them. The two numbers always agree.

On a past-year recap the following year's loans sit above the year you asked for and consume the request first. Say a lender opens their 2025 recap in February 2026 having already made 20 loans in 2026. Target 63, so we ask for 63 purchases: 20 come back from 2026 and 43 from 2025. But they made 100 loans in 2025.

- progressForYear: min(100, 63) = 63, so 100% correct, they finished it
- yearPurchases.length: min(43, 63) = 43, so 68% on a goal they completed

That's hypothetical.. its biggest year is 50 against a request of 60. So this guard is just defensive.

src/util/goalInReview.js
christian14b: good call!

dyersituations (approved):

cristhianDt (approved):

PR #7177, opened 2026-08-19
feat(live-loan-ads): align the google feed with the merchant center product spec

roger-in-kiva:
## What & why
MP-3136 — adjusts the live-loan Google ads feed to conform to the [Google Merchant Center product data spec](https://support.google.com/merchants/answer/7052112) (the previous TSV used Google Ads dynamic-display attribute names Merchant Center doesn't recognize).

### Feed shape
- **Columns renamed** to Merchant product attributes: `id, title, description, google_product_category, image_link, link` (dropped `Item subtitle`).
- **Added** `price` (`25.00 USD`), `availability` (`in_stock`), `identifier_exists` (`no`). All 7 always-required attributes are now present; `identifier_exists=no` covers the brand/GTIN/MPN requirement (loans have no product identifiers).

### Copy
- **`description`** is now the borrower's own loan-use text, formatted into a sentence (`This loan is special because …`, mirroring `formatWhySpecial`), falling back to the sector name when the use text is absent or would exceed the 5000-char limit.
- **Ad-safety** now scans the **raw borrower inputs** (first name, loan-use, sector) rather than the composed row, so authored template words (e.g. "loan" in the lead-in) don't drop every row.

### Images
- **Served from Kiva's own WebP CDN** (`…/img/w1200h1200/<hash>.webp`), dropping the Cloudinary transform — WebP is Merchant-accepted and needs no ICC step. Prod host is hardcoded (like `link`) so Google always gets the live image; `w1200h1200` caps large originals without upscaling.
- **500x500 eligibility gate** — loans whose image is smaller on either side are dropped (Merchant minimum; we don't upscale).

### Volume
- **Candidate pull raised 100 -> 200** (the FLSS single-page maximum) to backfill after the eligibility/ad-safety drops.

## Testing
- All live-loan-ads unit specs green (109) + live-loan-router (35); full suite shows only the pre-existing `possibleTypesCacheIntegrity` flake.
- Verified against the dev gateway and Fastly: WebP delivery + extension match, the size gate drops ~12% (88 kept), and FLSS honors `limit:200` (errors at 201).

## Known / out of scope
- `google_product_category` currently carries the free-text sector name (Google expects its taxonomy; likely ignored, harmless).
- The banned-word ad-safety list may over-drop loan-use text containing `buy`/`invest`/`earn` — accepted for now.

### DIFF
--- a/server/util/live-loan/ads/google-display/feed-row.js
+++ b/server/util/live-loan/ads/google-display/feed-row.js
@@ -1,12 +1,23 @@
 import { primaryFirstName } from '../ads-eligibility.js';
-import { KIVA_PROD_HOST, CLOUDINARY_AD_IMAGE_BASE, CLOUDINARY_AD_IMAGE_TRANSFORM } from '../constants.js';
+import { KIVA_PROD_HOST } from '../constants.js';
 
 const UTM_QUERY = 'utm_medium=paid&utm_source=google&utm_campaign=liveloans';
 const KIVA_LEND_BASE = `${KIVA_PROD_HOST}/lend`;
-// Cloudinary's `remote/` mapping fetches the loan's stored original at <hash>.jpg to transform on the fly.
-const AD_IMAGE_BASE = `${CLOUDINARY_AD_IMAGE_BASE}/${CLOUDINARY_AD_IMAGE_TRANSFORM}/remote`;
-// Google caps Item title / subtitle / description at 25 characters.
-const CAP = 25;
+// Ad image: Kiva's own image CDN in WebP at a fixed size. WebP is a Google-Merchant-accepted format and
+// the `.webp` extension matches the served bytes, so no image-conversion step is needed. The prod host is
+// hardcoded (not env-derived) so Google is always served the live image, even from a non-prod feed run.
+const AD_IMAGE_SIZE = 'w1200h1200';
+const KIVA_IMAGE_BASE = `${KIVA_PROD_HOST}/img/${AD_IMAGE_SIZE}`;
+// Google Merchant Center max lengths for the title and description fields.
+const TITLE_MAX = 150;
+const DESCRIPTION_MAX = 5000;
+
+// Static Google Merchant attributes every row carries. A Kiva loan is not a purchasable good, so price
+// and availability are fixed placeholders, and identifier_exists=no tells Google the item intentionally
+// has no brand/GTIN/MPN (which it otherwise requires on new products).
+const DEFAULT_PRICE = '25.00 USD';
+const DEFAULT_AVAILABILITY = 'in_stock';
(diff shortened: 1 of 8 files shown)


server/util/live-loan/ads/google-display/feed-row.js
```diff
@@ -1,12 +1,23 @@
...
 const KIVA_LEND_BASE = `${KIVA_PROD_HOST}/lend`;
-// Cloudinary's `remote/` mapping fetches the loan's stored original at <hash>.jpg to transform on the fly.
-const AD_IMAGE_BASE = `${CLOUDINARY_AD_IMAGE_BASE}/${CLOUDINARY_AD_IMAGE_TRANSFORM}/remote`;
-// Google caps Item title / subtitle / description at 25 characters.
-const CAP = 25;
+// Ad image: Kiva's own image CDN in WebP at a fixed size. WebP is a Google-Merchant-accepted format and
+// the `.webp` extension matches the served bytes, so no image-conversion step is needed. The prod host is
+// hardcoded (not env-derived) so Google is always served the live image, even from a non-prod feed run.
+const AD_IMAGE_SIZE = 'w1200h1200';
+const KIVA_IMAGE_BASE = `${KIVA_PROD_HOST}/img/${AD_IMAGE_SIZE}`;
```
dyersituations, line 10: Are these images still served by Cloudinary?

roger-in-kiva: No, this switches to Kiva's own image CDN served by our Fastly + media-service.
 In terms of ads is better as we are pointing to kiva's url instead of Cloudinary.

dyersituations: Why was this not possible before?

roger-in-kiva: B/C new documentation allows us to use webp image type.
Past spike documentation required to converts to sRGB, and force-embeds an sRGB ICC profile then finally use JPG image type.
This is a simpler approach as already cached and not draining Cloudinary requests.

dyersituations: Nice, can we remove the ICC profile work?

roger-in-kiva: That part was on the fly, and the main reason why we used Cloudinary paths:
`https://res.cloudinary.com/kiva/c_limit,w_1200,h_1200,f_jpg,cs_srgb,fl_force_icc/remote/5c258e1d730889cdb1af37be1196ba00.jpg`

These specs basically: `f_jpg,cs_srgb,fl_force_icc`.

So, nothing to remove, transparent.

dyersituations (approved):

PR #7179, opened 2026-08-20
fix: make KvWwwHeaderBasic the default global header

mcstover:
Stage 1 of the global header experiment cleanup.

## What changed

`TheHeader.vue` no longer reads the `home_page` experiment. A new `showBasicHeader` computed (`!minimal && !corporate`) picks the header, making `KvWwwHeaderBasic` the default.

**Precedence is inverted.** The experiment gate previously sat above the minimal and corporate branches, so a variant-`b` user got the basic header even on `/cc/*` pages. Those pathways now win over the basic header, which is the intended behavior.

The experiment's Apollo operation, `trackExperimentVersion` call, key constant, imports, and data flag are **commented out rather than deleted** — this wiring gets reused for future header experiments. An anchor comment above the constant names all five sites so they can be restored together.

The legacy default header markup stays in the file but is now unreachable, marked for stage 2 deletion. Removing it, and removing the Fastly `forceHeader` plumbing, are follow-up PRs.

## Testing

`TheHeader.spec.js` is rewritten as three pathway tests (default / minimal / corporate) with `KvWwwHeaderBasic` stubbed. The old test asserted on the legacy `SearchBar`, which no longer renders on the default path.

Full suite: 328 files, 5114 tests passing. `npm run lint`: 0 errors.

Manually verified on `/`, `/lend`, a `/cc/*` corporate route, logged-in and logged-out, and the `WwwPageMinimal` Storybook story (no route reaches the minimal pathway).

## Note for reviewers

The `home_page` experiment stays active server-side; this only stops `ui` from reading it. Until `cms-page-server` makes the same flip, a variant-`a` user could see the new header on `ui` routes and the legacy one on CPS routes in the same session. Ramping `home_page` to 100% `b` before this merges would close that window.

Spec: `~/kiva/docs/ui-docs/specs/2026-08-20-global-header-experiment-cleanup-design.md`
Plan: `~/kiva/docs/ui-docs/plans/2026-08-20-global-header-default-stage-1.md`

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -4,7 +4,7 @@
 		:class="isInExperimentPages ? 'tw-sticky tw-top-0 tw-z-sticky' : ''"
 	>
 		<KvWwwHeaderBasic
-			v-if="isNavUpdateExp"
+			v-if="showBasicHeader"
 			class="tw-bg-primary tw-border-b tw-border-tertiary tw-relative"
 			ref="newExpHeader"
 			:logged-in="!isVisitor"
@@ -104,7 +104,12 @@
 					</div>
 				</template>
 
-				<!-- Default Header -->
+				<!--
+					Default Header — UNREACHABLE as of the stage 1 header default flip.
+					The enclosing <nav> only renders when `minimal` or `corporate` is set,
+					so this `v-else` can never match. Retained for stage 2 deletion; see
+					docs/ui-docs/specs/2026-08-20-global-header-experiment-cleanup-design.md
+				-->
 				<template v-else>
 					<div
 						class="header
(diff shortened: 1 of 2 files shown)


src/components/WwwFrame/TheHeader.vue
```diff
@@ -104,7 +104,12 @@
 					</div>
 				</template>
 
-				<!-- Default Header -->
+				<!--
+					Default Header — UNREACHABLE as of the stage 1 header default flip.
+					The enclosing <nav> only renders when `minimal` or `corporate` is set,
```
dyersituations, line 109: How come never of these could be set?

mcstover:
Same as the CPS pr, just force the new header to always show with cleanup in follow up tickets.

src/components/WwwFrame/TheHeader.vue
mcstover: This comment isn't very clear. It's just saying the code below isn't operable any longer b/c the `showBasicHeader` property only flips if minimal or corporate are set and the v-ifs above would capture those cases...

dyersituations (approved):
Seems fine with follow-up cleanup.

PR #7180, opened 2026-08-20
feat: [mp-3091] - vqa goal in review

cristhianDt:
# MP-3091

## Summary

Several test cases were run; you can find them in the attached .md file. This pull request fixes some minor bugs.

https://kiva.atlassian.net/browse/MP-3091

## Results

[test-plan.md](https://github.com/user-attachments/files/31284206/test-plan.md)

## Evidences / videos

https://github.com/user-attachments/assets/eb999881-016b-40a0-9845-9ad34611d545

Captura de pantalla 2026-08-20 a la(s) 15 36 30

### DIFF
--- a/src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
+++ b/src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
@@ -59,7 +59,7 @@
 					:year="data?.year"
 					:current-year="currentYear"
 					:feedback-submitted="feedbackSubmitted"
-					@back-to-kiva="handleCta('back-to-kiva')"
+					@goal-recap-back-to-kiva="handleCta('goal-recap-back-to-kiva')"
 					@finish-goal="handleCta('finish-goal')"
 					@set-goal="handleCta('set-goal')"
 					@feedback-submitted="handleFeedbackSubmitted"
@@ -120,7 +120,7 @@ const props = defineProps({
 	},
 });
 
-const emit = defineEmits(['close', 'back-to-kiva', 'finish-goal', 'set-goal', 'feedback-submitted']);
+const emit = defineEmits(['close', 'goal-recap-back-to-kiva', 'finish-goal', 'set-goal', 'feedback-submitted']);
 const $kvTrackEvent = inject('$kvTrackEvent', () => {});
 
 // Single source of truth for "now". Add ?recapDate=YYYY-MM-DD to the url for QA specific dates
@@ -147,7 +147,6 @@ const handleCta = event => {
 		$kvTrackEvent('portfolio', 'click', `goal-recap-finish-my-${props.data?.year}-goal`);
 	} else {
 		$kvTrackEvent('portfolio', 'click', 'goal-recap-back-to-kiva');
-		emit('close');
 	}
 	emit(event);
 };
--- a/src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
+++ b/src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
@@ -252,8 +252,8 @@ onBeforeUnmount(teardownSlideObserver);
 
 	.goal-in-review-slides > :first-child > * {
 		@screen md {
-		min-height: var(--recap-page-height);
-	}
+			min-height: var(--recap-page-height);
+		}
 	}
 
 	[data-test=kv-lightbox] {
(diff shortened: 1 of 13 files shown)


src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
```diff
@@ -147,6 +147,7 @@ const handleCta = event => {
 		$kvTrackEvent('portfolio', 'click', `goal-recap-finish-my-${props.data?.year}-goal`);
 	} else {
 		$kvTrackEvent('portfolio', 'click', 'goal-recap-back-to-kiva');
+		emit('close');
```
christian14b: There's an emit after this. Shouldn't we close the modal in the parent logic instead when back to kiva event is emitted?

src/composables/useGoalInReview.js
```diff
@@ -166,6 +167,22 @@ export default function useGoalInReview({ apollo, cookieStore, goalData } = {})
...
+	 *
+	 * @param {object} router Vue router instance (getCtaHref reads the current route).
+	 * @returns {string|null} The loan-finding href, or null before the recap loads.
+	 */
+	function getFinishGoalHref(router) {
+		const summary = goalInReviewData.value?.goalSummary;
+		if (!summary?.category) {
+			return null;
+		}
+		return getCtaHref(summary.target, summary.category, router, summary.count ?? 0);
```
christian14b: I think we can move the above validations to this method instead so every component that use it can be benefited

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
```diff
@@ -187,6 +191,17 @@ export default {
 		async handleGoalInReviewFeedbackSubmitted() {
 			await this.setGoalFeedbackSubmittedPreference(this.goalInReviewData?.year);
 		},
+		// "Finish my goal" routes to the goal category's loan-finding page, the same
+		// destination as the goal cards' continue CTA (tracking fires in the modal).
+		handleGoalInReviewFinishGoal() {
```
christian14b, line 196: This method is duplicated in my kiva above, Can we move it to the composable or util and call it directly?

christian14b (commented):
Looks good! Just a couple of comments on my side

src/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.vue
cristhianDt: I appreciate the suggestion, but I don't think it deserves to be moved; it's just a couple of lines, and if different behavior is needed here or in MyKiva in the future, it can be done quickly

src/composables/useGoalInReview.js
cristhianDt: sure

src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
```diff
@@ -59,7 +59,7 @@
 					:year="data?.year"
 					:current-year="currentYear"
 					:feedback-submitted="feedbackSubmitted"
-					@back-to-kiva="handleCta('back-to-kiva')"
+					@goal-recap-back-to-kiva="handleCta('goal-recap-back-to-kiva')"
```
christian14b, line 62: Do we need `goal-recap-back-to-kiva` event? Seems like we can replace it with `close`  without altering the logic around it

src/composables/useGoalInReview.js
cristhianDt: before that I would like to get an opinion from @dyersituations about `getCtaHref`, looks like this method hasn't any guard I guess the idea is redirect the user to `/lend/filter` even if there is no category or target, so moving the validation to there we'll affect the main feature of that method

christian14b: Yes, it's probably an edge (almost impossible) case that a user goal doesn't have a `category`, but instead of `null`, we should redirect them to the combo page without filters.

src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
cristhianDt: what do you mean?, we need to record the event with the tracker but in the end (for the user) we'll just close the modal

src/composables/useGoalInReview.js
cristhianDt: knowing that, the best thing to do is remove the validation, thanks @christian14b

dyersituations: Agreed regarding taking them to `/lend/filter` as a fallback.

dyersituations (approved):

src/composables/useGoalInReview.js
cristhianDt: removed, thanks guys, for pointing this out

src/components/MyKiva/GoalInReview/GoalInReviewModal.vue
cristhianDt: @christian14b  can we close the thread?, goal-recap-back-to-kiva and close events have the same final feedback to the user but the origins and flows are pretty different

PR #7185, opened 2026-08-21
fix(borrower-profile): show the full loan view to volunteers CIT-4820

emuvente:
Volunteers have access to loan data in all states, so they should get the full borrower profile view by default for all statuses. The user's volunteer status is not being checked, so volunteers are mistakenly being shown the minimal view first, putting an extra click between them and the full profile. This PR fixes that by getting the volunteer ID for the current user (which is null if they are not a volunteer) and checking that when deciding to show the full or minimal view. I also extracted the logic for that decision to a testable utility method so that there aren't divergences between the prefetch and mounted view behavior (AD-320).


dyersituations (approved):

kathrynlovett (approved):
LGTM! Thank you 🙏🏻

PR #7187, opened 2026-08-24
perf(MyKiva): prevent layout shift from section pop-in after mounted AD-439

emuvente:
The `LendingCategorySection`, `LendingStats` and `MyKivaBorrowerCarousel` sections are causing large layout shifts because they are hidden entirely until `mounted`. This PR addresses that by removing the `clientRendered` gate entirely and adjusting the loading placeholders. Most of the changed lines in this are stories to cover these sections and their loading states.

**LendingCategorySection**: the loan cards in this already had loading placeholders, so the only change that was needed was setting the slide max width in css so they display correctly before JS runs.

**LendingStats**: this section already had loading placeholders and only needed minor styling adjustments, but it also exposed a hydration mismatch issue in `GoalSettingModal`. That was addressed by adding a `v-if` inside the modal so it is only included in the page when the modal is opened.

**MyKivaBorrowerCarousel**: this content was SSR ready. The only issue that existed was the slides being the wrong width for their viewport size until JS runs and measures the window, which was fixed using the same css variable method used for `LendingCategorySection`.


dyersituations (approved):
Thanks @emuvente! Tested locally, looks good. I have some changes locally I've been experiment with that also moves more of these components to SSR, not a ton of extra data needed. Once this is merged, I'll integrate my findings in a follow-up PR.

roger-in-kiva (approved):

PR #7188, opened 2026-08-25
feat: mp-3076-kiva-credit-tip-toggle

infante-jaime:
## MP-3076 — Checkout: balance-toward-tip toggle (variant b)

Adds the checkout toggle that lets lenders choose whether their Kiva Credit pays
the tip, for the tip-from-balance experiment (epic MP-3000). The UI persists the
preference and renders the backend-owned manifest — no client-side allocation math.

**6 files, +413/−3.** Layout/cosmetic changes that were originally in this PR have
been split out to **MP-3154** at review request.

### How it works

**The toggle** — `KivaCreditTipToggle.vue`

Renders in the tip row for variant `b` when the lender is logged in, has balance
> $0, has a loan in the basket, and a tip > $0. Hidden at a $0 tip and returns
with its previous state when a tip comes back.

State is read from `Manifest.applyKivaCreditToDonation` (added to
`initializeCheckout`); changes call the merged `updateKivaCreditDonationPreference`
mutation, which returns the rebuilt manifest with totals, then emit the existing
`refreshtotals`. Failures toast and refresh so a stale basket recovers.

Tracking fires `basket / click / tip-from-balance-toggle-on|off` on explicit
changes only — nothing fires for the default state.

**Seeding the default-off** — the eng call the ticket left open

Treatment has to default OFF, but `applyKivaCreditToDonation` is a `Boolean!` that
can't distinguish "never chose" from "chose yes", so the value alone can't drive
seeding. On the first eligible render of a variant-b basket still reading `true`,
the component fires the mutation once with `false` and records the basket id in a
`kvtipseeded` cookie; a marked basket is never re-seeded, so an explicit lender
choice is never overwritten. The toggle stays hidden until the manifest and the
rendered state agree.

Known edge case: clearing that cookie while the preference is ON re-seeds it off
on the next eligible render. Accepted trade-off of client-side seeding.

**Assignment** — read once in `CheckoutPage.created()` and provided to descendants
via `provide/inject`, following the existing `customTipDefaultVersion` pattern in
the same file. No exposure tracking here; MP-3077 owns that.

### Testing

10 specs covering control, seeding, re-seed protection, each hide condition,
tip-zero reappearance, persistence + tracking, and the failure/revert path.
Full suite green (330 files / 5,166 tests); ESLint, Stylelint and GraphQL lint clean.

Manually verified on dev: seeding fires exactly once per basket, the choice
survives reloads, on/off events fire, $0-tip hides and restores, control unchanged.

### Before launch (not in this PR)

- Prod settings manager: `uiexp.checkout_tip_from_balance_toggle` plus appending
  `checkout_tip_from_balance_toggle` to `ui.active_experiments`. **No space after
  the comma** — the resolver splits on `,` without trimming, so a leading space
  makes the key silently unmatched. Dev is configured.
- Experiment tracking key (`EXP-MP-3006-Aug2026` vs a rename) still needs settling
  in the epic before MP-3077 wires exposure.
- FYI: Figma's `text-secondary` is #505050 (gray-600) while `@kiva/kv-tokens` ships
  #757575 — the label uses `tw-text-gray-600` to match the design.
  
  
image


src/components/Checkout/UpsellModule.vue
```diff
@@ -1,6 +1,24 @@
 <template>
-	<div class="tw-bg-brand-100 tw-rounded tw-p-4">
-		<div class="tw-w-full tw-text-right tw-h-3">
+	<div
```
dyersituations: Why the changes to the upsell component? Is it just to fix the content alignment on the component? If so, I think this work should be separated from the tip experiment.

src/components/Checkout/DonationItem.vue
```diff
@@ -104,13 +105,20 @@
 						</div>
 					</div>
 
-					<div>
+					<div
+						:class="{
+							'md:tw-my-1 md:tw-max-w-2xl': isTipToggleVariant,
+							'md:tw-flex md:tw-items-center md:tw-gap-0.5': isTipToggleVariant,
```
dyersituations: These two entries can be combined.

src/components/Checkout/DonationItem.vue
```diff
@@ -345,6 +367,9 @@ export default {
 		isCampaignDonation() {
 			return !!this.donation?.metadata?.campaignId;
 		},
+		isTipToggleVariant() {
+			return this.tipFromBalanceVersion === 'b' && !this.isCampaignDonation && !this.orderTotalVariant;
```
dyersituations, line 368: Here we have more checks than in `BasketItem`, do both components need the same checks?

src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -0,0 +1,170 @@
...
+<script>
+import _filter from 'lodash/filter';
+import _get from 'lodash/get';
+import numeral from 'numeral';
+import logFormatter from '#src/util/logFormatter';
+import initializeCheckout from '#src/graphql/query/checkout/initializeCheckout.graphql';
+import updateKivaCreditDonationPreference from '#src/graphql/mutation/updateKivaCreditDonationPreference.graphql';
+import { KvSwitch } from '@kiva/kv-components';
+
+export const TIP_FROM_BALANCE_EXP_KEY = 'checkout_tip_from_balance_toggle';
```
dyersituations, line 30: Please add this to the active experiments doc and prod + dev settings managers.

src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -0,0 +1,170 @@
...
+				this.seeding = false;
+			});
+		},
+		setPreference(value) {
+			if (this.updating || value === this.applyKivaCreditToDonation) return;
+			this.updating = true;
+			this.toggleValue = value;
+			this.$emit('updating-totals', true);
+			this.apollo.mutate({
+				mutation: updateKivaCreditDonationPreference,
```
dyersituations: Looks like much of the two usages here of `updateKivaCreditDonationPreference` could be combined into a single method.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1517,11 +1537,22 @@ export default {
 .upsellContainer > .loading-placeholder {
 	min-height: 250px;
 }
+
+/* Tip-from-balance variant: banner is 120px on desktop; mobile keeps the control reservation */
+.upsellContainerCompact,
```
dyersituations: Let's use `-` naming for consistency.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1517,11 +1537,22 @@ export default {
...
+.upsellContainerCompact > .loading-placeholder {
+	min-height: 120px;
+}
 @media screen and (width <= 733px) {
 	.upsellContainer,
 	.upsellContainer > .loading-placeholder {
 		min-height: 300px;
 	}
+
+	.upsellContainerCompact,
```
dyersituations: Also, why do we need upsell-specific changes for this experiment here?

src/graphql/query/checkout/shopBasketUpdate.graphql
```diff
@@ -17,6 +17,7 @@ query shopBasketUpdate($basketId: String) {
 		basket {
 			id
 			hasFreeCredits
+			applyKivaCreditToDonation
```
dyersituations: Is this version of the query using this field anywhere in this changeset?

src/components/Checkout/UpsellModule.vue
infante-jaime: yest trying to fullly maatch the designs and it looked weord just changing one let e log another task for cosmetics and split this.

src/components/Checkout/DonationItem.vue
infante-jaime: done

src/components/Checkout/DonationItem.vue
infante-jaime: since split happened this is moved to the other PR (layout),

src/components/Checkout/KivaCreditTipToggle.vue
infante-jaime: Done

src/components/Checkout/KivaCreditTipToggle.vue
infante-jaime: done

src/pages/Checkout/CheckoutPage.vue
infante-jaime: done

src/pages/Checkout/CheckoutPage.vue
infante-jaime: again to match design spliting

src/graphql/query/checkout/shopBasketUpdate.graphql
infante-jaime: Good catch — nothing reads it from this query, only from initializeCheckout. Removing it.

src/components/Checkout/UpsellModule.vue
dyersituations: @infante-jaime please tag Lauren and Laura in the split out ticket to ensure we want the upsell changes now and whether we roll those style changes out to all users or only users in the experiment variant.

src/components/Checkout/DonationItem.vue
```diff
@@ -104,13 +105,18 @@
 						</div>
 					</div>
 
-					<div>
+					<div
+						:class="[showTipFromBalanceVariant
```
dyersituations: Let's use the same syntax here for optional classes that we have on other components.

src/components/Checkout/DonationItem.vue
```diff
@@ -345,6 +365,9 @@ export default {
 		isCampaignDonation() {
 			return !!this.donation?.metadata?.campaignId;
 		},
+		showTipFromBalanceVariant() {
+			return this.tipFromBalanceVersion === 'b' && !this.isCampaignDonation && !this.orderTotalVariant;
```
dyersituations, line 368: What is `orderTotalVariant`?

src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -0,0 +1,174 @@
...
+		data-testid="tip-from-balance-toggle"
+	>
+		<kv-switch
+			class="tw-flex"
+			data-testid="tip-from-balance-switch"
+			:model-value="toggleValue"
+			:disabled="updating"
+			@update:model-value="setPreference"
+		>
+			<!-- tw--ml-1 nets KvSwitch's internal 16px label gap down to the design's 8px -->
```
dyersituations: Let's try not to override global component layout/styles. If this change is critical, let's ask design if we can centralize this in the shared component. For now, I think we can use the shared component design.

infante-jaime: Removed the override and using the shared switch as-is now. I also removed tw-text-base, so the !important is no longer needed.

The only difference from the design is the label spacing (8px vs the component's 16px). I'll check with design if that's worth updating in KvSwitch.

src/components/Checkout/DonationItem.vue
infante-jaime: It's an existing prop that marks the duplicate DonationItem rendered by OrderTotals. We use it here so the toggle only shows on the real tip row and doesn't appear twice.

src/components/Checkout/DonationItem.vue
infante-jaime: updated

dyersituations:
@infante-jaime this isn't a realistic use case, but I want to make sure it isn't evidence of any potential issues:
1. Enable variant
2. Load basket w/ loan so tip is not paid for by balance
3. Change to control with `setuiab`
4. Reload page
5. Notice below

This won't happen with normal users, but please take a look to see if that "hanging $5" doesn't show evidence of any potential contamination between control and variant.

image

infante-jaime:
@dyersituations 

Yes I did had a bug now solved an fullu tested end to end locally 

Basket throughout: **$25 loan + $5 tip = $30.**

| Balance | Toggle | Backend reports | Screen shows | Sum | Balance change | New money | Evidence |
| --- | --- | --- | --- | ---: | ---: | ---: | --- |
| **$75** covers everything | ON | applied $30, due $0 | `−$30.00 / $0.00` | $30 | −$30 | $0 | **live** |
| | OFF | applied $30, due $5 | `−$25.00 / $5.00` | $30 | −$25 | $5 | money **live**, number **unit** |
| **$3** below the tip | ON | applied $3, due $27 | `−$3.00 / $27.00` | $30 | −$3 | $27 | derived |
| | OFF | applied $3, due $27 | `−$3.00 / $27.00` | $30 | −$3 | $27 | clamp no-op **unit** |
| **$27** covers loan, not tip | ON | applied $27, due $3 | `−$27.00 / $3.00` | $30 | −$27 | $3 | derived |
| | OFF | applied $27, due $5 | `−$25.00 / $5.00` | $30 | −$25 | $5 | clamp **unit** |

### What this demonstrates

Two things hold in all six cases:

1. **Credit shown + Total = basket total.** No more confusing $35 math.
2. **Credit shown = actual balance used.** The clamp keeps the displayed credit consistent with what is actually deducted.

The toggle effect is also clear between each ON/OFF pair:

- **$75 balance:** $5 shifts to new money.
- **$27 balance:** $2 shifts to new money.
- **$3 balance:** no change because the balance wasn't enough to cover the tip anyway.
toggle on
image
Toggle off 
image

src/components/Checkout/OrderTotals.vue
```diff
@@ -270,16 +276,29 @@ export default {
...
 		},
 		showKivaCredit() {
 			return this.showRemoveKivaCredit || this.showApplyKivaCredit;
 		},
+		appliedKivaCredit() {
+			const applied = numeral(this.totals.kivaCreditAppliedTotal).value() ?? 0;
+			// The lender has chosen to pay the tip with new money, so creditAmountNeeded is floored at the
+			// tip: that deposit lands in the balance and is spent again as credit, leaving the balance to
+			// fund only the non-tip items. Reporting the raw applied credit alongside the floored amount due
+			// reads as more than the basket total. See the tip floor in MP-3074's getAmountDue().
```
dyersituations: Remove ticket mention from comments.

infante-jaime: done

dyersituations:
I tried locally with the latest changes, and the inconsistency when moving back to control on the same basket still exists. Since you fixed something last time I mentioned this, I assume this might be evidence of a potential issue? Or is this just an edge case we shouldn't worry about?

Here I did as mentioned previously
- Open checkout as variant
- Add to basket
- See balance not applied like expected
- Use `setuiab` to change to control
- Basket is in weird state where balance is still not applied (control applies balance to entire basket

image

infante-jaime:
not an edge case, you're right. the toggle value stays in the basket and nothing ties it to the assignment, so it sticks even if you end up in control, and control has no toggle to undo it. I added some logic to reset it to the default when the lender isn't in the variant.

dyersituations:
I tested the same steps above locally with the latest changes, and in more than one occasion I encountered the below error when returning to the control (as you can see in the screenshot below, no toggle). Please use `setuiab` locally to switch between control and variant a handful of times to try to repro. The error isn't obvious at first glance, so I don't have more details at the moment.

image

infante-jaime:
> I tested the same steps above locally with the latest changes, and in more than one occasion I encountered the below error when returning to the control (as you can see in the screenshot below, no toggle). Please use `setuiab` locally to switch between control and variant a handful of times to try to repro. The error isn't obvious at first glance, so I don't have more details at the moment.
> 
> image

fixed a race condition, the payment form was mounting while the basket was still resetting so braintree couldn't build the drop in. also removed a duplicate call the toggle was making on every load. couldn't repro after that

src/components/Checkout/OrderTotals.vue
```diff
@@ -270,16 +276,29 @@ export default {
...
+		appliedKivaCredit() {
+			const applied = numeral(this.totals.kivaCreditAppliedTotal).value() ?? 0;
+			// The lender has chosen to pay the tip with new money, so creditAmountNeeded is floored at the
+			// tip: that deposit lands in the balance and is spent again as credit, leaving the balance to
+			// fund only the non-tip items. Reporting the raw applied credit alongside the floored amount due
+			// reads as more than the basket total. The floor lives in ManifestViewHelper::getAmountDue().
+			if (this.applyKivaCreditToDonation === false) {
+				const nonTipTotal = (numeral(this.totals.loanReservationTotal).value() ?? 0)
+					+ (numeral(this.totals.kivaCardTotal).value() ?? 0);
+				return Math.min(applied, nonTipTotal);
```
dyersituations, line 296: I'm confused by this logic: it returns the credit applied or the total of items in basket that aren't tips, whichever is the smaller amount. So we could have more credit applied than items in the basket. When would that happen? Wouldn't that be a bug?

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -933,6 +937,33 @@ export default {
...
+			};
+		},
+		pendingTipPreferenceReset() {
+			// The basket is about to be put back to the default, which changes the amount due. The
+			// payment form must not mount against a total we are already replacing: it tears its own
+			// container out of the DOM mid-initialization and Braintree fails to build the drop-in.
+			// Once an attempt has failed, stop withholding: a lender who cannot pay is worse off than
+			// one looking at a basket that still charges for the tip
+			// An unknown version is not "outside the variant": the assignment arrives after the first
+			// render when it has to be queried, and resetting on that would undo a treatment basket
```
dyersituations: This comment is long and confusing.

src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1129,6 +1160,7 @@ export default {
 				this.cookieStore.remove(TIP_FROM_BALANCE_SEEDED_COOKIE, { path: '/' });
 				this.refreshTotals();
 			}).catch(error => {
+				this.tipPreferenceResetFailed = true;
```
dyersituations, line 1160: What happens if `tipPreferenceResetFailed` happens? I don't see a recovery path.

src/components/Checkout/OrderTotals.vue
infante-jaime: Not a bug — this is intentional. The backend keeps credit applied to the whole basket, including the tip. So a $25 loan + $5 tip can come back as $30 in credit, even though only $25 is going toward loans.

When the toggle is off, the lender pays the $5 tip with new money. That money lands in the balance and is immediately spent again as credit, so showing $30 credit + $5 due makes it look like a $35 basket.

The UI instead shows the balance actually funding the loans/Kiva Cards. If the balance doesn't even cover those, we leave the credit amount unchanged.

src/pages/Checkout/CheckoutPage.vue
infante-jaime: There is a retry, just not in the same session. On the next checkout load the state resets and it tries again.

I didn't want to retry immediately and keep payment blocked if the mutation is failing. If it never succeeds, the tip is simply paid with new money instead of balance. That felt better than blocking checkout.

src/pages/Checkout/CheckoutPage.vue
infante-jaime: simplified comments

src/pages/Checkout/CheckoutPage.vue
dyersituations: Why would you retry on the next session? There would be a new basket.

infante-jaime: sorry I meant next page load,  not next session. same basket. it stays in the kvbskt cookie through reloads and coming back later, so if the mutation failed and they land on checkout again, the preference is still false and it tries again.

a brand new basket only happens after they check out, and that one starts with the preference unset, so there's nothing to reset there anyway.

dyersituations: What happens if the user checks out after this step fails? Do they see something different in the UI from what the BE uses for the checkout? Any other potential issues?

infante-jaime: the amount due comes straight from the BE (creditAmountNeeded), we don't do any math on it, so what they see is what they're charged. the credit clamp isn't variant-gated either, so a failed reset still shows −$25 / $5 in control. no mismatch.

dyersituations: In that case what is happening here if/when `tipPreferenceResetFailed` is set to `true`? I assume this is an edge case, but the fact that it's coded into the flow makes me think it is still important.

infante-jaime: always very important to thing of all the possible scenarios and thank you for that, It only blocks while the reset is running, so Braintree doesn't build against a total we're about to change. If the reset fails, the block is removed and they can still pay.

It shouldn't fail normally since it uses the same API as the rest of checkout. The flag is really just for one-off failures like a timeout or stale basket, where everything else is still working.

dyersituations: Looking at the code again, the fallback here is that they would be checking out with whatever the state was before. And this failure could be triggered when `applyKivaCreditToDonation` is changed by the toggle. So a user could: click toggle, mutation fails, then can checkout with previous state. What would the user see in this case? Before or after the toggle failed? I just want to make sure the user sees what the transaction will be sending to their credit card.

infante-jaime: Good question. They see the original state, and that's what gets charged.

If the toggle fails, we reset it to the basket's actual value, show an error toast, and refresh the totals from the server. So there's no point where the toggle shows one thing but we charge another — worst case, they see the error and the toggle goes back to where it started.

And it's a very edge case anyway. It needs that one mutation to fail while everything else still works, and if it does they just click again.

dyersituations (commented):
Changes look good, and I tested it locally, but I want to get at least one more set of eyes on the changes before I approve due to this being such a critical part of Kiva. @infante-jaime thanks for digging into the questions I added comments for!

infante-jaime:
> Changes look good, and I tested it locally, but I want to get at least one more set of eyes on the changes before I approve due to this being such a critical part of Kiva. @infante-jaime thanks for digging into the questions I added comments for!

thanks for reviewing! good questions make better code.

roger-in-kiva:
Looks like tablet and medium screens (1024px) is not fitting well:
Screenshot 2026-09-02 at 9 50 33 a m
Screenshot 2026-09-02 at 9 50 55 a m

Also ensure mobile is like expected:
Screenshot 2026-09-02 at 9 52 05 a m

infante-jaime:
> Looks like tablet and medium screens (1024px) is not fitting well: Screenshot 2026-09-02 at 9 50 33 a m Screenshot 2026-09-02 at 9 50 55 a m
> 
> Also ensure mobile is like expected: Screenshot 2026-09-02 at 9 52 05 a m

fixed. there's also a copy ticket coming that shortens the cta to "learn more", so this'll look neater once that lands — worth another look then.

updated the text early just to make it fit better

cristhianDt:
@infante-jaime probably is not related but I saw the toggle first then I decided `No donation to Kiva` then this checkbox broke the design (desktop and tablet), look

Captura de pantalla 2026-09-02 a la(s) 11 49 40

src/components/Checkout/DonationItem.vue
```diff
@@ -104,13 +108,17 @@
...
 					</div>
 
-					<div>
+					<div
+						:class="{ 'md:tw-flex md:tw-items-center md:tw-gap-0.5': showTipFromBalanceVariant }"
+					>
 						<div
-							class="donation-tagline tw-my-1 tw-max-w-2xl"
+							class="tw-my-1 tw-max-w-2xl"
+							:class="{ 'md:tw-min-w-0': showTipFromBalanceVariant }"
```
cristhianDt: we need to break up the copy for tablet variation, check this out:

Image

src/components/Checkout/DonationItem.vue
```diff
@@ -138,6 +147,12 @@
 						</button>
 					</div>
 				</div>
+
+				<kiva-credit-tip-toggle
```
cristhianDt, line 154: spacing is greater than mocks also margin-top, try to remove the margin-bottom of the copy container (should fix the margin-top issue)

Image

infante-jaime:
> @infante-jaime probably is not related but I saw the toggle first then I decided `No donation to Kiva` then this checkbox broke the design (desktop and tablet), look
> 
> Captura de pantalla 2026-09-02 a la(s) 11 49 40

actually as weird as it looks is how it is in dev checking with design if we can update that by taking the toggle space

src/components/Checkout/DonationItem.vue
infante-jaime: we can address this one in the updated copy ticket because I guess it will look better on tablet with shorter text in the cta 

updated the text early just to make it fit better

src/components/Checkout/DonationItem.vue
infante-jaime: fixed

src/components/Checkout/OrderTotals.vue
cristhianDt, line 473: @infante-jaime after search for some edge case with the AI, there is one to double check in your end and confirm whether that's intended or should use appliedKivaCredit (otherwise ignore it then):

> Analytics basket / show / kiva-credit still reports the raw applied credit, not the corrected display value
> 
> In this PR you introduced appliedKivaCredit ([OrderTotals.vue:286-299](vscode-webview://1hjn9f94pgbogecqdq4mpl9nol4t65bfdvja6f3ui3q95vlugt2m/repos/kiva/ui/src/components/Checkout/OrderTotals.vue#L286-L299)) so the displayed Kiva Credit line reflects what actually comes off the balance when the toggle is OFF — e.g. a $25 loan + $5 tip shows $25, not the backend's $30. The visible kivaCredit computed now uses it.
> 
> But the tracking event still reads the raw backend field:
> 
> 
> // OrderTotals.vue — trackBasketInfoEvents(), line ~475
> const creditValue = 100 * (!this.showApplyKivaCredit ? this.totals.kivaCreditAppliedTotal
>     : this.totals.creditAvailableTotal);
> this.$kvTrackEvent('basket', 'show', 'kiva-credit', creditProperty, creditValue);
> So for a toggle-OFF basket ($25 loan + $5 tip):
> 
> UI displays: $25.00 (via appliedKivaCredit)
> Event sends: 3000 (raw kivaCreditAppliedTotal × 100 = $30)
> The displayed number and the tracked number diverge by exactly the tip amount for every treatment basket that pays the tip with new money — which is the whole treatment population. If this event feeds the experiment's Looker funnel or any credit-applied metric, it will over-report credit-on-basket for variant b.
> 
> Question: is this intentional (track the raw backend allocation) or should it use appliedKivaCredit to match what the lender sees? If the latter, the one-line fix is to swap this.totals.kivaCreditAppliedTotal → this.appliedKivaCredit in trackBasketInfoEvents. Either way, worth a deliberate call rather than an accident of which computed each path happened to reference.

src/components/Checkout/DonationItem.vue
cristhianDt: please link the ticket in this thread, thanks

cristhianDt:
the feature and integration looks good, great job @infante-jaime, just an edge case to confirm (already commented)

here my TCs executed:

**Legend:** ✅ OK (tested & passing)

| # | Test case | Precondition | Expected | Status |
|---|-----------|--------------|----------|--------|
| 1 | Toggle appears for eligible user | Variant `b`, logged in, **balance > $0**, loan in basket, tip > $0 | Toggle renders in the tip row | ✅ |
| 2 | Seed cookie saved on first render | N°1, `kvtipseeded` absent | `kvtipseeded` cookie set to basket id | ✅ |
| 3 | First render seeds OFF | N°1, basket preference untouched | One `updateKivaCreditDonationPreference(false)` fires; toggle shows **OFF**; no `tip-from-balance-toggle-on/off` analytics on default | ✅ |
| 3b | No re-seed on second refresh (cookie present) | After N°3, **don't** clear cookie or touch toggle → refresh again | **No** mutation fires; toggle stays OFF (proves once-per-basket seeding) |  ✅  |
| 4 | No balance → toggle hidden | Variant `b`, user with **no Kiva credit / $0 balance** | Toggle not rendered; no seeding mutation | ✅ |
| 5 | Zero the tip → toggle hides | Toggle visible, then set tip/donation to **$0** | Toggle disappears (or disabled); no errant mutation | ✅ |
| 6 | Re-add positive tip → toggle reappears with persisted state | After N°5, set tip back to a positive amount | Toggle reappears in **previously selected** state; no re-seed mutation | ✅ |
| 7 | Toggle **ON** applies balance to tip | Toggle visible OFF | `...(true)` fires; credit rises, amount due drops; event `basket / click / tip-from-balance-toggle-on` | ✅ |
| 8 | Toggle **OFF** returns to deposit state | Toggle ON | `...(false)` fires; tip becomes amount due, payment form appears; event `basket / click / tip-from-balance-toggle-off` | ✅ |
| 9 | Persistence across reload | Toggle set to a non-default state | After full reload, toggle returns in same state; totals match | ✅ |
| 10 | Hidden with no loan (toggle **ON** path) | Add loans + tip → toggle ON → remove all loans | Toggle hidden; donation stays with credit applied (= platform default, not a bug) | ✅ |
| 10b | Hidden with no loan (toggle **OFF** path) | Add loans + tip → toggle OFF → remove all loans | Toggle hidden; credit withheld from donation (honors explicit choice) | ✅ |
| 11 | Control (variant `a`) unchanged | Force variant `a`, otherwise eligible | No toggle; no mutation; balance applies to tip as today | ✅ |
| 12 | Reset outside the variant | Basket at preference `false` in `b`, then render under variant `a` | One `...(true)` fires; `kvtipseeded` removed; totals revert; payment form waits for new amount | ✅ |

src/components/Checkout/DonationItem.vue
cristhianDt: definitely we need to do smt about the copy; the design looks strange with it and the checkbox

Captura de pantalla 2026-09-02 a la(s) 12 52 36

src/components/Checkout/DonationItem.vue
cristhianDt: (optional) no a big issue just check the gab I keep seeing 16px, mocks has 8px > https://www.figma.com/design/H0RJzgOoeuFmdH8ggDX2c1/Donations-from-deposits?node-id=122-4616&m=dev and margin-top 16px not 8px

src/components/Checkout/OrderTotals.vue
infante-jaime: good catch, it's deliberate — the event has always meant "credit consumed" ($30 really does leave your balance before $5 comes back as the deposit), while the new UI line means "net change to your balance" ($25). different questions, both right.

flagged on MP-3174 for lauren rather than swapped, since that event has history and redefining it mid-flight breaks before/after comparisons. also narrower than it looks — it only diverges when the balance covers the loans and spills into the tip.

src/components/Checkout/DonationItem.vue
infante-jaime: https://kiva.atlassian.net/browse/MP-3087

src/components/Checkout/OrderTotals.vue
cristhianDt: understood, thanks for checking

cristhianDt (approved):
I dont have any other feedback, just the spacing thread

dyersituations:
@infante-jaime the spacing issues are not currently affecting control? Just want to make sure if this gets merged that the non-experiment/current experience isn't changed. We'll be releasing `ui` soon.

infante-jaime:
> @infante-jaime the spacing issues are not currently affecting control? Just want to make sure if this gets merged that the non-experiment/current experience isn't changed. We'll be releasing `ui` soon.

This shouldnt be an issue by design there are no ui changes for the control just in the experiment and in this particular case is the toggle spacing.

So I guess we are good to go and any layout concerns can be addressed [here](https://kiva.atlassian.net/browse/MP-3087)

dyersituations (approved):

PR #7193, opened 2026-08-26
fix: show goal recap cta from the completed goal card

christian14b:
### Problem
After completing a goal, the goal card on MyKiva showed **"View impact progress"** instead of **"View goal recap"**, so the recap was unreachable from that card.

Two things made it look intermittent:
- The completed tile renders **once**. `hideGoalCard` is persisted after the first completed view, so it is hidden on every later load.
- It only renders in the **non-goals-row** experience. When `goalsRowEnabled` is on, `LendingStats` hides the tile and `MyKivaFeaturedSlot` takes over. So the unwired card was the only one some lenders saw.

`NextYearGoalCard` / `GoalProgressRing` had no goal-in-review wiring at all; the recap entry point had been added to the featured slot and the Impact Progress row

### DIFF
--- a/src/components/MyKiva/GoalProgressRing.vue
+++ b/src/components/MyKiva/GoalProgressRing.vue
@@ -69,7 +69,8 @@
 
 		<KvButton
 			class="tw-w-full goal-button"
-			v-kv-track-event="['portfolio', 'click', 'continue-towards-goal']"
+			:key="`goal-cta-${showRecapCta}`"
+			v-kv-track-event="showRecapCta ? null : ['portfolio', 'click', 'continue-towards-goal']"
 			@click="handleButtonClick"
 		>
 			{{ buttonText }}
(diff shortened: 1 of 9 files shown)


src/components/MyKiva/GoalProgressRing.vue
```diff
@@ -69,7 +69,8 @@
 
 		<KvButton
 			class="tw-w-full goal-button"
-			v-kv-track-event="['portfolio', 'click', 'continue-towards-goal']"
+			:key="`goal-cta-${showRecapCta}`"
+			v-kv-track-event="showRecapCta ? null : ['portfolio', 'click', 'continue-towards-goal']"
```
dyersituations, line 73: Do we want a different click event here for recap?

christian14b: Yep, it fires. `openGoalRecapFromCard` on the page sends `portfolio / click / view-goal-recap`, same as the
`Impact Progress` row. It's null here just so we don't also send `continue-towards-goal` on that same click.

Although, I do believe this can be cleaner by refactoring the progress ring component

dyersituations (approved):

PR #7203, opened 2026-08-27
fix: mykiva initial loading performance improvements

dyersituations:
https://kiva.atlassian.net/browse/MP-3139

V2 of PR, last branch history got messy with merging in other branches.

General improvements to the initial loading of MyKiva and MyKiva Next Steps pages, following stakeholder feedback that the loading felt slow. Improvements include focusing on improving above-fold loading:

- Reduced image size of top element background image
- Removed fields from MyKiva query that are no longer being used
- Split out queries to reduce data loaded from a shared stats query
- Added a little more data to prefetch so above-fold components can be server-rendered
- Added hydrate from cache process for goal data composable to use the prefetched data when available


emuvente (approved):
This is a great set of improvements!

mcstover (approved):

christian14b (approved):

PR #7204, opened 2026-08-28
feat(live-loan): render compact bundle card as transparent PNG

roger-in-kiva:
## What
The `compact-bundle` live-loan card now renders as a **transparent PNG** instead of a JPEG baked onto a white margin.

## Why
The card's drop shadow was baked onto a white background, so it only read on white email templates. Dark-mode templates need the margin transparent so the card + shadow composite onto the email's own background.

## How
- `drawCompact`: keep the canvas alpha channel, `clearRect` the canvas to transparent (instead of a white `fillRect`), and export `toBuffer('image/png')`. The white card body and drop shadow are unchanged; only the surrounding margin is now transparent.
- Content type is resolved per style via a new `contentTypeForStyle(style)` helper, used by the router for **both** fresh draws and cache hits (the cache stores only the buffer, so the header is derived from the style). Only `compact-bundle` is PNG; all other styles stay JPEG.
- **Sizing:** verified against Figma node 19534-4929 — every dimension already matches; no changes needed. Card height stays fixed at the 4-line worst case (the Figma frame is just its 3-line sample).

## Testing
- Compact render asserts PNG signature bytes + a fully transparent margin corner.
- `contentTypeForStyle` mapping unit test; router content-type assertions flipped to `image/png` for the three compact routes, JPEG retained for the others.
- Full unit suite green (one pre-existing, unrelated flake); lint clean.

## Known follow-up (not in this PR)
On a dark background the existing `rgba(0,0,0,0.08)` **black** shadow is very faint (a black shadow has little contrast against dark). This PR delivers the transparency the dark template needs; tuning the shadow treatment for dark (stronger shadow / light glow / per-template) is a separate design decision still open.

### DIFF
--- a/server/live-loan-router.js
+++ b/server/live-loan-router.js
@@ -1,7 +1,7 @@
 import express from 'express';
 import { error, warn } from './util/log.js';
 import { getFromCache, setToCache } from './util/memJsUtils.js';
-import drawLoanCard from './util/live-loan/live-loan-draw.js';
+import drawLoanCard, { contentTypeForStyle } from './util/live-loan/live-loan-draw.js';
 import fetchLoansByType, { QUERY_TYPE } from './util/live-loan/live-loan-fetch.js';
 import { trace } from './util/mockTrace.js';
 import { resolveBundleSize, DEFAULT_BUNDLE_COUNT, MAX_BUNDLE_COUNT } from './util/live-loan/bundle-size.js';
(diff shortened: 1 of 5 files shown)


cristhianDt:
(doubt) this feat is for the dark mode in some devices?

roger-in-kiva:
> (doubt) this feat is for the dark mode in some devices?

that's correct!

cristhianDt:
@roger-in-kiva nice! could you please add some screenshots for references/evidences ?, then I will approve it 🎉

roger-in-kiva:
> @roger-in-kiva nice! could you please add some screenshots for references/evidences ?, then I will approve it 🎉

cristhianDt (approved):
nice feature ☺️ !!

christian14b (approved):

dyersituations (commented):
👍

PR #7206, opened 2026-08-31
fix(header): sync ESI CSS variables for the new header states

mcstover:
Restores the ESI CSS-variable personalization that `KvWwwHeaderBasic` lost when it shipped. The variables were already emitted and already bridged; the hooks that read them were missing, so both ends were inert.

- The bridge only applied while user data was loading, so `--basket-display` never reached the header during a basket-only load.
- `userAvatarData` emitted `user-avatar` unconditionally, leaving the real-avatar element shown with an empty image whenever there was nothing to paint. It now emits the url only when it will be shown, and hides that element otherwise so the avatar icon carries the state.
- Adds the logged-out signal (`user-loading-display`) that the logged-in cluster hook needs.

## Scope of effect

`useCDNCaching` is route meta and only `/ui-site-map` sets it (`src/router/routes.js:696-701`), so the header binds `display` to these variables on that route alone. Live today: the basket hook and the cluster hook.

The `user-avatar` polarity fix has no live consumer yet. `avatarMode` cannot reach `esi` until `KvWwwHeaderBasic` exposes `useEsiAvatar` (CIT-5084), and `MyKivaButton` — the other component that reads these variables, directly under the `--ui-data-` prefix — only renders on minimal and corporate pages, which are never CDN-cached, so its variable branch is unreachable. That part of the change is groundwork for CIT-5085 rather than a visible fix.

One intended behaviour change where the hooks are live: on a stale logged-in-filled cached shell served to a logged-out visitor, the cluster is now suppressed pre-hydration. The visitor links are `v-if`-driven and stay hidden (design decision D4), so that visitor briefly sees neither the cluster nor a Log in link, where before they saw placeholder chrome.

## Verifying

ESI resolves locally, not only at Fastly: `server/vue-render.js:24-26` runs a `nodesi` processor over the rendered HTML when `config.server.simulateCDN` is set, and `index.html:5` carries the `/esi-ui/head` include. Under that flag `cdnNotedLoggedIn` comes from the `kvls` login-sync cookie rather than the Fastly header (`server/vue-middleware.js:101-105`), so both loading states are reachable locally.

With a `simulateCDN: true` config, on `/ui-site-map`:

- logged out with no `kvbskt` cookie → no basket link at any point;
- `kvls` set to anything other than `o` while not actually logged in → the logged-in cluster stays suppressed.

Logged-in user state confirmed loading correctly.

Covers CIT-5069 items 1, 2 and 4, and CIT-5071 item 3. The remaining items need `kv-ui-elements` changes and are tracked in CIT-5084 and CIT-5085.

https://kiva.atlassian.net/browse/CIT-5069
https://kiva.atlassian.net/browse/CIT-5071


eddieferrer (approved):

emuvente (approved):
Great fix! The description claims that this doesn't change local behavior because "local dev has no ESI", but that's not true. We process ESI locally: https://github.com/kiva/ui/blob/131e3501f72f4d8222fb5b629e107f86bd91cf2d/server/vue-render.js#L51. Please verify the behavior locally and not just in a Fastly-fronted environment.

mcstover:
@emuvente I had already verified the work locally and have once more. Also had claude update the description

PR #7215, opened 2026-09-02
fix: compact live loan card fix

roger-in-kiva:
- compact live loan card fix
- https://kiva.atlassian.net/browse/MP-3137


server/util/live-loan/compact-card-constants.js
```diff
@@ -4,9 +4,9 @@
 export const compactResizeFactor = 3;
 export const compactCardWidth = 271;
 export const compactCardPadding = 12;
-export const compactCardRadius = 16;
+export const compactCardRadius = 0;
```
dyersituations, line 7: The screenshot makes it look like the card has a tiny bit of border radius, should that be removed?

roger-in-kiva: `export const compactCardRadius = 0;` ensures no border radius.
Issue is b/c inset, fixed.

roger-in-kiva: Here an screenshot, best example than embedded above.
Screenshot 2026-09-02 at 1 53 48 p m

dyersituations: In that screenshot I'm seeing 4 lines, don't we want to try a max of 3 lines in loan use?

roger-in-kiva: Kat supposed that could fix the height issue, but is not.
Issue was a template padding setting, so we can keep the original requirement, more explicit for the user.

dyersituations (approved):
Sounds good to try these changes.

cristhianDt (approved):
LGTM

PR #7218, opened 2026-09-02
fix: logic to goal recap and goal card in my kiva

christian14b:
The recap CTA appeared the instant a goal completed, and clicking it made the goal card disappear. Both are now keyed to the MyKiva visit count instead.

  | | Goal Card | Confetti | CTA (View goal recap) | Goal Recap Modal |
  |---|---|---|---|---|
  | Visit 1 | shows | yes | no | no |
  | Visit 2 | shows | no | yes | opens; card survives closing it |
  | Visit 3+ | retires | no | no | no |


src/composables/useGoalInReview.js
```diff
@@ -71,25 +73,97 @@ export function getGoalInReviewCurrentYear() {
...
+ *
+ * @param {object} options Entry point inputs, describing the lender's active goal.
+ * @param {import('vue').MaybeRefOrGetter<boolean>} options.enabled The goal_in_review_enable setting.
+ * @param {import('vue').MaybeRefOrGetter<string>} options.goalStatus The goal's status.
+ * @param {import('vue').MaybeRefOrGetter<number|null>} options.goalYear The goal's year.
+ * @param {import('vue').MaybeRefOrGetter<boolean>} options.announced The hideGoalCard preference,
+ *   which MyKiva writes on the visit that announces the completion.
+ * @param {import('vue').MaybeRefOrGetter<boolean>} options.hasViewedRecap Whether this year's
+ *   recap has already been opened.
+ * @param {import('vue').MaybeRefOrGetter<number>} [options.loansTowardGoal] Loans made toward the goal.
```
dyersituations: The types on these parameters are strange.

christian14b: Good catch. Rephrased it entirely

src/composables/useGoalData.js
```diff
@@ -965,9 +963,6 @@ export default function useGoalData({ apollo } = {}) {
 		}
 
 		if (isGoalComplete) {
-			// Marked as soon as the goal looks complete, even if the achievement service has
-			// not confirmed it yet, so the recap cannot appear later in this same session.
-			markGoalCompletedThisSession(cookieStore, GOALS_CURRENT_YEAR);
```
dyersituations, line 970: So if we're not marking complete here (and not using the cookie store for knowing if a user sees the modal), when do these two steps happen now?

christian14b: We store a flag if the user sees the modal on `userPreferences` (That's our only reliable source to validate if a user already saw the goal recap), this cookie only helped us to prevent showing after thanks page

dyersituations: What happens in that case now (coming from TY)?

christian14b: User will see confetti + goal card completed state

Screenshot 2026-09-02 at 3 18 33 PM

If the user reloads the page, goal recap appears + goal card CTA will be "View goal recap". So, if you close it (by mistake or intended) you can click the card again to see the goal recap.

Screenshot 2026-09-02 at 3 08 31 PM

christian14b: if you completed your goal on the mobile app, you'll experience the same situation from above when you visit my kiva on a browser

dyersituations (approved):

cristhianDt:
👍🏽

PR #7220, opened 2026-09-03
feat: [MP-3177] integrate goal in review query

christian14b:
Integrating new goal query to goal in review work. https://kiva.atlassian.net/browse/MP-3177

### How to test

Go to `/mykiva?goto=goal-recap` to fire the modal and show the data. Compare with dev environment, both modals should have the same information


dyersituations (approved):
Tested locally, and it is a little faster for my personal account, nice!

PR #7225, opened 2026-09-04
feat: mp-3151-kiva-upsell component update

infante-jaime:
[MP-3154](https://kiva.atlassian.net/browse/MP-3154) — Checkout upsell banner: compact layout
Applies the compact "Support another borrower" design to the checkout upsell, and the matching type treatment to the loan row title.

What changed
Upsell banner — a shorter banner on desktop: smaller borrower avatar (w-10 instead of w-16), tighter padding, content vertically centred, a full-width progress meter instead of the short one, and a primary CTA rather than a link. The close button moves to the top-right corner on desktop while mobile keeps the existing inline one. Container min-height drops from 250px to 120px on desktop; mobile keeps 300px.

Loan row title — picks up the variant's heading style, so the borrower name matches the tip row above it.

Gated on the tip toggle's experiment
Everything is behind showTipFromBalanceVariant, reading the assignment MP-3076 already provides. No second experiment, no extra initializeExperiment call, no additional prefetch key — the compact banner appears for exactly the lenders who see the balance toggle.

Control is untouched
Every change is an additive class binding on the variant branch. The control path keeps its original classes verbatim — including tw-my-1 on the tagline and the original upsellContainer min-heights — because moving shared spacing regressed control earlier in this feature.

Testing
Full suite: 340 files, 5,333 tests passing. ESLint and Stylelint clean. Token classes only, no arbitrary values.

Merged with main (no conflicts). Verified no leftover references to the abandoned checkout_upsell_layout experiment.

image

[MP-3154]: https://kiva.atlassian.net/browse/MP-3154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -1573,9 +1582,18 @@ export default {
 .upsellContainer > .loading-placeholder {
 	min-height: 250px;
 }
+
+/* The compact banner is 120px on desktop; mobile keeps the taller reservation */
+.upsell-container-compact,
+.upsell-container-compact > .loading-placeholder {
+	min-height: 120px;
```
roger-in-kiva: tw-min-h-15

infante-jaime: done TY

src/components/Checkout/BasketItem.vue
```diff
@@ -12,7 +12,11 @@
 		<div class="tw-flex-auto borrower-info-wrapper">
 			<div class="borrower-info" data-testid="basket-loan-info">
 				<div class="tw-flex tw-mb-1">
-					<h2 class="tw-flex-grow" data-testid="basket-loan-name">
+					<h2
+						class="tw-flex-grow"
+						:class="{ 'tw-text-h3 !tw-font-medium tw-text-primary': showTipFromBalanceVariant }"
```
dyersituations: I noticed when testing locally that these upsell component changes appear when the user is in the variant even when they don't have a balance (so the toggle isn't appearing). Is that expected or do we only want to show the upsell styling changes when the toggle is also visible?

infante-jaime: Yes, that's what's happening. The styling only checks the arm, but the toggle also needs balance, loans, and a tip.

It doesn't affect the experiment results since exposure only fires for eligible lenders, but I agree they should match. If you can't see the toggle, you should just get the original design.

dyersituations: I see the latest changes, it looks like it might conflict with the other PR?
https://github.com/kiva/ui/pull/7226

infante-jaime: yes let me merge the other one so that I can resolve conflicts correctly

infante-jaime: conflicts resolved

dyersituations (approved):

cristhianDt (approved):

PR #7226, opened 2026-09-04
feat: mp 3174 tip toggle analytics

infante-jaime:
## MP-3174 — Tip toggle: exposure + audience exclusions

Adds the missing experiment exposure tracking and audience exclusions for the tip toggle.

### Exposure

* Fires once per eligible view in **both** control and variant.
* Variant fires once the toggle is ready to render.
* Control fires when the same eligibility conditions are met.
* Uses the existing experiment assignment and `trackExperimentVersion`.

### Audience

Excludes:

* Team members
* Lenders with **$1,000+ lifetime deposits**

Excluded lenders don't see the toggle or fire exposure.

### Implementation

* The toggle component now mounts in both arms so it can handle exposure consistently.
* Control remains inert: no toggle renders and no mutation runs.
* Prevents duplicate exposure from the `OrderTotals` copy.

### Testing

* **340 files / 5,350 tests passing**
* ESLint clean
* Coverage includes both arms, audience exclusions, and fire-once behavior.


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -591,6 +592,9 @@ export default {
 			this.myBalance = _get(data, 'my.userAccount.balance');
 			this.myId = _get(data, 'my.userAccount.id');
 			this.teams = _get(data, 'my.lender.teams.values');
+			// Falls back to zero: the field is nullable, and null must read as "no deposits" rather
+			// than "not loaded", or a lender funded only by promo credit is never in the audience
+			this.lifetimeDeposits = numeral(_get(data, 'my.lendingStats.totalAmountDeposited')).value() ?? 0;
```
dyersituations: For safety, and another entry below in this method, I wonder if we should have the `?? 0` inside `numeral`. See `this.depositIncentiveAmountToLend` below as an example.

infante-jaime: Good call. This also makes it fail safe for an invalid value. Matched it to `depositIncentiveAmountToLend`.

dyersituations (approved):

cristhianDt (approved):

PR #7229, opened 2026-09-05
refactor: use loan category ids instead of loan channel ids

eddieferrer:
Ticket: https://kiva.atlassian.net/browse/CIT-4130

## Summary
- Swaps the deprecated `lend.loanChannelsById` for `categoriesByIds` in the five Contentful-driven category components, reading `loanCategories` from the uiSetting `dataObject`
- Normalises the response so `savedSearch.loans` becomes a top-level `loans`, leaving the existing merge and selection logic unchanged
- `KivaMultiCategoryGrid` now tracks `selectedCategoryId` as a string rather than a numeric channel id
- Fixes a latent crash: `data?.lend?.loanChannelsById[0]` had no optional chaining on the index and would throw if the field returned null

## Notes
- The `loanFields` fragment is declared `on LoanBasic` and `savedSearch.loans.values` is `[LoanBasic]`, so the home page grid keeps its single-query prefetch — verified against the dev gateway
- `combinedLoanChannelData`, `selectedChannel` and `fetchLoanChannel` keep their names on purpose: they cross into `KivaClassicLoanCategorySelector`, `KivaLoanCardCategory` and `LoanCategorySelectorHomeExp`, which are shared with components covered by CIT-5107/5108/5111/5117
- No spec covers any of these five components, so lint and GraphQL document validation are the only automated checks here

## Blocked
`categoriesByIds` is dev-only right now — production returns `Variable "$ids" is never used`. Do not merge until it is deployed. The Contentful content is already migrated and published in both environments. Pairs with cms-page-server#3287.

### DIFF
--- a/src/components/Contentful/HomePage/NewHomeLoansByCategoryGrid.vue
+++ b/src/components/Contentful/HomePage/NewHomeLoansByCategoryGrid.vue
@@ -8,7 +8,7 @@
 			<kv-page-container>
 				<div>
 					<kiva-multi-category-grid
-						:contentful-loan-channels="loanChannels"
+						:contentful-loan-categories="loanCategories"
 						:loan-display-settings="loanDisplaySettings"
 						:new-home-exp="true"
 					/>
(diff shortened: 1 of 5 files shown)


mcstover (approved):
This should be fine. I'm not sure any of these are still in use at this time though they may be if if traced back to mykiva...

PR #7238, opened 2026-09-09
fix: mp-3193-show the switch only when credit can reach the tip

infante-jaime:
What this does

Fixes Lauren's finding: the tip-from-balance toggle appeared for lenders whose basket is as large as (or larger than) their balance. For them the switch does nothing — the amount due is the raw shortfall whether the tip comes from balance or not — so it read as a broken control.

The gate

Eligibility now requires the balance to exceed everything in the basket except the tip, instead of merely being positive. Below that line the switch can't change what's charged, so it doesn't render.

That "everything except the tip" is computed as itemTotal minus the tip donation, so a giving-fund (campaign) donation counts as something the balance has to pay for before it can reach the tip — and the definition holds if a new basket item type ever appears.

Exposure follows

Both arms read the same criteria, so control exposure narrows with the treatment and the arms stay comparable. Lenders excluded by this change never fire the experiment event in either arm.

Testing

Toggle spec: new cases for a balance below and exactly equal to the non-tip basket total.
Page spec: new coverage that checkout actually supplies the basket state the criteria read — including a campaign-donation basket where the loan alone would qualify but the donation pushes the tip out of reach. Previously nothing failed if that wiring was lost.
Verified by mutation, not just green tests: loosening the threshold back to balance > 0, changing > to >=, reverting the gate to loans + cards, and dropping the provided field each fail tests now.
Full unit suite and lint pass.
To see it: variant-b lender with, say, $30 balance and a $30+ basket → no toggle; drop the basket below the balance → toggle appears.

### DIFF
--- a/src/components/Checkout/KivaCreditTipToggle.vue
+++ b/src/components/Checkout/KivaCreditTipToggle.vue
@@ -35,16 +35,19 @@ const AUDIENCE_DEPOSIT_LIMIT = 1000;
 export const TIP_FROM_BALANCE_SEEDED_COOKIE = 'kvtipseeded';
 
 /**
- * Whether the lender is in the experiment audience, ignoring which arm they are in.
- * Unknown deposits read as ineligible until the basket query lands. The checkout page reads
- * this too, so the variant treatment only appears where the switch itself can.
+ * Whether the lender is in the experiment audience, ignoring which arm they are in. The checkout
+ * page reads this too, so the variant treatment only appears where the switch itself can.
+ *
+ * The balance has to exceed the loans and Kiva Cards, not merely be positive: below that the
+ * amount due is the raw shortfall either way, so the switch cannot change what is charged.
+ * Unknown deposits read as ineligible until the basket query lands.
  *
  * @param {Object} state The basket state provided by the checkout page
  * @returns {boolean} Whether the lender is in the experiment audience
  */
 export function meetsTipFromBalanceCriteria(state = {}) {
 	return !!state.myId
-		&& state.balance > 0
+		&& state.balance > state.nonTipTotal
 		&& state.hasLoans
 		&& state.tipAmount > 0
 		&& !state.onTeam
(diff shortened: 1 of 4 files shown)


src/pages/Checkout/CheckoutPage.vue
```diff
@@ -951,13 +951,17 @@ export default {
...
 				myId: this.myId,
 				balance: numeral(this.myBalance).value() ?? 0,
 				hasLoans: this.loans.length > 0,
-				tipAmount: numeral(tip?.price).value() ?? 0,
+				tipAmount,
 				basketId: this.basketId,
 				applyKivaCreditToDonation: this.applyKivaCreditToDonation,
+				// Everything the balance has to pay for before the tip, campaign donations included
+				nonTipTotal: (numeral(this.totals.itemTotal).value() ?? 0) - tipAmount,
+
```
dyersituations: Remove extra newline.

src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -35,16 +35,19 @@ const AUDIENCE_DEPOSIT_LIMIT = 1000;
...
+ * that the amount due is the raw shortfall either way, so the switch cannot change what is
+ * charged. Unknown deposits read as ineligible until the basket query lands.
  *
  * @param {Object} state The basket state provided by the checkout page
  * @returns {boolean} Whether the lender is in the experiment audience
  */
 export function meetsTipFromBalanceCriteria(state = {}) {
 	return !!state.myId
-		&& state.balance > 0
+		&& state.balance > state.nonTipTotal
```
dyersituations, line 50: I thought the balance needed to be greater than the total (loans, kiva cards, and tips).

infante-jaime: Good catch, but it's intentional. The idea of this check is to hide the toggle when it doesn't change anything.

As soon as the balance covers the loans/Kiva Cards, it can affect the charge. For example: $25 loan + $5 tip + $27 balance → on charges $3, off charges $5.

src/pages/Checkout/CheckoutPage.vue
infante-jaime: done

dyersituations (approved):

PR #7250, opened 2026-09-10
refactor: pin the header by route match instead of path substring

mcstover:
The header's sticky rule came from `isInExperimentPages`, which matched any path containing "lend" or "mykiva". That caught eight routes incidentally — `/lender/:publicId`, `/settings/autolending`, `/portfolio/lending-stats`, `/instant-lending-error`, `/process-instant-lending/…`, `/possibility/12-days-of-lending`, `/lend/saved-search` and `/lend-vue` — none of which support add to basket. Those are no longer sticky.

Stickiness is now one route rule (`isStickyHeaderRoute`) feeding one `stickyHeader` prop on `TheHeader`, so the route-driven case and the corporate case pin the same element through the same code path. `/cc/*` renders identically to before. `isInExperimentPages` is gone; its backing flag was removed in 19fc0b469 and only the route substring was left behind.

Sticky set confirmed with Casey on the ticket: My Kiva, the borrower profile, and the cc pages.

https://kiva.atlassian.net/browse/CIT-5031

### DIFF
--- a/src/components/WwwFrame/TheHeader.vue
+++ b/src/components/WwwFrame/TheHeader.vue
@@ -1,7 +1,7 @@
 <template>
 	<header
 		class="tw-transition-all tw-duration-1000 tw-ease-in-out"
-		:class="isInExperimentPages ? 'tw-sticky tw-top-0 tw-z-sticky' : ''"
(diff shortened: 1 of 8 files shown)


dyersituations (approved):
Thanks for the cleanup!

PR #7251, opened 2026-09-10
feat: mp-3194 layout fixes

infante-jaime:
What this does

Design-review fixes for the tip-from-balance checkout variant, from Lauren's QA pass (MP-3194) and Laura's design review (MP-3203). Everything renders only in the variant — control checkout is untouched (each change sits inside the variant branch of a ternary, behind v-if, or scoped to the toggle component, which only exists in the variant).

Layout (MP-3194)

Learn more alignment — the link sat higher than the tip copy beside it. The copy carried a top margin inside a centered flex row, and items-center counts margins, so the copy was pushed down. The margin now sits on the row and both children center together.
CTA arrow — dropped after "Learn more" in the variant, per design.
Upsell banner spacing — 8px sides / 24px vertical on desktop, with the container and loading placeholder at 128px so nothing shifts when the upsell loan loads in.
Design polish (MP-3203)

All matched against Laura's Figma CSS, using existing tokens only:

Borrower/donation names — the design system's Title style (tw-text-h4: 20px/26px on desktop) instead of tw-text-h3 with an !important weight override. Same spec, cleaner class.
Toggle label — tw-text-button-link (16px / 21px line-height, medium), matching the Figma "button" style. Color stays tw-text-gray-600, which is her exact #505050.
Toggle size — 48×28 with a 20px knob, per spec. KvSwitch hardcodes 56×32 with no size prop, so this is a scoped :deep override next to the existing gap override — structural selectors, token values, and one ! on the knob travel since the peer-checked rule outranks a scoped selector. A size prop on KvSwitch would be the long-term home for this.
Upsell background — tw-bg-secondary (#EDF4F1) instead of tw-bg-brand-100 (#EAF6F0), which read too bright.
Upsell progress meter — 6px via the meter's existing medium style, passed from outside the same way RecommendedLoanCard does. Was 8px.
Known gap, by design-system limitation

Laura's specs ask for font-weight 621. The site only loads PostGrotesk in book and medium, so the title and toggle label render at 500 — the heaviest real face. Getting her exact weight means adding a font to @kiva/kv-tokens; flagged to design as separate work.

Testing

No spec changes needed — presentation-only diff; full unit suite and lint pass.
To see it: a variant-b lender with a basket showing the toggle. Check the toggle at both states (knob should sit flush at both ends), the banner on desktop, and control against production — it should be identical.

validated with Laura and it looks as she expected.


src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -214,4 +214,19 @@ export default {
 :deep(label) {
 	@apply tw-gap-1;
 }
+
+/* The design toggle is 48x28 with a 20px knob; KvSwitch hardcodes 56x32 with no size prop.
```
mcstover: I don't think we should do this here. If we need a variant on the form toggle size, please add it to the toggle in kv-components.

src/components/Checkout/UpsellModule.vue
```diff
@@ -63,6 +63,7 @@
 						:is-funded="loan.status==='funded'"
 						:short-meter="!showTipFromBalanceVariant"
 						class="tw-mb-1"
+						:class="{ medium: showTipFromBalanceVariant }"
```
dyersituations, line 66: What is `medium`?

infante-jaime: It's the meter component's existing medium size, which sets the bar to 6px as per Laura.

src/components/Checkout/KivaCreditTipToggle.vue
infante-jaime: Removed it from UI and added it to `kv-components`. Just waiting for that to merge so we can update the version in UI.

dyersituations (approved):

roger-in-kiva (approved):

PR #7253, opened 2026-09-14
fix: mp-3206-say donation instead of tip in the variant copy

infante-jaime:
[MP-3206](https://kiva.atlassian.net/browse/MP-3206)

What this does

Swaps "tip" for "donation" in the two variant copy strings, per feedback Lauren collected — the rest of the site (control included) has always said "donation", and the new word was adding friction.

"your tip helps Kiva get it there" → "your donation helps Kiva get it there"
"Use my balance to cover this tip instead of future loans" → "Use my balance to cover this donation instead of future loans"
Both strings are variant-only; control copy is untouched.

Out of scope

Analytics labels and test ids containing "tip" (tip-from-balance-toggle-on/off, tip-donation-amount) stay as they are — renaming tracking mid-experiment would break data continuity.

Testing

Spec assertions updated to the new strings; full unit suite and lint pass.

image

[MP-3206]: https://kiva.atlassian.net/browse/MP-3206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


roger-in-kiva (approved):

christian14b (approved):

dyersituations (commented):
👍

PR #7256, opened 2026-09-15
feat: mp 3212 variant copy at zero tip

infante-jaime:
[MP-3212](https://kiva.atlassian.net/browse/MP-3212)

What this does

Keeps the variant copy and layout in place when a lender zeroes the tip — only the toggle and its label disappear. Previously the whole page flipped back to the control treatment mid-checkout (donation copy, compact upsell banner), which read as the UI changing its mind.

The audience criteria split in two: the variant treatment renders without the tip requirement, while the toggle and the exposure event keep the full criteria including the tip. No lender without a tip fires exposure in either arm, so the experiment read is unchanged.

Also bumps @kiva/kv-components to 10.4.1 (separate commit) to pick up the centered switch knob, which the wrapped mobile label was pushing off the track.

Testing

New page-level specs pin the zero-tip state; the old spec asserting the revert now asserts the opposite.
Verified by mutation: reverting either gate or re-adding the row's price check each fails tests.
QA note: at $0 the "donate loan repayments" prompt renders inside the kept variant layout (screenshots on the ticket) — new combination, verified on mobile.

image

image

[MP-3212]: https://kiva.atlassian.net/browse/MP-3212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Checkout/KivaCreditTipToggle.vue
```diff
@@ -36,26 +36,38 @@ const AUDIENCE_DEPOSIT_LIMIT = 1000;
...
 
+/**
+ * Whether the lender is in the experiment audience, ignoring which arm they are in. On top of
+ * the treatment criteria this needs a tip for the balance to cover: with none there is no
+ * decision to make, so neither arm renders the switch or fires exposure.
+ *
+ * @param {Object} state The basket state provided by the checkout page
+ * @returns {boolean} Whether the lender is in the experiment audience
+ */
+export function meetsTipFromBalanceCriteria(state = {}) {
```
dyersituations, line 67: This isn't being used?

infante-jaime: It is — both the variant switch and control exposure still use it, so the audience still requires a tip. The copy/layout no longer does, which is the intended fix: zeroing the tip keeps the treatment but doesn't fire exposure.

The export is now only used by the page spec to test the audience check against the basket state.

dyersituations (approved):

PR #7257, opened 2026-09-15
refactor: use categories for the loan spotlight

eddieferrer:
Ticket: https://kiva.atlassian.net/browse/CIT-5108

## Summary
- `allChannelsQuery` becomes `allCategoriesQuery`, moving from `lend.loanChannels` to the root `browsingCategories` field
- `spotlightLoanQuery` moves from `lend.loanChannelsById` to `categoriesByIds`, reading loans through `savedSearch(limit, pageNumber)` since categories page rather than offset
- The operation name `spotlightLoanQuery` is kept because `test/e2e/specs/CategoriesPage.cy.js` waits on `@gqlspotlightLoanQueryQuery`
- `$ids` is now `[String!]!`, so an unresolved slug skips the spotlight query rather than sending a null id
- Replaces `KvLoadingParagraph` with `KvLoadingText` from kv-components in the spotlight: the old component hardcodes 1rem lines, so the skeleton rendered taller than the text it stands in for and overflowed its own 5.5rem height

## Verification
- Both queries run against the dev gateway, exercising the resolved, fallback and unresolved slug paths
- The list query returns 32 browsing categories in roughly 850ms, against 100 channels in roughly 4000ms for the query it replaces
- Loading placeholder change verified on the dev server


src/components/Categories/LoanSpotlight.vue
```diff
@@ -73,69 +76,72 @@ const allChannelsQuery = gql`
...
+	// no category that matches the targeted name
+	if (targetedCategory.length === 0) {
+		// return id for fallback category
+		return fallbackCategory[0]?.id || null;
 	}
-	// targeted channel exists but no loans exist within it
-	if (targetedLoanChannel.length !== 0 && targetedLoanChannel[0].loans.totalCount === 0) {
-		return fallbackLoanChannel[0]?.id || null;
+	// targeted category exists but no loans exist within it
+	if (targetedCategory.length !== 0 && targetedCategory[0].savedSearch?.loans?.totalCount === 0) {
```
infante-jaime: This only catches `totalCount === 0`. Now that `savedSearch` can be `undefined`, we can skip the fallback and render an empty spotlight.

`if (!targetedCategory[0].savedSearch?.loans?.totalCount)` would cover both cases.

eddieferrer: Thanks, fixed

src/components/Categories/LoanSpotlight.vue
```diff
@@ -73,69 +76,72 @@ const allChannelsQuery = gql`
...
 		}
 	}
 `;
 
-function filterChannelsForRoute(routePath, loanChannels) {
-	const filteredChannels = loanChannels.filter(
-		loanChannel => loanChannel.url.split('/').pop() === routePath
+function filterCategoriesForRoute(routePath, categories) {
+	const filteredCategories = categories.filter(
+		category => category.url.split('/').pop() === routePath
```
mcstover, line 121: @eddieferrer @meganmcmillan Wondering if the forthcoming query to get category by slug will help use skip the "getAllCategories" query in this component. This seems quite inefficient at present to have to call for all categories when it's loaded...

dyersituations (approved):

mcstover (approved):
Probably be a little faster, but more expensive to query all categories than all loan channels but hopefully we can find a better way soon.

src/components/Categories/LoanSpotlight.vue
meganmcmillan: Yes, it looks like that will help here - we will provide a way to look up by the routePath/slug so you can get the one category you want.

meganmcmillan: https://kiva.atlassian.net/browse/CIT-5189

infante-jaime (approved):

eddieferrer:
> Probably be a little faster, but more expensive to query all categories than all loan channels but hopefully we can find a better way soon.

Its actually over 3x faster with the new query! I can just keep this open until we get the slug lookup.

PR #7255, opened 2026-09-14
feat(MP-3159): a/b test express vs non express checkout

christian14b:
https://kiva.atlassian.net/browse/MP-3159

PR #7260, opened 2026-09-15
feat: mp-3158 track the non-express checkout path from lend after goal set

roger-in-kiva:
## What

Adds two analytics events so we can tell when a lender reached checkout through the lend-after-goal-set feature rather than on their own.

| Page | Category | Action | Label |
|---|---|---|---|
| Checkout | `basket` | `view` | `from-lend-after-goal-set` |
| Thanks | `post-checkout` | `view` | `lend-from-after-goal-set` |

The label asymmetry (`from-lend-…` vs `lend-from-…`) is taken verbatim from the ticket.

## Why it needs a cookie

Both events fire on pages reached after the app has been left behind. `/basket` is not a route here: `router.push('/basket')` falls through to the catch-all, which does `window.location = to.fullPath`, the monolith resolves it to `/checkout`, and completing that order is another full page load to `/checkout/thanks`. Nothing in memory survives, so a cookie (`lend_after_goal_set`) carries the mark.

## Why the cookie holds a transaction id

The thanks page the lender leaves from and the one that ends the new order are the same component. `/checkout/thanks` matches `'checkout'` in `server/util/protectedRoutes.js`, so it is served `no-store`, which disables bfcache — pressing Back from checkout genuinely re-mounts the first thanks page and re-runs `onMounted`.

With a boolean, that Back press fired the event for the order the lender came *from* and consumed the cookie, so the order they actually completed was never counted. The cookie now stores the `kiva_transaction_id` it was marked on, and `isLendAfterGoalSetOrder` fires only for a different order, leaving the mark intact on a Back visit.

It also expires after an hour, so a lender who abandons checkout and never reaches a thanks page to clear it stops being credited.

## Changes

- `src/util/thanksPage/expressCheckoutUtils.js` — `LEND_AFTER_GOAL_SET_COOKIE` and the set / has / is-other-order / clear helpers.
- `src/composables/useExpressCheckoutModal.js` — `redirectToBasket()` folds the two basket redirects together and marks on the navigation, so a failed add-to-basket leaves nothing behind.
- `src/pages/Checkout/CheckoutPage.vue` — fires the basket view and deliberately does not clear it.
- `src/components/Thanks/ThanksPageSingleVersion.vue` — fires and clears, read above the goal-data awaits so a slow or failing load can't strand the mark.

## Scope

Thanks-page flow only, covering both branches that end in the basket: the express-disabled path and the non-empty-basket redirect. The MyKiva goal modal and Goals page already emit their own `go-to-checkout` events from MP-2748.

Note this is not strictly variant-A traffic once the A/B test is live. A variant-B lender whose basket already holds items is still redirected to the basket and marked, which is the intended scope: they reached checkout through the feature.

## Tests

23 added, all green.

- `expressCheckoutUtils.spec.js` — the cookie records the originating order and its one-hour expiry, falls back to a placeholder that stays truthy, credits only a different order, and survives a missing cookie store.
- `useExpressCheckoutModal.spec.js` — both redirect branches mark; the express modal path and a failed add do not.
- `CheckoutPage.spec.js` — fires with the mark, silent without, never clears.
- `ThanksPageSingleVersion.spec.js` — fires and clears; still fires when goal data never settles; stays silent on the order it was marked from.

Two pre-existing suite-wide flakes (`possibleTypesCacheIntegrity.spec.js`, `GoalInReviewModal.spec.js`) fail on a clean `main` worktree too and pass in isolation. Neither file is touched here.

## Known gaps, pre-existing and out of scope

The `lendingRewardOffered` redirect skips `CheckoutPage` before mount, and a `DONATION_ONLY_VIEW` / `LOGIN_REQUIRED_VIEW` thanks page never mounts the component.

[MP-3158](https://kiva.atlassian.net/browse/MP-3158) · epic [MP-2675](https://kiva.atlassian.net/browse/MP-2675) · A/B test [MP-3159](https://kiva.atlassian.net/browse/MP-3159) ([#7255](https://github.com/kiva/ui/pull/7255)) · prior tracking [MP-2748](https://kiva.atlassian.net/browse/MP-2748)

[MP-3158]: https://kiva.atlassian.net/browse/MP-3158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/thanksPage/expressCheckoutUtils.js
```diff
@@ -7,6 +8,18 @@ const DONATION_TYPENAME = 'Donation';
...
+ * Cookie set when the goal-set recommended loan sends the lender to the basket
+ * instead of the express checkout modal. The checkout and thanks pages read it to
+ * attribute their views to that flow.
+ *
+ * Its value is the order the lender was on when it was set, because the thanks page
+ * they leave from is the same component that later ends the order and Back can
+ * return to it. It expires after an hour so a lender who abandons checkout, and so
+ * never reaches a thanks page to clear it, stops being credited.
+ */
+export const LEND_AFTER_GOAL_SET_COOKIE = 'lend_after_goal_set';
```
christian14b, line 16: Can we move some of that explanation to the methods that apply this logic like expiring the cookie after an hour

dyersituations (approved):
Looks good pending some minor comment cleanup already commented on.

dyersituations (approved):

PR #7261, opened 2026-09-16
chore(header): delete the legacy default header branch

mcstover:
Deletes the unreachable default `<template v-else>` in `TheHeader.vue` and the state, methods, styles and imports only it used. The enclosing `<nav v-else>` renders only for `minimal` and `corporate`, both of which branch above it.

With that branch gone, `SearchBar.vue`, `Header/TeamsMenu.vue` and the `LendMenu/` directory have no importer left, so they go too, along with the GraphQL documents and the `searchEngine` util left with no consumer. `KvWwwHeaderBasic` ships its own search bar and lend menu.

`lendMenuData.graphql` and `countryList.graphql` are kept — `BailoutChips` and `MyKivaPage` still use them.

https://kiva.atlassian.net/browse/CIT-5028
https://kiva.atlassian.net/browse/CIT-5029


mcstover:
Going to add a little more cleanup/removal to this and test locally before opening up for review.

eddieferrer (approved):
🧹 !

dyersituations (approved):

mcstover:
Verified locally with and without login

PR #7265, opened 2026-09-16
feat: [MP-3208] - done show goal days left if there 90 days or less t…

cristhianDt:
# MP-3208

## Summary

Where

The goal card in the "Next steps recommended for you" carousel on MyKiva, and the same card on the /mykiva/next-steps page. This is the card with the progress ring, the goal title and the "Edit" button.

What is shown

A countdown text N days left, where N is the number of calendar days remaining in the goal year, counting the user's current local day through December 31 inclusive.

Singular copy on the last day: 1 day left. Never "1 days left".

Placement, typography and spacing per the Figma design. https://www.figma.com/design/ARJWLwdnFJ84FNSvPX4wYz/Goals-V2?node-id=9272-3809&m=dev

https://kiva.atlassian.net/browse/MP-3208

### Completed

Captura de pantalla 2026-09-16 a la(s) 12 03 34

### Start a new goal

https://github.com/user-attachments/assets/fde4dc2e-a82e-4288-87b9-3df06f7e0509

----

Captura de pantalla 2026-09-16 a la(s) 11 50 03
Captura de pantalla 2026-09-16 a la(s) 11 49 56
Captura de pantalla 2026-09-16 a la(s) 11 49 32
Captura de pantalla 2026-09-16 a la(s) 11 48 26
Captura de pantalla 2026-09-16 a la(s) 11 47 10
Captura de pantalla 2026-09-16 a la(s) 11 46 31
Captura de pantalla 2026-09-16 a la(s) 11 46 18
Captura de pantalla 2026-09-16 a la(s) 11 45 05


.storybook/stories/NextYearGoalCard.stories.js
```diff
@@ -6,14 +6,24 @@ export default {
...
 	const template = () => ({
 		components: { MyKivaNextYearGoalCard },
 		setup() {
+			// The card reads the QA date override from the URL after mount (as on MyKiva), so a story
+			// pins "today" by setting ?recapDate before mount. Storybook's own params are preserved.
+			const url = new URL(window.location.href);
+			if (recapDate) {
+				url.searchParams.set('recapDate', recapDate);
+			} else {
+				url.searchParams.delete('recapDate');
```
christian14b, line 23: If `recapDate` is missing, why do we need to delete it from the url?

src/composables/useGoalData.js
```diff
@@ -72,6 +72,32 @@ export const LAST_YEAR_KEY = GOALS_CURRENT_YEAR - 1;
...
+ * inclusive (Dec 31 → 1). Both ends are built as LOCAL midnights so the count stays whole across
+ * DST changes and never drifts to the UTC date. Returns null when the goal has no usable start
+ * date, its year is already over (local Jan 1 onward), or more than the countdown window remains.
+ *
+ * @param {Object|null} goal Goal with an ISO `dateStarted`; its calendar year is the goal year.
+ * @param {Date} [now] The effective "now". Pass getGoalInReviewNow() to honor the QA date
+ *   override; defaults to the browser clock.
+ * @returns {number|null} Days left (1..GOAL_YEAR_COUNTDOWN_WINDOW_DAYS), or null when hidden.
+ */
+export function getGoalYearDaysLeft(goal, now = new Date()) {
```
christian14b, line 90: I believe `date-fns` library has some method that can simplify this work or just use it directly. Probably `differenceInDays`?

src/components/MyKiva/NextYearGoalCard.vue
```diff
@@ -112,6 +114,21 @@ const completionAlreadyAnnounced = computed(() => Boolean(goalData?.hideGoalCard
 
 const userHasGoal = computed(() => !!props.userGoal && Object.keys(props.userGoal).length > 0);
 
+// The countdown is browser-only: the server knows neither the user's timezone nor the
+// ?recapDate QA override, so computing it during SSR would hydrate to different copy.
+// Waiting for mount renders nothing on the server and fills the value in after hydration.
+const hasMounted = ref(false);
+onMounted(() => {
+	hasMounted.value = true;
```
dyersituations: Seems like this could be simplified by just setting a `ref` on mount, rather than a `computed`

cristhianDt: good question, we need to use `computed` for cases where the user edits the goal, we must to update the ProgressRing, hide or show, the copy after edit the goal, so with `ref` we are not available to refresh the state if the goal change

.storybook/stories/NextYearGoalCard.stories.js
cristhianDt: its job is clearing a stale pin from the previous story, for example:

1. open UserGoalDaysLeft → the factory writes recapDate=2026-11-15 into the iframe URL.
2. click UserGoalWithProgress (passes no recapDate) → the param is still there from step 1.
3. without the delete, the card reads it after mount and shows 47 days left! on a story that must not have a countdown

make sense?

src/components/MyKiva/NextYearGoalCard.vue
dyersituations: But the goal is related to the year, so editing won't affect days left.

src/composables/useGoalData.js
cristhianDt: good call!, changed

src/components/MyKiva/NextYearGoalCard.vue
cristhianDt: well, just in case, when the goal changed to completed, but you're right, it's almost impossible, changing...

cristhianDt: okay, I've retested some cases, and there's one where, when the user sets the goal, because NextYearGoalCard.vue is already mounted, the value (using `ref`) won't change, so the copy won't appear when the card is updated. I moved the logic to GoalProgressRing.vue to cover this + any other cases and following your recommendation, lmk any other comment, thanks!

src/components/MyKiva/GoalProgressRing.vue
```diff
@@ -225,6 +248,25 @@ const progressCircleDesc = computed(() => {
...
 const isModalVariant = computed(() => props.variant === 'modal');
 
+// Days left in the goal year, resolved once in the browser. A single read after mount is
+// enough: the card renders this ring only while a goal exists (so mounting coincides with the
+// goal appearing), the goal year never changes while mounted, and completion is gated
+// reactively in showDaysLeft. It is never computed during SSR, which knows neither the
+// user's timezone nor the ?recapDate QA override.
+const daysLeft = ref(null);
+onMounted(() => {
+	daysLeft.value = getGoalYearDaysLeft({ dateStarted: props.goalDateStarted }, getGoalInReviewNow());
```
infante-jaime, line 258: Optional, but I think this logic fits better in NextYearGoalCard. The ring really just needs daysLeft.

Could we move it back to a computed in the parent and pass daysLeft down? A computed will react when the goal is set, so it avoids the issue the ref had.

dyersituations (approved):

src/components/MyKiva/GoalProgressRing.vue
cristhianDt: agree, I dont have a problem to revert the last commit, @dyersituations just to confirm is ok if we leave the computed in the parent?

dyersituations: Whatever makes the most sense!

christian14b (approved):

src/components/MyKiva/GoalProgressRing.vue
cristhianDt: both versions work well so leaving as it is for now, thanks team!

PR #7271, opened 2026-09-17
feat: add fundraisers entry point to MyKiva

eddieferrer:
Ticket: [CIT-5212](https://kiva.atlassian.net/browse/CIT-5212)

## Summary

- Adds a fundraisers section to MyKiva between "Next steps recommended for you" and "Impact progress", with two states depending on whether the lender owns a fund.
- Lenders with no fund see a carousel of occasion cards deep-linking into `/gf/configure?occasion=<id>`, plus a tooltip and a closing card pointing at `/lp/event-fundraiser`. Card copy is hardcoded; the images reuse the Contentful assets already serving that landing page.
- Lenders who own funds see a card per fund — title, avatar, participants, dollars raised and lives touched — scrolling in a carousel, alongside a card to start another.
- Card titles resolve through the same fallback chain the giving fund page uses, so a lender sees one name for their fund in both places. That chain is reimplemented in `getGivingFundTitle` rather than shared, so adding an occasion in cms-page-server without mirroring it here would let the two surfaces disagree.
- Owned funds come from the page's existing `myKivaQuery` prefetch rather than a component-level fetch, retiring `myGivingFunds.graphql` and its unused fields.
- Removes the "Check in on your giving funds" row, superseded by the new section's Manage All link. That link keys off owning a fund where the old row keyed off participation, so lenders who have only contributed to someone else's fund now reach `/gfm` from `/portfolio` instead. The card and its `/portfolio` use are untouched.

[CIT-5212]: https://kiva.atlassian.net/browse/CIT-5212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2026-09-17 at 2 54 18 PM
Screenshot 2026-09-17 at 2 54 24 PM
Screenshot 2026-09-17 at 3 04 18 PM


eddieferrer:
This PR also bumps up the giving fund query to prefetch. Initially the idea was to avoid duplicate queries since it was powering 2 rows related to giving funds. Now that it is only powering this 1 row we can go back to a client side fetch if we feel that expanding the mykiva query will slow the page down too much

src/composables/useGivingFund.js
```diff
@@ -1,7 +1,6 @@
 import logFormatter from '#src/util/logFormatter';
 // eslint-disable-next-line max-len
 import myGivingFundParticipationDonations from '#src/graphql/query/portfolio/myGivingFundParticipationDonations.graphql';
-import myGivingFundsQuery from '#src/graphql/query/portfolio/myGivingFunds.graphql';
```
mcstover, line 4: Is this not still in use on the portfolio page?

eddieferrer: Its not. The portfolio card uses these from the composable: https://github.com/kiva/ui/blob/119439839b419f9b7edfc2d8a8112bf9e4871269/src/components/GivingFunds/MyGivingFundsCard.vue#L67-L70
 which fire 2 smaller queries here: 
https://github.com/kiva/ui/blob/main/src/composables/useGivingFund.js#L4-L5

src/components/MyKiva/YourFundraisersSection.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+	<div>
+		<div class="tw-flex tw-items-center tw-gap-1 tw-mb-2">
```
mcstover, line 3: This block may need some bottom margin adjustments for mobile when the carousel controls get close to the headline and manage link...

mcstover (approved):

PR #7274, opened 2026-09-18
feat: gate the MyKiva fundraisers row behind an experiment

eddieferrer:
Ticket: [CIT-5265](https://kiva.atlassian.net/browse/CIT-5265)

## Summary

- Product wants to know whether the fundraisers row added in CIT-5212 draws engagement away from the rest of MyKiva — lending, goals, achievements — rather than adding to it. The row now renders only for version `b` of `mykiva_fundraisers_row`.
- Follows the client-side experiment pattern this page already runs twice for `mykiva_goals_row` and `mykiva_co_recovery_fund`: the assignment is prefetched alongside them, `initializeExperiment` resolves the version from cookie or cache in `created()` so the row does not pop in, and the flag gates the whole `#mykiva-fundraisers` section — both the occasion carousel and the owner row.
- The flag defaults to `false`, so the row is hidden until assignment returns `b`. `uiexp.mykiva_fundraisers_row` and the `active_experiments` entry are live in both dev and prod, at a 50/50 split across the full population.

[CIT-5265]: https://kiva.atlassian.net/browse/CIT-5265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dyersituations (approved):
While I know this isn't exactly a B2C exp, could you add an entry here for tracking?

https://kiva.atlassian.net/wiki/spaces/B2C/pages/3591274497/Active+Experiments+Features+Emails

mcstover (approved):

PR #7277, opened 2026-09-18
fix: ensure goal summary query is not fired if the user does not meet…

christian14b:
… the conditions to see goal in recap

https://kiva.atlassian.net/browse/MP-3236

## What this fixes

`loadAutoOpenRecap` loads the whole recap and *then* checks whether the pop-up is allowed to appear. So MyKiva and the Portfolio - Impact Dashboard load it on every visit where it was never going to open:

- the goal ran in a previous year
- the goal expired
- this is the visit that announces the completion (that one is for the confetti)
- the goal is still in progress, before its release date

That data is expensive to build for lenders with a lot of loans, and we threw it away without showing it.

## The solution

All but one of those rules can be answered from the user preferences the function already loads a line earlier and was ignoring. Those checks now run before the load.

The exception is whether the lender has actually lent anything, which needs the data. That check stays where it was.

`shouldAutoOpenRecap` is split in two: `canAutoOpenRecapBeforeLoad` holds everything except that last check, and the full function calls it, so the two can't drift apart.

No change to who sees the recap or when. It picks the same goal the same way, and the two fields it now reads early (status and start date) come back from the server unchanged.


src/composables/useGoalInReview.js
```diff
@@ -283,20 +286,26 @@ export default function useGoalInReview({ apollo, goalData } = {}) {
...
-		const data = await loadGoalInReview({ year });
-		const goalYear = new Date(data?.goalSummary?.dateStarted).getFullYear();
-
-		const shouldOpen = shouldAutoOpenRecap({
+		// The status and year the decision needs are already in the preferences just loaded,
+		// so visits that cannot pop up are turned away before the payload is paid for.
+		const goal = findMostRecentActiveGoal(parsedPrefs?.goals ?? []);
+		const decision = {
 			enabled,
-			isEligible: Boolean(data?.isEligible),
```
dyersituations, line 291: We no longer need to check `isEligible`?

src/util/goalInReview.js
```diff
@@ -50,11 +48,10 @@ function hasInProgressReleaseStarted(startDate, now) {
  * @param {Date|string|null} options.inProgressStartDate The goal_in_review_in_progress_start
  *   setting, the date in-progress goal setters become eligible.
  * @param {Date} options.now The effective current date.
- * @returns {boolean} Whether to open the recap automatically.
+ * @returns {boolean} Whether everything but the goal's own progress allows the pop-up.
  */
-export function shouldAutoOpenRecap({
+export function canAutoOpenRecapBeforeLoad({
```
dyersituations: This method name is confusing, since the modal doesn't actually open before loading data, right?

src/util/goalInReview.js
```diff
@@ -87,6 +84,19 @@ export function shouldAutoOpenRecap({
...
+ * Decides whether the recap should open by itself, for MyKiva and Portfolio alike.
+ * Both pages call this so the pop-up happens once per user across the two, rather
+ * than once per page.
+ *
+ * @param {object} options Trigger inputs; canAutoOpenRecapBeforeLoad documents the rest.
+ * @param {boolean} options.isEligible Whether the recap has a goal with progress.
+ * @returns {boolean} Whether to open the recap automatically.
+ */
+export function shouldAutoOpenRecap({ isEligible = false, ...rest } = {}) {
+	return Boolean(isEligible) && canAutoOpenRecapBeforeLoad(rest);
```
dyersituations: `canAutoOpenRecapBeforeLoad` is already called before this method is called.

src/composables/useGoalInReview.js
christian14b: It's still checked, I'll make the change to make it visible in both calls rather than hidden in the object. Good catch that it read as deleted

src/util/goalInReview.js
christian14b: Agreed, the function is gone. I'll rename it

src/composables/useGoalInReview.js
```diff
@@ -283,20 +286,26 @@ export default function useGoalInReview({ apollo, goalData } = {}) {
 			return null;
 		}
 
-		const data = await loadGoalInReview({ year });
-		const goalYear = new Date(data?.goalSummary?.dateStarted).getFullYear();
-
-		const shouldOpen = shouldAutoOpenRecap({
+		// The status and year the decision needs are already in the preferences just loaded,
```
tiarascoleman: Recommending this comment also to be rephrased

src/composables/useGoalInReview.js
```diff
@@ -283,20 +286,26 @@ export default function useGoalInReview({ apollo, goalData } = {}) {
...
+		// so visits that cannot pop up are turned away before the payload is paid for.
+		const goal = findMostRecentActiveGoal(parsedPrefs?.goals ?? []);
+		const decision = {
```
tiarascoleman: Maybe rename to payloadDecision?

src/util/goalInReview.js
christian14b: You're right... that was pointles, the second pass could only differ on eligibility. Removed the split entirely; `goalInReview.js` is back to untouched. Now it's rules once before the load, then a plain if (!data?.isEligible) after

src/composables/useGoalInReview.js
```diff
@@ -283,22 +285,25 @@ export default function useGoalInReview({ apollo, goalData } = {}) {
...
 			now,
-		});
+		};
+		if (!shouldAutoOpenRecap({ ...autoOpenRules, isEligible: true })) {
+			return null;
+		}
 
-		if (!shouldOpen) {
+		const data = await loadGoalInReview({ year });
+		if (!data?.isEligible) {
```
dyersituations, line 306: How can we always pass `isEligible: true` above, but use `data?.isEligible` here?

christian14b: Yeah, that is confusing. I dropped `isEligible` from `shouldAutoOpenRecap`, and the progress check now happens after the load

src/composables/useGoalInReview.js
```diff
@@ -283,22 +285,25 @@ export default function useGoalInReview({ apollo, goalData } = {}) {
 			return null;
 		}
 
-		const data = await loadGoalInReview({ year });
-		const goalYear = new Date(data?.goalSummary?.dateStarted).getFullYear();
-
-		const shouldOpen = shouldAutoOpenRecap({
+		// These rules need nothing but the preferences just loaded, so visits that cannot pop
+		// up are turned away before the payload is paid for.
```
dyersituations: What does this comment mean?

dyersituations (approved):
Thanks for the comment rewrites! Makes sense now!

PR #7269, opened 2026-09-17
feat: mp-3209 keep completed goal at front until its goal year ends

roger-in-kiva:
## Summary

On the MyKiva "Impact progress" row, a completed annual goal now stays in the front slide for the rest of its goal year. Before this change, the tile moved to the end of the row the instant the goal was completed. Now it only moves back, alongside the prior-year goals, once the user's local clock crosses into January 1 of the following year.

The row's front/back rule in `BadgesSectionV2.vue` changes from "is the goal achieved" to "is the goal achieved, and has its year not yet ended." The year check reuses `getIsPastGoalYear`/`getGoalInReviewCurrentYear`, the same helpers MP-3208's goal-card countdown already uses, so both features honor the `?recapDate=YYYY-MM-DD` QA override together.

## Approach notes

- The achieved check stays in the condition on purpose: an in-progress goal should never move to the back regardless of date, and keeping that check means the new year-boundary logic can only ever affect an *achieved* goal, not an in-progress one.
- No changes to `useGoalData.js`'s goal-renewal or `completedGoalsHistory` logic. Those already remove a prior-year completed goal from `userGoal` on the real January 1, so this row never shows one goal twice; the new code only changes where an achieved goal sits before that happens.
- The moved tile keeps its normal "active goal" formatting rather than switching to the historical-tile format; only its position in the row changes, matching the ticket's "no visual change" note.
- Badge ordering and prior-year goal ordering are untouched: only the one `if` that decides the goal tile's position changed.

## Related links

- Jira story: [MP-3209](https://kiva.atlassian.net/browse/MP-3209), keep a completed goal at the front until the goal year ends
- Epic: [MP-3199](https://kiva.atlassian.net/browse/MP-3199), Goals mykiva updates
- Related, shares the "goal year ends" clock definition: [MP-3208](https://kiva.atlassian.net/browse/MP-3208) / #7265, goal-card countdown
- Downstream: [MP-3210](https://kiva.atlassian.net/browse/MP-3210) (QA functional verification), [MP-3211](https://kiva.atlassian.net/browse/MP-3211) (VQA)

## Related tests

`test/unit/specs/components/MyKiva/BadgesSectionV2.spec.js`, new `describe('completed goal placement across the goal year')`:

- keeps a completed goal at the front through December 31 of its goal year
- moves a completed goal to the end once its goal year has ended
- keeps an in-progress goal at the front even after its start year has ended
[...]

PR #7278, opened 2026-09-21
fix: mp-3209 make the recapDate override reach the server so hydration matches

roger-in-kiva:
## Summary

QA hit three things on a production build at `/mykiva?recapDate=2027-01-01`: the impact progress row's order jumped after load, the first card rendered with no icon, and the console showed `Hydration completed but contains mismatches` followed by `TypeError: Cannot read properties of null (reading 'nodeType')` inside `__asyncHydrate`.

`getGoalInReviewNow()` read the `?recapDate` override from `window.location.search`, which the server has no access to. So anything that let the override decide its markup rendered one thing on the server and hydrated another. On the row that meant a different tile order, and since the row's icons are `defineAsyncComponent`s, reordering it mid-hydration stranded one of them on a detached node. It threw, and that tile lost its icon.

The override now comes off the route, which is populated during SSR: `vue-middleware` passes `req.url` into `router.push`, and the render waits on `router.isReady()`. Server and client resolve the same date on the first render, so the order no longer changes under hydration.

## What changed

- `useRecapDateOverride()` reads the override off the route. `getGoalInReviewNow` and `getGoalInReviewCurrentYear` take it as an optional argument, defaulting to the old address-bar behavior so untouched callers keep working.
- `useHideGoalSignup()` owns the sign up ask's window for the four surfaces that offer it. That block was copy-pasted in all four, and two of them had it wrong.
- `useGoalRecapEntryPoint` resolves the year it asks `hasViewedRecap` about. Its two callers each resolved one of their own and had drifted apart: `MyKivaFeaturedSlot` was answering from a module-load-time constant that ignored the override entirely.
- `ImpactDashboardPage` was a live instance of the same bug, gating `<GoalEntrypoint v-if>` on a bare `getGoalInReviewNow()` on a server rendered page. It threads the override by hand rather than through the composable, because its start date lives in Options API `data`, which `setup()` cannot reach.

The address-bar fallback stays for the reads that only run after user interaction (`loadAutoOpenRecap`, `GoalInReviewModal`, `GoalProgressRing`'s post-mount read). I verified each of those never renders on the server.

## Notes for review

- Production behavior for real lenders is unchanged. Nobody passes `recapDate`, so both sides resolve the real clock exactly as before.
- `/mykiva` and `/portfolio/lending-stats` are auth-required and not CDN-cached, so letting a query param influence SSR output carries no cache risk. Only `/ui-site-map` is CDN-cached.
- `useRecapDateOverride` optional-chains the router on purpose. An unguarded `useRouter()` returns undefined with no router provided and broke 41 existing tests; this carries a QA param only, so a router-less mount reads no override instead of failing.
- An earlier attempt deferred the override read to `onMounted`. It made things worse by guaranteeing a reorder right as the icon chunks were resolving, which is what crashed them, so it was dropped.

## Related links

- Jira story: [MP-3209](https://kiva.atlassian.net/browse/MP-3209), keep a completed goal at the front until the goal year ends
- Epic: [MP-3199](https://kiva.atlassian.net/browse/MP-3199), Goals mykiva updates
- The PR this follows up on: #7269
- Shares the "goal year ends" clock: [MP-3208](https://kiva.atlassian.net/browse/MP-3208) / #7265, the goal card countdown
- Downstream: [MP-3210](https://kiva.atlassian.net/browse/MP-3210) (QA functional verification), [MP-3211](https://kiva.atlassian.net/browse/MP-3211) (VQA)

## Related tests

74 tests across the five files, all passing. The two SSR specs fail on `main`; I confirmed that by reverting the fix.

`test/unit/specs/components/MyKiva/BadgesSectionV2.server.spec.js` (new, 3 tests, node environment):
- places an achieved goal ahead of the badges while its year runs
- honors the recapDate override so the client hydrates the same order
- asserts `typeof window === 'undefined'`, so the SSR specs cannot slip into being DOM specs that pass for the wrong reason

`test/unit/specs/components/MyKiva/MyKivaFeaturedSlot.server.spec.js` (2 added, 7 total):
- hides the signup ask when the override lands inside the window
- leaves the signup ask up when the override lands outside the window

`test/unit/specs/composables/useGoalInReview.spec.js` (7 added, 36 total):
- takes the override as an argument, for callers that also run on the server
- prefers the argument over the address bar
- `useRecapDateOverride`: reads it off the route, reads none when the route carries none, reads none rather than failing with no router
- `useGoalRecapEntryPoint`: asks about the year it resolved itself, asks about the year the route overrides to, falls back to the address bar with no route override

`test/unit/specs/pages/Portfolio/ImpactDashboard/ImpactDashboardPage.spec.js` (3 added, 8 total):
- hides the ask when the route override lands inside the window
- leaves the ask up once the override passes the goal year
- falls back to the clock with no override

`test/unit/specs/components/MyKiva/BadgesSectionV2.spec.js` (20 total): the placement tests now drive the route rather than only the address bar.

Repo-wide: 5531 passed, 6 skipped. The one failure is `possibleTypesCacheIntegrity.spec.js`, a pre-existing suite-order flake that passes in isolation. `GoalInReviewModal.spec.js` is a second pre-existing flake, failing only under parallel file load because its seven async imports exceed Testing Library's default 1000ms timeout; it passes with `--no-file-parallelism` and with this branch's files removed.

## Dev verification

Needs a production build, since dev never hydrates. The dev server leaves the container empty and Vue full-mounts instead, which is why none of this reproduces there.

```
npm run build
npm run start -- --config=dev-custom-host
```

- `/mykiva?recapDate=2026-12-31` with a completed current-year goal: goal tile first, no console errors
- `/mykiva?recapDate=2027-01-01`: goal tile last, ahead of the older prior-year goals, icon intact, no hydration warning
- `/mykiva` with no param: unchanged

## Known follow-up, worth a ticket before the New Year

The override was one of two triggers for this crash. The row's order still depends on client-local time resolved separately per side: `setGoalState` uses `new Date().getFullYear()`, which is the server's timezone during SSR, and `GOALS_CURRENT_YEAR` is module scope, so on the server it freezes at worker boot.

A lender in `America/Los_Angeles` at 2026-12-31 20:00 local against a UTC server can get the goal tile missing from the server HTML but first on the client: same detached-node crash, plus a tile that vanishes. The window is about a day a year, centered on exactly the transition MP-3209 exists to handle. Hardening `KvIcon`'s async hydration would cover the whole class rather than this one trigger.

[MP-3209]: https://kiva.atlassian.net/browse/MP-3209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/composables/useGoalInReview.js
```diff
@@ -99,6 +146,9 @@ export function useGoalRecapEntryPoint({
 	hasViewedRecap,
 	loansTowardGoal = 0,
 }) {
+	const { recapDate } = useRecapDateOverride();
+	const now = () => getGoalInReviewNow(recapDate.value);
```
dyersituations: I think `now` isn't descriptive enough

src/components/MyKiva/BadgesSectionV2.vue
```diff
@@ -107,19 +107,31 @@ const userHasGoal = computed(() => !!userGoal.value && Object.keys(userGoal.valu
...
-	currentYear: getGoalInReviewCurrentYear(),
-	loansTowardGoal,
-	activeGoalYear: userGoalYear.value,
-	now: getGoalInReviewNow(),
-});
+// The server needs this override too, since it sets the row's tile order. The carousel's icons
+// hydrate asynchronously, and reordering the row mid-hydration strands one of them on a
+// detached node, which throws and leaves that tile with no icon.
+const { recapDate } = useRecapDateOverride();
+const resolveNow = () => getGoalInReviewNow(recapDate.value);
```
dyersituations: Similar, same value, but this one reads more like a method.

I wonder if something like `resolveGoalInReviewNow`

src/composables/useGoalInReview.js
roger-in-kiva: Agreed, and it was worse than just vague: this and the one in BadgesSectionV2 were the same one-line helper under two different names. Both are `resolveGoalInReviewNow` now, in f54e61380.

src/components/MyKiva/BadgesSectionV2.vue
roger-in-kiva: Took your suggestion, `resolveGoalInReviewNow` in both files so they match and read against the `getGoalInReviewNow` they wrap. f54e61380.

dyersituations (approved):

